PR #6, opened 2022-05-04
refactor(controller): add candidateBundles type for getting latest Bundle

tylerslaton:
Starting with a small change to address a `TODO` comment about breaking some logic out into its own function. Also cleaning up some code generation.

### DIFF
--- a/controllers/platformoperator_controller.go
+++ b/controllers/platformoperator_controller.go
@@ -19,7 +19,6 @@ package controllers
 import (
 	"context"
 	"fmt"
-	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


controllers/platformoperator_controller.go
```diff
@@ -54,6 +53,8 @@ type PlatformOperatorReconciler struct {
 	RegistryClient registryClient.Interface
 }
 
+type candidateBundles []*api.Bundle
```
timflannagan, line 56: Do we want to make this a struct that stores an array of api.Bundles?

timflannagan (approved):

PR #9, opened 2022-05-09
Add BundleInstance Generation

### LINKED ISSUE #1: Update implementation to use generate and managed BundleInstance resources
The current implementation generates and manages an individual Bundle and BundleInstance resource for each PlatformOperator custom resource, but with [embedded Bundles](https://github.com/operator-framework/rukpak/pull/293) on the horizon, we should update the implementation to generate and manage BundleInstance resources, and let the plain provisioner handle the rest.

### LINKED ISSUE #2: Add support for generating BundleInstance resources
Goal: update the current implementation which only generates a Bundle resource to also generate a BundleInstance that matches that corresponding Bundle's metadata.Name.

When a new platform-operator Bundle version is available (e.g. it has the highest semver), the controller is responsible for generating a new Bundle resource, and updating the existing BundleInstance resource to point to that newly generated Bundle's metadata.Name.

### LINKED ISSUE #5: Add support for pivoting to new Bundle resources when a new version is available
Goal: ensure that new Bundles are being generated when a new version for a PlatformOperator becomes available (either through a file-based catalog, or an API that builds on top of file-based catalogs like the CatalogSource API).

tylerslaton:
Closes #5
Closes #2
Closes #1

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,6 +39,10 @@ help: ## Display this help.
 
 ##@ Development
 
+.PHONY: manifests
+manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
+	$(CONTROLLER_GEN) rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
+
 .PHONY: generate
 generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
 	$(CONTROLLER_GEN) crd:crdVersions=v1 output:crd:artifacts:config=config/crd/bases paths=./api/...
@@ -63,10 +67,6 @@ test: manifests generate fmt vet envtest ## Run tests.
 build: generate fmt vet ## Build manager binary.
 	go build -o bin/manager main.go
 
-.PHONY: run
-run: manifests generate fmt vet ## Run a controller from your host.
-	go run ./main.go
-
 .PHONY: docker-build
 docker-build: test ## Build docker image with the manager.
 	docker build -t ${IMG} .
@@ -81,6 +81,15 @@ ifndef ignore-not-found
   ignore-not-found = false
 endif
 
+.PHONY: demo
+# NOTE: This will fail as the currently available version of RukPak (v0.3.0) does not have 
+#       the requisite code.
+demo: run install-samples 
+
+.PHONY: kind-load
+kind-load: docker-build
+	kind load docker-image $(IMG)
+
 .PHONY: install
 install: manifests kustomize ## Install CRDs into the K8s cluster specified in ~/.kube/config.
 	$(KUSTOMIZE) build config/crd | kubectl apply -f -
@@ -90,13 +99,19 @@ uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified 
 	$(KUSTOMIZE) build config/crd | kubectl delete --ignore-not-found=$(ignore-not-found) -f -
 
 .PHONY: deploy
-deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
+deploy: manifests rukpak kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
 	$(KUSTOMIZE) build config/default | kubectl apply -f -
 
 .PHONY: undeploy
 undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/config. Call with ignore-not-found=true to ignore resource not found errors during deletion.
 	$(KUSTOMIZE) build config/default | kubectl delete --ignore-not-found=$(ignore-not-found) -f -
+
+run: kind-load install deploy ## Install and deploy all of the necessary resources for platform-operators
+
+.PHONY: install-samples
+install-samples:  ## Install the sample manifests found in config/samples/manifests
+	kubectl apply -f config/samples
 
 CONTROLLER_GEN = $(shell pwd)/bin/controller-gen
 .PHONY: controller-gen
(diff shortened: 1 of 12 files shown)


Makefile
```diff
@@ -39,6 +39,10 @@ help: ## Display this help.
 
 ##@ Development
 
+.PHONY: manifests
+manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
+	$(CONTROLLER_GEN) rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
```
timflannagan, line 44: I think we can remove this target entirely as there's overlap with the `generate` target. I was under the impression that the issue here was that I had consolidated the generate/manifest targets into a single target, but forgot to remove any target that specifies the `manifests` target as a prequisite?

Makefile
```diff
@@ -63,10 +67,6 @@ test: manifests generate fmt vet envtest ## Run tests.
 build: generate fmt vet ## Build manager binary.
 	go build -o bin/manager main.go
 
-.PHONY: run
-run: manifests generate fmt vet ## Run a controller from your host.
-	go run ./main.go
```
timflannagan, line 68: I'm personally a fan of this target, but I won't hold up these changes because of this.

controllers/platformoperator_controller.go
```diff
@@ -99,24 +125,22 @@ func (r *PlatformOperatorReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
-	log.Info("finding the bundles that contain the highest semver")
-
 	latestBundle, err := cb.latest()
 	if err != nil || latestBundle == nil {
 		log.Info("failed to find the bundle with the highest semver")
 		return ctrl.Result{RequeueAfter: 5 * time.Second}, err
 	}
 
-	log.Info("candidate bundles", "highest semver", latestBundle.Version)
+	log.Info("candidate bundle", "latest "+latestBundle.CsvName+" version", latestBundle.Version)
```
timflannagan, line 134:
```suggestion
	log.Info("candidate bundle", "latest", latestBundle.CsvName, "version", latestBundle.Version)
```

controllers/platformoperator_controller.go
```diff
@@ -125,47 +149,57 @@ func (r *PlatformOperatorReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
+		if err := controllerutil.SetOwnerReference(po, bi, r.Scheme); err != nil {
+			return err
 		}
-		if err := controllerutil.SetOwnerReference(po, bundle, r.Scheme); err != nil {
+		if err := r.createBundleInstance(ctx, bundleName, bundleImage); err != nil {
 			return err
 		}
-		if err := r.Create(ctx, bundle); err != nil {
+		po.Status.InstalledBundleInstanceName = bi.Name
+	} else if po.Status.InstalledBundleInstanceName != bundleName {
```
timflannagan, line 166: nit: return early in the if conditional to avoid having to use an if/else conditional structure?

controllers/platformoperator_controller.go
```diff
@@ -125,47 +149,57 @@ func (r *PlatformOperatorReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	return ctrl.Result{}, nil
 }
 
-func (r *PlatformOperatorReconciler) ensureBundle(ctx context.Context, po *platformopenshiftiov1alpha1.PlatformOperator, bundleName, bundleImage string) error {
-	if po.Status.InstalledBundleName != "" {
-		return nil
-	}
-	bundle := &rukpakv1alpha1.Bundle{}
-	if err := r.Get(ctx, types.NamespacedName{Name: bundleName}, bundle); err != nil {
+func (r *PlatformOperatorReconciler) ensureBundleInstance(ctx context.Context, po *platformopenshiftiov1alpha1.PlatformOperator, bundle *api.Bundle) error {
```
timflannagan, line 152: I think we can implement a CreateOrUpdate BundleInstance helper function that mirrors rukpak's CreateOrRecreate logic. This can be done as a follow-up though.

controllers/platformoperator_controller.go
```diff
@@ -125,47 +149,57 @@ func (r *PlatformOperatorReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
 }
 
-// SetupWithManager sets up the controller with the Manager.
-func (r *PlatformOperatorReconciler) SetupWithManager(mgr ctrl.Manager) error {
-	return ctrl.NewControllerManagedBy(mgr).
-		For(&platformopenshiftiov1alpha1.PlatformOperator{}).
-		Watches(&source.Kind{Type: &operatorsv1alpha1.CatalogSource{}}, handler.EnqueueRequestsFromMapFunc(requeuePlatformOperators(mgr.GetClient()))).
-		Complete(r)
+// updateBundleInstance is responsible for taking a BundleInstance and pivoting to a new Bundle
+func (r *PlatformOperatorReconciler) updateBundleInstance(ctx context.Context, bi *rukpakv1alpha1.BundleInstance, name, image string) error {
```
timflannagan, line 178: I'd inline this logic into the `r.ensureBundleInstance(...)` method but this is fine for now.

Makefile
tylerslaton: Mainly overwrote it to foster [functionality similar](https://github.com/operator-framework/rukpak/blob/77c1be1d734f82a0c07ca4b7893ac83f1033aa5b/Makefile#L107) to `RukPak`. Maybe we can keep this target as is and change `make run` => `make start`?

timflannagan: I don't have a strong opinion but I thought the `run` command was automatically scaffolded out for us when kubebuilder initialized this project?

controllers/platformoperator_controller.go
tylerslaton: Let's handle that in a follow-up for now.

tylerslaton: https://github.com/timflannagan/platform-operators/issues/10

timflannagan (approved):

controllers/platformoperator_controller.go
```diff
@@ -50,8 +50,37 @@ type PlatformOperatorReconciler struct {
 	RegistryClient registryClient.Interface
 }
 
+// SetupWithManager sets up the controller with the Manager.
+func (r *PlatformOperatorReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&platformopenshiftiov1alpha1.PlatformOperator{}).
+		Watches(&source.Kind{Type: &operatorsv1alpha1.CatalogSource{}}, handler.EnqueueRequestsFromMapFunc(requeuePlatformOperators(mgr.GetClient()))).
+		// TODO: Only apply for PlatformOperators related to this BundleInstance's OwnerReference
+		Watches(&source.Kind{Type: &rukpakv1alpha1.BundleInstance{}}, handler.EnqueueRequestsFromMapFunc(requeuePlatformOperators(mgr.GetClient()))).
```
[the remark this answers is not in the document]
tylerslaton: For sure.

PR #11, opened 2022-05-11
fix: improving pivoting logic for when a catalogsource gets updated

### LINKED ISSUE #10: Add CreateOrUpdate helper function for ensureBundleInstance function.
We should implement a CreateOrUpdate BundleInstance helper function that mirrors rukpak's [CreateOrRecreate logic](https://github.com/operator-framework/rukpak/blob/67cc29d77ed062bd97a3fa92dda6654d734c8981/internal/util/util.go#L181). This would be added into the `ensureBundleInstance` function inside of the `platformoperator_controller`.

tylerslaton:
Closes #10


api/v1alpha1/platformoperator_types.go
```diff
@@ -27,8 +27,14 @@ type PlatformOperatorSpec struct {
 
 // PlatformOperatorStatus defines the observed state of PlatformOperator
 type PlatformOperatorStatus struct {
-	InstalledBundleName         string `json:"installedBundleName"`
-	InstalledBundleInstanceName string `json:"installedBundleInstanceName"`
+	InstalledBundle         InstalledBundle `json:"installedPackage"`
+	InstalledBundleInstance string          `json:"installedBundleInstance"`
```
timflannagan, line 31: I'm not sure either of these fields are needed at this point. I think my expectation is that we'll generated a BundleInstance where the metdata.Name matches the parent PlatformOperator's metadata.Name. And I don't think we need to track an installed Bundle field because that information is already present in the BI's status once the plain provisioner starts reconciling that resource.

timflannagan:
Another thing I realize: we have this CatalogSource custom resource sample, but when I was originally bootstrapping this controller I was using the magic catalog CLI as it's easier to quickly check whether you're pivoting logic works as intended without having to deal with rebuilding the FBC index image, and waiting for the next registry poll.

tylerslaton:
@timflannagan 
> I think we can drop the the status fields, and only have a status field for conditions
So these fields are currently used to determine if the version we've installed with a BundleInstance is different than the most recently reconciled package with the highest semver in the catalog. Should we instead move to just always update the BundleInstance, even if the information is the same? The effect is probably the same but with a few wasted cycles.

> We can update the r.ensureBundleInstance method to simply wrap the controllerutil.CreateOrRecreate function to ensure that resource matches our desired specification
That sounds doable, let me pencil around with this locally and see what I can do. 

> Another thing I realize: we have this CatalogSource custom resource sample, but when I was originally bootstrapping this controller I was using the magic catalog CLI as it's easier to quickly check whether you're pivoting logic works as intended without having to deal with rebuilding the FBC index image, and waiting for the next registry poll.

I'll have to look into that as I haven't used it before, myself. Shouldn't be too bad and I'll take a look at its documentation.

timflannagan:
> Should we instead move to just always update the BundleInstance, even if the information is the same

I think we'd have to always perform this work of calculating what's the most recent registry bundle from that platform-operator's package unless I'm missing something obvious. Whether or not we update that BundleInstance is another story as that CreateOrUpdate logic is responsible for checking whether the current and desired client.Object(s) are semantically equal.

tylerslaton:
Gotcha, this is mainly just my lack knowledge about that function then. I'll mess around with it locally and come up with some new code.

tylerslaton:
@timflannagan Alright I got this all setup as you said. I think we should do a follow-up where we handle status-management, what do you think?

controllers/platformoperator_controller.go
```diff
@@ -149,55 +148,33 @@ func (r *PlatformOperatorReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
-							Ref: image,
-						}}}}},
-	}
+func buildBundleInstance(name, image string) *rukpakv1alpha1.BundleInstanceSpec {
+	return &rukpakv1alpha1.BundleInstanceSpec{
+		ProvisionerClassName: "core.rukpak.io/plain",
+		Template: &rukpakv1alpha1.BundleTemplate{
+			ObjectMeta: metav1.ObjectMeta{
+				Labels: map[string]string{"app": name},
+			},
```
timflannagan, line 170: We can technically drop this metadata as it's unclear what kind of labels/annotations we'll need to propagate right now.

timflannagan (approved):

PR #14, opened 2022-05-12
controllers: Add support for watching BundleInstance resources

timflannagan:


timflannagan:
Oh right I didn't enable actions for this repository :laughing: 

This should just work :tm: but I did manually test this by creating a BI, and then editing that BI and adding another label and watches the container logs to ensure that the owning PlatformOperator custom resource got successfully requeued.

PR #16, opened 2022-05-25
*: Refactor the platform operator's controller implementation

timflannagan:


api/v1alpha1/platformoperator_types.go
```diff
@@ -22,9 +22,11 @@ import (
 
 // PlatformOperatorSpec defines the desired state of PlatformOperator
 type PlatformOperatorSpec struct {
-	// PackageName specifies the name of the package to be installed from the provided CatalogSource.
-	// PackageName is required and must equal the exact name of the package in the catalog.
-	PackageName string `json:"packageName"`
+	// Package specifies the name of the package to be installed from the provided CatalogSource.
+	// Package is required and must equal the exact name of the package in the catalog.
+	Package string `json:"package"`
```
exdx, line 27: This can be in a separate PR, but I was thinking maybe we want to add additional kubebuilder markers for these fields. For example, even though they are required, `""` is still technically a valid value for these fields and would pass validation. Maybe we could add something like 

```
+kubebuilder:validation:MinLength=1
```

to ensure that some kind of value must be provided? (Go treats the length of `""` as zero per https://go.dev/play/p/63fGgUxvFOm)

internal/sourcer/source.go
```diff
@@ -0,0 +1,89 @@
...
+		client: c,
+	}
+}
+
+func (cs catalogSource) Source(ctx context.Context, po *v1alpha1.PlatformOperator) (*Bundle, error) {
+	bundles, err := getBundles(ctx, cs.client, po)
+	if err != nil {
+		return nil, err
+	}
+	filteredBundle, err := filterBundles(bundles, func(b1, b2 *Bundle) bool {
```
exdx, line 38: Could we add a `len(bundles)==0` check here and return early in that case? Otherwise I believe we end up with a nil pointer as the desiredBundle

main.go
```diff
@@ -93,20 +95,21 @@ func main() {
 
 	// Create a registryClient by referencing the requisite catalogSource's service and then
 	// check that it is healthy
-	c, err := registryClient.NewClient(catalogSourceService)
+	c, err := registryClient.NewClient("localhost:50051")
```
exdx, line 98: nit: pull this variable into a const?

exdx (approved):

api/v1alpha1/platformoperator_types.go
```diff
@@ -22,9 +22,11 @@ import (
...
+	// Channel specifies the current channel that's being used.
+	Channel string `json:"channel"`
```
tylerslaton, line 29: Do we want a user to be able to specify what channel a PO is watching on? Maybe we make it optional and have a logical default of `platform`.

controllers/platformoperator_controller.go
```diff
@@ -96,128 +59,32 @@ func (r *PlatformOperatorReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	log.Info("reconciling request", "req", req.NamespacedName)
 	defer log.Info("finished reconciling request", "req", req.NamespacedName)
 
+	// TODO: flesh out status condition management
```
tylerslaton, line 62: This still needs to get through #13

internal/applier/applier.go
```diff
@@ -0,0 +1,65 @@
...
+	rukpakv1alpha1 "github.com/operator-framework/rukpak/api/v1alpha1"
+	"github.com/timflannagan/platform-operators/api/v1alpha1"
+	"github.com/timflannagan/platform-operators/internal/sourcer"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+const (
+	plainProvisionerID    = "core.rukpak.io/plain"
```
tylerslaton, line 15: Maybe we could pull these [constants](https://github.com/operator-framework/rukpak/blob/main/internal/provisioner/plain/types/types.go) in from Rukpak?

internal/sourcer/source.go
```diff
@@ -0,0 +1,89 @@
...
+			Image:    b.GetBundlePath(),
+			Skips:    b.GetSkips(),
+			Replaces: b.GetReplaces(),
+		})
+	}
+	return cb, nil
+}
+
+func filterBundles(bundles []Bundle, f func(b1, b2 *Bundle) bool) (*Bundle, error) {
+	// TODO(tflannag): Respect skips/replaces semantics?
```
tylerslaton, line 79: We'll want some form of graph traversal in place but it can be a follow-up.

tylerslaton (commented):
Nice, I like the implementation and have a few thoughts.

internal/applier/applier.go
timflannagan: The rukpak constants are defined in internal packages, so we'd either need to promote those packages to pkg/ or manually vendor them for now.

exdx:
I think there's still a lot of value in this PR, even if we don't necessarily want the proposed spec change (to make channel configurable, since it seems we will still be looking at one channel (or a `platform-*` pattern of channels) for the specific use-case). The interfaces are really nice and help minimize the complexity quite a bit. Could we get this through without the spec changes?

timflannagan:
@exdx I can revive this then. I was playing around with another implementation locally that I think we can massage into this source interface, and mainly closed this as it was created before I added a CI pipeline to this repository.

PR #21, opened 2022-05-26
Ensure the platform operator stack is deployed before running e2e tests

timflannagan:
Home for any commits that are needed in order to get a basic e2e testing suite introduced.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM gcr.io/distroless/static:debug
+FROM gcr.io/distroless/static:debug-nonroot
 WORKDIR /
 
 COPY manager manager
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,27 +1,5 @@
-# Build the manager binary
-FROM golang:1.17 as builder
-
-WORKDIR /workspace
-# Copy the Go Modules manifests
-COPY go.mod go.mod
-COPY go.sum go.sum
-# cache deps before building and copying source so that we don't need to re-download as much
-# and so that source changes don't invalidate our downloaded layer
-RUN go mod download
-
-# Copy the go source
-COPY main.go main.go
-COPY api/ api/
-COPY controllers/ controllers/
-
-# Build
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -1,27 +1,5 @@
...
-COPY api/ api/
-COPY controllers/ controllers/
-
-# Build
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -o manager main.go
-
-# Use distroless as minimal base image to package the manager binary
-# Refer to https://github.com/GoogleContainerTools/distroless for more details
-FROM gcr.io/distroless/static:nonroot
+FROM gcr.io/distroless/static:debug-nonroot
```
timflannagan, line 1: Update from the debug tag that matched the rukpak container image to debug-nonroot as I ran into the following error locally:

```
container has runasnonroot and image will run as root
```

And that's because kubebuilder/operator-sdk automatically scaffold out that securityContext for our manager resources by default (and we didn't use that SDK tooling in rukpak).

controllers/platformoperator_controller.go
```diff
@@ -102,8 +105,27 @@ func (r *PlatformOperatorReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
+	if err := r.List(ctx, css); err != nil {
+		log.Error(err, "failed to list the catalogsource resources in the cluster")
+		return ctrl.Result{}, err
+	}
+	if len(css.Items) == 0 {
+		log.Info("unable to query catalog content as no catalogsources are available")
+		return ctrl.Result{}, nil
+	}
+	// TODO(tflannag): properly handle multiple catalogsources in a cluster
+	cs := css.Items[0]
```
timflannagan, line 118: This is a super short term hack that's needed in order to work with the basic e2e tests that I have sitting locally, which deploy a magic catalog instance, so hardcoding the CatalogSource reference, and even worse hardcode that logic at the main.go level as it fails the container at startup, wouldn't work with that setup.

PR #24, opened 2022-05-27
*: Refactor controller implementation

timflannagan:

### DIFF
--- a/controllers/platformoperator_controller.go
+++ b/controllers/platformoperator_controller.go
@@ -18,69 +18,28 @@ package controllers
 
 import (
 	"context"
-	"fmt"
 
-	"github.com/blang/semver/v4"
 	operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
-	"github.com/operator-framework/operator-registry/pkg/api"
-	registryClient "github.com/operator-framework/operator-registry/pkg/client"
 	rukpakv1alpha1 "github.com/operator-framework/rukpak/api/v1alpha1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/types"
-	utilerror "k8s.io/apimachinery/pkg/util/errors"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	logr "sigs.k8s.io/controller-runtime/pkg/log"
-	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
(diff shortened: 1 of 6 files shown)


controllers/platformoperator_controller.go
```diff
@@ -100,162 +59,32 @@ func (r *PlatformOperatorReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
-			cb = append(cb, b)
-		}
-	}
-	if len(errors) != 0 {
-		agg := utilerror.NewAggregate(errors)
-		log.Error(agg, "failed to build catalog snapshot")
-		return ctrl.Result{}, agg
+	desiredBundle, err := r.Sourcer.Source(ctx, po)
+	if err != nil {
+		log.Error(err, "failed to source content for platform operator")
```
exdx, line 69: nit: should we include the name of the PO in a formatted string in the error log for more visibility? 
```
log.Error(err, "failed to source content for platform operator", po.GetName())
```

controllers/platformoperator_controller.go
```diff
@@ -100,162 +59,32 @@ func (r *PlatformOperatorReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
-	if len(cb) == 0 {
-		log.Info("failed to find any candidate bundles")
-		return ctrl.Result{}, nil
-	}
-
-	latestBundle, err := cb.latest()
-	if err != nil || latestBundle == nil {
-		log.Info("failed to find the bundle with the highest semver")
+	if err := r.Applier.Apply(ctx, po, desiredBundle); err != nil {
+		log.Error(err, "failed to apply the desired bundle")
```
exdx, line 73: nit: same here, maybe more context for the user?
```
log.Error(err, "failed to apply the desired bundle", desiredBundle.GetName(), "for platform operator", po.GetName())

File: internal/sourcer/source.go [hunk 3]
```diff
@@ -0,0 +1,105 @@
...
+
+type Bundle struct {
+	Version  string
+	Image    string
+	Replaces string
+	Skips    []string
+}
+
+func (b Bundle) String() string {
+	return fmt.Sprintf("Version: %s; Image: %s; Repliaces; %s", b.Version, b.Image, b.Replaces)
```
exdx, line 27:
```suggestion
	return fmt.Sprintf("Version: %s; Image: %s; Replaces; %s", b.Version, b.Image, b.Replaces)
```

exdx (approved):
LGTM. Having these interfaces here so early on will make iterating a lot nicer.

controllers/platformoperator_controller.go
timflannagan: We should already have this information from the logger given we create a logger from the Reconcile ctx provided to us: `log := logr.FromContext(ctx)`.

controllers/platformoperator_controller.go
timflannagan: We should already have this information from the logger given we create a logger from the Reconcile ctx provided to us: `log := logr.FromContext(ctx)`.

internal/sourcer/source.go
timflannagan: Whoops - this was a debug artifact, but I'll keep it in anyways. Updated.

PR #37, opened 2022-06-06
Add channel configuration

exdx:
Adds the ability to configure the channel the PO controller uses when
querying catalogs at runtime via an environmental variable CHANNEL_NAME.


exdx:
Note: I'm working on getting a kustomize patch in for the makefile as well such that something like 
```
make run CHANNEL=beta
```
ends up being respected in the PO deployment

exdx:
Closing for now -- I think we can pass the channel information in the form of a constraint on the deppy side.

PR #46, opened 2022-10-25
*: Sync downstream o/platform-operators changes

timflannagan:
Update the v0 fork with the current downstream o/platform-operators changes.


timflannagan:
Merging as fixing the e2e is going to take some work. Need to distinguish between upstream/downstream kustomize overlay?

PR #47, opened 2022-10-25
*: Introduce the Operator API

timflannagan:
Builds on top of #46 and introduces a barebones Operator API and a controller watching that new API. The PO controller implementation has been refactored to maintain an Operator resource under-the-hood, and delegates the process of sourcing/applying registry+v1 bundles to the Operator provisioner. 

The Operator provisioner uses the same implementation as the PO controller with some slight adaptations to the underlying function signatures, which may point to a brittle implementation that could be shored up over time.

Note: There's no support yet for bubbling up underlying Operator resource status back to the the PO resource that's managing it. The PO API is also defined in o/api which means that the currently po.Status.ActiveBundleDeployment field is misnamed and will need some love once the implementation is solidified.


timflannagan:
I'm also not entirely sold on this PlatformOperator <-> Operator API 1:1 mapping right now. It feels like the correct API surface is "I want this Operator to be installed as a platform operator" long term.

PR #48, opened 2022-10-29
*: Implement support for reconciling Operator version constraints

timflannagan:
Pushing up local changes that I had staged. Updates the Operator API and introduces an optional spec.package.version string field. When this field has been unspecified, the expected behavior is that the highest semver bundle that meets the package constraints will be returned.

The spec.package.version field supports a user specifying either an explicit version (e.g. "2.0.1") or a version range (e.g. "<2.0.0 || >=3.0.0"). When reconciling this value, we attempt to first check whether an explicit version has been specified through a simple regex, and then fall back to parsing a version range.

### DIFF
--- a/api/v1alpha1/operator_types.go
+++ b/api/v1alpha1/operator_types.go
@@ -32,7 +32,8 @@ type CatalogSpec struct {
 }
 
 type PackageSpec struct {
-	Name string `json:"name"`
+	Name    string `json:"name"`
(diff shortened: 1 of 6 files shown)


timflannagan:
@joelanford this is basically what I had in mind for a super, super rough implementation after our conversation during demos. Just pushing up what I had hacked on friday afternoon.

PR #50, opened 2022-11-02
*: Add initial support for upgrading an operator after installation

timflannagan:
During reconciliation, the operator controller will attempt to generate a BundleDeployment under-the-hood that will manage the registry+v1 bundle that was selected for installation. In the case that this BundleDeployment already exists, the current implementation avoids updating that resource, or attempt to query the catalog(s) in the cluster to find whether a new registry+v1 bundle is available that meets the desired package name configuration.

That behavior was a short stop-gap to ensure the barebones framework was in place as continuously firing off gRPC registry requests per reconciliation attempt can be expensive over time. And this behavior is also amplified given the controller places a watch on the CatalogSource API, and that API is known to have some inconsistencies with it's state transitions. A notable limitation with this approach is that it opened the door for users to modify the underlying BundleDeployment given the reconciliation loop didn't have logic that compared the current and desired state of the generated resource.

These changes implement initial support for upgrading an operator after initial installation. The reconciliation loop has removed the safeguards in place that were in place to prevent unnecessary work, and now bundles are sourced every reconciliation. While the functionality is there, this approach has some clear drawbacks, and implementing a catalog cache should help smooth out the performance hit, while making the state transitions for the CatalogSource API more predictable.

With these changes, the controller now accurately compares the current and desired state of the generated BundleDeployment resource attached to the operator  resource being reconciled. When a user modifies this resource, the controller will revert those changes back to the desired state.

After an operator has been installed on the cluster and a new bundle version is available in the configured catalog, then the controller will attempt to get the highest semver bundle that meets the desired package name constraint. That highest semver bundle metadata will be propagated to the generated BundleDeployment, which will result in a pivot being performed by rukpak's core provisioners.

> Note: there's some ambiguity and open questions in this domain around what's the desired behavior we're looking for long term.

While these changes add the ability to simply upgrade operators, there's some limitations in the current behavior:

- The controller doesn't respect any upgrade graph semantics defined for the selected package.
- There's no way to delineate between "this operator is failing at runtime" or an operator signaling "don't upgrade me right now, I'm in the middle of a migration" which influences our upgrade model.
- Introduces the potential for implicitly downgrading/rolling back an operator version given there's no catalog cache, and a catalog could revert back to an older version of an operator.
- Introduces the potential for non-deterministic behavior. In the case that an operator doesn't configure a spec.catalog configuration, then the current behavior is to source a bundle that meets the package name criteria from any catalogs available in the cluster. This opens the door for another catalog (i.e. the one that was selected during installation) to be updated with the highest semver bundle that matches the configured package name specified, and that package may defined different upgrade graph semantics than the originally chosen one.

A follow-up implementation may start to address the want/need for a catalog cache. This should help make the implementation a bit more robust, and help inform some of the open questions in this domain.


internal/applier/bundledeployment_applier.go
```diff
@@ -1,35 +1,50 @@
...
-	bd.SetOwnerReferences([]metav1.OwnerReference{*controllerRef})
 
-	bd.Spec = buildBundleDeployment(image)
-	return bd
+	labels := map[string]string{
+		"core.olm.io/package":          o.Spec.Package.Name,
+		"core.olm.io/version":          source.Version,
+		"core.olm.io/source-name":      source.SourceInfo.Name,
+		"core.olm.io/source-namespace": source.SourceInfo.Namespace,
+	}
```
timflannagan, line 30: This propagates this relevant metadata returned by the sourcing layer, and while it's nice to have this auxiliary data available for visibility and to fuel list queries, we really should be bubbling this up to the operator's status.

api/v1alpha1/operator_types.go
```diff
@@ -45,6 +45,18 @@ type OperatorStatus struct {
 	// then it means the Operator has either not been installed yet or is failing to install.
 	// +optional
 	ActiveBundleDeployment ActiveBundleDeployment `json:"activeBundleDeployment,omitempty"`
+
+	// sourceInfo is the reference to the previously sourced catalog data that's being
+	// installed by this Operator resource.
+	SourceInfo SourceInfo `json:"sourceInfo,omitempty"`
```
timflannagan, line 51: This might introduce some potential coupling in the internal implementations as my current plan is to leverage this information if it's already present to avoid cross-catalog sourcing after an operator has been installed.

timflannagan:
This is what the status looks like now when persisted on the cluster:

```yaml
...
spec:
  package:
    name: prometheus-operator
status:
  activeBundleDeployment:
    name: prometheus-operator
  conditions:
  - lastTransitionTime: "2022-11-02T20:15:54Z"
    message: Successfully applied the prometheus-operator BundleDeployment resource
    reason: InstallSuccessful
    status: "True"
    type: Installed
  sourceInfo:
    name: magiccatalog
    namespace: default
    version: 0.2.0
```

This seems reasonable for a first pass, but something we should aim to re-evaluate over time.

timflannagan:
Added a custom printer column for that sourceInfo integration:

```bash
$ k get operators.core.olm.io 
NAME                  ACTIVE BUNDLEDEPLOYMENT   INSTALL STATE       INSTALLED VERSION   AGE
prometheus-operator   prometheus-operator       InstallSuccessful   0.2.0               47m
```

This was useful for local dev purposes, but I imagine it's would be useful for admins anyways.

timflannagan:
I was also going to add an optional `spec.profile` field to the API surface that could be used to configure how the operator handles failure modes and non-deterministic situations. The only profiles here would be "BestEffort" and "Deterministic" and we'd be able to leverage this internally during the sourcing layer to handle situations where 1/3 catalogs are currently offline. Ended up moving those changes into it's own branch in the meantime.

timflannagan:
List of open questions I had recorded locally too:

```go
// Open Questions:
// 1. Once a catalog is initially selected, continue to use that catalog during upgrades?
// 2. Use a catalog cache that refreshes periodically with some eviction policy to
//    avoid firing gRPC registry requests every reconciliation loop. What does this cache
//    look like? How is it implemented?
// 3. How do failure modes work when sourcing catalog contents? Is it okay if a single
//    catalog fails to successfully list the bundles in the catalog when another catalog
//    could meet the desired configuration? Is this behavior something we'd need to surface
//    in the API?
// 4. Do we care about cross-channel upgrades right now? Do we care about catalog priority?
// 5. How do we best track the currently installed version, and the desired version?
// 6. When following upgrade graph semantics, do we need to give time for migrations to occur
//    or are we simply running through the path?
// 7. How do we effectively delineate between BundleDeployment runtime health/status checks, and an
//    operator being able to express "don't upgrade me right now"? How does this influence pivoting
//    behavior, and is there a way to implement this robustly without introducing new complexity?
// 8. What does a resolveset look like? Is it a single YAML stream of all the operators + dependencies
//    that need to exist on the cluster, or would it be a subset of [operator + it's dependencies] N
//    times? For the former, are there any scaling implications when using local storage like a ConfigMap?
//    Do we need to implement the storage layer in a more robust/scalable way?
// 9. We currently place a watch on the CatalogSource API, which is renown for having weird state transistions.
//    How does this influence a reconciliation model where we attempt to source a bundle to accommodate upgrade
//    situations given this is already a potentially expensive process, while ensuring we don't reconcile
//    transient catalog source events that could affect performance? Does the combination of a catalog cache
//    and a separate goroutine help smooth this issue out?
// 10. During upgrade paths where an operator resource doesn't specify a catalog configuration, what's the
//     desired behavior here? There's potential for multiple catalogs on the cluster to share the same desired
//     package name. After the initial installation, would the right behavior be ensuring that the same package
//     is being used to follow the defined upgrade graph semantics?
// 11. What happens when a catalog has been updated and reverted back to a previous catalog? In the case that
//     an operator has been installed at v0.3.0, and the new catalog only has a v0.1.0 bundle available in the
//     desired package name, the current implementation would roll back to that version. Is this behavior something
//     we have any desired of supporting? Note: this is where a catalog cache would be useful.
```

PR #56, opened 2022-11-08
internal,test: Ensure cross-catalog upgrades are prohibited

timflannagan:
Update the CatalogSource registry implementation, and ensure that an Operator tracks the catalog source that was initially selected during installation when no spec.Catalog configuration has been specified by the user.

In the case that multiple catalogs exist on the cluster (e.g. c1, c2, ..., cN) and a subset of those catalogs all contain the same package name, then the expected behavior is that one of those catalogs that contains the highest semver bundle will be selected during installation. After installation, a different catalog has been updated and now contains the highest semver bundle in the desired package name. We'd like to prohibit an Operator upgrading to another catalog source in this case to avoid the scenario where packages within those different catalogs define different upgrade graphs.

Note: We may want to do something similar with cross-channel updates, but it's unclear whether this cross-catalog UX is something worth enforcing as a long term behavior.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,7 +72,7 @@ e2e: deploy test-e2e
 .PHONY: test-e2e
 FOCUS := $(if $(TEST),-v -focus "$(TEST)")
 JUNIT_REPORT := $(if $(ARTIFACT_DIR), -output-dir $(ARTIFACT_DIR) -junit-report junit_e2e.xml)
-test-e2e: ginkgo ## Run e2e tests
+test-e2e: ginkgo ## Run e2e tests.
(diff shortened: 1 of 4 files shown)


timflannagan:
```
Using the kubectl kubectl binary
Using the artifacts/should-eventually-pivot-to-the-most-recent-version output directory
Storing the test artifact output in the artifacts/should-eventually-pivot-to-the-most-recent-version directory
Collecting get operators.core.olm.io -o yaml output...
W1108 16:32:47.169751   24758 loader.go:223] Config not found: ~/kube/.config
```

Whoops :facepalm:

PR #57, opened 2022-11-08
*: Ensure testing artifacts are gathered when running in CI

timflannagan:


timflannagan:
Nice, looks like that finally worked. Removing the debug commit.

PR #81, opened 2023-01-04
Adding e2e make target, makefile adjustments, and e2e test framework

dtfranz:
Adds the `e2e` target for `make` along with the associated github action. Also adjusts `run` to be more in-line with what we do in RukPak, running the `operator-controller` inside a kind cluster instead of the kubebuilder generated default behavior (run locally)
To keep the changes to a minimum I've not fully adjusted the makefile to match RukPak's, so there are many functional and stylistic differences which we may want to resolve in the near future.

Signed-off-by: dtfranz <dfranz@redhat.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,15 @@
-
+###########################
+# Configuration Variables #
+###########################
 # Image URL to use all building/pushing image targets
-IMG ?= controller:latest
+export IMAGE_REPO ?= quay.io/operator-framework/operator-controller
+export IMAGE_TAG ?= latest
+export GO_BUILD_TAGS ?= upstream
+IMG?=$(IMAGE_REPO):$(IMAGE_TAG)
+
+OPERATOR_CONTROLLER_NAMESPACE ?= operator-controller-system
+KIND_CLUSTER_NAME ?= operator-controller
+
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
 ENVTEST_K8S_VERSION = 1.25.0
 
@@ -54,9 +63,26 @@ fmt: ## Run go fmt against code.
 vet: ## Run go vet against code.
 	go vet ./...
 
-.PHONY: test
-test: manifests generate fmt vet envtest ## Run tests.
+.PHONY: test test-e2e e2e kind-load kind-cluster kind-cluster-cleanup
+test: manifests generate fmt vet envtest test-e2e ## Run all tests.
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test ./... -coverprofile cover.out
+
+FOCUS := $(if $(TEST),-v -focus "$(TEST)")
+E2E_FLAGS ?= ""
+test-e2e: ginkgo ## Run the e2e tests
+	$(GINKGO) --tags $(GO_BUILD_TAGS) $(E2E_FLAGS) -trace -progress $(FOCUS) test/e2e
+
+e2e: run test-e2e kind-cluster-cleanup ## Run e2e test suite on local kind cluster
+
+kind-load: kind ## Loads the currently constructed image onto the cluster
+	$(KIND) load docker-image $(IMG) --name $(KIND_CLUSTER_NAME)
+
+kind-cluster: kind kind-cluster-cleanup ## Standup a kind cluster
+	$(KIND) create cluster --name ${KIND_CLUSTER_NAME}
+	$(KIND) export kubeconfig --name ${KIND_CLUSTER_NAME}
+
+kind-cluster-cleanup: kind ## Delete the kind cluster
+	$(KIND) delete cluster --name ${KIND_CLUSTER_NAME}
 
 ##@ Build
 
@@ -65,14 +91,17 @@ build: manifests generate fmt vet ## Build manager binary.
 	go build -o bin/manager main.go
 
 .PHONY: run
-run: manifests generate fmt vet ## Run a controller from your host.
-	go run ./main.go
+run: docker-build kind-cluster kind-load install deploy wait ## Build the operator-controller then deploy it into a new kind cluster.
+
+.PHONY: wait
+wait:
+	kubectl wait --for=condition=Available --namespace=$(OPERATOR_CONTROLLER_NAMESPACE) deployment/operator-controller-controller-manager --timeout=60s
 
 # If you wish built the manager image targeting other platforms you can use the --platform flag.
 # (i.e. docker build --platform linux/arm64 ). However, you must enable docker buildKit for it.
 # More info: https://docs.docker.com/develop/develop-images/build_enhancements/
 .PHONY: docker-build
-docker-build: test ## Build docker image with the manager.
+docker-build: generate ## Build docker image with the operator-controller.
 	docker build -t ${IMG} .
 
 .PHONY: docker-push
@@ -129,11 +158,23 @@ $(LOCALBIN):
 ## Tool Binaries
 KUSTOMIZE ?= $(LOCALBIN)/kustomize
 CONTROLLER_GEN ?= $(LOCALBIN)/controller-gen
+KIND ?= $(LOCALBIN)/kind
+GINKGO ?= $(LOCALBIN)/ginkgo
 ENVTEST ?= $(LOCALBIN)/setup-envtest
 
 ## Tool Versions
 KUSTOMIZE_VERSION ?= v4.5.7
 CONTROLLER_TOOLS_VERSION ?= v0.10.0
+
+.PHONY: kind
+kind: $(KIND) ## Download kind locally if necessary.
+$(KIND): $(LOCALBIN)
+	test -s $(LOCALBIN)/kind || GOBIN=$(LOCALBIN) go install sigs.k8s.io/kind@v0.15.0
+
+.PHONY: ginkgo
+ginkgo: $(GINKGO) ## Download ginkgo locally if necessary.
+$(GINKGO): $(LOCALBIN)
+	test -s $(LOCALBIN)/ginkgo || GOBIN=$(LOCALBIN) go install github.com/onsi/ginkgo/v2/ginkgo@v2.1.4
 
 KUSTOMIZE_INSTALL_SCRIPT ?= "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh"
 .PHONY: kustomize
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@
 ###########################
 # Image URL to use all building/pushing image targets
 export IMAGE_REPO ?= quay.io/operator-framework/operator-controller
-export IMAGE_TAG ?= latest
+export IMAGE_TAG ?= devel
 export GO_BUILD_TAGS ?= upstream
 IMG?=$(IMAGE_REPO):$(IMAGE_TAG)
 
@@ -72,6 +72,7 @@ E2E_FLAGS ?= ""
 test-e2e: ginkgo ## Run the e2e tests
 	$(GINKGO) --tags $(GO_BUILD_TAGS) $(E2E_FLAGS) -trace -progress $(FOCUS) test/e2e
 
+e2e: KIND_CLUSTER_NAME=operator-controller-e2e
 e2e: run test-e2e kind-cluster-cleanup ## Run e2e test suite on local kind cluster
 
 kind-load: kind ## Loads the currently constructed image onto the cluster
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,9 @@ endif
 SHELL = /usr/bin/env bash -o pipefail
 .SHELLFLAGS = -ec
 
+# Disable -j flag for make
+.NOTPARALLEL:
+
 .PHONY: all
 all: build
 
--- a/config/crd/bases/operators.operatorframework.io_operators.yaml
+++ b/config/crd/bases/operators.operatorframework.io_operators.yaml
@@ -13,7 +13,7 @@ spec:
     listKind: OperatorList
     plural: operators
     singular: operator
-  scope: Cluster
+  scope: Namespaced
   versions:
   - name: v1alpha1
     schema:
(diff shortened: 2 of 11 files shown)


test/e2e/e2e_suite_test.go
```diff
@@ -0,0 +1,61 @@
...
+	Expect(err).To(BeNil())
+
+	err = corev1.AddToScheme(scheme)
+	Expect(err).To(BeNil())
+
+	err = appsv1.AddToScheme(scheme)
+	Expect(err).To(BeNil())
+
+	err = apiextensionsv1.AddToScheme(scheme)
+	Expect(err).To(BeNil())
```
joelanford, line 54: At this point, the only thing we need to add to the scheme is the operatorv1alpha1 API since that's the only API we interact with during the test. If/when we need the other APIs, we can add them later.

test/e2e/e2e_suite_test.go
```diff
@@ -0,0 +1,61 @@
...
+
+	c, err = client.New(cfg, client.Options{Scheme: scheme})
+	Expect(err).To(BeNil())
+
+	kubeClient, err = kubernetes.NewForConfig(cfg)
+	Expect(err).To(BeNil())
```
joelanford, line 60: Don't think we need `kubeClient` yet either. I'd remove and let a future PR re-introduce if we need it.

test/e2e/reconcile_test.go
```diff
@@ -0,0 +1,44 @@
...
+	"context"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	operatorv1alpha1 "github.com/operator-framework/operator-controller/api/v1alpha1"
+)
+
+var _ = Describe("Basic Reconciliation", func() {
```
joelanford, line 13: Mainly a terminology nit...

Basic reconciliation tests might be better suited for more unit-style tests in the controllers package where we call `Reconcile` once per `It` and can fine tune the scenarios we want to cover. I always like to point to this package as an example: https://github.com/operator-framework/helm-operator-plugins/tree/main/pkg/reconciler

For example of the difference:
- A unit test could pre-create an `Operator` and use a "fake" deppy entity source, then run `Reconcile` once to check that the expected `BundleDeployment` gets created and the `Operator` status reflects the new state. Or with a different fake entity source, it could check that a specific error is applied to the `Operator` status if deppy can't resolve.
- An e2e would start an entire Operator controller pod, create an `Operator` and then check that, eventually, the requested operator is actually running.

The e2e would require rukpak to be installed, running, and reconciling. And it would need to wait for rukpak to unpack the bundle, apply it, check its status, etc. 

Both are valuable, but my suggestion would be that we put more focus and energy into the faster unit tests and only add e2es when a unit test seems insufficient.

TL;DR: Maybe reword the e2e `Describes` to align with this somehow? And then (in a follow-up) do this same sort of thing in the `controllers` package, but with an example `It` that actually calls `Reconcile` instead of relying on the controller to do it with its event handler.

joelanford (commented):
Nice! I'm a big fan of getting this setup early so that there's a place for contributors to fairly easily add tests.

One other thing I think we should try to do early on is encourage more unit testing than we did in rukpak for the core functionality. Right now, much of rukpak's functional tests are in e2e's. In e2e's we can't (easily) get coverage metrics, so it's difficult to ensure we're covering critical sections in tests, e2e's generally take an order of magnitude more time to run (making dev cycles take longer way longer), and it's basically impossible to test individual state transitions.

Makefile
```diff
@@ -1,6 +1,15 @@
-
+###########################
+# Configuration Variables #
+###########################
 # Image URL to use all building/pushing image targets
-IMG ?= controller:latest
+export IMAGE_REPO ?= quay.io/operator-framework/operator-controller
+export IMAGE_TAG ?= latest
```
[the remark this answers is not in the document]
dtfranz: Of course! Good suggestion.

Makefile
```diff
@@ -54,25 +63,45 @@ fmt: ## Run go fmt against code.
...
+.PHONY: test test-e2e e2e kind-load kind-cluster kind-cluster-cleanup
+test: manifests generate fmt vet envtest test-e2e ## Run all tests.
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test ./... -coverprofile cover.out
 
+FOCUS := $(if $(TEST),-v -focus "$(TEST)")
+E2E_FLAGS ?= ""
+test-e2e: ginkgo ## Run the e2e tests
+	$(GINKGO) --tags $(GO_BUILD_TAGS) $(E2E_FLAGS) -trace -progress $(FOCUS) test/e2e
+
+e2e: run test-e2e kind-cluster-cleanup ## Run e2e test suite on local kind cluster
```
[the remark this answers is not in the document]
dtfranz: Yes, you're right. I missed that - will update.

test/e2e/reconcile_test.go
```diff
@@ -0,0 +1,44 @@
...
+			ctx    context.Context
+			err    error
+		)
+		BeforeEach(func() {
+			ctx = context.Background()
+
+			bundle = &operatorv1alpha1.Operator{
+				ObjectMeta: metav1.ObjectMeta{
+					GenerateName: "operator",
+					Namespace:    "operator-controller-system",
```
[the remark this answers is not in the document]
dtfranz: It is currently namespace scoped, which should be fixed... I wouldn't mind taking care of that in this PR if you think it's appropriate.

test/e2e/reconcile_test.go
dtfranz: Makes perfect sense. I will adjust the phrasing on these tests to make it more clear that this is much more than just testing the Reconcile func. 
Regarding your thought on e2e test requiring rukpak be present: I had been thinking that we would have two categories of e2e. One where each component is e2e tested alone in a vacuum (where possible), and then another for integration with all the components working together. In the case of operator-controller, a simple e2e test that I would envision is one where we create an `Operator` CR then expect the presence of a `BundleDeployment` in the cluster for Operator installation as a result. But if the value-add of a test like that is negligible when we also have a unit test for `Reconcile` which is doing the same thing with fake client then I could see why including only integration style tests makes sense.

test/e2e/reconcile_test.go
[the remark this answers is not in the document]
joelanford: Oh yeah, that should be fixed. I can submit a separate PR to keep things well-scoped.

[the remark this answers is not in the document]
dtfranz: Sorry I forgot to mention, I just went ahead and added that to this PR since it was pretty simple.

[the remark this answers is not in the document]
joelanford: https://github.com/operator-framework/operator-controller/pull/82

Makefile
```diff
@@ -129,12 +159,24 @@ $(LOCALBIN):
...
 
+.PHONY: kind
+kind: $(KIND) ## Download kind locally if necessary.
+$(KIND): $(LOCALBIN)
+	test -s $(LOCALBIN)/kind || GOBIN=$(LOCALBIN) go install sigs.k8s.io/kind@v0.15.0
+
+.PHONY: ginkgo
+ginkgo: $(GINKGO) ## Download ginkgo locally if necessary.
+$(GINKGO): $(LOCALBIN)
+	test -s $(LOCALBIN)/ginkgo || GOBIN=$(LOCALBIN) go install github.com/onsi/ginkgo/v2/ginkgo@v2.1.4
```
joelanford, line 178: The `test -s $(LOCALBIN)/<binName>` tests are sufficient for CI because they'll always fail and we'll always get the correct version. 

But locally, these can be annoying when we bump the version of one of these and you still have an old version from an earlier commit. If there's some diff in behavior between your local version and the one specified here, you'll be scratching your head trying to figure out why.

There are two solutions I've seen for this:
1. Rukpak has a separate [`tools` go module](https://github.com/operator-framework/rukpak/tree/main/hack/tools) with a separate go.mod that lists all the tool versions with [make targets](https://github.com/operator-framework/rukpak/blob/c527e657b4dabfd228068fbe458b72acb013bf00/Makefile#L223-L258) that always `go build`
2. A [somewhat curated script](https://github.com/operator-framework/helm-operator-plugins/blob/main/tools/scripts/fetch) that knows how to check the presence and version of the existing binary to see if a new one is needed [called like this](https://github.com/operator-framework/helm-operator-plugins/blob/dd32bac92534db6a00fe30252cd944c361639ec5/Makefile#L81). It's _mostly_ set and forget.

Option 1 is good if all of our tools are Go-based
Option 2 is probably slightly faster when actually using the tools, but requires some tweaks every now and then when a tool changes its release artifacts.

config/crd/bases/operators.operatorframework.io_operators.yaml
```diff
@@ -13,7 +13,7 @@ spec:
     listKind: OperatorList
     plural: operators
     singular: operator
-  scope: Namespaced
+  scope: Cluster
```
joelanford, line 16: Oh sorry, I submitted #82 for this. Should we merge that to keep separation of concerns and then rebase this one?

Makefile
joelanford: This could be a follow-up for sure.

test/e2e/e2e_suite_test.go
```diff
@@ -0,0 +1,38 @@
...
+
+func TestE2E(t *testing.T) {
+	RegisterFailHandler(Fail)
+	SetDefaultEventuallyTimeout(1 * time.Minute)
+	SetDefaultEventuallyPollingInterval(1 * time.Second)
+	RunSpecs(t, "E2E Suite")
+}
+
+var _ = BeforeSuite(func() {
+	cfg = ctrl.GetConfigOrDie()
```
varshaprasad96, line 30: Could these tests go under [suite_test.go](https://github.com/operator-framework/operator-controller/blob/main/controllers/suite_test.go)? Just because we have the test framework already scaffolded by the tool?

varshaprasad96: Just saw Joe's comment https://github.com/operator-framework/operator-controller/pull/81#discussion_r1061098399, please feel free to ignore my review :)

Only one nit (we could do in a follow up), we could use envtest even for the unit (case 1) tests (though it is mainly intended for integration) if we foresee the interaction of multiple individual components, and would have to test "Reconcile" anyway. This in the case (1) where we need deppy entity source, and essentially mock the functioning of controller. This is just because previously we had faced version incompatibility issues by using locally available binaries and had a hard time figuring the cause.

varshaprasad96 (approved):

awgreene (approved):
Nice work on this PR, please address Joe's open comments. Looks good otherwise!

Makefile
```diff
@@ -54,25 +63,46 @@ fmt: ## Run go fmt against code.
...
+
+kind-load: kind ## Loads the currently constructed image onto the cluster
+	$(KIND) load docker-image $(IMG) --name $(KIND_CLUSTER_NAME)
+
+kind-cluster: kind kind-cluster-cleanup ## Standup a kind cluster
+	$(KIND) create cluster --name ${KIND_CLUSTER_NAME}
+	$(KIND) export kubeconfig --name ${KIND_CLUSTER_NAME}
+
+kind-cluster-cleanup: kind ## Delete the kind cluster
+	$(KIND) delete cluster --name ${KIND_CLUSTER_NAME}
```
tmshort, line 86: Nit: Looking at consistency, we might want to rename these. If we wanted to go with k8s' V-O verbiage:
e.g. 
`kind-cluster` -> `kind-create-cluster`
`kind-cluster-cleanup` -> `kind-delete-cluster`
`kind-load` -> `kind-load-image`

Makefile
```diff
@@ -54,25 +63,46 @@ fmt: ## Run go fmt against code.
...
 ##@ Build
 
 .PHONY: build
 build: manifests generate fmt vet ## Build manager binary.
 	go build -o bin/manager main.go
 
 .PHONY: run
-run: manifests generate fmt vet ## Run a controller from your host.
-	go run ./main.go
+run: docker-build kind-cluster kind-load install deploy wait ## Build the operator-controller then deploy it into a new kind cluster.
```
tmshort, line 95: Do we really want to change the behavior of the `run` target? Would another target make sense?

Also a bit concerned on ordering here, if `make` is run with `-j`. (I not sure if order-only prerequisites will make it any better, but the prerequisites of all these prerequisites need to be proper.)

tmshort: A parallel build will definitely change the order here.
Might want to add:
```
.NOTPARALLEL: run

config/crd/bases/operators.operatorframework.io_operators.yaml
dtfranz: Yes, no problem! I'll back out that change here then rebase after yours merges.

Makefile
dtfranz: I agree with your reasoning for these, but I also named them this way for the sake of consistency with rukpak, so we'd need to change them in both places if we also wanted to be internally consistent. If it's alright with you, I'd prefer to leave these at least for now.

Makefile
dtfranz: This change was also motivated by consistency with rukpak, where `make run` does exactly the same thing. I could add a `run-local` perhaps if we want to keep this capability?
For the `.NOTPARALLEL` thing, I could try adding a line in the makefile that covers all the commands with dependencies, since nearly all of them (not just `run`) would be broken with `-j`.

dtfranz: Updated, I added `.NOTPARALLEL:` which disables it for all targets. They all pretty much all need to be run sequentially so I think this is fine.

Makefile
dtfranz: I did notice number 1 there in rukpak's makefile. I initially was working with that but it ended up turning into a bit of a rabbit-hole with the amount of changes needed, since it appeared as though it would require re-working every built-in make target we got from kubebuilder, and then at that point it felt like I was going to end up with just a copy-paste of rukpak's makefile which would require re-testing all of the targets. We definitely need to take care of it at some point, if you think it's a good thing to take care of now I could give it another go. It just felt like it was getting a bit out-of-scope to me.

dtfranz: I've created issue #85 to track this.

test/e2e/e2e_suite_test.go
varshaprasad96: No worries, I have created an issue to follow up https://github.com/operator-framework/operator-controller/issues/86. We can start using envtest when we get to writing tests for reconciler depending on the complexity of those. Till then its good :)

PR #82, opened 2023-01-05
make Operator API cluster-scoped

joelanford:
Signed-off-by: Joe Lanford <joe.lanford@gmail.com>


dtfranz (approved):
Approved, thanks for doing that

PR #91, opened 2023-01-11
Add Conditions to status

### LINKED ISSUE #73: Update operator api resources with a status to indicate solver success/failure
**User Story:**

As a user, if I create an Operator CR that cannot be fulfilled for any reason, I would like the reason for the failure to be surfaced in the Operator CR's status.

**Open Questions:**

- Are Operator CR's resolved independently or as a group, which should be solved in https://github.com/operator-framework/deppy/issues/75

**Acceptance Criteria:**

- Unmet constraints are surfaced in the Operator CR's status.

tmshort:
* Adds an initial Condition: "Ready" with a "False" status.
* Fixed #73
* Include unit-tests

For: https://issues.redhat.com/browse/OLM-2854


api/v1alpha1/operator_types.go
```diff
@@ -27,8 +27,14 @@ type OperatorSpec struct {
...
 }
 
+const (
+	TypeReady = "Ready"
+)
+
 // OperatorStatus defines the observed state of Operator
-type OperatorStatus struct{}
+type OperatorStatus struct {
+	Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type" protobuf:"bytes,1,rep,name=conditions"`
```
joelanford, line 36: I haven't seen (and I haven't looked very hard) patchStrategy and patchMergeKey on conditions before. Is this something suggested in official k8s docs somewhere?

I'm wondering what the implications are? It seems like this would permit two writers to server-side apply two different types concurrently? Do we want that?

controllers/operator_controller.go
```diff
@@ -47,10 +49,33 @@ type OperatorReconciler struct {
...
+	}
+	apimeta.SetStatusCondition(&operator.Status.Conditions, readyCondition)
+	log.Info("set not ready Condition")
 
 	// TODO(user): your logic here
 
+	// This is something that needs to happen if any Status is updated,
+	// even if an error occurs
+	if err := r.Status().Update(ctx, &operator); err != nil {
+		log.Error(err, "unable to update Operator status")
```
joelanford, line 75: No need to log this error. By default, the controller will log errors returned from `Reconcile`

api/v1alpha1/operator_types.go
tmshort: Yes, these are the recommended attributes:
https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#Condition
What it should mean is that you can only have a single instance of each type.

controllers/operator_controller.go
tmshort: In this case, `err` isn't returned. It can be, but I didn't want to override any other error condition that may be present. (The Reconcile function isn't really set up this way right now.)

controllers/operator_controller.go
```diff
@@ -47,10 +49,33 @@ type OperatorReconciler struct {
...
+	var operator operatorsv1alpha1.Operator
+	if err := r.Get(ctx, req.NamespacedName, &operator); err != nil {
+		log.Error(err, "unable to fetch Operator")
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	readyCondition := metav1.Condition{
+		Type:               operatorsv1alpha1.TypeReady,
+		Status:             metav1.ConditionFalse,
+		Reason:             "startUp",
```
joelanford, line 63: `Reason` should be a `const` in the API package and should be "CamelCased". Clients sometimes use `Reason` in their logic, so this is part of our API.

Also, I know this is probably not meant to stay this way, but the way I would suggest structuring condition modification is:
1. Anytime there's an error during reconciliation, set the conditions relevant to that error.
2. Anytime reconciliation has progressed to the point where you know what a condition will be regardless of any future errors, set that condition.
3. No matter which code path happens, the `Reconcile` function should call `SetStatusCondition` exactly once for every single condition type.

Lastly, what is "startUp" supposed to tell a user about the state of the Operator object? I'd suggest something like reason "NotImplemented" with a message along the lines of "Operator reconciliation is not yet implemented." And then a TODO comment that tells us to remove this once we start building out the implementation.

controllers/operator_controller.go
```diff
@@ -47,10 +49,33 @@ type OperatorReconciler struct {
...
+		ObservedGeneration: operator.GetGeneration(),
+	}
+	apimeta.SetStatusCondition(&operator.Status.Conditions, readyCondition)
+	log.Info("set not ready Condition")
 
 	// TODO(user): your logic here
 
+	// This is something that needs to happen if any Status is updated,
+	// even if an error occurs
+	if err := r.Status().Update(ctx, &operator); err != nil {
```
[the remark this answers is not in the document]
tmshort: I had to look at that twice to see that it's doing what I would actually prefer (i.e. \calling a helper function and then handling all the status updates in a single place).

I can set that up.

controllers/operator_controller.go
tmshort: I was kinda going along the lines of "we hit the reconcile function". But it was not meant to stay that way. Will update with TODO, etc.

controllers/suite_test.go
```diff
@@ -78,3 +83,60 @@ var _ = AfterSuite(func() {
...
+			ctx      context.Context
+			opName   string
+			pkgName  string
+			err      error
+		)
+		BeforeEach(func() {
+			ctx = context.Background()
+
+			opName = "operator-test"
+			pkgName = "package-test"
```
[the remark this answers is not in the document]
tmshort: Thats fair, I had removed originally had random names, but decided later to keep it simple.

tmshort:
And now the basic unit tests are actually run...

tmshort:
@joelanford I believe I've addressed all comments

api/v1alpha1/operator_types.go
dtfranz: Do we also need the other tags for `Conditions` that are listed in the example `FooStatus` on that page? I'm not actually sure, but it's definitely worth adding them and running `make generate manifests` to see what they do to the generated code and/or CRD file.
```yaml
type FooStatus struct{
    // Represents the observations of a foo's current state.
    // Known .status.conditions.type are: "Available", "Progressing", and "Degraded"
    // +patchMergeKey=type
    // +patchStrategy=merge
    // +listType=map
    // +listMapKey=type
    Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type" protobuf:"bytes,1,rep,name=conditions"`

    // other fields
}
```

tmshort: It does add something (`x-kubernetes-list`) values. So I might as well.

tmshort:
This is for: https://issues.redhat.com/browse/OLM-2854

api/v1alpha1/operator_types.go
```diff
@@ -27,8 +27,20 @@ type OperatorSpec struct {
 	PackageName string `json:"packageName"`
 }
 
+const (
+	TypeReady = "Ready"
+
+	StatusNotImplemented = "NotImplemented"
```
awgreene, line 33: Can we add TODOs to define these at some point?

controllers/operator_controller.go
```diff
@@ -47,7 +51,48 @@ type OperatorReconciler struct {
...
+		}
+	}
+
+	// Compare resources - ignoring status - and update if required
+	existingOp.Status, reconciledOp.Status = operatorsv1alpha1.OperatorStatus{}, operatorsv1alpha1.OperatorStatus{}
+	if !equality.Semantic.DeepEqual(existingOp, reconciledOp) {
+		if updateErr := r.Update(ctx, reconciledOp); updateErr != nil {
+			return res, utilerrors.NewAggregate([]error{reconcileErr, updateErr})
+		}
+	}
```
awgreene, line 79: Could you explain when the controller would modify something other than the Operator CR's status? I can't imagine a scenario where we'd want to modify the resource's spec, is this to update something in the metadata (annotations, labels, etc..)? If the point of this update call is to modify something outside the spec, does it make sense to exclude it?

controllers/suite_test.go
```diff
@@ -78,3 +85,60 @@ var _ = AfterSuite(func() {
...
+			operator *operatorsv1alpha1.Operator
+			ctx      context.Context
+			opName   string
+			pkgName  string
+			err      error
+		)
+		BeforeEach(func() {
+			ctx = context.Background()
+
+			opName = fmt.Sprintf("perator-test-%s", rand.String(8))
```
awgreene, line 101:
```suggestion
			opName = fmt.Sprintf("operator-test-%s", rand.String(8))
```

controllers/operator_controller.go
tmshort: This was suggested by @joelanford; it's copied from the rukpak reconciler.

tmshort: (And possibly metadata)

joelanford: Finalizers are the only thing that comes to mind. Defaulting should be handled for us by the apiserver during admission. Not reconciliation.

This is a good question though. Perhaps we should have the code panic if it detects a change we made to non-status, non-finalizers, as that would indicate a programming mistake.

tmshort: Should rukpak be updated as well?

controllers/operator_controller.go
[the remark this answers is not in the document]
joelanford: > handling all the status updates in a single place

Yes, exactly. And only once per reconcile.

api/v1alpha1/operator_types.go
joelanford: > you can only have a single instance of each type.

Another follow-up/TODO/ground rule/invariant:

For every `condition.Type` that we define, that type has a condition set no matter what code path is taken through the reconciler. Every single time reconcile is called, we make sure to evaluate what the value of each `condition.Type` should be and we make sure it is set that way.

TL;DR:
1. Reconcile is idempotent. Same input always results in the same output.
2. Output always includes exactly one of each condition type.

Where input is defined as: "Operator CR (minus its status) + existing cluster state"
And output is defined as: "Operator CR status + new desired cluster state"

controllers/operator_controller.go
tmshort: @joelanford: something like?
```go
	existingOp.Status, reconciledOp.Status = operatorsv1alpha1.OperatorStatus{}, operatorsv1alpha1.OperatorStatus{}
	existingOp.Finalizers, reconciledOp.Finalizers = []string{}, []string{}
	if !equality.Semantic.DeepEqual(existingOp, reconciledOp) {
		panic("spec or metadata changed by reconciler")
	}

Comment (awgreene) [reply controllers/operator_controller.go:79]: This would strip finalizers from the object, which we may not want to. Something like this may work.
```
existingOp.Status, reconciledOp.Status = operatorsv1alpha1.OperatorStatus{}, operatorsv1alpha1.OperatorStatus{}
	if !equality.Semantic.DeepEqual(existingOp.Spec, reconciledOp.Spec) {
		panic("spec or metadata changed by reconciler")
	}
``

tmshort: Are you sure? We aren't doing an `r.Update()` in this case, although that might still be necessary due to updated finalizers. We can't just compare the `Specs`, as that ignores other `metadata`.

tmshort: OK, weirdness! `r.Status().Update()` actually makes additional modifications to the structure. So, do all the comparisons _first_, then update the status, `panic()` if required, then update the spec (should just be the finalizers).
The `reconciledOp` `existingOp` and `compareOp` are all local variables, and k8s is unmodified until an `Update()` call is made.

controllers/operator_controller.go
```diff
@@ -47,7 +51,59 @@ type OperatorReconciler struct {
...
+	reconciledOp := existingOp.DeepCopy()
+	res, reconcileErr := r.reconcile(ctx, reconciledOp)
+
+	// Do checks before any Update()s, as Update() may modify the resource structure!
+	updateStatus := !equality.Semantic.DeepEqual(existingOp.Status, reconciledOp.Status)
+	updateFinalizers := !equality.Semantic.DeepEqual(existingOp.Finalizers, reconciledOp.Finalizers)
+
+	// Compare resources - ignoring status & metadata.finalizers
+	compareOp := reconciledOp.DeepCopy()
+	existingOp.Status, compareOp.Status = operatorsv1alpha1.OperatorStatus{}, operatorsv1alpha1.OperatorStatus{}
```
awgreene, line 72: @joelanford I forget, does controller runtime send a DeepCopy of the resource retrieved from the cache when a get call is made?

controllers/operator_controller.go
awgreene: >Are you sure? We aren't doing an r.Update() in this case, although that might still be necessary due to updated finalizers. We can't just compare the Specs, as that ignores other metadata.

Ah sorry, I didn't notice that you changed the code between the time that I initially reviewed the PR and made that comment.

controllers/operator_controller.go
awgreene: Not an issue, Controller-Runtime does the deepCopy for you.

controllers/operator_controller.go
```diff
@@ -47,7 +51,59 @@ type OperatorReconciler struct {
...
+	existingOp.Finalizers, compareOp.Finalizers = []string{}, []string{}
+	specDiffers := !equality.Semantic.DeepEqual(existingOp, compareOp)
```
awgreene, line 74: Is compareOp needed? Could you not compare the specs directly?
```suggestion
	specDiffers := !equality.Semantic.DeepEqual(&existingOp.Spec, &reconciled.Spec)
```

tmshort: We're not comparing just specs, but specs and the metadata as well. Want to make sure no annotations or labels or etc were added.

tmshort: If we didn't care about the metadata, then that would be correct.

controllers/operator_controller.go
```diff
@@ -47,7 +51,59 @@ type OperatorReconciler struct {
...
+
+	if specDiffers {
+		panic("spec or metadata changed by reconciler")
+	}
+
+	if updateFinalizers {
+		if updateErr := r.Update(ctx, reconciledOp); updateErr != nil {
+			return res, utilerrors.NewAggregate([]error{reconcileErr, updateErr})
+		}
+	}
```
awgreene, line 90: Rather than "updateFinalizers", should this be "updateMetadata"? CC @joelanford

awgreene: I assume that there may be instances where we want to change metadata on the object we're reconciling. If we're suggesting that there is not, then we can leave it as is.

controllers/operator_controller.go
awgreene: This assumes that we want to allow changes to the metadata field.

controllers/operator_controller.go
tmshort: A prior comment from @joelanford implies the only thing that the reconciler could change are the Finalizers and the Status. So those are checked explicitly to see if they changed. If anything else was updated, then there's a `panic()`.

tmshort: Admittedly, `updateFinalizers` causes all of the resource (sans Status) to be updated, but if anything else were to change, it'd `panic()` first.

controllers/operator_controller.go
tmshort: In this case, it's _only_ the finalizers that can change. Any other metadata change will cause a `panic()`

tmshort: `specDiffers` could be better named.

controllers/operator_controller.go
awgreene: Got it. Thanks.

awgreene (approved):
Great work @tmshort, it may be worth adding a note that we're choosing not to allow changes to the object's metadata other than its finalizers, but this is fine.

/approve

dtfranz (approved):
Looks good to me!

controllers/operator_controller.go
joelanford: How about `unexpectedFieldsChanged`?

joelanford: Unresolved only because I'm not sure you'd see my comment. Not merge blocking though.

tmshort: I see it!

controllers/operator_controller.go
```diff
@@ -47,7 +51,59 @@ type OperatorReconciler struct {
...
+
+	if updateStatus {
+		if updateErr := r.Status().Update(ctx, reconciledOp); updateErr != nil {
+			return res, utilerrors.NewAggregate([]error{reconcileErr, updateErr})
+		}
+	}
+
+	if specDiffers {
+		panic("spec or metadata changed by reconciler")
+	}
```
joelanford, line 84: There are probably good arguments on both sides, but did we consider moving this before the status update?

tmshort: I gave it some thought. But I wasn't sure. Having the `Status` update at least gives an idea of what error may have lead to this condition. But this is "something that shouldn't happen"™, and should never be seen after release.

controllers/operator_controller.go
joelanford: Yeah, I think it's a good idea to only allow finalizers to be changed. As a rule, reconcilers should only change non-status, non-finalizer fields on objects they _manage_/_own_, not their primary objects.

controllers/operator_controller.go
```diff
@@ -47,7 +51,59 @@ type OperatorReconciler struct {
...
+// Helper function to do the actual reconcile
+func (r *OperatorReconciler) reconcile(ctx context.Context, op *operatorsv1alpha1.Operator) (ctrl.Result, error) {
+
+	// TODO(user): change ReasonNotImplemented when functionality added
+	readyCondition := metav1.Condition{
+		Type:               operatorsv1alpha1.TypeReady,
+		Status:             metav1.ConditionFalse,
+		Reason:             operatorsv1alpha1.ReasonNotImplemented,
+		Message:            "The Reconcile operation is not implemented",
+		ObservedGeneration: op.GetGeneration(),
```
joelanford, line 104: Oh this reminds me of another way to test for making sure we update all conditions. Unit tests of the `Reconcile` method should make sure that all condition types have condition objects AND all of those objects have observedGeneration that matches the metadata.generation of the Operator that was reconciled.

controllers/suite_test.go
```diff
@@ -78,3 +85,60 @@ var _ = AfterSuite(func() {
...
+
+			err = k8sClient.Get(ctx, client.ObjectKey{
+				Name: opName,
+			}, getOperator)
+			Expect(err).To(Not(HaveOccurred()))
+
+			// There should always be a "Ready" condition, regardless of Status.
+			conds := getOperator.Status.Conditions
+			Expect(conds).To(Not(BeEmpty()))
+			Expect(conds).To(ContainElement(HaveField("Type", operatorsv1alpha1.TypeReady)))
```
joelanford, line 141: Should we also test for the specific Status and Reason we expect?

tmshort: I decided to not bother with that. As I wanted to minimize changes (vs. additions) to the tests as code is added.

tmshort:
Thank you all. I'll do a follow up to address some of minor issues/nits.

PR #97, opened 2023-01-24
[resolution pt.3] bundles and deps variable source

perdasilva:
This PR adds the `Bundles and Dependencies` variable source. It generate `BundleVariables` for all `RequiredPackage` variables found from executing the `inputVariableSource`s. It does so by inspecting the `BundleEntities` included in each `RequiredPackageVariable`, resolving dependencies, and outputting `BundleVariables` for all `BundleEntities` and their resolved dependencies.

### DIFF
--- a/internal/resolution/variable_sources/bundles_and_dependencies/bundles_and_dependencies.go
+++ b/internal/resolution/variable_sources/bundles_and_dependencies/bundles_and_dependencies.go
@@ -1,0 +1,166 @@
+package bundles_and_dependencies
+
+import (
+	"context"
+	"fmt"
+	"sort"
+
+	"github.com/blang/semver/v4"
+	"github.com/operator-framework/deppy/pkg/deppy"
+	"github.com/operator-framework/deppy/pkg/deppy/constraint"
+	"github.com/operator-framework/deppy/pkg/deppy/input"
+
+	olmentity "github.com/operator-framework/operator-controller/internal/resolution/variable_sources/entity"
+	"github.com/operator-framework/operator-controller/internal/resolution/variable_sources/required_package"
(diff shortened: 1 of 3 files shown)


internal/resolution/variable_sources/bundles_and_dependencies/bundles_and_dependencies.go
```diff
@@ -0,0 +1,166 @@
...
+		dependencyIDs = append(dependencyIDs, bundle.ID)
+	}
+	var constraints []deppy.Constraint
+	if len(dependencyIDs) > 0 {
+		constraints = append(constraints, constraint.Dependency(dependencyIDs...))
+	}
+	return &BundleVariable{
+		SimpleVariable: input.NewSimpleVariable(bundleEntity.ID, constraints...),
+		bundleEntity:   bundleEntity,
+		dependencies:   dependencyBundleEntities,
```
awgreene, line 45: For my understanding:
- SimpleVariables seems to take each dependency bundleID for purposes of resolution.
- dependencies are the actual entities, which we seem to store here for quick reference later rather than retrieving the entity from the source, is that correct?

awgreene (approved):
Looks good!

internal/resolution/variable_sources/bundles_and_dependencies/bundles_and_dependencies.go
perdasilva: `SimpleVariable` is the simplest implementation of `deppy.Variable` possible. It's just a convenience. Basically, it just has an id and a list of constraints and you can `Add`  constraints. It will contain everything the resolver needs to convert this to some boolean expression.

`dependencies` are the actual bundle dependencies - exactly so we don't need to go back to the backing store to get them later.

PR #99, opened 2023-01-24
[resolution pt.5] olm variable source

perdasilva:
This PR adds the `OLMVariableSource` which orchestrates variable creation by using the other variables sources (e.g. `RequiredPackage`, `BundlesAndDeps`, and `GlobalConstraints`)

### DIFF
--- a/internal/resolution/variable_sources/olm/olm_test.go
+++ b/internal/resolution/variable_sources/olm/olm_test.go
@@ -1,0 +1,156 @@
+package olm_test
+
+import (
+	"context"
+	"fmt"
+	"sort"
+	"strings"
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/operator-framework/deppy/pkg/deppy"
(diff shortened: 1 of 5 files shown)


internal/resolution/variable_sources/olm/olm_test.go
```diff
@@ -0,0 +1,156 @@
...
+			"packageA package uniqueness",
+			"prometheus package uniqueness",
+		})))
+	})
+
+	It("should return an errors when they occur", func() {
+		olmVariableSource := olm.NewOLMVariableSource("prometheus", "packageA")
+		_, err := olmVariableSource.GetVariables(context.Background(), FailEntitySource{})
+		Expect(err).To(HaveOccurred())
+	})
```
awgreene, line 123: Does this sufficiently handle the case where an OLMVariableSource is provided that can't be fulfilled by the entity source?

awgreene (approved):
Looks great!

internal/resolution/variable_sources/olm/olm_test.go
perdasilva: Yeah, it should basically throw it's hand up in the air XD

PR #100, opened 2023-01-24
[resolution pt.6] operator resolver

perdasilva:
This PR adds the operator resolver, which queries the cluster for operator resources, converts them to package names and runs resolution.


awgreene (approved):
Great work!

PR #102, opened 2023-01-25
fixup unit test description

perdasilva:
This PR just addresses a comment in a previous PR where a unit test description was found to be mismatched

Signed-off-by: perdasilva <perdasilva@redhat.com>

### DIFF
--- a/internal/resolution/variable_sources/utils/predicates/predicates_test.go
+++ b/internal/resolution/variable_sources/utils/predicates/predicates_test.go
@@ -42,7 +42,7 @@ var _ = Describe("Predicates", func() {
 	})
 
 	Describe("InChannel", func() {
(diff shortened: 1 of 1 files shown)


internal/util/logging.go
```diff
@@ -0,0 +1,27 @@
+package util
```
tmshort, line 1: Is this used?

tmshort: (meaning the whole file)

perdasilva: oh crap - nah that snuck in

perdasilva: was also trying to think about logging then gave up - ty!

awgreene (approved):
/approve

PR #103, opened 2023-01-26
Added e2e test for operator resolution

dtfranz:
Resolver e2e test for Operator Controller. Creates an Operator Resource specifying prometheus as the packageName then expects resolution to succeed and provide a bundle path.


test/e2e/install_test.go
```diff
@@ -0,0 +1,61 @@
+package e2e
```
tmshort, line 1:
```
package e2e_test
```
??

test/e2e/install_test.go
```diff
@@ -0,0 +1,61 @@
...
+			err      error
+		)
+		BeforeEach(func() {
+			ctx = context.Background()
+			operator = &operatorv1alpha1.Operator{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: fmt.Sprintf("operator-%s", rand.String(8)),
+				},
+				Spec: operatorv1alpha1.OperatorSpec{
+					PackageName: "prometheus",
```
awgreene, line 36: Nit, use a constant.

test/e2e/install_test.go
perdasilva: I think package e2e is fine - there's only tests and test tooling here...no?

test/e2e/install_test.go
dtfranz: Updated :smiley:

test/e2e/install_test.go
dtfranz: Yeah that's right, I feel like I usually just see that convention when we have `actualpkg` and then when we test it `actualpkg_test`

grokspawn (approved):

PR #104, opened 2023-01-27
BundleDeployment creation with BundleImage

dtfranz:
Combines @perdasilva 's latest commits with a selection from @varshaprasad96 and @ankitathomas 's [controller_prototype](https://github.com/varshaprasad96/operator-controller/tree/controller_prototype) to enable the operator-controller to create bundledeployments using a (currently hard-coded) CatalogSource. Also adds additional Makefile targets to enable simpler installation of dependencies (cert-manager and rukpak).

It should be possible to demonstrate milestone 1 by running the following two commands:
```bash
$ make run
$ kubectl apply -f config/samples/operators_v1alpha1_operator.yaml
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,8 @@
 export IMAGE_REPO ?= quay.io/operator-framework/operator-controller
 export IMAGE_TAG ?= devel
 export GO_BUILD_TAGS ?= upstream
+export CERT_MGR_VERSION ?= v1.9.0
+export WAIT_TIMEOUT ?= 60s
 IMG?=$(IMAGE_REPO):$(IMAGE_TAG)
 
 OPERATOR_CONTROLLER_NAMESPACE ?= operator-controller-system
@@ -99,11 +101,11 @@ build: manifests generate fmt vet ## Build manager binary.
 	go build -o bin/manager main.go
 
 .PHONY: run
-run: docker-build kind-cluster kind-load install deploy wait ## Build the operator-controller then deploy it into a new kind cluster.
+run: docker-build kind-cluster kind-load cert-mgr rukpak install deploy wait ## Build the operator-controller then deploy it into a new kind cluster.
 
 .PHONY: wait
 wait:
-	kubectl wait --for=condition=Available --namespace=$(OPERATOR_CONTROLLER_NAMESPACE) deployment/operator-controller-controller-manager --timeout=60s
+	kubectl wait --for=condition=Available --namespace=$(OPERATOR_CONTROLLER_NAMESPACE) deployment/operator-controller-controller-manager --timeout=$(WAIT_TIMEOUT)
 
 # If you wish built the manager image targeting other platforms you can use the --platform flag.
 # (i.e. docker build --platform linux/arm64 ). However, you must enable docker buildKit for it.
@@ -138,6 +140,17 @@ docker-buildx: test ## Build and push docker image for the manager for cross-pla
 ifndef ignore-not-found
   ignore-not-found = false
 endif
+
+## TODO dfranz: replace cert-mgr and rukpak targets with our chosen method of distribution when available
+.PHONY: cert-mgr
+cert-mgr: ## Install cert-manager
+	kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/$(CERT_MGR_VERSION)/cert-manager.yaml
+	kubectl wait --for=condition=Available --namespace=cert-manager deployment/cert-manager-webhook --timeout=$(WAIT_TIMEOUT)
+
+.PHONY: rukpak
+rukpak: ## Install rukpak
+	kubectl apply -f https://github.com/operator-framework/rukpak/releases/latest/download/rukpak.yaml
+	kubectl wait --for=condition=Available --namespace=rukpak-system deployment/core --timeout=$(WAIT_TIMEOUT)
 
 .PHONY: install
 install: manifests kustomize ## Install CRDs into the K8s cluster specified in ~/.kube/config.
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -6,6 +6,17 @@ metadata:
   name: manager-role
 rules:
 - apiGroups:
+  - core.rukpak.io
+  resources:
+  - bundledeployments
+  verbs:
+  - create
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
   - operators.operatorframework.io
   resources:
   - operators
--- a/controllers/operator_controller.go
+++ b/controllers/operator_controller.go
@@ -19,10 +19,12 @@ package controllers
 import (
 	"context"
 
+	rukpakv1alpha1 "github.com/operator-framework/rukpak/api/v1alpha1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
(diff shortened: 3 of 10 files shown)


Makefile
```diff
@@ -139,6 +140,16 @@ ifndef ignore-not-found
...
+.PHONY: cert-mgr
+cert-mgr: ## Install cert-manager
+	kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/$(CERT_MGR_VERSION)/cert-manager.yaml
+	kubectl wait --for=condition=Available --namespace=cert-manager deployment/cert-manager-webhook --timeout=60s
+
+.PHONY: rukpak
+rukpak: ## Install rukpak
+	kubectl apply -f https://github.com/operator-framework/rukpak/releases/latest/download/rukpak.yaml
+	kubectl wait --for=condition=Available --namespace=rukpak-system deployment/core --timeout=60s
+
```
joelanford, line 152: Add a TODO to swap this out with whatever we come up with for the "distribution" or install script or whatever that ends up being?

controllers/operator_controller.go
```diff
@@ -145,11 +154,47 @@ func (r *OperatorReconciler) reconcile(ctx context.Context, op *operatorsv1alpha
 	return ctrl.Result{}, nil
 }
 
+func (r *OperatorReconciler) generateExpectedBundleDeployment(o operatorsv1alpha1.Operator, bundlePath string) (*rukpakv1alpha1.BundleDeployment, error) {
+	dep := &rukpakv1alpha1.BundleDeployment{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: o.GetName(),
+		},
+		Spec: rukpakv1alpha1.BundleDeploymentSpec{
+			//TODO: Don't assume plain provisioner
```
joelanford, line 163: +1 

Long term, I envision a separate configuration API that sets up mappings from bundle properties to provisioner class names.

Mid term, perhaps we get some mileage around a property that sets the provisioner class name directly. We have to be careful with this though because I expect `ProvisionerClass` could end up being a new API in rukpak, which means provisionerClassName becomes a runtime thing, not something hardcoded in rukpak.

Short term, what you have here will be fine for quite awhile, I imagine.

controllers/operator_controller.go
```diff
@@ -145,11 +154,47 @@ func (r *OperatorReconciler) reconcile(ctx context.Context, op *operatorsv1alpha
...
+			Template: &rukpakv1alpha1.BundleTemplate{
+				ObjectMeta: metav1.ObjectMeta{
+					// TODO: Remove
+					Labels: map[string]string{
+						"app": "my-bundle",
+					},
+				},
+				Spec: rukpakv1alpha1.BundleSpec{
+					Source: rukpakv1alpha1.BundleSource{
+						// TODO: Don't assume image type
```
joelanford, line 174: IMO, we should make that bundle path property something a bit more definitive (e.g. bundle image reference). At that point, it would be totally safe to assume that we use the rukpak image type if we have that property. And that would then leave the door open for bundles to have different properties that map to a different source type.

controllers/operator_controller.go
```diff
@@ -145,11 +154,47 @@ func (r *OperatorReconciler) reconcile(ctx context.Context, op *operatorsv1alpha
...
+	}
+	return dep, nil
+}
+
 // SetupWithManager sets up the controller with the Manager.
 func (r *OperatorReconciler) SetupWithManager(mgr ctrl.Manager) error {
-	r.resolver = resolution.NewOperatorResolver(mgr.GetClient(), resolution.HardcodedEntitySource)
-
 	err := ctrl.NewControllerManagedBy(mgr).
+		Owns(&rukpakv1alpha1.BundleDeployment{}).
```
joelanford, line 197: Convention/nit/my opinion: Always declare `For` as the first watch. Its the primary CR type. Then Owns, then Watches.

Sort of ordering from most coupling (self) to least coupling (things I watch, but don't own).

dtfranz: Yeah, that makes sense to me too. I'll update it.

controllers/operator_controller.go
```diff
@@ -38,13 +40,15 @@ type OperatorReconciler struct {
 	client.Client
 	Scheme *runtime.Scheme
 
-	resolver *resolution.OperatorResolver
+	Resolver *resolution.OperatorResolver
```
awgreene, line 43: Does this need to be exported? Could we just create a GetEntity(id DeppyID) function for the OperatorReconciler type?

dtfranz: Not sure if this is answering your second question, but for the first one it was done this way to allow the resolver to be set up separately from the controller in `main()`, which you can see in the prototype [here](https://github.com/varshaprasad96/operator-controller/blob/controller_prototype/main.go#L150). This seems to make it easier to set up the catalog source cache.

controllers/operator_controller.go
dtfranz: Agreed here. I've created another issue here #108 to address this, since I was hoping it wouldn't be a blocker for our shorter term goals.

controllers/operator_controller.go
awgreene: That could be done with a `NewOperatorReconciler(c client, s scheme, r resolver) &OperatorReconciler` function, right? Point being that once the cache is set we shouldn't allow it to be manipulated directly.

awgreene: Well, avoid allowing people to manipulate the resolver that is.

dtfranz: Yeah I see your point there, I'll update :+1:

config/rbac/role.yaml
```diff
@@ -5,6 +5,17 @@ metadata:
   creationTimestamp: null
   name: manager-role
```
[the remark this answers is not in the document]
dtfranz: Yes, it is kubebuilder - and it also gets prepended with a prefix set in `config/defaults/kustomization.yaml`: `namePrefix: operator-controller-`, so it ends up as `operator-controller-manager-role`. Personally I also don't like this way of doing it much but it's also kind of a pain to go through and replace all the names with different values. Plus some of them don't get prepended and I've never quite figured out the rule for when that happens...

controllers/operator_controller.go
```diff
@@ -124,9 +128,14 @@ func (r *OperatorReconciler) reconcile(ctx context.Context, op *operatorsv1alpha
 					bundlePath, err := v.BundleEntity().BundlePath()
 					if err != nil {
 						return ctrl.Result{}, err
+					} // Create bundleDeployment
```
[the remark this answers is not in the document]
dtfranz: Updated

controllers/operator_controller.go
dtfranz: Updated

Makefile
dtfranz: Updated

config/rbac/role.yaml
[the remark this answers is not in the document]
awgreene: Let's avoid touching it for now, keep in mind that all kubebuilder controllers follow this convention, so there are some assumptions about the name of this role.

Makefile
```diff
@@ -139,6 +140,16 @@ ifndef ignore-not-found
   ignore-not-found = false
 endif
 
+.PHONY: cert-mgr
+cert-mgr: ## Install cert-manager
+	kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/$(CERT_MGR_VERSION)/cert-manager.yaml
+	kubectl wait --for=condition=Available --namespace=cert-manager deployment/cert-manager-webhook --timeout=60s
```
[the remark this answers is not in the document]
dtfranz: Updated, can now be overridden with `WAIT_TIMEOUT`

awgreene (approved):

awgreene:
Oh, can we fix the commit header?

PR #105, opened 2023-01-27
Solution refactor

perdasilva:
This PR updates the deppy dependency and refactors the resolver and controller to deal with the new `Solution` structure.
It also takes the opportunity to make some small updates to the variable source unit tests.


awgreene (approved):
/approve

PR #107, opened 2023-01-28
controller code/test refactoring/additions

joelanford:
I noticed y'all made some great progress this week, and got the itch to stay in the loop and pitch in. This PR is a whole bunch of changes all wrapped up together, so I have no illusions about it being merged as is. I'm curious what folks think!

Main bullet points are:
- [x] remove unnecessary RBAC rules for create/update/patch/delete Operator
  objects. this leaves only get, list, watch, update status, and update
  finalizers. (#116)
- [x] use server-side apply to manage bundle deployments (#114)
- [x] unit test refactoring (#115)
   - move reconciler unit tests to a separate file from the suite. if we add other controllers in the future, we'll likely want separate test files for each controller.
   - export the reconciler's resolver field to make it easy to swap out for tests (also add a test entity source)
   - remove the manager setup in the unit tests. this ensures we can test individual runs of the Reconcile function, which is good for two reasons: 
       1. no need to use `Eventually`, no polling, etc. so it speeds up the tests 
       2. it ensures that we fully understand the individual transitions that occur from a beginning state to an end state when multiple Reconcile calls are necessary. (e.g. r1: create BD, r2: update Op based on BD status change to unpacking, r3: update Op again based on BD status change to installed successfully)
   - added more unit tests to increase code coverage.
   - moved conditions and reasons checks to invariants that run after every test of reconciling an existing Operator object.
- [x] decrease nesting in main reconcile code by setting conditions and returning early when handling unexpected errors. (#118)
- [ ] make certain bundle entity properties optional. not all bundles will provide/require GVKs or require other packages. lookups for those entities should return nil slices rather than an error.

EDIT: I split each of these main things out into separate PRs now that it seems like folks are generally onboard.

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -21,12 +21,8 @@ rules:
   resources:
   - operators
   verbs:
-  - create
-  - delete
   - get
   - list
-  - patch
-  - update
   - watch
 - apiGroups:
   - operators.operatorframework.io
--- a/controllers/operator_controller.go
+++ b/controllers/operator_controller.go
@@ -20,10 +20,12 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/operator-framework/deppy/pkg/deppy/solver"
 	rukpakv1alpha1 "github.com/operator-framework/rukpak/api/v1alpha1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
@@ -35,6 +37,7 @@ import (
 	operatorsv1alpha1 "github.com/operator-framework/operator-controller/api/v1alpha1"
 	"github.com/operator-framework/operator-controller/internal/resolution"
 	"github.com/operator-framework/operator-controller/internal/resolution/variable_sources/bundles_and_dependencies"
+	"github.com/operator-framework/operator-controller/internal/resolution/variable_sources/entity"
 )
 
 // OperatorReconciler reconciles a Operator object
@@ -104,7 +107,6 @@ func checkForUnexpectedFieldChange(a, b operatorsv1alpha1.Operator) bool {
 
 // Helper function to do the actual reconcile
 func (r *OperatorReconciler) reconcile(ctx context.Context, op *operatorsv1alpha1.Operator) (ctrl.Result, error) {
-
 	// run resolution
 	solution, err := r.Resolver.Resolve(ctx)
 	if err != nil {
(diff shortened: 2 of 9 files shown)


config/rbac/role.yaml
```diff
@@ -21,12 +21,8 @@ rules:
   resources:
   - operators
   verbs:
-  - create
-  - delete
   - get
   - list
-  - patch
-  - update
```
tmshort, line 29: Related but kinda. I still think that "manager-role" is too generic for a `ClusterRole`. Every single operator made with kubebuilder is going to create, by default, a `ClusterRole` named "manager-role", and this _will_ lead to conflicts and debugging issues.

This is apparently done via the Makefile:
https://github.com/kubernetes-sigs/kubebuilder/discussions/2297

tmshort: These changes will not survive a `make manifests`.
AFAICT, the proper mechanism is to update the kubebuilder tags:
https://github.com/operator-framework/operator-controller/blob/13ea200dfa877b268791ccd77089e505ee40fc8d/controllers/operator_controller.go#L54

joelanford: The final name gets prefixed with project name via kustomize, iirc.

joelanford: And if we don't like `manager-role` suffix, we can change it here: https://github.com/operator-framework/operator-controller/blob/13ea200dfa877b268791ccd77089e505ee40fc8d/Makefile#L57

But I'd suggest sticking with the defaults unless we have a good reason not to because that's what other contributors and cluster admins familiar with kubebuilder-based operators will be used to.

tmshort: As of right now, it does not appear that operator-controller is setup with the proper kustomization.
Examples I seeing (e.g. https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/src/reference/controller-gen.md) are using different name for `rbac:roleName`
Something to do in another PR, but not here.

controllers/operator_controller.go
```diff
@@ -38,23 +40,13 @@ import (
...
-		Client:   c,
-		Scheme:   s,
-		resolver: r,
-	}
+	Scheme   *runtime.Scheme
+	Resolver *resolution.OperatorResolver
 }
 
-//+kubebuilder:rbac:groups=operators.operatorframework.io,resources=operators,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=operators.operatorframework.io,resources=operators,verbs=get;list;watch
```
tmshort, line 47: I must not have scrolled down far enough! This was in the original commit.

controllers/operator_controller.go
```diff
@@ -112,61 +104,105 @@ func checkForUnexpectedFieldChange(a, b operatorsv1alpha1.Operator) bool {
...
-			case *bundles_and_dependencies.BundleVariable:
-				packageName, err := v.BundleEntity().PackageName()
+		apimeta.SetStatusCondition(&op.Status.Conditions, metav1.Condition{
+			Type:               operatorsv1alpha1.TypeReady,
+			Status:             metav1.ConditionFalse,
+			Reason:             operatorsv1alpha1.ReasonResolutionFailed,
+			Message:            err.Error(),
+			ObservedGeneration: op.GetGeneration(),
+		})
+		return ctrl.Result{}, nil
```
tmshort, line 118: Not returning `err` here? Other places set `Message: err.Error()` and also return `err`.

controllers/operator_controller.go
```diff
@@ -193,11 +229,6 @@ func (r *OperatorReconciler) generateExpectedBundleDeployment(o operatorsv1alpha
 			},
 		},
 	}
-
-	if err := ctrl.SetControllerReference(&o, dep, r.Scheme); err != nil {
```
tmshort, line 197: Does this need to be done anywhere else?

dtfranz: Looks like it's being done manually above on 196. Follow-up though: what's the reason behind doing it manually vs using the `ctrl.SetControllerReference` func?

joelanford: Avoid having to handle an error.

controllers/operator_controller.go
joelanford: I think I kept the same return value for this code path.

Maybe we should return the error though? It would mean exponential back off reconciliation, which I think is almost always a good idea unless we're absolutely sure something else will trigger another reconcile in the future to get past this error.

joelanford: I updated in the most recent commit. Now, all errors should be returned.

joelanford:
I added another commit, which I think improves the controller code a bit more.  It also adds more unit test scenarios. I'm fairly happy with those, but they do introduce some duplication that I think could be reduced a bit, so I think its worth a follow up to focus on that.

main.go
```diff
@@ -92,11 +88,11 @@ func main() {
...
 
-	if err = controllers.NewOperatorReconciler(
-		mgr.GetClient(),
-		mgr.GetScheme(),
-		resolution.NewOperatorResolver(mgr.GetClient(), resolution.HardcodedEntitySource),
-	).SetupWithManager(mgr); err != nil {
+	if err = (&controllers.OperatorReconciler{
+		Client:   mgr.GetClient(),
+		Scheme:   mgr.GetScheme(),
+		Resolver: resolution.NewOperatorResolver(mgr.GetClient(), resolution.HardcodedEntitySource),
```
perdasilva, line 94: much better!

internal/resolution/variable_sources/entity/bundle_entity.go
```diff
@@ -154,7 +154,7 @@ func (b *BundleEntity) loadRequiredGVKs() error {
 	b.mu.Lock()
 	defer b.mu.Unlock()
 	if b.requiredGVKs == nil {
-		requiredGVKs, err := loadFromEntity[[]GVKRequired](b.Entity, property.TypeGVKRequired)
+		requiredGVKs, err := loadFromEntity[[]GVKRequired](b.Entity, property.TypeGVKRequired, false)
```
perdasilva, line 157: as we discussed online, I think we should only push these booleans to constants to make it a little more readable and that's basically the cherry on top ^^

might be worth splitting the suite_test changes to a different PR. But not strictly necesssary...

PR #114, opened 2023-02-02
Use server-side apply to ensure the desired bundle deployments exist

joelanford:
One facet of #107

Server-side apply is the preferred way to manage owned resources on a cluster. With forced ownership of the fields we care about, it give the operator controller the ability to always ensure the desired fields are set, while also allowing other controllers to manipulate fields we don't care about and that are irrelevant to the core functionality of the operator controller.


tmshort (approved):
/lgtm

PR #115, opened 2023-02-02
refactor and add unit tests for operator reconciler

### LINKED ISSUE #86: Start using envtest for testing reconciler
With the introduction of a separate suite to test the working of reconciler in [here](https://github.com/operator-framework/operator-controller/pull/81/files#diff-93b46b0715181c9cf397eed60862261b7839ff721599fcc393a393b9edea610bR32-R37), it would be useful to spin up an isolated test server with envtest instead of directly taking in the config of the kind cluster and using locally available binaries. This will help us in avoiding any version incompatibility issues that may occur.

joelanford:
Another facet of #107, covering the following:
- move reconciler unit tests to a separate file from the suite. if we add other controllers in the future, we'll likely want separate test files for each controller.
- export the reconciler's resolver field to make it easy to swap out for tests (also add a test entity source)
- remove the manager setup in the unit tests. this ensures we can test individual runs of the Reconcile function, which is good for two reasons:
  - no need to use Eventually, no polling, etc. so it speeds up the tests
  - it ensures that we fully understand the individual transitions that occur from a beginning state to an end state when multiple Reconcile calls are necessary. (e.g. r1: create BD, r2: update Op based on BD status change to unpacking, r3: update Op again based on BD status change to installed successfully)
added more unit tests to increase code coverage.
- moved conditions and reasons checks to invariants that run after every test of reconciling an existing Operator object.

Closes #86

### DIFF
--- a/controllers/operator_controller.go
+++ b/controllers/operator_controller.go
@@ -40,17 +40,8 @@ import (
 // OperatorReconciler reconciles a Operator object
 type OperatorReconciler struct {
 	client.Client
-	Scheme *runtime.Scheme
-
-	resolver *resolution.OperatorResolver
-}
-
-func NewOperatorReconciler(c client.Client, s *runtime.Scheme, r *resolution.OperatorResolver) *OperatorReconciler {
-	return &OperatorReconciler{
-		Client:   c,
-		Scheme:   s,
-		resolver: r,
-	}
+	Scheme   *runtime.Scheme
+	Resolver *resolution.OperatorResolver
 }
 
 //+kubebuilder:rbac:groups=operators.operatorframework.io,resources=operators,verbs=get;list;watch
@@ -120,9 +111,9 @@ func (r *OperatorReconciler) reconcile(ctx context.Context, op *operatorsv1alpha
 	var message = "resolution was successful"
 
 	// run resolution
-	solution, err := r.resolver.Resolve(ctx)
+	solution, err := r.Resolver.Resolve(ctx)
 	if err != nil {
-		status = metav1.ConditionTrue
+		status = metav1.ConditionFalse
 		reason = operatorsv1alpha1.ReasonResolutionFailed
 		message = err.Error()
 	} else {
--- a/controllers/operator_controller_test.go
+++ b/controllers/operator_controller_test.go
@@ -1,0 +1,344 @@
+package controllers_test
+
+import (
+	"context"
+	"fmt"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/operator-framework/deppy/pkg/deppy"
+	"github.com/operator-framework/deppy/pkg/deppy/input"
+	rukpakv1alpha1 "github.com/operator-framework/rukpak/api/v1alpha1"
+	apimeta "k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/rand"
+	"k8s.io/utils/pointer"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	operatorsv1alpha1 "github.com/operator-framework/operator-controller/api/v1alpha1"
+	"github.com/operator-framework/operator-controller/controllers"
+	"github.com/operator-framework/operator-controller/internal/resolution"
+	operatorutil "github.com/operator-framework/operator-controller/internal/util"
+)
+
+var _ = Describe("Reconcile Test", func() {
+	var (
+		ctx        context.Context
+		reconciler *controllers.OperatorReconciler
+	)
+	BeforeEach(func() {
+		ctx = context.Background()
+		reconciler = &controllers.OperatorReconciler{
+			Client:   cl,
+			Scheme:   sch,
+			Resolver: resolution.NewOperatorResolver(cl, testEntitySource),
+		}
+	})
+	When("the operator does not exist", func() {
+		It("returns no error", func() {
+			res, err := reconciler.Reconcile(context.Background(), ctrl.Request{NamespacedName: types.NamespacedName{Name: "non-existent"}})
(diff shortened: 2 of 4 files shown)


controllers/operator_controller.go
```diff
@@ -118,9 +109,9 @@ func (r *OperatorReconciler) reconcile(ctx context.Context, op *operatorsv1alpha
 	var message = "resolution was successful"
 
 	// run resolution
-	solution, err := r.resolver.Resolve(ctx)
+	solution, err := r.Resolver.Resolve(ctx)
 	if err != nil {
-		status = metav1.ConditionTrue
+		status = metav1.ConditionFalse
```
tmshort, line 114: Shouldn't this remain `ConditionTrue` as resolution _did_ fail?

main.go
```diff
@@ -20,22 +20,18 @@ import (
 	"flag"
 	"os"
 
-	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
-	// to ensure that exec-entrypoint and run can make use of them.
-	_ "k8s.io/client-go/plugin/pkg/client/auth"
-
+	rukpakv1alpha1 "github.com/operator-framework/rukpak/api/v1alpha1"
```
tmshort, line 23: Wondering why this moved up here (i.e. away from the other `github.com` imports). I have seen `go fmt` move things around, but it seems fine with this.

tmshort (commented):
`make test-unit` locally succeeds
`make e2e` locally succeeds
But `make test` locally fails (even after a rebase to `main`). It passes on `main`, however.
(`make test` should apparently be part of the CI)
It's actually failing on `make test-e2e`...

controllers/operator_controller.go
joelanford: `ConditionTrue` is saying `Ready: True`.  Since resolution failed, shouldn't we be saying `Ready: false`?

main.go
joelanford: Oh yeah, that my IDE running `gosimports` (https://github.com/rinchsan/gosimports) to automatically organize imports into groups and sort them.

joelanford:
Re: `make test`, I think that's broken? If I clear out my kind clusters and run `make test`, it fails on `main`.

`make test` only include `test-e2e`, but not `e2e`, so the kind cluster setup doesn't happen.

controllers/operator_controller.go
joelanford: As an aside, I think the structure of how these condition values gets set makes it difficult to reason about. I've fixed this in #107, so as soon as this change lands, I can layer in another PR that simplifies the `Reconcile` function and makes it more obvious how/when the ready condition is set.

varshaprasad96 (approved):
/lgtm

controllers/operator_controller_test.go
```diff
@@ -0,0 +1,344 @@
...
+					Expect(bd.Spec.Template.Spec.Source.Type).To(Equal(rukpakv1alpha1.SourceTypeImage))
+					Expect(bd.Spec.Template.Spec.Source.Image).NotTo(BeNil())
+					Expect(bd.Spec.Template.Spec.Source.Image.Ref).To(Equal("quay.io/operatorhubio/prometheus@sha256:5b04c49d8d3eff6a338b56ec90bdf491d501fe301c9cdfb740e5bff6769a21ed"))
+				})
+				It("sets resolution success status", func() {
+					cond := apimeta.FindStatusCondition(operator.Status.Conditions, operatorsv1alpha1.TypeReady)
+					Expect(cond).NotTo(BeNil())
+					Expect(cond.Status).To(Equal(metav1.ConditionTrue))
+					Expect(cond.Reason).To(Equal(operatorsv1alpha1.ReasonResolutionSucceeded))
+					Expect(cond.Message).To(Equal("resolution was successful"))
```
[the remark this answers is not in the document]
joelanford: I'm not opposed to constants for messages we reuse, but I'm pretty sure condition messages are not considered stable or part of the API a consumer is supposed to depend on.

controllers/operator_controller.go
tmshort: You're right. I was conflating "Resolution failed" with the type.

OK, I think the confusion is that it's not obvious this is for the "Ready" type, since `TypeReady` is not very local, it's down 30+ lines.

controllers/operator_controller_test.go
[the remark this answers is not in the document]
tmshort: I agree with joe. `Reason` is the programatic interface.

tmshort:
> Re: `make test`, I think that's broken? If I clear out my kind clusters and run `make test`, it fails on `main`.
> 
> `make test` only include `test-e2e`, but not `e2e`, so the kind cluster setup doesn't happen.

See #119

controllers/operator_controller_test.go
```diff
@@ -0,0 +1,344 @@
...
+					Spec:       operatorsv1alpha1.OperatorSpec{PackageName: pkgName},
+				}
+				err := cl.Create(ctx, operator)
+				Expect(err).NotTo(HaveOccurred())
+			})
+			It("sets resolution failure status", func() {
+				By("running reconcile")
+				res, err := reconciler.Reconcile(ctx, ctrl.Request{NamespacedName: opKey})
+				Expect(res).To(Equal(ctrl.Result{}))
+				// TODO: should resolution failure return an error?
```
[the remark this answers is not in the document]
joelanford: +1, I've got that fixed up in #118. When this merges, I'll rebase that one.

controllers/operator_controller.go
joelanford: ^ Exactly - That problem will be resolved in #118.

joelanford:
@tmshort anything else you want to review/discuss on this one?

tmshort:
> @tmshort anything else you want to review/discuss on this one?

Nope, I fixed the tests in #119, and got my head around the conditions, I think you're good.

Are you concerned about the go-apidiff failure (assuming not, since this is still in development)?

/lgtm

joelanford:
Not concerned about the apidiff failure. While we're in development, the purpose of that is to inform us of the changes we're making to the public API, and potentially have discussion about it. Honestly, the main thing I'm watching for at this stage is unnecessarily adding new surface area to the public API. Not really at all concerns about removals or changes at this point.

PR #116, opened 2023-02-02
reduce RBAC permissions of operator-controller

joelanford:
Another facet of #107 

operator-controller does not (and should not ever) need permission to create/update/patch/delete Operator objects. It needs:
- write on Operator status
- write on Operator finalizers
- read on all other operator fields


tmshort (approved):
/approve

These smaller, scoped, PRs are easier to review! :)

Issue #83, opened 2023-01-05
Tie `ENVTEST_K8S_VERSION` to the version of `k8s.io/client-go`


joelanford:
Something like this: https://github.com/operator-framework/helm-operator-plugins/blob/dd32bac92534db6a00fe30252cd944c361639ec5/Makefile#L41-L46

_Originally posted by @joelanford in https://github.com/operator-framework/operator-controller/pull/81#discussion_r1061080134_

joelanford:
Closed by #103

PR #117, opened 2023-02-02
Makefile: remove unused ENVTEST_K8S_VERSION variable

joelanford:
It looks like this should have been removed as part of the fix for #83, which seems to have already made it in in #103


tmshort (approved):
/approve
LGTM

PR #118, opened 2023-02-03
Reduce reconcile complexity

joelanford:
This PR reduces the [cyclomatic complexity](https://en.wikipedia.org/wiki/Cyclomatic_complexity) of the `reconcile` method of the operator controller from 8 to 5 (as measured by gocyclo), mainly by factoring out the bundle entity lookup from the solution into a separate function.

It also ensures all code paths set the `Ready` condition, which required two new reasons (`BundleLookupFailed` and `BundleDeploymentFailed`)

This is built on top of #115, so
/hold
until that has merged and we've rebased this one.

### DIFF
--- a/api/v1alpha1/operator_types.go
+++ b/api/v1alpha1/operator_types.go
@@ -17,8 +17,9 @@ limitations under the License.
 package v1alpha1
 
 import (
-	operatorutil "github.com/operator-framework/operator-controller/internal/util"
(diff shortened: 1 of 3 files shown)


tmshort (approved):
/lgtm

Removing the loop, and doing a complete `SetStatusCondition()` in every error location makes this a lot clearer.

PR #119, opened 2023-02-03
Fix `make test`

tmshort:
`make test` runs all the tests: `e2e` and `test-unit` The `test-e2e` is only the test; it does not include setup and teardown, whereas `e2e` does.


joelanford (approved):
/lgtm

PR #120, opened 2023-02-08
Add targets to Makefile

varshaprasad96:
This commit adds two additional Makefile targets to Makefile:

`cleanup`: Since operator-controller is the single point of entry and `make run` installs Rukpak and Cert-manager, it would be good to have a target that cleans up the cluster of controller and all its dependencies.

`run-local`: To be able to run the controller locally, it is useful for developments so that we need not always have to build image to test operator. This is similar to `make run` target in SDK generated projects (https://github.com/operator-framework/operator-sdk/blob/5cbdad9209332043b7c730856b6302edc8996faf/testdata/go/v3/memcached-operator/Makefile#L118)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -97,8 +97,11 @@ kind-cluster-cleanup: kind ## Delete the kind cluster
 build: manifests generate fmt vet ## Build manager binary.
 	go build -o bin/manager main.go
 
+.PHONY: run-image
+run-image: docker-build kind-load install-cert-mgr install-rukpak install deploy wait ## Build the operator-controller then deploy it into an existing kind cluster.
+
 .PHONY: run
-run: docker-build kind-cluster kind-load cert-mgr rukpak install deploy wait ## Build the operator-controller then deploy it into a new kind cluster.
+run: kind-cluster run-image ## Build the operator-controller then deploy it into a new kind cluster.
 
 .PHONY: run-local
 run-local: manifests generate fmt vet  ## Run a controller from your host. Make sure that necessary CRDs are installed in the cluster.
@@ -143,15 +146,23 @@ ifndef ignore-not-found
 endif
 
 ## TODO dfranz: replace cert-mgr and rukpak targets with our chosen method of distribution when available
-.PHONY: cert-mgr
-cert-mgr: ## Install cert-manager
+.PHONY: install-cert-mgr
+install-cert-mgr: ## Install cert-manager
 	kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/$(CERT_MGR_VERSION)/cert-manager.yaml
 	kubectl wait --for=condition=Available --namespace=cert-manager deployment/cert-manager-webhook --timeout=$(WAIT_TIMEOUT)
 
-.PHONY: rukpak
-rukpak: ## Install rukpak
+.PHONY: uninstall-cert-mgr
+uninstall-cert-mgr: ## Install cert-manager
+	kubectl delete -f https://github.com/cert-manager/cert-manager/releases/download/$(CERT_MGR_VERSION)/cert-manager.yaml
+
+.PHONY: install-rukpak
+install-rukpak: ## Install rukpak
 	kubectl apply -f https://github.com/operator-framework/rukpak/releases/latest/download/rukpak.yaml
 	kubectl wait --for=condition=Available --namespace=rukpak-system deployment/core --timeout=$(WAIT_TIMEOUT)
+
+.PHONY: uninstall-rukpak
+uninstall-rukpak: ## Install rukpak
+	kubectl delete -f https://github.com/operator-framework/rukpak/releases/latest/download/rukpak.yaml
 
 .PHONY: install
 install: manifests kustomize ## Install CRDs into the K8s cluster specified in ~/.kube/config.
@@ -172,9 +183,7 @@ undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/confi
 
 # cleans up all the operator resources and their dependents like cert-manager and rukpak
 .PHONY: cleanup
-cleanup: undeploy uninstall
-	kubectl delete --ignore-not-found=$(ignore-not-found) -f https://github.com/cert-manager/cert-manager/releases/download/$(CERT_MGR_VERSION)/cert-manager.yaml
-	kubectl delete --ignore-not-found=$(ignore-not-found) -f https://github.com/operator-framework/rukpak/releases/latest/download/rukpak.yaml
+cleanup: undeploy uninstall uninstall-rukpak uninstall-cert-mgr 
 
 ##@ Build Dependencies
 
--- a/Makefile
+++ b/Makefile
@@ -100,6 +100,10 @@ build: manifests generate fmt vet ## Build manager binary.
 .PHONY: run
 run: docker-build kind-cluster kind-load cert-mgr rukpak install deploy wait ## Build the operator-controller then deploy it into a new kind cluster.
 
+.PHONY: run-local
+run-local: manifests generate fmt vet  ## Run a controller from your host. Make sure that necessary CRDs are installed in the cluster.
+	go run ./main.go
+
 .PHONY: wait
 wait:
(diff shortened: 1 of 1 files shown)


varshaprasad96:
cc: @dtfranz @tmshort could you please have a look.

Makefile
```diff
@@ -100,6 +100,10 @@ build: manifests generate fmt vet ## Build manager binary.
 .PHONY: run
 run: docker-build kind-cluster kind-load cert-mgr rukpak install deploy wait ## Build the operator-controller then deploy it into a new kind cluster.
 
+.PHONY: run-local
+run-local: manifests generate fmt vet  ## Run a controller from your host. Make sure that necessary CRDs are installed in the cluster.
```
tmshort, line 104: Would it make sense to do an `install` here, so that CRDs are install on the default cluster?

Makefile
```diff
@@ -166,6 +170,12 @@ deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in
 undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/config. Call with ignore-not-found=true to ignore resource not found errors during deletion.
 	$(KUSTOMIZE) build config/default | kubectl delete --ignore-not-found=$(ignore-not-found) -f -
 
+# cleans up all the operator resources and their dependents like cert-manager and rukpak
+.PHONY: cleanup
+cleanup: undeploy uninstall
+	kubectl delete --ignore-not-found=$(ignore-not-found) -f https://github.com/cert-manager/cert-manager/releases/download/$(CERT_MGR_VERSION)/cert-manager.yaml
+	kubectl delete --ignore-not-found=$(ignore-not-found) -f https://github.com/operator-framework/rukpak/releases/latest/download/rukpak.yaml
```
tmshort, line 177: Wondering if these should be targets instead, and rename the `rukpak` and `cert-mgr` targets?
e.g.
`rukpak` -> `install-rukpak`
`cert-mgr` -> `install-cert-manager`

Then, add `uninstall-rukpak` and `uninstall-cert-mgr` recipes and then as dependencies to `cleanup`?

Makefile
varshaprasad96: I was thinking about it, then realized that we need rukpak and cert-manager CRDs too. Which I why I excluded everything instead of selectively installing only operator CRDs.

Makefile
varshaprasad96: That's a good idea! Thoughts on bundling cert-manager and rukpak together in a single install and uninstall target?
Because cert-manager is a requirement for Rukpak, we won't have to install cert-manager alone ever (I guess).

tmshort: Probably keep them separate, fine-grained control is more flexible.

joelanford: for cleanup, what about just deleting the kind cluster?

joelanford: Is there a use case for actually deleting individual dependencies or subcomponents? If not, my preference is to do the simplest/smallest thing that devs will _actually_ use. Whenever I contribute to an unfamiliar project, I'm pretty put off when there are tons and tons of make targets because its never clear what is useful to me and what isn't without building a full mental picture of the whole thing.

joelanford: But `make run` always creates a fresh cluster, so un-doing `make run` is to delete the cluster. We don't really support a developer experience around deploying the current state of this project in a pre-existing cluster. Doing that open's up a pandora's box of complexity, mainly around pushing/pulling images in an easy/repeatable way for different developers.

Ultimately the reason for `make run` working the way it does is that we (a) want an easy button to get everything deployed from the current state of the repo and (b) want to be opinionated about the cluster used in dev/test to make sure developers are working against the same environments. We use `kind` so that we can:
1. pin to a specific kubernetes version
2. have mostly hermetic dev/test environments.
3. load local images into the cluster's image store without having to push them to a remote registry.
4. not have to deal with supporting a whole bunch of finnicky developer workflows that have different requirements for cluster interactions.

`operator-sdk olm install` only ever deals with OLM releases, so it doesn't need to worry about the same sorts of things we worry about for dev/test. Images are already in an official public repo).

But backing up for a second, what are we actually trying to accomplish with this? In rukpak, I personally never needed to do a cleanup on its own, and just stop there. Is it "I want a fast dev cycle so rebuild/redeploy a code change and do it as quickly as possible"? If so, I can see keeping the kind cluster around and I definitely think that's valuable. I have a script on the side that does that for rukpak development.

tmshort: I agree with @joelanford  in that if you have `make X` you should also have a `make undo-X` (or equivalent). So, in the case here, there are separate installs for `rukpak` and `cert-mgr`, and subsequently should be separate uninstalls. If these installs were combined into a single recipe, then the uninstall should also be a separate recipe.

That being said, where @joelanford and I disagree is on the granularity of the makefile. I personally prefer multiple targets with simple, small recipes that can be combined.

joelanford: @tmshort I think that _can_ be okay. But a Makefile with `uninstall`, `undeploy` and `cleanup` is confusing. For the "off the beaten path" targets, I'd suggest more descriptive names.

And I'd reserve short, catch-all names for the "easy button" things that require no extra finnagling.

And lastly, I think I'd somehow highlight the easy button targets, perhaps via a separate section at the top of `make help`

joelanford: So I guess I'm okay with a bunch of targets if we make sure its apparent which of them are the primary ones we expect contributors to use. Once they're comfortable with those, then they can dig more into the Makefile and use bits and pieces of the secondary targets to fine-tune their workflows.

varshaprasad96: @joelanford @tmshort Got it, I agree that if we are if we are considering this target to do the opposite of `make run`, then it should tear the cluster down. 

Though as a developer (IMO), I feel its still useful to have an option to entirely remove the installed olm components instead of forcefully shutting the cluster down and creating a new one every time. This is just for faster cycles of development, where I don't care of the existing polluted cluster state (unless it interferes causing issues) - (though spinning up a cluster doesn't take much effort, I would avoid doing it repeatedly for every small cycle of development). 
But as it may not be valuable to make it as a separate Makefile target and can end up causing more confusion, I'll modify the PR to just tear down the kind cluster.

joelanford: That's what I was getting at with my question @varshaprasad96. Would it make sense to have a target that:
- build the images
- loads them into kind
- kubectl delete --ignore-not-present the manifest
- kubectl apply the manifest

Then `make run` becomes:
- setup the kind cluster
- do the above

varshaprasad96: So something like this:

```Makefile
install-rukpak:

install-certmgr:

uninstall-rukpak:

uninstall-certmgr:

run-image:
	docker-build kind-load install deploy

run: 
    kind-cluster run-image 

clean:
	kind delete <cluster>

run-local: manifests generate fmt vet
         go run ./main.go
```

Use cases:

For developer, they could:
1. For running it in an existing cluster: `make run-image`
2. For running operator-controller locally: `make install-certmgr`,`make install-rukpak`,`make install`,`make run-local`


For developer who were running operator controller locally or would like to clean existing cluster:
- `make uninstall` `make uninstall-rukpak` `make uninstall-certmgr`

For a someone trying to run OLM in a single shot:
- `make run` - does everything for you.
- `make clean` - tear down the cluster

varshaprasad96: As @tmshort suggested, we could bundle up `make run-local` to be:
```
run-local: manifests generate fmt vet install-certmgr install-rukpak install
         go run ./main.go
```
to make it easier for devs, or leave it as is so that they are aware of what components are being installed (that way its transparent on what to uninstall). Its a tradeoff, I'm fine either way

perdasilva:
@varshaprasad96 I think these changes are really nice, could I ask you to please rebase?

varshaprasad96:
Modified and rebased the PR to address changes according to https://github.com/operator-framework/operator-controller/pull/120#discussion_r1103229092

tmshort:
Rebase needed again

tmshort:
@varshaprasad96 this needs a rebase. Do you plan to continue this work? It's been several months.

m1kola:
This PR has been silent for several months. I'm closing it, but please don't hesitate to re-open, if needed.

PR #122, opened 2023-02-10
api: add tests that ensure types and reasons are registered correctly

joelanford:
We currently export constants for the universe of types and reasons that the operator controller uses to set conditions. And we append them to a list that is used in reconciler tests to ensure that `Reconcile` sets each condition type during any given call. See [here](https://github.com/operator-framework/operator-controller/blob/520219c1c4812d3446c85cf8383de3572ebbbf9d/api/v1alpha1/operator_types.go#L32-L54).

However there's still a small gap. It's possible to add a new type/reason without appending it into the respective operatorutil list, which means the assumptions we make (i.e. that the operatorutil lists are exhaustive) are not enforced. And that means our invariant test for "all types must be set at the end of a reconcile call" might produce false negatives.

This PR closes that gap by adding a test that ensures that:
1. For every constant that begins with `Type` or `Reason`, its value exists in the respective operatorutil slice.
2. For every value in the operatorutil slices, there exists a respective `Type` or `Reason` prefixed constant.

### DIFF
--- a/api/v1alpha1/operator_types_test.go
+++ b/api/v1alpha1/operator_types_test.go
@@ -1,0 +1,93 @@
+package v1alpha1_test
+
+import (
+	"fmt"
+	"go/ast"
+	"go/parser"
+	"go/token"
+	"io/fs"
+	"strconv"
+	"strings"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	operatorutil "github.com/operator-framework/operator-controller/internal/util"
+)
+
+var _ = Describe("OperatorTypes", func() {
+	Describe("Condition Type and Reason constants", func() {
+		It("should register types in operatorutil.ConditionTypes", func() {
+			types, err := parseConstants("Type")
+			Expect(err).NotTo(HaveOccurred())
+
+			for _, t := range types {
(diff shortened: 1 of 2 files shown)


api/v1alpha1/operator_types_test.go
```diff
@@ -0,0 +1,83 @@
...
+				Expect(r).To(BeElementOf(operatorutil.ConditionReasons), "Append Reason%s to operatorutil.ConditionReasons in this package's init function.", r)
+			}
+			for _, r := range operatorutil.ConditionReasons {
+				Expect(r).To(BeElementOf(reasons), "There must be a Reason%[1]s string literal constant for reason %[1]q (i.e. 'const Reason%[1]s = %[1]q')", r)
+			}
+		})
+	})
+})
+
+func parseConstants(prefix string) ([]string, error) {
```
perdasilva, line 45: nice! could be cool to add a couple of comments for those unfamiliar with the the ast package

perdasilva (approved):
lgtm - just one small non-blocking suggestion

perdasilva:
@joelanford could you please rebase?

joelanford:
Done! Thanks for the review @perdasilva!

api/v1alpha1/operator_types_test.go
perdasilva: :man_cook: :kiss:

PR #124, opened 2023-02-10
Introduce Pull Request Template

awgreene:
Signed-off-by: Alexander Greene <greene.al1991@gmail.com>

### DIFF
--- a/.github/workflows/pull_request_template.md
+++ b/.github/workflows/pull_request_template.md
@@ -1,0 +1,14 @@
+# Description
+
+<!--
+Thank you for your contribution!
+
+Please provide a summary of the changes and the motivation behind the change.
+-->
+
+## Reviewer Checklist
+
+- [ ] API Go Documentation
+- [ ] Tests: Unit Tests (and E2E Tests, if appropriate)
+- [ ] Comprehensive Commit Messages
+- [ ] Links to related GitHub Issue(s)


anik120 (commented):
/lgtm

.github/workflows/pull_request_template.md
```diff
@@ -0,0 +1,13 @@
+# Description
+
+Thank you for your contribution!
+
+Please provide a summary of the changes and the motivation behind the change.
+
+## Reviewer Checklist
+
+- [ ] Documentation
```
joelanford, line 9: This seems a bit non-actionable at the moment, right? If we want commiters to write docs, we probably first need to boostrap some docs and write something about what requires docs, where to put them, etc.

.github/workflows/pull_request_template.md
```diff
@@ -0,0 +1,13 @@
...
+- [ ] Unit Tests
+- [ ] E2E Tests
+- [ ] Commits are prefixed with one of [`Bug Fix:` , `Feature:`, `Doc:`]
```
joelanford, line 12: Do we _really_ care about commit prefixes at this point? Do we have automation already that:
- enforces this
- uses it for something (i assume the purpose is to build changelogs)

Does this mean we're ready to start forcing PRs to do just one of thses things?

.github/workflows/pull_request_template.md
awgreene: I can create a `/docs` dir in this commit, do we really need a fully fleshed out doc with examples?

awgreene: Having a designated spot and allowing someone to add to it when appropriate seems reasonable

.github/workflows/pull_request_template.md
joelanford: +1 on removing it commit prefix stuff from existing milestone, IMO.

.github/workflows/pull_request_template.md
joelanford: I think some structure or at least some guidelines about structure is important too. Otherwise, we'll end up with a whole bunch of stuff duplicated or in random-ish places. I say this from experience with rukpak docs evolution.

Maybe we start with something that talks about ensuring Go Docs exist for the CRD API fields. I honestly feel like we would get a huge amount of mileage if that's _all_ we documented, for now/starting out.

We can simultaneously figure out a docs strategy around OLMv1, but I feel like that deserves more planning and attention.

.github/workflows/pull_request_template.md
```diff
@@ -0,0 +1,13 @@
...
+
+Thank you for your contribution!
+
+Please provide a summary of the changes and the motivation behind the change.
+
+## Reviewer Checklist
+
+- [ ] Documentation
+- [ ] Unit Tests
+- [ ] E2E Tests
```
joelanford, line 11: I'm not sure I'd include E2E tests in this list. I think we should _heavily_ push contributors/reviewers towards unit tests.

Obviously we'll need e2e's, but I don't think it makes sense on a PR-to-PR basis.

.github/workflows/pull_request_template.md
awgreene: Removed.

.github/workflows/pull_request_template.md
awgreene: Thoughts on changing this to `- [ ] E2E Tests (if appropriate)`

.github/workflows/pull_request_template.md
awgreene: Removed the `getting-started doc` commit.

awgreene: Updated the documentation check for Go Doc exists for CRD api.

.github/workflows/pull_request_template.md
joelanford: WDYT about condensing down to this?

```
- [ ] Tests: unit tests (and e2es, if necessary)
```

In most cases I think we'll want unit tests only. So what's the best way to get that point across, while also reminding contributors and reviewers?

This:
- [x] Unit tests
- [ ] E2e tests (if appropriate)

Or this:
- [x] Tests: unit tests (and e2es, if appropriate)

The former might give the impression that a "good" PR checks all the boxes.
The latter might give the impression that a "good" PR just has to add some tests, doesn't matter which.

In fact a good PR will use _just_ a unit test if that's all that's required to test the change. And we'll only need an e2e to test the main happy paths or other functionality that just can't technically be covered in a unit test.

I think I lean toward the latter option, personally. Once we have a `CONTRIBUTING.md` we could have a section about testing, and we could then update the PR template to link the word `appropriate` to the CONTRIBUTING.md section that talks about what's appropriate for e2es.

As a follow-up, should we add a `CONTRIBUTING.md` that describes all this stuff in more detail?

joelanford (approved):
Looks good pending what others think about the discussion of a single line item for testing vs separate ones for unit/e2e.

I stated my preference for a single line item, but it isn't a strong preference or objectively better. So I'm good either way.

/lgtm

.github/workflows/pull_request_template.md
```diff
@@ -0,0 +1,11 @@
+# Description
+
+Thank you for your contribution!
+
+Please provide a summary of the changes and the motivation behind the change.
```
tmshort, line 5: I think you want to put this into a comment, so that it's not automatically pulled into the PR description without being deleted.
See: https://raw.githubusercontent.com/openssl/openssl/master/.github/PULL_REQUEST_TEMPLATE.md

tmshort (changes requested):

.github/workflows/pull_request_template.md
awgreene: Done!

awgreene: Done!

tmshort (approved):
/lgtm

ankitathomas (approved):
/lgtm

PR #125, opened 2023-02-10
[controller] resurface bundledeployment errors

varshaprasad96:
This PR enables the operator object to resurface the errors in Rukpak by setting its conditions as "BundleDeploymentFailed" if any of the BD errors.


controllers/operator_controller.go
```diff
@@ -277,3 +289,21 @@ func (r *OperatorReconciler) existingBundleDeploymentUnstructured(ctx context.Co
 	}
 	return &unstructured.Unstructured{Object: unstrExistingBundleDeploymentObj}, nil
 }
+
+func (r *OperatorReconciler) verifySuccessfulBundleDeployments(ctx context.Context, name string) error {
+	bundleDeploymentExists := &rukpakv1alpha1.BundleDeployment{}
+	if err := r.Client.Get(ctx, types.NamespacedName{Name: name}, bundleDeploymentExists); err != nil {
+		return fmt.Errorf("bundleDeployment %q not found %v", name, err)
```
joelanford, line 296: This could give us stale data since we've _just_ potentially patched the BD. We should use the BD updated via the `ensureBundleDeployment` function.

And I think we need to tweak that to handle the case where we don't actually make a patch call by making sure the desired BD is updated to point to the existing BD.

controllers/operator_controller.go
```diff
@@ -277,3 +289,21 @@ func (r *OperatorReconciler) existingBundleDeploymentUnstructured(ctx context.Co
...
+	}
+
+	conditions := bundleDeploymentExists.Status.Conditions
+	ers := []error{}
+
+	for _, c := range conditions {
+		if c.Status == metav1.ConditionFalse {
+			ers = append(ers, fmt.Errorf("error observed by Rukpak: %v", c.Message))
+		}
+	}
```
joelanford, line 306: I don't think it's safe to assume that all Conditions impact the determination of whether the Operator Ready condition is failing.

I also don't think its safe to assume that a `False` condition indicates an error (e.g. what if we add `Progressing` as a new condition type to rukpak)

I think what we should probably do here is specifically check for known condition types and implement logic based on the various possibilities.

test/e2e/install_test.go
```diff
@@ -64,4 +68,45 @@ var _ = Describe("Operator Install", func() {
 		err = c.Delete(ctx, operator)
 		Expect(err).ToNot(HaveOccurred())
 	})
+
+	It("errors when Rukpak is not able to install the operator", func() {
```
joelanford, line 72: Could we move this test to `controllers/operator_controller_test.go`? We don't actually need rukpak running for this, right? We just need to create BDs with various  status condition types/reason/message combos so that we exercise the operator controller's code paths.

The motivation is to reserve e2e tests for things that really do require the full system running.

controllers/operator_controller.go
varshaprasad96: Tweaked it a bit to check the existing and desired BD after patch.

controllers/operator_controller.go
varshaprasad96: For now I'm checking only the `TypeHasValidBundle` and `TypeInstalled` conditions to be false explicitly. Is it sufficient, or can there ever be a situation in rukpak where the status are left `Unknown` for which we need to error?

joelanford: Would it work to check for `!= True` for now?

I think we'll want to tighten this up to handle all those possibilities though. E.g. if installed and had valid bundle are both unknown, then ready should also be unknown.

controllers/operator_controller.go
```diff
@@ -257,10 +257,16 @@ func (r *OperatorReconciler) ensureBundleDeployment(ctx context.Context, desired
 	}
 
 	// If the existing BD already has everything that the desired BD has, no need to contact the API server.
+	// Make sure the status of the existingBD from the server is as expected.
 	if equality.Semantic.DeepDerivative(desiredBundleDeployment, existingBundleDeployment) {
-		return nil
+		return verifySuccessfulBDStatus(existingBundleDeployment)
```
joelanford, line 262: I was thinking more along these lines:
```go
*desiredBundleDeployment = *existingBundleDeployment
return nil
```

And then have a separate function that maps a BD status to a Operator ready condition. So we wouldn't be verifying success so much as we'd be saying: "this BD status results in this Operator Ready condition"

WDYT?

controllers/operator_controller.go
```diff
@@ -277,3 +283,30 @@ func (r *OperatorReconciler) existingBundleDeploymentUnstructured(ctx context.Co
...
+	}
+
+	for _, condition := range conditions {
+		// Currently we are checking for only these two types, if any other condition is added to Rukpak, validate accordingly.
+		if condition.Type == rukpakv1alpha1.TypeHasValidBundle || condition.Type == rukpakv1alpha1.TypeInstalled {
+			if condition.Status == metav1.ConditionFalse {
+				ers = append(ers, fmt.Errorf("error observed by Rukpak: %v", condition.Message))
+			}
+		}
+	}
```
joelanford, line 310: I wonder if we should have a series of specific short circuiting checks. e.g.

```
if condBundle == nil && condInstalled == nil {
    // Ready = false (or maybe unknown?), Message: waiting for BD to be reconciled
}
if condInstalled != nil && condInstalled.Status == True { 
  // Ready == True, Message: condInstalled.Message
}
if condInstalled != nil && condInstalled.Status == False { 
  // Ready == False, Message: condInstalled.Message
}
if condBundle!= nil && condBundle.Status == False { 
  // Ready == False, Message: condBundle.Message
}
// Unknown
```

controllers/operator_controller.go
varshaprasad96: It does, but just wanted to make sure if `Unknown` status is something for which we want to error for? Usually its an intermediate status before either failing or succeeding. But if there is a case where the status is just left Unknown, then we should be specifically checking that and erroring.

joelanford: See my other comment about thinking of this more like a mapping of BD status to Operator Ready. Not "is this BD status an error or not". Maybe that helps?

varshaprasad96: @joelanford modified the PR based on the review comments. 

I ran a few scenarios and have some questions regarding this:

1. Previously, when we were inferring a false status as error, we were also returning the error on the controller, not just updating the status. Currently, the controller does not surface any error when bundleDeployment fails (imo it should) instead just silently only updates the status.
2. When a request is requeued for an Unknown condition, we are running the resolver again which is not necessary and could be an expensive operation by itself. Instead what we actually need, is to just wait for the rukpak to finish the steps for creating bundleDeployment. An option is to have a random backoff with requeueAfter (a fixed interval of time), but I’m not sure if that is still a valid approach to avoid running reconciliation for the same state again. 
3. Do we need to requeue when the operator’s Ready status is false if we are not triggering it with a non-nil error? I think yes, because in future there maybe a change in catalogsource’s cache to let’s say include the package we want that changes resolver's entities. But it again comes back to Q1, where the user will have no idea on why the request is getting requeued again when we do not return a valid error.

joelanford: > Currently, the controller does not surface any error when bundleDeployment fails (imo it should) instead just silently only updates the status.

I think this is actually okay. In my mind, if the operator reconcile actually fails in some way, it makes sense to return an error. But if we did everything we we're supposed to do successfully, and another component (in this case rukpak) is failing in some way, we can reflect that other component's error in our status, and then just wait to reconcile until we get another event on the the BD, the assumptions being that:
1. If we did everything successfully this time, its pretty likely we'll do everything successfully again if we reconcile immediately again with no other change in state (ergo `requeue: true` is likely to keep resulting in more `requeue: true` until some other state actually changes).
2. The other component will likely be the thing returning an error, re-reconciling, and updating the status of the underlying thing. We'll see those status updates and reconcile again based on that.

> Instead what we actually need, is to just wait for the rukpak to finish the steps for creating bundleDeployment.

Exactly. The way we wait is to requeue events on the BD (with out `Owns(BD)` watch) as rukpak progresses on it and updates its status accordingly.

> Do we need to requeue when the operator’s Ready status is false if we are not triggering it with a non-nil error? I think yes, because in future there maybe a change in catalogsource’s cache to let’s say include the package we want that changes resolver's input entities.

I think a better approach might be to have a way for the catalogsource cache to signal: "Hey, I've received a change. You should requeue _all_ of the BDs"

controllers/operator_controller.go
```diff
@@ -162,14 +162,35 @@ func (r *OperatorReconciler) reconcile(ctx context.Context, op *operatorsv1alpha
 		return ctrl.Result{}, err
 	}
 
-	// update operator status
+	// convert existing unstructured object into bundleDeployment for easier mapping of status.
+	existingTypedBundleDeployment := &rukpakv1alpha1.BundleDeployment{}
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(dep.UnstructuredContent(), existingTypedBundleDeployment); err != nil {
+		return ctrl.Result{}, err
```
joelanford, line 168: We need to set the `Ready` status condition here before returning the error.

controllers/operator_controller.go
```diff
@@ -162,14 +162,35 @@ func (r *OperatorReconciler) reconcile(ctx context.Context, op *operatorsv1alpha
...
+	// 2. If the Operator "Ready" status is "True": Update the "successful resolution" status and return the result.
+	// 3. If the Operator "Ready" status is "False": There is error observed from Rukpak. Update the status accordingly.
+	status, message := verifyBDStatus(existingTypedBundleDeployment)
+	var reason string
+	if status == metav1.ConditionTrue || status == metav1.ConditionUnknown {
+		reason = operatorsv1alpha1.ReasonResolutionSucceeded
+	} else {
+		reason = operatorsv1alpha1.ReasonBundleDeploymentFailed
+	}
+
```
joelanford, line 182: Nit: Put this in it's own function, and have that function return the entire `metav1.Condition` object?

Then the `apimeta.SetStatusCondition` below looks a bit simpler:
```go
readyCondition := mapBDStatusToReadyCondition(existingBD)
apimeta.SetStatusCondition(&op.Status.Conditions, readyCondition)
```

controllers/operator_controller.go
```diff
@@ -162,14 +162,35 @@ func (r *OperatorReconciler) reconcile(ctx context.Context, op *operatorsv1alpha
...
+	// update operator status:
+	// 1. If the Operator "Ready" status is "Unknown": The status of successful bundleDeployment is unknown, requeue the request.
+	// 2. If the Operator "Ready" status is "True": Update the "successful resolution" status and return the result.
+	// 3. If the Operator "Ready" status is "False": There is error observed from Rukpak. Update the status accordingly.
+	status, message := verifyBDStatus(existingTypedBundleDeployment)
+	var reason string
+	if status == metav1.ConditionTrue || status == metav1.ConditionUnknown {
+		reason = operatorsv1alpha1.ReasonResolutionSucceeded
```
joelanford, line 178: I'm thinking `ReasonResolutionSucceeded` is no longer the right thing. We might need to rename this to something like `ReasonInstallSucceeded`

varshaprasad96: +1! Not sure if we want to do in this PR, but I've created a separate [commit](https://github.com/operator-framework/operator-controller/pull/125/commits/543b7a4d26d21d5ff8a800884ca3b6917c6def7d) with the change of `ReasonResolutionSucceeded` to `ReasonInstallationSucceeded`. If we want to remove it out of this PR, I can create a follow-up. We could make another iteration of re-naming condition type and reasons for addressing https://github.com/operator-framework/operator-controller/issues/126 in a follow-up.

controllers/operator_controller.go
```diff
@@ -162,14 +162,21 @@ func (r *OperatorReconciler) reconcile(ctx context.Context, op *operatorsv1alpha
...
-		Reason:             operatorsv1alpha1.ReasonResolutionSucceeded,
-		Message:            "resolution was successful",
-		ObservedGeneration: op.GetGeneration(),
-	})
+	// convert existing unstructured object into bundleDeployment for easier mapping of status.
+	existingTypedBundleDeployment := &rukpakv1alpha1.BundleDeployment{}
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(dep.UnstructuredContent(), existingTypedBundleDeployment); err != nil {
+		apimeta.SetStatusCondition(&op.Status.Conditions, metav1.Condition{
+			Type:               operatorsv1alpha1.TypeReady,
+			Status:             metav1.ConditionUnknown,
```
varshaprasad96, line 170: Leaving it to `Unknown` instead of `False`, since an error in converting to typed object does not necessarily mean that install was unsuccessful.

controllers/operator_controller_test.go
```diff
@@ -293,10 +300,239 @@ var _ = Describe("Reconcile Test", func() {
...
+				Expect(cond.Status).To(Equal(metav1.ConditionUnknown))
+				Expect(cond.Reason).To(Equal(operatorsv1alpha1.ReasonInstallationSucceeded))
+				Expect(cond.Message).To(ContainSubstring(`waiting for rukpak to install bundleDeployment successfully`))
+			})
+		})
 		AfterEach(func() {
 			verifyInvariants(ctx, operator)
 
-			err := cl.Delete(ctx, operator)
+			err := cl.DeleteAllOf(ctx, operator)
```
joelanford, line 535: Curious about this change. Does DeleteAllOf do something different?

One thing I realize... We probably need to be cleaning up the BD. Garbage collection doesn't run in envtest.

controllers/operator_controller_test.go
```diff
@@ -112,11 +112,13 @@ var _ = Describe("Reconcile Test", func() {
 					Expect(bd.Spec.Template.Spec.Source.Image.Ref).To(Equal("quay.io/operatorhubio/prometheus@sha256:5b04c49d8d3eff6a338b56ec90bdf491d501fe301c9cdfb740e5bff6769a21ed"))
 				})
 				It("sets resolution success status", func() {
-					cond := apimeta.FindStatusCondition(operator.Status.Conditions, operatorsv1alpha1.TypeReady)
-					Expect(cond).NotTo(BeNil())
-					Expect(cond.Status).To(Equal(metav1.ConditionTrue))
-					Expect(cond.Reason).To(Equal(operatorsv1alpha1.ReasonResolutionSucceeded))
-					Expect(cond.Message).To(Equal("resolution was successful"))
+					Eventually(func() {
```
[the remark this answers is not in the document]
varshaprasad96: Yes, when the bundleDeployment is first created the conditions do on appear immediately so it moves to `Unknown` state. Removing `Eventually` can cause race conditions wherein the test fails existing at Unknown state.

controllers/operator_controller.go
```diff
@@ -277,3 +287,50 @@ func (r *OperatorReconciler) existingBundleDeploymentUnstructured(ctx context.Co
...
+
+// mapBDStatusToReadyCondition returns the operator object's "TypeReady" condition based on the bundle deployment statuses.
+func mapBDStatusToReadyCondition(existingBD *rukpakv1alpha1.BundleDeployment, observedGeneration int64) metav1.Condition {
+	// update operator status:
+	// 1. If the Operator "Ready" status is "Unknown": The status of successful bundleDeployment is unknown, wait till Rukpak updates the BD status.
+	// 2. If the Operator "Ready" status is "True": Update the "successful resolution" status and return the result.
+	// 3. If the Operator "Ready" status is "False": There is error observed from Rukpak. Update the status accordingly.
+	status, message := verifyBDStatus(existingBD)
+	var reason string
+	if status == metav1.ConditionTrue || status == metav1.ConditionUnknown {
```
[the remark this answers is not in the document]
varshaprasad96: I'm fine either way, since its either saying that we don't know whether the state is going to successful or not. A better approach would be have a new reason, I can do that in follow up.

controllers/operator_controller_test.go
varshaprasad96: In one of the tests where the creation of duplicate package is tested, we create another operator object (https://github.com/operator-framework/operator-controller/blob/520219c1c4812d3446c85cf8383de3572ebbbf9d/controllers/operator_controller_test.go#L267)  which is left behind, that is why modified it to be `DeleteAllOf` to be on the safer side of cleaning up all operator objects.

That's a good point, though the owner references are set, we would still to delete BD, will modify the PR.

controllers/operator_controller_test.go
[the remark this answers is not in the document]
varshaprasad96: @joelanford  When the first time reconciler results, usually the status of the Operator is `Unknown` because bundle deployment would not have immediately updated its `HasValidBundle` and `InstallSuccessful` status. Eventually doesn't do anything here. I can tweak all other the unit tests to remove conditions checks, and have them in one place (where we verify if bundle status is being translated as right operator status)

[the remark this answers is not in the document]
joelanford: These tests don't run a controller, so it's up to us to setup the preconditions for each scenario.

Scenario 1: there is no BD, we expect one to be created if resolution succeeds

Scenario 2: there is a BD, but it's spec is wrong. We expect the BD to be updated.

Scenario 3: there is a BD with a specific status. We expect the Operator Ready condition to reflect that status

Other variations of Scenario 3...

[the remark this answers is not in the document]
joelanford: So I'd expect the condition checks to remain. Just no need for the Eventually wrapper.

controllers/operator_controller_test.go
joelanford: Ah. In that case, we should have a separate AfterEach (at the same level as that BeforeEach) that deletes that other Operator.

controllers/operator_controller_test.go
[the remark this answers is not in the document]
varshaprasad96: In each of the scenarios, when the BD is created it initially does not have any conditions set (or it maybe set but Unknown). Which is why the operator returns an Unknown condition, since BD creation is still in progress. As and when the status in BD changes, the corresponding Operator status changes. In the controller, we are checking the BD status as soon as it is created and returning the controller result. The operator status will move from Unknown (Waiting) --> True/False/Unknown. We would not be necessarily getting a true status immediately after the first reconcile because of which the current tests (where BD does not exist before) which check the operator ready condition will fail. 

Agree, eventually wrapper doesn't make sense. But the conditions checks need to be removed.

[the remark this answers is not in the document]
joelanford: In the scenarios where the BD does not exist or does not have a status before Reconcile starts, we still have deterministic output of the Ready condition state though. We should still be asserting on whatever that output is.

controllers/operator_controller.go
```diff
@@ -162,14 +162,21 @@ func (r *OperatorReconciler) reconcile(ctx context.Context, op *operatorsv1alpha
...
-		Message:            "resolution was successful",
-		ObservedGeneration: op.GetGeneration(),
-	})
+	// convert existing unstructured object into bundleDeployment for easier mapping of status.
+	existingTypedBundleDeployment := &rukpakv1alpha1.BundleDeployment{}
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(dep.UnstructuredContent(), existingTypedBundleDeployment); err != nil {
+		apimeta.SetStatusCondition(&op.Status.Conditions, metav1.Condition{
+			Type:               operatorsv1alpha1.TypeReady,
+			Status:             metav1.ConditionUnknown,
+			Reason:             operatorsv1alpha1.ReasonInstallationSucceeded,
```
joelanford, line 171: For follow-up: change this to BD failed or maybe another new reason

controllers/operator_controller.go
```diff
@@ -277,3 +287,51 @@ func (r *OperatorReconciler) existingBundleDeploymentUnstructured(ctx context.Co
 	}
 	return &unstructured.Unstructured{Object: unstrExistingBundleDeploymentObj}, nil
 }
+
+// verifyBDStatus reads the various possibilities of status in bundle deployment and translates
+// into corresponding operator condition status and message.
+func verifyBDStatus(dep *rukpakv1alpha1.BundleDeployment) (metav1.ConditionStatus, string) {
+	isValidBundleCond := apimeta.FindStatusCondition(dep.Status.Conditions, rukpakv1alpha1.TypeHasValidBundle)
+	isInstalledCond := apimeta.FindStatusCondition(dep.Status.Conditions, rukpakv1alpha1.TypeInstalled)
```
joelanford, line 295: For follow up: add code to handle case where these conditions are out of date (based on comparison of bd.metadata.generation and bd.status.observedGeneration).

controllers/operator_controller.go
```diff
@@ -277,3 +287,51 @@ func (r *OperatorReconciler) existingBundleDeploymentUnstructured(ctx context.Co
...
+	// 1. If the Operator "Ready" status is "Unknown": The status of successful bundleDeployment is unknown, wait till Rukpak updates the BD status.
+	// 2. If the Operator "Ready" status is "True": Update the "successful resolution" status and return the result.
+	// 3. If the Operator "Ready" status is "False": There is error observed from Rukpak. Update the status accordingly.
+	status, message := verifyBDStatus(existingBD)
+	var reason string
+	// TODO: introduce a new reason for condition Unknown, instead of defaulting it to Installation Succeeded.
+	if status == metav1.ConditionTrue {
+		reason = operatorsv1alpha1.ReasonInstallationSucceeded
+	} else {
+		reason = operatorsv1alpha1.ReasonBundleDeploymentFailed
```
joelanford, line 327: For follow-up: don't leak rukpak names in Operator API (i.e. change this to "InstallationFailed")

joelanford (approved):
Great work! I've jotted down a few follow-ups we discussed, but I think we should merge this and tackle those separately.

/lgtm

varshaprasad96:
Have created an issue for follow-ups https://github.com/operator-framework/operator-controller/issues/127. Will have a PR ready for those as soon as this merges.

awgreene (approved):
/approve
Looks good!

PR #130, opened 2023-02-20
Modify Operator Conditions

varshaprasad96:
This PR brings in 2 changes:

1. Introduce a new reason for Unknown status: Currently, we specify BundleDeploymentFailure or InstallationSuccess as reasons for an unknown status. This means that we are not sure if the install is successfull or we aren't sure of a failure either. To make it more accurate, we introduce a new reason "InstallStatusUnknown" to mean that the status cannot be determined at that particular instant.

2. Rename "BundleDeploymentFailed" to "InstallationUnsuccessful": This prevents leaking of Rukpak API names. BundleDeployment Failed can be one of the reason for unsuccessful installation.


api/v1alpha1/operator_types.go
```diff
@@ -33,10 +33,11 @@ const (
...
 	TypeReady = "Ready"
 
-	ReasonInstallationSucceeded  = "InstallationSucceeded"
-	ReasonResolutionFailed       = "ResolutionFailed"
-	ReasonBundleLookupFailed     = "BundleLookupFailed"
-	ReasonBundleDeploymentFailed = "BundleDeploymentFailed"
+	ReasonInstallationSucceeded     = "InstallationSucceeded"
+	ReasonResolutionFailed          = "ResolutionFailed"
+	ReasonBundleLookupFailed        = "BundleLookupFailed"
+	ReasonInstallationFailed        = "ReasonInstallationFailed"
```
joelanford, line 39:
```suggestion
	ReasonInstallationFailed        = "InstallationFailed"
```

controllers/operator_controller.go
```diff
@@ -323,8 +323,10 @@ func mapBDStatusToReadyCondition(existingBD *rukpakv1alpha1.BundleDeployment, ob
 	// TODO: introduce a new reason for condition Unknown, instead of defaulting it to Installation Succeeded.
```
joelanford, line 323: Remove this TODO now?

controllers/operator_controller.go
```diff
@@ -323,8 +323,10 @@ func mapBDStatusToReadyCondition(existingBD *rukpakv1alpha1.BundleDeployment, ob
...
 	if status == metav1.ConditionTrue {
 		reason = operatorsv1alpha1.ReasonInstallationSucceeded
+	} else if status == metav1.ConditionUnknown {
+		reason = operatorsv1alpha1.ReasonInstallationStatusUnknown
 	} else {
-		reason = operatorsv1alpha1.ReasonBundleDeploymentFailed
+		reason = operatorsv1alpha1.ReasonInstallationFailed
 	}
```
joelanford, line 330: WDYT about changing this to a switch/case structure with cases for True and False, and then a default case which would catch Unknown and any other unexpected status value?

api/v1alpha1/operator_types.go
varshaprasad96: Sorry, my bad!

controllers/operator_controller.go
varshaprasad96: @joelanford Thoughts on defaulting to False instead of Unknown? The only possible set of values for the status could be True, False, Unknown or nil (unless new states are introduced which is out of scope for now). In [verifyBDStatus](https://github.com/operator-framework/operator-controller/blob/a6a8091d0680cbd4114cfcd8e1f6b9baa2c29f14/controllers/operator_controller.go#L293) we make sure that it defaults to Unknown for any unexpected case. The only other unexpected value could be empty status, for which the API server would anyway give a [validation](https://github.com/kubernetes/apimachinery/blob/117bd9b56ec3fe6b4891d86097971e49bf9f68f4/pkg/apis/meta/v1/types.go#L1365) error. Given the various scenarios we have mapped, an empty status definitely means that bundle is not installed in cluster. Unknown seems reasonable, but given that we know definitely that the bundle is not present in cluster isn't it better to set false.

perdasilva:
Could you please rebase @varshaprasad96 - I can override the go-apidiff.

varshaprasad96:
@perdasilva done! Thanks :)

Issue #132, opened 2023-02-22
Create release pipeline for operator-controller


dtfranz:
In order to create the release pipeline for OLM v1, each component needs its own release pipeline, including the operator-controller.

Acceptance criteria:
- github release action is created for operator-controller
- operator-controller can be installed from the resulting quay image

dtfranz:
Addressing in #133

dtfranz:
Closed with #133

Issue #109, opened 2023-02-01
Create installation yaml for operator-controller


dtfranz:
Each component of OLM v1 should be installed through a consistent format. RukPak, for instance, is currently installed using a single downloadable yaml file ([here](https://github.com/operator-framework/rukpak/releases/download/v0.11.0/rukpak.yaml)). We should create a similar if not identical installation method for the operator-controller. If we choose to go another route, an additional issue must be created to do the same for RukPak (and probably cert-manager too). 

Acceptance criteria:
- operator-controller can be installed from a single downloadable yaml file.

Note that the output of this issue should not install dependencies such as RukPak itself, just as RukPak does not install its dependencies through the same yaml. This should be taken care of in the next issue: #110 Also note that installing the operator-controller on its own through this yaml without installing the dependencies will not result in a successful installation. This information should be easily available so that users don't install this yaml only hoping for a successful OLM V1 installation. We should also have links to documentation for #110 available.

Possible idea: Install operator-controller via BundleDeployment

dtfranz:
Addressing in #133

dtfranz:
Closed with #133

Issue #110, opened 2023-02-01
Create install command to install operator-controller and all dependencies


dtfranz:
In order to facilitate an easy release process, we should develop a simple method of installation for the operator-controller as well as its dependencies. So far we have proposed doing this through a shell script to allow this with a single command, similar to the golangcli-lint install command here:
```bash
curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.50.1
```
One alternative idea was to package the yaml files into a single OCI image then install through that.

Acceptance criteria:
- operator-controller and all dependencies may be installed with a single command
- ~~Should exist in the repo created in #111~~ Obsolete: The operator-controller repo will be the source-of-truth for OLM v1 releases, install steps, etc.
- Must install Cert-Manager, RukPak, and the operator-controller. Each component should be installed successfully and reach a running state before moving to the next component.
- ~~The script should default to installing the latest released version of each component, but allow users to specify released versions for each component.~~ Obsolete: A release of operator-controller should have the versions of each component enforced in the go.mod file, Makefile, etc.

PR #133, opened 2023-02-23
Operator Controller Release

dtfranz:
Adds a release github action and implements goreleaser, borrowing heavily from rukpak but stripped down a bit.

See my fork here for an example release output: https://github.com/dtfranz/operator-controller/releases/tag/v0.0.2

Some of the additional build targets found in rukpak have been stripped out to keep this smaller.

Addresses #132 and #109

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,4 +1,0 @@
-# More info: https://docs.docker.com/engine/reference/builder/#dockerignore-file
-# Ignore build and test binaries.
-bin/
-testbin/
--- a/.goreleaser.template.yml
+++ b/.goreleaser.template.yml
@@ -1,0 +1,50 @@
+env:
+- GOPROXY=https://proxy.golang.org|direct
+- GO111MODULE=on
+- CGO_ENABLED=0
+before:
+  hooks:
+    - go mod tidy
+    - go mod download
+builds:
+  - id: operator-controller
+    main: ./
+    binary: bin/manager
+    tags: $GO_BUILD_TAGS
+    goos:
+    - linux
+    ldflags:
+    - -X main.Version={{ .Version }}
+dockers:
+- image_templates:
+  - "{{ .Env.IMAGE_REPO }}:{{ .Env.IMAGE_TAG }}"
+  dockerfile: Dockerfile
+  goos: linux
+docker_manifests:
+- name_template: "{{ .Env.IMAGE_REPO }}:{{ .Env.IMAGE_TAG }}"
+  image_templates:
+  - "{{ .Env.IMAGE_REPO }}:{{ .Env.IMAGE_TAG }}"
+checksum:
+  name_template: 'checksums.txt'
+snapshot:
+  name_template: "{{ incpatch .Version }}-next"
+changelog:
+  use: github-native
+  skip: $DISABLE_RELEASE_PIPELINE
+release:
+  disable: $DISABLE_RELEASE_PIPELINE
+  extra_files:
+  - glob: 'operator-controller.yaml'
+  header: |
+    ## Installation
+
+    ```bash
+    kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/{{ .Env.CERT_MGR_VERSION }}/cert-manager.yaml
+    kubectl wait --for=condition=Available --namespace=cert-manager deployment/cert-manager-webhook --timeout=60s
+    kubectl apply -f https://github.com/operator-framework/rukpak/releases/latest/download/rukpak.yaml
+    kubectl wait --for=condition=Available --namespace=rukpak-system deployment/core --timeout=60s
+    kubectl wait --for=condition=Available --namespace=rukpak-system deployment/helm-provisioner --timeout=60s
+    kubectl wait --for=condition=Available --namespace=rukpak-system deployment/rukpak-webhooks --timeout=60s
+    kubectl apply -f https://github.com/operator-framework/operator-controller/releases/download/{{ .Tag }}/operator-controller.yaml
+    kubectl wait --for=condition=Available --namespace=operator-controller-system deployment/operator-controller-controller-manager --timeout=60s
+    ```
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,34 +1,14 @@
-# Build the manager binary
-FROM golang:1.19 as builder
-ARG TARGETOS
-ARG TARGETARCH
-
-WORKDIR /workspace
-# Copy the Go Modules manifests
-COPY go.mod go.mod
-COPY go.sum go.sum
-# cache deps before building and copying source so that we don't need to re-download as much
-# and so that source changes don't invalidate our downloaded layer
-RUN go mod download
-
-# Copy the go source
-COPY main.go main.go
-COPY api/ api/
-COPY controllers/ controllers/
-COPY internal/ internal/
-
-# Build
-# the GOARCH has not a default value to allow the binary be built according to the host where the command
-# was called. For example, if we call make docker-build in a local env which has the Apple Silicon M1 SO
-# the docker BUILDPLATFORM arg will be linux/arm64 when for Apple x86 it will be linux/amd64. Therefore,
-# by leaving it empty we can ensure that the container and binary shipped on it will have the same platform.
-RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -o manager main.go
-
-# Use distroless as minimal base image to package the manager binary
-# Refer to https://github.com/GoogleContainerTools/distroless for more details
+# Note: This dockerfile does not build the binaries 
+# required and is intended to be built only with the
+# 'make build' or 'make release' targets.
 FROM gcr.io/distroless/static:nonroot
+
 WORKDIR /
-COPY --from=builder /workspace/manager .
+
+COPY manager manager
+
+EXPOSE 8080
+
 USER 65532:65532
 
 ENTRYPOINT ["/manager"]
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,7 @@ export IMAGE_REPO ?= quay.io/operator-framework/operator-controller
 export IMAGE_TAG ?= devel
 export GO_BUILD_TAGS ?= upstream
 export CERT_MGR_VERSION ?= v1.9.0
+export GORELEASER_VERSION ?= v1.15.2
 export WAIT_TIMEOUT ?= 60s
 IMG?=$(IMAGE_REPO):$(IMAGE_TAG)
 
@@ -94,8 +95,8 @@ kind-cluster-cleanup: kind ## Delete the kind cluster
 ##@ Build
 
 .PHONY: build
-build: manifests generate fmt vet ## Build manager binary.
-	go build -o bin/manager main.go
+build: manifests generate fmt vet substitute goreleaser ## Build manager binary using goreleaser for current GOOS and GOARCH.
+	${GORELEASER} build ${GORELEASER_ARGS} --single-target -o bin/manager
 
 .PHONY: run
 run: docker-build kind-cluster kind-load cert-mgr rukpak install deploy wait ## Build the operator-controller then deploy it into a new kind cluster.
@@ -108,12 +109,9 @@ wait:
 # (i.e. docker build --platform linux/arm64 ). However, you must enable docker buildKit for it.
 # More info: https://docs.docker.com/develop/develop-images/build_enhancements/
 .PHONY: docker-build
-docker-build: generate ## Build docker image with the operator-controller.
-	docker build -t ${IMG} .
-
-.PHONY: docker-push
-docker-push: ## Push docker image with the manager.
-	docker push ${IMG}
+docker-build: export GOOS=linux
+docker-build: build ## Build docker image with the operator-controller.
+	docker build -t ${IMG} -f Dockerfile ./bin/
 
 # PLATFORMS defines the target platforms for  the manager image be build to provide support to multiple
 # architectures. (i.e. make docker-buildx IMG=myregistry/mypoperator:0.0.1). To use this option you need to:
@@ -131,6 +129,24 @@ docker-buildx: test ## Build and push docker image for the manager for cross-pla
 	- docker buildx build --push --platform=$(PLATFORMS) --tag ${IMG} -f Dockerfile.cross .
 	- docker buildx rm project-v3-builder
 	rm Dockerfile.cross
+
+###########
+# Release #
+###########
+
+##@ Release:
+export DISABLE_RELEASE_PIPELINE ?= true
+export GORELEASER_ARGS ?= --snapshot --rm-dist
+
+substitute:
+	envsubst < .goreleaser.template.yml > .goreleaser.yml
+
+release: substitute goreleaser ## Runs goreleaser for the operator-controller. By default, this will run only as a snapshot and will not publish any artifacts unless it is run with different arguments. To override the arguments, run with "GORELEASER_ARGS=...". When run as a github action from a tag, this target will publish a full release.
+	$(GORELEASER) $(GORELEASER_ARGS)
+
+quickstart: VERSION ?= $(shell git describe --abbrev=0 --tags)
+quickstart: kustomize generate ## Generate the installation release manifests
+	kubectl kustomize config/default | sed "s/:devel/:$(VERSION)/g" > operator-controller.yaml
 
 ##@ Deployment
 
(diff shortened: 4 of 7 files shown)


Makefile
```diff
@@ -132,6 +132,24 @@ docker-buildx: test ## Build and push docker image for the manager for cross-pla
 	- docker buildx rm project-v3-builder
 	rm Dockerfile.cross
 
+###########
```
perdasilva, line 135: can we get some `make help` docs around release - when/how it should be used?

internal/version/version.go
```diff
@@ -0,0 +1,8 @@
+package version
+
+// GitCommit indicates which commit the rukpak binaries were built from
+var GitCommit string
```
perdasilva, line 4: does the value get injected at build time? is it worth adding a comment?

perdasilva (approved):
lgtm - I just had a couple of small questions/suggestions. Nothing blocking.

Dockerfile
```diff
@@ -1,34 +1,7 @@
-# Build the manager binary
-FROM golang:1.19 as builder
-ARG TARGETOS
-ARG TARGETARCH
+FROM gcr.io/distroless/static:debug-nonroot
```
joelanford, line 1: I don't think we need the debug variant for operator-controller do we?

```suggestion
FROM gcr.io/distroless/static:nonroot
```

Dockerfile
```diff
@@ -1,34 +1,7 @@
...
-# Use distroless as minimal base image to package the manager binary
-# Refer to https://github.com/GoogleContainerTools/distroless for more details
-FROM gcr.io/distroless/static:nonroot
 WORKDIR /
-COPY --from=builder /workspace/manager .
-USER 65532:65532
 
-ENTRYPOINT ["/manager"]
+COPY manager manager
+
```
joelanford, line 6: Do we need to explicitly specify the user? Honestly not sure. I think my suggestion would be to specify it explicitly rather than inherit it from the base image.

We should also specify the entrypoint since this is the only binary we'll be using in this image.
```suggestion
USER 65532:65532
ENTRYPOINT ["/manager"]
```

Makefile
```diff
@@ -95,10 +95,10 @@ kind-cluster-cleanup: kind ## Delete the kind cluster
 
 .PHONY: build
 build: manifests generate fmt vet ## Build manager binary.
-	go build -o bin/manager main.go
+	CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -o bin/manager main.go
```
joelanford, line 98: I discovered [`goreleaser build`](https://goreleaser.com/cmd/goreleaser_build/#synopsis) recently. WDYT about using that instead? The big benefit is that we essentially get local builds that use the exact same configuration as the release builds (same env, same ldflags, etc.)

Makefile
```diff
@@ -109,7 +109,7 @@ wait:
 # More info: https://docs.docker.com/develop/develop-images/build_enhancements/
 .PHONY: docker-build
 docker-build: generate ## Build docker image with the operator-controller.
-	docker build -t ${IMG} .
+	docker build -t ${IMG} -f Dockerfile ./bin/
```
joelanford, line 112: This seems like it'll fail because the Dockerfile expects the binary to be pre-built by goreleaser. The previous Dockerfile included the binary build as a pre-step in a multi-stage build.

I think we need to add `build` to the list of pre-reqs?

Makefile
```diff
@@ -109,7 +109,7 @@ wait:
...
 
 .PHONY: docker-push
 docker-push: ## Push docker image with the manager.
```
joelanford, line 115: WDYT about removing `docker-push`? We don't use it anywhere else, and it isn't part of the normal dev flow.

Makefile
```diff
@@ -132,6 +132,24 @@ docker-buildx: test ## Build and push docker image for the manager for cross-pla
...
 
+###########
+# Release #
+###########
+
+##@ release:
+
+export DISABLE_RELEASE_PIPELINE ?= true
+substitute:
+	envsubst < .goreleaser.template.yml > .goreleaser.yml
```
joelanford, line 143: Is this used anywhere other than `release`? If not, I'd propose to consolidate this into the `release` recipe.

Having a top-level target called `substitute` seems like it could be both too vague and potentially confusing for contributors wondering if/when they should use it.

internal/version/version.go
```diff
@@ -0,0 +1,8 @@
...
+
+func String() string {
+	return GitCommit
+}
```
joelanford, line 8: I don't see this used anywhere. Remove it? Seems like anything that needs to know the git commit can just use `version.GitCommit` directly.

joelanford (commented):
Looks pretty good overall! The Rukpak setup mostly works, but I think it was somewhat rushed and probably is in need of some fine tuning.

I left _some_ comments along those lines, but there's probably more to do as well. Regardless, I think further improvements could be done in one or more follow-ons.

Makefile
```diff
@@ -132,6 +132,24 @@ docker-buildx: test ## Build and push docker image for the manager for cross-pla
...
+
+release: GORELEASER_ARGS ?= --snapshot --rm-dist
+release: goreleaser substitute ## Run goreleaser
+	$(GORELEASER) $(GORELEASER_ARGS)
+
+quickstart: VERSION ?= $(shell git describe --abbrev=0 --tags)
+quickstart: generate ## Generate the installation release manifests
+	kubectl kustomize config/default | sed "s/:latest/:$(VERSION)/g" > operator-controller.yaml
```
joelanford, line 151: Kustomize can do image ref substitution directly. We may want to consider whether that is a viable alternative. If we aren't very careful, this approach could silently replace _other_ image refs in our manifests if they happen to use `:latest`.

Really _anything_ referencing `latest` in our manifests would be a problem, so maybe in practice this would be fine. But I'd rather not compound that problem if it does somehow happen.

Makefile
joelanford: Also, the binary built in this case needs to be GOOS=linux, but in the non-docker build case should just take the default settings of the environment. That way, mac and linux machines can both build local binaries and container image binaries correctly.

internal/version/version.go
joelanford: +1.

Also, I don't see this used anywhere else in the Go binary. Maybe we should take this out of this PR and create a separate issue to remind us to get version info into the binary and have a way to print it out.

IIRC the go compiler as of 1.18 or 1.19 now injects this git info automatically. We might not need to do it ourselves anymore.

.goreleaser.template.yml
```diff
@@ -0,0 +1,51 @@
...
+  - "{{ .Env.IMAGE_REPO }}:{{ .Env.IMAGE_TAG }}"
+checksum:
+  name_template: 'checksums.txt'
+snapshot:
+  name_template: "{{ incpatch .Version }}-next"
+changelog:
+  use: github-native
+  skip: $DISABLE_RELEASE_PIPELINE
+release:
+  disable: $DISABLE_RELEASE_PIPELINE
```
joelanford, line 36: I'm trying to remember why we do this convoluted envsubst thing rather than use the `--snapshot` flag when invoking goreleaser.

.goreleaser.template.yml
```diff
@@ -0,0 +1,51 @@
...
+    ```bash
+    kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/{{ .Env.CERT_MGR_VERSION }}/cert-manager.yaml
+    kubectl wait --for=condition=Available --namespace=cert-manager deployment/cert-manager-webhook --timeout=60s
+    kubectl apply -f https://github.com/operator-framework/rukpak/releases/latest/download/rukpak.yaml
+    kubectl wait --for=condition=Available --namespace=rukpak-system deployment/core --timeout=60s
+    kubectl wait --for=condition=Available --namespace=rukpak-system deployment/helm-provisioner --timeout=60s
+    kubectl wait --for=condition=Available --namespace=rukpak-system deployment/rukpak-webhooks --timeout=60s
+    kubectl apply -f https://github.com/operator-framework/operator-controller/releases/download/{{ .Tag }}/operator-controller.yaml
+    kubectl wait --for=condition=Available --namespace=operator-controller-system deployment/operator-controller --timeout=60s
+    ```
```
joelanford, line 51: This is getting somewhat long. I suppose this is basically our "distribution", right? I wonder if we should think about consolidating this into a script and then having the header be a one-liner that shows how to `curl | sh` the script? (could be a follow-on)

dtfranz: Yeah, this is basically the distribution method at the moment. We have an upcoming issue though that should accomplish what you're asking for here: #110

Makefile
dtfranz: Yeah I can definitely see that being a problem. Let me fix this up to make sure we're just using the substitutions provided by kustomize.

Makefile
dtfranz: I've updated the `build` target to use `goreleaser build`, and I believe that's also resolved your concern here. By default, when running `make build`, goreleaser will set GOARCH and GOOS to the user's setup. This won't apply though when building for release.

Makefile
dtfranz: OK, with the latest push I have this being done through kustomize using the `VERSION` env set during release. Let me know what you think :)

Makefile
joelanford: For `docker-build` I _think_ we need to explicitly set GOOS=linux, right? If we don't won't a macOS user end up with a docker container that contains a darwin binary?

Dockerfile
```diff
@@ -1,34 +1,11 @@
-# Build the manager binary
-FROM golang:1.19 as builder
-ARG TARGETOS
-ARG TARGETARCH
+FROM gcr.io/distroless/static:nonroot
```
joelanford, line 1: Can we add a comment to this file that says it's intended to be used with `make docker-build` or `make release`?

Makefile
```diff
@@ -194,6 +207,11 @@ ginkgo: $(GINKGO) ## Download ginkgo locally if necessary.
 $(GINKGO): $(LOCALBIN)
 	test -s $(LOCALBIN)/ginkgo || GOBIN=$(LOCALBIN) go install github.com/onsi/ginkgo/v2/ginkgo@v2.1.4
 
+.PHONY: goreleaser
+goreleaser: $(GORELEASER) ## Builds a local copy of goreleaser
+$(GORELEASER): $(LOCALBIN)
+	test -s $(LOCALBIN)/goreleaser || GOBIN=$(LOCALBIN) go install github.com/goreleaser/goreleaser@v1.11.2
```
joelanford, line 213: For follow-up, but I'm noticing right now -- we should move LOCALBIN somewhere else because the docker build copies all these build tool binaries into the build context unnecessarily, which makes builds take longer.

Rukpak puts these in `./tools/bin` and reserves `./bin` for binaries built from the repo itself.

dtfranz: Great point, I'll create a follow-up issue to take care of both these items.

dtfranz: Added #135

dtfranz:
@joelanford @perdasilva I believe I've addressed all your comments, please take another look when you have a moment.

joelanford:
e2e is still failing for me on my Mac. To fix it for me, I added this line, directly above the declaration of the `docker-build` target:
```
docker-build: export GOOS=linux
```

Makefile
tmshort: TBH, building within the container generally improves compatibility within the container and improves build reproducibility.

.dockerignore
```diff
@@ -1,4 +0,0 @@
-# More info: https://docs.docker.com/engine/reference/builder/#dockerignore-file
-# Ignore build and test binaries.
-bin/
-testbin/
```
tmshort, line 4: Why delete?

dtfranz:
> e2e is still failing for me on my Mac. To fix it for me, I added this line, directly above the declaration of the `docker-build` target:
> 
> ```
> docker-build: export GOOS=linux
> ```

I've pushed up your fix, please pull and test again when you can :)

joelanford:
Still failing for me on macOS (arm64, but I don't think that matters). After I run `make e2e`, I see this:
```
$ file ./bin/manager
./bin/manager: Mach-O 64-bit executable arm64
```

So I'm still getting a darwin binary. Same fix from before still fixes it for me.

.goreleaser.template.yml
joelanford: Ah I remember: we want to build/push images/manifests -- but NOT release -- for builds of the main branch.

It seems like this problem is half solved. It's possible to template `release.disable`, but still not possible to template `changelog.skip`.

I opened an issue in goreleaser to see if we can make `changelog.skip` templatable as well: https://github.com/goreleaser/goreleaser/issues/3828

dtfranz:
> Still failing for me on macOS (arm64, but I don't think that matters). After I run `make e2e`, I see this:
> 
> ```
> $ file ./bin/manager
> ./bin/manager: Mach-O 64-bit executable arm64
> ```
> 
> So I'm still getting a darwin binary. Same fix from before still fixes it for me.

Strange, it's not clear to me why that fix is necessary since the goos is set to linux in the goreleaser template for both the binary and docker build. I've updated the branch again with your fix in the meantime.

joelanford:
> Strange, it's not clear to me why that fix is necessary since the goos is set to linux in the goreleaser template for both the binary and docker build.

I think the `goreleaser build --single-target` _always_ uses GOOS and GOARCH, regardless of what's in the goreleaser file.

As far as the docker build, the config in the goreleaser file is only used during a release. The `docker-build` Makefile target doesn't use the goreleaser config at all. 

This docker build stuff could _maybe_ be improved some as well (maybe the `docker-build` target should also invoke goreleaser somehow? But that seems better for a follow-up.

.goreleaser.template.yml
joelanford: Sweet. The maintainer already merged a PR that fixes this. Once there's a release, we can directly reference `{{ .Env.DISABLE_RELEASE_PIPELINE }}` as a template value in the actual goreleaser config file. No need for the envsubst shenanigans after that.

.goreleaser.template.yml
```diff
@@ -0,0 +1,50 @@
...
+  hooks:
+    - go mod tidy
+    - go mod download
+builds:
+  - id: operator-controller
+    main: ./
+    binary: bin/manager
+    tags: $GO_BUILD_TAGS
+    goos:
+    - linux
```
joelanford, line 15: Not sure if GH ate my comment or if I managed not to submit it.

This seems to build only a single binary for `linux/$GOARCH`. But we need to build binaries for the following:
- `linux/amd64`
- `linux/arm64`
- `linux/ppc64le`
- `linux/s390x`

joelanford: We then need to propagate these to separate individual docker images, and then build the manifest list that combines all of them.

joelanford: All this is probably fine to do in a follow-up, but might be good to do that before we tag this repo the first time.

Makefile
joelanford: GoReleaser doesn't support that, so if we want binary builds happening inside docker, we can't use goreleaser I don't think.

I don't love this restriction, but my main interests are:
- we have exactly one way to build the the binaries snd containers 
- the binaries we build ARE the binaries that end up in the container.

GoReleaser can be made to ensure reproducibility (outside of Go compiler version) so I think we end up at the same place. It just takes a bit more care to get there.

joelanford (approved):
/lgtm

Nice work! A few more things are better left for smaller follow-ups:
1. Multiarch binaries and containers
2. Remove the `envsubst` and goreleaser template file when there's a new release of goreleaser
3. Move tools binaries to separate directory.
4. Investigate calling goreleaser as part of the `docker-build` target.

Anything I missed?

dtfranz:
> /lgtm
> 
> Nice work! A few more things are better left for smaller follow-ups:
> 
>     1. Multiarch binaries and containers
> 
>     2. Remove the `envsubst` and goreleaser template file when there's a new release of goreleaser
> 
>     3. Move tools binaries to separate directory.
> 
>     4. Investigate calling goreleaser as part of the `docker-build` target.
> 
> 
> Anything I missed?

Thanks @joelanford! Yes, I think that about sums it up. 
I did some testing with all the latest changes and found some issues when running the releaser in my fork. goreleaser doesn't play well with using kustomize to modify the manifest files, so after thinking about it probably for too long and not seeing a clean and easy solution I went with a compromise of keeping rukpak's `sed` replace for the image tag but used `:devel` instead of `:latest` so we'll be far less likely to hit a tag we didn't intend. If that's a dealbreaker let me know, otherwise I'll merge once I get another LGTM after the latest changes :smiley:

joelanford:
Totally reasonable. I think that's another small cleanup thing we can do once this heavy lift has merged.

/lgtm

Issue #127, opened 2023-02-17
Follow-ups for #125 


varshaprasad96:
#125 enables operator objects to resurface errors when Bundle deployment fails. Some TODOs:

- [ ] Introduce a new reason for `metav1.Unknown` condition: https://github.com/operator-framework/operator-controller/pull/125#discussion_r1109890031
- [ ] Add code to handle the case when conditions are out of date: https://github.com/operator-framework/operator-controller/pull/125#discussion_r1109892499
- [ ] Introduce new reason for failure of Bundle Deployment: https://github.com/operator-framework/operator-controller/pull/125#discussion_r1109894494

varshaprasad96:
@joelanford For (2), had a few questions. Comparing bundle deployment's `status.ObservervedGeneration` and `spec.Generation` seems to be very fragile. `bd.Status.ObserverdGeneration` is supposed to be eventually consistent with the spec when it changes. During the first reconcile, there is very high possibility that rukpak is still working on the BD, and status has not been updated yet. This means the BDs observed status generation will always be less than spec.Generation. It eventually becomes consistent, when the bd is created successfully. This is not necessarily the case when BD is out of date, but when it is just getting created.

The scenario when BD is stale and we end up mapping a stale BD condition to operator as in below should again eventually be correct (though not immediate) when BD changes its status after rukpak has updated it with the new image. 
1. operator is created, bd is installed
2. Operator object or catalog gets updated such that a different image pops out of the resolver
3. Reconcile patches the BD with the new image.
4. The BD already has a status from the initial install.
5. We map that out of date status to the new operator status.

Given this, do we need to handle this case in Operator Controller or Rukpak? Wouldn't it be better if rukpak changes the BD status when it patches it with a new image (with a new condition?) so that we can stamp it in Operator status, than us comparing the status and spec generations?

joelanford:
> During the first reconcile, there is very high possibility that rukpak is still working on the BD, and status has not been updated yet.

Yep, agreed -- except its a 100% chance. During the very first reconcile of an Operator object, its a guarantee that the BD we are looking at after we apply the SSA patch will not have a status. The data populated into the BD by that apply call by the API server is literally just whatever the result of admission webhooks + apiserver metadata changes (uid, generation, resourceVersion, etc).

> This means the BDs observed status generation will always be less than spec.Generation ... This is not necessarily the case when BD is out of date.

[...]

Issue #126, opened 2023-02-13
Rename `Ready` condition?


joelanford:
`Ready` implies that the operator is active, watches have been started, controller is running, any services it includes have live endpoints, etc.

Our Ready condition is not that. Ours is "There is a BD created". And with #125, that will improve to "There is a BD that successfully installed." Rukpak currently does not provide a signal about whether the operator is truly ready.

My proposal:
1. In the Operator API, rename `Ready` to `Installed`.
2. (At some point, not now), start thinking about https://github.com/operator-framework/rukpak/issues/419 (or other alternate solutions for answering the question: "Is this Operator _actually_ ready?"

PR #134, opened 2023-03-01
Conditions: Map out-of-date bundleDeployment with an operator condition

varshaprasad96:
This PR brings in the change that sets the operator condition to Unknown when the BundleDeployment's conditions are out od date.

### DIFF
--- a/controllers/operator_controller.go
+++ b/controllers/operator_controller.go
@@ -294,10 +294,6 @@ func verifyBDStatus(dep *rukpakv1alpha1.BundleDeployment) (metav1.ConditionStatu
 	isValidBundleCond := apimeta.FindStatusCondition(dep.Status.Conditions, rukpakv1alpha1.TypeHasValidBundle)
 	isInstalledCond := apimeta.FindStatusCondition(dep.Status.Conditions, rukpakv1alpha1.TypeInstalled)
 
-	if isValidBundleCond == nil && isInstalledCond == nil {
-		return metav1.ConditionUnknown, fmt.Sprintf("waiting for bundleDeployment %q status to be updated", dep.Name)
-	}
-
 	if isValidBundleCond != nil && isValidBundleCond.Status == metav1.ConditionFalse {
 		return metav1.ConditionFalse, isValidBundleCond.Message
 	}
@@ -309,11 +305,21 @@ func verifyBDStatus(dep *rukpakv1alpha1.BundleDeployment) (metav1.ConditionStatu
 	if isInstalledCond != nil && isInstalledCond.Status == metav1.ConditionTrue {
 		return metav1.ConditionTrue, "install was successful"
 	}
-	return metav1.ConditionUnknown, fmt.Sprintf("could not determine the state of bundleDeployment %s", dep.Name)
+	return metav1.ConditionUnknown, fmt.Sprintf("could not determine the state of BundleDeployment %s", dep.Name)
 }
 
 // mapBDStatusToReadyCondition returns the operator object's "TypeReady" condition based on the bundle deployment statuses.
 func mapBDStatusToReadyCondition(existingBD *rukpakv1alpha1.BundleDeployment, observedGeneration int64) metav1.Condition {
+	// if BundleDeployment status is stale, return an unknown condition.
+	if isBundleDepStale(existingBD) {
+		return metav1.Condition{
+			Type:               operatorsv1alpha1.TypeReady,
+			Status:             metav1.ConditionUnknown,
+			Reason:             operatorsv1alpha1.ReasonInstallationStatusUnknown,
+			Message:            fmt.Sprintf("waiting for BundleDeployment %q status to be updated. BundleDeployment conditions out of date.", existingBD.Name),
+			ObservedGeneration: observedGeneration,
+		}
+	}
 	// update operator status:
 	// 1. If the Operator "Ready" status is "Unknown": The status of successful bundleDeployment is unknown, wait till Rukpak updates the BD status.
 	// 2. If the Operator "Ready" status is "True": Update the "successful resolution" status and return the result.
@@ -337,4 +343,9 @@ func mapBDStatusToReadyCondition(existingBD *rukpakv1alpha1.BundleDeployment, ob
 		Message:            message,
 		ObservedGeneration: observedGeneration,
 	}
+}
+
+// isBundleDepStale returns true if conditions are out of date.
+func isBundleDepStale(bd *rukpakv1alpha1.BundleDeployment) bool {
+	return bd != nil && bd.Status.ObservedGeneration != bd.GetGeneration()
 }
--- a/controllers/operator_controller_test.go
+++ b/controllers/operator_controller_test.go
@@ -121,9 +121,10 @@ var _ = Describe("Reconcile Test", func() {
 				})
 			})
 			When("the expected BundleDeployment already exists", func() {
+				var bd *rukpakv1alpha1.BundleDeployment
 				BeforeEach(func() {
 					By("patching the existing BD")
-					err := cl.Create(ctx, &rukpakv1alpha1.BundleDeployment{
+					bd = &rukpakv1alpha1.BundleDeployment{
 						ObjectMeta: metav1.ObjectMeta{
 							Name: opKey.Name,
 							OwnerReferences: []metav1.OwnerReference{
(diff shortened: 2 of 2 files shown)


controllers/operator_controller.go
```diff
@@ -175,6 +175,17 @@ func (r *OperatorReconciler) reconcile(ctx context.Context, op *operatorsv1alpha
...
+		apimeta.SetStatusCondition(&op.Status.Conditions, metav1.Condition{
+			Type:               operatorsv1alpha1.TypeReady,
+			Status:             metav1.ConditionUnknown,
+			Reason:             operatorsv1alpha1.ReasonInstallationStatusUnknown,
+			Message:            fmt.Sprintf("waiting for bundleDeployment %q status to be updated. BundleDeployment conditions out of date.", existingTypedBundleDeployment.Name),
+			ObservedGeneration: op.GetGeneration(),
+		})
+		return ctrl.Result{}, nil
+	}
+
```
joelanford, line 188: Should we put this inside the `mapBDStatusToReadyCondition` function?

controllers/operator_controller_test.go
```diff
@@ -348,8 +348,31 @@ var _ = Describe("Reconcile Test", func() {
...
+				cond := apimeta.FindStatusCondition(op.Status.Conditions, operatorsv1alpha1.TypeReady)
+				Expect(cond).NotTo(BeNil())
+				Expect(cond.Status).To(Equal(metav1.ConditionUnknown))
+				Expect(cond.Reason).To(Equal(operatorsv1alpha1.ReasonInstallationStatusUnknown))
+				Expect(cond.Message).To(Equal(fmt.Sprintf("waiting for bundleDeployment %q status to be updated. BundleDeployment conditions out of date.", bd.Name)))
+			})
+
 			It("verify operator status when bundle deployment is waiting to be created", func() {
 				By("running reconcile")
+				bd.Status.ObservedGeneration = bd.GetGeneration()
```
varshaprasad96, line 371: This is hacky, but needed since we are trying to mimic >1 reconcile.

controllers/operator_controller.go
```diff
@@ -175,6 +175,17 @@ func (r *OperatorReconciler) reconcile(ctx context.Context, op *operatorsv1alpha
 		return ctrl.Result{}, err
 	}
 
+	if isBundleDepStale(existingTypedBundleDeployment) {
+		apimeta.SetStatusCondition(&op.Status.Conditions, metav1.Condition{
+			Type:               operatorsv1alpha1.TypeReady,
+			Status:             metav1.ConditionUnknown,
+			Reason:             operatorsv1alpha1.ReasonInstallationStatusUnknown,
```
varshaprasad96, line 182: @joelanford (continuing the [discussion](https://github.com/operator-framework/operator-controller/issues/127#issuecomment-1449127217) here) - I am leaning towards leaving the status Unknown. If we for sure know that status is out of date because of an error where it can't anymore change to true, we could make it `False`. My understanding is:

`True` - We know for sure that installation was successful.
`Unknown` - Can't determine the status because either we don't have enough data (BD conditions aren't created), or rukpak is still in progress.
`False` - We know for sure that the package is not available (probably neither will it be in future given the current state persists).

 Which is why I was wondering if we need Rukpak to have a new status for this particular scenario of upgrades, rather than us inferring an out of date bd status. Till we reach to a point of introducing any more status conditions (addressing  #126), could we make this scenario to be "Unknown". 

I'm fine with making it "False" too, but just wanted to check again based on what we discussed [here](https://github.com/operator-framework/operator-controller/pull/130#discussion_r1114626377).

controllers/operator_controller.go
varshaprasad96: Fine either ways, just made it separate to make it clear that we are not mapping the BD status since its stale.

controllers/operator_controller.go
joelanford: +1 on unknown for now.

controllers/operator_controller.go
```diff
@@ -314,6 +314,16 @@ func verifyBDStatus(dep *rukpakv1alpha1.BundleDeployment) (metav1.ConditionStatu
 
 // mapBDStatusToReadyCondition returns the operator object's "TypeReady" condition based on the bundle deployment statuses.
 func mapBDStatusToReadyCondition(existingBD *rukpakv1alpha1.BundleDeployment, observedGeneration int64) metav1.Condition {
+	// if BundleDeployment status is stale, return an unknown condition.
+	if isBundleDepStale(existingBD) {
+		return metav1.Condition{
+			Type:               operatorsv1alpha1.TypeReady,
+			Status:             metav1.ConditionUnknown,
+			Reason:             operatorsv1alpha1.ReasonInstallationStatusUnknown,
+			Message:            fmt.Sprintf("waiting for bundleDeployment %q status to be updated. BundleDeployment conditions out of date.", existingBD.Name),
```
joelanford, line 323: nit: use consistent case on BundleDeployment? I realize the first use is in the middle of a sentence and the second is at the beginning, but we aren't capitalizing "Waiting" in the first sentence.

My suggestion is that if we're talking about a kind, capitalize it always, just like its defined in the API.

```suggestion
			Message:            fmt.Sprintf("waiting for BundleDeployment %q status to be updated. BundleDeployment conditions out of date.", existingBD.Name),
```

controllers/operator_controller.go
```diff
@@ -338,3 +348,8 @@ func mapBDStatusToReadyCondition(existingBD *rukpakv1alpha1.BundleDeployment, ob
 		ObservedGeneration: observedGeneration,
 	}
 }
+
+// isBundleDepStale returns true if conditions are out of date.
+func isBundleDepStale(existingTypedBundleDeployment *rukpakv1alpha1.BundleDeployment) bool {
```
joelanford, line 353: Maybe just call it `bd` inside this function to make it a bit easier to read?

```suggestion
func isBundleDepStale(bd *rukpakv1alpha1.BundleDeployment) bool {
```

controllers/operator_controller_test.go
```diff
@@ -348,8 +348,31 @@ var _ = Describe("Reconcile Test", func() {
 				Expect(err).NotTo(HaveOccurred())
 			})
 
+			It("verify operator status when bundle deployment status is stale while being created", func() {
```
joelanford, line 351: nit:
```suggestion
			It("sets operator ready condition to unknown when bundle deployment status is stale", func() {
```

controllers/operator_controller_test.go
joelanford: A better approach might be to have separate `When` blocks. Something along these lines, maybe?

```go
When("The bundle deployment status is stale", func() {})
When("The bundle deployment status is up-to-date", func() {
	BeforeEach(func() {
		bd.Status.ObservedGeneration = bd.GetGeneration()
		By("updating the status of bundleDeployment")
		err := cl.Status().Update(ctx, bd)
		Expect(err).NotTo(HaveOccurred())
	}
})
```

Also, I find it easier to think about these tests not as mimicking multiple reconciles, but more like "what happens during a given Reconcile when the state of the world is X?". And then we would ideally have tests for all the various X's such that we get really good code coverage in `Reconcile`.

controllers/operator_controller.go
joelanford, line 323: Here, when both conditions are nil, I think we may now have a slightly misleading condition message: https://github.com/operator-framework/operator-controller/blob/cc2c7375262ce0a225ea9d00351ea1bcc498ced3/controllers/operator_controller.go#L297-L299

Before, we were assuming "those conditions not being present means it just hasn't reconciled yet"

Now, we know that the BD has been reconciled because we've made it past the staleness check unscathed. The status is up-to-date, but _still_ the expected conditions aren't present. I'd suggest just removing lines 297-299, and then if both conditions are `nil`, we'll end up at line 312, which seems more appropriate now I think. WDYT?

controllers/operator_controller_test.go
```diff
@@ -374,6 +397,7 @@ var _ = Describe("Reconcile Test", func() {
 					Message: "failed to unpack",
 					Reason:  rukpakv1alpha1.ReasonUnpackFailed,
 				})
+				bd.Status.ObservedGeneration = bd.GetGeneration()
```
joelanford, line 400: Oh yeah, now that I'm seeing this repeated multiple times, it definitely seems ripe for being inside a `BeforeEach` inside a `When("bd status is up-to-date")`

controllers/operator_controller_test.go
varshaprasad96: Made the change, just left the status update call in `It` blocks, so that we update the status only once after setting the right conditions.

controllers/operator_controller_test.go
```diff
@@ -348,180 +348,210 @@ var _ = Describe("Reconcile Test", func() {
 				Expect(err).NotTo(HaveOccurred())
 			})
 
-			It("verify operator status when bundle deployment is waiting to be created", func() {
```
joelanford, line 351: Perhaps as a follow-up. We have a `Describe("the expected bundle deployment already exists")` block at line 123. It seems like all of the tests in the `When("the existing operator status is based on bundleDeployment")` block at line 307 belong inside that block, right? That might reduce the LoC necessary to setup the tests since all of these status mapping tests assume the BD spec is already correct.

WDYT?

joelanford (approved):
/lgtm

controllers/operator_controller_test.go
varshaprasad96: Made the change and removed a duplicate test

controllers/operator_controller_test.go
```diff
@@ -151,37 +152,234 @@ var _ = Describe("Reconcile Test", func() {
...
+					}
+					err := cl.Create(ctx, bd)
 					Expect(err).NotTo(HaveOccurred())
+				})
 
-					By("running reconcile")
-					res, err := reconciler.Reconcile(ctx, ctrl.Request{NamespacedName: opKey})
-					Expect(res).To(Equal(ctrl.Result{}))
-					Expect(err).NotTo(HaveOccurred())
+				When("the BundleDeployment status is stale", func() {
```
joelanford, line 160: I think within the `When("the expected BundleDeployment already exists")`, maybe we need a slightly different organization?

When the BD spec is out of date:
 - we expect to patch (and therefore we expect the status to be outdated after the patch)

When the BD spec is up-to-date:
  - we expect NOT to patch (that's the `PIt` that needs to be implemented with a wrapper or fake client)
  - we expect the various BD status possibilities to be mapped to the appropriate Operator status.

WDYT?

varshaprasad96: Re-arranged the tests in that block for check if the spec is updated.

varshaprasad96:
@joelanford Re-organized the test cases, please feel free to let me know if you would like to see any other changes. I can do those in follow-ups, merging this for now.

Issue #123, opened 2023-02-10
The resolver should treat certain properties/constraints as optional


joelanford:
See https://kubernetes.slack.com/archives/C0181L6JYQ2/p1675349508746979

TL;DR: We should pass a boolean value to [`loadFromEntity`](https://github.com/operator-framework/operator-controller/blob/520219c1c4812d3446c85cf8383de3572ebbbf9d/internal/resolution/variable_sources/entity/bundle_entity.go#L212) to indicate whether the property is required (`true`) or optional (`false`). Currently the function is implemented as if everything it is asked to load is required.

If this new parameter is optional/false, we should return the default value of the type and a nil error.

Lastly, instead of using `true`/`false` at call sites, @perdasilva suggested we define constants (perhaps `required, optional = true, false`), and then use the constant at call sites to make it more obvious what the semantics of the boolean value are.

PR #138, opened 2023-03-22
[entity_resolution] Not error when optional properties are not present in entity

varshaprasad96:
This PR introduces the optional field while loading entities from bundles. If the property is optional and is not present, we do not error.

Fixes: #123


joelanford (approved):
Nice! Thanks @varshaprasad96!

/approve
/lgtm

PR #142, opened 2023-03-28
Add .spec.version

perdasilva:
Adds `.spec.version` property to the Operator CRD. Some small changes were made to the resolver and top level variable sources to accommodate for this change and, hopefully, future ones. When reviewing, we should be asking ourselves how easy it would be to add the next property.

This PR also adds a small admissions unit test, so we can validate the OpenApi validation rules we apply using the kubebuilder annotations.

I've also nerfed the stackdumps down to panic. Otherwise we just get logs pointing to internal controller-runtime code...

### DIFF
--- a/api/v1alpha1/operator_types.go
+++ b/api/v1alpha1/operator_types.go
@@ -29,8 +29,8 @@ type OperatorSpec struct {
 	PackageName string `json:"packageName"`
 
 	//+kubebuilder:validation:MaxLength:=64
-	//+kubebuilder:validation:Pattern:=^(\|\||\s+)?([\s~^><=]*)v?(\d+)(\.(\d+))?(\.(\d+))?(\-(.+))?$
-	//+kubebuilder:optional
+	//+kubebuilder:validation:Pattern=^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(-(0|[1-9]\d*|[0-9]*[a-zA-Z-][0-9a-zA-Z-]*)(\.(0|[1-9]\d*|[0-9]*[a-zA-Z-][0-9a-zA-Z-]*))*)?(\+([0-9a-zA-Z-]+(\.[0-9a-zA-Z-]+)*))?$
+	//+kubebuilder:Optional
 	// Version is an optional is semver constraint on the package version. If not specified, the latest version available of the package will be installed.
 	// If specified, the specific version of the package will be installed so long as it is available in any of the content sources available.
 	// Examples:
@@ -51,6 +51,7 @@ const (
 	ReasonBundleLookupFailed        = "BundleLookupFailed"
 	ReasonInstallationFailed        = "InstallationFailed"
 	ReasonInstallationStatusUnknown = "InstallationStatusUnknown"
+	ReasonInvalidSpec               = "InvalidSpec"
 )
 
 func init() {
@@ -65,6 +66,7 @@ func init() {
 		ReasonBundleLookupFailed,
 		ReasonInstallationFailed,
 		ReasonInstallationStatusUnknown,
+		ReasonInvalidSpec,
 	)
 }
 
--- a/api/v1alpha1/operator_types.go
+++ b/api/v1alpha1/operator_types.go
@@ -31,12 +31,9 @@ type OperatorSpec struct {
 	//+kubebuilder:validation:MaxLength:=64
 	//+kubebuilder:validation:Pattern=^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(-(0|[1-9]\d*|[0-9]*[a-zA-Z-][0-9a-zA-Z-]*)(\.(0|[1-9]\d*|[0-9]*[a-zA-Z-][0-9a-zA-Z-]*))*)?(\+([0-9a-zA-Z-]+(\.[0-9a-zA-Z-]+)*))?$
 	//+kubebuilder:Optional
-	// Version is an optional is semver constraint on the package version. If not specified, the latest version available of the package will be installed.
+	// Version is an optional semver constraint on the package version. If not specified, the latest version available of the package will be installed.
 	// If specified, the specific version of the package will be installed so long as it is available in any of the content sources available.
-	// Examples:
-	//  - 1.2.3
-	//  - 1.0.0-alpha
-	//  - 1.0.0-rc.1
+	// Examples: 1.2.3, 1.0.0-alpha, 1.0.0-rc.1
 	//
 	// For more information on semver, please see https://semver.org/
 	Version string `json:"version,omitempty"`
--- a/api/v1alpha1/operator_types.go
+++ b/api/v1alpha1/operator_types.go
@@ -31,6 +31,14 @@ type OperatorSpec struct {
 	//+kubebuilder:validation:MaxLength:=64
 	//+kubebuilder:validation:Pattern:=^(\|\||\s+)?([\s~^><=]*)v?(\d+)(\.(\d+))?(\.(\d+))?(\-(.+))?$
 	//+kubebuilder:optional
+	// Version is an optional is semver constraint on the package version. If not specified, the latest version available of the package will be installed.
+	// If specified, the specific version of the package will be installed so long as it is available in any of the content sources available.
+	// Examples:
+	//  - 1.2.3
+	//  - 1.0.0-alpha
+	//  - 1.0.0-rc.1
+	//
+	// For more information on semver, please see https://semver.org/
 	Version string `json:"version,omitempty"`
 }
 
--- a/api/v1alpha1/operator_types.go
+++ b/api/v1alpha1/operator_types.go
@@ -27,6 +27,11 @@ type OperatorSpec struct {
 	//+kubebuilder:validation:MaxLength:=48
 	//+kubebuilder:validation:Pattern:=^[a-z0-9]+(-[a-z0-9]+)*$
 	PackageName string `json:"packageName"`
+
+	//+kubebuilder:validation:MaxLength:=64
+	//+kubebuilder:validation:Pattern:=^(\|\||\s+)?([\s~^><=]*)v?(\d+)(\.(\d+))?(\.(\d+))?(\-(.+))?$
+	//+kubebuilder:optional
+	Version string `json:"version,omitempty"`
 }
 
 const (
--- a/config/crd/bases/operators.operatorframework.io_operators.yaml
+++ b/config/crd/bases/operators.operatorframework.io_operators.yaml
@@ -39,6 +39,10 @@ spec:
                 maxLength: 48
                 pattern: ^[a-z0-9]+(-[a-z0-9]+)*$
                 type: string
+              version:
+                maxLength: 64
+                pattern: ^(\|\||\s+)?([\s~^><=]*)v?(\d+)(\.(\d+))?(\.(\d+))?(\-(.+))?$
+                type: string
             required:
             - packageName
             type: object
--- a/config/crd/bases/operators.operatorframework.io_operators.yaml
+++ b/config/crd/bases/operators.operatorframework.io_operators.yaml
@@ -47,7 +47,7 @@ spec:
                   sources available. Examples: - 1.2.3 - 1.0.0-alpha - 1.0.0-rc.1
                   \n For more information on semver, please see https://semver.org/"
                 maxLength: 64
-                pattern: ^(\|\||\s+)?([\s~^><=]*)v?(\d+)(\.(\d+))?(\.(\d+))?(\-(.+))?$
+                pattern: ^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(-(0|[1-9]\d*|[0-9]*[a-zA-Z-][0-9a-zA-Z-]*)(\.(0|[1-9]\d*|[0-9]*[a-zA-Z-][0-9a-zA-Z-]*))*)?(\+([0-9a-zA-Z-]+(\.[0-9a-zA-Z-]+)*))?$
                 type: string
             required:
             - packageName
--- a/config/crd/bases/operators.operatorframework.io_operators.yaml
+++ b/config/crd/bases/operators.operatorframework.io_operators.yaml
@@ -40,12 +40,12 @@ spec:
                 pattern: ^[a-z0-9]+(-[a-z0-9]+)*$
                 type: string
               version:
-                description: "Version is an optional is semver constraint on the package
+                description: "Version is an optional semver constraint on the package
                   version. If not specified, the latest version available of the package
                   will be installed. If specified, the specific version of the package
                   will be installed so long as it is available in any of the content
-                  sources available. Examples: - 1.2.3 - 1.0.0-alpha - 1.0.0-rc.1
-                  \n For more information on semver, please see https://semver.org/"
+                  sources available. Examples: 1.2.3, 1.0.0-alpha, 1.0.0-rc.1 \n For
+                  more information on semver, please see https://semver.org/"
                 maxLength: 64
                 pattern: ^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(-(0|[1-9]\d*|[0-9]*[a-zA-Z-][0-9a-zA-Z-]*)(\.(0|[1-9]\d*|[0-9]*[a-zA-Z-][0-9a-zA-Z-]*))*)?(\+([0-9a-zA-Z-]+(\.[0-9a-zA-Z-]+)*))?$
                 type: string
--- a/config/crd/bases/operators.operatorframework.io_operators.yaml
+++ b/config/crd/bases/operators.operatorframework.io_operators.yaml
@@ -40,6 +40,12 @@ spec:
                 pattern: ^[a-z0-9]+(-[a-z0-9]+)*$
                 type: string
               version:
+                description: "Version is an optional is semver constraint on the package
+                  version. If not specified, the latest version available of the package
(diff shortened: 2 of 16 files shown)


api/v1alpha1/operator_types.go
```diff
@@ -27,6 +27,11 @@ type OperatorSpec struct {
 	//+kubebuilder:validation:MaxLength:=48
 	//+kubebuilder:validation:Pattern:=^[a-z0-9]+(-[a-z0-9]+)*$
 	PackageName string `json:"packageName"`
+
+	//+kubebuilder:validation:MaxLength:=64
+	//+kubebuilder:validation:Pattern:=^(\|\||\s+)?([\s~^><=]*)v?(\d+)(\.(\d+))?(\.(\d+))?(\-(.+))?$
```
joelanford, line 32: Will this match the full set of (but not more than) version ranges supported by `semver.ParseRange`?

controllers/operator_controller_test.go
```diff
@@ -80,6 +80,38 @@ var _ = Describe("Operator Controller Test", func() {
 				Expect(cond.Message).To(Equal(fmt.Sprintf("package '%s' not found", pkgName)))
 			})
 		})
+		When("the operator specifies a version range for which there are no bundles", func() {
+			var pkgName string
+			BeforeEach(func() {
+				By("initializing cluster state")
+				pkgName = fmt.Sprintf("non-existent-%s", rand.String(6))
```
joelanford, line 87: Should this be a package name that _does_ exist, but one that doesn't have a bundle in this range?

joelanford:
> I've also nerfed the stackdumps down to panic. Otherwise we just get logs pointing to internal controller-runtime code...

Typically those logs contain the error message our reconciler returns (if we return a non-nil error). Do those errors still come through, and we're just no longer _also_ barfing out the stack trace? If so, 👍

api/v1alpha1/operator_types.go
```diff
@@ -27,6 +27,11 @@ type OperatorSpec struct {
...
+	//+kubebuilder:optional
+	Version string `json:"version,omitempty"`
```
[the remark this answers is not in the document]
tmshort: The UX document had `packageVersion`. I'm less happy with `versionRange` as it doesn't imply a single version value is acceptable.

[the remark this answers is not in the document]
bparees: >  it doesn't imply a single version value is acceptable.

and `Version` doesn't imply a range is acceptable.

how about `VersionConstraint` which is an vague enough name that people will feel forced to read the api doc to understand the acceptable values/format, but also a name which encompasses both use cases?

api/v1alpha1/operator_types.go
tmshort: AFACT, this doesn't support the full syntax.

tmshort: The semver package explicitly mentions no regex... Where did that one come from?

tmshort: https://ihateregex.io/expr/semver/ then add the comparison operators

tmshort: Either way, it's not correct as-is, and needs to account for the range prefix elements.

joelanford:
Just as an update, we discussed the tradeoffs of versionRange, version, versionConstraint and -- for now -- landed on `spec.version` and it being optional and a single version (no range). The semantics are hopefully implicitly defined by the demo script in #162.

api/v1alpha1/operator_types.go
joelanford: Well, based on the milestone 3 consensus, it sounds like we're going with `spec.version` and a single explicit version for now. So I think we can literally use the linked regex?

internal/resolution/variable_sources/olm/olm.go
```diff
@@ -14,24 +14,36 @@ import (
...
 
 	// build variable source pipeline
 	variableSource := crd_constraints.NewCRDUniquenessConstraintsVariableSource(bundles_and_dependencies.NewBundlesAndDepsVariableSource(inputVariableSources...))
 	return variableSource.GetVariables(ctx, entitySource)
 }
+
+func (o *OLMVariableSource) requiredPackageFromOperator(operator *operatorsv1alpha1.Operator) (*required_package.RequiredPackageVariableSource, error) {
+	var opts []required_package.RequiredPackageOption
+	if operator.Spec.Version != "" {
+		opts = append(opts, required_package.InVersionRange(operator.Spec.Version))
```
joelanford, line 46: @perdasilva given the direction set out in #162, we're saying that only individual versions will be allowed in the API. However, an individual version is a valid range, and I am still of the belief that we'll want to give users the ability to specify a range at some point.

So my advice is:
1. Let's make the `spec.version` regex be a regex that permits only individual versions
2. Let's keep `InVersionRange` as the under-the-hood implementation of the constraint.

api/v1alpha1/operator_types.go
perdasilva: Based on our upstream convo yesterday, I've left the name as "version" (I'm also happy with PackageVersion, if that's prefered - I personally like the succinctness of `version`. I've added some comments with examples and a link to the semver.org site.

internal/resolution/variable_sources/olm/olm.go
perdasilva: Yeah, I agree with everything you said. The original regex was for a single version (got my wires crossed - managed to get a pretty decent one for ranges, but I don't think there's a regex that matches all cases - it could be that we'll need to match *most* but still verify under the hood. ChatGPT ftw XD

controllers/operator_controller_test.go
perdasilva: Gotcha now. I've addressed it. Changed the name of the package and put a comment to say the version of the package is doesn't exist

perdasilva:
I've gone ahead and made the semver validation more robust. I also decided to add validation in the reconciler in case the regex fails. I haven't managed to find a combination that does. But, as soon as we release someone will XD. I've made the validation additive, so we can just add new validation functions.

api/v1alpha1/operator_types.go
```diff
@@ -27,6 +27,19 @@ type OperatorSpec struct {
 	//+kubebuilder:validation:MaxLength:=48
 	//+kubebuilder:validation:Pattern:=^[a-z0-9]+(-[a-z0-9]+)*$
 	PackageName string `json:"packageName"`
+
+	//+kubebuilder:validation:MaxLength:=64
+	//+kubebuilder:validation:Pattern=^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(-(0|[1-9]\d*|[0-9]*[a-zA-Z-][0-9a-zA-Z-]*)(\.(0|[1-9]\d*|[0-9]*[a-zA-Z-][0-9a-zA-Z-]*))*)?(\+([0-9a-zA-Z-]+(\.[0-9a-zA-Z-]+)*))?$
+	//+kubebuilder:Optional
+	// Version is an optional is semver constraint on the package version. If not specified, the latest version available of the package will be installed.
```
tmshort, line 34: Nit: grammar: 
`Version is an optional is semver constraint`

config/crd/bases/operators.operatorframework.io_operators.yaml
```diff
@@ -39,6 +39,16 @@ spec:
                 maxLength: 48
                 pattern: ^[a-z0-9]+(-[a-z0-9]+)*$
                 type: string
+              version:
+                description: "Version is an optional is semver constraint on the package
```
tmshort, line 43: Same grammar nit.

perdasilva: fixed ^^ nice catch. I also modified the "Example" because the rendering in the description looked funky

controllers/validators/validators.go
```diff
@@ -0,0 +1,38 @@
...
+}
+
+// ValidateOperatorSpec validates the operator spec, e.g. ensuring that .spec.version, if provided, is a valid SemVer
+func ValidateOperatorSpec(operator *operatorsv1alpha1.Operator) error {
+	validators := []operatorCRValidatorFunc{
+		validateSemver,
+	}
+	for _, validator := range validators {
+		if err := validator(operator); err != nil {
+			return err
```
m1kola, line 34: For now it is ok as there is only one field which requires validation. Just curious how you see UX for it once we have more validators - Will we bereturning one error at a time or use some sort of aggregated error (so we surface all issues at once)?

controllers/validators/validators_test.go
```diff
@@ -0,0 +1,63 @@
...
+
+import (
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"github.com/operator-framework/operator-controller/api/v1alpha1"
+	"github.com/operator-framework/operator-controller/controllers/validators"
+)
+
+var _ = Describe("Validators", func() {
```
m1kola, line 11: Maybe I'm missing something - why not to use simple unit tests here? Looks like a classic use case for a table driven test.

m1kola (commented):
Still reviewing this one. Just a few questions so far

internal/resolution/variable_sources/required_package/required_package_test.go
```diff
@@ -120,6 +122,33 @@ var _ = Describe("RequiredPackageVariableSource", func() {
 		}))
 	})
 
+	It("should filter by version range", func() {
+		// recreate source with version range option
+		var err error
+		rpvs, err = required_package.NewRequiredPackage(packageName, required_package.InVersionRange(">=1.0.0 !1.0.0 <3.0.0"))
```
tmshort, line 128: Is this valid range? i.e. include 1.0.0 then exclude it?

config/samples/operators_v1alpha1_operator.yaml
```diff
@@ -11,3 +11,4 @@ metadata:
 spec:
   # TODO(user): Add fields here
   packageName: prometheus
+  version: 3.0.0
```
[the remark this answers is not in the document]
perdasilva: sorry - that shouldn't have been committed - I'll omit it.

internal/resolution/variable_sources/required_package/required_package_test.go
perdasilva: That's a fair point. I think it's a valid range, even if it's superfluous (just collapses to >1.0.0 <3.0.0. I'll change it to !2.0.0 to avoid this confusion.

controllers/validators/validators_test.go
perdasilva: I think we decided to go for ginkgo for the unit tests. However, I'm starting to think it might have been a mistake. They aren't as easy to execute as go tests, and carry the cruft of the *_suite.go files. From my perspective at the moment, I'm not a huge fan of table driven tests. I think they do have a nice additive property to them and can make it easier to change many tests at once. But from previous experience (which is just anecdotal at best) they carry a higher complexity in trying to understand whats going later on in the project. Especially if you have massive test case structures with huge inputs. On the other hand, singular tests are easier to wrap your head around later on. However, they can carry a higher maintenance burden during refactoring/maintenance. So, I find it to be a tradeoff between readability and and maintainability (which are also interrelated XD). Now with copilot and chatGPT it's easier to write singular tests hahahah. On a serious side, I'm easy either way. I'd say we should do the following:
 - I'll create a github discussion around this so we can decide what we should do (I think it's important to have some consistency)
 - Let's leave it as it is for now, make a join decision, and we can always refactor out ginkgo (at least for the unit tests) in favor of gotests and make a decision on whether we want to do tabular or singular tests and keep that decision as a compass bearing
 
 wdyt?

controllers/validators/validators.go
perdasilva: That's a great call out. Personally, I think we should try as much as possible to add all of the validation errors as conditions. It would suck to keep going back and forth trying to get it right. What I'm not sure of right now is if it's "acceptable" to add multiple conditions of the same type (I don't see why not, but I don't know what the best practice is here). According to my good friend ChatGPT:

```
It is generally best to keep the status of a Kubernetes resource concise and clear. If you have multiple conditions of the same type, it might be better to consolidate them into a single condition that provides an accurate representation of the overall state of the resource.

Here are some guidelines to consider when working with conditions in the status of a Kubernetes resource:

Each condition should have a distinct type that describes the aspect of the resource being observed.
Conditions should be used to provide insights into the state of the resource and any issues that might be affecting it.
Use the Reason and Message fields within a condition to provide more detailed information about the state of the resource and any issues.
Update conditions as the state of the resource changes, ensuring that the status stays up-to-date and accurate.
By following these guidelines, you can maintain a clear and concise status that communicates the state of the resource effectively. In cases where you need to represent multiple aspects of the resource's state, it's better to use separate conditions with different types. However, if you have multiple instances of the same type of condition, consider consolidating them into a single condition with a clear reason and message.
```
It seems to suggest we should try to group everything into a single condition. But that seems ugly at first sight. Another option from the top of my head might be to create a condition for each validation error. That also seems ugly. Personally, I'm in favor of just having multiple conditions of the same type each describing its own reason for failure.

I suggest we leave it as it is for now. I will add a comment and a ticket so we can follow up on this after a more thorough discussion upstream. wdyt?

controllers/validators/validators.go
```diff
@@ -0,0 +1,37 @@
...
+	operatorsv1alpha1 "github.com/operator-framework/operator-controller/api/v1alpha1"
+)
+
+type operatorCRValidatorFunc func(operator *operatorsv1alpha1.Operator) error
+
+// validateSemver validates that the operator's version is a valid SemVer.
+// this validation should already be happening at the CRD level. But, it depends
+// on a regex that could possibly fail to validate a valid SemVer. This is added as an
+// extra measure to ensure a valid spec before the CR is processed for resolution
+func validateSemver(operator *operatorsv1alpha1.Operator) error {
```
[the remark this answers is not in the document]
perdasilva: > If the CRD regex validation fails, will the value even be set?
If I understood the question correctly (and please forgive me if I'm going over stuff you already know), my understanding of the way it works is: there's both client-side and server-side validation of the field value against the regex before it reaches the controller. The intent of putting that the validation is to automatically reject anything that isn't a semver so it doesn't even make it to the controller. But, after looking at the regex, I couldn't get any confidence that it would work 100% of the time. I tried to validate with different negative cases, but I couldn't convince myself that it will definitely catch everything. So, it could be possible that an invalid spec slips through. So, I thought it best to handle that case in code (even if it's unlikely that it will ever be executed). Hopefully, for most of the field that come along in the API we'll be able to get away with relatively simple regexs and lengths and we'll make use of this reconciler validation sparingly.

Since I'm rubber ducking this with you now, it occured to me that should a false positive slip through, we'd get a resolution failure on trying to parse the semver. So, maybe I'm being overzealous? Though a part of me thinks that ensuring that the input is clean before reaching the resolver might be worthwhile.

controllers/validators/validators.go
m1kola: I have definitevely seen conditions like `ValidSomething` true/false with message containing concatenated errors. I think it is very common in OCP. But I also seen condition per issue approach.

IMO - if condition is meant to be consumed by humans - we might just want to concatenate errors. We will later be able to split them into separate conditions. I personally not a fan of scrolling through 10s of conditions: I would rather see all validation errors in one place.

However if we want to programmatically consume these conditions or want to give our users this ability - then it will definitevely be easier to consume if we have conditions for each issue. But in this specific case I do not see a lot of value in this feature (but I'm happy to be proven wrong).

> I suggest we leave it as it is for now. I will add a comment and a ticket so we can follow up on this after a more thorough discussion upstream. wdyt?

Agreed

perdasilva:
demo

controllers/validators/validators.go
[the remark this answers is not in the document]
m1kola: It indeed feels like belt and braces, but I think I'm in favour of keeping this validation as it prevents us hittng rest of the code in case of discrapancies between the regex and the library (e.g. bugs in library, for example).

m1kola (approved):
From my very limited experience with v1 - looks good.

tmshort:
> The demo image...
Although it might have been easier to use a script with `set -x`

controllers/validators/validators.go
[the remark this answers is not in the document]
tmshort: Assuming you got the regex from a reliable source, this scenario should never be hit, but better safe than sorry.

tmshort (approved):

perdasilva:
> > The demo image...
> 
> Although it might have been easier to use a script with `set -x`

Yeh, I did that originally - but it didn't look as nice =(

controllers/validators/validators_test.go
joelanford: Ginkgo supports table driven test fwiw. Maybe not quite as straightforward as a hand-written one, but maybe a reasonable compromise?

controllers/validators/validators.go
joelanford: The condition type needs to be unique in the list of conditions.

Type and reason are part of the API so once we GA, those can't be removed or changed.

My recommendation is generally to keep the number of types and reasons as small as possible based on what clients specifically require.

In this case, I'd start with whatever type we're already using, maybe even use the existing failure reason, and then put the validation errors concatenated in the message.

But +1 to capturing this in a separate issue.

joelanford:
I thing I didn't see in the demo (or maybe missed?) is what happens when you have an existing version already installed and set `spec.version` to a non-existent version (Items 10 and 11 from #162)?

Another thing we could show (and yes I realize this isn't technically in the scope of #162): what happens when trying to create or update the `Operator` with an `spec.version` that is not parseable semver?

PR #147, opened 2023-03-30
bump goreleaser, remove obsolete substitute task

joelanford:
Go releaser 1.16.1 made it possible to use templates in the `release.disable` and `changelog.skip` fields. As a result, we can simplify our release process by directly templating our release pipeline environment variable into the goreleaser config file.

I also changed the polarity of the `DISABLE_RELEASE_PIPELINE ?= true` environment variable to be `ENABLE_RELEASE_PIPELINE ?= false`. This seemed more intuitive (to me at least).


awgreene (approved):
Looks good to me!

/approve

PR #150, opened 2023-04-03
Introduce the Apache License

awgreene:


LICENSE
```diff
@@ -0,0 +1,201 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [yyyy] [name of copyright owner]
```
everettraven, line 189: Do we need to specify the copyright information?
```suggestion
   Copyright 2023 Operator Framework
```

everettraven: Nevermind, I realize now that is part of the section you apply to the top of each file :)

Is there going to be a separate PR to add the copyright information to all the files?

joelanford: +1 we should probably put the copyright info in source code headers.

And we should also probably have a sanity test that checks for the license headers. That could probably be a follow-up though.

tmshort (approved):
This appears to be identical to the  Apache license provided by GitHub when you add a license as your initial commit to a repository.

I also compared it to the apache online version at https://www.apache.org/licenses/LICENSE-2.0.txt, and with the exception of a single blank line at the beginning, they are identical.

I will note that OpenSSL is Apache 2.0 licensed, but their license differs:
1. The URL at the top is `http` vs `https`
2. They don't have the APPENDIX

PR #157, opened 2023-04-04
Release Install script

dtfranz:
Adds an install script which installs the required sub-components of OLM v1 in a single command. The install script is built on-the-fly when the release is run, and the appropriate versions of all sub-components are set inside the Makefile or grabbed from the go.mod file.

Addresses #110

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,7 @@ export GO_BUILD_TAGS ?= upstream
 export CERT_MGR_VERSION ?= v1.9.0
 export GORELEASER_VERSION ?= v1.16.2
 export OLM_V0_VERSION ?= v0.24.0
+export RUKPAK_VERSION=$(shell go list -mod=mod -m -f "{{.Version}}" github.com/operator-framework/rukpak)
 export WAIT_TIMEOUT ?= 60s
 IMG?=$(IMAGE_REPO):$(IMAGE_TAG)
 
@@ -100,7 +101,7 @@ build: manifests generate fmt vet goreleaser ## Build manager binary using gorel
 	${GORELEASER} build ${GORELEASER_ARGS} --single-target -o bin/manager
 
 .PHONY: run
-run: docker-build kind-cluster kind-load cert-mgr rukpak install deploy wait ## Build the operator-controller then deploy it into a new kind cluster.
+run: docker-build kind-cluster kind-load install ## Build the operator-controller then deploy it into a new kind cluster.
 
 .PHONY: wait
 wait:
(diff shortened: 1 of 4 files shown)


scripts/install.tpl.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
+
+olm_version=$OLM_V0_VERSION
```
dtfranz, line 3: These are "supposed" to be set from the envsubst command in the makefile's `quickstart` target but I added some simple checks just in case.

scripts/install.tpl.sh
```diff
@@ -0,0 +1,45 @@
...
+
+kubectl apply -f "https://github.com/cert-manager/cert-manager/releases/download/${cert_mgr_version}/cert-manager.yaml"
+kubectl_wait "cert-manager" "deployment/cert-manager-webhook" "60s"
+
+kubectl apply -f "https://github.com/operator-framework/rukpak/releases/download/${rukpak_version}/rukpak.yaml"
+kubectl_wait "rukpak-system" "deployment/core" "60s"
+kubectl_wait "rukpak-system" "deployment/helm-provisioner" "60s"
+kubectl_wait "rukpak-system" "deployment/rukpak-webhooks" "60s"
+
+kubectl apply -f "https://github.com/operator-framework/operator-controller/releases/download/${operator_controller_version}/operator-controller.yaml"
```
joelanford, line 44: One thing that could be useful here is if we used this install script as part of the `e2e` and `run` targets. That way we actually exercise this as part of our dev cycle and testing.

I think all we would need to do is make the entire operator-controller manifest path configurable.

For `make run`, we'd generate the manifest without the image ref substitution, and then this would be `kubectl apply -f ./operator-controller.yaml`

For the release manifest, we'd substitute the entire release URL in like you're doing here.

If this is not a trivial change, its totally non-blocking and would be fine to do in a follow-up.

scripts/install.tpl.sh
```diff
@@ -0,0 +1,45 @@
...
+if [[ -z "$olm_version" || -z "$cert_mgr_version" || -z "$rukpak_version" || -z "$operator_controller_version" ]]; then
+    err="Error: Missing component version(s) for: "
+    if [[ -z "$olm_version" ]]; then
+        err+="operator-lifecycle-manager "
+    fi 
+    if [[ -z "$cert_mgr_version" ]]; then
+        err+="cert-manager "
+    fi 
+    if [[ -z "$rukpak_version" ]]; then
+        err+="rukpak "
```
awgreene, line 17: Nit, extra spaces.
```suggestion
        err+="operator-lifecycle-manager"
    fi 
    if [[ -z "$cert_mgr_version" ]]; then
        err+="cert-manager"
    fi 
    if [[ -z "$rukpak_version" ]]; then
        err+="rukpak"
```

Makefile
```diff
@@ -8,6 +8,7 @@ export GO_BUILD_TAGS ?= upstream
 export CERT_MGR_VERSION ?= v1.9.0
 export GORELEASER_VERSION ?= v1.16.2
 export OLM_V0_VERSION ?= v0.24.0
+export RUKPAK_VERSION=$(shell go list -mod=mod -m -f "{{.Version}}" github.com/operator-framework/rukpak)
```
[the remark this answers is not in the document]
dtfranz: Credit to @ankitathomas for this one! :smile:

scripts/install.tpl.sh
dtfranz: They were intentional, it's a lazy way to format the final error string when component versions are unset.

scripts/install.tpl.sh
dtfranz: Cool! I'll give this a shot, it's a good idea.

dtfranz: Added :)

awgreene (approved):
Looks good to me!

PR #166, opened 2023-04-12
Convert OLM V1 MVP share doc to markdown

tlwu2013:
This commit add the OLM v1 doc in markdown which is converted from the shared doc: [OLM V1 Product Requirements Document](https://docs.google.com/document/d/1-vsZ2dAODNfoHb7Nf0fbYeKDF7DUqEzS9HqgeMCvbDs/edit)

### DIFF
--- a/docs/olmv1_roadmap.md
+++ b/docs/olmv1_roadmap.md
@@ -1,0 +1,134 @@
+# OLM v1 roadmap
+
+## OLM overview and use cases
+
+OLMs purpose is to manage cluster extensions centrally and declaratively on Kubernetes clusters. Its mission is to make installing, running, and updating functional add-ons to the cluster easy, safe, and reproducible for cluster administrators and PaaS administrators throughout the lifecycle of the underlying cluster as well as those of add-ons. The scope of OLM is currently the one cluster it is running on.
+
+OLM has unique support for the specific needs of cluster extensions, which are commonly referred to as operators. These are classified as one or more Kubernetes controllers shipping with one or more API extensions (CustomResourceDefinitions) to provide additional functionality to the cluster (though there are deviations from this coupling of CRDs and controllers, discussed below). They are managed centrally by OLM running on the cluster, where OLMs functionality is implemented following the Kubernetes operator pattern as well.
+
+OLM defines a lifecycle for these extensions in which they get installed, potentially causing other extensions to be installed as well, a limited set of customization of configuration at runtime, an update model following a path defined by the extension developer, and eventual decommission and removal.
+
+There is a dependency model in which extensions can rely on each other for required services that are out of scope of the primary purpose of the extension, allowing each extension to focus on a specific purpose. OLM also prevents conflicting extensions from running on the cluster, either with conflicting dependency constraints or conflicts in ownership of services provided via APIs.
+
+Since cluster extensions need to be supported with an enterprise-grade product lifecycle the specific scenarios in which an extension can be installed or updated are going to be limited by the author of the extension, primarily to align with what was tested by their QE processes. OLM allows the author to enforce these support  limitations in the form of additional constraints.
+
+During their lifecycle on the cluster, OLM also manages the permissions and capabilities extensions have on the cluster as well as the permission and access tenants on the cluster have to the extensions. This is done using the Kubernetes RBAC system in combination with tenant isolation using Kubernetes namespaces. The interaction surface of the extensions is solely composed of Kubernetes APIs the extensions define.
+
+OLM also defines a packaging model in which catalogs of extensions, usually containing the entire version history of each extension, are made available to clusters for cluster users to browse and select from. Via new versions of extensions delivered with this packaging system, OLM is able to apply updates to existing running extensions on the cluster in a way where the integrity of the cluster is maintained and constraints and dependencies are kept satisfied.
+
+The scope of all these features is a single cluster so far with namespace-based handling of catalog access and extension API accessibility and discoverability. Expansion of this scope is indirectly expected through the work of the Kubernetes Control Plane (kcp) project, which in its first incarnation will likely use its own synchronization mechanism to get OLM-managed extensions deployed eventually on one or more physical clusters from a shared, virtual control plane called a “workspace”. This is an area under active development and subject to change, OLM might need to become aware of kcp in a future state.
+
+**Verdict:** The purpose of OLM remains the same. The scope of OLM will increase to span multiple clusters following the kcp model, though likely many aspects of this will become transparent to OLM itself through the workspace abstraction that kcp provides. What needs to change in OLM 1.0 is how the above tasks are carried out from the user perspective and how much control users have in the process and which persona is involved.
+
+## Functional Requirements
+_Priority Rating: 1 highest, 2 medium, 3 lower (e.g. P2 = Medium Priority)_
+
+**F1 - Extension catalogs (P1):** The existing OLM concepts around catalogs, packages and channels is to be used as a basis for below functional requirements.
+
+**F2 - Extension catalog discovery (P2):** Unprivileged tenants need to have the ability to discover extensions that are available to install. In particular users need to be able to discover all versions in all channels that an extension package defines in a catalog. The privilege should be given at the discretion of the cluster administrator.
+
+**F3 - Dependency preview (P3):** Before extension installation, OLM needs to introspect the dependencies of an extension and present a preview of the resolution result to the user to let the user confirm they are ok with this.
+
+**F4 - Dependency review (P3):** For installed extensions OLM needs to surface the dependency relationship to other installed extensions and also highlight which other extensions depend on a particular extension so users are informed about the relationships at runtime.
+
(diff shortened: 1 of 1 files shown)


docs/olmv1_mvp.md
```diff
@@ -0,0 +1,136 @@
...
+- OLM will move to a descoped, cluster-wide singleton model for cluster extensions, extension management isn’t namespaced
+
+## Constraints
+- Only operator bundles with “AllNamespace” mode installation support can be lifecycled with the new APIs / flows in OLM 1.0
+
+## Dependencies
+- RBAC templating (OLM-2158) and discrete visibility control (OLM-2366) will land prior to OLM 1.0 and unblock most operators that do not support AllNamespace installation mode today
+
+- Bandwidth in the Console engineering team to surface the new flows / forms / visualizations associated with any functional requirements
+
```
everettraven, line 120: Are these relevant to include in the doc? These seem to have some references to Jira tickets and internal Red Hat teams. Also I'm not sure the RBAC templating is still a dependency (could be wrong here).

docs/olmv1_mvp.md
```diff
@@ -0,0 +1,136 @@
...
+**Verdict:** The purpose of OLM remains the same. The scope of OLM will increase to span multiple clusters following the kcp model, though likely many aspects of this will become transparent to OLM itself through the workspace abstraction that kcp provides. What needs to change in OLM 1.0 is how the above tasks are carried out from the user perspective and how much control users have in the process and which persona is involved.
+
+## Functional Requirements
+_Priority Rating: 1 highest, 2 medium, 3 lower (e.g. P2 = Medium Priority)_
+
+**F1 - Extension catalogs (P1):** The existing OLM concepts around catalogs, packages and channels is to be used as a basis for below functional requirements.
```
everettraven, line 26: From a contributor perspective, I think it would be neat to set these up as check boxes so that when we have met these requirements we can check them off. Would be helpful to easily tell which things have been achieved and which have not.

everettraven (commented):
I know this is a direct port from the Google doc, but I have a couple non-blocking comments:

docs/olmv1_mvp.md
```diff
@@ -0,0 +1,136 @@
...
+
+OLM has unique support for the specific needs of cluster extensions, which are commonly referred to as operators. These are classified as one or more Kubernetes controllers shipping with one or more API extensions (CustomResourceDefinitions) to provide additional functionality to the cluster (though there are deviations from this coupling of CRDs and controllers, discussed below). They are managed centrally by OLM running on the cluster, where OLMs functionality is implemented following the Kubernetes operator pattern as well.
+
+OLM defines a lifecycle for these extensions in which they get installed, potentially causing other extensions to be installed as well, a limited set of customization of configuration at runtime, an update model following a path defined by the extension developer, and eventual decommission and removal.
```
varshaprasad96, line 9: Would it be helpful to call the extensions in this particular context as "operators" directly for easier understanding?

docs/olmv1_mvp.md
```diff
@@ -0,0 +1,136 @@
...
+
+**B6 - Human-readable status extensions information (P2):** Whenever OLM is in the process of or having reached or failed to reach a desired state it needs to update the user about what is happening / what has happened without assuming knowledge about OLM internal or implementation details.
+
+**B7 - Scalability & Resource consumption (P1):** OLM is used on clusters with hundreds to thousands of namespaces and tenants. Its API controls, specifically for F2 and F12 need to be built in such a way that resource consumption scales linearly with usage and cluster size and the overall resource usage envelope stays within manageable bounds that does not put the cluster stability, especially that of the API server at risk. System memory especially is a scarce resource.
+
```
varshaprasad96, line 97: Do we also need to talk about safer installs and rollback in case there is install is unsuccessful.

docs/olmv1_mvp.md
```diff
@@ -0,0 +1,136 @@
...
+**C1 - Compatibility with existing extensions  (P1):** OLM should be able to manage extensions packaged with the current bundle format in the way described by the functional and behavior requirements when the bundle supports AllNamespace installation mode.
+
+**C2 - Compatibility with existing catalogs (P1):** OLM should be able to retrieve and update extensions that adhere to C1 from the currently supported catalog formats (File-based catalogs).
+
+**C3 - Incompatibility with existing extensions  (P1):** OLM 1.0 does not support managing bundles or extension versions that do not support AllNamespace installation mode with the new set of APIs or flows
+
+## Assumptions
+- No additional tenancy model will be introduced at the control plane / API layer of Kubernetes upstream
+
+- kcp doesn’t fundamentally change OLMs role and responsibilities around managing extensions (at least initially)
```
varshaprasad96, line 109:
```suggestion
- No additional tenancy model will be introduced at the control plane / API layer of Kubernetes upstream

- kcp doesn’t fundamentally change OLMs role and responsibilities around managing extensions (at least initially)
```

Do we need these assumptions explicitly mentioned for an user?

docs/olmv1_mvp.md
```diff
@@ -0,0 +1,136 @@
+# OLM v1 MVP
```
tlwu2013, line 1: The functional/behavioral requirements included in this doc are actually beyond the scope of the MVP.  What do you all think if we title this as "OLM v1 Project Requirement Document (PRD)"?





Tony Wu
  [5 minutes ago](https://kubernetes.slack.com/archives/C0181L6JYQ2/p1681442993466969?thread_ts=1681333000.404779&cid=C0181L6JYQ2)
if we want to swap out Product Requirement Document (PRD) to something less business focused, should we call it something like “project requirement document”? (edited) 





Tony Wu
  [4 minutes ago](https://kubernetes.slack.com/archives/C0181L6JYQ2/p1681443027065769?thread_ts=1681333000.404779&cid=C0181L6JYQ2)
this means its coverage is beyond the MVP scope (edited) 





Tony Wu
  [4 minutes ago](https://kubernetes.slack.com/archives/C0181L6JYQ2/p1681443044380709?thread_ts=1681333000.404779&cid=C0181L6JYQ2)
and the bonus point is the shorthand is still “PRD” (edited) 





Tony Wu
  [4 minutes ago](https://kubernetes.slack.com/archives/C0181L6JYQ2/p1681443051958609?thread_ts=1681333000.404779&cid=C0181L6JYQ2)
what do you all think?


Tony Wu
  [1 minute ago](https://kubernetes.slack.com/archives/C0181L6JYQ2/p1681443223709359?thread_ts=1681333000.404779&cid=C0181L6JYQ2)
let me add that to the PR too

everettraven: Maybe "OLMv1 Roadmap"? IMO it would be nice to break this down into different sections that include what is needed for the MVP and then what is planned for the future. Maybe a structure like:

```md
# OLM v1 Roadmap

## MVP

### Functional Requirements
### Behavioral Requirements

## Future Work

### Functional Requirements
### Behavioral Requirements
```

That way we can have a clear line between what is necessary for the MVP and what is not.

docs/olmv1_mvp.md
tlwu2013: Right, this PR was intended to be a direct porting; hence, I didn't look into the details, so thanks for catching this.

The first bullet here is meant to call out: for OLM 1.0 to succeed, it is crucial to address "F13 - Extension permissions management (P1)" and "F12 - Installed extension discovery (P1)" first so cluster admins can grant Operators the minimum necessary permissions to fulfill their intended function rather than always granting permissions at the cluster level.  I'll remove those old Jira ticket IDs and reword them, and I'll remove the console-related work.

docs/olmv1_mvp.md
tlwu2013: I like the idea of having a visual checklist to track progress, but at the same time, I worry the rollout of certain items may occur in multiple phases, which could make it difficult to promptly check items off.  As such, should we leave the bullet list as is for the time being and submit a pull request for an update once there are features that can be checked off?

docs/olmv1_mvp.md
tlwu2013: I've changed the title and the filename to "roadmap" in this PR, and should we defer defining the MVP and Future Work scope in the future PR?

tlwu2013:
hey folks, thanks for looking at this PR, and I've addressed some of the feedback quickly.  Since the intention for this PR is a direct porting from shared doc to markdown so we can have other feature requirements being discussed and captured in the upstream requirement markdown, should we merge this one first and address some feedback in later PRs that needs extended discussions?  Thank you all!

docs/olmv1_mvp.md
everettraven: IMO an item should not be checked off until it is fully implemented. I'm fine with checking off after the fact if we are okay with that

joelanford: I would at least make the individual items markdown headers so that we get link anchors out of them. But that can easily happen in a follow-up.

tlwu2013:
> This looks good to me. I agree that we should go ahead and merge then and then make further suggestions in follow-on PRs. Otherwise, I fear there will always be _something_ debatable here that keeps this PR from merging. Let's have those individual debates in smaller PRs.
> 
> @tlwu2013 Could you amend your commit to get the DCO to pass? Instructions here: https://probot.github.io/apps/dco/

Done, thanks @joelanford !

PR #179, opened 2023-04-20
(gitignore) gitignore only stuff included in project

anik120:


.gitignore
```diff
@@ -1,35 +1,16 @@
...
 
 # Kubernetes Generated files - skip generated files, except for vendored files
 
 !vendor/**/zz_generated.*
 
-# editor and IDE paraphernalia
-.idea
-*.swp
-*.swo
-*~
```
tmshort, line 32: Removing this can be an issue with EMACS users (such as myself).

tmshort (commented):
Why are these files being removed from gitignore? There is no reason to check any of these in, so it's not unreasonable to have them in the file.

.gitignore
```diff
@@ -1,35 +1,16 @@
...
-dist/**
 operator-controller.yaml
-install.sh
 
 # Kubernetes Generated files - skip generated files, except for vendored files
 
 !vendor/**/zz_generated.*
 
-# editor and IDE paraphernalia
-.idea
```
joelanford, line 29: I use GoLand, which creates this directory in the repo root. Please leave this in.

.gitignore
```diff
@@ -1,35 +1,16 @@
...
-*.swp
-*.swo
-*~
+# Dependency directories
+vendor/
 
-# TODO dfranz remove this line and the bin folder when tools binaries are moved to their own folder
-!bin/.dockerignore
```
joelanford, line 35: Are the tool binaries in separate directories now? If not, this (and the `.dockerignore` file) should stay.

.gitignore
```diff
@@ -1,35 +1,16 @@
...
 *.out
 
 # Release output
-dist/**
 operator-controller.yaml
-install.sh
 
 # Kubernetes Generated files - skip generated files, except for vendored files
 
 !vendor/**/zz_generated.*
```
joelanford, line 12: We're not vendoring anything, so I think we want to remove this as well?

.gitignore
```diff
@@ -1,35 +1,16 @@
...
+bin/
 
-# Output of the go coverage tool, specifically when used with LiteIDE
+# Output of the go coverage tool
 *.out
 
 # Release output
-dist/**
 operator-controller.yaml
-install.sh
```
joelanford, line 22: Don't get generate `install.sh` as part of `make run|e2e|release`? And goreleaser creates the `dist` directory when we build binaries and run the release. So I think these lines need to stay.

PR #181, opened 2023-04-21
Tools Bin

dtfranz:
Separates the tools binaries out to their own folder so we can stop doing funky stuff to ignore them when building the operator-controller image.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ vendor
 *.dylib
 bin/*
 testbin/*
+/hack/tools/bin/*
 Dockerfile.cross
 
 # Test binary, build with `go test -c`
@@ -33,5 +34,3 @@ install.sh
 \#*\#
 .\#*
 
-# TODO dfranz remove this line and the bin folder when tools binaries are moved to their own folder
-!bin/.dockerignore
--- a/Makefile
+++ b/Makefile
@@ -98,6 +98,7 @@ kind-cluster-cleanup: kind ## Delete the kind cluster
 
 .PHONY: build
 build: manifests generate fmt vet goreleaser ## Build manager binary using goreleaser for current GOOS and GOARCH.
+	mkdir -p bin
 	${GORELEASER} build ${GORELEASER_ARGS} --single-target -o bin/manager
 
 .PHONY: run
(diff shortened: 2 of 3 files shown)


dtfranz:
/hold
I'd like to try caching the tools directory as well for a little extra speed boost on the builds.

Makefile
```diff
@@ -174,10 +174,12 @@ deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in
...
 	$(KUSTOMIZE) build config/default | kubectl delete --ignore-not-found=$(ignore-not-found) -f -
 
-##@ Build Dependencies
+################
+# Hack / Tools #
+################
 
 ## Location to install dependencies to
-LOCALBIN ?= $(shell pwd)/bin
+LOCALBIN ?= $(shell pwd)/hack/tools/bin
```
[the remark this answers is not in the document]
dtfranz: It's setup this way to give us an option in the future of declaring our go tools via `go.mod` file in `hack/tools/`, then when those are pulled down they'll go in `hack/tools/bin/`. Makes it a bit easier to manage with our `.gitignore` files IMO.

.goreleaser.yml
```diff
@@ -9,6 +9,9 @@ before:
 builds:
   - id: operator-controller
     main: ./
+    hooks:
+      pre:
+        - mkdir -p bin
```
joelanford, line 14: Is `bin` only a concern for `goreleaser` because of the `make build` task? If so, WDYT about calling `mkdir -p bin` in the `make build` recipe prior to calling `goreleaser`?

.gitignore
```diff
@@ -8,6 +8,7 @@ vendor
 *.dylib
 bin/*
 testbin/*
+hack/tools/bin/*
```
joelanford, line 11: IMO, we should be more explicit in some of these cases. IIRC, leaving the leading slash out means that we will ignore any path under the current dir that ends with `hack/tools/bin/*`. Whereas including the leading slash means _just_ this specific directory from the root (based on the location of the gitignore file). IMO explicitness is almost always better.

There's a reasonable argument one could make about doing this more ubiquitously for all relevant .gitignore paths in a single PR, so this is non-blocking, regardless.

```suggestion
/hack/tools/bin/*
```

.goreleaser.yml
dtfranz: It's needed for everything that runs `goreleaser`, including the `release` target, so I figured I'd build it into the yaml so it would always be there regardless of how/which `make` was called.

.gitignore
dtfranz: Fully in agreement here, I'll add your suggestion. Thanks!

.goreleaser.yml
joelanford: > including the release target

Ah, I didn't know this. How does it come into play with `make release`? (I thought that target used `<projectRoot>/dist` exclusively.

.goreleaser.yml
```diff
@@ -26,33 +29,21 @@ dockers:
   dockerfile: Dockerfile
   goos: linux
   goarch: amd64
-  use: buildx
-  build_flag_templates:
-    - "--platform=linux/amd64"
```
joelanford, line 31: Need to rebase to pick this fix back up?

.goreleaser.yml
dtfranz: I stand corrected, you were right on this one; I've updated it to just call `mkdir -p bin` when running the `build` target.

joelanford (approved):

Issue #161, opened 2023-04-11
Use a catalogd-defined catalog to install an Operator


joelanford:
This issue should be closed when a demo recording is posted to the #olm-dev kubernetes slack channel that follows the below script:

Demo Script:
1. Install operator-controller from the `main` branch or its most recently tagged release using operator-controller's standard install script and manifest.
2. Create a `Catalog` object referencing `quay.io/operatorhubio/catalog:latest`
3. Show that the catalog status reflects that it was successfully unpacked.
4. Show that the catalogd metadata api reports objects for all packages, channels, and bundles in the catalog.
5. Create an `Operator` object referencing package `strimzi-kafka-operator`.
6. Show that a`BundleDeployment` is created to reflect a bundle image reference with the highest semver bundle defined in the package (as of 4/5/23, that version is 0.34.0):  `quay.io/operatorhubio/strimzi-kafka-operator@sha256:42a3dce711492fae22f2d919299a0594f152396f6c2fef65ffc47226ba9c786e`
7. Delete the `Operator` and `Catalog` objects.
8. Show that no catalog metadata or bundle deployment objects remain.

Related Issues:
- [ ] https://github.com/operator-framework/operator-controller/pull/141

joelanford:
@anik120 Demoed this integration in yesterday's OLM working group meeting. Watch it here: https://youtu.be/RR88HeoCyfE?t=2322

Closing as complete.

PR #182, opened 2023-04-24
(entitySource) Use catalogd provided content for sourcing entities

anik120:
This PR replaces OLM v0 catalog operator with https://github.com/operator-framework/catalogd as the source of entities for deppy.
 
Addresses #161


controllers/operator_controller.go
```diff
@@ -121,7 +116,15 @@ func (r *OperatorReconciler) reconcile(ctx context.Context, op *operatorsv1alpha
 		})
 		return ctrl.Result{}, nil
 	}
-
+	if r.Resolver == nil {
```
anik120, line 119: Keeping `Resolver` as a dependency for Operator reconciler, and passing in a nil value in main.go, since the tests pass in a fakeResolver with a `testEntitySource`. Ideally I'd like to use mocking for interfaces instead of passing in a nil value, but then kept it like this since this is the first pass implementation that just makes a new entitySource every time the reconciler runs, I.e we'll probably optimize the entire structure soon and it's probably worth waiting and seeing which way this molds before restricting the test suite to fit any mold.

internal/resolution/entitysources/catalogdsource.go
```diff
@@ -0,0 +1,142 @@
...
+				}
+			}
+		}
+		channelValue, _ := jsonMarshal(property.Channel{ChannelName: channelName, Priority: 0})
+		props[property.TypeChannel] = string(channelValue)
+		entity := input.Entity{
+			ID:         deppy.IdentifierFromString(fmt.Sprintf("%s%s%s", bundle.Name, bundle.Spec.Package, channelName)),
+			Properties: props,
+		}
+		entities[entity.ID] = entity
```
anik120, line 71: This isn't the greatest, but since we have work planned for in the catalogd side that'll change these APIs, I didn't optimize this at all, and only included the bare minimum that deppy reconciles with today (I.e did not include any properties other than package name, version, bundle image and channel it belongs to)

scripts/install.tpl.sh
```diff
@@ -11,7 +11,7 @@ olm_version=$OLM_V0_VERSION
 cert_mgr_version=$CERT_MGR_VERSION
 rukpak_version=$RUKPAK_VERSION
 
-if [[ -z "$olm_version" || -z "$cert_mgr_version" || -z "$rukpak_version" ]]; then
+if [[ -z "$cert_mgr_version" || -z "$rukpak_version" ]]; then
     err="Error: Missing component version(s) for: "
     if [[ -z "$olm_version" ]]; then
```
dtfranz, line 16: Lines 16 through 18 can also be removed since we aren't checking the `olm_version` anymore :)

controllers/operator_controller.go
stevekuznetsov: Playing devil's advocate: if your test suite is so dependent on the particular implementation details, it's too fragile and perhaps is testing at the wrong level of abstraction. A test that sets up inputs, runs a reconcile loop, and asserts over outputs should be robust to choices like how optimized the entity source is.

internal/resolution/resolver.go
```diff
@@ -7,9 +7,8 @@ import (
 	"github.com/operator-framework/deppy/pkg/deppy/solver"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
-	"github.com/operator-framework/operator-controller/internal/resolution/variable_sources/olm"
-
 	"github.com/operator-framework/operator-controller/api/v1alpha1"
+	"github.com/operator-framework/operator-controller/internal/resolution/variable_sources/olm"
```
stevekuznetsov, line 11: I strongly recommend using https://github.com/openshift-eng/openshift-goimports in CI

internal/resolution/entitysources/catalogdsource.go
```diff
@@ -0,0 +1,142 @@
...
+				}
+				props[property.TypePackage] = string(pkgValue)
+			}
+		}
+
+		imgValue, err := jsonMarshal(bundle.Spec.Image)
+		if err != nil {
+			return nil, err
+		}
+		props["olm.bundle.path"] = string(imgValue)
```
stevekuznetsov, line 55: Should this refer to the package constant?

internal/resolution/variable_sources/entity/bundle_entity.go
```diff
@@ -120,17 +118,10 @@ func (b *BundleEntity) ChannelProperties() (*ChannelProperties, error) {
...
-	return "", fmt.Errorf("error determining bundle path for entity '%s': property '%s' not found or empty", b.ID, propertyBundlePath)
-}
-
-func (b *BundleEntity) DefaultChannel() (string, error) {
-	if defaultChannel, ok := b.Entity.Properties[propertyDefaultChannel]; ok && defaultChannel != "" {
-		return defaultChannel, nil
+	if err := b.loadBundlePath(); err != nil {
+		return "", err
 	}
-	return "", fmt.Errorf("error determining default channel for entity '%s': property '%s' not found or empty", b.ID, propertyDefaultChannel)
```
stevekuznetsov, line 133: If this was a useful error before, why did we lose it?

internal/resolution/variable_sources/entity/bundle_entity.go
```diff
@@ -212,6 +203,19 @@ func (b *BundleEntity) loadChannelProperties() error {
...
+		if err != nil {
+			return fmt.Errorf("error determining bundle path for entity '%s': %w", b.ID, err)
+		}
+		b.bundlePath = bundlePath
+	}
+	return nil
+}
+
 func loadFromEntity[T interface{}](entity *input.Entity, propertyName string, required propertyRequirement) (T, error) {
 	deserializedProperty := *new(T)
```
stevekuznetsov, line 220: Especially if consuming the property is deserializing it, it does not seem like serializing it in the first place and storing the serialized form is getting us anything?

internal/resolution/variable_sources/entity/bundle_entity_test.go
```diff
@@ -243,7 +243,7 @@ var _ = Describe("BundleEntity", func() {
 	Describe("BundlePath", func() {
 		It("should return the bundle channel properties if present", func() {
 			entity := input.NewEntity("operatorhub/prometheus/0.14.0", map[string]string{
-				"olm.bundle.path": "bundle.io/path/to/bundle",
+				"olm.bundle.path": `"bundle.io/path/to/bundle"`,
```
stevekuznetsov, line 246: Is this a change in semantic?

main.go
```diff
@@ -43,12 +42,12 @@ var (
...
 	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
+
 	utilruntime.Must(operatorsv1alpha1.AddToScheme(scheme))
 	utilruntime.Must(rukpakv1alpha1.AddToScheme(scheme))
-	//+kubebuilder:scaffold:scheme
+	utilruntime.Must(catalogd.AddToScheme(scheme))
 
-	// required by the catalog source cache controller
-	utilruntime.Must(olmv0v1alpha1.AddToScheme(scheme))
+	//+kubebuilder:scaffold:scheme
```
stevekuznetsov, line 50: dangling comment?

everettraven (commented):
First round - most of my comments are around trying to understand the decision to go with a static entity source implementation rather than a dynamic one.

test/e2e/install_test.go
```diff
@@ -63,6 +44,8 @@ var _ = Describe("Operator Install", func() {
 		err := c.Create(ctx, operator)
 		Expect(err).ToNot(HaveOccurred())
 
+		// TODO dfranz: This test currently relies on the hard-coded CatalogSources found in bundle_cache.go
```
dtfranz, line 47: Shouldn't this comment be removed instead of added back since the hard-coded catalog sources are gone?

controllers/operator_controller.go
anik120: > Playing devil's advocate: if your test suite is so dependent on the particular implementation details, it's too fragile and perhaps is testing at the wrong level of abstraction. A test that sets up inputs, runs a reconcile loop, and asserts over outputs should be robust to choices like how optimized the entity source is.

Generally yes I agree. Looks like when the entitySources was either [static](https://github.com/operator-framework/operator-controller/blob/e596002d68d6cbf8f1408ca06f04909ed8e0161d/main.go#L94), or had the [appearance of being static](https://github.com/operator-framework/operator-controller/blob/main/main.go#L108) since there was a controller updating the content of the entitySource, it made sense to structure the unit tests in a way that takes advantage of the [assumption that entitySource is going to be static](https://github.com/operator-framework/operator-controller/blob/a252719a953c59063947977f88d8bdbcde96887f/controllers/operator_controller_test.go#L36) from the reconciler's point of view. With [our wish to do away with that controller structure for creating entities](https://github.com/operator-framework/operator-controller/pull/141#issuecomment-1516407385) to get us started, it's also highlighting what @stevekuznetsov mentioned here. Hence the call out from me about passing in `nil` for that field in the main caller, because we'll probably have to rethinking the layout of the test suite (which I'd argue is out of scope for this PR but def should be addressed in a follow up PR)

internal/resolution/entitysources/catalogdsource.go
```diff
@@ -0,0 +1,142 @@
...
+
+		for _, prop := range bundle.Spec.Properties {
+			switch prop.Type {
+			case property.TypePackage:
+				type pkg struct {
+					Name    string `json:"packageName"`
+					Version string `json:"version"`
+				}
+				tmp := pkg{}
+				if err := json.Unmarshal(prop.Value, &tmp); err != nil {
```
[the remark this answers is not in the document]
tmshort: Mistakes happen? And we should always validate user-derived input?

controllers/operator_controller_test.go
```diff
@@ -24,7 +24,7 @@ import (
 	operatorutil "github.com/operator-framework/operator-controller/internal/util"
 )
 
-var _ = Describe("Operator Controller Test", func() {
+var _ = Describe("Reconcile Test", func() {
```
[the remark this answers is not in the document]
anik120: When I was removing the catalogsource controller I wanted to enforce in my head that there's only one, main reconciler now, but it doesn't have to reflect in the code base :) reverting this...

internal/resolution/entitysources/catalogdsource.go
```diff
@@ -0,0 +1,142 @@
...
+)
+
+// catalogdEntitySource is a source for(/collection of) deppy defined input.Entity, built from content
+// made accessible on-cluster by https://github.com/operator-framework/catalogd.
+// It is an implementation of deppy defined input.EntitySource
+type catalogdEntitySource map[deppy.Identifier]input.Entity
+
+func NewCatalogdEntitySource(ctx context.Context, client client.Client) (catalogdEntitySource, error) {
+	entities := catalogdEntitySource{}
+	bundleMetadatas, packageMetdatas, err := onClusterMetadata(ctx, client)
```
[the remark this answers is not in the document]
anik120: > `onCluster` as opposed to what? Why is that meaningful?

As opposed to [static hardcoded entitySource](https://github.com/operator-framework/operator-controller/blob/a252719a953c59063947977f88d8bdbcde96887f/main.go#L94). I'm open to a better name suggestion though, but like I mentioned, this code is definitely subject to change soon

internal/resolution/entitysources/catalogdsource.go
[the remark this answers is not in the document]
anik120: So right now we just throw in whatever we're getting from the [declCfg](https://github.com/operator-framework/catalogd/blob/main/pkg/controllers/core/catalogsource_controller.go#L230) [into the CR](https://github.com/operator-framework/catalogd/blob/main/pkg/controllers/core/catalogsource_controller.go#L257-L260), which ends up looking something like this: 

```yaml
spec:
  catalogSource: catalogsource-sample
  image: quay.io/operatorhubio/argocd-operator@sha256:5f4533de114e9301e836155190dc2c66a969f55085af03a4959a2fb0af74e1f4
  package: argocd-operator
  properties:
  - type: olm.gvk
    value: ewogICAgICAgICAgICAgICAgImdyb3VwIjogImFyZ29wcm9qLmlvIiwKICAgICAgICAgICAgICAgICJraW5kIjogIkFwcFByb2plY3QiLAogICAgICAgICAgICAgICAgInZlcnNpb24iOiAidjFhbHBoYTEiCiAgICAgICAgICAgIH0=
  - type: olm.gvk
    value: ewogICAgICAgICAgICAgICAgImdyb3VwIjogImFyZ29wcm9qLmlvIiwKICAgICAgICAgICAgICAgICJraW5kIjogIkFwcGxpY2F0aW9uIiwKICAgICAgICAgICAgICAgICJ2ZXJzaW9uIjogInYxYWxwaGExIgogICAgICAgICAgICB9
  - type: olm.gvk
    value: ewogICAgICAgICAgICAgICAgImdyb3VwIjogImFyZ29wcm9qLmlvIiwKICAgICAgICAgICAgICAgICJraW5kIjogIkFwcGxpY2F0aW9uU2V0IiwKICAgICAgICAgICAgICAgICJ2ZXJzaW9uIjogInYxYWxwaGExIgogICAgICAgICAgICB9
  - type: olm.gvk
    value: ewogICAgICAgICAgICAgICAgImdyb3VwIjogImFyZ29wcm9qLmlvIiwKICAgICAgICAgICAgICAgICJraW5kIjogIkFyZ29DRCIsCiAgICAgICAgICAgICAgICAidmVyc2lvbiI6ICJ2MWFscGhhMSIKICAgICAgICAgICAgfQ==
  - type: olm.gvk
    value: ewogICAgICAgICAgICAgICAgImdyb3VwIjogImFyZ29wcm9qLmlvIiwKICAgICAgICAgICAgICAgICJraW5kIjogIkFyZ29DREV4cG9ydCIsCiAgICAgICAgICAgICAgICAidmVyc2lvbiI6ICJ2MWFscGhhMSIKICAgICAgICAgICAgfQ==
  - type: olm.package
    value: ewogICAgICAgICAgICAgICAgInBhY2thZ2VOYW1lIjogImFyZ29jZC1vcGVyYXRvciIsCiAgICAgICAgICAgICAgICAidmVyc2lvbiI6ICIwLjUuMCIKICAgICAgICAgICAgfQ==
  relatedImages:
  - image: gcr.io/kubebuilder/kube-rbac-proxy:v0.8.0
    name: ""
  - image: quay.io/argoprojlabs/argocd-operator@sha256:2f5c0d4567607266ccacb91d8c2e3b18c2afe0edaf55855dcb1a06b02173b520
    name: ""
  - image: quay.io/operatorhubio/argocd-operator@sha256:5f4533de114e9301e836155190dc2c66a969f55085af03a4959a2fb0af74e1f4
    name: ""
```  

We'd discussed correcting this and changing the plumbing to make sure we throw in unmarshalled structs there instead, but again, that was superseded with https://github.com/operator-framework/catalogd/issues/38. Hence the tap dance with marshal/unmarshal: just a stop gap so that we're not sweating over details that are going to change soon.

internal/resolution/entitysources/catalogdsource.go
```diff
@@ -0,0 +1,142 @@
...
+				}
+				pkgValue, err := jsonMarshal(tmp)
+				if err != nil {
+					return nil, err
+				}
+				props[property.TypePackage] = string(pkgValue)
+			}
+		}
+
+		imgValue, err := jsonMarshal(bundle.Spec.Image)
```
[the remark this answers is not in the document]
anik120: https://github.com/operator-framework/operator-controller/pull/182#discussion_r1176960714

I.e tap dance as stop gap

internal/resolution/variable_sources/entity/bundle_entity.go
anik120: Oops. Looks like this was part of the v0 CatalogSource integration PR (that got reverted as part of the first commit of this PR which reverts the entire thing). @ankitathomas @perdasilva if you could help me with cherry-picking the relevant parts of that PR that should be included back, that'll save me some time from sifting through the v0 integration PR 😀

internal/resolution/variable_sources/entity/bundle_entity_test.go
anik120: Again part of the v0 CatalogSource integration that the first commit reverts. cc: @ankitathomas @perdasilva 

So looks like bundle_entity.go and bundle_entity_test.go gets affected by the revert. I'm happy to just...revert..the...revert for these two files, but it's probably prudent to have you guys chime in with the context for these kind of changes first so that I am sure of what I'm doing :)

internal/resolution/entitysources/catalogdsource.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+func (es catalogdEntitySource) Iterate(ctx context.Context, fn input.IteratorFunction) error {
+	for _, entity := range es {
+		if err := fn(&entity); err != nil {
+			return err
+		}
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
anik120: > The biggest change is that it makes the list call and builds entities every time these calls are made. A future optimization would be to implement caching

So what is the objective we're trying to achieve with this, today? This definitely is an interesting idea. My brain just molded into the format of the two entitySources I observed before this (removed as a part of this PR, and replaced with the catalogd entitysource), but I'm not sure I understand the value of doing this in achieve the goal "first pass over of the integration with catalogd"

internal/resolution/resolver.go
anik120: This definitely is a larger conversation to bring up in the upstream community meetings so that it can be discussed and endorsed by the community first

internal/resolution/entitysources/catalogdsource.go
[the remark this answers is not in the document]
stevekuznetsov: @tmshort if it's required, the validation should occur at creation or mutation of the object, not downstream every time the value is used.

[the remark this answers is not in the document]
stevekuznetsov: @anik120 is there an understanding of how long that effort will take? Leaving the software in a more stable spot while the change happens might be useful if it's a long-running effort?

internal/resolution/entitysources/catalogdsource.go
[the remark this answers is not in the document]
stevekuznetsov: If this only has meaning as compared to some hardcoded thing being ripped out right now I'd just say leave the extra adjectives out.

internal/resolution/resolver.go
stevekuznetsov: Sure - the tool runs in a couple milliseconds so I don't imagine it will be a hard thing to convince, but definitely out of scope for this PR

test/e2e/install_test.go
anik120: Consolidating comments about e2e here. 

@stevekuznetsov : "Looks like an OLMv0 end-to-end test was removed - is there a new one for this integration?"

I realized after removing the hardcoded entity source that the v0 CatalogSource integration was using the v0 OLM installation to create a CatalogSource with an [index image](https://github.com/operator-framework/operator-controller/blob/4d6757a2b0fe7dda779d03d44f37d20872218711/test/e2e/catalogsource_install_test.go#L78), but then for us to include prod grade e2e tests we probably have to set up the e2e test infrastructure first. That means
* Discuss if we want to install catalogd v0.0.1 as part of the operator-controller installation 
* Set up test infrastructure like a mock image registry, that can be used to build and post an index image into so that the tests can pull from it etc. 

My first instinct is to add the hardcoded entitySource back, so that the deppy tests can go on with the hardcoded entitySource, and we don't increase the scope of this PR with "set up e2e test infrastructure for this project", but then if we feel like we must do it as a part of this PR then I'm happy to do it, with an agreement that it'll increase the scope of this PR. 

cc: @joelanford

anik120: > Discuss if we want to install catalogd v0.0.1 as part of the operator-controller installation

My guess for this is "yes we want this to be included in this PR, just like we install rukpak", but I'll wait for others to chime in first, to make sure we're comfortable including v0.1.0 release of catalogd as a part of the operator-contorller installation

internal/resolution/entitysources/catalogdsource.go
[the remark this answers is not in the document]
anik120: It's definitely a "the very next step" for the project. We'd planned it for milestone 3 but it got pushed out in favor of us focusing our effort on the integration for m3, so I'd say milestone 4 is a pretty accurate timeline for when it's about to happen. (We can probably do a v0.2.0 release after m4 and in m5 we can update this integration with the v0.2.0 release, but don't quote me on that yet I'd like to make sure others working in the project are in agreement with "we can do these things we're promising in that timeline" too :) ). But the gist of it is, these are going to change pretty soon.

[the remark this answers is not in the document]
anik120: > Mistakes happen? And we should always validate user-derived input?

> @tmshort if it's required, the validation should occur at creation or mutation of the object, not downstream every time the value is used.

All makes sense. I've made a note of it [here](https://github.com/operator-framework/catalogd/issues/38#issuecomment-1522391862) so that we don't miss this

internal/resolution/resolver.go
anik120: It isn't so much about the run time, than it is about "we're about to use an openshift product for this project". We've been having conversations about getting away from open shift native products for our projects (prow/openshift-ci etc) in favor of more cncf native projects, so this feels like a good item to include in that list too.

internal/resolution/entitysources/catalogdsource.go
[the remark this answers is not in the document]
everettraven: > So what is the objective we're trying to achieve with this, today?

The main objective is to not have to instantiate a new entity source every single time you want to resolve.

>My brain just molded into the format of the two entitySources I observed before this (removed as a part of this PR, and replaced with the catalogd entitysource)

I'm not sure I'm following here - from what I could tell the way that the two entity sources before this one were both designed with the idea in mind that you create it once not create it every time you need to resolve something. In my mind this implementation seems to change the pattern introduced by the other entity sources. The hardcoded entity source was only "static" in the sense that it hardcoded its responses but appeared to be "dynamic" to the controller since the idea was you instantiate it once and use it everywhere. The same goes for the OLMv0 integration - it used a controller to keep track of `CatalogSource` CRs and update its cache accordingly.

Maybe the verbiage of "static" and "dynamic" is misleading here but what I am meaning is that *I* feel that we should be following the pattern set by the previous entity sources of using one instantiation everywhere instead of instantiating every single time we want to use the resolver.

> but I'm not sure I understand the value of doing this in achieving the goal "first pass over of the integration with catalogd"

I don't think there is much difference in value but from what I can tell, just that this PR appears to change the pattern established by the previous two entity sources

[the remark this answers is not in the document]
everettraven: Another note on this - based on [this comment](https://github.com/operator-framework/operator-controller/pull/141#issuecomment-1516407385) I don't think we are following the "on the fly" approach proposed since we are loading everything at instantiation and instantiating it every time we need to resolve. 

*Maybe* it falls into the optimization territory but I think taking the approach of instantiate once and use a client to fetch when certain functions are called is more aligned with what was proposed by the comment I linked

controllers/operator_controller.go
everettraven: Going to point to a couple comments I made as a response to another comment:
- https://github.com/operator-framework/operator-controller/pull/182#discussion_r1177068140
- https://github.com/operator-framework/operator-controller/pull/182#discussion_r1177071919

The gist is that I feel we are going against the entity source pattern established by the previous two entity sources by requiring an instantiation for every time we want to resolve vs instantiating once and using it everywhere. If this is really something we don't mind and want to address as an optimization that's fine by me, just something I wanted to point out since it stood out to me.

.gitignore
```diff
@@ -1,3 +1,4 @@
+vendor
```
joelanford: Is there some reason that this needs to be `gitignore`'d? I don't think anything in the repo generates a vendor dir, so I don't think we ever expect it to exist.

Makefile
```diff
@@ -147,7 +146,7 @@ release: goreleaser ## Runs goreleaser for the operator-controller. By default,
 quickstart: export MANIFEST="https://github.com/operator-framework/operator-controller/releases/download/$(VERSION)/operator-controller.yaml"
 quickstart: kustomize generate ## Generate the installation release manifests and scripts
 	kubectl kustomize config/default | sed "s/:devel/:$(VERSION)/g" > operator-controller.yaml
-	envsubst '$$OLM_V0_VERSION,$$CERT_MGR_VERSION,$$RUKPAK_VERSION,$$MANIFEST' < scripts/install.tpl.sh > install.sh
+	envsubst '$$CERT_MGR_VERSION,$$RUKPAK_VERSION,$$MANIFEST' < scripts/install.tpl.sh > install.sh
```
joelanford, line 149: Do we need to add the `catalogd` version here?

controllers/operator_controller.go
joelanford: I'm a little confused as well.
- In our `main.go`, I'd expect to create the resolver with the "real" catalogd entity source, and then set the `reconciler.Resolver` in main.go
- In our tests, we're also seting up the reconciler, but there, we are using a mock (hardcoded) entity source. In that case, we setup the resolver with that entity source and set the `reconciler.Resolver` in the test file.

In both cases, the actual reconciler expects the resolver to be non-nil.

internal/resolution/entitysources/catalogdsource.go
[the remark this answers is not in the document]
joelanford: Agree with @everettraven. I think the `New` function should return a struct that has a single `client.Client` field. Then we use that client during calls to the entity source functions to actually fetch the metadata on the fly. This theoretically ensures we get up-to-date metadata every time (and yes, I realize our existing CRD approach means there are probably some holes in that theory, but I'm not terribly concerned since we know we're going back to that API in a future milestone anyway).

internal/resolution/entitysources/catalogdsource.go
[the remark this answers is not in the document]
joelanford: I'm not sure I'm understanding the need to unmarshal/marshal. I know that deppy wants maps of properties, not lists, so I would expect some handling to do that conversion. But isn't this just a matter of collecting up the properties that have the same type into `map[string][]json.Message`, and then building the deppy entities from that?

internal/resolution/entitysources/catalogdsource.go
joelanford: Good question. This PR is highlighting to me that we need some specific focus on the translation of FBC to what deppy needs. It seems there's an (unnecessary?) impedance mismatch that is necessitating some of this.

internal/resolution/entitysources/catalogdsource.go
```diff
@@ -0,0 +1,142 @@
...
+
+func jsonMarshal(p interface{}) ([]byte, error) {
```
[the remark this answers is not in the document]
joelanford: json.Marshal always escapes for HTML.

joelanford (commented):
This isn't really what I had in mind when we talked about it before. I was anticipating the dynamic entity source that Bryce was alluding to in comments. Having said that, I'm not necessarily opposed to the "read everything into memory at init time" approach because it _technically_ seems like it will meet the bar set out in the demo script.

I _do_ think we should setup the entity source once and remove the `if r.Resolver == nil` block in the reconciler though. I know doing that may necessitate making this more dynamic though.

controllers/operator_controller_test.go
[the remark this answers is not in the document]
joelanford: And just to put a bow on this, this package can contain multiple controllers and therefore multiple controller test roots. So its important for each of those roots to be descriptive about which controller the tests are for.

internal/resolution/entitysources/catalogdsource.go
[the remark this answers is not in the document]
joelanford: Sorry! Read right over the fact that you asked specifically about the HTML aspect. HTML encoding subs out `<`, `>`, and `&` for unicode escape sequences, which isn't super human-friendly. In the case of our FBC code in operator-registry (where I suspect @anik120 found this?), this is important because we expect humans to be looking at the output.

In this case, I'm not sure whether humans will ever see anything downstream of this marshaling. Theoretically, I suppose it could show up in resolution error messages?

internal/resolution/resolver.go
joelanford: Just because something exists under an openshift repo doesn't mean its a problem. If it solves a problem and doesn't give RH or RH-products preferential treatment, it's fine.

There are also other reasonable upstream tools if we're averse to that for some reason. I've used https://github.com/rinchsan/gosimports

internal/resolution/entitysources/catalogdsource.go
joelanford: To expand on my previous comment, this is not an actual FBC property. This is something sort of only happening at the deppy layer. I _think_ this will go away when @perdasilva removes the entity source concept and replaces it with a variable source.

internal/resolution/variable_sources/entity/bundle_entity.go
joelanford: Default channel isn't a concept in OLMv1 yet anyway (and it may never be). So perhaps we just fully remove anything referencing default channels?

test/e2e/install_test.go
joelanford: > Discuss if we want to install catalogd v0.0.1 as part of the operator-controller installation

Yes, 100%.

> Set up test infrastructure like a mock image registry, that can be used to build and post an index image into so that the tests can pull from it etc.

Agree with doing this separately. There's a whole bunch wrapped up in this question:
1. Should we do the `catalog.spec.source` union type first so that we don't need all the image infrastructure everywhere we want to test it?
2. What tests should we put in `catalogd` vs `operator-controller`?
3. What's the appropriate layer for the tests? Unit or e2e? Or both?

internal/resolution/entitysources/catalogdsource.go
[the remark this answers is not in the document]
stevekuznetsov: Wouldn't we expect to HTML-escape just before putting the output into a webpage or something? I am still lost as to why we're not operating over raw values here.

internal/resolution/resolver.go
stevekuznetsov: Yeah it's a super trivial utility .... is OpenShift is that much of a dirty word?

internal/resolution/entitysources/catalogdsource.go
```diff
@@ -0,0 +1,159 @@
...
+	entities, err := getEntities(ctx, es.client)
+	if err != nil {
+		return nil, err
+	}
+	for _, entity := range entities {
+		if filter(&entity) {
+			resultSet = append(resultSet, entity)
+		}
+	}
+	return resultSet, nil
```
joelanford, line 46: Nit: in `Filter` and `GroupBy`, is it possible to call `Iterate` with an iterate lambda function that implements the filter and groupBy logic? That way we don't have to repeat ourselves quite as much (and thus reduce the chance for drift between implementations)

internal/resolution/entitysources/catalogdsource.go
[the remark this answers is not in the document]
joelanford: This is _disabling_ HTML escaping, which is the hardcoded only option in `json.Marshal`. 

However, I think we could probably avoid doing any json marshaling in this entity source. I mentioned in a different comment that there _is_ some impedance mismatch between FBC properties (`[]Property`) and deppy properties (`map[string]string`, I think?), which does seem awkward, but since we already have `json.RawMessage`, it seems like we could do the conversion without needing to marshal/unmarshal anything.

internal/resolution/entitysources/catalogdsource.go
```diff
@@ -0,0 +1,159 @@
...
+				}
+				tmp := pkg{}
+				if err := json.Unmarshal(prop.Value, &tmp); err != nil {
+					return nil, err
+				}
+				pkgValue, err := jsonMarshal(tmp)
+				if err != nil {
+					return nil, err
+				}
+				props[property.TypePackage] = string(pkgValue)
```
joelanford, line 116: This needs a comment explaining what's going on. To my eye, it's literally unmarshalling and then re-marshaling, which seems like a no-op. Why not consolidate this down to:
```go
case property.TypePackage:
	props[property.TypePackage] = string(prop.Value)
```

Better yet, I'd imagine something more like this:
```go
valsByType := map[string][]json.RawMessage{}
for _, prop := range bundle.Spec.Properties {
	valsByType[prop.Type] = append(valsByType[prop.Type], prop.Value)
}
entityProps := map[string]string{}
for typ, vals := range valsByType {
	switch typ {
	case property.TypePackage:
		// We expect exactly on "olm.package" property
		if len(vals) != 1 { return err }
		entityProps[typ] = string(vals[0])
	default:
		// All other properties can have 0 or more values, so always use a list
		entityProps[typ] = fmt.Sprintf("[%s]", bytes.Join(vals, []byte(",")))
	}
}
```

internal/resolution/entitysources/catalogdsource.go
```diff
@@ -0,0 +1,159 @@
...
+				}
+			}
+		}
+		channelValue, _ := jsonMarshal(property.Channel{ChannelName: channelName, Priority: 0})
+		props[property.TypeChannel] = string(channelValue)
+		entity := input.Entity{
+			ID:         deppy.IdentifierFromString(fmt.Sprintf("%s%s%s", bundle.Name, bundle.Spec.Package, channelName)),
+			Properties: props,
+		}
+		entities = append(entities, entity)
```
joelanford, line 140: This should go inside the channel loop.

For a given bundle, deppy expects the entity source to return an entity for each channel that the bundle is in (under a different ID). We can't just pick one of the channels and call it a day. So if a bundle is an entry in three channels, that should produce three entities.

internal/resolution/entitysources/catalogdsource.go
```diff
@@ -0,0 +1,159 @@
...
+// catalogdEntitySource is a source for(/collection of) deppy defined input.Entity, built from content
+// made accessible on-cluster by https://github.com/operator-framework/catalogd.
+// It is an implementation of deppy defined input.EntitySource
+type catalogdEntitySource struct {
+	client client.Client
+}
+
+// type catalogdEntitySource map[deppy.Identifier]input.Entity
+
+func NewCatalogdEntitySource(ctx context.Context, client client.Client) catalogdEntitySource {
```
joelanford, line 26: No need to pass a context here anymore.

internal/resolution/entitysources/catalogdsource.go
```diff
@@ -0,0 +1,159 @@
...
+	"github.com/operator-framework/operator-registry/alpha/property"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	catalogd "github.com/operator-framework/catalogd/pkg/apis/core/v1beta1"
+)
+
+// catalogdEntitySource is a source for(/collection of) deppy defined input.Entity, built from content
+// made accessible on-cluster by https://github.com/operator-framework/catalogd.
+// It is an implementation of deppy defined input.EntitySource
+type catalogdEntitySource struct {
```
joelanford, line 20: I think I'd recommend exporting this type, at least if we're going to return this type from the `New` function. The other option is to return `input.EntitySource`.

Returning an unexported type can be awkward for callers because they can use it in the scope of where it was returned but they can't pass it to other functions because they can't refer to it due to it being unexported.

main.go
joelanford: That comment is a sentinel comment used by kubebuilder/operator-sdk to mutate this section of code when you run `operator-sdk create api ...`. It's probably safe to remove, tbh.

joelanford (commented):
Nice update Anik! This is shaping up pretty well, I think.

controllers/operator_controller_test.go
```diff
@@ -875,19 +875,19 @@ func verifyConditionsInvariants(op *operatorsv1alpha1.Operator) {
 
 var testEntitySource = input.NewCacheQuerier(map[deppy.Identifier]input.Entity{
 	"operatorhub/prometheus/0.37.0": *input.NewEntity("operatorhub/prometheus/0.37.0", map[string]string{
-		"olm.bundle.path": "quay.io/operatorhubio/prometheus@sha256:3e281e587de3d03011440685fc4fb782672beab044c1ebadc42788ce05a21c35",
+		"olm.bundle.path": `"quay.io/operatorhubio/prometheus@sha256:3e281e587de3d03011440685fc4fb782672beab044c1ebadc42788ce05a21c35"`,
```
stevekuznetsov, line 878: This file still has some double quoting going on, unclear if expected?

internal/resolution/entitysources/catalogdsource.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+// type catalogdEntitySource map[deppy.Identifier]input.Entity
+
+func NewCatalogdEntitySource(ctx context.Context, client client.Client) catalogdEntitySource {
+
+	return catalogdEntitySource{client: client}
+}
+
+func (es catalogdEntitySource) Get(ctx context.Context, id deppy.Identifier) *input.Entity {
```
stevekuznetsov, line 31: Similarly, I think we want a pointer receiver, or these methods will run with a value-copy of `client`

internal/resolution/entitysources/catalogdsource.go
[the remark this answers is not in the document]
joelanford: We do something like this to disable HTML encoding in `opm render` because we expect humans to be looking at the output.

Here, the resolver will see this. The only place I could see a human being exposed to resolver input is in resolver error messages, and it's conceivable that property values will end up there.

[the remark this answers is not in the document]
stevekuznetsov: I think I'm still really lost. Even in the resolver if I'm reading this code correctly a `json.Unmarshal` happens first so it would be back to raw values ... In user-facing output of JSON (which in and of itself is a bit of an oxymoron) this makes sense but I think unless there's some clear need, starting with a simple call to `json.Marshal` makes sense ...

internal/resolution/variable_sources/entity/bundle_entity.go
```diff
@@ -212,6 +203,19 @@ func (b *BundleEntity) loadChannelProperties() error {
...
+		bundlePath, err := loadFromEntity[string](b.Entity, PropertyBundlePath, required)
+		if err != nil {
+			return fmt.Errorf("error determining bundle path for entity '%s': %w", b.ID, err)
+		}
+		b.bundlePath = bundlePath
+	}
+	return nil
+}
+
 func loadFromEntity[T interface{}](entity *input.Entity, propertyName string, required propertyRequirement) (T, error) {
```
perdasilva, line 219: Not related to this PR, but it would be awesome if you could change this to `[T any]` =D

main.go
```diff
@@ -92,20 +94,10 @@ func main() {
...
-	if err := catsrcReconciler.SetupWithManager(mgr); err != nil {
-		setupLog.Error(err, "unable to create catalog source controller", "controller", "CatalogSource")
-		os.Exit(1)
-	}
-
 	if err = (&controllers.OperatorReconciler{
 		Client:   mgr.GetClient(),
 		Scheme:   mgr.GetScheme(),
-		Resolver: resolution.NewOperatorResolver(mgr.GetClient(), catsrcReconciler),
+		Resolver: resolution.NewOperatorResolver(mgr.GetClient(), entitysources.NewCatalogdEntitySource(context.TODO(), mgr.GetClient())),
```
perdasilva, line 100: would it make sense to use the same context (or child context) as the manager `ctrl.SetupSignalHandler()` ?

internal/resolution/entitysources/catalogdsource.go
anik120: I'm going with a pointer receiver anyway since the struct will likely grow to include other stuff (eg a cache)

config/rbac/role.yaml
```diff
@@ -6,30 +6,31 @@ metadata:
   name: manager-role
 rules:
 - apiGroups:
-  - ""
+  - catalogd.operatorframework.io
   resources:
-  - events
+  - bundlemetadata
```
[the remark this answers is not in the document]
anik120: They're not. This is part of the first commit, which reverts v0 integration. I'm guessing events were watched somewhere in that implementation.

controllers/operator_controller_test.go
anik120: @perdasilva so do I need to put get rid of the quotes (that the revert adds in)? Fwiw the unit test cases pass without any problem now.

main.go
anik120: Thank you! Wasn't sure which context was available to us here 🙏🏽

internal/resolution/entitysources/catalogdsource.go
```diff
@@ -0,0 +1,157 @@
...
+)
+
+// catalogdEntitySource is a source for(/collection of) deppy defined input.Entity, built from content
+// made accessible on-cluster by https://github.com/operator-framework/catalogd.
+// It is an implementation of deppy defined input.EntitySource
+type catalogdEntitySource struct {
+	client client.Client
+}
+
+func NewCatalogdEntitySource(ctx context.Context, client client.Client) *catalogdEntitySource {
```
joelanford, line 24: Can't tell if my comment was lost. But this `ctx` parameter is no longer necessary in this function signature.

internal/resolution/entitysources/catalogdsource.go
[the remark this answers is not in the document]
joelanford: :+1: Seems reasonable to me to start with `json.Marshal` and re-evaluate if there's a need.

Makefile
```diff
@@ -159,7 +159,7 @@ endif
 install: export MANIFEST="./operator-controller.yaml"
 install: manifests kustomize generate ## Install CRDs into the K8s cluster specified in ~/.kube/config.
 	kubectl kustomize config/default > operator-controller.yaml
-	envsubst '$$OLM_V0_VERSION,$$CERT_MGR_VERSION,$$RUKPAK_VERSION,$$MANIFEST' < scripts/install.tpl.sh | bash -s
+	envsubst '$$CATALOGD_VERSION,$$CERT_MGR_VERSION,$$RUKPAK_VERSION,$$MANIFEST' < scripts/install.tpl.sh | bash -s
```
joelanford, line 162: OCD nit: use the same ordering of envvar names here and in quickstart.

internal/resolution/entitysources/catalogdsource.go
anik120: I feel like I might have experienced the awkwardness myself somewhere, but I am not sure if it was exactly this? With what I have in my mind right at this moment, this is just the OOPs concept of "Objects should not be directly accessed by anything outside the package it's defined in, but should be accessible to the world only through exported methods, that can control what information is exposed to the world"

anik120: But having said that, I don't think I've heard that principle repeated in the context of Golang yet though. I'm still inclined to keep it unexported, and export it when we see we need to (and that'll be added to my TIL moments too), but if you insist on doing this change now I'm happy to go ahead and do it anyway.

main.go
anik120: Actually, looks like that function has some dealings with a channel that gets closed when you call that function, and trying to use that context variable twice just leads to: 

```
2023-04-27T20:07:30Z	INFO	controller-runtime.metrics	Metrics server is starting to listen	{"addr": "127.0.0.1:8080"}
2023-04-27T20:07:30Z	INFO	setup	starting manager
panic: close of closed channel

goroutine 1 [running]:
sigs.k8s.io/controller-runtime/pkg/manager/signals.SetupSignalHandler()
	/Users/anbhatta/go/src/github.com/operator-framework/operator-controller/vendor/sigs.k8s.io/controller-runtime/pkg/manager/signals/signal.go:31 +0x25
main.main()
	/Users/anbhatta/go/src/github.com/operator-framework/operator-controller/main.go:116 +0x7cc
```

I'm just reverting this to context.TODO() for now and we can definitely look into it later.

anik120: Oh wait, we don't need this anymore anyway https://github.com/operator-framework/operator-controller/pull/182#discussion_r1179603532 XD

Makefile
anik120: Done

internal/resolution/entitysources/catalogdsource.go
anik120: Oh yea that is a very good catch. Updating.

controllers/operator_controller_test.go
joelanford: Leave as is. I think it's expected to be a JSON string, which needs to retain quotes.

internal/resolution/entitysources/catalogdsource.go
joelanford: We're in `internal` here, so it really doesn't matter all that much since we're the only clients. Let's just go with this for now.

test/e2e/install_test.go
```diff
@@ -27,62 +28,62 @@ var _ = Describe("Operator Install", func() {
...
+				Spec: operatorv1alpha1.OperatorSpec{
+					PackageName: pkgName,
+				},
+			}
+			operatorCatalog := &catalogd.CatalogSource{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "test-catalog",
+				},
+				Spec: catalogd.CatalogSourceSpec{
+					Image: "quay.io/operatorhubio/catalog:latest",
```
joelanford, line 49: This is a very large catalog to use in an e2e. Okay for now, but can you add a TODO comment and an issue in this repo to replace this with something we build locally/on-the-fly as part of the e2e test and make available during the e2e?

test/e2e/install_test.go
```diff
@@ -27,62 +28,62 @@ var _ = Describe("Operator Install", func() {
 		operatorName string
 		operator     *operatorv1alpha1.Operator
 	)
+	When("An operator is installed from an operator catalog", func() {
+		BeforeEach(func() {
```
joelanford, line 32: I think we need an `AfterEach` to cleanup this catalog that we create? Probably should move the operator delete to the `AfterEach` as well to mirror where it gets created.

test/e2e/install_test.go
```diff
@@ -27,62 +28,62 @@ var _ = Describe("Operator Install", func() {
...
-		testNamespace = createTestNamespace(ctx, c, "registry-grpc-")
-		cleanup = applyCRDifNotPresent(ctx)
-		testPrefix := "registry-grpc-"
+			By("eventually reporting a successful resolution and bundle path")
+			Eventually(func(g Gomega) {
+				err = c.Get(ctx, types.NamespacedName{Name: operator.Name}, operator)
+				g.Expect(err).ToNot(HaveOccurred())
+				g.Expect(len(operator.Status.Conditions)).To(Equal(1))
+				g.Expect(operator.Status.Conditions[0].Message).To(Equal("install was successful"))
+			}).WithTimeout(defaultTimeout).WithPolling(defaultPoll).Should(Succeed())
```
joelanford, line 72: It it necessary to pass `g Gomega` to the function? I would expect us to be able to do something like this:
```go
Eventually(func() {
    Expect(...)
})
```

internal/resolution/entitysources/catalogdsource.go
```diff
@@ -0,0 +1,155 @@
...
+)
+
+// catalogdEntitySource is a source for(/collection of) deppy defined input.Entity, built from content
+// made accessible on-cluster by https://github.com/operator-framework/catalogd.
+// It is an implementation of deppy defined input.EntitySource
+type catalogdEntitySource struct {
+	client client.Client
+}
+
+func NewCatalogdEntitySource(client client.Client) *catalogdEntitySource {
```
stevekuznetsov, line 24: Is there a compelling reason not to return `input.EntitySource`?

internal/resolution/entitysources/catalogdsource.go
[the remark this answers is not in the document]
stevekuznetsov: Can we file an issue to clean this up?

internal/resolution/entitysources/catalogdsource.go
joelanford: There's a go proverb "Accept interfaces, return concrete types". So my initial argument was "let's just export this", which is fine because its internal. But since its internal, its hard to care that much because we can always change it.

test/e2e/install_test.go
anik120: Yup I'd discussed this yesterday and cc:@theishshah I'm adding the TODO and creating the issue so that we can keep track of this

test/e2e/install_test.go
anik120: Turns out: 

```
 [FAILED] The function passed to Eventually had an invalid signature of func().  Functions passed to Eventually must either:

  	(a) have return values or
  	(b) take a Gomega interface as their first argument and use that Gomega instance to make assertions.

  You can learn more at https://onsi.github.io/gomega/#eventually
```

anik120: I know we're used to the first approach in our existing test suites, but if we want to discuss the merits of a) vs b) we can do that in a separate PR. For now I've reverted back to using b)

joelanford: Oh, that function needs to return something so it can be asserted against.

internal/resolution/variable_sources/entity/bundle_entity.go
```diff
@@ -212,6 +203,19 @@ func (b *BundleEntity) loadChannelProperties() error {
 	return nil
 }
 
+func (b *BundleEntity) loadBundlePath() error {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	if b.bundlePath == "" {
+		bundlePath, err := loadFromEntity[string](b.Entity, PropertyBundlePath, required)
```
everettraven, line 210: I'm not sure that this is necessary. From my testing, this particular line is what makes us have to do weird marshalling and unmarshalling logic [here](https://github.com/operator-framework/operator-controller/pull/182/files#diff-1b647d2198aaf1054b9a90ee4b1ec25fbcc7ed2b7cd68d63e6a89008a3dfe6f8R92-R96)

I think using the `loadFromEntity()` function makes the most sense when you are attempting to load something you are expecting to follow a specific structure and gets unmarshalled into a `struct`. For this case, since we are expecting an unstructured `string` (`"hello world!"` as opposed to `"{"test": "test"}"`) and the `b.Entity.Properties` is a `map[string]string` I think we are better off just doing something like:
```suggestion
		bundlePath, ok := b.Entity.Properties[PropertyBundlePath]
```

Using the above suggestion will allow us to change [this](https://github.com/operator-framework/operator-controller/pull/182/files#diff-1b647d2198aaf1054b9a90ee4b1ec25fbcc7ed2b7cd68d63e6a89008a3dfe6f8R92-R96) to:
```go
props["olm.bundle.path"] = bundle.Spec.Image
```

Makefile
```diff
@@ -6,8 +6,8 @@ export IMAGE_REPO ?= quay.io/operator-framework/operator-controller
 export IMAGE_TAG ?= devel
 export GO_BUILD_TAGS ?= upstream
 export CERT_MGR_VERSION ?= v1.9.0
+export CATALOGD_VERSION ?= v0.1.2
```
everettraven, line 9:
```suggestion
export CATALOGD_VERSION ?= v0.1.3
```

New catalogd version released with fixes to the image architectures

test/e2e/install_test.go
anik120: Yea that's what I meant in my previous comment, that we're using the func signature you mentioned in other test suites but changing the current approach in the code base seems like a follow up PR

.gitignore
anik120: You have to `go mod vendor` to build the manager binary, and you either have to delete it every time you're about to push to your PR or never do `git add .` :) I don't think it hurts to keep it here

joelanford: > You have to `go mod vendor` to build the manager binary

I don't have to do that...

anik120:
Follow up sketch here: https://github.com/operator-framework/operator-controller/issues/187

config/rbac/role.yaml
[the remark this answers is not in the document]
tmshort: Yes, events were create/patch in the original catsrc commit 2973a14

controllers/operator_controller.go
```diff
@@ -121,7 +115,6 @@ func (r *OperatorReconciler) reconcile(ctx context.Context, op *operatorsv1alpha
 		})
 		return ctrl.Result{}, nil
 	}
-
```
tmshort, line 124: Nit: deletion of a blank line/whitepsace. This goes against the "minimizing" the diff, but I don't care enough.

tmshort:
Just a note that catsrc integration was added in 2973a14, this PR deletes all the added files from that commit (at least), and removes other edits of that commit.

tmshort (approved):
LGTM, although so many other people reviewed this, I'd at least like to get one of them to give it 👍

controllers/operator_controller.go
```diff
@@ -55,15 +55,9 @@ type OperatorReconciler struct {
...
-// Reconcile is part of the main kubernetes reconciliation loop which aims to
-// move the current state of the cluster closer to the desired state.
-// TODO(user): Modify the Reconcile function to compare the state specified by
-// the Operator object against the actual cluster state, and then
-// perform operations to make the cluster state reflect the state specified by
-// the user.
-//
-// For more details, check Reconcile and its Result here:
-// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.13.1/pkg/reconcile
+//+kubebuilder:rbac:groups=catalogd.operatorframework.io,resources=bundlemetadata,verbs=get;list;watch
```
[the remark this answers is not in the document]
varshaprasad96: @anik120 took another pass, I maybe wrong but do we also _need_ get and list for those GVKs? The catalog controller is fetching bundle metadata and packages, we are getting the content through the manager's client and passing in the [`entities`](https://github.com/operator-framework/operator-controller/pull/182/files#diff-1b647d2198aaf1054b9a90ee4b1ec25fbcc7ed2b7cd68d63e6a89008a3dfe6f8R112) to the controller. The operator controller doesn't seem to directly interact with both the APIs (pkgs or BMD). Given that, does it need the rbac to get and list those resources?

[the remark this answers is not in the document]
joelanford: We need list and watch if we use the caching client, right?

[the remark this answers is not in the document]
joelanford: We can put these rbac comments in the entity source implementation and update the rbac generator invocation to find them there if we want to keep the rules nearest the place that needs them. But that could be a follow up.

varshaprasad96 (approved):
/lgtm

controllers/operator_controller.go
```diff
@@ -55,15 +55,9 @@ type OperatorReconciler struct {
...
-// move the current state of the cluster closer to the desired state.
-// TODO(user): Modify the Reconcile function to compare the state specified by
-// the Operator object against the actual cluster state, and then
-// perform operations to make the cluster state reflect the state specified by
-// the user.
-//
-// For more details, check Reconcile and its Result here:
-// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.13.1/pkg/reconcile
+//+kubebuilder:rbac:groups=catalogd.operatorframework.io,resources=bundlemetadata,verbs=get;list
+//+kubebuilder:rbac:groups=catalogd.operatorframework.io,resources=packages,verbs=get;list
```
joelanford, line 59: Because we're using `mgr.GetClient()` when we setup the catalogd entity source, that means we're using the caching client. The caching client needs `list` and `watch` verbs. `get` isn't necessary because the caching client always gets from the cache and never actually issues get requests to the apiserver.

If you create the sample catalog and operator and then watch the logs based on this commit, you'll see lines like this:
```
E0505 12:53:49.867260       1 reflector.go:140] pkg/mod/k8s.io/client-go@v0.26.1/tools/cache/reflector.go:169: Failed to watch *v1beta1.Package: unknown (get packages.catalogd.operatorframework.io)
E0505 12:53:50.229906       1 reflector.go:140] pkg/mod/k8s.io/client-go@v0.26.1/tools/cache/reflector.go:169: Failed to watch *v1beta1.BundleMetadata: unknown (get bundlemetadata.catalogd.operatorframework.io)
E0505 12:53:52.993949       1 reflector.go:140] pkg/mod/k8s.io/client-go@v0.26.1/tools/cache/reflector.go:169: Failed to watch *v1beta1.Package: unknown (get packages.catalogd.operatorframework.io)
E0505 12:53:53.228923       1 reflector.go:140] pkg/mod/k8s.io/client-go@v0.26.1/tools/cache/reflector.go:169: Failed to watch *v1beta1.BundleMetadata: unknown (get bundlemetadata.catalogd.operatorframework.io)
```

```suggestion
//+kubebuilder:rbac:groups=catalogd.operatorframework.io,resources=bundlemetadata,verbs=list;watch
//+kubebuilder:rbac:groups=catalogd.operatorframework.io,resources=packages,verbs=list;watch
```

anik120: Yikes. Updated.

Issue #196, opened 2023-05-09
Update e2e package tests


perdasilva:
## Description
---
Follow the testing guidelines set forth in #188 and refactor the [e2e package](https://github.com/operator-framework/operator-controller/tree/main/test/e2e) unit tests.

### Pre-Conditions
---
- [ ] #188 

### Acceptance Criteria
---
- [ ] All test files strictly under the resolution package updated to conform to the guidelines in #188

Issue #188, opened 2023-05-09
Document unit testing conventions


perdasilva:
### Discussed in https://github.com/operator-framework/operator-controller/discussions/168

<div type='discussions-op-text'>

<sup>Originally posted by **perdasilva** April 14, 2023</sup>
A [recent discussion](https://github.com/operator-framework/operator-controller/pull/142#discussion_r1165747298) in #142 raised a couple of questions around what our unit testing conventions ought to be:
 - should we still use ginkgo for unit tests or stick with gotest?
 - should we default to tabular tests, singular tests, or are we happy with either?

For the sake of consistency, I'm putting this out there to our community so we can establish some best practices and conventions. IMHO:
 - I think we should rollback on ginkgo for unit testing. Adds additional files (*_suite.go), it's not as nicely integrated with my IDE (or I haven't found a way to have as nice an experience as I do with gotests).
 - I personally think there is a trade-off between tabular and singular tests, and I'm of the camp that thinks that for unit tests readability > conciseness (and I personally have a hard time reading tabular tests as there is a physical distance between the input and the logic). Having said that, I'm happy to default to what out community decides. Ultimately, consistency is a better goal in my mind.</div>

Acceptance Criteria:
[] document describing development conventions in `/docs` created
[] development conventions doc covers our (unit) testing conventions:
  * No ginkgo for unit tests - use go test
  * Use tabular tests when it makes sense - call out trade-offs
  * Minimize the use of shared helper functions as they can often go forgotten

For more info, see discussion https://github.com/operator-framework/operator-controller/discussions/168

PR #207, opened 2023-05-09
Removed Ginkgo in require_package_test.go

matthew-hagemann:
This PR:
Aims to align tests with guidelines in #188 
Let me know if I misunderstood any and I can make adjustments.

Closes:
#196

### DIFF
--- a/internal/resolution/variable_sources/required_package/required_package_test.go
+++ b/internal/resolution/variable_sources/required_package/required_package_test.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"testing"
 
-	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	"github.com/operator-framework/deppy/pkg/deppy"
 	"github.com/operator-framework/deppy/pkg/deppy/input"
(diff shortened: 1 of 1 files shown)


internal/resolution/variable_sources/required_package/required_package_test.go
```diff
@@ -15,90 +14,80 @@ import (
...
+			property.TypePackage: `{"packageName": "test-package", "version": "2.0.0"}`,
+			property.TypeChannel: `{"channelName":"stable","priority":0}`,
+		})),
+		olmentity.NewBundleEntity(input.NewEntity("bundle-3", map[string]string{
+			property.TypePackage: `{"packageName": "test-package", "version": "3.0.0"}`,
+			property.TypeChannel: `{"channelName":"stable","priority":0}`,
+		})),
+	}
+	rpv = required_package.NewRequiredPackageVariable(packageName, bundleEntities)
+	t.Run("ShouldReturnCorrectPackageName", func(t *testing.T) {
```
ncdc, line 40: I would suggest that you don't do a literal translation of `It()` to `t.Run()`. It's ok to have multiple assertions in a single block of testing code that aren't each in their own t.Run subtest.

perdasilva: +1 and if we could use testify for the assertions, even better ^^
@matthew-hagemann thanks so much for this!

ncdc:
@matthew-hagemann hi, do you think you'll have time to come back to this?

tmshort:
I believe this was resolved in #402.

Issue #205, opened 2023-05-09
Operators should be reconciled on any CatalogSources change


awgreene:
Background:

When reconciling operator CRs, the operator-controller builds a list of constraints based on operator CRs and attempts to solve those constraints by considering content available in catalogSources. If the content within a catalogSource changes, a new solution may be available for a set of constraints that had failed earlier.

The operator controller should be updated to watch catalogSources. On changes that affect the catalogContent, the operator controller should queue all operator CRs to be reconciled. For the purposes of this ticket, we will queue all operator CRs for any catalog event, even if content might not have changed.

Demo Script:

install OLM V1
Remove the default catalogSource
Create an Operator CR that exists in the catalog
Check that it fails
Recreate the default catalogSource
Check that the package defined in the operator CR is deployed

awgreene:
Demoed in the June 13th, 2023 Working Group and Issue Triage Meeting, viewable [here](https://youtu.be/j4j-5XRH8Kw).

PR #216, opened 2023-05-16
WIP: resolve operators on catalog availability change

ankitathomas:
This PR triggers resolution whenever the available catalogsources on cluster change. Resolution is triggered for all operators on cluster when:
- a new catalogsource becomes available
- a catalogsource is deleted
- the state of any catalogsource on cluster changes (ready/not ready)

See: https://github.com/operator-framework/operator-controller/issues/205

- [ ] API Go Documentation


controllers/catalog_predicates.go
```diff
@@ -0,0 +1,115 @@
...
+	if err != nil {
+		fmt.Println(err)
+		return false
+	}
+	c.catalogReady[e.Object.GetName()] = catalogReady
+	return catalogReady
+}
+
+func (c *catalogReadyTransitionPredicate) Update(e event.UpdateEvent) bool {
+	fmt.Println("UpdateEvent CatalogSource", e.ObjectOld.GetName(), e.ObjectNew.GetName())
```
awgreene, line 43: Names cannot be changed in an update event.

controllers/catalog_predicates.go
```diff
@@ -0,0 +1,115 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+// Predicate for reconciling operators when available (i.e. ready) catalogsources on cluster change
+type catalogReadyTransitionPredicate struct {
+	predicate.Funcs
+	catalogReady map[string]bool
```
awgreene, line 22: I'm hesitant to include a cache of catalog statuses, could you explain why this is needed?

controllers/catalog_predicates.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+}
+
+func (c *catalogReadyTransitionPredicate) Create(e event.CreateEvent) bool {
+	fmt.Println("CreateEvent CatalogSource", e.Object.GetName())
+	catalogReady, err := isCatalogReady(e.Object)
+	if err != nil {
+		fmt.Println(err)
+		return false
+	}
```
joelanford, line 37: If the catalog is not ready, is that something we would want to reflect in the status of the Operator (either now, or in the future)?

I _think_ I'm of the opinion that we should just blindly forward along all CatalogSource events and deal with the state of it during reconciliation where we have a change to touch the Operator status if necessary.

WDYT?

controllers/catalog_predicates.go
```diff
@@ -0,0 +1,115 @@
...
+			if cond.Type == catalogd.TypeReady && cond.Status == v1.ConditionTrue {
+				return true, nil
+			}
+		}
+	}
+	return false, nil
+}
+
+// Generate reconcile requests for all operators affected by a catalog change
+func operatorRequestsForCatalog(ctx context.Context, c client.Client, logger logr.Logger) handler.MapFunc {
```
joelanford, line 95: Since we only need to read, it would make this function a little more flexible to require only a `client.Reader` instead of a `client.Client` (which must implement read and write methods).
```suggestion
func operatorRequestsForCatalog(ctx context.Context, c client.Reader, logger logr.Logger) handler.MapFunc {
```

controllers/catalog_predicates.go
ankitathomas: Depending on whether we have a caching layer for catalog contents at the operator controller level, we may not want to reflect a non-ready state for catalogs at the operator-controller level. I don't think it makes sense to run the resolver to trigger on any catalog change especially if the catalog is not ready, but I agree that we can have that check moved further down the resolution process. We'd need to verify that none of our sources are stale/non-ready to avoid resolving unnecessarily.

controllers/catalog_predicates.go
ankitathomas: I added this as while I was testing for generic events on a cluster, this isn't a necessary piece of the code. Thanks for pointing that out, I'll clean that up.

test/e2e/install_test.go
```diff
@@ -93,7 +93,75 @@ var _ = Describe("Operator Install", func() {
...
+				},
+				Spec: catalogd.CatalogSourceSpec{
+					// (TODO): Set up a local image registry, and build and store a test catalog in it
+					// to use in the test suite
+					Image: "quay.io/operatorhubio/catalog:latest",
+				},
+			}
+		})
+
+		It("resolves again when a new catalog is available", func() {
```
joelanford, line 129: I wonder if we can avoid all the long timeouts here by simplifying the test expectations. I think all we really want to assert is "whenever a catalog is created/updated/deleted, all of the existing operators are reconciled". Is there something that would show up in the `Operator` CR that we could use to verify that a reconcile happened after the catalog event?

That might be organized something like this:
```
Describe("reconcile Operators on catalog events")
   - BeforeEach() -> create 3 operators
   - AfterEach() -> delete 3 operators
   - When("catalog is created")
        - It("reconciles 3 operators")
   - When("catalog exists")
       - BeforeEach() -> create a catalog
       - AfterEach() -> delete the catalog
       - When("catalog is updated")
           - It("reconciles 3 operators")
       - When("catalog is deleted")
           - It("reconciles 3 operators")
```

joelanford: Or maybe even this is too much to do in the e2e? Is there a way to test the controller and the watches without a full-blown e2e?

If so, we could verify all these scenarios there, and then just focus on the happy path in the e2e?

I'm honestly not sure what's the best option here, but I think we should _really_ focus on keeping our overall e2e run as short (in terms of overall run time) as possible.

everettraven: I agree that it would be ideal if we could test this outside of the e2e tests, but as far as I am aware there isn't a good way to test watches being triggered without actually starting the controller and performing operations against a cluster that would trigger those watches. I think it *could* be done with envtest but would likely involve a lot of stubbing out resources and would essentially boil down to the question of "do we prefer complex unit tests or longer running e2e tests?"

controllers/operator_controller.go
```diff
@@ -400,3 +409,26 @@ func mapBDStatusToReadyCondition(existingBD *rukpakv1alpha1.BundleDeployment, ob
 func isBundleDepStale(bd *rukpakv1alpha1.BundleDeployment) bool {
 	return bd != nil && bd.Status.ObservedGeneration != bd.GetGeneration()
 }
+
+// Generate reconcile requests for all operators affected by a catalog change
+func operatorRequestsForCatalog(ctx context.Context, c client.Reader, logger logr.Logger) handler.MapFunc {
+	return func(object client.Object) []reconcile.Request {
+		// no way of associating an operator to a catalog so create reconcile requests for everything
```
everettraven, line 416: If we included the catalog in an `Operator.Status` field could this be filtered that way via a field selector? Probably out of scope for this work, but I definitely think it would be ideal in the future to have some way of saying `Catalog` X was changed so only re-reconcile `Operator` resources that were resolved to a package in `Catalog` X

controllers/operator_controller.go
```diff
@@ -20,6 +20,9 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/go-logr/logr"
+	operatorv1 "github.com/operator-framework/api/pkg/operators/v1"
```
tmshort, line 24: Why are we pulling in the `Operator` from `operator-framework/api` (OLMv0) rather than using the one included in this repo (OLMv1)? (i.e. `operatorsv1alpha1` below?) Are we conflating the two Operator CRDs?

controllers/operator_controller.go
```diff
@@ -400,3 +409,26 @@ func mapBDStatusToReadyCondition(existingBD *rukpakv1alpha1.BundleDeployment, ob
 func isBundleDepStale(bd *rukpakv1alpha1.BundleDeployment) bool {
 	return bd != nil && bd.Status.ObservedGeneration != bd.GetGeneration()
 }
+
+// Generate reconcile requests for all operators affected by a catalog change
+func operatorRequestsForCatalog(ctx context.Context, c client.Reader, logger logr.Logger) handler.MapFunc {
+	return func(object client.Object) []reconcile.Request {
+		// no way of associating an operator to a catalog so create reconcile requests for everything
+		operators := operatorv1.OperatorList{}
```
tmshort, line 417: shouldn't this be `operatorsv1alpha1`?

controllers/operator_controller.go
ankitathomas: you're right, this should be `operatorsv1alpha1`, thanks!

test/e2e/install_test.go
ankitathomas: Using a minimal catalog cuts the wait down by a good chunk, so the e2e suite shouldn't be as long lived. I'm not sure if we'd want to expose resolution/reconcile details on the operator CR just for the sake of CI, in case it can be confusing to a user.

controllers/operator_controller.go
ankitathomas: It'd be useful to have this field but that excludes the possibility of cross catalog updates. I can see this happening if a catalog an operator is installed from is deleted and reinstalled, we'd still need to trigger reconcile to catch any upgrades.

test/e2e/install_test.go
tmshort: If we're concerned about adding to the status API, how about examining the logs. We control the logs, and the tests can examine them for particular log entries. It wouldn't necessarily be exposing the logs as an API since it would be for e2e/unit tests, which can be modified when the logs are.

controllers/operator_controller.go
everettraven: >excludes the possibility of cross catalog updates

So with this current implementation we are saying "if there is an update in any catalog we will update the operator"? Are we sure that is something we want to do without some form of explicit action from the user side? What if there is a catalog with a package with the same name in another catalog that is actually a completely different package?

These questions are coming from a lack of knowledge on how operator-controller handles that case today and if there have already been discussions around this so apologies in advance if I am rehashing something

joelanford: > if there is an update in any catalog we will update the operator?

No, we're saying we'll resolve, evaluate constraints, and apply the results of resolution. There's definitely more work we need to do to provide sufficient knobs on the Operator API to help admins avoid that sort of problem, but that's a problem for resolution to solve, so IMO, we shouldn't preempt that right now by not even giving the resolver a chance to take a look.

test/e2e/install_test.go
joelanford: The changes in this PR boil down to an updated controller configuration with a new watch that adds some extra stuff to the workqueue and calls a reconciler. So I'm suggesting it might be possible to isolate that specific behavior if we refactored a little bit (e.g. what if instead of `SetupWithManager` being a method on the reconciler, we made it a function that we could pass a reconciler to). In our `main.go`, we'd pass the real reconciler, but here, we'd pass a fake reconciler that does the "did I get triggered" assertions

Then we could put some envtest-based unit tests together alongside the reconciler and we would essentially split the testing into two chunks:
1. Does the controller configuration enqueue the expected reconcile requests?
2. Does the reconciler handle reconcile requests in the expected way?

We'd still want an e2e that runs through some scenarios of the integrated/real main.go, but we would avoid the combinatoric problem here, which would save _lots_ of time in e2e runs.

joelanford: > how about examining the logs

Yeah +1 if the unit test path isn't taken. We do this in the SDK helm-operator e2e tests.

controllers/operator_controller.go
everettraven: Sounds good. Thanks for the clarification!

tmshort (changes requested):
I think this generally looks good, but would like to see @joelanford's last comment addressed, whether any changes are made or not.

tmshort (approved):

tmshort:
Not sure how I clicked "requested changes" instead of "approve"!

test/e2e/install_test.go
ankitathomas: I'm opening another issue to track the refactoring https://github.com/operator-framework/operator-controller/issues/247. I don't think it makes sense to block this PR on the refactor, I'm in favor of having that in a separate PR.

PR #218, opened 2023-05-18
Operator CR Installed status condition

### LINKED ISSUE #199: Add status.installedBundleSource
Introduce the status.installedBundleSource which should reflect the bundle image sha/git commit which can be derived from [this structure](https://github.com/operator-framework/rukpak/blob/0876db6a30de0aaa6ce5651e2e24f5f6d5766e22/api/v1alpha1/bundle_types.go#L61-L78) in the RukPak bundle status.

A corresponding condition should exist which represents whether or not the operator is installed:
```
 Status:
  conditions:
    - type: "Installed"
       status: "False"
       reason: "PrettyBigBad"
       message: "The package does not exist"
       observedGeneration: 2
```

jmprusi:
# Description

This PR extends the operator CR with a new condition: "Installed" which exposes the bundle deployment status at the operator CR level.

It also adds a new field: "installedBundleResource" that contains the installed image/git ref.

Closes #199 

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/api/v1alpha1/operator_types.go
+++ b/api/v1alpha1/operator_types.go
@@ -81,7 +81,7 @@ func init() {
 // OperatorStatus defines the observed state of Operator
 type OperatorStatus struct {
 	// +optional
-	InstalledBundleSource string `json:"installedBundleSource,omitempty"`
+	InstalledBundleResource string `json:"installedBundleResource,omitempty"`
 	// +optional
 	ResolvedBundleResource string `json:"resolvedBundleResource,omitempty"`
 
--- a/api/v1alpha1/operator_types.go
+++ b/api/v1alpha1/operator_types.go
@@ -46,7 +46,8 @@ type OperatorSpec struct {
 
 const (
 	// TODO(user): add more Types, here and into init()
-	TypeResolved = "Resolved"
+	TypeInstalled = "Installed"
+	TypeResolved  = "Resolved"
 
 	ReasonBundleLookupFailed        = "BundleLookupFailed"
 	ReasonInstallationFailed        = "InstallationFailed"
@@ -61,6 +62,7 @@ const (
 func init() {
 	// TODO(user): add Types from above
 	operatorutil.ConditionTypes = append(operatorutil.ConditionTypes,
+		TypeInstalled,
 		TypeResolved,
 	)
 	// TODO(user): add Reasons from above
@@ -78,6 +80,8 @@ func init() {
 
 // OperatorStatus defines the observed state of Operator
 type OperatorStatus struct {
+	// +optional
+	InstalledBundleSource string `json:"installedBundleSource,omitempty"`
 	// +optional
 	ResolvedBundleResource string `json:"resolvedBundleResource,omitempty"`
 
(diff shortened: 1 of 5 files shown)


api/v1alpha1/operator_types.go
```diff
@@ -78,6 +80,8 @@ func init() {
 
 // OperatorStatus defines the observed state of Operator
 type OperatorStatus struct {
+	// +optional
+	InstalledBundleSource string `json:"installedBundleSource,omitempty"`
```
everettraven, line 84: Nit: I'm assuming this has a relation to the field below in that they both point to `Bundle` resources on cluster (please correct me if I am wrong). To keep the naming similar I would prefer the name `InstalledBundleResource`

controllers/operator_controller.go
```diff
@@ -176,20 +210,93 @@ func (r *OperatorReconciler) reconcile(ctx context.Context, op *operatorsv1alpha
...
+			ObservedGeneration: op.GetGeneration(),
+		}, bundleDeploymentSource.Git.Repository + "@" + bundleDeploymentSource.Git.Ref.Commit
+	default:
+		return metav1.Condition{
+			Type:               operatorsv1alpha1.TypeInstalled,
+			Status:             metav1.ConditionUnknown,
+			Reason:             operatorsv1alpha1.ReasonInstallationStatusUnknown,
+			Message:            fmt.Sprintf("unknown bundledeployment source type %q", bundleDeploymentSource.Type),
+			ObservedGeneration: op.GetGeneration(),
+		}, ""
```
everettraven, line 296: Nit: 

Coming into this a bit unfamiliar with the actual code base, but from what I can tell this is the 7th use of creating this condition with only the `Condition.Message` field being different. It might be nice to create a function like:
```go
func installStatusUnknownCondition(msg string, generation int64) metav1.Condition {
  return metav1.Condition{
 	Type:               operatorsv1alpha1.TypeInstalled,
 	Status:             metav1.ConditionUnknown,
 	Reason:             operatorsv1alpha1.ReasonInstallationStatusUnknown,
 	Message:            msg,
 	ObservedGeneration: generation,
 }
}
```

I don't know if this is a followed pattern in the code base, but IMO this causes there to be a bit less bloat to read through when looking at the code and it will make it so you only have to change a few lines if in the future we wanted to tweak the values for this condition at all

everettraven (approved):
Overall looks good to me, just some nits:

api/v1alpha1/operator_types.go
everettraven: Alternatively, changing the field `ResolvedBundleResource` --> `ResolvedBundleSource` would also work

jmprusi: Makes sense! Thanks! Let me change it to `InstalledBundleResource`.

controllers/operator_controller.go
jmprusi: I don't have a strong opinion on this. Still, after trying the code change, I'm unsure about only creating a helper for the InstallStatus Unknown condition, as the code looks "weird/harder to read" having some conditions declared differently.

How is this usually solved in other codebases?

api/v1alpha1/operator_types.go
jmprusi: fixed

controllers/operator_controller.go
everettraven: > having some conditions declared differently.
I think this is the biggest thing about *only* doing this for the one condition. IIRC both catalogd and rukpak use helper functions for *all* the status conditions, so I'd be fine with this being a follow up item to clean up the code for all the status conditions

tmshort (approved):

controllers/operator_controller.go
jmprusi: @everettraven makes sense; I will start a PR to cleanup this code by creating the proper helpers for all the conditions.

PR #221, opened 2023-05-18
Makefile: don't globally export GOOS=linux

### LINKED ISSUE #220: 'go run' uses GOOS=linux for local tools, even on non-linux platforms
https://github.com/operator-framework/operator-controller/blob/4ea83e6fbe666af5023d4e43f34110ac2cd27e24/Makefile#L115
makes GOOS 'sticky' for follow-on operations, and since it is the first dependency for the 'run' target, it /might/ be run first and carry that value over to other targets.

joelanford:
# Description
Fixes #220

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -96,10 +96,18 @@ kind-cluster-cleanup: kind ## Delete the kind cluster
 
 ##@ Build
 
+BUILDCMD = sh -c 'mkdir -p $(BUILDBIN) && ${GORELEASER} build ${GORELEASER_ARGS} --single-target -o $(BUILDBIN)/manager'
+BUILDDEPS = manifests generate fmt vet goreleaser
+
 .PHONY: build
-build: manifests generate fmt vet goreleaser ## Build manager binary using goreleaser for current GOOS and GOARCH.
-	mkdir -p bin
-	${GORELEASER} build ${GORELEASER_ARGS} --single-target -o bin/manager
+build: BUILDBIN = bin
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -112,26 +120,8 @@ wait:
...
-.PHONY: docker-buildx
-docker-buildx: test ## Build and push docker image for the manager for cross-platform support
-	# copy existing Dockerfile and insert --platform=${BUILDPLATFORM} into Dockerfile.cross, and preserve the original Dockerfile
-	sed -e '1 s/\(^FROM\)/FROM --platform=\$$\{BUILDPLATFORM\}/; t' -e ' 1,// s//FROM --platform=\$$\{BUILDPLATFORM\}/' Dockerfile > Dockerfile.cross
-	- docker buildx create --name project-v3-builder
-	docker buildx use project-v3-builder
-	- docker buildx build --push --platform=$(PLATFORMS) --tag ${IMG} -f Dockerfile.cross .
-	- docker buildx rm project-v3-builder
-	rm Dockerfile.cross
+docker-build: build-linux ## Build docker image with the operator-controller.
```
tmshort, line 123: Should the comment here indicate GOOS/GOARCH behavior?

tmshort (approved):
Looks good, with one minor question.

Makefile
joelanford: Yeah, good idea.

joelanford: Updated.

tmshort:
Still looks good, waiting for CI

Issue #206, opened 2023-05-09
Dependency Library + CLI


awgreene:
We want to be able to perform deppy resolution off cluster. The Deppy library leaves it to users to define the EntitySource/VariableSources, so it wouldn't make sense to couple the CLI with the Deppy library. Instead, this ticket proposes that we create a new library that:
- Handles the implementation details of OLM V1's EntitySource/VariableSource implementation.
- Defines an input API that is OLM V1 specific:
-- FBC Content
-- Operator CR Constraints
-- Cluster Properties
- Introduces a CLI that allows off cluster resolution

A stretch goal could be to provide a KUBECONFIG and generate a file that can be used as an input API to the CLI.

m1kola:
We have a proof of concept in this PR: https://github.com/operator-framework/operator-controller/pull/246

The next steps here:
* Add a message into the CLI tool saying something along the lines: "This is the proof of concept of off-cluster resolution and will likely change. We are gathering community feedback here: [link to GitHub discussion]"
* Get it reviewed and merged
* Gather more feedback from the community using this GitHub discussion - https://github.com/operator-framework/operator-controller/discussions/262
* Plan next iterations based on feedback

Also there are a few things going in parallel. For example, we will likely change significantly how we build resolution problems (see https://github.com/operator-framework/deppy/issues/96 and [relevant RFC](https://docs.google.com/document/d/1PizF1uQcMrrAFsLgcx2O0Coq7mei4xe9hxpz-QRp-x0/edit#heading=h.x3tfh25grvnv)). Once we have more or less stable way to build the resolution problem - we will be able to create some sort of library to share between operator-controller and resolution CLI.

m1kola:
#246 got merged. We have PoC CLI which is:
* Capable of resolving off-cluster (doesn't require a cluster at all)
* Takes package name, channel (optional) and version (optional) as input for resulution
* It also take either a local directory with FBC or an image ref with FBC catalog (just one for now)
* Optionally it also takes a directory with yaml manifests. If manifests contain `Operator` objects, they will be used as input into the solver (list of required packages - similar to Operator objects installed in the cluster when running on-cluster resoltion).

Next steps:
* Gather feedback from the community using this GitHub discussion - https://github.com/operator-framework/operator-controller/discussions/262
* Plan next iterations based on feedback


[...]

PR #222, opened 2023-05-19
Moves `./controllers` into `./internal` and `./main.go` into `./cmd`

m1kola:
# Description

* Moves `./controllers` into `./internal`
  Unless we have intention to export `./controllers` as a library - it makes sense to move this package into `./internal`. If we want to expose `./controllers` at some point - I would suggest moving it into `./pkg`.
* Moves `./main.go` into `./cmd`
  We want to add more commands into the repo (#206, for example) so would be great to avoid having one of the entry points in the repo root

**Note:** `go-apidiff` will complain since we no longer export stuff from `./controllers`.

## Reviewer Checklist

- [ ] API Go Documentation


tmshort:
I question moving `main.go`. As #206 could be done in the `deppy` repo (or even the `kubectl-operator` repo), and not necessarily in this repo.
I agree we probably don't need to export most (if any) of `operator-controller` _now_, but maybe in the future?

everettraven:
The actual move looks fine to me, but I agree with @tmshort that we may not need to export anything now, but we could in the future. I think this gets into the discussion of the different supported use cases and form factors that OLMv1 could come in and is probably a discussion much to large for this PR.

That being said, if we find in the future that some stuff needs to be exported we could always refactor it again

m1kola:
> I question moving `main.go`. As #206 could be done in the `deppy` repo (or even the `kubectl-operator` repo), and not necessarily in this repo.

@tmshort I would still suggest to move it to `cmd` as it is very command to place entry points in `cmd` in Go projects.


> agree we probably don't need to export most (if any) of `operator-controller` _now_, but maybe in the future?

> That being said, if we find in the future that some stuff needs to be exported we could always refactor it again

@tmshort @everettraven I think it is easier to export something than remove publicly available API. I'm in favour of starting with as few exported things as possible and export as the need arises.

everettraven:
>I'm in favour of starting with as few exported things as possible and export as the need arises.

@m1kola I think that is reasonable

everettraven (approved):

tmshort:
> > I'm in favour of starting with as few exported things as possible and export as the need arises.
> 
> @m1kola I think that is reasonable

According to Spock, "I was not attempting to evaluate its moral implications, Doctor. As a matter of cosmic history, it has always been easier to destroy than to create." but not in the case of APIs.

So, yeah.

PR #223, opened 2023-05-19
Removes `util` package

m1kola:
# Description

`util` is too generic for a package name.

Let's try and avoid packages like `utils`, `common`, etc as packages like this tend to become dumping grounds.

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/api/v1alpha1/operator_types.go
+++ b/api/v1alpha1/operator_types.go
@@ -19,7 +19,7 @@ package v1alpha1
 import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
-	operatorutil "github.com/operator-framework/operator-controller/internal/util"
+	"github.com/operator-framework/operator-controller/internal/conditionsets"
 )
 
 // OperatorSpec defines the desired state of Operator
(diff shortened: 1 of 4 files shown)


tmshort:
I'm going to whine about the length of the name (`conditionsets` vs `util`), perhaps abbreviate (within the code, not necessarily the directory name) to `cond`?

m1kola:
@tmshort previously it was imported as `operatorutil` which is 1 char longer than `conditionsets`. I would prefer `conditionsets` over `cond` as it is less ambiguous. But yeah...

> There are only two hard things in Computer Science: cache invalidation and naming things.

😄

m1kola:
@tmshort I was consindering calling it `cond` or `condition`, but this is not about condition itself. I don't know how to make it clear that this is not about condition but rather about a set/collection of condition types. And at the same time not introduce confusion with API package and other common abbreviations.

Overall - I don't like this thing. Especially the fact that we tightly coupled API (`api/v1alpha1/operator_types.go`) with an utility package which only used for tests.

tmshort (approved):

m1kola:
@tmshort rebased. Please take another look.

tmshort (approved):

PR #224, opened 2023-05-19
Add additional prometheus support

tmshort:
Kustomize the necessary role/rolebinding to work with kube-prometheus

# Description

Installs kube-promethus when certain variables are set during the build.
Update README.md with this information.

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,13 @@ export XDG_DATA_HOME ?= /tmp/.local/share
 # bingo manages consistent tooling versions for things like kind, kustomize, etc.
 include .bingo/Variables.mk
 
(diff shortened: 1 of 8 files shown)


joelanford:
This is really cool! I'm a big prometheus fan, but I have to be honest. This strikes me as something that will languish, potentially stop working, be a distraction, or all of the above. Not because this isn't well done, but because it isn't our bread and butter, and there doesn't seem to be a push from users for us to take this on and include it as a feature.

m1kola:
It looks like we are not making progress here. I'm closing it, but please don't hesitate to re-open it if we still intend to go forward with this change in.

PR #227, opened 2023-05-22
Update Makefile targets for consistency

### LINKED ISSUE #226: Makefile inconsistencies
While reviewing the makefile I noticed several potential issues:

1. The `install` target, which does not build anything depends on the `generate` target which invokes `controller-gen` to update `DeepCopy, DeepCopyInfo and `DeepCopyObject` methods. The `generate` command _should_ probably be a dependency of `build-linux` or `build`.

2. The `install` target does not allow for specifying any `IMG`, so it assumes the default image name. The `deploy` target  can be used to install 

3. The `uninstall` target removes CRDs, but the `install` target installs everything (and is used by `run`).

4. The `deploy` target installs the controller, the `undeploy` target removes everything. (This is OK.)

5. The `quickstart` target creates a manifest (operator-controller.yaml), but ignores it in favor of a released version.

tmshort:
# Description

Update Makefile targets for consistency

Fixes #226

* Rename the `install` target to `kind-deploy`
* Remove `uninstall` target
* Rename the `kind-cluster-cleanup` target to `kind-clean`
* Remove the `generate` target from `kind-deploy`
* Swap `e2e`/`test-e2e` targets
* Update help descriptions
* Add `help-extended`,
* Change e2e workflow to use `test-e2e`

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -1,0 +1,29 @@
+name: deploy-test
+
+on:
+  workflow_dispatch:
+  pull_request:
+  push:
+    branches:
+    - main
+
+jobs:
+  deploy-basic:
+    runs-on: ubuntu-latest
+    steps:
+
+    - uses: actions/checkout@v3
+
+    - uses: actions/setup-go@v4
+      with:
+        go-version-file: go.mod
+
+    - name: Run basic deploy/undeploy
+      run: |
+        make kind-cluster
+        make deploy
+        kubectl get crds operators.operators.operatorframework.io
+        kubectl get ns operator-controller-system
+        make undeploy
+        ! kubectl get ns operator-controller-system
+        ! kubectl get crds operators.operators.operatorframework.io
--- a/Makefile
+++ b/Makefile
@@ -43,102 +43,112 @@ SHELL = /usr/bin/env bash -o pipefail
 
 .DEFAULT_GOAL := build
 
-##@ General
+#SECTION General
 
 # The help target prints out all targets with their descriptions organized
-# beneath their categories. The categories are represented by '##@' and the
-# target descriptions by '##'. The awk commands is responsible for reading the
+# beneath their categories. The categories are represented by '#SECTION' and the
+# target descriptions by '#HELP' or '#EXHELP'. The awk commands is responsible for reading the
 # entire set of makefiles included in this invocation, looking for lines of the
-# file as xyz: ## something, and then pretty-format the target and help. Then,
-# if there's a line with ##@ something, that gets pretty-printed as a category.
+# file as xyz: #HELP something, and then pretty-format the target and help. Then,
+# if there's a line with #SECTION something, that gets pretty-printed as a category.
 # More info on the usage of ANSI control characters for terminal formatting:
 # https://en.wikipedia.org/wiki/ANSI_escape_code#SGR_parameters
 # More info on the awk command:
 # http://linuxcommand.org/lc3_adv_awk.php
+# The extended-help target uses '#EXHELP' as the delineator.
 
 .PHONY: help
-help: ## Display this help.
-	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[a-zA-Z_0-9-]+:.*?##/ { printf "  \033[36m%-15s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
+help: #HELP Display essential help.
+	@awk 'BEGIN {FS = ":[^#]*#HELP"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n\n"} /^[a-zA-Z_0-9-]+:.*#HELP / { printf "  \033[36m%-15s\033[0m %s\n", $$1, $$2 } ' $(MAKEFILE_LIST)
 
-##@ Development
+.PHONY: help-extended
+help-extended: #HELP Display extended help.
+	@awk 'BEGIN {FS = ":.*#(EX)?HELP"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[a-zA-Z_0-9-]+:.*#(EX)?HELP / { printf "  \033[36m%-25s\033[0m %s\n", $$1, $$2 } /^#SECTION / { printf "\n\033[1m%s\033[0m\n", substr($$0, 10) } ' $(MAKEFILE_LIST)
+
+#SECTION Development
 
 .PHONY: lint
-lint: $(GOLANGCI_LINT) ## Run golangci linter.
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -84,13 +84,18 @@ test-unit: envtest ## Run the unit tests
...
 	$(KIND) load docker-image $(IMG) --name $(KIND_CLUSTER_NAME)
 
 kind-cluster: kind kind-cluster-cleanup ## Standup a kind cluster
 	$(KIND) create cluster --name ${KIND_CLUSTER_NAME}
 	$(KIND) export kubeconfig --name ${KIND_CLUSTER_NAME}
 
+kind-deploy: export MANIFEST="./operator-controller.yaml"
+kind-deploy: kind manifests kustomize ## Install controller and dependencies onto the kind cluster
+	kubectl kustomize config/default > operator-controller.yaml
+	envsubst '$$CATALOGD_VERSION,$$CERT_MGR_VERSION,$$RUKPAK_VERSION,$$MANIFEST' < scripts/install.tpl.sh | bash -s
```
varshaprasad96, line 97: Just curious, if we have this, then do we need `deploy` target. I see that `make kind-deploy` would controller and the dependencies and the latter would only install the controller, but is there a use case where we would like to support just the installation of the controller with (probably) some other version of residual dependencies running on cluster. 

Also, thoughts on having the naming as `kind-deploy-e2e`? Just so that it's easier to interpret the difference between the two.

joelanford: How much of this is theoretical? I'd very much like to limit our Makefile to having as few supported use-cases as possible, for the sake of our own sanity.

I personally use (pretty much exclusively):
- `make build` to build the binaries locally.
- `make run` to rebuild/redeploy everything in a fresh kind cluster
- `make test-unit` (or whatever we call it) to run the unit tests
- `make test-e2e` to run the e2e test

Would it be worthwhile to do a quick contributor survey to see what people _actually_ use and then go from there?

tmshort: https://kubernetes.slack.com/archives/C0181L6JYQ2/p1685127976360479

.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,31 @@
...
+      run: |
+        make kind-cluster
+        make install
+        kubectl get crds operators.operators.operatorframework.io
+        make uninstall
+        ! kubectl get crds operators.operators.operatorframework.io
+        make deploy
+        kubectl get ns operator-controller-system
+        make undeploy
+        ! kubectl get ns operator-controller-system
```
joelanford, line 31: These steps essentially happen during the course of running the e2e, don't they? I suppose we don't try `uninstall` and `undeploy` during the e2e, but maybe we could?

Makefile
```diff
@@ -133,24 +146,26 @@ release: goreleaser ## Runs goreleaser for the operator-controller. By default,
 	$(GORELEASER) $(GORELEASER_ARGS)
 
 quickstart: export MANIFEST="https://github.com/operator-framework/operator-controller/releases/download/$(VERSION)/operator-controller.yaml"
-quickstart: kustomize generate ## Generate the installation release manifests and scripts
+quickstart: kustomize ### Generate the installation release manifests and scripts with dependencies
```
joelanford, line 149: Curious why we're removing `generate` as a dependency. Without it, the generated `operator-controller.yaml` may be out-dated.

joelanford: Nevermind, I see `generate` only generates go code. But perhaps `manifests` should be a dependency of `quickstart`?

Makefile
joelanford, line 1: Can't comment on the line since it is unchanged, but I'd propose moving `release` to extended help.

Makefile
```diff
@@ -133,24 +146,26 @@ release: goreleaser ## Runs goreleaser for the operator-controller. By default,
...
 ifndef ignore-not-found
   ignore-not-found = false
 endif
 
 .PHONY: install
-install: export MANIFEST="./operator-controller.yaml"
-install: manifests kustomize generate ## Install CRDs into the K8s cluster specified in ~/.kube/config.
-	kubectl kustomize config/default > operator-controller.yaml
-	envsubst '$$CATALOGD_VERSION,$$CERT_MGR_VERSION,$$RUKPAK_VERSION,$$MANIFEST' < scripts/install.tpl.sh | bash -s
+install: manifests kustomize ### Install CRDs into the K8s cluster specified in ~/.kube/config.
```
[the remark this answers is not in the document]
tmshort: `install` just does the CRDs... which aren't of much use without the controller. There are some processes that install CRDs first, then everything else... but operator-controller kinda needs to be first anyways.

.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,29 @@
...
+    runs-on: ubuntu-latest
+    steps:
+
+    - uses: actions/checkout@v3
+
+    - uses: actions/setup-go@v4
+      with:
+        go-version-file: go.mod
+
+    - name: Run basic unit tests
```
everettraven, line 21: calling this a "unit" test would be confusing to me if  all I saw was the name of the step and not what was actually happening

everettraven (approved):
A nit in the action but other than that the changes look reasonable to me

Makefile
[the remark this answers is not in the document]
joelanford: > There are some processes that install CRDs first

Our own processes, or theoretical third party processes? If the latter, and only the latter, I'm still of the opinion that we stick with _just_ the deploy/undeploy targets.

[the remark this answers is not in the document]
tmshort: Third-party; it's how we tried to service-mesh at my last job.

.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,29 @@
...
+      run: |
+        make kind-cluster
+        make deploy
+        kubectl get crds operators.operators.operatorframework.io
+        kubectl get ns operator-controller-system
+        make undeploy
+        ! kubectl get ns operator-controller-system
+        ! kubectl get crds operators.operators.operatorframework.io
```
joelanford, line 29: Two things:
1. nit: it looks like there's a mix of spaces and tabs here for indentation (unless GH UI is tricking me). Change tabs to spaces (only because that's how the rest of the file looks, not trying to start a holy war 😛 )?
2. My last comment I think was eaten by a new commit, repeating here: These steps essentially happen during the course of running the e2e, don't they? I suppose we don't try uninstall and undeploy during the e2e, but maybe we could?

tmshort: Since removing `install`/`uninstall`, I put in `undeploy`

tmshort: There are no tabs... so GH UI must be tricking you?

grokspawn:
I embrace the thesis of this PR, but IMHO the difference between two-hashes and three-hashes is too subtle to avoid confusion.

tmshort:
> I believe the file separator for awk is arbitrary. Could we change to something like `#extended` or something which more intuitively indicates the distinction from `##`?
```
#HELP
#EXHELP
#SECTION

grokspawn (approved):

Issue #215, opened 2023-05-16
e2e flake: [FAILED] Operator Install when An operator is installed from an operator catalog


joelanford:
See: https://github.com/operator-framework/operator-controller/actions/runs/4990417752/jobs/8938348906#step:4:522
```
• [FAILED] [181.699 seconds]
Operator Install when An operator is installed from an operator catalog [It] resolves the specified package with correct bundle path
/home/runner/work/operator-controller/operator-controller/test/e2e/install_test.go:64

  Timeline >>
  STEP: creating the Operator resource @ 05/16/23 12:15:10.141
  STEP: eventually reporting a successful resolution and bundle path @ 05/16/23 12:15:10.145
  [FAILED] in [It] - /home/runner/work/operator-controller/operator-controller/test/e2e/install_test.go:75 @ 05/16/23 12:15:40.147
  << Timeline

  [FAILED] Timed out after 30.001s.
  The function passed to Eventually failed at /home/runner/work/operator-controller/operator-controller/test/e2e/install_test.go:74 with:
  Expected
      <string>: package 'argocd-operator' not found
  to equal
      <string>: install was successful
  In [It] at: /home/runner/work/operator-controller/operator-controller/test/e2e/install_test.go:75 @ 05/16/23 12:15:40.147
------------------------------

Summarizing 1 Failure:
  [FAIL] Operator Install when An operator is installed from an operator catalog [It] resolves the specified package with correct bundle path
  /home/runner/work/operator-controller/operator-controller/test/e2e/install_test.go:75
```

tmshort:
Things to look at (from @joelanford):

- The image is pulled from a public image registry over the internet (is quay up? are we being rate-limited?)
- The image is big (is it taking too long to download/unpack?)
- The image is always changing and we don’t control the content of it (did the thing we depend on change?)

joelanford:
> Put the image contents + image build in the repo (and then use kind load to get it into the e2e cluster)

Doing this part requires a catalogd release that includes https://github.com/operator-framework/catalogd/pull/65

PR #233, opened 2023-05-25
Download external manifests and kind-load containers for e2e

jmprusi:
# Description

This PR adds new steps before running e2e tests, download the manifests of external dependencies, looks for container images, pulling them and preloading them into the e2e kind server.

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -110,7 +110,15 @@ build-linux: $(BUILDDEPS) ## Build manager binary using goreleaser for GOOS=linu
 	GOOS=linux $(BUILDCMD)
 
 .PHONY: run
-run: docker-build kind-cluster kind-load install ## Build the operator-controller then deploy it into a new kind cluster.
+run: docker-build kind-cluster kind-load ensure-external-manifests ensure-external-images-kindload install ## Build the operator-controller then deploy it into a new kind cluster.
+
+.PHONY: ensure-external-manifests ## Ensure that the external manifests are present in the manifests directory.
+ensure-external-manifests:
+	./scripts/pull-manifests.sh
+
+.PHONY: ensure-external-images-kindload
+ensure-external-images-kindload: ensure-external-manifests ## Ensure that the external images are loaded into the kind cluster.
+	./scripts/kind-load-images.sh
 
 .PHONY: wait
 wait:
@@ -133,7 +141,7 @@ release: goreleaser ## Runs goreleaser for the operator-controller. By default, 
 	$(GORELEASER) $(GORELEASER_ARGS)
 
 quickstart: export MANIFEST="https://github.com/operator-framework/operator-controller/releases/download/$(VERSION)/operator-controller.yaml"
-quickstart: kustomize generate ## Generate the installation release manifests and scripts
+quickstart: kustomize generate ensure-external-manifests ## Generate the installation release manifests and scripts
 	kubectl kustomize config/default | sed "s/:devel/:$(VERSION)/g" > operator-controller.yaml
 	envsubst '$$CATALOGD_VERSION,$$CERT_MGR_VERSION,$$RUKPAK_VERSION,$$MANIFEST' < scripts/install.tpl.sh > install.sh
 
--- a/Makefile
+++ b/Makefile
@@ -72,6 +72,10 @@ fmt: ## Run go fmt against code.
 vet: ## Run go vet against code.
 	go vet ./...
 
+.PHONY: verify
+verify: fmt vet manifests generate ensure-external-manifests ## Run all verification steps.
+	git diff --exit-code
+
 .PHONY: test test-e2e e2e kind-load kind-cluster kind-cluster-cleanup
 test: manifests generate fmt vet test-unit e2e ## Run all tests.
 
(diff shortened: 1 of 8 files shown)


scripts/pull-manifests.sh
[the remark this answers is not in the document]
tmshort: Under normal circumstances, since the versions are specified, nothing will happen (the manifests are already downloaded).
Perhaps a README.md file in `testdata/manifests` indicating how to update the manifests:
e.g. 
> How to update the test manifests:
> 1. Start in the root of the repository
> 2. Update versions in the `Makefile`
> 3. Run `make ensure-external-manifests`
> 4. Run `git add .`
> 5. Run `git commit -s`

This will certainly lead me to update my prometheus PR, as I download the git repo to install.

Makefile
```diff
@@ -110,7 +114,14 @@ build-linux: $(BUILDDEPS) ## Build manager binary using goreleaser for GOOS=linu
...
 
 .PHONY: run
-run: docker-build kind-cluster kind-load install ## Build the operator-controller then deploy it into a new kind cluster.
+run: docker-build kind-cluster kind-load ensure-external-manifests ensure-external-images-kindload install ## Build the operator-controller then deploy it into a new kind cluster.
+
+.PHONY: ensure-external-manifests ## Ensure that the external manifests are present in the manifests directory.
+ensure-external-manifests:
+	./scripts/pull-manifests.sh
+
+ensure-external-images-kindload: ensure-external-manifests ## Ensure that the external images are loaded into the kind cluster.
```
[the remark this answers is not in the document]
jmprusi: done!

Makefile
```diff
@@ -5,10 +5,14 @@
 export IMAGE_REPO ?= quay.io/operator-framework/operator-controller
 export IMAGE_TAG ?= devel
 export GO_BUILD_TAGS ?= upstream
+
+# If any of these versions are modified, remember to run `make ensure-external-manifests` to update
```
joelanford, line 9: Do we have a target like `make verify` that runs things like generators, formaters, etc. and then runs `git diff -e` to check that nothing changed?

If so, it seems like we should add `ensure-external-manifests` to the list?

If not, capture a follow-up to add a `verify` target and run it in CI?

scripts/pull-manifests.sh
```diff
@@ -0,0 +1,39 @@
...
+shopt -s extglob
+
+catalogd_version=$CATALOGD_VERSION
+cert_mgr_version=$CERT_MGR_VERSION
+rukpak_version=$RUKPAK_VERSION
+manifestsToClean=()
+
+# Let's check if the manifests for cert-manager, rukpak, and catalogd are available in the local filesystem
+# If they are, we'll use those instead of pulling from GitHub
+if [[ -f "testdata/manifests/cert-manager-${cert_mgr_version}.yaml" ]]; then
```
joelanford, line 13: Is there a better landing spot for these than `testdata`? What about `third_party` or `external` or `dependencies` or something like that? Or maybe something under `hack`, sort of as a sibling of `hack/tools`?

scripts/install.tpl.sh
```diff
@@ -36,16 +36,16 @@ function kubectl_wait() {
     kubectl wait --for=condition=Available --namespace="${namespace}" "${runtime}" --timeout="${timeout}"
 }
 
-kubectl apply -f "https://github.com/cert-manager/cert-manager/releases/download/${cert_mgr_version}/cert-manager.yaml"
+kubectl apply -f testdata/manifests/cert-manager-${cert_mgr_version}.yaml
```
joelanford, line 39: I don't think this works because we re-use this script in the actual github release. So in that case, users installing operator controller will likely need to use the URL like we have it today.

joelanford:
This is doing two things, right:
-  it makes our `make run` and `make e2e` resilient to github being down because we've saved off the manifests in our repo (but I don't recall this ever being a real concern)
- it means the waits in our install script are no longer waiting for image pulls.

Are there other problems this solves?

If not, I wonder if we should really be thinking about is whether our wait could be a bit smarter. This seems like a whole bunch of machinery and stuff to maintain, so I just want to make sure we think this is the best solution, and that its worth it.

scripts/pull-manifests.sh
jmprusi: I was thinking on something like `contrib` or `third_party` https://github.com/operator-framework/operator-controller/pull/233#discussion_r1205726677

jmprusi:
> This is doing two things, right?
> 
> * it makes our `make run` and `make e2e` resilient to github being down because we've saved off the manifests in our repo (but I don't recall this ever being a real concern)
> * it means the waits in our install script are no longer waiting for image pulls.
> 
> Are there other problems this solves?

Just the preloading of docker images, if GitHub is down, game over as our CI runs on it. 

> 
> If not, I wonder if we should really be thinking about is whether our wait could be a bit smarter. This seems like a whole bunch of machinery and stuff to maintain, so I just want to make sure we think this is the best solution, and that its worth it.

I agree, started this to drive some discussion. Also, this PR is not useful enough as the e2e are failing... I think that first, we should improve the e2e tests logging (capture pods logs and save them into the test artifact)

tmshort:
Odd... `e2e-kind` still failing after 30 seconds?

tmshort:
This should be referencing #215, which references an updated `catalogd`

jmprusi:
We can drop this one. It doesn't fix the issue it's supposed to.

PR #242, opened 2023-05-31
add support for handling plain+v0 bundle types

### LINKED ISSUE #238: Support installation of different bundle formats
Support installation of different bundle formats in a single catalog.
https://github.com/operator-framework/operator-controller/blob/d6aa5e9133139674b240782ea39d7dea34f129f2/controllers/operator_controller.go#L257 

Proposed implementation:

*Lookup bundle's format via bundle property `olm.bundle.mediatype`
*Use a switch statement when choosing the Bundle + BD provisioner class names
      If `olm.bundle.mediatype` is not found/unset, default to registry+plain
      If `olm.bundle.mediatype` is "plain+v0", use plain+plain
      If `olm.bundle.mediatype` is "registry+v1", use registry+plain
      if unknown mediatype, fail

everettraven:
# Description
- Adds logic for differentiating between `plain+v0` and `registry+v1` bundle formats and configuring the proper `provisionerClassName` for the `Bundle` template in the `BundleDeployment.Spec.Template` field.
- Adds unit tests where necessary
- Adds e2e tests for installing a `plain+v0` bundle
- Adds a `plain+v0` bundle to the testdata catalog

- resolves #238 
## Reviewer Checklist

- [ ] API Go Documentation


internal/resolution/entitysources/catalogdsource.go
```diff
@@ -86,6 +87,8 @@ func getEntities(ctx context.Context, client client.Client) (input.EntityList, e
 				// this is already a json marshalled object, so it doesn't need to be marshalled
 				// like the other ones
 				props[property.TypePackage] = string(prop.Value)
+			case entity.PropertyBundleMediaType:
+				props[entity.PropertyBundleMediaType] = string(prop.Value)
```
joelanford, line 91: Unrelated, but we need to capture a follow-up here to make sure our FBC -> entity translation is not lossy with respect to property type. We should be forwarding all properties through.

I think we need to go back to figure out what deppy expects here. Properties can be duplicated (some are expected to be, some are not). In this case, we expect probably 0 or 1 mediatype properties. We expect exactly one package property. But then we allow 0-n for `olm.gvk`, `olm.gvk.required`, `olm.package.required`, etc.

internal/resolution/variable_sources/entity/bundle_entity.go
```diff
@@ -11,6 +11,14 @@ import (
...
 
 const PropertyBundlePath = "olm.bundle.path"
+const PropertyBundleMediaType = "olm.bundle.mediatype"
+
+type MediaType string
+
+const (
+	MediaTypePlain    = "plain+v0"
+	MediaTypeRegistry = "registry+v1"
+)
```
joelanford, line 21: This is likely not the right place to define these in the long run, but probably fine for now. Add a TODO comment about finding a more permanent home?

internal/controllers/operator_controller.go
```diff
@@ -357,6 +367,19 @@ func isBundleDepStale(bd *rukpakv1alpha1.BundleDeployment) bool {
 	return bd != nil && bd.Status.ObservedGeneration != bd.GetGeneration()
 }
 
+// mapBundleMediaTypeToBundleProvisioner maps an olm.bundle.mediatype property to a
+// rukpak bundle provisioner class name that is capable of unpacking the bundle type
+func mapBundleMediaTypeToBundleProvisioner(mediaType string) (string, error) {
```
joelanford, line 372: Maybe have this return the bundle and BD provisioner? That way, if we ever have something that needs a non-plain BD provisioner, it slots in really easily.

internal/controllers/operator_controller.go
```diff
@@ -357,6 +367,19 @@ func isBundleDepStale(bd *rukpakv1alpha1.BundleDeployment) bool {
...
+	switch mediaType {
+	case entity.MediaTypePlain:
+		return "core-rukpak-io-plain", nil
+	case entity.MediaTypeRegistry, "":
+		return "core-rukpak-io-registry", nil
```
joelanford, line 377: It's probably a good idea to make a comment about the empty string and explain why `""` means registry+plain.

everettraven: Comment added in https://github.com/operator-framework/operator-controller/pull/242/commits/012fa1475eb4819386fa2f00e0e98222a1f06a50

internal/controllers/operator_controller.go
```diff
@@ -245,12 +255,13 @@ func (r *OperatorReconciler) getBundleEntityFromSolution(solution *solver.Soluti
 	return nil, fmt.Errorf("entity for package %q not found in solution", packageName)
 }
 
-func (r *OperatorReconciler) generateExpectedBundleDeployment(o operatorsv1alpha1.Operator, bundlePath string) *unstructured.Unstructured {
+func (r *OperatorReconciler) generateExpectedBundleDeployment(o operatorsv1alpha1.Operator, bundlePath string, bundleProvisioner string) *unstructured.Unstructured {
 	// We use unstructured here to avoid problems of serializing default values when sending patches to the apiserver.
```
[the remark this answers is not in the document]
everettraven: Looks like it does here: https://github.com/operator-framework/operator-controller/blob/22afecf8a5c50a8d4c46cb6826392fb0ad39b74c/internal/controllers/operator_controller.go#L324 Although I could be misunderstanding this patch call as doing SSA and it isn't. This `generateExpectedBundleDeployment` is generating an `unstructured.Unstructured` to use for that patch request. I'd be curious to see if it is worth it to create applyconfigurations for the rukpak APIs to improve the creation of the patch information here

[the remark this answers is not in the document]
stevekuznetsov: > I'd be curious to see if it is worth it to create applyconfigurations for the rukpak APIs to improve the creation of the patch information here

We updated the generators upstream so you should get those free with newer codegen tools. I do think the apply you linked will be SSA - and the defaults worries from this block should be obviated.

testdata/bundles/registry-v1/prometheus-operator.v0.47.0/metadata/annotations.yaml
```diff
@@ -1,4 +1,5 @@
 annotations:
+  operators.operatorframework.io.bundle.channel.default.v1: beta
```
anik120, line 2: Why is this being added back? https://github.com/operator-framework/operator-controller/pull/250#discussion_r1214652172

internal/controllers/operator_controller_test.go
```diff
@@ -960,6 +960,108 @@ var _ = Describe("Operator Controller Test", func() {
...
+				bd := &rukpakv1alpha1.BundleDeployment{}
+				Expect(cl.Get(ctx, types.NamespacedName{Name: opKey.Name}, bd)).NotTo(HaveOccurred())
+				Expect(bd.Spec.ProvisionerClassName).To(Equal("core-rukpak-io-plain"))
+				Expect(bd.Spec.Template.Spec.ProvisionerClassName).To(Equal("core-rukpak-io-plain"))
+				Expect(bd.Spec.Template.Spec.Source.Type).To(Equal(rukpakv1alpha1.SourceTypeImage))
+				Expect(bd.Spec.Template.Spec.Source.Image).NotTo(BeNil())
+				Expect(bd.Spec.Template.Spec.Source.Image.Ref).To(Equal("quay.io/operatorhub/plain@sha256:plain"))
+			})
+		})
+		When("the operator specifies a package with a bade bundle mediatype", func() {
```
anik120, line 1018: nit
```suggestion
		When("the operator specifies a package with a bad bundle mediatype", func() {
```

test/e2e/install_test.go
```diff
@@ -27,103 +27,147 @@ const (
...
-							Ref: testCatalogRef,
-						},
-					},
-				},
-			}
-		})
-		It("resolves the specified package with correct bundle path", func() {
-			err := c.Create(ctx, operatorCatalog)
+			var err error
+			operatorCatalog, err = createTestCatalog(ctx, testCatalogName, testCatalogRef)
```
anik120, line 38: Why does the operator catalog need to be created before each spec?

anik120 (changes requested):

test/e2e/install_test.go
everettraven: You're right that it doesn't have to be created before each spec. I can move it to the `BeforeSuite()` process to create  the catalog before the test suite even runs and then clean it up in an `AfterSuite()`

everettraven: https://github.com/operator-framework/operator-controller/pull/242/commits/95771f9a2f470e3a21c4f309995ed3d00cdf6525 makes it so the catalog is created in `BeforeSuite()` and cleaned up in `AfterSuite()`

anik120 (approved):
/lgtm

internal/controllers/operator_controller.go
joelanford: Sounds good! Definitely non-blocking for this PR.

test/e2e/e2e_suite_test.go
```diff
@@ -42,4 +53,76 @@ var _ = BeforeSuite(func() {
...
+		cond := apimeta.FindStatusCondition(operatorCatalog.Status.Conditions, catalogd.TypeUnpacked)
+		g.Expect(cond).ToNot(BeNil())
+		g.Expect(cond.Status).To(Equal(metav1.ConditionTrue))
+		g.Expect(cond.Reason).To(Equal(catalogd.ReasonUnpackSuccessful))
+
+		// Ensure some packages exist before continuing so the
+		// operators don't get stuck in a bad state
+		pList := &catalogd.PackageList{}
+		err = c.List(ctx, pList)
+		g.Expect(err).ToNot(HaveOccurred())
```
joelanford, line 73: For anything that just returns an error, I find this is shorter/easier to parse:
```suggestion
		g.Expect(c.List(ctx, pList)).To(Succeed())
```

testdata/catalogs/test-catalog/catalog.yaml
```diff
@@ -0,0 +1,48 @@
...
+entries:
+  - name: plain.0.1.0
+---
+schema: olm.bundle
+name: plain.0.1.0
+package: plain
+image: localhost/testdata/bundles/plain-v0/plain:v0.1.0
+relatedImages:
+  - name: ""
+    image: ""
```
joelanford, line 41: We could remove this if we merge https://github.com/operator-framework/operator-controller/pull/259 first.

test/e2e/e2e_suite_test.go
everettraven: https://github.com/operator-framework/operator-controller/pull/242/commits/82499689cb2eac6191e184ddd8875be0972e4690

everettraven: Oops, missed some: https://github.com/operator-framework/operator-controller/pull/242/commits/7f4c1bff9d5e053576eddd29d57c063bd0c02b6f

internal/controllers/operator_controller_test.go
```diff
@@ -904,58 +904,160 @@ var _ = Describe("Operator Controller Test", func() {
...
-				Expect(cond.Status).To(Equal(metav1.ConditionUnknown))
-				Expect(cond.Reason).To(Equal(operatorsv1alpha1.ReasonInstallationStatusUnknown))
-				Expect(cond.Message).To(Equal("installation has not been attempted as spec is invalid"))
+				Expect(cond.Status).To(Equal(metav1.ConditionFalse))
+				Expect(cond.Reason).To(Equal(operatorsv1alpha1.ReasonInstallationFailed))
+				Expect(cond.Message).To(Equal("unknown bundle mediatype: badmedia+v1"))
 			})
 		})
+	})
+	When("an invalid semver is provided that bypasses the regex validation", func() {
```
joelanford, line 1009: This may have been missed in the merge. I moved this in #257.

joelanford (approved):

PR #246, opened 2023-06-01
PoC: Resolution CLI

m1kola:
# Description

This is a PoC for #206.

Right now the tool is capable of resolving a package without a cluster: it is possible to provide a FBC (dir or an image) into the tool as well as a directory containing manifests with `Operator` objects.

The resolution tool will use `Operator` from the input manifests and consider them required (as if they were installed on a cluster).

Once we merge this, the next steps will be:

* Gather feedback from the community using this GitHub discussion - https://github.com/operator-framework/operator-controller/discussions/262
* Plan next iterations based on feedback

Also there are a few things going in parallel. For example, we will likely change significantly how we build resolution problems (see https://github.com/operator-framework/deppy/issues/96 and [relevant RFC](https://docs.google.com/document/d/1PizF1uQcMrrAFsLgcx2O0Coq7mei4xe9hxpz-QRp-x0/edit#heading=h.x3tfh25grvnv)). Once we have more or less stable way to build the resolution problem - we will be able to create some sort of library to share between operator-controller and resolution CLI.

## Example usage:

```
# Resolve prometheus from catalog quay.io/operatorhubio/catalog:latest from default channel
$ go run ./cmd/resolutioncli/ -index-ref quay.io/operatorhubio/catalog:latest -package-name prometheus
quay.io/operatorhubio/prometheus@sha256:85decebfdfe993484834bd722d93fb44725b3a56cb4f9ebc4e2fd398bdcbad54

# Resolve prometheus from catalog dir ~/Downloads/index from default channel
$ go run ./cmd/resolutioncli/ -index-ref ~/Downloads/index -package-name prometheus
quay.io/operatorhubio/prometheus@sha256:85decebfdfe993484834bd722d93fb44725b3a56cb4f9ebc4e2fd398bdcbad54

# Resolve prometheus version 0.65.1 from default channel
$ go run ./cmd/resolutioncli/ -index-ref quay.io/operatorhubio/catalog:latest -package-name prometheus -package-version 0.65.1
quay.io/operatorhubio/prometheus@sha256:85decebfdfe993484834bd722d93fb44725b3a56cb4f9ebc4e2fd398bdcbad54

# Resolve prometheus version 0.47.0 from default channel
$ go run ./cmd/resolutioncli/ -index-ref quay.io/operatorhubio/catalog:latest -package-name prometheus -package-version 0.47.0
quay.io/operatorhubio/prometheus@sha256:47c659b352bbc4b75b64fbe0077d19afaea80cf691c62d803d44fdfc88dc9c6f

# Resolve prometheus version 0.47.0 from stable channel which doesn't exist (beta is default channel)
$ go run ./cmd/resolutioncli/ -index-ref quay.io/operatorhubio/catalog:latest -package-name prometheus -package-version 0.47.0 -package-channel stable
package 'prometheus' at version '0.47.0' in channel 'stable' not found
exit status 1

# Resolve prometheus version 0.47.0 from beta channel
$ go run ./cmd/resolutioncli/ -index-ref quay.io/operatorhubio/catalog:latest -package-name prometheus -package-version 0.65.1 -package-channel beta
quay.io/operatorhubio/prometheus@sha256:85decebfdfe993484834bd722d93fb44725b3a56cb4f9ebc4e2fd398bdcbad54

# Resolve prometheus from default channel and reading input manifests from ./operator-input/ dir
# Note: at the moment, resolution tool only takes Operator kind as an input and considers it to be
# a required package.
$ mkdir operator-input
$ cat <<EOT > operator-input/operator_v1alpha1.yaml
kind: Operator
metadata:
  name: operator-sample
spec:
  packageName: argocd-operator
EOT
$ go run ./cmd/resolutioncli/ -index-ref quay.io/operatorhubio/catalog:latest -package-name prometheus -input-dir ./operator-input/
quay.io/operatorhubio/prometheus@sha256:85decebfdfe993484834bd722d93fb44725b3a56cb4f9ebc4e2fd398bdcbad54

# Resolve prometheus from default channel and reading input manifests from ./operator-input/ dir
# and prometheus package already exists in the input (e.g. already installed on the cluster)
$ mkdir operator-input
$ cat <<EOT > operator-input/operator_v1alpha1.yaml
kind: Operator
metadata:
  name: operator-sample
spec:
  packageName: prometheus
  version: 0.47.0
EOT
$ go run ./cmd/resolutioncli/ -index-ref quay.io/operatorhubio/catalog:latest -package-name prometheus -input-dir ./operator-input/
duplicate identifier "required package prometheus" in input
exit status 1
```

## Reviewer Checklist

- [ ] API Go Documentation


cmd/resolutioncli/fbcsource.go
```diff
@@ -0,0 +1,156 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "offline-fbc-catalog",
+		},
+	}
+
+	bundleMetadatas := ToBundleMetadata(fbc, catalog)
+	packages, err := ToPackages(fbc, catalog)
+	if err != nil {
+		return nil, nil, err
+	}
```
m1kola, line 153: Should probably use `github.com/operator-framework/operator-registry/alpha/model` instead

cmd/resolutioncli/entity_source.go
```diff
@@ -0,0 +1,154 @@
...
+			return nil, err
+		}
+
+		es.entitiesCache = entities
+	}
+
+	return es.entitiesCache, nil
+}
+
+func modelToEntities(model model.Model) (input.EntityList, error) {
```
m1kola, line 119: This somehow needs to share what we have in here:

https://github.com/operator-framework/operator-controller/blob/098778ea18bd1ea501795b1a68c5e52dbdff2515/internal/resolution/entitysources/catalogdsource.go#L75-L120

So we don't have to maintain two translations in parallel.

cmd/resolutioncli/main.go
```diff
@@ -0,0 +1,184 @@
...
+	if packageName == "" {
+		return fmt.Errorf("missing required -%s flag", flagNamePackageName)
+	}
+
+	if indexRef == "" {
+		return fmt.Errorf("missing required -%s flag", flagNameIndexRef)
+	}
+
+	return nil
+}
```
everettraven, line 110: Nit: If these flags are required would it make more sense to make them positional arguments? IMO flags should be reserved for configurable options. 

I noticed we aren't using `spf13/cobra` for building the CLI commands, but it has some nice support for setting required number of arguments that we could use here and automatically generates a pretty nice help message that could be used to show the required positional arguments.

everettraven (approved):
Overall this looks good to me with the idea in mind that this still has a good bit of follow up work to do to get it production ready. I had a minor nit but nothing that should block this PR IMO. 

I'm not very well versed in the variable source / deppy stuff so I'd definitely prefer someone more familiar with that review it prior to merging in case I've missed something.

Great work @m1kola !

cmd/resolutioncli/main.go
tmshort: I'm a fan of explicit arguments; it can get confusing with positional unless there's an explicit logic (e.g. `kubectl`'s _verb_ _type_ [_name_])

tmshort (changes requested):
Some small changes (mostly copyright years). However, I would like to see this incorporated in the `Makefile` as either a target that's automatically built, or a separate target.

cmd/resolutioncli/main.go
m1kola: Package name probably makes sense as a positional argument and everything else as flags. But I suggest that we leave it as is for now and come back to discuss details of the interface once start work on finalising this CLI.

We are not using cobra right now, but I think we should use it in the production-ready version.

internal/resolution/variable_sources/olm/olm.go
```diff
@@ -42,10 +62,8 @@ func (o *VariableSource) GetVariables(ctx context.Context, entitySource input.En
 		if err != nil {
 			return nil, err
 		}
-		inputVariableSources = append(inputVariableSources, rps)
+		variableSources = append(variableSources, rps)
 	}
 
-	// build variable source pipeline
-	variableSource := crdconstraints.NewCRDUniquenessConstraintsVariableSource(bundlesanddependencies.NewBundlesAndDepsVariableSource(inputVariableSources...))
```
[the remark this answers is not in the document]
m1kola: I had to remove this chaining within the variable source becase for off-cluster usage we need a bit different set of variable soruces (e.g. we need a variable source which creates a variable from CLI inputs).

instead of chaining I now use `NestedVariableSource` in `NewOLMVariableSource` (note after #273 it is `NewOperatorVariableSource`). `NewOLMVariableSource` is used in on-cluster operator and it is not in this diff (I did not change the calling code).


https://github.com/operator-framework/operator-controller/blob/9a1a15d7d7c7e069dab21ff5930052d13553e1fc/cmd/manager/main.go#L108C19-L108C19

But `NewCRDUniquenessConstraintsVariableSource` and `NewBundlesAndDepsVariableSource` are also used in off-cluster resolution [here within `olm.NestedVariableSource`](https://github.com/operator-framework/operator-controller/pull/246/files#diff-0e1413d2458238cf94143692a8b6990ff6d5da773e6ea780707586fb1d609409R125-R139).

I suspect `NestedVariableSource` and probably whole `composite.go` will go away once we switch to the solution proposed in the [RFC](https://docs.google.com/document/d/1PizF1uQcMrrAFsLgcx2O0Coq7mei4xe9hxpz-QRp-x0/edit#heading=h.x3tfh25grvnv).

cmd/resolutioncli/main.go
```diff
@@ -0,0 +1,184 @@
...
+
+	codecs = serializer.NewCodecFactory(scheme)
+)
+
+func init() {
+	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
+	utilruntime.Must(operatorsv1alpha1.AddToScheme(scheme))
+	utilruntime.Must(rukpakv1alpha1.AddToScheme(scheme))
+	utilruntime.Must(catalogd.AddToScheme(scheme))
+}
```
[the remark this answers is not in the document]
m1kola: We can do it in `main` I think, but I was following the pattern from operator-controller's `main.go`:

https://github.com/operator-framework/operator-controller/blob/ebc5cfec16be3c9512c2c34ce0b6877ee3ffe040/cmd/manager/main.go#L49-L57

m1kola (commented):
@tmshort I think I addressed your feedback + made sure this PR doesn't decrease test coverage.

> However, I would like to see this incorporated in the Makefile as either a target that's automatically built, or a separate target.

I did not want to add this into makefile or goreleaser because it is still a PoC and a lot will change. Didn't want to make it "official" for now.

But if you think that we should still do that - I suggest that we do it in a separate PR.

tmshort:
I would like to see a m`Makefile` entry for this, but perhaps not now.

tmshort (approved):

PR #248, opened 2023-06-01
Replacing ginkgo test with go testing

jubittajohn:
# Description

Changes for #191


internal/resolution/resolver_test.go
```diff
@@ -19,112 +18,115 @@ import (
...
+		{
+			Name: "should resolve the packages described by the available Operator resources",
+			Client: FakeClient(testResource...),
+			EntitySource: testEntitySource,
+			SelectedVariableCnt: 4,
+			ExpectedError: nil,
+		},
+		{
+			Name: "should not return an error if there are no Operator resources",
+			Client: FakeClient([]client.Object{}...),
```
m1kola, line 80: This should also work:

```suggestion
			Client: FakeClient(),
```

internal/resolution/resolver_test.go
```diff
@@ -19,112 +18,115 @@ import (
...
+	}
+	
+	errorClient := NewFailClientWithError(fmt.Errorf("something bad happened"))
+
+	for _, tt := range []struct {
+		Name				string
+		Client				client.Client
+		EntitySource		input.EntitySource
+		SelectedVariableCnt	int
+		ExpectedError		error
```
m1kola, line 69: I think this file wasn't `go fmt`-ed.

internal/resolution/resolver_test.go
```diff
@@ -19,112 +18,115 @@ import (
...
+				assert.Nil(t, solution)
+			} else {
+				assert.Len(t, solution.SelectedVariables(), tt.SelectedVariableCnt)
+				if len(solution.SelectedVariables()) == 4 {
+					assert.True(t, solution.IsSelected("operatorhub/packageA/2.0.0"))
+					assert.True(t, solution.IsSelected("operatorhub/prometheus/0.47.0"))
+					assert.True(t, solution.IsSelected("required package packageA"))
+					assert.True(t, solution.IsSelected("required package prometheus"))
+					assert.False(t, solution.IsSelected("operatorhub/prometheus/0.37.0"))
+				}
```
m1kola, line 115: What happens if we add another test case where we expect 3 variables to be selected? Or 4, but different 4?

internal/resolution/resolver_test.go
```diff
@@ -19,112 +18,115 @@ import (
...
+		},
+		{
+			Name: "should return an error if the entity source throws an error",
+			Client: FakeClient(testResource...),
+			EntitySource: FailEntitySource{},
+			ExpectedError: fmt.Errorf("error calling filter in entity source"),
+		}, 
+		{
+			Name: "should return an error if the client throws an error",
+			Client: errorClient,
```
m1kola, line 93: It is not a long line, so I would just do this inline:

```suggestion
			Client: NewFailClientWithError(fmt.Errorf("something bad happened")),
```

m1kola (changes requested):
Need to tidy up a bit, but it is a good start!

internal/resolution/resolver_test.go
jubittajohn: Sure I will do that. This would cover an already covered portion of the code that tests if packages described by the available Operator resources are resolved.

m1kola (commented):
I need to properly review updated version - so far just a small suggestion so far.

I will take a good look tomorrow.

internal/resolution/resolver_test.go
```diff
@@ -19,112 +18,113 @@ import (
...
+					assert.True(t, solution.IsSelected(identifier))
+				}
+				assert.NoError(t, err)
+			}
+
+		})
+	}
+}
+
+func FakeClient(objects ...client.Object) client.Client {
```
ankitathomas, line 121: It doesn't seem like FakeClient is used anywhere else, this doesn't necessarily need to be exported

internal/resolution/resolver_test.go
```diff
@@ -19,112 +18,113 @@ import (
...
+			ExpectedError: fmt.Errorf("something bad happened"),
+		},
+	} {
+		t.Run(tt.Name, func(t *testing.T) {
+			client := tt.Client
+			entitySource := tt.EntitySource
+			resolver := resolution.NewOperatorResolver(client, entitySource)
+			solution, err := resolver.Resolve(context.Background())
+
+			if err != nil {
```
ankitathomas, line 106: This check should probably be looking at tt.ExpectedError. The way it's currently written, it can miss a case where err is nil when it should be non-nil.

internal/resolution/resolver_test.go
```diff
@@ -19,112 +18,113 @@ import (
...
-		"olm.package":     "{\"packageName\":\"prometheus\",\"version\":\"0.47.0\"}",
-	}),
-	"operatorhub/packageA/2.0.0": *input.NewEntity("operatorhub/packageA/2.0.0", map[string]string{
-		"olm.bundle.path": `"foo.io/packageA/packageA:v2.0.0"`,
-		"olm.channel":     "{\"channelName\":\"stable\",\"priority\":0}",
-		"olm.gvk":         "[{\"group\":\"foo.io\",\"kind\":\"Foo\",\"version\":\"v1\"}]",
-		"olm.package":     "{\"packageName\":\"packageA\",\"version\":\"2.0.0\"}",
-	}),
-}
+		"operatorhub/prometheus/0.47.0": *input.NewEntity("operatorhub/prometheus/0.47.0", map[string]string{
```
ankitathomas, line 28: nit: does this need the additional space?

ankitathomas (commented):
I have a few small suggestions, but great work so far!

internal/resolution/resolver_test.go
jubittajohn: go fmt creates this additional spacing.

internal/resolution/resolver_test.go
jubittajohn: The FakeClient is being used inside the structure by the Client field for creating fake clients.

ankitathomas: As long as it's being used within the same package, it doesn't need to be exported

ankitathomas (approved):
/lgtm

internal/resolution/resolver_test.go
```diff
@@ -19,112 +18,114 @@ import (
...
+			},
+		},
+	}
+
+	for _, tt := range []struct {
+		Name              string
+		Client            client.Client
+		EntitySource      input.EntitySource
+		SelectedVariables []deppy.Identifier
+		ExpectedError     error
```
m1kola, line 69: Also minor things (no need to change):

* These fields in don't have to be exported as they are in the anonimous struct & never exported
* To distinguish between inputs and expected results I would prefix all the expected data with `want` or `expected`. E.g.:

```suggestion
		name                      string
		client                    client.Client
		entitySource              input.EntitySource
		expectedselectedVariables []deppy.Identifier
		expectedError             error
```

internal/resolution/resolver_test.go
```diff
@@ -19,112 +18,114 @@ import (
...
+			Name:          "should return an error if the client throws an error",
+			Client:        NewFailClientWithError(fmt.Errorf("something bad happened")),
+			EntitySource:  testEntitySource,
+			ExpectedError: fmt.Errorf("something bad happened"),
+		},
+	} {
+		t.Run(tt.Name, func(t *testing.T) {
+			client := tt.Client
+			entitySource := tt.EntitySource
+			resolver := resolution.NewOperatorResolver(client, entitySource)
```
m1kola, line 104: There is no need in these two varaibles, I think.

```suggestion
			resolver := resolution.NewOperatorResolver(tt.client, tt.entitySource)
```

m1kola (approved):
I only managed to come up with a few minor suggestions, but nothing blocking.

internal/resolution/resolver_test.go
```diff
@@ -19,112 +18,114 @@ import (
...
-	RunSpecs(t, "Operator Resolver Suite")
-}
 
-func FakeClient(objects ...client.Object) client.Client {
-	scheme := runtime.NewScheme()
-	if err := v1alpha1.AddToScheme(scheme); err != nil {
-		panic(fmt.Sprintf("error creating fake client: %s", err))
+	testEntityCache := map[deppy.Identifier]input.Entity{"operatorhub/prometheus/0.37.0": *input.NewEntity(
+		"operatorhub/prometheus/0.37.0", map[string]string{
+			"olm.bundle.path": `"quay.io/operatorhubio/prometheus@sha256:3e281e587de3d03011440685fc4fb782672beab044c1ebadc42788ce05a21c35"`,
```
anik120, line 24: This is an actual production image, doesn't looks like there's any reason we can't just use "foo.io/bar/baz"?

internal/resolution/resolver_test.go
```diff
@@ -19,112 +18,114 @@ import (
...
+			"olm.channel":     "{\"channelName\":\"beta\",\"priority\":0,\"replaces\":\"prometheusoperator.0.32.0\"}",
+			"olm.gvk":         "[{\"group\":\"monitoring.coreos.com\",\"kind\":\"Alertmanager\",\"version\":\"v1\"}, {\"group\":\"monitoring.coreos.com\",\"kind\":\"Prometheus\",\"version\":\"v1\"}]",
```
anik120, line 26: These aren't actually being used right now, right? It'll be cleaner to only include the properties that are actually being used (`olm.package`, `olm.bundle.path`), and include other fields like gvk iteratively as we start actually using them

anik120 (changes requested):

internal/resolution/resolver_test.go
ankitathomas: As long as FBC populates those properties, I think it's still useful to at least have them mocked on test objects.

anik120 (approved):
/lgtm
/approve

PR #249, opened 2023-06-02
Updates deppy

m1kola:
# Description

Bumps deppy to the latest commit in master.

## Reviewer Checklist

- [ ] API Go Documentation


internal/resolution/variable_sources/olm/olm_test.go
```diff
@@ -177,8 +177,8 @@ var _ input.EntitySource = &FailEntitySource{}
 type FailEntitySource struct {
 }
 
-func (f FailEntitySource) Get(ctx context.Context, id deppy.Identifier) *input.Entity {
-	return nil
+func (f FailEntitySource) Get(ctx context.Context, id deppy.Identifier) (*input.Entity, error) {
+	return nil, fmt.Errorf("error executing get")
```
varshaprasad96, line 181: Not a blocker at all: but do we want to define this as constants?
panicEntitySrcErr, FailEntitySrcErr etc.. just in case we want to use it in tests to assert the errors.

varshaprasad96 (approved):
I think I missed the change where removed returning error from `NewDeppySolver`. 

Else it looks good to me!
/lgtm

internal/resolution/variable_sources/olm/olm_test.go
m1kola: I see how it might be useful, but I would like to avoid doing it in this PR. Focus of this is to upgrade Deppy with minial changes to operator controller.

This function follows the existing pattern for `Filter`, `GroupBy`, etc.

Issue #247, opened 2023-06-01
Refactor catalogsource reconcile e2e test to a unit test


ankitathomas:
In order to keep the e2e test suite focused purely on the essentials, refactor the catalog reconcile test into a set of unit tests checking for whether operator-controller reconciles correctly on:
- catalog creation
- catalog update
- catalog delete

See: https://github.com/operator-framework/operator-controller/pull/216/files#r1199511183

PR #250, opened 2023-06-02
(e2e) Use custom built test catalog for e2e testing

### LINKED ISSUE #215: e2e flake: [FAILED] Operator Install when An operator is installed from an operator catalog

anik120:
This PR
* Stands up a local image registry in the cluster
* Builds bundles and catalog images and uploads them to the image registry
* Uses the custom images in the e2e test suite

* Also introduces a `pullSecret` field for the Operator API's Spec struct, to allow installation of operators on cluster whose bundles require imagePullSecret to be provisioned. This is required because the bundle images built and uploaded to the local registry above requires a pull secret for the local registry.

closes #215

# Description

## Reviewer Checklist

- [ ] API Go Documentation


joelanford:
> - Stands up a local image registry in the cluster
> - Builds bundles and catalog images and uploads them to the image registry

I think this is overkill. We should be able to simply build the images, `kind load` them, and then refer to them directly in the e2e. No need to run an extra image registry, IMO.

joelanford:
I think we should emulate the way rukpak set this up:
Build and load here: https://github.com/operator-framework/rukpak/blob/fdc3ebd85c928072fc6f6b59a2debf752159b791/Makefile#L185

Directly refer to those image in e2e's. For example: https://github.com/operator-framework/rukpak/blob/fdc3ebd85c928072fc6f6b59a2debf752159b791/test/e2e/plain_provisioner_test.go#L71

everettraven:
>Stands up a local image registry in the cluster

This feels very heavy handed - is there a reason we have to stand up an image registry on cluster instead of just `kind load`ing the images? As long as we aren't using images with `:latest` tags the [default Kube pull policy](https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting) should just allow us to use `kind load` to load images and use them with the unpack pods.

In my head the most simple and straightforward approach to using a custom built test catalog would be to:
1. Build a custom test catalog & pull the bundle image, tagging with a non `:latest` tag
2 `kind load docker-image ...` for both images from 1
3. Change the catalog image in the e2e tests
4. Run the e2e tests

anik120:
Looks like my use of the `latest` tag was in fact the problem. 

>:latest tags the [default Kube pull policy](https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting)

TIL. 

Updated the PR to just use built and loaded images

Makefile
```diff
@@ -94,6 +97,12 @@ kind-cluster: kind kind-cluster-cleanup ## Standup a kind cluster
 kind-cluster-cleanup: kind ## Delete the kind cluster
 	$(KIND) delete cluster --name ${KIND_CLUSTER_NAME}
 
+kind-load-test-artifacts: kind ## Load the e2e testdata container images into a kind cluster
+	$(CONTAINER_RUNTIME) build $(TESTDATA_DIR)/bundles/registry-v1/prometheus-operator.v0.47.0 -t localhost/testdata/bundles/registry-v1/prometheus-operator:v0.47.0
```
everettraven, line 101: For fetching the bundle we are probably going to want this to have a tag other than latest as well:
```suggestion
	$(CONTAINER_RUNTIME) build $(TESTDATA_DIR)/bundles/registry-v1/prometheus-operator.v0.47.0 -t localhost/testdata/bundles/registry-v1/prometheus-operator:v0.47.0:e2e
```

testdata/catalogs/test-catalog/catalog.yaml
```diff
@@ -0,0 +1,104 @@
...
+        {
+            "name": "prometheusoperator.0.47.0"
+        }
+    ]
+}
+{
+    "schema": "olm.bundle",
+    "name": "prometheusoperator.0.47.0",
+    "package": "prometheus",
+    "image": "localhost/testdata/bundles/registry-v1/prometheus-operator:v0.47.0",
```
everettraven, line 20: We should specify a non-latest tag here:
```suggestion
    "image": "localhost/testdata/bundles/registry-v1/prometheus-operator:v0.47.0:e2e",
```

testdata/catalogs/test-catalog/catalog.yaml
```diff
@@ -0,0 +1,104 @@
...
+    "relatedImages": [
+        {
+            "name": "",
+            "image": "quay.io/operatorhubio/prometheus@sha256:47c659b352bbc4b75b64fbe0077d19afaea80cf691c62d803d44fdfc88dc9c6f"
+        },
+        {
+            "name": "",
+            "image": "quay.io/prometheus-operator/prometheus-operator:v0.47.0"
+        }
+    ]
```
everettraven, line 103: Do we need the related images?

everettraven (changes requested):

Makefile
everettraven: Misread this, this is fine - my bad!

everettraven (approved):
Overall looks good to me, only have a question about the related images in the bundle blob

anik120:
> Overall looks good to me, only have a question about the related images in the bundle blob

Removed the related images, they aren't needed 👍🏽

testdata/catalogs/test-catalog.Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+ENTRYPOINT ["/bin/opm"]
+CMD ["serve", "/configs", "--cache-dir=/tmp/cache"]
+
+# Copy declarative config root into image at /configs and pre-populate serve cache
+ADD test-catalog /configs
+RUN ["/bin/opm", "serve", "/configs", "--cache-dir=/tmp/cache", "--cache-only"]
+
+# Set DC-specific label for the location of the DC root directory
+# in the image
+LABEL operators.operatorframework.io.index.configs.v1=/configs
```
joelanford, line 15: This should be able to be much simpler in OLMv1:

```Dockerfile
FROM scratch
ADD test-catalog /configs
```

testdata/catalogs/test-catalog/catalog.yaml
joelanford, line 1: Rename file to `catalog.json`?

testdata/catalogs/test-catalog/catalog.yaml
```diff
@@ -0,0 +1,94 @@
...
+            }
+        },
+        {
+            "type": "olm.gvk",
+            "value": {
+                "group": "monitoring.coreos.com",
+                "kind": "ThanosRuler",
+                "version": "v1"
+            }
+        },
```
joelanford, line 85: I'd kinda prefer that we add properties only as we decide we're going to use them and test something about them. That way we don't have a bunch of inert stuff in here that our future selves have to wonder if its necessary to keep.

testdata/bundles/registry-v1/prometheus-operator.v0.47.0/metadata/annotations.yaml
```diff
@@ -0,0 +1,7 @@
+annotations:
+  operators.operatorframework.io.bundle.channel.default.v1: beta
```
joelanford, line 2: Remove this line? Default channel is not a concept in OLMv1 (I don't think)

testdata/catalogs/test-catalog/catalog.yaml
everettraven: Instead of renaming can we actually convert all the JSON in this file to YAML? I rebased https://github.com/operator-framework/operator-controller/pull/242 on this PR locally and having this as JSON was causing some issues due to it being a json stream in a single .json file. YAML will let us keep it all in the same file

everettraven: Although this was likely a problem on my end as the tests seem to pass on this PR. I also personally prefer YAML over JSON but that personal preference shouldn't block this PR

anik120: Done

testdata/catalogs/test-catalog/catalog.yaml
anik120: Makes total sense 👍🏽

testdata/catalogs/test-catalog/catalog.yaml
```diff
@@ -0,0 +1,94 @@
+{
+    "schema": "olm.package",
+    "name": "prometheus",
+    "defaultChannel": "beta"
```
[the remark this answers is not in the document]
anik120: Looks like adding comments to a json file isn't that easy. 
One way would be to add a key/value of the form "comment": "TODO:...", but that's possibly a bit confusing?

testdata/catalogs/test-catalog.Dockerfile
anik120: Looks like it is 🎉

everettraven (approved):
All this seems fine for now. There is likely going to be some modifications as part of #242

testdata/catalogs/test-catalog/catalog.json
```diff
@@ -0,0 +1,33 @@
...
+            "type": "olm.package",
+            "value": {
+                "packageName": "prometheus",
+                "version": "0.47.0"
+            }
+        }
+    ],
+    "relatedImages": [
+        {}
+    ]
```
joelanford, line 32: Nit: remove `"relatedImages`" entirely?

test/e2e/install_test.go
```diff
@@ -43,158 +43,182 @@ var _ = Describe("Operator Install", func() {
...
-				g.Expect(cond.Status).To(Equal(metav1.ConditionTrue))
-				g.Expect(cond.Reason).To(Equal(rukpakv1alpha1.ReasonUnpackSuccessful))
-
-				cond = apimeta.FindStatusCondition(bd.Status.Conditions, rukpakv1alpha1.TypeInstalled)
-				g.Expect(cond).ToNot(BeNil())
-				g.Expect(cond.Status).To(Equal(metav1.ConditionTrue))
-				g.Expect(cond.Reason).To(Equal(rukpakv1alpha1.ReasonInstallationSucceeded))
-			}).WithTimeout(defaultTimeout).WithPolling(defaultPoll).Should(Succeed())
-		})
-		It("resolves again when a new catalog is available", func() {
```
joelanford, line 118: Are these tests essentially removed from this PR? I know @ankitathomas has #247 open, that shuffles things around, but I didn't expect to see this removed in the context of this PR.

test/e2e/install_test.go
```diff
@@ -43,158 +43,182 @@ var _ = Describe("Operator Install", func() {
...
+			Expect(bd.Spec.Template.Spec.Source.Type).To(Equal(rukpakv1alpha1.SourceTypeImage))
+			Expect(bd.Spec.Template.Spec.Source.Image.Ref).To(Equal("localhost/testdata/bundles/registry-v1/prometheus-operator:v0.47.0"))
+			Expect(operator.Status.ResolvedBundleResource).To(Equal(bd.Spec.Template.Spec.Source.Image.Ref))
+			Expect(operator.Status.InstalledBundleResource).To(Equal(bd.Spec.Template.Spec.Source.Image.Ref))
 		})
 	})
+	// TODO: Turn this test on once a operators in the plain bundle format are installable from a catalog, and use the
+	// plain bundle included in a separate catalog than the registryV1-catalog
+
+	// When("An operator installation is attempted that isn't available in any catalog", func() {
```
joelanford, line 93: Should we remove these comments from this PR and have them added back uncommented in the plain-bundle-support PR? Though it seems these are the tests Ankita added, but now they need plain-bundle support? ... Maybe it's because it's late on Friday, but I'm thoroughly confused.

testdata/catalogs/test-catalog/catalog.yaml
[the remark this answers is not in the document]
joelanford: Oh yeah, forgot about that. Let's just leave it as is.

testdata/catalogs/test-catalog/catalog.json
anik120: It looks like spec.relatedImages field is not optional: 

```
condition "Unpacked" is False/UnpackFailed; message: "create bundle metadata objects: applying bundle metadata \"test-catalog-prometheusoperator.0.47.0\": BundleMetadata.catalogd.operatorframework.io \"test-catalog-prometheusoperator.0.47.0\" is invalid: spec.relatedImages: Invalid value: \"null\": spec.relatedImages in body must be of type array: \"null\""
```

Admittedly I'd have preferred to keep it at `"relatedImages": []` but even that is read as null, and had to do some json jujitsu to come up with `[{}]`

test/e2e/install_test.go
anik120: Apologies, I'd discussed the context for this change in the [slack thread](https://kubernetes.slack.com/archives/C0181L6JYQ2/p1685622093221839) but should have left a comment here. 

Essentially, with the change from `WithTimeout(5 * time.Minute)` to `WithTimeout(defaultTimeout)`, the tests were stepping on each other's toes. There was a ton of 
* "delete resources, and wait for upto 5 mins for the resources to be cleaned up"
* "do this, this and this to speed up the resource cleanup" 
etc etc. 

When I'd pushed this change, I'd thought about using a different catalog for this test, so the three specs in this suite can be independent from each other ([in line with ginkgo best practices](https://onsi.github.io/ginkgo/#:~:text=This%20allows%20Ginkgo%20to%20shuffle,and%20run%20specs%20in%20parallel.)). However, in the conversation in the slack thread, it became more apparent that not only will these tests benefit from using a different catalog, but they most likely belong to a different test suite altogether. Yet however, I noticed https://github.com/operator-framework/operator-controller/issues/247 which is converting these to unit tests anyway, so I left them commented out, so that https://github.com/operator-framework/operator-controller/issues/247 can just remove the commented out lines. 

Commenting them in in this PR requires either 
A) the building of another catalog with a new operator so that the specs can remain independent of each other, or 
B) having them run sequentially by adding safeguards such as "wait for resources to clean up" and "do a, b and c to speed up resource clean up" etc etc

option A will be wasted effort in this PR since 
a) a new bundle will be added when [plain bundle installation capability is added](https://github.com/operator-framework/operator-controller/pull/242)
b) the test is going away anyway. 

option B....doesn't help the "bring the overall time for the test suite run" objective of this PR at all. 

So leaving this commented out in this PR so that https://github.com/operator-framework/operator-controller/issues/247 can simply get rid of these commented out lines.

testdata/catalogs/test-catalog/catalog.json
everettraven: ^ Shouldn't be an issue after https://github.com/operator-framework/catalogd/issues/38 is done

joelanford (approved):
/approve

testdata/catalogs/test-catalog/catalog.json
anik120: We'll have to create a follow up issue to fix this.

PR #251, opened 2023-06-02
🌱 Catalog unit test

ankitathomas:
This PR refactors the e2e catalog reconcile test to a unit test checking whether reconcile is triggered on catalog events (create, update, delete events). Reconcile is detected by scanning controller logs for logs from the reconcile loop for all operators.

See: https://github.com/operator-framework/operator-controller/issues/247

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -94,14 +94,14 @@ func main() {
 		os.Exit(1)
 	}
 
-	if err = (&controllers.OperatorReconciler{
+	if err = controllers.SetupWithManager(&controllers.OperatorReconciler{
 		Client: mgr.GetClient(),
 		Scheme: mgr.GetScheme(),
 		Resolver: solver.NewDeppySolver(
 			entitysources.NewCatalogdEntitySource(mgr.GetClient()),
 			olm.NewOLMVariableSource(mgr.GetClient()),
 		),
-	}).SetupWithManager(mgr); err != nil {
+	}, mgr); err != nil {
 		setupLog.Error(err, "unable to create controller", "controller", "Operator")
 		os.Exit(1)
 	}
--- a/internal/controllers/catalog_test.go
+++ b/internal/controllers/catalog_test.go
@@ -1,0 +1,106 @@
+package controllers_test
+
+import (
+	"context"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	catalogd "github.com/operator-framework/catalogd/api/core/v1alpha1"
+	operatorsv1alpha1 "github.com/operator-framework/operator-controller/api/v1alpha1"
+	"github.com/operator-framework/operator-controller/internal/controllers"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+func collectRequests(reqChan <-chan string) func() []string {
+	var reqNames []string
+	return func() []string {
+		select {
+		case req := <-reqChan:
+			reqNames = append(reqNames, req)
+		default:
+		}
+		return reqNames
+	}
+}
+
+func operatorForPackage(name, pkg string) *operatorsv1alpha1.Operator {
+	return &operatorsv1alpha1.Operator{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: name,
+		},
+		Spec: operatorsv1alpha1.OperatorSpec{
+			PackageName: pkg,
+		},
+	}
+}
+
+var _ = Describe("SetupWithManager", func() {
+	When("catalog events occur", func() {
+		var cancel context.CancelFunc
+		var ctx context.Context
+		var opNames []string
+		var reqChan chan string
+		BeforeEach(func() {
+			ctx = context.Background()
+			mgr, err := ctrl.NewManager(cfg, ctrl.Options{Scheme: sch})
+			Expect(err).To(BeNil())
+
+			opNames = []string{"prometheus", "project-quay"}
+			reqChan = make(chan string)
+			var fakeReconciler reconcile.Func = func(_ context.Context, request ctrl.Request) (ctrl.Result, error) {
+				reqChan <- request.Name
+				return ctrl.Result{}, nil
+			}
+			Expect(controllers.SetupWithManager(fakeReconciler, mgr)).To(Succeed())
+
+			var mgrCtx context.Context
+			mgrCtx, cancel = context.WithCancel(ctx)
+			go func() {
+				Expect(mgr.Start(mgrCtx)).To(Succeed())
+			}()
+
+			for _, p := range opNames {
+				Expect(cl.Create(ctx, operatorForPackage(p, p))).To(Succeed())
+			}
+			By("verifying initial reconcile logs for operator creation")
+			Eventually(collectRequests(reqChan)).Should(ConsistOf(opNames))
+		})
+		It("reconciles all affected operators on cluster", func() {
+			By("creating a new catalog")
+			catalog := &catalogd.Catalog{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "t",
+				},
+				Spec: catalogd.CatalogSpec{
+					Source: catalogd.CatalogSource{
+						Type:  catalogd.SourceTypeImage,
+						Image: &catalogd.ImageSource{},
+					},
+				},
+			}
+			Expect(cl.Create(ctx, catalog)).To(Succeed())
+			By("verifying operator reconcile logs on catalog create")
+			Eventually(collectRequests(reqChan)).Should(ConsistOf(opNames))
+
+			By("updating a catalog")
+			catalog.Spec.Source.Image.Ref = "s"
+			Expect(cl.Update(ctx, catalog)).To(Succeed())
+			By("verifying operator reconcile logs on catalog update")
+			Eventually(collectRequests(reqChan)).Should(ConsistOf(opNames))
+
+			By("deleting a catalog")
+			Expect(cl.Delete(ctx, catalog)).To(Succeed())
+			By("verifying operator reconcile logs on catalog delete")
+			Eventually(collectRequests(reqChan)).Should(ConsistOf(opNames))
+		})
+		AfterEach(func() {
+			cancel() // stop manager
+			close(reqChan)
+			for _, p := range opNames {
+				Expect(cl.Delete(ctx, operatorForPackage(p, p))).To(Succeed())
+			}
+		})
+	})
+})
--- a/internal/controllers/operator_controller_test.go
+++ b/internal/controllers/operator_controller_test.go
@@ -3,13 +3,9 @@ package controllers_test
 import (
 	"context"
 	"fmt"
-	"strings"
-	"time"
 
-	"github.com/go-logr/logr/funcr"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
-	catalogd "github.com/operator-framework/catalogd/api/core/v1alpha1"
 	"github.com/operator-framework/deppy/pkg/deppy"
 	"github.com/operator-framework/deppy/pkg/deppy/input"
 	"github.com/operator-framework/deppy/pkg/deppy/solver"
@@ -22,7 +18,6 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
-	"sigs.k8s.io/controller-runtime/pkg/log"
 
 	operatorsv1alpha1 "github.com/operator-framework/operator-controller/api/v1alpha1"
 	"github.com/operator-framework/operator-controller/internal/conditionsets"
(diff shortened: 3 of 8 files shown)


internal/controllers/operator_controller_test.go
```diff
@@ -1018,6 +1023,94 @@ var _ = Describe("Operator Controller Test", func() {
 			Expect(cond.Message).To(Equal("installation has not been attempted as spec is invalid"))
 		})
 	})
+	When("a catalog changes on cluster", func() {
```
joelanford, line 1026: We may want to consider sectioning these tests off in a separate test file, or at least root level `Describe` block. Reasoning is that everything else here is assuming no manager, no controller, and instead manually running `Reconcile` with pre-setup and post-assertions.

These tests start a manager, use `Eventually`, etc.

internal/controllers/operator_controller_test.go
```diff
@@ -1018,6 +1023,94 @@ var _ = Describe("Operator Controller Test", func() {
...
+		var logCount int
+		BeforeEach(func() {
+			l := funcr.New(func(prefix, args string) {
+				if prefix == "operator-controller" &&
+					strings.Contains(args, `"controller"="operator"`) &&
+					strings.Contains(args, `"msg"="ending"`) {
+					// filter for only relevant logs
+					testLogs = append(testLogs, fmt.Sprintf("%s", args))
+				}
+			}, funcr.Options{Verbosity: 1})
```
joelanford, line 1038: TIL, this `funcr` is super cool.

I'm torn though. On one hand, doing things this way means we can pretty much test our controller setup logic as-is. However it seems very brittle because changing our logs makes our tests for a completely unrelated thing break.

Another option is to slightly restructure the `SetupWithManager` like this:
```go
// SetupWithManager sets up the controller with the Manager.
func (r *OperatorReconciler) SetupWithManager(mgr ctrl.Manager) error {
	_, err := NewOperatorController(mgr, r)
	return err
}

// NewOperatorController sets up a controller for Operator objects using the provided reconciler.
func NewOperatorController(mgr ctrl.Manager, r reconcile.Reconciler) (controller.Controller, error) {
	return ctrl.NewControllerManagedBy(mgr).
		For(&operatorsv1alpha1.Operator{}).
		Watches(source.NewKindWithCache(&catalogd.Catalog{}, mgr.GetCache()),
			handler.EnqueueRequestsFromMapFunc(operatorRequestsForCatalog(context.TODO(), mgr.GetClient(), mgr.GetLogger()))).
		Owns(&rukpakv1alpha1.BundleDeployment{}).
		Build(r)
}
```

And then our tests could use `NewOperatorController` with a custom reconciler that collects the `reconcile.Requests` and makes assertions based on that. WDYT?

internal/controllers/operator_controller_test.go
```diff
@@ -1018,6 +1023,94 @@ var _ = Describe("Operator Controller Test", func() {
...
+				Expect(err).To(BeNil())
+			}
+			Eventually(func(g Gomega) {
+				By("verifying initial reconcile logs for operator creation")
+				g.Expect(len(testLogs) >= len(opNames)).To(BeTrue())
+				for _, p := range opNames {
+					g.Expect(testLogs[len(testLogs)-len(opNames):]).To(ContainElement(ContainSubstring(fmt.Sprintf("\"Operator\"={\"name\":\"%s\"}", p))))
+				}
+				logCount = len(testLogs)
+			}).WithTimeout(2 * time.Second).WithPolling(1 * time.Second).WithArguments().Should(Succeed())
```
joelanford, line 1065: Are these custom timeouts and polling configurations necessary?

ankitathomas: The timeouts are necessary. The default 1s timeout isn't enough time for the operatorReconciler to consistently go through all queued objects, we'd get a lot of flakes without it.

The timeouts are less necessary when switching to a fake reconciler though.

internal/controllers/catalog_test.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/operator-framework/operator-controller/internal/controllers"
+
+	catalogd "github.com/operator-framework/catalogd/pkg/apis/core/v1beta1"
+	operatorsv1alpha1 "github.com/operator-framework/operator-controller/api/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+var _ = Describe("Operator Controller Catalog Reconciler", func() {
```
joelanford, line 17: WDYT about making this whole section the `SetupWithManager` tests. So we're essentially making this the place that we test our watches (and anything else we might setup).
```suggestion
var _ = Describe("SetupWithManager", func() {
```

internal/controllers/catalog_test.go
```diff
@@ -0,0 +1,97 @@
...
+	When("a catalog changes on cluster", func() {
```
joelanford, line 18: Based on the previous suggestion, WDYT about adding an extra layer of hierarchy under the `Describe("SetupWithManager")` block. This way it'll be setup a bit easer to handle more watches or other configurations we add in the future.

```
BeforeEach() // setup and start the controller+manager
When("Operator events occur")
    It("reconciles self") // do Operator create, update, delete
When("Catalog events occur")
    BeforeEach() // create a few operators
    It("reconciles all Operators") // do Catalog create, update, delete
    AfterEach() // delete the operators
AfterEach() // cancel the context to stop the manager
```

internal/controllers/catalog_test.go
```diff
@@ -0,0 +1,97 @@
...
+				op := &operatorsv1alpha1.Operator{ObjectMeta: metav1.ObjectMeta{Name: p}, Spec: operatorsv1alpha1.OperatorSpec{PackageName: p}}
+				err := cl.Create(ctx, op)
+				Expect(err).To(BeNil())
+			}
+			Eventually(func(g Gomega) {
+				By("verifying initial reconcile logs for operator creation")
+				g.Expect(len(reconcileReqs) >= len(opNames)).To(BeTrue())
+				g.Expect(reconcileReqs).To(ContainElements(opNames))
+				logCount = len(reconcileReqs)
+			}).Should(Succeed())
```
joelanford, line 55: Not positive this works, but docs make it seem like it should:
```suggestion
			Eventually(reconcileReqs).Should(Equal(opNames))
```
`Equal` instead of `ContainElements` because (I'm pretty sure) we can expect a specific order of reconciles since we don't currently configure concurrent reconciles. 

For the `logCount` thing, I wonder if there is an alternative that lets us:
- not need it
- increase clarity of the subsequent test assertions

Perhaps we could "reset" the reconciler (i.e. make the slice 0 length) after each assertion? Might need a mutex to avoid a race condition.

But if we did that, then all of the `Eventually` checks would (I think?) look like the suggestion in this comment.

internal/controllers/catalog_test.go
```diff
@@ -0,0 +1,97 @@
...
+			Eventually(func(g Gomega) {
+				By("verifying operator reconcile logs on catalog delete")
+				g.Expect(reconcileReqs).To(HaveLen(logCount + len(opNames)))
+				g.Expect(reconcileReqs[len(reconcileReqs)-len(opNames):]).To(ConsistOf(opNames))
+			}).Should(Succeed())
+		})
+		AfterEach(func() {
+			for _, p := range opNames {
+				op := &operatorsv1alpha1.Operator{ObjectMeta: metav1.ObjectMeta{Name: p}, Spec: operatorsv1alpha1.OperatorSpec{PackageName: p}}
+				Expect(cl.Delete(ctx, op)).To(BeNil())
```
joelanford, line 92: Nit:
```suggestion
				Expect(cl.Delete(ctx, op)).To(Succeed())
```

testdata/crds/catalogd.operatorframework.io_catalogs.yaml
joelanford, line 1: This concerns me a little bit because it seems like something that will become outdated and potentially cause false negatives or false positives from our tests.

I see that there's already precedent for this with the BD CRD though, so I think it's okay to merge with this.

However, could you capture an issue for us to look into keeping these CRDs up-to-date based on what our `go.mod` says?

joelanford (commented):
Looks really good. Just a few (hopefully easy to implement) suggestions that I think will make things more clear and result in fewer lines of code.

internal/controllers/catalog_test.go
ankitathomas: I've been thinking about this, but do we want to test controller-runtime objects? Testing for reconcile on catalog events makes sense with catalogs since we're using a custom handler, but with operators, it feels like we're testing whether the manager works, that should be on controller-runtime.

internal/controllers/catalog_test.go
ankitathomas: I think we should consider the possibility of reconciles happening out of order, especially if we're introducing a potential flake.

testdata/crds/catalogd.operatorframework.io_catalogs.yaml
ankitathomas: https://github.com/operator-framework/operator-controller/issues/254

internal/controllers/catalog_test.go
joelanford: It isn't really testing controller-runtime. It's testing our specific watch setup. I agree that almost all of the code that _actually_ runs is controller-runtime code, but it's all triggered based on our setup of the manager.

I think there's pretty good long term value here from a regression standpoint, especially if we can get the framework setup while things are simple. Anecdotally, the rukpak watches are actually fairly complex due to the way its APIs reference provisioner classes. There has been discussion in the context of oeprator-controller to implement things like upgrade policies via referenced objects, so I think it is inevitable that our watches become more complex over time.

internal/controllers/catalog_test.go
joelanford: Ah, yeah. That makes sense. I think the handler make a `List` call and that returns Operators in a non-deterministic order.

I checked out this PR locally to continue investigating the channel idea. I ran these updated tests with some extra debug logging and noticed that there are specs in operator_controller_test that don't clean up after themselves and are causing _these_ tests to have to deal with that pollution.

I have a fix for that here: #257 

I think we should merge that, rebase this, and then see again if we can improve on the way we assert for reconciles. I still think there's a way to do this with channels (perhaps receive from channel into a slice and then assert on the slice, but without the slice being reused for multiple assertions?).

internal/controllers/catalog_test.go
everettraven: I think @joelanford's suggestion here makes a lot of sense for testing that the watches we are setting up are getting triggered as expected. That being said, I don't think it should block this PR specifically unless @ankitathomas really felt like including it. At least the foundation is laid by this PR and we can always increase the robustness of our tests in follow ups

internal/controllers/catalog_test.go
```diff
@@ -0,0 +1,109 @@
...
+				Expect(err).To(BeNil())
+			}
+			By("verifying initial reconcile logs for operator creation")
+			var reqNames []string
+			for range opNames {
+				var req string
+				Eventually(reqChan).Should(Receive(&req))
+				reqNames = append(reqNames, req)
+			}
+			Expect(reqNames).To(ConsistOf(opNames))
```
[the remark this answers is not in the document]
joelanford: +1, here's what I had come up with:

```go
func collectRequests(reqChan <-chan string) func() []string {
	var reqNames []string
	return func() []string {
		select {
		case req := <-reqChan:
			reqNames = append(reqNames, req)
		default:
		}
		return reqNames
	}
}
```

And then:
```go
Eventually(collectRequests(reqChan)).Should(ConsistOf(opNames))
```

internal/controllers/catalog_test.go
```diff
@@ -0,0 +1,109 @@
...
+		var opNames = []string{"prometheus", "project-quay"}
+		var reqChan = make(chan string, 4*len(opNames))
+		BeforeEach(func() {
+			ctx = context.Background()
+			mgr, err := ctrl.NewManager(cfg, ctrl.Options{Scheme: sch})
+			Expect(err).To(BeNil())
+
+			var fakeReconciler reconcile.Func = func(_ context.Context, request ctrl.Request) (ctrl.Result, error) {
+				// ignore reconcile requests from other tests
+				if !strings.HasPrefix(request.Name, "operator-test") {
```
joelanford, line 31: We shouldn't do this, IMO. This should hopefully be solved by #257 .

internal/controllers/catalog_test.go
```diff
@@ -0,0 +1,109 @@
...
+				reqNames = append(reqNames, req)
+			}
+			Expect(reqNames).To(ConsistOf(opNames))
+		})
+		AfterEach(func() {
+			for _, p := range opNames {
+				op := &operatorsv1alpha1.Operator{ObjectMeta: metav1.ObjectMeta{Name: p}, Spec: operatorsv1alpha1.OperatorSpec{PackageName: p}}
+				Expect(cl.Delete(ctx, op)).To(Succeed())
+			}
+			cancel() // stop manager
```
joelanford, line 106: We should stop the manager before the clean up. That way we don't attempt any further reconciles based on these deletions (and cause undesired channel send attempts).

internal/controllers/catalog_test.go
```diff
@@ -0,0 +1,109 @@
...
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+var _ = Describe("SetupWithManager", func() {
+	When("catalog events occur", func() {
+		var cancel context.CancelFunc
+		var ctx context.Context
+		var opNames = []string{"prometheus", "project-quay"}
+		var reqChan = make(chan string, 4*len(opNames))
```
joelanford, line 23: A few minor improvements:
1. We should probably initialize the channel in `BeforeEach`
2. If we cancel the manager context before deleting the Operator objects in `AfterEach`, I think we can use an unbuffered channel
3. We should probably close the channel in `AfterEach` _after_ we cancel the manager context.

testdata/crds/catalogd.operatorframework.io_catalogs.yaml
```diff
@@ -0,0 +1,172 @@
...
+spec:
+  group: catalogd.operatorframework.io
+  names:
+    kind: Catalog
+    listKind: CatalogList
+    plural: catalogs
+    singular: catalog
+  scope: Cluster
+  versions:
+    - name: v1beta1
```
joelanford, line 16: Sorry about this, but I just merged in catalogd v0.3.0, which changed the CRDs. We should re-pull this one. I think it was just the version changing from v1beta1 to v1alpha1, but probably best to just copy the whole thing over again.

cmd/manager/main.go
```diff
@@ -94,14 +94,13 @@ func main() {
...
-	if err = (&controllers.OperatorReconciler{
+	if err = controllers.SetupWithManager(&controllers.OperatorReconciler{
 		Client: mgr.GetClient(),
 		Scheme: mgr.GetScheme(),
 		Resolver: solver.NewDeppySolver(
 			entitysources.NewCatalogdEntitySource(mgr.GetClient()),
-			olm.NewOLMVariableSource(mgr.GetClient()),
-		),
-	}).SetupWithManager(mgr); err != nil {
+			olm.NewOLMVariableSource(mgr.GetClient())),
```
everettraven, line 102: Nit: Keep the `solver.NewDeppySolver` closing parenthesis on it's own line. IMO it makes it easier to see where the call stops at a glance

everettraven (commented):
Added another nit and agree with @joelanford's suggestions

internal/controllers/catalog_test.go
```diff
@@ -0,0 +1,94 @@
...
+		var ctx context.Context
+		var opNames []string
+		var reqChan chan string
+		BeforeEach(func() {
+			ctx = context.Background()
+			mgr, err := ctrl.NewManager(cfg, ctrl.Options{Scheme: sch})
+			Expect(err).To(BeNil())
+
+			opNames = []string{"prometheus", "project-quay"}
+			reqChan = make(chan string, 4*len(opNames))
```
everettraven, line 40: Do we still need the buffer to be `4*len(opNames)`? I don't think its a huge deal, but IIUC we could probably reduce this to just `len(opNames)` since a receiver will be using the channel after each "reconcile" that occurs.

internal/controllers/catalog_test.go
```diff
@@ -0,0 +1,94 @@
...
+			var fakeReconciler reconcile.Func = func(_ context.Context, request ctrl.Request) (ctrl.Result, error) {
+				reqChan <- request.Name
+				return ctrl.Result{}, nil
+			}
+
+			err = controllers.SetupWithManager(fakeReconciler, mgr)
+			Expect(err).To(BeNil())
```
everettraven, line 47: Nit (stolen from a nit @joelanford left on one of my PRs but I really liked it so :shrug: ): For functions that just return an error, I think something like:
```suggestion
			Expect(controllers.SetupWithManager(fakeReconciler, mgr)).To(Succeed())
```
looks a bit nicer

internal/controllers/catalog_test.go
```diff
@@ -0,0 +1,94 @@
...
+				op := &operatorsv1alpha1.Operator{ObjectMeta: metav1.ObjectMeta{Name: p}, Spec: operatorsv1alpha1.OperatorSpec{PackageName: p}}
+				err := cl.Create(ctx, op)
+				Expect(err).To(BeNil())
+			}
+			By("verifying initial reconcile logs for operator creation")
+			Eventually(collectRequests(reqChan)).Should(ConsistOf(opNames))
+		})
+		It("reconciles all affected operators on cluster", func() {
+			By("creating a new catalog")
+			catalog := &catalogd.Catalog{ObjectMeta: metav1.ObjectMeta{Name: "t"}, Spec: catalogd.CatalogSpec{Source: catalogd.CatalogSource{Type: catalogd.SourceTypeImage, Image: &catalogd.ImageSource{}}}}
```
everettraven, line 67: Nit: Make this take up multiple lines - this is a really long line and is more difficult to follow than if it was on multiple lines

internal/controllers/catalog_test.go
```diff
@@ -0,0 +1,94 @@
...
+			Eventually(collectRequests(reqChan)).Should(ConsistOf(opNames))
+
+			By("deleting a catalog")
+			err = cl.Delete(ctx, catalog)
+			Expect(err).To(BeNil())
+			By("verifying operator reconcile logs on catalog delete")
+			Eventually(collectRequests(reqChan)).Should(ConsistOf(opNames))
+		})
+		AfterEach(func() {
+			cancel() // stop manager
```
everettraven, line 87: Do we need to close the channel as well?

perdasilva:
@ankitathomas can you please rebase?

PR #257, opened 2023-06-08
operator_controller_test: ensure we cleanup after each spec

joelanford:
# Description

I was reviewing #251 and noticed that there were leftover items in the envtest etcd after several operator_controller_test specs completed.

This PR:
1. Updates operator_controller_test.go to clean up after itself
2. Adds a check in AfterSuite to verify that no operators or bundle deployments remain in the cluster after all tests complete.

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/controllers/operator_controller_test.go
+++ b/internal/controllers/operator_controller_test.go
@@ -52,6 +52,11 @@ var _ = Describe("Operator Controller Test", func() {
 		BeforeEach(func() {
 			opKey = types.NamespacedName{Name: fmt.Sprintf("operator-test-%s", rand.String(8))}
 		})
+		AfterEach(func() {
+			verifyInvariants(ctx, reconciler.Client, operator)
+			Expect(cl.DeleteAllOf(ctx, &operatorsv1alpha1.Operator{})).To(Succeed())
+			Expect(cl.DeleteAllOf(ctx, &rukpakv1alpha1.BundleDeployment{})).To(Succeed())
+		})
 		When("the operator specifies a non-existent package", func() {
 			var pkgName string
 			BeforeEach(func() {
@@ -137,7 +142,6 @@ var _ = Describe("Operator Controller Test", func() {
 				err := cl.Create(ctx, operator)
 				Expect(err).NotTo(HaveOccurred())
 			})
-
 			When("the BundleDeployment does not exist", func() {
 				BeforeEach(func() {
 					By("running reconcile")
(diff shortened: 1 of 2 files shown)


internal/controllers/operator_controller_test.go
```diff
@@ -665,51 +659,6 @@ var _ = Describe("Operator Controller Test", func() {
 				Expect(cond.Message).To(Equal("installation has not been attempted as resolution failed"))
 			})
 		})
-		When("the existing operator status is based on bundleDeployment", func() {
```
joelanford, line 668: As of https://github.com/operator-framework/operator-controller/pull/134, this block no longer has any specs in it. So it is unused. Therefore removing it.

internal/controllers/operator_controller_test.go
```diff
@@ -954,75 +903,64 @@ var _ = Describe("Operator Controller Test", func() {
...
+				fakeClient client.Client
+			)
+			BeforeEach(func() {
+				opKey = types.NamespacedName{Name: fmt.Sprintf("operator-validation-test-%s", rand.String(8))}
 
-			err := cl.Delete(ctx, operator)
-			Expect(err).To(Not(HaveOccurred()))
-		})
-	})
-	When("an invalid semver is provided that bypasses the regex validation", func() {
```
joelanford, line 964: Moved this block into the `When("the operator exists")` so that it also passes through the `verifyInvariants` check that we apply to operators after calling `Reconcile`.

internal/controllers/suite_test.go
```diff
@@ -77,7 +79,27 @@ var _ = BeforeSuite(func() {
...
 
 var _ = AfterSuite(func() {
+	var operators operatorsv1alpha1.OperatorList
+	var bundleDeployments rukpakv1alpha1.BundleDeploymentList
+
+	Expect(cl.List(context.Background(), &operators)).To(Succeed())
+	Expect(cl.List(context.Background(), &bundleDeployments)).To(Succeed())
+
+	Expect(namesFromList(&operators)).To(BeEmpty(), "operators left in the cluster")
+	Expect(namesFromList(&bundleDeployments)).To(BeEmpty(), "bundle deployments left in the cluster")
```
everettraven, line 89: Would something like:
```suggestion
	Expect(operators.Items).To(BeEmpty(), "operators left in the cluster")
	Expect(bundleDeployments.Items).To(BeEmpty(), "bundle deployments left in the cluster")
```

suffice? Upside I see of your method is that it would show the names of the operator objects leftover, but I'm not sure what the `.Items` approach would output off the top of my head (maybe additional helpful info?)

everettraven (approved):
Just a question, but this looks good to me overall

internal/controllers/suite_test.go
joelanford: I had that before, and it shows a _huge_ output of the entire structure of the missing objects. To the point where the deluge of info made it seem less helpful.

With this PR getting the current number down to 0 and this test passing, any future failures should be pretty obvious where to look: in the code that was changed.

PR #275, opened 2023-06-27
Implement upgrade edges support

### LINKED ISSUE #267: Solver: Introduce constraints for "upgrade edges"
For an operator upgrade to be valid, the new operator version should comply with a set of constraints:

- Channel: The new operator version should be part of the same channel 
- Replaces: The currently installed operator version should be included in the "replace" field of the new desired operator version (see skips)
- Skips: Skips define from what version it is possible to jump to without upgrading one version at a time. This should be considered and not only the "replaces" field.  

related to https://github.com/operator-framework/operator-controller/issues/231

### LINKED ISSUE #268: Solver: Introduce variable sources for installed bundles

jmprusi:
# Description

Adds upgrade edges support to operator-controller

- Annotates bundleDeployments with the original operator information (packageName, Channel, Version)
- Adds a new constraint that uses the packages "replaces" value to ensure that the upgrade path between the previous version and the new one is valid.  

Closes #268
Closes #267

## Reviewer Checklist

- [ ] API Go Documentation


internal/resolution/entities/bundle_entity.go
```diff
@@ -121,14 +119,35 @@ func (b *BundleEntity) ChannelName() (string, error) {
...
+		return "", err
+	}
+	return b.replaces.Replaces, nil
+}
+
+func (b *BundleEntity) loadReplaces() error {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	if b.replaces == nil {
+		// TODO: move property name to constant
```
perdasilva, line 143: should we address this todo?

internal/resolution/entities/bundle_entity_test.go
```diff
@@ -269,6 +265,27 @@ var _ = Describe("BundleEntity", func() {
 		})
 	})
 
+	Describe("Replaces", func() {
```
perdasilva, line 268: nice!

internal/resolution/entitysources/catalogdsource.go
```diff
@@ -106,6 +110,9 @@ func getEntities(ctx context.Context, client client.Client) (input.EntityList, e
 				if catalogScopedEntryName == bundle.Name {
 					channelValue, _ := json.Marshal(property.Channel{ChannelName: ch.Name, Priority: 0})
 					props[property.TypeChannel] = string(channelValue)
+					// TODO(jmprusi): Add the proper PropertyType for this
```
perdasilva, line 113: should we address this todo?

internal/resolution/variables/installed_package.go
```diff
@@ -0,0 +1,32 @@
...
+type InstalledPackageVariable struct {
+	*input.SimpleVariable
+	bundleEntities []*olmentity.BundleEntity
+}
+
+func (r *InstalledPackageVariable) BundleEntities() []*olmentity.BundleEntity {
+	return r.bundleEntities
+}
+
+func NewInstalledPackageVariable(bundleImage string, bundleEntities []*olmentity.BundleEntity) *InstalledPackageVariable {
```
perdasilva, line 22: should we add a unit test for this function?

internal/resolution/variablesources/bundle_deployment.go
```diff
@@ -0,0 +1,58 @@
...
+		variableSources = append(variableSources, o.inputVariableSource)
+	}
+
+	bundleDeployments := rukpakv1alpha1.BundleDeploymentList{}
+	if err := o.client.List(ctx, &bundleDeployments); err != nil {
+		return nil, err
+	}
+
+	processed := map[string]struct{}{}
+	if len(bundleDeployments.Items) > 1 {
```
perdasilva, line 39: debug statement, should remove

internal/resolution/variablesources/bundle_deployment.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+func NewBundleDeploymentVariableSource(cl client.Client, inputVariableSource input.VariableSource) *BundleDeploymentVariableSource {
+	return &BundleDeploymentVariableSource{
+		client:              cl,
+		inputVariableSource: inputVariableSource,
+	}
+}
+
+func (o *BundleDeploymentVariableSource) GetVariables(ctx context.Context, entitySource input.EntitySource) ([]deppy.Variable, error) {
```
perdasilva, line 27: we should probably have good coverage of this function, wdyt?

internal/resolution/variablesources/installed_package.go
```diff
@@ -0,0 +1,83 @@
...
+	resultSet, err := entitySource.Filter(ctx, predicates.WithBundleImage(r.bundleImage))
+	if err != nil {
+		return nil, err
+	}
+	if len(resultSet) == 0 {
+		return nil, r.notFoundError()
+	}
+
+	// sort by channel and version
+	// TODO: this is a bit of a hack and it assumes a well formed catalog.
```
perdasilva, line 33: could you please create issue/s for these  todos and link them in the comments here?

perdasilva (changes requested):
Looks really good ^^ Thank you for this! I know it was tough. I think we just need to address a couple of the todos, add issue/s for the others, and maybe a little bit more unit test coverage. Great work!

internal/resolution/entitysources/catalogdsource.go
joelanford: I know the following suggestion is a bigger refactor, and I'm not suggesting we do it now, but I think we should refactor things such that there are:
- variable source for bundles
- variable source for channels

Bundle entities would no longer carry upgrade edge or channel membership information. Rather, channel variables would incorporate bundle membership and upgrade edges.

joelanford: IMO, something like my suggestion should happen almost immediately after this PR lands so that we set ourselves up for a more extensible future where different channel implementations can exist and we don't have to keep adding synthetic properties to bundles (which might have undesirable effects if/when we support more arbitrary contraint definitions in the future)

internal/resolution/variablesources/bundle_deployment.go
jmprusi: done, CI failing due to codecov issues..

internal/resolution/variablesources/bundle_deployment.go
jmprusi: removed

internal/resolution/entities/bundle_entity.go
jmprusi: Removing this TODO as it is tied to: https://github.com/operator-framework/operator-controller/pull/275#discussion_r1265236837

awgreene (approved):
/approve

internal/resolution/entitysources/catalogdsource.go
```diff
@@ -106,6 +110,9 @@ func getEntities(ctx context.Context, client client.Client) (input.EntityList, e
...
+					replacesValue, _ := json.Marshal(replacesProperty{Replaces: b.Replaces})
+					props["olm.replaces"] = string(replacesValue)
```
perdasilva, line 115: should we move this to a  constant? (olm.replaces)

perdasilva (approved):
lgtm - small non-blocking issue around a magic string. But we can tackle that on in the next PR ^^

internal/resolution/variables/installed_package.go
```diff
@@ -0,0 +1,32 @@
...
+	id := deppy.IdentifierFromString(fmt.Sprintf("installed package %s", bundleImage))
+	entityIDs := make([]deppy.Identifier, 0, len(bundleEntities))
+	for _, bundle := range bundleEntities {
+		entityIDs = append(entityIDs, bundle.ID)
+	}
+	return &InstalledPackageVariable{
+		SimpleVariable: input.NewSimpleVariable(id, constraint.Mandatory(), constraint.Dependency(entityIDs...)),
+		bundleEntities: bundleEntities,
+	}
+}
```
m1kola, line 32: @jmprusi what is the difference between this and `RequiredPackageVariable`?

perdasilva: They are semantically different concepts that get translated to the same sort of structure (variable with mandatory and dependency constraints). I guess we could separate that out as a concept if we want to simplify the code here a bit. wdyt?

m1kola: > They are semantically different concepts that get translated to the same sort of structure (variable with mandatory and dependency constraints).

@perdasilva Is it just semantic difference? Or are they/will they be processed differently? As far as I understnad - they are exactly the same today.

> I guess we could separate that out as a concept if we want to simplify the code here a bit. wdyt?

For now I'm just trying to understand how it works. But I don't follow - what do you mean by separating? They are already separate structures.

perdasilva: For now it is. It could be that these two evolve on different paths. I meant separate out the pattern of a Variable with a Mandatory and Dependency constraints. I think it could be a common pattern. Then we can just typedef the required and installed package variables as that.

m1kola: Thanks for the clarification.

> I meant separate out the pattern of a Variable with a Mandatory and Dependency constraints. I think it could be a common pattern. Then we can just typedef the required and installed package variables as that.

Sounds like a good idea.

PR #283, opened 2023-07-03
Cleans up kustomize files & gets rid of deprecated fields

m1kola:
# Description

* `patchesStrategicMerge` is deprecated
* `config/default/manager_config_patch.yaml` is unused
* Removes some unused scaffolding

## Reviewer Checklist

- [ ] API Go Documentation


config/crd/kustomization.yaml
```diff
@@ -5,15 +5,15 @@ resources:
...
 # patches here are for enabling the conversion webhook for each CRD
-#- patches/webhook_in_operators.yaml
+#- path: patches/webhook_in_operators.yaml
 #+kubebuilder:scaffold:crdkustomizewebhookpatch
 
 # [CERTMANAGER] To enable cert-manager, uncomment all the sections with [CERTMANAGER] prefix.
 # patches here are for enabling the CA injection for each CRD
-#- patches/cainjection_in_operators.yaml
+#- path: patches/cainjection_in_operators.yaml
 #+kubebuilder:scaffold:crdkustomizecainjectionpatch
```
everettraven, line 17: Should we include this section at all? Do we anticipate needing to use these sections? It looks like everything under this section is commented out and is effectively unused for now.

m1kola: I'm not aware of the background for this so I decided to leave it as is. But I'm happy to clean this up too if we are sure we don't need this.

everettraven: based on the `+kubebuilder:` markers in there this looks like it came from initially scaffolding the project with Kubebuilder. I we anticipate adding webhooks we can leave it in but if not it should be safe to remove.

m1kola: Good point. Removed this stuff.

everettraven (approved):

PR #284, opened 2023-07-03
Adds E2E coverage reporting

m1kola:
# Description

Adds E2E coverage reporting.

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -22,4 +22,17 @@ jobs:
 
     - name: Run e2e tests
       run: |
-        make e2e
+        # By default make stops building on first non-zero exit code which
+        # in case of E2E tests will mean that code coverage will only be
+        # collected on successful runs. We want to collect coverage even
+        # after failing tests.
+        # With -k flag make will continue the build, but will return non-zero
+        # exit code in case of any errors.
+        make -k e2e
+
+    - uses: codecov/codecov-action@v3
+      with:
+        files: e2e-cover.out
+        flags: e2e
+        fail_ci_if_error: true
+        functionalities: fixes
--- a/Makefile
+++ b/Makefile
@@ -131,8 +131,9 @@ export GO_BUILD_ASMFLAGS ?= all=-trimpath=${PWD}
 export GO_BUILD_LDFLAGS  ?= -s -w -X $(shell go list -m)/version.Version=$(VERSION)
 export GO_BUILD_GCFLAGS  ?= all=-trimpath=${PWD}
 export GO_BUILD_TAGS     ?= upstream
+export GO_BUILD_FLAGS    ?=
 
-BUILDCMD = go build -tags '$(GO_BUILD_TAGS)' -ldflags '$(GO_BUILD_LDFLAGS)' -gcflags '$(GO_BUILD_GCFLAGS)' -asmflags '$(GO_BUILD_ASMFLAGS)' -o $(BUILDBIN)/manager ./cmd/manager
+BUILDCMD = go build $(GO_BUILD_FLAGS) -tags '$(GO_BUILD_TAGS)' -ldflags '$(GO_BUILD_LDFLAGS)' -gcflags '$(GO_BUILD_GCFLAGS)' -asmflags '$(GO_BUILD_ASMFLAGS)' -o $(BUILDBIN)/manager ./cmd/manager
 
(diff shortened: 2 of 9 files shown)


Makefile
```diff
@@ -103,7 +105,16 @@ test-unit: $(SETUP_ENVTEST) ## Run the unit tests
 	eval $$($(SETUP_ENVTEST) use -p env $(ENVTEST_VERSION)) && go test -tags $(GO_BUILD_TAGS) -count=1 -short $(UNIT_TEST_DIRS) -coverprofile cover.out
 
 e2e: KIND_CLUSTER_NAME=operator-controller-e2e
-e2e: run kind-load-test-artifacts test-e2e kind-cluster-cleanup ## Run e2e test suite on local kind cluster
+e2e: KUSTOMIZE_BUILD_DIR=config/e2e
+e2e: GO_BUILD_FLAGS=-cover
+e2e: run kind-load-test-artifacts test-e2e ## Run e2e test suite on local kind cluster
+	# Coverage-instrumented binary produces coverage on termination, so we scale down the manager before gathering the coverage
```
perdasilva, line 111: great use of comments ^^ thank you!

m1kola:
@perdasilva thanks for the review.

> What do you think is missing before we can take it off draft state?

There were few things which I wanted to change before marking ready for review:

* I didn't want to have a bunch of `kubectl` commands in the `Makefile` They are now in `./hack/e2e-coverage.sh`.
* I didn't like the very long json in `--overrides` in `kubectl run`. I moved pod creation into `config/e2e/manager_e2e_coverage_copy_pod.yaml`.
* I had to change github action to make sure that we collect coverage even if tests fail. In the previous version coverage was only collected on success due to `make` aborting execution on first non-zero exit code. I added `-k`/`--keep-going` flag to `make -k e2e` which will continue make build, but will return non-zero exit code in case of failures.

The PR should be ready now. I'm just checking few things before I actually take it off the draft state.

.github/workflows/e2e.yaml
```diff
@@ -22,4 +22,11 @@ jobs:
 
     - name: Run e2e tests
       run: |
-        make e2e
+        make -k e2e
```
ncdc, line 25: Mind commenting why `-k` is here? I liked your explanation in the comments and think it would be worth putting in here.

Makefile
```diff
@@ -103,7 +105,13 @@ test-unit: $(SETUP_ENVTEST) ## Run the unit tests
 	eval $$($(SETUP_ENVTEST) use -p env $(ENVTEST_VERSION)) && go test -tags $(GO_BUILD_TAGS) -count=1 -short $(UNIT_TEST_DIRS) -coverprofile cover.out
 
 e2e: KIND_CLUSTER_NAME=operator-controller-e2e
-e2e: run kind-load-test-artifacts test-e2e kind-cluster-cleanup ## Run e2e test suite on local kind cluster
```
ncdc, line 106: I see that the help text `## Run e2e test suite on local kind cluster` is gone - did you mean to remove it?

hack/e2e-coverage.sh
```diff
@@ -0,0 +1,33 @@
...
+    rm -rf "$COVERAGE_DIR"
+}
+trap cleanup EXIT
+
+# Coverage-instrumented binary produces coverage on termination,
+# so we scale down the manager before gathering the coverage
+kubectl -n "$OPERATOR_CONTROLLER_NAMESPACE" scale deployment/"$OPERATOR_CONTROLLER_MANAGER_DEPLOYMENT_NAME" --replicas=0
+
+# Wait for the copy pod to be ready
+kubectl -n "$OPERATOR_CONTROLLER_NAMESPACE" wait --for=condition=ready pod "$COPY_POD_NAME"
```
ncdc, line 25: Is it possible for the copy pod to get scheduled and be ready before the manager pod, i.e. "locking" the PVC and preventing the manager pod from starting?

Makefile
m1kola: Good catch! Didn't mean to remove help comment.

hack/e2e-coverage.sh
ncdc: Thanks, I was incorrect in my thinking that ReadWriteOnce meant it would only bind to 1 pod at a time.

ncdc (approved):

PR #286, opened 2023-07-05
Cross-component E2E for operator framework

jubittajohn:
# Description
This PR is for a cross-component demo with all OLM v1 repositories. The ginkgo test does the following: 
-  Uses operator-sdk and kustomize to build `plain+v0` bundles and create catalogs to include the bundles.
- Installs, upgrades and deletes a `plain+v0` operator.
- Uses operator-sdk to build `registry+v1` bundles and create catalogs to include the bundles.
- Installs, upgrades and deletes a `registry+v1` operator.

## Purpose of the E2E tests.
- Development on OLM v1 is split across multiple repositories, and the list of relevant repositories may grow over time. While we have demos showing improvements in functionality of components over time, it can be difficult to have a picture of the state of OLM v1 at any given time for someone not following its development closely. Having a single source to look for OLM v1 behavior can provide more clarity about the state of the project.
- With the scale of the OLM v1 project, it is useful to have a means to test components in the operator development + lifecycle pipeline together to create a more cohesive experience for all users.
- Test workflows for common personas as per user documentation to guarantee the accuracy of documented steps.

## Workflow

1. start with an empty directory
2. call operator-sdk to initialize and generate an operator
3. generate a bundle directory
4. build/push/kind load bundle images from the bundle directories
5. repeat steps 2-4 as necessary to get bundles for multiple operator versions
6. generate a catalog directory
7. build/push/kind load the catalog
8. create a Catalog CR (with kubectl operator)
9. create an Operator CR (with kubectl operator)
10. trigger Operator upgrades (with kubectl operator)
11. delete the Operator CR (with kubectl operator)
12. delete the Catalog CR (with kubectl operator)
13. repeat steps 2-12 for each bundle format (e.g. registry+v1 and plain+v0)
   
 - Checks to be ensured:
   - Ensuring packages and bundlemetadata are created for the catalog.
   - Ensuring operator operations succeed.
   - Verifying presence/absence of relevant manifests.
   
 - Added new Makefile target/Github action:
   - makefile target: `operator-developer-e2e`
   - Github action: /workflows/operator-developer-e2e.yaml
   - Note: The current `test-e2e` are only for testing the functionality of operator-controller.  The new e2e test is meant for an operator-framework e2e or an OLM v1 e2e and to capture the developer experience. 

## How to run

- Makefile target `operator-developer-e2e` : Runs the entire E2E setup.
- Makefile target `test-op-dev-e2e`: Runs the ginkgo test.
- Makefile target `deploy-local-registry`: Deploys local registry
- Makefile target `cleanup-local-registry` : Stops and removes local registry
- Makefile target `kind-cluster-cleanup`: Deletes the kind cluster
  
## Getting Started

Below are the input values used in the test.

- Building operator-controller, deploying it into the cluster and rest of the configuration is done in the `MakeFile` of this repo under the target `operator-developer-e2e`. This includes:
    
    - Setting up a kind cluster named `operator-controller-op-dev-e2e`.
    - Installing the operator controller onto the cluster.
    - Setting up `opm`, `operator-sdk` and `kustomize` using bingo.
    - Setting up a local registry server for building and loading images.

- The following structs defined are required as input for both plain+v0 and registry+v1 bundles:
   - For getting bundle related inputs:
        ```
        type BundleInfo struct {
            baseFolderPath string 
            bundles        []BundleContent
        }

        type BundleContent struct {
            bInputDir     string
            bundleVersion string
            imageRef      string
        }
        ```
        - `baseFolderPath` - Base/root path of the folder for the specific bundle type input data.[path to plain-v0 or registry-v1 bundles testdata]
        - `bundles` - Stores the data relevant to different versions of the bundle.
        - `bInputDir` - The input directory containing the specific version of the bundle data.
        - `bundleVersion` - The specific version of the bundle data.
        - `imageRef` - This is formed. Stores the bundle image reference which will be of the format `localhost:5001/<operator_name>-bundle:v.<bundleVersion>`
    - For getting catalog related inputs:
        ```
        type CatalogDInfo struct {
            baseFolderPath     string
            catalogDir         string
            operatorName       string
            desiredChannelName string
            imageRef           string
            fbcFileName        string
        }
        ```
        - `baseFolderPath` - Base/root path of the folder for the catalogs.
        - `operatorName` - Name of the operator to be installed from the bundles.
        - `desiredChannelName` - Desired channel name for the operator.
        - `catalogDir` - This is formed. The directory to store the FBC. The formed value will be of the format: `<operator-name>-catalog`
        - `imageRef` - This is formed. Stores the FBC image reference which will be of the format: `localhost:5001/<username>/<catalogDir>:test`
        - `fbcFileName` - Name of the FBC file. This is hard-coded as `catalog.yaml`.
 - For getting information related to the install/upgrade action for operators:
        ```
        type OperatorActionInfo struct {
            installVersion string 
            upgradeVersion string
        }
        ``` 
        - `installVersion` - Version of the operator to be installed on the cluster.
        - `upgradeVersion` - Version of the operator to be upgraded on the cluster.
        
    - The below input is used to form the bundle using operator-sdk.
        ```
        type SdkProjectInfo struct {
            projectName string
            domainName  string
            group       string
            version     string
            kind        string
        }

<!--
Thank you for your contribution!

Please provide a summary of the changes and the motivation behind the change.
-->

## Reviewer Checklist

- [ ] API Go Documentation


Makefile
```diff
@@ -123,6 +129,43 @@ kind-load-test-artifacts: $(KIND) ## Load the e2e testdata container images into
 	$(KIND) load docker-image localhost/testdata/bundles/plain-v0/plain:v0.1.0 --name $(KIND_CLUSTER_NAME)
 	$(KIND) load docker-image localhost/testdata/catalogs/test-catalog:e2e --name $(KIND_CLUSTER_NAME)
 
+.PHONY: opm
+OPM = ./bin/opm
+OPM_VERSION = v1.28.0
+opm: ## Download opm locally if necessary.
```
dtfranz, line 135: Preferably if we're going to be using the opm binary in operator-controller then we should add it with [`bingo`](https://github.com/bwplotka/bingo) as we've done with the other binary dependencies. You should be able to add that with:  
```
$ bingo get github.com/operator-framework/operator-registry/cmd/opm@v1.28.0
```
The same applies to `operator-sdk` down below, you should be able to add that to the project as well with:
```
$ bingo get github.com/operator-framework/operator-sdk/cmd/operator-sdk@v1.31.0
```

Makefile
```diff
@@ -123,6 +129,43 @@ kind-load-test-artifacts: $(KIND) ## Load the e2e testdata container images into
...
+install-operator-sdk:  ## Install operator-sdk
+	if [ ! -f ~/bin/operator-sdk ]; then \
+		curl -LO $(OPERATOR_SDK_DL_URL)/operator-sdk_${OS}_${ARCH} ; \
+		chmod +x operator-sdk_${OS}_${ARCH} ; \
+		mv operator-sdk_$(OS)_$(ARCH) /usr/local/bin/operator-sdk ; \
+	fi
+
+.PHONY: deploy-local-registry
+deploy-local-registry: ## Deploy local docker registry
+	docker run -d -p 5000:5000 --restart=always --name local-registry registry:2
```
dtfranz, line 163: Our makefile isn't doing this consistently yet but we should probably start using `$(CONTAINER_RUNTIME)` here instead of `docker`.

test/operator-framework-e2e/README.md
```diff
@@ -0,0 +1,122 @@
+# Cross-component E2E for operator framework
```
dtfranz, line 1: Thanks for adding a readme for this!

test/operator-framework-e2e/operator_framework_test.go
```diff
@@ -0,0 +1,970 @@
...
+		})
+	})
+	When("Delete an operator", func() {
+		It("Delete an operator", func() {
+			err = deleteOperator(ctx, catalogDInfo.operatorName)
+			Expect(err).ToNot(HaveOccurred())
+
+			By("Verifying the operator doesn't exist")
+			Eventually(func(g Gomega) {
+				err = checkOperatorDeleted(catalogDInfo.operatorName)
```
dtfranz, line 223: Given the func name `checkOperatorDeleted(...)`, it might make more sense for it to return a bool and do something like `return err != nil && errors.IsNotFound(err)` inside of the func, which would also simplify the checks here a bit.

test/operator-framework-e2e/operator_framework_test.go
```diff
@@ -0,0 +1,970 @@
...
+			By("Creating FBC for plain bundle using custom routine")
+			var imageRefs []string
+			var bundleVersions []string
+			for _, b := range bundleInfo.bundles {
+				imageRefs = append(imageRefs, b.imageRef)
+				bundleVersions = append(bundleVersions, b.bundleVersion)
+			}
+			fbc := CreateFBC(catalogDInfo.operatorName, catalogDInfo.desiredChannelName, imageRefs, bundleVersions)
+			err = WriteFBC(*fbc, catalogDInfo.baseFolderPath+"/"+catalogDInfo.catalogDir, catalogDInfo.fbcFileName)
+			Expect(err).To(Not(HaveOccurred()))
```
dtfranz, line 170: Just a minor style nitpick but perhaps makes these consistently `Expect(err).ToNot(HaveOccurred())` like you have it in some other places.

.github/workflows/operator-developer-e2e.yaml
```diff
@@ -0,0 +1,23 @@
+name: operator-developer-e2e
+
+on:
+  workflow_dispatch:
+  pull_request:
+  push:
+   branches:
+   - main
```
tmshort, line 8: Inconsistent indentation: should be 2 spaces (i.e. more than 1) to match the rest of the file.

Makefile
```diff
@@ -97,6 +97,9 @@ E2E_FLAGS ?= ""
 test-e2e: $(GINKGO) ## Run the e2e tests
 	$(GINKGO) --tags $(GO_BUILD_TAGS) $(E2E_FLAGS) -trace -progress $(FOCUS) test/e2e
 
+test-operator-developer-e2e: $(GINKGO) ## Run operator create, upgrade and delete tests
+	$(GINKGO) --tags $(GO_BUILD_TAGS) $(E2E_FLAGS) -trace -progress test/operator-framework-e2e
```
tmshort, line 101: The name's a bit unwieldy... maybe `test-op-dev-e2e`?

tmshort: Does this need `$(FOCUS)` as in the `test-e2e` case?

Makefile
```diff
@@ -105,6 +108,9 @@ test-unit: $(SETUP_ENVTEST) ## Run the unit tests
 e2e: KIND_CLUSTER_NAME=operator-controller-e2e
 e2e: run kind-load-test-artifacts test-e2e kind-cluster-cleanup ## Run e2e test suite on local kind cluster
 
+operator-developer-e2e: KIND_CLUSTER_NAME=operator-controller-e2e  ## Run operator-developer e2e on local kind cluster
```
tmshort, line 111: Should this have a different `kind` cluster name than the `e2e` test?

Makefile
```diff
@@ -123,6 +129,43 @@ kind-load-test-artifacts: $(KIND) ## Load the e2e testdata container images into
...
+		mv operator-sdk_$(OS)_$(ARCH) /usr/local/bin/operator-sdk ; \
+	fi
+
+.PHONY: deploy-local-registry
+deploy-local-registry: ## Deploy local docker registry
+	docker run -d -p 5000:5000 --restart=always --name local-registry registry:2
+
+.PHONY: stop-local-registry
+stop-local-registry: ## Stop and remove local registry
+	docker container stop local-registry && docker container rm -v local-registry
```
tmshort, line 167: Use `$(CONTAINER_RUNTIME)` here
Split this into two lines, no need for the `&&` as the `Makefile` will terminate (or at least it should) on error

test/operator-framework-e2e/create_fbc_helper.go
```diff
@@ -0,0 +1,155 @@
...
+			Value: json.RawMessage(fmt.Sprintf(`{"packageName": "%s", "version": "%s"}`, pkgName, versions[i])),
+		})
+		properties = append(properties, property.Property{
+			Type:  SchemaBundleMediatype,
+			Value: json.RawMessage(fmt.Sprintf(`"%s"`, BundleMediatype)),
+		})
+
+		bundleFormed = append(bundleFormed, declcfg.Bundle{
+			Schema:     SchemaBundle,
+			Name:       pkgName + ".v" + versions[i],
```
tmshort, line 90: A bit concerned here that `i` is an index for `imageRefs`, yet it is also being used as an index for `versions`

test/operator-framework-e2e/create_fbc_helper.go
```diff
@@ -0,0 +1,155 @@
...
+			Replaces: replace,
+		})
+	}
+	return channelEntries
+}
+
+// Forms bundle schema for the FBC
+func formBundle(pkgName string, versions, imageRefs []string) []declcfg.Bundle {
+	bundleFormed := make([]declcfg.Bundle, 0, len(imageRefs))
+	for i := 0; i < len(imageRefs); i++ {
```
tmshort, line 77: `range` could probably be used here instead of a "traditional C" `for` loop

test/operator-framework-e2e/create_fbc_helper.go
```diff
@@ -0,0 +1,155 @@
...
+			Package:    pkgName,
+			Image:      imageRefs[i],
+			Properties: properties,
+		})
+	}
+	return bundleFormed
+}
+
+// Writes the formed FBC into catalog.yaml file
+func WriteFBC(fbc declcfg.DeclarativeConfig, fbcFilePath, fbcFileName string) error {
```
tmshort, line 100: In this function, the assignments to `err` could be done within the `if` statement, since it's a fairly consistent structure.
Actually, this could be done in a number of places. This is mostly a style thing, but it does appear to be idiomatic.

test/operator-framework-e2e/generate_dockerfile.go
```diff
@@ -0,0 +1,36 @@
...
+	err = t.Execute(file, struct{ YamlDir string }{yamlFolderName})
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+const dockerfileTmpl = `ADD {{.YamlDir}} /configs/{{.YamlDir}}
+`
```
tmshort, line 36: Just making sure you want a blank line here? Perhaps make it more obvious by using `\n` instead? Because, this could be taken for a typo (and use double-quotes instead of back-ticks)

test/operator-framework-e2e/operator_framework_test.go
```diff
@@ -0,0 +1,982 @@
...
+}
+
+// Checks if bundle deployment succeeded
+func checkPackageInstalled(operator *operatorv1alpha1.Operator) error {
+	bd := rukpakv1alpha1.BundleDeployment{}
+	err := c.Get(ctx, types.NamespacedName{Name: operator.Name}, &bd)
+	if err != nil {
+		return fmt.Errorf("Error retrieving the bundle deployments for the operator %v:%v", operator.Name, err)
+	}
+	if len(bd.Status.Conditions) != 2 {
```
ankitathomas, line 892: you might want to specifically look for conditions with `api.FindStatusCondition` here

ankitathomas (commented):
Great work so far!

test/operator-framework-e2e/operator_framework_test.go
```diff
@@ -0,0 +1,977 @@
...
+			installVersion: "0.1.0",
+			upgradeVersion: "0.1.1",
+		}
+		for i, b := range bundleInfo.bundles {
+			bundleInfo.bundles[i].imageRef = remoteRegistryRepo + catalogDInfo.operatorName + "-bundle:v" + b.bundleVersion
+		}
+		catalogDInfo.catalogDir = catalogDInfo.operatorName + "-catalog"
+		catalogDInfo.imageRef = remoteRegistryRepo + catalogDInfo.catalogDir + ":test"
+	})
+	When("Build and load plain+v0 bundle images into the test environment", func() {
```
joelanford, line 149: Rather than directly committing plain bundle manifests, it should be possible to generate the manifests from an SDK project with `kustomize build config/default > ${bundleRoot}/manifests/manifest.yaml`

joelanford:
Firstly, this looks great! It's awesome to have that full coast-to-coast test that starts from scratch that we've been talking about.

I looks like we cover the following steps (correct me if I'm wrong):
1. start with an empty directory
2. call `operator-sdk` to initialize and generate an operator
3. generate a bundle directory
4. build/push/kind load bundle images from the bundle directories
5. repeat steps 2-4 as necessary to get bundles for multiple operator versions
6. generate a catalog directory
7. build/push/kind load the catalog
8. create a `Catalog` CR (with `kubectl operator`)
9. create an `Operator` CR (with `kubectl operator`)
10. trigger `Operator` upgrades (with `kubectl operator`)
11. delete the `Operator` CR (with `kubectl operator`)
12. delete the `Catalog` CR (with `kubectl operator`)
13. repeat steps 2-12 for each bundle format (e.g. registry+v1 and plain+v0)

What I'm _really_ interested in is which of these steps is more difficult than one or two shell commands, because that is where I think we should focus our efforts on UX to make things easier for our users. It might be nice to organize the test such that each of these steps is a function. If the function calls a single shell command that calls one of our tools, that's the ideal. If the function is a bunch of go code, that likely means that we have a tooling gap.

Makefile
jubittajohn: Have used a different `kind` cluster name : `operator-controller-op-dev-e2e`. Is this a good name?

test/operator-framework-e2e/read_manifests.go
```diff
@@ -0,0 +1,59 @@
...
+	"sigs.k8s.io/yaml"
+)
+
+type Object struct {
+	Kind       string `yaml:"kind"`
+	APIVersion string `yaml:"apiVersion"`
+	Metadata   struct {
+		Name string `yaml:"name"`
+	} `yaml:"metadata"`
+}
```
m1kola, line 18: I think you are looking for something like `k8s.io/apimachinery/pkg/runtime/serializer`. Here is an example usage in our project:

https://github.com/operator-framework/operator-controller/blob/fa84c733ade0f71f33085a978ae0163f570d48b7/cmd/resolutioncli/input_manifests.go#L44-L45

test/operator-framework-e2e/operator_framework_test.go
```diff
@@ -0,0 +1,975 @@
...
+	cmd.Stdout = w
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("Error building Docker container image %s : %v", tag, err)
+	}
+
+	cmd = exec.Command("docker", "push", tag)
+	cmd.Stderr = w
+	cmd.Stdout = w
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("Error pushing Docker container image: %s to the registry: %v", tag, err)
```
m1kola, line 575: I'm gettign this error:
```
Error pushing Docker container image: localhost:5001/plain-catalog:test to the registry: exit status 125
```

I use `podman` with an alias from `docker` to `podman`.

I changed the local port from 5000 to 5001 due to issue with control center on macos mentioned above, but still fails.

test/operator-framework-e2e/operator_framework_test.go
```diff
@@ -0,0 +1,975 @@
...
+			var toDelete []string
+			toDelete = append(toDelete, filepath.Join(catalogDInfo.baseFolderPath, catalogDInfo.catalogDir))               // delete the FBC formed
+			toDelete = append(toDelete, filepath.Join(catalogDInfo.baseFolderPath, catalogDInfo.catalogDir+".Dockerfile")) // delete the catalog Dockerfile generated
+			err = deleteFolderFile(toDelete)
+			Expect(err).ToNot(HaveOccurred())
+		})
+	})
+})
+
+var _ = Describe("Operator Framework E2E for registry+v1 bundles", func() {
```
m1kola, line 243: It looks like this is almost identical to `Operator Framework E2E for plain bundles`. Can we have different test setup, but run the same test checks?

test/operator-framework-e2e/operator_framework_test.go
```diff
@@ -0,0 +1,975 @@
...
+			for _, b := range bundleInfo.bundles {
+				dockerContext := filepath.Join(bundleInfo.baseFolderPath, b.bInputDir)
+				dockerfilePath := filepath.Join(dockerContext, "Dockerfile")
+				err = buildPushLoadContainer(b.imageRef, dockerfilePath, dockerContext, kindServer, GinkgoWriter)
+				Expect(err).ToNot(HaveOccurred())
+			}
+		})
+	})
+	When("Create the FBC", func() {
+		It("Create a FBC", func() {
```
m1kola, line 160: Ideally specs should read as a sentance. Currently we have the following:

```
Operator Framework E2E for registry+v1 bundles when Create FBC and validate FBC Create a FBC
Operator Framework E2E for plain bundles when Build and load plain+v0 bundle images into the test environment Build the plain bundle images and load them
Operator Framework E2E for registry+v1 bundles when Generate docker file and FBC image and load the FBC image into test environment Create the docker file
Operator Framework E2E for plain bundles when Create the FBC Create a FBC
Operator Framework E2E for registry+v1 bundles when Install an operator and check if the operator operations succeed Create an operator object and install it
Operator Framework E2E for plain bundles when Install an operator and check if the operator operations succeed Create an operator object and install it
Operator Framework E2E for registry+v1 bundles when Build registry+v1 bundles with operator-sdk Generate manifests and CSV for the operator
Operator Framework E2E for registry+v1 bundles when Clearing up catalog object and other files formed for the test Clearing up data generated for the test
Operator Framework E2E for plain bundles when Build and load the FBC image into the test environment Generate the docker file, build and load FBC image
Operator Framework E2E for registry+v1 bundles when Build registry+v1 bundles with operator-sdk Generate and build registry+v1 bundle
Operator Framework E2E for plain bundles when Upgrade an operator to higher version and check if the operator operations succeed Upgrade to a higher version of the operator
Operator Framework E2E for registry+v1 bundles when Create FBC and validate FBC Validate FBC
Operator Framework E2E for registry+v1 bundles when Build registry+v1 bundles with operator-sdk Initialize new operator-sdk project and create new api and controller
Operator Framework E2E for registry+v1 bundles when Upgrade an operator to higher version and check if the operator operations succeed Upgrade to a higher version of the operator
Operator Framework E2E for registry+v1 bundles when Create a catalog object and check if the resources are created Create catalog object for the FBC and check if catalog, packages and bundle metadatas are created
Operator Framework E2E for plain bundles when Create a catalog object and check if the resources are created Create catalog object for the FBC and check if catalog, packages and bundle metadatas are created
Operator Framework E2E for plain bundles when Delete an operator Delete an operator
Operator Framework E2E for plain bundles when Clearing up catalog object and other files formed for the test Clearing up data generated for the test
Operator Framework E2E for registry+v1 bundles when An operator is deleted Delete and operator

```

m1kola (commented):
I think this is super useful. I left a few comments/suggestions.

Most of them are minor. Major themes are:

* I didn't manage to run it locally: macos uses port 5000 + something is up with pushing images to the local registry
* If I understnad correctly, `It` specs depend on each other and it might be an issue when ran out of expected order.
* There is quite a lot of code duplication in `test/operator-framework-e2e/operator_framework_test.go`. I'm wondering if we can reduce it.

test/operator-framework-e2e/operator_framework_test.go
```diff
@@ -0,0 +1,975 @@
...
+			upgradeVersion: "0.1.1",
+		}
+		for i, b := range bundleInfo.bundles {
+			bundleInfo.bundles[i].imageRef = remoteRegistryRepo + catalogDInfo.operatorName + "-bundle:v" + b.bundleVersion
+		}
+		catalogDInfo.catalogDir = catalogDInfo.operatorName + "-catalog"
+		catalogDInfo.imageRef = remoteRegistryRepo + catalogDInfo.catalogDir + ":test"
+	})
+	When("Build and load plain+v0 bundle images into the test environment", func() {
+		It("Build the plain bundle images and load them", func() {
```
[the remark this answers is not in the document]
jubittajohn: Yes. The `Describe` blocks are run in parallel and they execute the tests for `registry+v1` and `plain+v0` bundle formats. However, the each `It` block  in the test depend on the output of the previous `It` block. I observed the above behavior playing around with the test. (I could be wrong).  Saw this regarding `It ` blocks. [Ginkgo will normally run It blocks synchronously](https://pkg.go.dev/github.com/onsi/ginkgo#It:~:text=Ginkgo%20will%20normally%20run%20It%20blocks%20synchronously.).

test/operator-framework-e2e/operator_framework_test.go
jubittajohn: The test ran successfully on my macos. So I guess I am missing out something here. A local registry was used over the container registry `quay.io` for improved performance. What could be an ideal solution for the issue with the port?

m1kola: I'm on version Ventura 13.4.1 right now. It looks like it started listening on port 5000 starting from Monterey.

> What could be an ideal solution for the issue with the port?

I would change the localhost port to something else other than 5000. For example, [kind suggests](https://kind.sigs.k8s.io/docs/user/local-registry/) using port 5001. And it looks like kind changed from 5000 to 5001 precisely due to this reason:

* https://github.com/kubernetes-sigs/kind/pull/2621
* https://github.com/kubernetes/test-infra/pull/25134

jubittajohn: Updated the port to use 5001 on the host. But the port 5001 on the host maps to port 5000 on the container. This is because, the docker container uses the image `registry:2` which listens on port 5000.  This works for me. I am on version Ventura 13.4. Could you please confirm if the error is now resolved.

test/operator-framework-e2e/operator_framework_test.go
jubittajohn: The bundle structure is different for `registry+v1` and `plain+v0`. Additionally, FBC formation also differs for both as plain bundle doesn't have `opm` support yet. Because of these differences, I guess separate tests for each bundle type would be better. This helps us easily find and track errors(if they occur), as separate tests show each step clearly, instead of combining steps into functions(for running same test checks), which might make things less clear.

test/operator-framework-e2e/read_manifests.go
jubittajohn: Updated the code using the universal decoder. However, in cases where a single file contains multiple YAML documents separated by `---`, the universal decoder recognizes only the first resource. This situation is relevant,  `plain+v0` bundle generated through `kustomize`, the manifest has multiple YAML documents are combined into one file using `--- `separators. Is there a decoder solution that can effectively handle scenarios involving multiple YAML documents within a single file? This is now handled by splitting the content of the YAML file and decoding each of them using the universal decoder.

test/operator-framework-e2e/operator_framework_test.go
m1kola: > But the port 5001 on the host maps to port 5000 on the container.

@jubittajohn that's good. This is what we want. I just quickly chekced - registry is running fine now, but I'm still getting an error on an attempt to push the image:

```bash
  [FAILED] in [It] - /Users/mradchuk/workspace/src/github.com/operator-framework/operator-controller/test/operator-framework-e2e/operator_framework_test.go:318 @ 08/09/23 10:13:38.056
  << Timeline

  [FAILED] Unexpected error:
      <*errors.errorString | 0x1400014f4f0>:
      Error pushing bundle image localhost:5001/registry-operator-bundle:v0.1.0 with tag 0.1.0 : make[1]: Entering directory '/Users/mradchuk/workspace/src/github.com/operator-framework/operator-controller/test/operator-framework-e2e/registry-operator'
      make docker-push IMG=localhost:5001/registry-operator-bundle:v0.1.0
      make[2]: Entering directory '/Users/mradchuk/workspace/src/github.com/operator-framework/operator-controller/test/operator-framework-e2e/registry-operator'
      docker push localhost:5001/registry-operator-bundle:v0.1.0
      Getting image source signatures
      Copying blob sha256:dea1f3bba92bc840b14e6ae4e75f6d1538be9c7d11b13e27dadb691cd16c6a04
      Copying blob sha256:b6d433180825c5881b9565b9cbff353bace1926fe429c2cc4c67b75b203c8706
      Copying blob sha256:7e13d0a7484e3297efc0eb92610081b75a115f7a24b2ffde5e25bde7f6619634
      Error: trying to reuse blob sha256:b6d433180825c5881b9565b9cbff353bace1926fe429c2cc4c67b75b203c8706 at destination: pinging container registry localhost:5001: Get "https://localhost:5001/v2/": http: server gave HTTP response to HTTPS client
      make[2]: *** [Makefile:131: docker-push] Error 125
      make[2]: Leaving directory '/Users/mradchuk/workspace/src/github.com/operator-framework/operator-controller/test/operator-framework-e2e/registry-operator'
      make[1]: *** [Makefile:240: bundle-push] Error 2
      make[1]: Leaving directory '/Users/mradchuk/workspace/src/github.com/operator-framework/operator-controller/test/operator-framework-e2e/registry-operator'
      : exit status 2
      {
          s: "Error pushing bundle image localhost:5001/registry-operator-bundle:v0.1.0 with tag 0.1.0 : make[1]: Entering directory '/Users/mradchuk/workspace/src/github.com/operator-framework/operator-controller/test/operator-framework-e2e/registry-operator'\nmake docker-push IMG=localhost:5001/registry-operator-bundle:v0.1.0\nmake[2]: Entering directory '/Users/mradchuk/workspace/src/github.com/operator-framework/operator-controller/test/operator-framework-e2e/registry-operator'\ndocker push localhost:5001/registry-operator-bundle:v0.1.0\nGetting image source signatures\nCopying blob sha256:dea1f3bba92bc840b14e6ae4e75f6d1538be9c7d11b13e27dadb691cd16c6a04\nCopying blob sha256:b6d433180825c5881b9565b9cbff353bace1926fe429c2cc4c67b75b203c8706\nCopying blob sha256:7e13d0a7484e3297efc0eb92610081b75a115f7a24b2ffde5e25bde7f6619634\nError: trying to reuse blob sha256:b6d433180825c5881b9565b9cbff353bace1926fe429c2cc4c67b75b203c8706 at destination: pinging container registry localhost:5001: Get \"https://localhost:5001/v2/\": http: server gave HTTP response to HTTPS client\nmake[2]: *** [Makefile:131: docker-push] Error 125\nmake[2]: Leaving directory '/Users/mradchuk/workspace/src/github.com/operator-framework/operator-controller/test/operator-framework-e2e/registry-operator'\nmake[1]: *** [Makefile:240: bundle-push] Error 2\nmake[1]: Leaving directory '/Users/mradchuk/workspace/src/github.com/operator-framework/operator-controller/test/operator-framework-e2e/registry-operator'\n: exit status 2",
      }
  occurred
  In [It] at: /Users/mradchuk/workspace/src/github.com/operator-framework/operator-controller/test/operator-framework-e2e/operator_framework_test.go:318 @ 08/09/23 10:13:38.056
```

tmshort:
**I tried running this on my Ventura Mac (M1 Pro):**
* With docker installed via `brew install docker --cask`, I was able to get the tests ran, but there were timeout and others errors:
```
Summarizing 2 Failures:
  [FAIL] Operator Framework E2E for registry+v1 bundles [It] should succeed
  /Users/tshort/git/operator-controller/test/operator-framework-e2e/operator_framework_test.go:846
  [FAIL] Operator Framework E2E for plain+v0 bundles [It] should succeed
  /Users/tshort/git/operator-controller/test/operator-framework-e2e/operator_framework_test.go:840

Ran 2 of 2 Specs in 309.603 seconds
FAIL! -- 0 Passed | 2 Failed | 0 Pending | 0 Skipped
--- FAIL: TestOperatorFramework (309.60s)
```
* I couldn't get docker working via `brew install docker docker-machine virtualbox`, as it requires x86_64 (qemu would probably work, but I didn't try).
* With podman installed via `brew install podman`; it fails not finding the `docker` command. After creating a `docker` symlink to `podman`, it had the same behavior as docker, the tests ran, and encountered the same error.
(This was after a bunch of cleanup of my podman/docker config.)

**I tried running this on my F39 (x86_64):**
I have `docker` installed on here, because I was running into `podman` issues. Ran into the same errors.

Makefile
```diff
@@ -141,6 +149,27 @@ kind-load-test-artifacts: $(KIND) ## Load the e2e testdata container images into
...
+deploy-local-registry: ## Deploy local docker registry
+	$(CONTAINER_RUNTIME) run -d -p 5001:5000 --restart=always --name local-registry registry:2
+
+.PHONY: stop-local-registry
+stop-local-registry: ## Stop local registry
+	$(CONTAINER_RUNTIME) container stop local-registry
+
+.PHONY: remove-local-registry
+remove-local-registry: ## Remove local registry
+	$(CONTAINER_RUNTIME) container rm -v local-registry
```
tmshort, line 162: Having these as separate targets means two things:
They aren't cleaned up when an error occurs, which is good if you want to look at the status of an error, but bad if you want to run the tests again. You might want to force stop in the deploy-local-registry. Or possibly combine these into a single "cleanup" target to make it easier to clean up after failure,

jubittajohn: Added `cleanup-local-registry` which combines both the targets mentioned above. Need to look into force stop in deploy-local-registry if that is required to be addressed.

test/operator-framework-e2e/operator_framework_test.go
jubittajohn: @m1kola 
The issue is with using `podman`.

The error in pushing was due to `docker` being hard coded as the container runtime in the test. This is now changed to get the current container runtime using the environment variable `CONTAINER_RUNTIME`.  The default value of `CONTAINER_RUNTIME` defined in the `Makefile` is `docker`. Therefore the correct runtime has to be assigned to the variable `CONTAINER_RUNTIME` before calling the make target if `docker` is what is not being used. The test routine also assumes the runtime as `docker` if it is unable to retrieve the value of the environment variable.

But this is only a partial fix to the problem. With this change the test for `plain+v0` bundles will pass but for `registry+v1` will fail. This is because `registry+v1` uses `operator-sdk` support. Thus to mimic the user experience, the targets `bundle-build` and `bundle-push`  from the generated Makefile by `operator-sdk` tool, which has `docker` being hard coded as the container runtime, is used to build and push the bundle images. This could be marked as an issue and addressed when hard coding `docker` as container runtime is addressed by `operator-sdk`.

In conclusion, the current test will only work for `docker`.

jubittajohn:
> **I tried running this on my Ventura Mac (M1 Pro):**
> 
> * With docker installed via `brew install docker --cask`, I was able to get the tests ran, but there were timeout and others errors:
> 
> ```
> Summarizing 2 Failures:
>   [FAIL] Operator Framework E2E for registry+v1 bundles [It] should succeed
>   /Users/tshort/git/operator-controller/test/operator-framework-e2e/operator_framework_test.go:846
>   [FAIL] Operator Framework E2E for plain+v0 bundles [It] should succeed
>   /Users/tshort/git/operator-controller/test/operator-framework-e2e/operator_framework_test.go:840
> 
> Ran 2 of 2 Specs in 309.603 seconds
> FAIL! -- 0 Passed | 2 Failed | 0 Pending | 0 Skipped
> --- FAIL: TestOperatorFramework (309.60s)
> ```
> 
> * I couldn't get docker working via `brew install docker docker-machine virtualbox`, as it requires x86_64 (qemu would probably work, but I didn't try).
> * With podman installed via `brew install podman`; it fails not finding the `docker` command. After creating a `docker` symlink to `podman`, it had the same behavior as docker, the tests ran, and encountered the same error.
>   (This was after a bunch of cleanup of my podman/docker config.)
> 
> **I tried running this on my F39 (x86_64):**
> 
> * I have `docker` installed on here, because I was running into `podman` issues. Ran into the same errors.

The podman issue is due to reasons specified in the (comment)[https://github.com/operator-framework/operator-controller/pull/286#discussion_r1289567123]. The test case failure is not being reproduced consistently. Looking into the issue with the test failures.

test/operator-framework-e2e/operator_framework_test.go
m1kola: I have a symlink from `docker` to `podman` and Podman's CLI is compatible with docker. So `make operator-developer-e2e CONTAINER_RUNTIME=podman` and `make operator-developer-e2e` give me the same results.

It has to be something else.

I don't know how often (if at all) we will be running this test locally. It might be not worth spinning wheels on this (at least in context of this PR).

test/operator-framework-e2e/read_manifests.go
m1kola: Good point!

I think `YAMLToJSONDecoder` from `k8s.io/apimachinery/pkg/util/yaml` should do the trick.

I think you should be able to get kind, api version and name if you decode into `Unstructured` from `k8s.io/apimachinery/pkg/apis/meta/v1/unstructured`.

m1kola (commented):
CI if failing at the moment:

```
Status conditions for the operator registry-operator for the version 0.1.1 is not as expected:Expected status: True, but got: Unknown
```

Once this failure is deailt with - I think we will be in a good place to get it merged. After that we will be able to iterate in smaller PRs.

I [wasn't able to run](https://github.com/operator-framework/operator-controller/pull/286#discussion_r1289902366) this locally on ARM mac with podman, but I'm not sure whether we want to hold this PR because of this.

test/operator-framework-e2e/operator_framework_test.go
```diff
@@ -0,0 +1,1061 @@
...
+		imageRefsBundleVersions := make(map[string]string)
+		for _, b := range bundleInfo.bundles {
+			imageRefsBundleVersions[b.imageRef] = b.bundleVersion
+		}
+		err = genPlainCatalogDirectory(catalogDInfo, imageRefsBundleVersions)
+		Expect(err).ToNot(HaveOccurred())
+
+		By("building/pushing/kind loading the catalog images")
+		dockerContext := catalogDInfo.baseFolderPath
+		dockerfilePath := filepath.Join(dockerContext, catalogDInfo.catalogDir+".Dockerfile")
```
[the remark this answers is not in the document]
jubittajohn: If the catalog name is `test-catalog`, the expectation here is to  form the `Dockerfile` name as `test-catalog.Dockerfile`. Hence concatenation is used.  This is to distinguish between various Dockerfiles formed. I think we can't achieve this if we use `filepath.Join` as suggested above.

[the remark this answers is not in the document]
m1kola: Ahhh, I see now. I was under the impression that it will be something like: `/somePathToDockerContext/catalogDir/.Dockerfile`, but you want it to be `/somePathToDockerContext/catalog-name.Dockerfile`. Got it now!

I think I just got confused by field name `catalogDir`: I asumed that this is not a catalog name, but part of path to catalog directory. In this case, as far as I understand - it is both.

I think to reduce confusion a bit we can use `fmt.Sprintf("%s.Dockerfile", catalogDInfo.catalogDir)` instead of `catalogDInfo.catalogDir+".Dockerfile")`.

m1kola (approved):
Great work! I think we are in the state where we can merge and we will be able to iterate on top of what you built (e.g. fixing issues with local podman on mac os, etc).

Thank you for all your contributions, best of luck with your studies and I hope to see you around again at some point :)

dtfranz:
/lgtm
Great job Jubitta! We really appreciate your work here!

dtfranz (approved):

PR #299, opened 2023-07-27
Tidies up `Makefile`: fixes default target, remove unused ones and makes phony targets to by be truly phony

m1kola:
# Description

* Fixes default target. It looks like target `all` was previously define first to implicitly set `build` to be default target, but since introduction `include` above the `all` target - now `BINGO` is default target. This change explicitly sets `build` to be default.
* Sets several targets as phony as they are not targeting a file and moves `.PHONY` closer to target definition for better visibility.
* Removes unused `wait` target

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,8 +41,7 @@ SHELL = /usr/bin/env bash -o pipefail
 # Disable -j flag for make
 .NOTPARALLEL:
 
-.PHONY: all
-all: build
+.DEFAULT_GOAL := build
 
 ##@ General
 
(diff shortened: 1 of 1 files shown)


ncdc:
re the unused `wait` target - do we know if anyone was calling `make wait` as part of their workflow?

m1kola:
@ncdc I believe this [was introduced](https://github.com/operator-framework/operator-controller/commit/47485a7e8e678ff7f922321d014e8ae1adeef352) to by used by `run` target. At some point [`wait` was removed](https://github.com/operator-framework/operator-controller/commit/a553313d8d1a6470905b74dd6aef533b39d68087#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L103) from `run`. So `wait` seems to be obsolete.

---

I rebased the PR to fix conflicts.

ncdc (approved):

PR #306, opened 2023-08-01
Adds go module caching for bingo modules

m1kola:
# Description

We introduced [bingo](https://github.com/bwplotka/bingo) in OLMv1 repos for managing tools and it created a bunch of modules in the `.bingo` directory in each repo. It works well, but on every job we download these dependencies from internet instead of using cache. This happens becuase setup-go github action only looks into project root for `go.sum` and builds cache based on it.

`cache-dependency-path` needs to be added to all OLMv1 repos and it should speed up our builds (we've seen decrease in build time of ~3 minutes in https://github.com/operator-framework/olm-docs).

Relevant issue: #323

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -24,6 +24,9 @@ jobs:
           go.sum
           .bingo/**.sum
 
+    - name: Install all tools
+      run: make install-tools
+
     - name: Run e2e tests
       run: |
         # By default make stops building on first non-zero exit code which
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -20,6 +20,9 @@ jobs:
     - uses: actions/setup-go@v4
       with:
         go-version-file: go.mod
+        cache-dependency-path: |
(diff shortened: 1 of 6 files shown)


joelanford:
Do we need to install all of the tools for all of the workflows? Or is it enough to let `make` targets eventually install just the tools that are required for those workflows?

ncdc:
Lazy is fine

m1kola:
> Do we need to install all of the tools for all of the workflows? Or is it enough to let `make` targets eventually install just the tools that are required for those workflows?

> Lazy is fine

It is not that simple :) I described the issue [here](https://github.com/operator-framework/operator-controller/issues/323#issuecomment-1667920491).

I've decided to put this on hold for now and prioritise other work for now, but I'm hoping to come back to this.

m1kola:
I'll close it for now to avoid confusion. It is not working as expected. More details [here](https://github.com/operator-framework/operator-controller/issues/323#issuecomment-1667920491).

PR #307, opened 2023-08-01
Switch to using tilt-support repo

ncdc:
# Description

Requires https://github.com/operator-framework/tilt-support/pull/1

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -1,66 +1,23 @@
-# This loads a helper function that isn't part of core Tilt that simplifies restarting the process in the container
-# when files changes.
-load('ext://restart_process', 'docker_build_with_restart')
+if not os.path.exists('../tilt-support'):
+    fail('Please clone https://github.com/operator-framework/tilt-support to ../tilt-support')
 
-# Treat the main binary as a local resource, so we can automatically rebuild it when any of the deps change. This
-# builds it locally, targeting linux, so it can run in a linux container.
-local_resource(
-    'manager_binary',
-    cmd='''
-mkdir -p .tiltbuild/bin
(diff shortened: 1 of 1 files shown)


Tiltfile
```diff
@@ -1,66 +1,23 @@
-# This loads a helper function that isn't part of core Tilt that simplifies restarting the process in the container
-# when files changes.
-load('ext://restart_process', 'docker_build_with_restart')
+if not os.path.exists('../tilt-support'):
```
tmshort, line 1: I'm very leery about going outside the repo.

ncdc: This is for local development and iteration. It's pretty standard.

ncdc: This doesn't clone the tilt-support repo. That's something you have to do manually. It's definitely not forcing arbitrary code from the internet on you 😄 

An alternative is to consider `tilt-support` a Tilt extension:

```
v1alpha1.extension_repo(name='tilt-support', url='https://github.com/operator-framework/tilt-support')
v1alpha1.extension(name='tilt-support', repo_name='tilt-support', repo_path='.')
load('ext://tilt-support', 'deploy_repo')
```

Both would load the same code

tmshort: I just said I'm "leery". I haven't seen many (if any?) projects that look outside their repository unless explicitly configured to do so by the user (e.g. `--with-software=/path/to/software`). Then again, I'm also leery about pulling random software versions from the 'net (given my corporate security background).
That said, if it's just looking for something, then it's not so bad.

ncdc: Ok, cool. Yeah this is by far the easiest way.

m1kola (approved):

perdasilva (approved):
lgtm

PR #326, opened 2023-08-08
Fixes replace based upgrades

### LINKED ISSUE #321: `entity for package "project-quay" not found in solution` when trying to upgrade to the next version using replaces
I'm getting `entity for package "project-quay" not found in solution` when trying to upgrade to the next version. 

`project-quay` uses `replaces` to speficy upgrade edges and support for this was added #275. I see that PR has an [e2e test confirming an upgrade](https://github.com/operator-framework/operator-controller/blob/fa84c733ade0f71f33085a978ae0163f570d48b7/test/e2e/install_test.go#L171-L213) so it might be that I'm missing something.

## Steps to reproduce

1. Run the operators from master:
  ```bash
  make run
  ```
1. Create a catalog:
  ```bash
  kubectl apply -f config/samples/catalogd_operatorcatalog.yaml
  ```
1. Create an operator
  ```bash
  kubectl apply -f - <<EOF
  kind: Operator
  metadata:
    labels:
      app.kubernetes.io/name: operator
      app.kubernetes.io/instance: operator-sample
      app.kubernetes.io/part-of: operator-controller
      app.kubernetes.io/managed-by: kustomize
      app.kubernetes.io/created-by: operator-controller
    name: operator-sample
  spec:
    packageName: project-quay
    channel: stable-3.8
    version: 3.8.1
  EOF
  ```
1. Wait for the operator to be installed:
  ```bash
  kubectl wait operators operator-sample --for condition=Installed
  ```
1. Edit operator to have a next valid version (in this specific example - 3.8.2 replaces 3.8.1):
  ```bash
  kubectl patch operator operator-sample --type json -p '[{"op": "replace", "path": "/spec/version", "value": "3.8.2"}]'
  ```
1. Observe conditions:
  ```bash
  kubectl get operator operator-sample -o json | jq ".status"
  ```
  Output:
  
  ```json
  {
    "conditions": [
      {
        "lastTransitionTime": "2023-08-04T14:24:02Z",
        "message": "installation has not been attempted as resolution failed",
        "observedGeneration": 13,
        "reason": "InstallationStatusUnknown",
        "status": "Unknown",
        "type": "Installed"
      },
      {
        "lastTransitionTime": "2023-08-04T14:24:02Z",
        "message": "entity for package \"project-quay\" not found in solution",
        "observedGeneration": 13,
        "reason": "ResolutionFailed",
        "status": "False",
        "type": "Resolved"
      }
    ]
  }
  ```

## Expected results

Operator is correctly resolved and installed

m1kola:
# Description

Fixes replace based upgrades.

Closes #321

## Reviewer Checklist

- [ ] API Go Documentation


internal/resolution/entitysources/catalogdsource.go
```diff
@@ -110,9 +110,10 @@ func getEntities(ctx context.Context, client client.Client) (input.EntityList, e
 				if catalogScopedEntryName == bundle.Name {
 					channelValue, _ := json.Marshal(property.Channel{ChannelName: ch.Name, Priority: 0})
 					props[property.TypeChannel] = string(channelValue)
-					// TODO(jmprusi): Add the proper PropertyType for this
-					replacesValue, _ := json.Marshal(replacesProperty{Replaces: b.Replaces})
-					props["olm.replaces"] = string(replacesValue)
+					replacesValue, _ := json.Marshal(replacesProperty{
+						Replaces: fmt.Sprintf("%s-%s%s%s", bundle.Spec.Catalog.Name, b.Replaces, bundle.Spec.Package, ch.Name),
```
m1kola, line 114: So the idea is that: in the entity properties we have formated replaces (vs raw data from catalogd objects) and don't have to format it in `internal/resolution/variablesources/installed_package.go`: we just filter entity source by id of the currently installed entity.

However I'm not entirely sure whether this is what we want to do. This limits replaces to a specific catalog and channel.

internal/resolution/entities/bundle_entity.go
```diff
@@ -140,7 +138,7 @@ func (b *BundleEntity) loadReplaces() error {
 	b.mu.Lock()
 	defer b.mu.Unlock()
 	if b.replaces == nil {
-		replaces, err := loadFromEntity[Replaces](b.Entity, "olm.replaces", optional)
+		replaces, err := loadFromEntity[Replaces](b.Entity, PropertyBundleReplaces, optional)
```
perdasilva, line 141: thank you!

internal/resolution/entitysources/catalogdsource.go
awgreene: In OLM V0, which I believe is be the primary usecase for replace/skips/skipRange based upgrades, we supported cross channel updates. I suspect that we'll want to support cross channel upgrades here as well.

m1kola: Yeah, and for [semver upgrade support](https://docs.google.com/document/d/1KIhV8oqr22Bp8lf370Aej2n-3uW_xGcTptONHHPRVYA/edit) we also want:

* The bundle resolved for an operator during an upgrade can come from a different channel or catalog than the currently installed bundle. (i.e. successors can come from any channel and any catalog)
* The bundle resolved for an operator during an upgrade cannot come from a different package than the currently installed bundle.

So I'm looking into other options to solve this.

ankitathomas: I think it may be less restricting to have channel information be a new constraint rather than included into the replaces field keeping cross channel updates in mind, similar to what the upgrade support brief mentions.

m1kola:
Ok, here what I did:

* I added a new property into `CatalogdEntitySource`: which contains a name of the current bundle and (if present in catalog) name of the bundle it replaces. Currently in the `main` we only have replaces and no current bundle name.
* Updated `InstalledPackageVariableSource` and `Replaces` predicate to use the new prop
* Updated a bunch of tests (mostly `Entity` mocks to include new prop)

For now I'm trying to make minimal changes to fix the issue, but I have a feeling that things will change internall once we introduce [semver based upgrades](https://docs.google.com/document/d/1KIhV8oqr22Bp8lf370Aej2n-3uW_xGcTptONHHPRVYA/edit).

internal/resolution/entities/bundle_entity_test.go
```diff
@@ -268,21 +268,21 @@ var _ = Describe("BundleEntity", func() {
 	Describe("Replaces", func() {
 		It("should return the replaces property if present", func() {
```
perdasilva, line 269: nit: should we rephrase the test description?

perdasilva, line 269:
```suggestion
		It("should return the channel entry property if present", func() {
```

internal/resolution/entities/bundle_entity_test.go
```diff
@@ -268,21 +268,21 @@ var _ = Describe("BundleEntity", func() {
 	Describe("Replaces", func() {
```
perdasilva, line 268:
```suggestion
	Describe("ChannelEntry", func() {
```

perdasilva (approved):
lgtm - thanks for this!! We're having some discussions around channels atm, so it _could_ change, but getting this stable rn is super important!

PR #343, opened 2023-08-21
Switch to catalogd's `CatalogMetadata` APIs

### LINKED ISSUE #342: Switch to catalogd's `CatalogMetadata` APIs
Since #334 we have catalogd's `CatalogMetadata` APIs available for operator-controller.

As mentioned in #334:

> The `CatalogMetadata` API contains (almost) all of the FBC that the eventual catalogd file server will, so it is a good step forward, and is something that folks implementing some of the existing RFCs can use _right now_. This will ease an eventual migration to the catalogd file server.

As part of this issue we want to switch existing code to use `CatalogMetadata` and address the following to do:

https://github.com/operator-framework/operator-controller/blob/fd7cfb57680c3949d7317a4ad4c33dc7a42f09d5/cmd/resolutioncli/entity_source.go#L120-L125

But instead of converting declcfg into CRDs as initially suggested in the TODO we will operate on declcfg in both operator-controller and resolution CLI.

It will make supporting resolution CLI a bit easier.

m1kola:
# Description

Closes #342

This is a minimal set of changes required to switch from `BundleMetadata` API to `CatalogMetadata` and remove code duplication between resolution CLI and operator controller itself.

I expect more refactoring to be done as part of #347 where we will be removing entities/entity sources from operator controller and switching to `VariableSources`. E.g. we might want to create a wrapper for `declcfg` structs which makes it easier to traverse catalog (e.g. package might contain refs to channels and channels might contain refs to bundles). But this to be defined separately and out of scope for this PR.

The main goal of this PR is to make removal of entities and entity sources easier.

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/resolution/entitysources/catalogdsource.go
+++ b/internal/resolution/entitysources/catalogdsource.go
@@ -8,6 +8,7 @@ import (
 	catalogd "github.com/operator-framework/catalogd/api/core/v1alpha1"
 	"github.com/operator-framework/deppy/pkg/deppy"
 	"github.com/operator-framework/deppy/pkg/deppy/input"
+	"github.com/operator-framework/operator-registry/alpha/declcfg"
 	"github.com/operator-framework/operator-registry/alpha/property"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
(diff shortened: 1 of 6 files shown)


m1kola (commented):
Codecov is reporting decrease in coverage becasue there is a bunch of new `if err != nil { return err}` cases. Our existing tests do not cover failure scenarios at the moment.

I'm not adding any new tests for these code branches because this all will be changed as part of #347.

internal/resolution/entitysources/catalogdsource.go
```diff
@@ -71,70 +72,117 @@ func (es *CatalogdEntitySource) Iterate(ctx context.Context, fn input.IteratorFu
...
+
+	return allEntitiesList, nil
+}
+
+func MetadataToEntities(catalogName string, channels []declcfg.Channel, bundles []declcfg.Bundle) (input.EntityList, error) {
+	entityList := input.EntityList{}
+
+	bundlesMap := map[string]*declcfg.Bundle{}
+	for i := range bundles {
+		bundlesMap[bundles[i].Name] = &bundles[i]
```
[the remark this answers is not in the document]
m1kola: Good point! Should probably prefix map keys with a package name.

internal/resolution/entitysources/catalogdsource.go
```diff
@@ -71,70 +72,117 @@ func (es *CatalogdEntitySource) Iterate(ctx context.Context, fn input.IteratorFu
...
-			switch prop.Type {
-			case property.TypePackage:
-				// this is already a json marshalled object, so it doesn't need to be marshalled
-				// like the other ones
-				props[property.TypePackage] = string(prop.Value)
-			case entities.PropertyBundleMediaType:
-				props[entities.PropertyBundleMediaType] = string(prop.Value)
-			}
+	for _, catalog := range catalogList.Items {
+		channels, bundles, err := fetchCatalogMetadata(ctx, cl, catalog.Name)
```
everettraven, line 83: Do we need to fetch and loop through all the catalogs here? If we want to ensure we fetch _all_ `CatalogMetadata` for _all_ `Catalog` resources I think we can simplify this to listing all `CatalogMetadata` resources and then processing each one.

internal/resolution/entitysources/catalogdsource.go
```diff
@@ -71,70 +72,117 @@ func (es *CatalogdEntitySource) Iterate(ctx context.Context, fn input.IteratorFu
...
-	if err := client.List(ctx, &packageMetdatas); err != nil {
-		return catalogd.BundleMetadataList{}, nil, err
+func fetchCatalogMetadata(ctx context.Context, cl client.Client, catalogName string) ([]declcfg.Channel, []declcfg.Bundle, error) {
+	channels, err := fetchCatalogMetadataByScheme[declcfg.Channel](ctx, cl, declcfg.SchemaChannel, catalogName)
+	if err != nil {
+		return nil, nil, err
+	}
+	bundles, err := fetchCatalogMetadataByScheme[declcfg.Bundle](ctx, cl, declcfg.SchemaBundle, catalogName)
+	if err != nil {
+		return nil, nil, err
```
everettraven, line 160: Do we only care about channels and bundles? What about packages?

internal/resolution/entitysources/catalogdsource.go
```diff
@@ -71,70 +72,117 @@ func (es *CatalogdEntitySource) Iterate(ctx context.Context, fn input.IteratorFu
...
-					}
-					entityList = append(entityList, entity)
+
+		allEntitiesList = append(allEntitiesList, catalogEntitiesList...)
+	}
+
+	return allEntitiesList, nil
+}
+
+func MetadataToEntities(catalogName string, channels []declcfg.Channel, bundles []declcfg.Bundle) (input.EntityList, error) {
```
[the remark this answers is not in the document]
m1kola: Could you please explain your concerns?

I'm concerned about `CatalogdEntitySource` since it seems to call `getEntities` on each `Filter` call. I decided not to touch it since we will be removing entity sources in #347.

I think as part of #347 we should think about some cache which will be valid for the duration of one resolution.

In the CLI we cache entites into the enttiy source property:

https://github.com/operator-framework/operator-controller/blob/b568602e9daa7df5970090f8f5b791f36f30ddb5/cmd/resolutioncli/entity_source.go#L92-L109

In operator controlle we can not do the same since entity source is resused between reconciliations.

internal/resolution/entitysources/catalogdsource.go
```diff
@@ -71,70 +72,117 @@ func (es *CatalogdEntitySource) Iterate(ctx context.Context, fn input.IteratorFu
...
+				Name:     bundle.Name,
+				Replaces: chEntry.Replaces,
+			})
+			props[entities.PropertyBundleChannelEntry] = string(replacesValue)
+
+			catalogScopedEntryName := fmt.Sprintf("%s-%s", catalogName, bundle.Name)
+			entity := input.Entity{
+				ID:         deppy.IdentifierFromString(fmt.Sprintf("%s%s%s", catalogScopedEntryName, bundle.Package, ch.Name)),
+				Properties: props,
+			}
```
everettraven, line 145: The construction of this entity feels quite complex. Is there a particular reason it needs to be? I'm wondering if there is a better way to produce these entities such that there is one for each package, channel, and bundle without having to do this channel+bundle combination stuff.

i.e could the entity creation loop be simplified to something like:
```go
entityList := []input.Entity{}
for _, catalogMetadata := range catalogMetadataList {
  var entity input.Entity
  switch catalogMetadata.Schema {
    case "olm.package":
      //  doing some psuedocode here
      pack := CatalogMetadataToPackage(catalogMetadata)
      entity = input.Entity{
        ID: deppy.IdentifierFromString(IdForPackage(pack)),
        Properties: PropertiesForPackage(pack),
      }
    ...
  }
}
```

everettraven: I'm also not very knowledgeable on the resolution process and what the entity requirements are at this point so please do let me know if something like ^ is just not possible at the moment

internal/resolution/entitysources/catalogdsource.go
m1kola: We only need package name to create an entity and both channels and bundles have it. So for now channels and bundles are enough. But we might need to fetch packages at some point.

internal/resolution/entitysources/catalogdsource.go
everettraven: That's fair. My bad for not properly reading the description and jumping straight to the review :sweat_smile:

internal/resolution/entitysources/catalogdsource.go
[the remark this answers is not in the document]
m1kola: Updated

dtfranz (approved):
This /lgtm, I've taken note of the concerns and hopefully I can help take care of them during the next efforts on the refactor. Thanks Mikalai!

internal/resolution/entitysources/catalogdsource.go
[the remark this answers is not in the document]
joelanford: > fetching the data from API server on each reconcile & multiple times during one resolution.

Is that true? I thought we had informer caches setup for the catalogd data such that `cl.Get` and `cl.List` would only ever hit the cache (and never the real apiserver)?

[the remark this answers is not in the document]
m1kola: Ah, you might be right! I totally forgot about informers.

Issue #347, opened 2023-08-21
Update Operator-Controller to use VariableSources


awgreene:
Deppy [recently](https://github.com/operator-framework/deppy/pull/108) removed EntitySources in favor of VariableSources, the operator-controller should be updated to rely on the new VariableSource.

PR #352, opened 2023-08-22
fix: check for unsat resolutions in Reconcile

### LINKED ISSUE #330: Reconciler does not correctly handle UNSAT resolutions
Related to https://github.com/operator-framework/deppy/issues/139 and possibly #329 

In `Reconcile` we:
1. Check to see if `Solve` returned an error.
	- This currently can happen if a variable source returns an error when getting variables.
	- Notably, solve does NOT return an error for UNSAT. This error means "was not able to get variables or setup the solver"
2. Look for the matching variable in the solution

We skip the step of checking if the solution was SAT or UNSAT. UNSAT currently manifests as an error looking up the bundle entity from the solution. This "works", but it means users don't see the UNSAT reason that comes from `solution.Error()`

We should check `solution.Error()`, and handle it accordingly.

joelanford:
# Description

Closes #330 

## Reviewer Checklist

- [ ] API Go Documentation


Makefile
```diff
@@ -113,7 +113,7 @@ test-unit: $(SETUP_ENVTEST) ## Run the unit tests
 e2e: KIND_CLUSTER_NAME=operator-controller-e2e
 e2e: KUSTOMIZE_BUILD_DIR=config/e2e
 e2e: GO_BUILD_FLAGS=-cover
-e2e: run kind-load-test-artifacts test-e2e e2e-coverage kind-cluster-cleanup ## Run e2e test suite on local kind cluster
+e2e: kind-cluster-cleanup run kind-load-test-artifacts test-e2e e2e-coverage kind-cluster-cleanup ## Run e2e test suite on local kind cluster
```
joelanford, line 116: This is a preemptive cleanup of a pre-existing e2e cluster. I was running into the slight annoyance of the e2e failing, the last `kind-cluster-cleanup` not running, and then having to manually delete the cluster to be able to run the e2e again.

internal/controllers/operator_controller.go
```diff
@@ -136,6 +140,21 @@ func (r *OperatorReconciler) reconcile(ctx context.Context, op *operatorsv1alpha
 		return ctrl.Result{}, err
 	}
 
+	// TODO: Checking for unsat is awkward using the current version of deppy.
+	//    This awkwardness has been fixed in an unreleased version of deppy.
+	//    When there is a new minor release of deppy, we can revisit this and
+	//    simplify this to a normal error check.
+	//    See https://github.com/operator-framework/deppy/issues/139.
```
joelanford, line 147: On the working group call today, I mentioned that I would try to go ahead and bump deppy to the commit containing this fix.

Unfortunately that commit brings in some breaking changes with deppy's removal of the EntitySource API, so we'll have to wait until operator-controller accounts for that change. (see https://github.com/operator-framework/operator-controller/issues/347)

Makefile
joelanford: I see @tmshort has a more wide-ranging fix in #353, so removing the Makefile changes from this PR.

tmshort: Adding `kind-cluster-cleanup` early in the dependency line won't work; as `make` will have recognized that target as already completed by the time it gets to the second `kind-cluster-cleanup`. So it won't be run, and it will never clean up!

dtfranz (approved):

m1kola (approved):

Issue #346, opened 2023-08-21
Version Ranges: Introduce Masterminds/semver


tmshort:
### Summary
The **Masterminds/semver** will be used to parse **.spec.version** field of the Operator CR.The regular expression contained in the CRD that clients use to validate input before sending to the server must be updated to support version ranges. Due to the complexity and imprecision of the hyphen (-) operator, it will not be included.

* Update the validator for **.spec.version** to use **Masterminds/semver**.
* Update the variable source code to use **Masterminds/semver** APIs
* Update other code locations as necessary
* Use the alias `mmsemver` (may change in PR) to refer to **Masterminds/sever** to distinguish it from **blang/semver**. 
* Validate existing unit & e2e tests continue to work.
* Add unit tests to validate ranges, both positive and negative; this should be done _without_ a client (i.e. ginkgo). If necessary, remove the regex in the CRD definition to allow ranges to be used.

### Reference
[RFC: OLMv1 Version Ranges](https://docs.google.com/document/d/1Cf4vw6kdadNp3bHFxh5AMcq1xq0vsEUgC-JNEaVtU6g/edit?usp=sharing)

Update the Operator CR to support version ranges. There are no direct API changes.

### Related GitHub Issues

#345 

### Dependencies

https://github.com/operator-framework/operator-controller/issues/344

### Updates and Amendments

None.

tmshort:
> * update

sion

tmshort:
Fixed by #374 (apparently, you can't do `Fixes num and num`, and have GH detect that.

joelanford:
> apparently, you can't do Fixes num and num, and have GH detect that

Yeah separate lines should do it though

```md
Fixes #345
Fixes #346
```

PR #374, opened 2023-08-29
Introduce Masterminds/semver 

### LINKED ISSUE #345: Version Ranges: Update CRD Regex
### Summary
The regular expression contained in the CRD that clients use to validate input before sending to the server must be updated to support version ranges. Due to the complexity and imprecision of the hyphen (-) operator, it will not be included.

* Update the regex to support the *Masterminds/semver* syntax (except for hyphen ranges).
* Add unit tests to validate ranges, both positive and negative. This must use a client program (e.g. `oc` or `kubectl`).

### Reference

[RFC: OLMv1 Version Ranges](https://docs.google.com/document/d/1Cf4vw6kdadNp3bHFxh5AMcq1xq0vsEUgC-JNEaVtU6g/edit?usp=sharing)

Update the Operator CR to support version ranges. There are no direct API changes.

### Related GitHub Issues

See dependencies.

### Dependencies

#344 
#346 may remove the regex

### Updates and Amendments

None.

tmshort:
# Description

Add support for Masterminds/semver for .spec.Version
This is a bit more entangled into the code than I expected,
most instances of bsemver were replaced.

Fixes #345 and #346.

Includes the regex commit.

## Reviewer Checklist

- [ ] API Go Documentation


internal/resolution/variablesources/required_package_test.go
```diff
@@ -80,7 +80,7 @@ var _ = Describe("RequiredPackageVariableSource", func() {
 	It("should filter by version range", func() {
 		// recreate source with version range option
 		var err error
-		rpvs, err = variablesources.NewRequiredPackageVariableSource(packageName, variablesources.InVersionRange(">=1.0.0 !2.0.0 <3.0.0"))
+		rpvs, err = variablesources.NewRequiredPackageVariableSource(packageName, variablesources.InVersionRange(">=1.0.0 !=2.0.0 <3.0.0"))
```
tmshort, line 83: Masterminds/semver does not  support the `!` operator, instead it supports `!=`. This was a bit unexpected.

api/v1alpha1/operator_types.go
```diff
@@ -29,7 +29,7 @@ type OperatorSpec struct {
 	PackageName string `json:"packageName"`
 
 	//+kubebuilder:validation:MaxLength:=64
-	//+kubebuilder:validation:Pattern=^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(-(0|[1-9]\d*|[0-9]*[a-zA-Z-][0-9a-zA-Z-]*)(\.(0|[1-9]\d*|[0-9]*[a-zA-Z-][0-9a-zA-Z-]*))*)?(\+([0-9a-zA-Z-]+(\.[0-9a-zA-Z-]+)*))?$
+	//+kubebuilder:validation:Pattern=`^(\s*(=||!=|>|<|>=|=>|<=|=<|~|~>|\^)\s*(v?(0|[1-9]\d*|[x|X|\*])(\.(0|[1-9]\d*|x|X|\*]))?(\.(0|[1-9]\d*|x|X|\*))?(-([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?(\+([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?)\s*)((?:\s+|,\s*|\s*\|\|\s*)(=||!=|>|<|>=|=>|<=|=<|~|~>|\^)\s*(v?(0|[1-9]\d*|x|X|\*])(\.(0|[1-9]\d*|x|X|\*))?(\.(0|[1-9]\d*|x|X|\*]))?(-([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?(\+([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?)\s*)*$`
```
ncdc, line 32: Does this remove any previously-valid values? If so, we probably need to bump the API to v1alpha2.

internal/resolution/variablesources/required_package_test.go
ncdc: What is the blast radius of this impact? Operator spec.version field? FBC metadata?

ncdc:
@tmshort we don't have Prow on this repo; if you'd like to "hold" the PR, I'd recommend converting it to draft.

api/v1alpha1/operator_types.go
tmshort: No it doesn't. It expands the  *.spec.Version* to support ranges, where previously it supported only a single semver. This is directly from #359

internal/resolution/variablesources/required_package_test.go
tmshort: (Note this is test code.)
It _should_ only impact the **operator.spec.Version**, which previously only supported a single semver, not a range. But it is something to be aware of. Possibly something to port to Masterminds/semver if they're interested.

internal/resolution/variablesources/bundles_and_dependencies.go
```diff
@@ -87,7 +87,7 @@ func (b *BundlesAndDepsVariableSource) getEntityDependencies(ctx context.Context
 	// todo(perdasilva): disambiguate between not found and actual errors
 	requiredPackages, _ := bundleEntity.RequiredPackages()
 	for _, requiredPackage := range requiredPackages {
-		semverRange, err := bsemver.ParseRange(requiredPackage.VersionRange)
+		semverRange, err := mmsemver.NewConstraint(requiredPackage.VersionRange)
```
joelanford, line 90: Technically, this should remain `blang` if we don't want to change the meaning of the `olm.package.required` constraints.

tmshort: I can look at it; changes were only made if they propagated from `validation.go`s changes. This likely means that it would require intermixing **Masterminds** and **blang** code here.

joelanford: Yeah, looks like you'll need to convert a blang version to a masterminds verison somewhere along the line.

It looks like bundles_and_dependencies.go and required_package.go _both_ call `predicates.InVersionRange`, so it seems like we'll have to break that in two. Perhaps:
- `predicates.InMastermindsVersionRange`
- `predicates.InBlangVersionRange`

And then `predicates.InMasterMindsVersionRange` converts the bundle entity blang version to a masterminds version to check against the mastermind constraint?

tmshort: `blang.Range` is a function and cannot be converted into a `MasterMind.Constraints` (which is a struct). I'm woking out a couple different methods; none of which are pretty...

internal/controllers/admission_test.go
```diff
@@ -61,6 +61,15 @@ var _ = Describe("Operator Spec Validations", func() {
...
 			".1.2.3",
+			"<<1.2.3",
+			">>1.2.3",
+			">~1.2.3",
+			"==1.2.3",
+			"=!1.2.3",
+			"1.Y",
+			">1.2.3 && <2.3.4",
+			">1.2.3;<2.3.4",
+			"1.2.3 - 2.3.4",
```
grokspawn, line 72: worth including the blang '!=' equivalent here?  (e.g. "!1.2.3")

tmshort:
ping @grokspawn @joelanford this should be ready now

internal/resolution/entities/bundle_entity.go
```diff
@@ -87,13 +88,20 @@ func (b *BundleEntity) PackageName() (string, error) {
...
 		return nil, err
 	}
 	return b.semVersion, nil
 }
 
+func (b *BundleEntity) VersionMasterminds() (*mmsemver.Version, error) {
+	if err := b.loadPackage(); err != nil {
+		return nil, err
+	}
+	return mmsemver.NewVersion(b.bundlePackage.Version)
```
joelanford, line 102: Nit: the asymmetry of `VersionBlang()` stuff being parsed during `b.loadPackage()` and the masterminds version being parsed here every time is maybe a bit unexpected and feels a bit off?

Other potential options:
- add a new sibling field to `b.semVersion`, parse both during `loadPackage()`
- parse into a blang version and then just build a masterminds.Version struct (either here or in `loadPackage`) using the blang version fields (this might be nice because its one fewer error to handle)
- have this `Version()` method just return a string and expect that caller to do the parsing.

joelanford: Another option: do the (option 2) conversion from blang version to masterminds version in the `InMastermindsVersionRange()` function.

tmshort: TL;DR: this is what happens when you try to optimize code coverage!

1. I had created a new field, but all errors occurred during blang code processing (since it's just a simple version), and subsequently the masterminds error handling code was never executed, lowering code coverage. I could get rid of the error handling code, but that's bad form.
2. The `Masterminds.Version` struct has no user-accessible fields, so you can't create the structure that way. It actually makes more sense to do a `blang.Version.String()` and feed that into `Masterminds.NewVersion()`, but it's the same error handling problem as option 1.
3. This causes the version string to be parsed all the time, also non-optimal, and worse performance than option 1.

tmshort: > Another option: do the (option 2) conversion from blang version to masterminds version in the InMastermindsVersionRange() function.

AFAICT, this means passing an blang structure into a mastermind-named function, which seems bad form.

joelanford: > The Masterminds.Version struct has no user-accessible fields, so you can't create the structure that way.

I think you can do `masterminds.NewVersion(major, minor, patch, pre, metadata string)` and get back a pointer to the struct. It requires a bit of finagling because the blang pre and metadata fields are slices that need to be joined.

joelanford: But letting the caller do the parsing in option 3 seems fine to me too.

tmshort: You're right, blangToMasterminds() would basically be this: 
```
mmsemver.NewVersion(bundledVersion.String)
```
Which is just moving where the conversion takes place

joelanford: Sorry, I was thinking of this function: https://github.com/Masterminds/semver/blob/2f39fdc11c33c38e8b8b15b1f04334ba84e751f2/version.go#L214

tmshort: Although option 3 might be the cleanest, it is changing the behavior of `sortVersion()`.

Anything I do to try to change that _will_ negatively impact code coverage.

joelanford: IMO, trying to cover every error path just leads to brittle tests that make refactoring hard. I think as long as we're covering the happy paths, and the coverage "regression" is only happening because of some refactoring of when/how errors are handled, we shouldn't spend time dealing in the 1s or .1s of coverage percentages.

joelanford (approved):

internal/resolution/entities/bundle_entity.go
tmshort: Is there a way to specify a (minimum, err maximum? negative) threshold for CodeCov?

tmshort: >Sorry, I was thinking of this function: https://github.com/Masterminds/semver/blob/2f39fdc11c33c38e8b8b15b1f04334ba84e751f2/version.go#L214

Because of the pre and metadata fields, etc. it's just easier to do all conversions via `.String()`.

PR #375, opened 2023-08-31
WIP: Remove Entities and EntitySources

dtfranz:
# Description
WIP PR for collaboration on the removal of Entities and EntitySources.

## Reviewer Checklist

- [ ] API Go Documentation


awgreene:
Will Fix #347

internal/controllers/variable_source.go
```diff
@@ -38,5 +38,8 @@ func NewVariableSource(cl client.Client) variablesources.NestedVariableSource {
 		func(inputVariableSource input.VariableSource) (input.VariableSource, error) {
 			return variablesources.NewCRDUniquenessConstraintsVariableSource(inputVariableSource), nil
 		},
+		func(inputVariableSource input.VariableSource) (input.VariableSource, error) {
+			return variablesources.NewCatalogdVariableSource(cl), nil
+		},
```
m1kola, line 43: I think this might have to go higher in the slice. Each variable source in this slice nests the one above and some of the variable sources depend on variables received from the nested sources.

E.g. I believe `NewBundlesAndDepsVariableSource` depends on variables produced by catalog variable source.

internal/resolution/variables/catalogd.go
```diff
@@ -0,0 +1,13 @@
...
+	"github.com/operator-framework/deppy/pkg/deppy"
+	"github.com/operator-framework/deppy/pkg/deppy/input"
+)
+
+var _ deppy.Variable = &CatalogdVariable{}
+
+type CatalogdVariable struct {
+	*input.SimpleVariable
+	dependencies []*BundleVariable
+}
```
m1kola, line 13: Do we use this one? Can't find a reference.

internal/resolution/variablesources/catalogdsource.go
```diff
@@ -44,12 +64,12 @@ func (es *CatalogdEntitySource) Filter(ctx context.Context, filter input.Predica
 	return resultSet, nil
 }
 
-func (es *CatalogdEntitySource) GroupBy(ctx context.Context, fn input.GroupByFunction) (input.EntityListMap, error) {
+func (es *CatalogdVariableSource) GroupBy(ctx context.Context, fn input.GroupByFunction) (map[string][]deppy.Variable, error) {
```
m1kola, line 67: I believe we can remove `Get`, `GroupBy` and `Iterate` easily because we do not use them.

`Filter` has to be replaced somehow. It has a convenient API of accepting reusable predicate functions.

We might want to have some sort of `filter` package to help us work with variables. But I think it is reasonable to start without it and see whether we actually need it (e.g. do we have any repeating patterns?)

internal/resolution/variables/bundle.go
```diff
@@ -1,42 +1,107 @@
...
+	bundlePackage    *property.Package
+	providedGVKs     []GVK
+	requiredGVKs     []GVKRequired
+	requiredPackages []PackageRequired
+	channel          *property.Channel
+	channelEntry     *ChannelEntry
+	semVersion       *bsemver.Version
+	bundlePath       string
+	mediaType        string
+	mu               sync.RWMutex
```
m1kola, line 81: Wondering if we should separate bundle variable (meant for deppy resolution) from bundle data struct holding data from catalog.

m1kola (commented):
Good progress so far. Left a few comments.

internal/resolution/variablesources/catalogdsource.go
m1kola: I've been playing a bit with filtering & separate structures for catalog data. Here is what I came up with so far: https://github.com/operator-framework/operator-controller/pull/393

m1kola:
Let's close this. We have a new branch where we took a different approch. A PR will be created soon.

PR #393, opened 2023-09-04
Adds new `catalogmetadata` package

m1kola:
# Description

Adding a new package to deal with catalog metadata.

We will be using it to replace existing enttiy soruces. At the moment `catalogmetadata` package contains:
* Types which embed `declcfg.Package`, `declcfg.Channel` and `declcfg.Bundle`.
  Later we will be adding more methods to them (for unmarshalling properties, for example). This PR already extends `Bundle` type.
* A function to unmarshal catalog metadata into `Package`, `Channel` or `Bundle` structs.

`catalogmetadata/filter` subpackage:
* `Filter` function which can filter all these 3 types using a given predicate
* `And`, `Or` and `Not` composite predicates to be used with `Filter`
* A set of predicates for bundles

`catalogmetadata/client` subpackage:
* A client for fetching bundles (today we get info from kube API server, but later we will switch implementation to use catalogd HTTP server).

`catalogmetadata/sort` subpackage:
* Functions to sort `catalogmetadata` structs (e.g. byndles by version).

---

This will eventually replace the following:

* `internal/resolution/entities`
* `internal/resolution/entitysources`
* `internal/resolution/util/predicates`
* `internal/resolution/util/sort` (most likely)
* `cmd/resolutioncli/entity_source.go`

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/catalogmetadata/filter/filter.go
+++ b/internal/catalogmetadata/filter/filter.go
@@ -1,10 +1,14 @@
-package catalogmetadata
+package filter
+
+import (
+	"github.com/operator-framework/operator-controller/internal/catalogmetadata"
+)
 
 // Predicate returns true if the object should be kept when filtering
-type Predicate[T Schemas] func(entity *T) bool
+type Predicate[T catalogmetadata.Schemas] func(entity *T) bool
 
 // Filter filters a slice accordingly to
-func Filter[T Schemas](in []*T, test Predicate[T]) []*T {
+func Filter[T catalogmetadata.Schemas](in []*T, test Predicate[T]) []*T {
 	out := []*T{}
 	for i := range in {
 		if test(in[i]) {
@@ -14,7 +18,7 @@ func Filter[T Schemas](in []*T, test Predicate[T]) []*T {
 	return out
 }
 
-func And[T Schemas](predicates ...Predicate[T]) Predicate[T] {
+func And[T catalogmetadata.Schemas](predicates ...Predicate[T]) Predicate[T] {
 	return func(obj *T) bool {
 		for _, predicate := range predicates {
 			if !predicate(obj) {
@@ -25,7 +29,7 @@ func And[T Schemas](predicates ...Predicate[T]) Predicate[T] {
 	}
 }
 
-func Or[T Schemas](predicates ...Predicate[T]) Predicate[T] {
+func Or[T catalogmetadata.Schemas](predicates ...Predicate[T]) Predicate[T] {
 	return func(obj *T) bool {
 		for _, predicate := range predicates {
 			if predicate(obj) {
@@ -36,7 +40,7 @@ func Or[T Schemas](predicates ...Predicate[T]) Predicate[T] {
 	}
 }
 
-func Not[T Schemas](predicate Predicate[T]) Predicate[T] {
+func Not[T catalogmetadata.Schemas](predicate Predicate[T]) Predicate[T] {
 	return func(obj *T) bool {
 		return !predicate(obj)
 	}
--- a/internal/catalogmetadata/sort/sort.go
+++ b/internal/catalogmetadata/sort/sort.go
@@ -1,0 +1,33 @@
+package sort
+
+import (
+	"github.com/operator-framework/operator-controller/internal/catalogmetadata"
+)
+
+// ByVersion is a sort "less" function that orders bundles
+// in inverse version order (higher versions on top).
+func ByVersion(b1, b2 *catalogmetadata.Bundle) bool {
+	ver1, err1 := b1.Version()
+	ver2, err2 := b2.Version()
+	if err1 != nil || err2 != nil {
+		return compareErrors(err1, err2) < 0
+	}
+
+	// Check for "greater than" because
+	// we want higher versions on top
+	return ver1.GT(*ver2)
+}
+
+// compareErrors returns 0 if both errors are either nil or not nil
+// -1 if err1 is nil and err2 is not nil
+// +1 if err1 is not nil and err2 is nil
+func compareErrors(err1 error, err2 error) int {
+	if err1 != nil && err2 == nil {
+		return 1
+	}
+
+	if err1 == nil && err2 != nil {
+		return -1
+	}
+	return 0
+}
--- a/internal/catalogmetadata/types.go
+++ b/internal/catalogmetadata/types.go
@@ -1,7 +1,14 @@
 package catalogmetadata
 
 import (
+	"encoding/json"
+	"fmt"
+	"sync"
+
+	bsemver "github.com/blang/semver/v4"
+
 	"github.com/operator-framework/operator-registry/alpha/declcfg"
+	"github.com/operator-framework/operator-registry/alpha/property"
 )
 
 type Schemas interface {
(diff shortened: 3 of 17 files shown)


internal/catalogmetadata/types.go
```diff
@@ -0,0 +1,21 @@
...
+	declcfg.Package
+}
+
+type Channel struct {
+	declcfg.Channel
+}
+
+type Bundle struct {
+	declcfg.Bundle
+}
```
m1kola, line 21: Thinking about how we can re-implement all the [predicates](https://github.com/operator-framework/operator-controller/blob/edd5ad9b195e56455ad47d292612494930027e0d/internal/resolution/util/predicates/predicates.go) from entities to this struct.

These should be easy:

* `WithPackageName`
* `InMastermindsSemverRange` and `InBlangSemverRange`
* `ProvidesGVK`
* `WithBundleImage`

As all the data we need is in `declcfg.Bundle`.

However `InChannel` and `Replaces` are tricky as `declcfg.Bundle` doesn't include channel membership info required for both.

I'm currently thinking about adding a file like `InChannels []*Channel` to `Bundle` struct which we will be populating when reading catalog metadata.

This should enable us relatively easily answer whether a bundle is in a specific channel or not.

`Replaces` is a bit tricker: we would need to go read each channel from the bundle and make sure that there is (or is not) a `ChannelEntry` which contains a replace with a specific value.

stevekuznetsov:
Is there a quick summary of why this change is being proposed?

m1kola:
@stevekuznetsov we are looking into ways to remove entities and entity sources from operator-controller. This is just me playing with how we might interact with the catalog data. See #375 for more details.

internal/catalogmetadata/filter/filter.go
```diff
@@ -0,0 +1,51 @@
...
+func And[T catalogmetadata.Schemas](predicates ...Predicate[T]) Predicate[T] {
+	return func(obj *T) bool {
+		eval := true
+		for _, predicate := range predicates {
+			eval = eval && predicate(obj)
+			if !eval {
+				return false
+			}
+		}
+		return eval
```
dtfranz, line 30: Pretty nitpicky but could we not just say:
```suggestion
		for _, predicate := range predicates {
			if !predicate(obj) {
				return false
			}
		}
		return true
```

internal/catalogmetadata/filter/filter.go
```diff
@@ -0,0 +1,51 @@
...
+func Or[T catalogmetadata.Schemas](predicates ...Predicate[T]) Predicate[T] {
+	return func(obj *T) bool {
+		eval := false
+		for _, predicate := range predicates {
+			eval = eval || predicate(obj)
+			if eval {
+				return true
+			}
+		}
+		return eval
```
dtfranz, line 43: And similarly for `Or`:
```suggestion
		for _, predicate := range predicates {
			if predicate(obj) {
				return true
			}
		}
		return false
```

internal/catalogmetadata/sort/sort.go
```diff
@@ -0,0 +1,34 @@
+package sort
+
+import (
+	"github.com/operator-framework/operator-controller/internal/catalogmetadata"
+)
+
+// ByVersion is a sort "less" function that orders bundles
+// in inverse version order (higher versions on top).
+func ByVersion(b1, b2 *catalogmetadata.Bundle) bool {
```
dtfranz, line 9: Should we not also sort by package name as we did before with entities? If we don't, what I'm imagining is our list of:
```
- packageName: a-package
  version: 1.0.0
- packageName: b-package
  version: 2.0.0
- packageName: a-package
  version: 3.0.0
```
would get sorted to:
```
- packageName: a-package
  version: 3.0.0
- packageName: b-package
  version: 2.0.0
- packageName: a-package
  version: 1.0.0
```
instead of getting grouped up by name followed by version:
```
- packageName: a-package
  version: 3.0.0
- packageName: a-package
  version: 1.0.0
- packageName: b-package
  version: 2.0.0
```

internal/catalogmetadata/sort/sort.go
```diff
@@ -0,0 +1,34 @@
...
+	ver1, err1 := b1.Version()
+	ver2, err2 := b2.Version()
+	errComp := compareErrors(err1, err2)
+	if errComp != 0 {
+		return errComp < 0
+	}
```
dtfranz, line 15: If both versions come back nil and with errors, then line 19 will panic:
```suggestion
	errComp := compareErrors(err1, err2)
	if errComp != 0 {
		return errComp < 0
	} else if ver1 == nil && ver2 == nil {
		return 0
	}
```
If only one version comes back nil then the `errComp` check should catch it.

internal/catalogmetadata/types.go
```diff
@@ -0,0 +1,149 @@
...
+			return err
+		}
+		b.bundlePackage = &bundlePackage
+
+		semVer, err := bsemver.Parse(b.bundlePackage.Version)
+		if err != nil {
+			return fmt.Errorf("could not parse semver %q for bundle '%s': %s", b.bundlePackage.Version, b.Name, err)
+		}
+		b.semVersion = &semVer
+	}
```
dtfranz, line 92: IMO these should be separated out a bit so that on subsequent `Version()` calls we will continue to get an error back when we try to load it when missing. i.e:
1. `Version()` is called for `Bundle`, but version is missing from package info; line 89 returns an error which can be handled.
2. `Version()` is called again; `b.bundlePackage` is no longer `nil` so we skip loading both fields.
3. We return `b.semVersion` on line 60, which is nil, and no error.
This makes it more difficult to handle this case when doing ops like sorting. I think instead we should do:

```suggestion
	if b.bundlePackage == nil {
		bundlePackage, err := loadFromProps[property.Package](b, property.TypePackage, true)
		if err != nil {
			return err
		}
		b.bundlePackage = &bundlePackage
	}
	if b.semVersion == nil {
		semVer, err := bsemver.Parse(b.bundlePackage.Version)
		if err != nil {
			return fmt.Errorf("could not parse semver %q for bundle '%s': %s", b.bundlePackage.Version, b.Name, err)
		}
		b.semVersion = &semVer
	}
```
Since `b.semVersion` will remain nil we can continue giving the relevant error each time it's requested.

internal/catalogmetadata/filter/filter.go
m1kola: Good point! I just copy-pasted this from older [deppy commit](https://github.com/operator-framework/deppy/blob/4c15c99c52e42655253fcfac942d805d46cd8d22/pkg/deppy/input/query.go#L32-L56) without paying too much attention.

I'll update.

internal/catalogmetadata/sort/sort.go
m1kola: Good catch! I'll update tests and code.

Your suggested code won't work since we need to return bool (true if `b1` is less than `b2` and false otherwise). But we can do something like this:

```go
if err1 != nil || err2 != nil {
	return compareErrors(err1, err2) < 0
}
```

internal/catalogmetadata/sort/sort.go
m1kola: _I think_ ordering by package is not important. We have the following use cases for old bundle entity based `sort.ByChannelAndVersion` at the moment:
* `RequiredPackageVariableSource` - here we always filter by pacakge name so all bundles in sorting will have the same package name
* Twice in `InstalledPackageVariableSource` - on first occasion we call it on a slice filtered by bundle image (so most likely same package) and on second occasion we call it on a slice filtered by `replaces` which should also be within the same package
* `BundlesAndDepsVariableSource` - here is the use case where we might sort different packages. A bundle might require multiple GVKs from different pacakge. But, _I think_, as long as dependencies are sorted by version from newest to oldest we should be ok.

m1kola: I'm not 100% that this is correct, but let's go with it and fix it if it doesn't work. Trying to not overcomplicate it :)

dtfranz (approved):
LGTM!

Issue #189, opened 2023-05-09
[Epic] Refactor ginkgo unit tests to go test


perdasilva:
### Description
---
In discussion https://github.com/operator-framework/operator-controller/discussions/168, we've decided to avoid using ginkgo for unit testing. We need to move our current ginkgo unit tests to go test.

### Tasks
---
- [x] #190
- [x] #191
- [x] #192
- [x] #193
- [x] #194
- [x] #195
- [x] #196
- [x] #197

tmshort:
With the latest set of commits, operator-controller is now ginkgo free!

PR #399, opened 2023-09-06
Remove ginkgo from api tests

tmshort:
This is related to epic #189, although there doesn't seem to be a related issue for this directory.

# Description

## Reviewer Checklist

- [ ] API Go Documentation


tmshort:
@ncdc baby steps to removing Ginkgo...

ncdc (commented):
The conditionsets package variables don't seem to be used anywhere / these tests are ... 🤷 ?

tmshort:
> The conditionsets package variables don't seem to be used anywhere / these tests are ... 🤷 ?

@joelanford added these tests, see #122

api/v1alpha1/operator_types_test.go
```diff
@@ -8,39 +8,51 @@ import (
 	"io/fs"
 	"strconv"
 	"strings"
-
-	. "github.com/onsi/ginkgo/v2"
-	. "github.com/onsi/gomega"
+	"testing"
 
 	"github.com/operator-framework/operator-controller/internal/conditionsets"
+	// replaces with "slices" in go 1.21
```
joelanford, line 14: Nit: put this comment on the same line as the import to be more friendly with tooling that sorts/groups imports.

ncdc:
Interesting, I haven't seen anyone validate that every condition is set post-reconcile before.

api/v1alpha1/operator_types_test.go
```diff
@@ -8,39 +8,51 @@ import (
...
+	"golang.org/x/exp/slices"
 )
 
-var _ = Describe("OperatorTypes", func() {
-	Describe("Condition Type and Reason constants", func() {
-		It("should register types in conditionsets.ConditionTypes", func() {
-			types, err := parseConstants("Type")
-			Expect(err).NotTo(HaveOccurred())
+func TestOperatorType(t *testing.T) {
```
joelanford, line 18: Nit: better test name to convey test purpose a bit better. Same comment for the Reason tests.
```suggestion
func TestOperatorTypeRegistration(t *testing.T) {
```

joelanford:
> Interesting, I haven't seen anyone validate that every condition is set post-reconcile before.

WDYT about that?

I set that test up as a reaction to seeing various reconcilers not properly update conditions when Reconcile doesn't follow the happy path.  This essentially forces all tested code paths to think about how all of the conditions should be set, and it somewhat guards against the possibility of multiple reconciles being necessary to fully process a change/update.

ncdc:
In other projects, our reconcilers added conditions when they were relevant, and didn't ensure 100% of the possible conditions were present 100% of the time. I'm not against the validation, although the purist in me would rather not see exported package level variables that only exist to serve unit tests. But we don't need to debate the pros/cons of that in this PR.

ncdc (approved):

tmshort:
@ncdc, I had to rebase, and handle a `lint` issue due to a comment format issue!

ncdc (approved):

PR #400, opened 2023-09-07
Removes GVKRequired struct

m1kola:
# Description

It doesn't look like we need this duplication.

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/resolution/entities/bundle_entity.go
+++ b/internal/resolution/entities/bundle_entity.go
@@ -47,23 +47,13 @@ func (g GVK) String() string {
 	return fmt.Sprintf(`group:"%s" version:"%s" kind:"%s"`, g.Group, g.Version, g.Kind)
 }
 
-type GVKRequired property.GVKRequired
-
-func (g GVKRequired) String() string {
-	return fmt.Sprintf(`group:"%s" version:"%s" kind:"%s"`, g.Group, g.Version, g.Kind)
-}
-
-func (g GVKRequired) AsGVK() GVK {
-	return GVK(g)
-}
-
 type BundleEntity struct {
 	*input.Entity
 
 	// these properties are lazy loaded as they are requested
 	bundlePackage    *property.Package
 	providedGVKs     []GVK
-	requiredGVKs     []GVKRequired
+	requiredGVKs     []GVK
 	requiredPackages []PackageRequired
 	channel          *property.Channel
 	channelEntry     *ChannelEntry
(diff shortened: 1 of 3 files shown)


internal/resolution/entities/bundle_entity.go
```diff
@@ -47,23 +47,13 @@ func (g GVK) String() string {
 	return fmt.Sprintf(`group:"%s" version:"%s" kind:"%s"`, g.Group, g.Version, g.Kind)
 }
 
-type GVKRequired property.GVKRequired
```
stevekuznetsov, line 50: Why do we have the `type GVK property.GVK` type alias above? Why do we need one of these two aliases but not both?

m1kola: @stevekuznetsov I beleive this is only for use to define `String()` method on it.

I see no reason for `GVKRequired` to exist as both are esentially the same struct\schema.

m1kola: Another guess: probably we re-define structs so we don't spred `property.GVK` across controller-runtime codebase.

stevekuznetsov: a) if we need aliases here, we need all of them
b) if there's a semantic difference between `GVK` and `GVKRequired`, we need to keep that difference here
c) if there's not a difference, let's go to `"github.com/operator-framework/operator-registry/alpha/property"` and remove `GVKRequired`.

> so we don't spred property.GVK across controller-runtime codebase.

What does this mean? Why would we not want to do "spread"? What are the implications?

stevekuznetsov: > There is semantic difference between properties (they have different type), but not between structs: they both represent GVK.

If there is a semantic difference, I don't know that we can remove the use of one. Users are expected to understand the implication of each.

m1kola: We are not removing `olm.gvk.required` property type. User will still be using it. I'm only proposing to re-use a struct into which we unmarshal values of properties of both types `olm.gvk` and `olm.gvk.required`. `GVKRequired` is being converted into `GVK` anyway.

But I agree with you - this should be done on `operator-registry` side first. When creating this PR I didn't notice that that `operator-registry` also has two different structs for it.

m1kola: Ahh, I see. It looks like the only purpose of separate `GVKRequired` is to enable `operator-registry` to [get its type](https://github.com/operator-framework/operator-registry/blob/5ca4a9f7dcb7be737b18f952f72a3a4dc195d841/alpha/property/property.go#L255-L259) using reflection and [map it to a property type](https://github.com/operator-framework/operator-registry/blob/5ca4a9f7dcb7be737b18f952f72a3a4dc195d841/alpha/property/scheme.go#L13).

tmshort (approved):
/lgtm

PR #404, opened 2023-09-09
Revert "Removes GVKRequired struct"

stevekuznetsov:
Reverts operator-framework/operator-controller#400

### DIFF
--- a/internal/resolution/entities/bundle_entity.go
+++ b/internal/resolution/entities/bundle_entity.go
@@ -47,13 +47,23 @@ func (g GVK) String() string {
 	return fmt.Sprintf(`group:"%s" version:"%s" kind:"%s"`, g.Group, g.Version, g.Kind)
 }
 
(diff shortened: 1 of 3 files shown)


tmshort:
Still need the `-s`/Signed option.
So, why are we reverting this?

m1kola (commented):
Sorry, I should’ve marked #400 as a draft after [this thread](https://github.com/operator-framework/operator-controller/pull/400#discussion_r1318854976).

I think both operator-registry and operator-controller can get rid of it, but it is not a trivial thing to do on operator-registry side since it uses go type to map to a property type. It is not a difficult change either, but I don’t have time to look into it at the moment.

While operator-registry still has this this extra type, I think, operator-controller should also have it. There is no impact on functionality on operator-controller side without the second type, but it might be confusing that we have two types on operator-registry side and one in operator-controller.

I'm happy to approve it, but it needs a signoff.

PR #408, opened 2023-09-13
Remove provides+required GVK and modify/remove associated tests

dtfranz:
GVK Properties are not being set on bundle properties, and how we handle these dependencies has not been fully fleshed out. Removing this code will also simplify the transition away from Entity/EntitySource code.
Please also see slack discussion [here](https://kubernetes.slack.com/archives/C0181L6JYQ2/p1694532423074629).


internal/catalogmetadata/types.go
```diff
@@ -11,6 +11,12 @@ import (
 	"github.com/operator-framework/operator-registry/alpha/property"
 )
 
+const (
+	MediaTypePlain          = "plain+v0"
+	MediaTypeRegistry       = "registry+v1"
+	PropertyBundleMediaType = "olm.bundle.mediatype"
+)
```
m1kola, line 18: I was a bit confused since this (and `PackageRequired` an some other bits in this file) is not related to GVK removal. But we still need these changes for further entity source removal.

I'm ok to merge it like this.

internal/catalogmetadata/types_test.go
```diff
@@ -13,140 +14,176 @@ import (
...
-		gvk, err = invalidGVK.ProvidedGVKs()
-		assert.EqualError(t, err, "error determining provided GVKs for bundle \"fake-bundle.invalid\": property \"olm.gvk\" with value \"badGvkStructure\" could not be parsed: invalid character 'b' looking for beginning of value")
-		assert.Nil(t, gvk)
-
-		gvk, err = noGVK.ProvidedGVKs()
-		assert.NoError(t, err)
-		assert.Nil(t, gvk)
-	})
+			}},
+			wantRequiredPackages: nil,
```
m1kola, line 106: Nit: empty value for a pointer is `nil` so this can be ommited.

internal/catalogmetadata/types.go
```diff
@@ -94,28 +90,40 @@ func (b *Bundle) loadPackage() error {
...
 	return nil
 }
 
-func (b *Bundle) loadRequiredGVKs() error {
+func (b *Bundle) loadMediaType() error {
 	b.mu.Lock()
 	defer b.mu.Unlock()
-	if b.requiredGVKs == nil {
-		requiredGVKs, err := loadFromProps[[]GVKRequired](b, property.TypeGVKRequired, false)
+	if b.mediaType == "" {
```
m1kola, line 121: Hm. We should probably use a pointer here: if `nil` - try to unmarshal. If non-`nil` - then we already tried to load it (even if the string itself is empty).

m1kola (commented):
Looks good to me! A few minor things, but I think we can merge.

m1kola (approved):

internal/catalogmetadata/types.go
m1kola: Created #409 for this. Will rebase once this PR gets merged.

PR #411, opened 2023-09-14
Use catalogd HTTP Server instead of `CatalogMetadata` API

### LINKED ISSUE #293: Update catalog content retrieval method
Upon completion of https://github.com/operator-framework/catalogd/milestone/4 catalogd will have a new method for serving catalog content via a HTTP server as opposed to Kubernetes CustomResources. Once the milestone is complete and a release of catalogd containing the changes has been cut, this method for retrieving catalog content for resolution needs to be updated.

everettraven:
# Description
- Bumps `catalogd` dependency to `v0.7.0`
- Update the `internal/catalogmetadata` package to make HTTP requests to the catalogd HTTP server that serves catalog contents instead of using the deprecated `CatalogMetadata` API
- Minimally update the `internal/resolution/entitysources/catalogdsource.go` implementation to also make HTTP requests (didn't do any optimization work here since it is planned to be removed by #413)
- resolves #293 

## Reviewer Checklist

- [ ] API Go Documentation


internal/catalogmetadata/client/client.go
```diff
@@ -31,14 +63,47 @@ func (c *Client) Bundles(ctx context.Context) ([]*catalogmetadata.Bundle, error)
...
+			return nil, fmt.Errorf("error forming request: %s", err)
 		}
 
-		bundles, err := fetchCatalogMetadata[catalogmetadata.Bundle](ctx, c.cl, catalog.Name, declcfg.SchemaBundle)
+		resp, err := http.DefaultClient.Do(req)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("error performing request: %s", err)
+		}
+		defer resp.Body.Close()
```
joelanford, line 81: Data point: the original code hits an in-memory cache to fetch catalog metadata. This code makes a network call for the full response every time.

Should we go ahead and think about caching the full response locally and optimizing our network calls (either by using cache-related request headers, or by keying off something in the catalog API to avoid a network call altogether)

everettraven: If we want to think about caching as part of this that is fine, but IIRC there was some discussion that I thought resulted in "check performance without caching first and if it is poor add caching". I don't have a strong preference either way but the caching concept seems like it may need some more discussion around the best way to do it (i.e are we okay with an in-memory cache or is a filesystem based cache better in this scenario?)

joelanford: Yeah, I mainly want to make sure this is discussed/captured.

> check performance without caching first and if it is poor add caching

+1. I was thinking we might want to figure out how to reliably measure this and then make sure we don't accidentally regress

everettraven: We did add a way to add and enable pprof endpoints to catalogd here: https://github.com/operator-framework/catalogd/blob/f9c6bcfe7e0fcad1b36df18d06061afeafd29458/cmd/manager/main.go#L176-L182

everettraven: Created https://github.com/operator-framework/operator-controller/issues/415 to capture this item as a follow up

grokspawn (approved):

internal/catalogmetadata/client/client.go
```diff
@@ -12,15 +13,46 @@ import (
 	"github.com/operator-framework/operator-controller/internal/catalogmetadata"
 )
 
-func NewClient(cl client.Client) *Client {
-	return &Client{cl: cl}
+// TODO: When a catalogd release containing https://github.com/operator-framework/catalogd/pull/168
+// is made this can be removed in favor of parsing `Catalog.Status` for the
+// appropriate URL for fetching catalog contents.
+const catalogdOnClusterBaseURL = "http://catalogd-catalogserver.catalogd-system.svc"
+const catalogdOnClusterURLTemplate = "%s/catalogs/%s/all.json"
```
anik120, line 20: It's probably worth waiting for that to go in, and to release a v0.6.1 of catalogd for using here

anik120 (dismissed):

internal/catalogmetadata/client/client.go
everettraven: I'm fine with waiting if that's what we really want to hold this on that. 

It looks like the next catalogs release will have to be v0.7.0 since a release branch wasn't created for the v0.6.0 release and a change warranting a minor increment just merged to main. Looks like catalogd is also missing a release guide

internal/catalogmetadata/client/client.go
m1kola: > "check performance without caching first and if it is poor add caching"

We make multiple calls to `Bundles` during single resolution. I think downloading >100mb file multiple times is going to be an issue. Plus having to unmarshall it multiple times...

internal/catalogmetadata/client/client.go
```diff
@@ -31,14 +63,47 @@ func (c *Client) Bundles(ctx context.Context) ([]*catalogmetadata.Bundle, error)
...
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("error performing request: %s", err)
+		}
+		defer resp.Body.Close()
+
+		err = declcfg.WalkMetasReader(resp.Body, func(meta *declcfg.Meta, err error) error {
+			if err != nil {
+				return err
+			}
```
m1kola, line 86: I feel like this error is going to be super generic. Should we format it something like `"more context: %s", err`?

internal/catalogmetadata/client/client.go
```diff
@@ -31,14 +63,47 @@ func (c *Client) Bundles(ctx context.Context) ([]*catalogmetadata.Bundle, error)
...
+			case declcfg.SchemaChannel:
+				content, err := catalogmetadata.Unmarshal[catalogmetadata.Channel](meta)
+				if err != nil {
+					return fmt.Errorf("error unmarshalling catalog metadata: %s", err)
+				}
+				channels = append(channels, content)
+			case declcfg.SchemaBundle:
+				content, err := catalogmetadata.Unmarshal[catalogmetadata.Bundle](meta)
+				if err != nil {
+					return fmt.Errorf("error unmarshalling catalog metadata: %s", err)
```
m1kola, line 98: This and line 92 is the same error message which might be confusing. Can we have something like `error unmarshalling bundles from catalog metadata: %s` and `error unmarshalling channels from catalog metadata: %s`?

P.S.: I know previously we also had this issue when we were just passing an error unmodified from `fetchCatalogMetadata` into the calling code.

internal/catalogmetadata/client/client_test.go
```diff
@@ -60,59 +58,51 @@ func TestClient(t *testing.T) {
...
-							Content: json.RawMessage(`{"name":123123123}`),
-						},
-					})
+					catalogContentMap["catalog-1"] = append(catalogContentMap["catalog-1"], []byte(`{"schema": "olm.bundle", "name":123123123}`)...)
 
-					return objs, nil
+					return objs, nil, catalogContentMap
 				},
-				wantErr: "error unmarshalling catalog metadata: json: cannot unmarshal number into Go struct field Bundle.name of type string",
+				wantErr: "error processing response: expected value for key \"name\" to be a string, got %!t(float64=1.23123123e+08): 1.23123123e+08",
```
m1kola, line 76: I think this refactoring changes the meaning of the test case and probably should be a separate test case.

After refactoring we are hitting this:

```go
if err != nil {
	return err
}
```
in the very beginning of `walkFn` we pass into `WalkMetasReader`.

This means that `invalid bundle` test case and `invalid channel` test case below are basically testing the same code path.

The original was testing output of `Bundles` method when we had a overall valid looking response from the API, but marshalled bundle was invalid (or channel in case of `invalid channel` test case).

Screenshot 2023-09-18 at 10 30 56

internal/catalogmetadata/client/client_test.go
```diff
@@ -127,9 +117,9 @@ func TestClient(t *testing.T) {
 	})
 }
 
-func defaultFakeCatalog() ([]client.Object, []*catalogmetadata.Bundle) {
+func defaultFakeCatalog() ([]client.Object, []*catalogmetadata.Bundle, map[string][]byte) {
 	package1 := `{
-		"schema": "olm.bundle",
+		"schema": "olm.package",
```
m1kola, line 122: Good catch!

internal/catalogmetadata/unmarshal.go
```diff
@@ -3,18 +3,16 @@ package catalogmetadata
...
-	catalogd "github.com/operator-framework/catalogd/api/core/v1alpha1"
+	"github.com/operator-framework/operator-registry/alpha/declcfg"
 )
 
-func Unmarshal[T Schemas](cm []catalogd.CatalogMetadata) ([]*T, error) {
-	contents := make([]*T, 0, len(cm))
-	for _, cm := range cm {
-		var content T
-		if err := json.Unmarshal(cm.Spec.Content, &content); err != nil {
+func Unmarshal[T Schemas](meta *declcfg.Meta) (*T, error) {
```
m1kola, line 9: It was convinient to convert `[]catalogd.CatalogMetadata` into `[]*Bundle` or `[]*Scheme` within another generic.

But given that now call it multipe times on its own & it does not operate on a slice - I think we should probably just get rid of it.

internal/catalogmetadata/client/client.go
joelanford: > release a v0.6.1 of catalogd for using here

That change should _probably_ be v0.7.0 since we're adding a new feature/API field. It's generally a really good habit to only put actual bug fixes in patch releases.

> a release branch wasn't created for the v0.6.0

We can always go back and branch off from the `v0.6.0` commit if we need to patch that minor release.

internal/catalogmetadata/client/client.go
joelanford: ⬆️ this was one of the cases I was particularly worried about. Thanks for calling this out @m1kola!

internal/catalogmetadata/client/client.go
joelanford: To @anik120's point though. This does feel like the wrong abstraction knowing that we're planning to have the URL provided by the individual catalog objects.

It looks like this setup is _just_ to facilitate test code, right?

An alternative could be an `http.RoundTripper` wrapper that replaces the request base URL with the test server's URL. That way test-specific needs don't leak into the Client API.

internal/catalogmetadata/client/client_test.go
everettraven: Ah, good catch! I think this has to do with the actual unmarshalling behavior. It looks like if the response can't be unmarshalled into a `declcfg.Meta` object the first error case is triggered. I'll add some additional testing logic here to make sure we hit all the paths

internal/catalogmetadata/client/client.go
everettraven: >It looks like this setup is just to facilitate test code, right?

No. This setup is specifically because the version of catalogd this is based on doesn't have a status field to show the URL to hit. While that is definitely nice to have, it's not absolutely necessary if you know the pattern for building the correct URL for this version of catalogd. My intention here was to go ahead and get the use of the HTTP server in place by setting these constants that will be removed and replaced when a version of catalogd is used that gives the URL in a Catalog's status.

internal/catalogmetadata/unmarshal.go
everettraven: Got rid of it in https://github.com/operator-framework/operator-controller/pull/411/commits/79cf2386f67dfe6578dcf757139dde69385c748d

internal/catalogmetadata/client/client_test.go
everettraven: Updated the test cases in https://github.com/operator-framework/operator-controller/pull/411/commits/79cf2386f67dfe6578dcf757139dde69385c748d

internal/catalogmetadata/client/client.go
everettraven: Good catch - Updated in https://github.com/operator-framework/operator-controller/pull/411/commits/79cf2386f67dfe6578dcf757139dde69385c748d

internal/catalogmetadata/client/client.go
everettraven: I tried to add more context to it in https://github.com/operator-framework/operator-controller/pull/411/commits/79cf2386f67dfe6578dcf757139dde69385c748d - let me know if this is sufficient

internal/catalogmetadata/client/client.go
everettraven: I took a look at the `http.RoundTripper` and it seems like that would also be leaking test-specific needs into the Client API but I'm happy to use that if it is preferred over the current approach. That being said, it seems like all of this would be moot once Anik's PR in catalogd is merged to add the status field for the URL so if we are going to hold this PR on that I'd prefer we don't do any further work in this specific area of the Client API

internal/catalogmetadata/client/client.go
everettraven: I've added some basic in-memory caching logic as part of https://github.com/operator-framework/operator-controller/pull/411/commits/9618f691101fe3c62de7eb11aa3568685c1558f2

I'm holding off on any unit tests to test the caching functionality until there has been more discussion on the preferred caching approach

everettraven:
IIUC the plan is to hold this PR on a catalogd release that contains https://github.com/operator-framework/catalogd/pull/168 and that caching logic should be added. Due to this I will be placing a hold on this PR

/hold

ncdc:
@everettraven please mark it as a draft PR - we don't use Prow on this repo. Thanks!

internal/catalogmetadata/client/client.go
stevekuznetsov: What is the specific goal of having this repo (in an unreleased `main` state) depend on versioned/released commits from `catalogd`? What is lost if both repos float in an unreleased state between releases?

internal/catalogmetadata/client/client.go
```diff
@@ -12,15 +14,57 @@ import (
...
+	// to the catalogd HTTP server. When using NewClient() this defaults
+	// to http://catalogd-catalogserver.catalogd-system.svc
+	// TODO: When a catalogd release containing https://github.com/operator-framework/catalogd/pull/168
+	// is made this can be removed in favor of parsing `Catalog.Status` for the
+	// appropriate URL for fetching catalog contents.
+	baseURL string
+
+	// cacheInfo is a mapping to hold information necessary
+	// for caching catalog contents
+	cache map[string]CacheData
```
stevekuznetsov, line 67: what are the keys of this? is it easy to make that obvious in the name of the field?

internal/catalogmetadata/client/client.go
```diff
@@ -31,14 +75,75 @@ func (c *Client) Bundles(ctx context.Context) ([]*catalogmetadata.Bundle, error)
...
+		// is made this should be updated in favor of parsing `Catalog.Status` for the
+		// appropriate URL for fetching catalog contents.
+		req, err := http.NewRequestWithContext(ctx, http.MethodGet, fmt.Sprintf(catalogdOnClusterURLTemplate, c.baseURL, catalog.Name), nil)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("error forming request: %s", err)
 		}
 
-		bundles, err := fetchCatalogMetadata[catalogmetadata.Bundle](ctx, c.cl, catalog.Name, declcfg.SchemaBundle)
+		cacheData, ok := c.cache[catalog.Name]
```
stevekuznetsov, line 89: I imagine your `Client` is used concurrently? You likely need a `sync.RWMutex` around access to this member?

internal/catalogmetadata/client/client.go
```diff
@@ -31,14 +75,75 @@ func (c *Client) Bundles(ctx context.Context) ([]*catalogmetadata.Bundle, error)
...
+		if ok {
+			req.Header.Add("If-Modified-Since", cacheData.LastModifiedTime)
```
stevekuznetsov, line 91: Can you use information on the `CatalogSource` object to detect whether your client's cache is out-of-date, definitively? If not, is it possible to expose such information from `catalogd` to facilitate it?

internal/catalogmetadata/client/client_test.go
```diff
@@ -60,59 +61,64 @@ func TestClient(t *testing.T) {
...
-						},
-					})
+					catalogContentMap["catalog-1"] = append(catalogContentMap["catalog-1"],
+						[]byte(`{"schema": "olm.bundle", "name":"foo", "package":"bar", "image":123123123}`)...)
 
-					return objs, nil
+					return objs, nil, catalogContentMap
 				},
-				wantErr: "error unmarshalling catalog metadata: json: cannot unmarshal number into Go struct field Bundle.name of type string",
+				wantErr: "error processing response: error unmarshalling bundle from catalog metadata: json: cannot unmarshal number into Go struct field Bundle.image of type string",
```
stevekuznetsov, line 91: Global comment: errors are opaque values in Go. If a library author wants you to be able to inspect a particular error, they will export something and you will be able to use `errors.As` on it. If not, there is nothing to be gained from testing the specific strings emitted as part of the error - all it does is create extremely brittle unit tests that need to change every time the implementation does.

internal/catalogmetadata/client/client.go
perdasilva: If we have a similar case in the future, would it be sufficient to add caching (optimization) with a follow up PR? we should avoid boiling the ocean. I get that it will be slow. But, so what? It's not in production. Let's keep things small and quick to review. It's a great call out though, I just think next time we should just write up a ticket, link it in code if necessary, merge, then start on the next lot. If we cannot put up with the latency, could we just put it behind a feature gate until we are happy with it?

internal/catalogmetadata/client/client.go
joelanford: `Catalog` has `status.resolvedSource`, which will tell you the SHA resolution of whatever was in the spec. In theory, we can keep track of this resolved ref and only query if we have to cache or there's a mismatch.

internal/catalogmetadata/cache/cache.go
```diff
@@ -0,0 +1,152 @@
...
+// The Catalog provided to this function is expected to:
+// - Be non-nil
+// - Have a Catalog.Status.Conditions entry with type Unpacked and status True
+// - Have a non-nil Catalog.Status.ResolvedSource.Image
+// This ensures that we are only attempting to fetch catalog contents for Catalog
+// resources that have been successfully reconciled, unpacked, and are being served.
+// These requirements help ensure that we can rely on status conditions to determine
+// when to issue a request to update the cached Catalog contents.
+func (fsc *FilesystemCache) FetchCatalogContents(ctx context.Context, catalog *catalogd.Catalog) (io.ReadCloser, error) {
+	fsc.mutex.Lock()
```
stevekuznetsov, line 71: When you have to check whether or not the cache needs updating, be careful not to acquire the write lock too early, as it's exclusive. We can infer from our system architecture that reads will outnumber writes by a large clip, so a less intrusive implementation would look like:

```
read-lock
check preconditions
read-unlock

if preconditions met to write new entry
fetch new data
acquire write lock
check preconditions
ensure nobody raced and updated before us
if preconditions are still in the same place we thought they were at the time we had the read lock
write the data
write unlock
```

internal/catalogmetadata/cache/cache.go
```diff
@@ -0,0 +1,152 @@
...
+	defer fsc.mutex.Unlock()
+
+	shouldUpdateCache := false
+	data, cached := fsc.cacheDataByCatalogName[catalog.Name]
+	if cached {
+		// we should update the cache if the catalog generation has changed
+		// AND the status has been updated to reflect a new image ref
+		shouldUpdateCache = catalog.Generation != data.Generation && catalog.Status.ResolvedSource.Image.Ref != data.ResolvedRef
```
stevekuznetsov, line 79: If the catalog is configured to poll for updates, the generation will not change, but the image ref will. Why do we need to check for the generation here?

internal/catalogmetadata/cache/cache.go
```diff
@@ -0,0 +1,152 @@
...
+
+		// TODO: When a catalogd release containing https://github.com/operator-framework/catalogd/pull/168
+		// is made this should be updated in favor of parsing `Catalog.Status` for the
+		// appropriate URL for fetching catalog contents.
+		req, err := http.NewRequestWithContext(ctx, http.MethodGet, fmt.Sprintf(catalogdOnClusterURLTemplate, catalogdOnClusterBaseURL, catalog.Name), nil)
+		if err != nil {
+			return nil, fmt.Errorf("error forming request: %s", err)
+		}
+
+		// If we are updating the cache, make sure there are actually changes
```
stevekuznetsov, line 100: If we can be certain of our intent by looking at the `*catalog.Catalog` status for the image ref, do we need these headers?

internal/catalogmetadata/cache/cache.go
```diff
@@ -0,0 +1,152 @@
...
+		// If the response status code is a 200 OK, that means
+		// either the information has changed since we cached or
+		// we don't have any cached information for this catalog.
+		// If the response status code is a 304 Not Modified it means
+		// we injected the If-Modified-Since header and have a cache so
+		// we return the cached information. Any other response code is
+		// unexpected and should result in an error
+		switch resp.StatusCode {
+		case http.StatusOK:
+			file, err := os.OpenFile(cacheFilePath, os.O_CREATE|os.O_TRUNC|os.O_WRONLY, os.ModePerm)
```
stevekuznetsov, line 123: nit: why not `os.WriteFile()`?

stevekuznetsov: (one call as opposed to `OpenFile()` `Write()` and `Close()`)

internal/catalogmetadata/cache/cache.go
everettraven: Good point - removed in https://github.com/operator-framework/operator-controller/pull/411/commits/fe924c2c2f9c686d3720bb930d42f01590409df1

internal/catalogmetadata/cache/cache.go
everettraven: Updated this in https://github.com/operator-framework/operator-controller/pull/411/commits/fe924c2c2f9c686d3720bb930d42f01590409df1 - thanks for pairing with me on this to help me understand this better!

internal/catalogmetadata/client/client.go
everettraven: I believe this has been resolve as of https://github.com/operator-framework/operator-controller/pull/411/commits/fe924c2c2f9c686d3720bb930d42f01590409df1 so marking it as such. Feel free to unresolve if this is inaccurate

internal/catalogmetadata/client/client.go
everettraven: This got moved around but should be addressed. Resolving

internal/catalogmetadata/client/client.go
everettraven: Should be taken care of in the new cache implementation as of https://github.com/operator-framework/operator-controller/pull/411/commits/fe924c2c2f9c686d3720bb930d42f01590409df1

internal/resolution/entitysources/catalogdsource.go
```diff
@@ -15,15 +16,35 @@ import (
...
-	client client.Client
+	client  client.Client
+	baseURL string
 }
 
-func NewCatalogdEntitySource(client client.Client) *CatalogdEntitySource {
-	return &CatalogdEntitySource{client: client}
+type CatalogdEntitySourceOpt func(es *CatalogdEntitySource)
+
+func WithBaseURL(baseURL string) CatalogdEntitySourceOpt {
```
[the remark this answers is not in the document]
everettraven: I have it defaulting to a specific value. This option is to allow users to change from the default value. I can make it part of the required parameters if we would like, but I believe all this logic will be removed in #413 so I intentionally made this as barebones as possible.

internal/resolution/entitysources/catalogdsource.go
```diff
@@ -169,21 +190,40 @@ type declcfgSchema interface {
...
+const catalogdOnClusterURLTemplate = "%s/catalogs/%s/all.json"
+
 // TODO: Cleanup once https://github.com/golang/go/issues/45380 implemented
 // We should be able to get rid of the schema arg and switch based on the type passed to this generic
-func fetchCatalogMetadataByScheme[T declcfgSchema](ctx context.Context, cl client.Client, schema, catalogName string) ([]T, error) {
-	cmList := catalogd.CatalogMetadataList{}
-	if err := cl.List(ctx, &cmList, client.MatchingLabels{"schema": schema, "catalog": catalogName}); err != nil {
-		return nil, err
+func fetchCatalogMetadataByScheme[T declcfgSchema](ctx context.Context, baseURL, schema, catalogName string) ([]T, error) {
+	contents := []T{}
```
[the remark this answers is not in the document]
everettraven: IIUC this logic will be removed by #413 so not sure it is worth addressing this for it to be removed in a fairly short amount of time

internal/catalogmetadata/client/client.go
everettraven: >What is the specific goal of having this repo (in an unreleased main state) depend on versioned/released commits from catalogd? What is lost if both repos float in an unreleased state between releases?

No specific goal, just the way this project has been configured. As far as I am aware, all the testing infrastructure and scripts for installing and running operator-controller specifically look for releases from it's dependencies. More explicitly, this Makefile target: https://github.com/operator-framework/operator-controller/blob/6640aacac00f183d861a64d148f4636df1ffb964/Makefile#L139-L142

calls this shell script that installs manifests from a released version of catalogd, rukpak, and cert-manager: https://github.com/operator-framework/operator-controller/blob/6640aacac00f183d861a64d148f4636df1ffb964/scripts/install.tpl.sh#L39-L48

I don't see why we couldn't update everything to make it so operator-controller can depend on unversioned/unreleased commits from it's dependencies but that is significantly out of scope for this PR and, IMO, waiting for a new release is going to be the quickest path to pick up these changes.

internal/catalogmetadata/cache/cache.go
```diff
@@ -0,0 +1,136 @@
...
+
+	catalogd "github.com/operator-framework/catalogd/api/core/v1alpha1"
+
+	"github.com/operator-framework/operator-controller/internal/catalogmetadata/client"
+	catalogmetadataclient "github.com/operator-framework/operator-controller/internal/catalogmetadata/client"
+)
+
+var _ catalogmetadataclient.Fetcher = &filesystemCache{}
+
+func NewFilesystemCache(cachePath string, tripper http.RoundTripper) client.Fetcher {
```
everettraven, line 20: TODO: Add godoc comment

internal/catalogmetadata/cache/cache_test.go
```diff
@@ -0,0 +1,271 @@
...
+			{
+				"name": "fake1.v1.0.0"
+			}
+		]
+	}`
+)
+
+// TODO(everettraven): Add unit tests for testing the caching logic.
+// Holding off on this until the caching implementation is agreed upon
+// (i.e memory based vs file based caching)
```
everettraven, line 55: TODO: Remove this comment

internal/catalogmetadata/client/client_test.go
```diff
@@ -29,29 +33,34 @@ func init() {
 	utilruntime.Must(catalogd.AddToScheme(scheme))
 }
 
+// TODO(everettraven): Add unit tests for testing the caching logic.
+// Holding off on this until the caching implementation is agreed upon
+// (i.e memory based vs file based caching)
```
everettraven, line 38: TODO: remove this comment

everettraven (commented):
Some todos for myself

internal/catalogmetadata/client/client_test.go
everettraven: https://github.com/operator-framework/operator-controller/pull/411/commits/f23b319c87b1b3e87d3d1217c6488537989f0a2d

internal/resolution/entitysources/catalogdsource.go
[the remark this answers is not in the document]
everettraven: It totally went right over my head that you were referring to the contents variable :sweat_smile: . Moved it in https://github.com/operator-framework/operator-controller/pull/411/commits/f23b319c87b1b3e87d3d1217c6488537989f0a2d

internal/resolution/entitysources/catalogdsource.go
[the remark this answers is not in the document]
everettraven: All this was removed when bumping to catalogd v0.7.0

internal/catalogmetadata/client/client_test.go
everettraven: Done in https://github.com/operator-framework/operator-controller/pull/411/commits/1cbe6157a213246232930aa3ce5f7259e6574c61

internal/catalogmetadata/cache/cache.go
```diff
@@ -0,0 +1,143 @@
...
+		return nil, fmt.Errorf("error: provided catalog must be non-nil")
+	}
+
+	if catalog.Status.ResolvedSource == nil {
+		return nil, fmt.Errorf("error: catalog %q has a nil status.resolvedSource value", catalog.Name)
+	}
+
+	if catalog.Status.ResolvedSource.Image == nil {
+		return nil, fmt.Errorf("error: catalog %q has a nil status.resolvedSource.image value", catalog.Name)
+	}
```
joelanford, line 80: I know this may not actually be possible, but in theory shouldn't we be able to fetch catalog contents without a resolved source if we have the content URL?

It seems like these fields not being present means that we could fall back to a `If-Modified-Since` request.

stevekuznetsov: In what real-world scenario would that manifest?

internal/catalogmetadata/cache/cache.go
```diff
@@ -0,0 +1,143 @@
...
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, catalog.Status.ContentURL, nil)
+	if err != nil {
+		return nil, fmt.Errorf("error forming request: %s", err)
+	}
+
+	resp, err := fsc.tripper.RoundTrip(req)
+	if err != nil {
+		return nil, fmt.Errorf("error performing request: %s", err)
+	}
+	defer resp.Body.Close()
```
joelanford, line 103: Why is this not using an `http.Client`?

internal/catalogmetadata/cache/cache.go
```diff
@@ -0,0 +1,143 @@
...
+
+	switch resp.StatusCode {
+	case http.StatusOK:
+		contents, err := io.ReadAll(resp.Body)
+		if err != nil {
+			return nil, fmt.Errorf("error reading response body: %s", err)
+		}
```
joelanford, line 110: Rather than reading the entire file into a byte slice, can we `os.Create` a file and then io.Copy(file, resp.Body)` to stream the response directly into the file?

internal/catalogmetadata/client/client.go
```diff
@@ -31,14 +51,42 @@ func (c *Client) Bundles(ctx context.Context) ([]*catalogmetadata.Bundle, error)
 		return nil, err
 	}
 	for _, catalog := range catalogList.Items {
-		channels, err := fetchCatalogMetadata[catalogmetadata.Channel](ctx, c.cl, catalog.Name, declcfg.SchemaChannel)
+		// if the catalog has not been successfully unpacked, skip it
+		if !meta.IsStatusConditionPresentAndEqual(catalog.Status.Conditions, catalogd.TypeUnpacked, metav1.ConditionTrue) {
+			continue
+		}
```
joelanford, line 57: Follow-up: we probably should not ignore unavailable catalogs. If we do, we introduce non-determinism to resolution.

If a `Catalog` that is needed for resolution is unavailable (that's all `Catalogs` right now because there's no way to disable catalogs globally or for a particular `Operator`), I think we should fail resolution, report the error, and then requeue when the Catalog changes state.

internal/resolution/entitysources/catalogdsource.go
```diff
@@ -171,19 +174,36 @@ type declcfgSchema interface {
 
 // TODO: Cleanup once https://github.com/golang/go/issues/45380 implemented
 // We should be able to get rid of the schema arg and switch based on the type passed to this generic
-func fetchCatalogMetadataByScheme[T declcfgSchema](ctx context.Context, cl client.Client, schema, catalogName string) ([]T, error) {
-	cmList := catalogd.CatalogMetadataList{}
-	if err := cl.List(ctx, &cmList, client.MatchingLabels{"schema": schema, "catalog": catalogName}); err != nil {
-		return nil, err
+func fetchCatalogMetadataByScheme[T declcfgSchema](ctx context.Context, url, schema, _ string) ([]T, error) {
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
```
joelanford, line 178: I might totally be missing something, but why isn't this code using the caching client?

internal/catalogmetadata/cache/cache.go
everettraven: In theory yes, practically though I'm not sure that scenario would take place since the content URL is a side effect of successfully resolving and unpacking a source. 

I could be wrong, but I don't see a real world case where this happens.

internal/catalogmetadata/cache/cache.go
everettraven: No particular reason. I didn't do much looking around to see if there was a better option than an `http.RoundTripper`. Should be a fairly easy update, will look into this in the morning.

joelanford: Yeah, I think a client would be better here. You can still set a custom roundtripper on a client if necessary.

internal/catalogmetadata/cache/cache.go
joelanford: >> I know this may not actually be possible

> I could be wrong, but I don't see a real world case where this happens.

Very well!

internal/resolution/entitysources/catalogdsource.go
everettraven: This is in the entitysources/ package which I believe was removed as part of #413. I was trying to not do a bunch of refactoring of this logic because it was either going to make it in before this PR or shortly after.

internal/catalogmetadata/client/client.go
```diff
@@ -12,15 +16,31 @@ import (
...
+// catalog contents from catalogd.
+type Fetcher interface {
+	// FetchCatalogContents fetches contents from the catalogd HTTP
+	// server for the catalog provided. It returns an io.ReadCloser
+	// containing the FBC contents that the caller is expected to close.
+	// returns an error if any occur.
+	FetchCatalogContents(ctx context.Context, catalog *catalogd.Catalog) (io.ReadCloser, error)
+}
+
+func New(cl client.Client, fetcher Fetcher) *Client {
```
m1kola, line 29: I called this thing "client" due to lack of imagination on my side. Don't hesitate to rename, if you feel like it & have a better name in mind.

internal/catalogmetadata/cache/cache.go
```diff
@@ -0,0 +1,143 @@
...
+	if catalog.Status.ResolvedSource.Image == nil {
+		return nil, fmt.Errorf("error: catalog %q has a nil status.resolvedSource.image value", catalog.Name)
+	}
+
+	cacheDir := filepath.Join(fsc.cachePath, catalog.Name)
+	cacheFilePath := filepath.Join(cacheDir, "data.json")
+
+	fsc.mutex.RLock()
+	if data, ok := fsc.cacheDataByCatalogName[catalog.Name]; ok {
+		if catalog.Status.ResolvedSource.Image.Ref == data.ResolvedRef {
```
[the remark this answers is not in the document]
everettraven: Technically it could be any string, but with the current way catalogd updates this field it will always follow the format `image@sha256:somesha`

internal/catalogmetadata/cache/cache.go
everettraven: Done in https://github.com/operator-framework/operator-controller/pull/411/commits/fcb56fdd451b29de5ebcebd0fcfdb474872d9520

internal/catalogmetadata/cache/cache.go
everettraven: Using a client as of https://github.com/operator-framework/operator-controller/pull/411/commits/fcb56fdd451b29de5ebcebd0fcfdb474872d9520

internal/catalogmetadata/cache/cache.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	resp, err := fsc.client.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("error performing request: %s", err)
+	}
+	defer resp.Body.Close()
+
+	switch resp.StatusCode {
+	case http.StatusOK:
```
m1kola, line 106: I suspect the `switch` was here for `If-Modified-Since` implementation where we needed to handle `304` as well.

What do you think about to geting rid of it to avoid nesting?

E.g.

```go
if resp.StatusCode != http.StatusOK {
	return nil, fmt.Errorf("received unexpected response status code %d", resp.StatusCode)
}

fsc.mutex.Lock()
defer fsc.mutex.Unlock()
// ...
```

internal/catalogmetadata/client/client_test.go
everettraven: >library maintainer you are very unlikely to spot an issue where refactored test or refactored implementation starts hitting a different code path. This happened even in https://github.com/operator-framework/operator-controller/pull/411#discussion_r1328477900.

I think the issue in this PR was spotted due to investigating the test coverage and realizing that a particular path wasn't covered and not necessarily because of the error strings. 

That being said, I don't have a strong preference here (although I do lean more towards creating less brittle unit tests) and am happy to do whatever folks prefer.

internal/catalogmetadata/client/client.go
everettraven: That means that all `Operator` installs have a dependency on all `Catalog` resources being available. If someone creates a `Catalog` resource with an improper image (i.e not an actual catalog image) that would result in being unable to install any `Operator`s until that is remedied which feels like a bad UX IMO. The way I am viewing it is that at resolution time we only care about resolving based on the `Catalog`s that are available and have been successfully unpacked. If none of them provide what we need, fail resolution. The exception here is if a user specifies that an `Operator` should _only_ be installed via a specific `Catalog` (which I'm not sure if that is currently possible). 

Definitely agree that this is a follow-up item and needs further discussion. I created #428 to capture this.

everettraven:
Looks like actions are hung up - closing and re-opening to kick them

internal/catalogmetadata/client/client_test.go
m1kola: > I think the issue in this PR was spotted due to investigating the test coverage and realizing that a particular path wasn't covered and not necessarily because of the error strings.

No it was the fact that the expected error string changed made me wonder whether it was correct change or not. I used code coverage to confirm and illustrate the point.

This is something I catch fairly often this way and I'm not aware of the alternative.

This is not the first time I see the same opinions on error test (another example I could quickly find is this [SO answer](https://stackoverflow.com/a/42036847) too). However I've never heard of an alternative which prevents this and avoids exposing errors as an API (we do not need a custom error type in most of the cases).

This [blog post](https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully) also says "Never inspect the output of error.Error", however it also says:

> I know that sometimes this isn’t possible, and as someone pointed out on twitter, this advice doesn’t apply to writing tests. Never the less, comparing the string form of an error is, in my opinion, a code smell, and you should try to avoid it.

I realise that checking error string is not 100% guarantee from a regression + different code paths might lead to the same error string, but I think, in tests, it is still better than just "err != nil` check.

internal/catalogmetadata/cache/cache.go
everettraven: Good point, will update

internal/catalogmetadata/client/client_test.go
everettraven: >No it was the fact that the expected error string changed made me wonder whether it was correct change or not. I used code coverage to confirm and illustrate the point.

Fair enough. I'm not strongly opinionated on this matter so I'll go ahead and revert it to checking the strings. If there is further discussion desired on this I'd vote that we push this to a separate discussion around testing standards/expectations and keep it from blocking this PR.

everettraven: Reverted the change to `bool` in the latest push. Resolving this - if there is desire for further discussion please create a new issue/discussion about testing standards/expectations, thank you!

m1kola (commented):
I think it looks good to me, but I haven't looked into tests just yet. Hoping to come back to this tomorrow.

cmd/manager/main.go
```diff
@@ -100,7 +106,7 @@ func main() {
 	}
 
 	cl := mgr.GetClient()
-	catalogClient := catalogclient.New(cl)
+	catalogClient := catalogclient.New(cl, cache.NewFilesystemCache(cachePath, http.DefaultClient))
```
joelanford, line 109: https://medium.com/@nate510/don-t-use-go-s-default-http-client-4804cb19f779

joelanford: TL;DR: we need to set some sensible timeouts on the client.

everettraven: TIL

internal/catalogmetadata/cache/cache.go
```diff
@@ -0,0 +1,142 @@
...
+
+	if err = os.MkdirAll(cacheDir, os.ModePerm); err != nil {
+		return nil, fmt.Errorf("error creating cache directory for Catalog %q: %s", catalog.Name, err)
+	}
+
+	file, err := os.Create(cacheFilePath)
+	if err != nil {
+		return nil, fmt.Errorf("error creating cache file for Catalog %q: %s", catalog.Name, err)
+	}
+
```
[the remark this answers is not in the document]
joelanford: Actually, since we're returning this open file handle. Can we Seek(0) and just return `file`?

cmd/manager/main.go
everettraven: Updated in the latest push (https://github.com/operator-framework/operator-controller/pull/411/commits/1427f7950c3e0c3b2fa56916795b133cf149a4c7)

internal/catalogmetadata/cache/cache.go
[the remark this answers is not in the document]
everettraven: Done in the latest push (https://github.com/operator-framework/operator-controller/pull/411/commits/1427f7950c3e0c3b2fa56916795b133cf149a4c7)

internal/catalogmetadata/cache/cache_test.go
```diff
@@ -0,0 +1,269 @@
...
+				},
+				contents: []byte(strings.Join([]string{package1, bundle1, stableChannel}, "\n")),
+				tripper:  &MockTripper{},
+				extraTests: func(t *testing.T, c client.Fetcher, ctx context.Context, tst test) {
+					rc, err := c.FetchCatalogContents(ctx, tst.catalog)
+					assert.NoError(t, err)
+					defer rc.Close()
+					data, err := io.ReadAll(rc)
+					assert.NoError(t, err)
+					assert.Equal(t, tst.contents, data)
```
joelanford, line 106: I can see how these extra tests provide coverage of the cache code, but unless I'm missing something, they don't actually verify whether cached data or fetched data is returned.

internal/catalogmetadata/cache/cache_test.go
```diff
@@ -0,0 +1,269 @@
...
+
+func TestClient(t *testing.T) {
+	t.Run("Bundles", func(t *testing.T) {
+		type test struct {
+			name       string
+			catalog    *catalogd.Catalog
+			contents   []byte
+			wantErr    bool
+			tripper    *MockTripper
+			extraTests func(t *testing.T, c client.Fetcher, ctx context.Context, tst test)
```
joelanford, line 61: I've not seen this pattern before, and I'm worried it might be confusing, ripe for misuse later, or become an abstraction that is no longer useful and that we keep carrying because it's here.

WDYT of factoring these into tests into separate `Test*` functions or `t.Run` blocks that just do a series of things in order?

internal/catalogmetadata/client/client_test.go
```diff
@@ -33,25 +37,27 @@ func TestClient(t *testing.T) {
 	t.Run("Bundles", func(t *testing.T) {
 		for _, tt := range []struct {
 			name        string
-			fakeCatalog func() ([]client.Object, []*catalogmetadata.Bundle)
+			fakeCatalog func() ([]client.Object, []*catalogmetadata.Bundle, map[string][]byte)
 			wantErr     string
+			cache       *MockFetcher
```
joelanford, line 42: nit:
```suggestion
			fetcher       *MockFetcher
```

internal/catalogmetadata/cache/cache_test.go
everettraven: TLDR: I'd prefer this be done in a follow-up if acceptable since this follows the table driven testing the rest of the `catalogmetadata` package seems to follow for multiple tests of the same method.

The thinking here was that this would be a function signature used to run a few extra validation/test steps and at it's base, all of these test cases have to go through the same initial run. The idea was to reduce code repetition while still achieving the ability to do some additional testing steps for the, currently, 2 tests that need it. 

I'd prefer to keep this as is for now since it follows the same table driven testing pattern that exists in the rest of the `catalogmetadata` package and reduces code repetition.

internal/catalogmetadata/client/client_test.go
joelanford: I've encountered this same conundrum as well. Here was my answer:
- https://github.com/operator-framework/operator-registry/blob/eaebcf63f94dabdabc3a25ceeeedcb26aaa11d79/alpha/model/model_test.go#L84-L109
- https://github.com/operator-framework/operator-registry/blob/eaebcf63f94dabdabc3a25ceeeedcb26aaa11d79/alpha/model/model_test.go#L174-L198

internal/catalogmetadata/cache/cache_test.go
everettraven: You're correct on this one - I missed forcing an update to the contents returned by the mock http.RoundTripper. The test below verifies this by ensuring the contents returned by the function are the same as the contents that should be returned by the `http.RoundTripper`

everettraven: For more clarity, the logic behind this is:
- If tst.tripper.content is changed and the cache is not used the content that will be returned != tst.contents
- If tst.tripper.content is changed _and_ the status.resolvedSource.image.ref is changed it won't hit the cache and thus the response should be different than original contents and instead match what the tst.tripper is told to return.

everettraven: (I've updated the code a bit to try and make this more clear when reading it)

internal/catalogmetadata/cache/cache_test.go
```diff
@@ -0,0 +1,269 @@
...
+					},
+					Status: catalogd.CatalogStatus{
+						ResolvedSource: &catalogd.CatalogSource{
+							Image: nil,
+						},
+					},
+				},
+				contents:   []byte(strings.Join([]string{package1, bundle1, stableChannel}, "\n")),
+				tripper:    &MockTripper{serverError: true},
+				extraTests: func(_ *testing.T, _ client.Fetcher, _ context.Context, _ test) {},
```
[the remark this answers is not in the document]
everettraven: https://github.com/operator-framework/operator-controller/pull/411/commits/64c2111bb84f07e4fb4867f64499eeebe5e8bebc

internal/catalogmetadata/client/client_test.go
everettraven: https://github.com/operator-framework/operator-controller/pull/411/commits/64c2111bb84f07e4fb4867f64499eeebe5e8bebc

internal/catalogmetadata/cache/cache_test.go
joelanford: I'm not opposed to table-driven, but that sort of hits on my concern. This isn't totally table-driven because the table actually contains more test code rather then being more declarative/static test input/expectations.

joelanford: If I'm the only one with this concern, I won't hold this PR up on it though.

everettraven: That's a fair point - I'm still new to the table driven testing pattern so I wasn't aware this was actually an anti-pattern for it. Thanks for pointing it out, til :)

Changed it in https://github.com/operator-framework/operator-controller/pull/411/commits/76c9cc8412d8a4ff191989f841802243262b8274 - let me know if this addresses your concerns

internal/catalogmetadata/client/client_test.go
```diff
@@ -60,76 +66,92 @@ func TestClient(t *testing.T) {
...
 									}
 								]
-							}`),
-						},
-					})
+							}`)...)
+
+					return objs, nil, catalogContentMap
+				},
+				wantErr: "bundle \"fake1.v9.9.9\" not found in catalog \"catalog-1\" (package \"fake1\", channel \"channel-with-missing-bundle\")",
```
joelanford, line 73: Nit: when the string itself has double quotes, use backticks for the string literal:
```suggestion
				wantErr: `bundle "fake1.v9.9.9" not found in catalog "catalog-1" (package "fake1", channel "channel-with-missing-bundle")`,
```

joelanford (approved):
One more nit I mentioned about, but looks good. Thanks for sticking with it!

internal/catalogmetadata/client/client_test.go
everettraven: Addressed in https://github.com/operator-framework/operator-controller/pull/411/commits/e0aede3cbae8e0c366026d90a88bdd9dc85d1123

Issue #258, opened 2023-06-08
Ensure the FBC --> entity translation for bundle properties is not lossy


everettraven:
Currently the translation for catalog contents --> entity is lossy when it comes to bundle properties. As we can see from https://github.com/operator-framework/operator-controller/blob/80a5786d9a9335392d35e715b4956a93b9fd64c9/internal/resolution/entitysources/catalogdsource.go#L83-L90 we are only passing through bundle properties that are of type `olm.package` and therefore losing any other bundle property information.

We should update this to ensure that all bundle properties are being forwarded through during the entity translation so they can be used properly.

joelanford:
@perdasilva This seems like a rough edge of the translation from FBC to `input.Entity`. I know the `Properties []property.Property` field in FBC is problematic, but it seems more problematic that we don't have an opaque canonical translation function.

Another option would be to have `input.Entity` change its properties field to match (and maybe translate internally)?

Either way, I think we need to smooth out this edge.

m1kola:
Today properties on a Deppy `input.Entity`  are defined as [a map of strings](https://github.com/operator-framework/deppy/blob/cbf2c66b141b9968935a7f9e43d7269ea7fe9e6f/pkg/deppy/input/entity.go#L9). But in operator-controller I think we can pack anything we want into this map in any format we want: we don’t have to change Deppy for operator-controller to understand repeated properties as Deppy itself is not reading properties (but we consider change to avoid weird marshalling on controller side). And `BundleMetadataSpec` already has properties as a slice. So to me it looks like the ball is entirely on the operator-controller side.

perdasilva:
+1 to @m1kola: I think this fix should go in op-controller. Entities are dumb by design. Though, they might go away. I'd say the right way to fix this would be to just laydown a json string in the property - and copy all of the properties over.

m1kola:
In #413 we got rid of entity source & entity code and now have access to all bundle props, but there is currently an issue on read where we lose some values. Once #427 gets merged, I think we can consider this done.

m1kola:
#427 is merged now. For some reason GH didn't close the issue, so I'm doing it manually.

PR #413, opened 2023-09-14
WIP: Remove Entities and EntitySources

dtfranz:
WIP while the branch and tests are fixed up.

# Description

Bumps deppy version and removes use of Entity and EntitySource objects, in favor of a new  CatalogMetadata client.

Addresses #347


internal/resolution/variables/bundle.go
```diff
@@ -56,3 +58,17 @@ func NewBundleUniquenessVariable(id deppy.Identifier, atMostIDs ...deppy.Identif
 		SimpleVariable: input.NewSimpleVariable(id, constraint.AtMost(1, atMostIDs...)),
 	}
 }
+
+// BundleToBundleVariableIDs returns a list of all possible IDs for a bundle.
+// A bundle can be present in multiple channels and we need a separate variable
+// with a unique ID for each occurrence.
```
joelanford, line 64: I know this is the data model we already have, but I'm really not a fan. There should be a single unique ID for each bundle in a particular catalog. I'm worried that creating multiple entities per bundle (one for each channel), will result in very confusing resolution unsat error messages.

Also, generating all of these extra bundle variables and making the variable space larger presumably requires more memory and CPU for resolution.

Not saying we fix this now, but definitely something we should fix soon before it becomes impossible.

m1kola: @joelanford I wrote this code and I'm not a fan too! :) It is here because this is how it currently works in `main`: see `internal/resolution/entitysources/catalogdsource.go` where we create multiple instances of `Entity` - one per bundle occurnace in package channels.

However since I wrote this I was wondering whether it is actually necessary to have channel aware variables with all the filtering we do before we give it to the solver. I suggest that we leave it like this for now to stay close to what we already have in `main`, but look into it as a separate effort. I created #414 for this.

cmd/resolutioncli/main.go
```diff
@@ -150,22 +157,14 @@ func resolve(ctx context.Context, resolver *solver.DeppySolver, packageName stri
 	}
 
 	// Get the bundle image reference for the bundle
-	bundleImage, err := bundleEntity.BundlePath()
-	if err != nil {
-		return "", err
-	}
-
-	return bundleImage, nil
+	return bundleEntity.Image, nil
```
m1kola, line 160: Should probably rename `bundleEntity` -> `bundle`.

And below `getBundleEntityFromSolution` -> `bundleFromSolution` (in go we usually do not prefix with `get`).

internal/controllers/operator_controller.go
```diff
@@ -262,14 +251,11 @@ func mapBDStatusToInstalledCondition(existingTypedBundleDeployment *rukpakv1alph
 	}
 }
 
-func (r *OperatorReconciler) getBundleEntityFromSolution(solution *solver.Solution, packageName string) (*entities.BundleEntity, error) {
+func (r *OperatorReconciler) getBundleEntityFromSolution(solution *solver.Solution, packageName string) (*catalogmetadata.Bundle, error) {
```
m1kola, line 254:
```suggestion
func (r *OperatorReconciler) bundleFromSolution(solution *solver.Solution, packageName string) (*catalogmetadata.Bundle, error) {
```

internal/controllers/operator_controller.go
```diff
@@ -262,14 +251,11 @@ func mapBDStatusToInstalledCondition(existingTypedBundleDeployment *rukpakv1alph
...
 	for _, variable := range solution.SelectedVariables() {
 		switch v := variable.(type) {
 		case *olmvariables.BundleVariable:
-			entityPkgName, err := v.BundleEntity().PackageName()
-			if err != nil {
-				return nil, err
-			}
+			entityPkgName := v.BundleEntity().Package
```
m1kola, line 258:
```suggestion
			pkgName := v.Bundle().Package
```

internal/controllers/operator_controller.go
```diff
@@ -165,20 +165,9 @@ func (r *OperatorReconciler) reconcile(ctx context.Context, op *operatorsv1alpha
...
-		setResolvedStatusConditionFailed(&op.Status.Conditions, err.Error(), op.GetGeneration())
-		return ctrl.Result{}, err
-	}
-
-	// Now we can set the Resolved Condition, and the resolvedBundleSource field to the bundleImage value.
-	op.Status.ResolvedBundleResource = bundleImage
-	setResolvedStatusConditionSuccess(&op.Status.Conditions, fmt.Sprintf("resolved to %q", bundleImage), op.GetGeneration())
+	// Now we can set the Resolved Condition, and the resolvedBundleSource field to the bundleEntity.Image value.
+	op.Status.ResolvedBundleResource = bundleEntity.Image
+	setResolvedStatusConditionSuccess(&op.Status.Conditions, fmt.Sprintf("resolved to %q", bundleEntity.Image), op.GetGeneration())
```
m1kola, line 170:
```suggestion
	// Now we can set the Resolved Condition, and the resolvedBundleSource field to the bundle.Image value.
	op.Status.ResolvedBundleResource = bundle.Image
	setResolvedStatusConditionSuccess(&op.Status.Conditions, fmt.Sprintf("resolved to %q", bundle.Image), op.GetGeneration())
```

cmd/resolutioncli/main.go
```diff
@@ -150,22 +157,14 @@ func resolve(ctx context.Context, resolver *solver.DeppySolver, packageName stri
...
-func getBundleEntityFromSolution(solution *solver.Solution, packageName string) (*olmentity.BundleEntity, error) {
+func getBundleEntityFromSolution(solution *solver.Solution, packageName string) (*catalogmetadata.Bundle, error) {
 	for _, variable := range solution.SelectedVariables() {
 		switch v := variable.(type) {
 		case *olmvariables.BundleVariable:
-			entityPkgName, err := v.BundleEntity().PackageName()
-			if err != nil {
-				return nil, err
-			}
+			entityPkgName := v.BundleEntity().Package
```
m1kola, line 167:
```suggestion
			pkgName := v.Bundle().Package
```

internal/resolution/variables/bundle.go
```diff
@@ -1,42 +1,44 @@
...
 	*input.SimpleVariable
-	bundleEntity *olmentity.BundleEntity
-	dependencies []*olmentity.BundleEntity
+	bundle       *catalogmetadata.Bundle
+	dependencies []*catalogmetadata.Bundle
 }
 
-func (b *BundleVariable) BundleEntity() *olmentity.BundleEntity {
-	return b.bundleEntity
+func (b *BundleVariable) BundleEntity() *catalogmetadata.Bundle {
```
m1kola, line 21:
```suggestion
func (b *BundleVariable) Bundle() *catalogmetadata.Bundle {
```

internal/resolution/variables/installed_package.go
```diff
@@ -7,28 +7,28 @@ import (
...
 
 type InstalledPackageVariable struct {
 	*input.SimpleVariable
-	bundleEntities []*olmentity.BundleEntity
+	bundles []*catalogmetadata.Bundle
 }
 
-func (r *InstalledPackageVariable) BundleEntities() []*olmentity.BundleEntity {
-	return r.bundleEntities
+func (r *InstalledPackageVariable) BundleEntities() []*catalogmetadata.Bundle {
```
m1kola, line 20:
```suggestion
func (r *InstalledPackageVariable) Bundle() []*catalogmetadata.Bundle {
```

internal/resolution/variables/required_package.go
```diff
@@ -7,28 +7,28 @@ import (
...
 
 type RequiredPackageVariable struct {
 	*input.SimpleVariable
-	bundleEntities []*olmentity.BundleEntity
+	bundles []*catalogmetadata.Bundle
 }
 
-func (r *RequiredPackageVariable) BundleEntities() []*olmentity.BundleEntity {
-	return r.bundleEntities
+func (r *RequiredPackageVariable) BundleEntities() []*catalogmetadata.Bundle {
```
m1kola, line 20:
```suggestion
func (r *RequiredPackageVariable) Bundle() []*catalogmetadata.Bundle {
```

internal/resolution/variablesources/bundles_and_dependencies.go
```diff
@@ -5,111 +5,115 @@ import (
...
 
-		// get bundle dependencies
-		dependencyEntityBundles, err := b.getEntityDependencies(ctx, head, entitySource)
-		if err != nil {
-			return nil, fmt.Errorf("could not determine dependencies for entity with id '%s': %w", head.ID, err)
-		}
+			// get bundle dependencies
+			dependencies, err := b.filterBundleDependencies(allBundles, head)
+			if err != nil {
+				return nil, fmt.Errorf("could not determine dependencies for entity with id '%s': %w", id, err)
```
m1kola, line 77:
```suggestion
				return nil, fmt.Errorf("could not determine dependencies for bundle with id '%s': %w", id, err)
```

internal/resolution/variablesources/installed_package.go
```diff
@@ -3,82 +3,66 @@ package variablesources
...
-func (r *InstalledPackageVariableSource) GetVariables(ctx context.Context, entitySource input.EntitySource) ([]deppy.Variable, error) {
-	// find corresponding bundle entity for the installed content
-	resultSet, err := entitySource.Filter(ctx, predicates.WithBundleImage(r.bundleImage))
+func (r *InstalledPackageVariableSource) GetVariables(ctx context.Context) ([]deppy.Variable, error) {
+	allBundles, err := r.catalog.Bundles(ctx)
 	if err != nil {
 		return nil, err
 	}
+
+	// find corresponding bundle entity for the installed content
```
m1kola, line 30:
```suggestion
	// find corresponding bundle for the installed content
```

internal/resolution/variables/bundle.go
joelanford: > whether it is actually necessary to have channel aware variables with all the filtering we do before we give it to the solver.

There are at least two ways to do this:
1. As you said, when building bundle variables from an `Operator`, we know whether or not we need to filter by channels, and we know what all of the channel membership is. When we build the operator variable, we would end up with something close to `OperatorFoo` depends on `BundleVariables(<set-of-bundle-variables-that-match-spec-specified-constraints>)`. So we would be able to filter on package name, bundle version, channel, and upgrade candidates before we even hand a variable off to the resolver. This is pretty much what I did in my PoC starting around here: https://github.com/joelanford/operator-controller/blob/31082f42fa9fbe61a4c9155683d0956e2425b774/internal/resolution/v2/variablesources/operator.go#L71
2. Another option is to build variables for channels, like: `CatFoo/PackageBar/ChannelAlpha` depends on `BundleVariables(<set of bundles in the channel>)`. And then if an `Operator` specifies the alpha channel, then you'd make `OperatorFoo` depends on `CatFoo/PackageBar/ChannelAlpha`. (and if it doesn't specify a channel, then intuitively, it would have no dependencies on any channel variables).

One thing to worry about with option (2) is if we ever have an API that let's you specify that the bundle can come from multiple channels. With option 2, what's the preference ordering? Is it first by channel, then by version? Or is it first by version, then by channel? Either way, that preference order is easier to model/implement with option (1) because you can sort a flat list of bundles however you like.

internal/controllers/operator_controller_test.go
```diff
@@ -575,43 +580,6 @@ var _ = Describe("Operator Controller Test", func() {
 				})
 			})
 		})
-		When("the selected bundle's image ref cannot be parsed", func() {
```
dtfranz, line 578: Removed this test as the bundle path retrieval no longer has an error path with the new struct.

ncdc:
@dtfranz please update the description to say either "resolves" or "fixes" instead of "addresses". The former will result in the associated issue being closed when the PR is merged. That doesn't happen with "addresses". Thanks.

internal/catalogmetadata/client/client.go
```diff
@@ -12,6 +12,10 @@ import (
 	"github.com/operator-framework/operator-controller/internal/catalogmetadata"
 )
 
+type CatalogClient interface {
```
ncdc, line 15: https://github.com/golang/go/wiki/CodeReviewComments#interfaces

cmd/manager/main.go
```diff
@@ -99,12 +99,14 @@ func main() {
 		os.Exit(1)
 	}
 
+	cl := mgr.GetClient()
+	catalogClient := catalogclient.NewClient(cl)
```
ncdc, line 103: I know NewClient isn't net new in this PR, but given the package is named client, let's consider renaming NewClient to New.

cmd/resolutioncli/catalogmetadata_reader.go
```diff
@@ -0,0 +1,123 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	catalogd "github.com/operator-framework/catalogd/api/core/v1alpha1"
+	"github.com/operator-framework/operator-registry/alpha/action"
+	"github.com/operator-framework/operator-registry/alpha/declcfg"
+)
+
+type catalogmetadataReader struct {
```
ncdc, line 34: Metadata should be capitalized; i.e., catalogMetadataReader

cmd/resolutioncli/catalogmetadata_reader.go
```diff
@@ -0,0 +1,123 @@
...
+	renderer      action.Render
+	clientBuilder *fake.ClientBuilder
```
ncdc, line 36: Why is this using fakes?

cmd/resolutioncli/variable_source.go
```diff
@@ -19,12 +19,14 @@ package main
 import (
 	"github.com/operator-framework/deppy/pkg/deppy/input"
 
+	catalogclient "github.com/operator-framework/operator-controller/internal/catalogmetadata/client"
 	"github.com/operator-framework/operator-controller/internal/resolution/variablesources"
 )
 
-func newPackageVariableSource(packageName, packageVersion, packageChannel string) func(inputVariableSource input.VariableSource) (input.VariableSource, error) {
+func newPackageVariableSource(catalog catalogclient.CatalogClient, packageName, packageVersion, packageChannel string) func(inputVariableSource input.VariableSource) (input.VariableSource, error) {
```
ncdc, line 26: Global comments for all places we're referencing a CatalogClient in this PR:
1. Given the variable's type is CatalogClient, I would expect to see "client" somewhere in the variable's name
2. Maybe this is actually more like a "BundleProvider"? Or does that make it too abstract, and just calling this catalogClient would suffice?

internal/catalogmetadata/types_test.go
```diff
@@ -123,7 +123,13 @@ func TestBundleRequiredPackages(t *testing.T) {
 	} {
 		t.Run(tt.name, func(t *testing.T) {
 			packages, err := tt.bundle.RequiredPackages()
-			assert.Equal(t, tt.wantRequiredPackages, packages)
+			assert.Equal(t, len(tt.wantRequiredPackages), len(packages))
+			for i, pkg := range packages {
+				// Must custom compare due to Semver function type
+				assert.Equal(t, tt.wantRequiredPackages[i].PackageRequired, pkg.PackageRequired)
```
ncdc, line 129: Haven't looked at the full test ... are these guaranteed to sort consistently (packages and tt.wantRequiredPackages)?

ncdc (commented):
I am out of time for today, but wanted to submit what I had for now. Can try to review more tomorrow.

internal/catalogmetadata/types.go
```diff
@@ -129,6 +129,9 @@ func (b *Bundle) loadMediaType() error {
 }
 
 func (b *Bundle) propertyByType(propType string) *property.Property {
+	// TODO: Get rid of the map and support multiple properties with the same type
+	// Some property types such as `olm.gvk` and `olm.gvk.required`
+	// can appear in the b.Properties list multiple
```
joelanford, line 134: Is there an issue for this?

joelanford:
It looks like this may actually resolve the issue from #419 as well. In this PR (or maybe a follow-up is okay), can we update the controller unit test and the e2e test to add a new `alpha` channel for the prometheus package that contains a non-latest version, and then update the relevant assertions to make sure that 0.65.1 is still the resolved bundle?

See:
- https://github.com/operator-framework/operator-controller/pull/420/files#diff-b97ef0be75ddfb0d44613fbe3583548b7e4e75ca45faf8d1cbf25e7600e3ed8a
- https://github.com/operator-framework/operator-controller/pull/420/files#diff-a776550a4d41e3a27063ccc4c50e5cd7e0d90c5dc234cd96217b345a1e52ac54
- https://github.com/operator-framework/operator-controller/pull/420/files#diff-041c5c4339769a620410df332c23ce9b0aa3e4824b26aadda8c5a5c6003e57cf

cmd/manager/main.go
m1kola: Good catch. My editor used to suggest getting rid of names like this, but for some reaons no longer does it. Wondering if I'm missing some tools.

Anyway - I created #422 for this.

cmd/resolutioncli/catalogmetadata_reader.go
m1kola: Resolution CLI is a PoC for offline resolution. It doesn't interact with cluster. I used fakes here as a quick way to keep the rest of the code untouched, but now given that we will be removing kube client in #411 I think we should just create a second implementation of a Client which reads from a FS and has the same interfacte as the `Client` from `internal/catalogmetadata/client`.

Should save us some marshalling too.

internal/catalogmetadata/types_test.go
m1kola: It is guaranteed however I think this fix should be a separate PR. Created #423 for this.

cmd/resolutioncli/catalogmetadata_reader.go
m1kola: No longer relevant as I replaced it with a client

cmd/resolutioncli/catalogmetadata_reader.go
m1kola: Replaced with a client.

m1kola:
Converted back to draft to avoud accidental merge. It includes changes from #423 + some feedback still need to be addressed.

m1kola:
We probably also address #258 here

cmd/resolutioncli/variable_source.go
dtfranz: Personally I like `BundleProvider`, to me that makes it clear what the intended purpose of the type is.

internal/catalogmetadata/client/client.go
dtfranz: Updated; this interface is now declared in the `variablesources` pkg so we can create mocks for testing there.

joelanford:
> We probably also address https://github.com/operator-framework/operator-controller/issues/258 here

Is this that TODO comment I asked if there was a issue for. Definitely seems related. Seems like the `propertyByType` function is still lossy because, for duplicate property types, only the last one is found/kept/used.

dtfranz:
> It looks like this may actually resolve the issue from #419 as well. In this PR (or maybe a follow-up is okay), can we update the controller unit test and the e2e test to add a new `alpha` channel for the prometheus package that contains a non-latest version, and then update the relevant assertions to make sure that 0.65.1 is still the resolved bundle?

The changes are pretty big here already, I'd like to take care of this in a follow-up if that's alright. I think that you are correct though, that this should resolve that issue.

internal/resolution/variables/installed_package_test.go
```diff
@@ -1,43 +1,42 @@
...
-		})),
+
+	bundles := []*catalogmetadata.Bundle{
+		{Bundle: declcfg.Bundle{Name: "bundle-1", Properties: []property.Property{
+			{Type: property.TypePackage, Value: json.RawMessage(`{"packageName": "test-package", "version": "1.0.0"}`)},
+		}}},
+		{Bundle: declcfg.Bundle{Name: "bundle-2", Properties: []property.Property{
+			{Type: property.TypePackage, Value: json.RawMessage(`{"packageName": "test-package", "version": "2.0.0"}`)},
+		}}},
+		{Bundle: declcfg.Bundle{Name: "bundle-2", Properties: []property.Property{
```
joelanford, line 26:
```suggestion
		{Bundle: declcfg.Bundle{Name: "bundle-3", Properties: []property.Property{
```

internal/resolution/variables/required_package_test.go
```diff
@@ -1,43 +1,41 @@
...
-			property.TypeChannel: `{"channelName":"stable","priority":0}`,
-		})),
+	bundles := []*catalogmetadata.Bundle{
+		{Bundle: declcfg.Bundle{Name: "bundle-1", Properties: []property.Property{
+			{Type: property.TypePackage, Value: json.RawMessage(`{"packageName": "test-package", "version": "1.0.0"}`)},
+		}}},
+		{Bundle: declcfg.Bundle{Name: "bundle-2", Properties: []property.Property{
+			{Type: property.TypePackage, Value: json.RawMessage(`{"packageName": "test-package", "version": "2.0.0"}`)},
+		}}},
+		{Bundle: declcfg.Bundle{Name: "bundle-2", Properties: []property.Property{
```
joelanford, line 25:
```suggestion
		{Bundle: declcfg.Bundle{Name: "bundle-3", Properties: []property.Property{
```

joelanford (approved):
A few copy/pastas, but LGTM

internal/catalogmetadata/types.go
m1kola: I created a PR for this: https://github.com/operator-framework/operator-controller/pull/427

Will remove the to do from this PR soon as it affects nothing here.

cmd/resolutioncli/variable_source.go
m1kola: Eventually there might be a need to read other catalog info like channels and packages. So the best thing I came up with was "catalog client", but I don't really like it.

Happy to go with `BundleProvider` for now and we can rename later when/if we need to expand.

internal/resolution/variablesources/bundles_and_dependencies_test.go
```diff
@@ -166,34 +255,57 @@ var _ = Describe("BundlesAndDepsVariableSource", func() {
...
+							InChannels: []*catalogmetadata.Channel{{Channel: declcfg.Channel{Name: "stable"}}},
+						},
+					}),
+				},
+			},
+		)
+		_, err := bdvs.GetVariables(context.TODO())
 		Expect(err).To(HaveOccurred())
-		Expect(err).To(MatchError("could not determine dependencies for entity with id 'bundle-2': could not find package dependencies for bundle 'bundle-2'"))
+		Expect(err.Error()).To(ContainSubstring("could not determine dependencies for bundle with id '-test-package-stable-bundle-2': could not find package dependencies for bundle 'bundle-2'"))
```
m1kola, line 300: Nit

> id '-test-package-stable-bundle-2'

This starts with `-` because `CatalogName` is missing in the fake bundle.

I'll update.

internal/resolution/variablesources/bundles_and_dependencies_test.go
```diff
@@ -222,15 +340,7 @@ func VariableWithID(id deppy.Identifier) func(vars []*olmvariables.BundleVariabl
 func CollectBundleVariableIDs(vars []*olmvariables.BundleVariable) []string {
 	ids := make([]string, 0, len(vars))
 	for _, v := range vars {
-		ids = append(ids, v.Identifier().String())
+		ids = append(ids, v.Bundle().Name)
```
m1kola, line 343: I think this one should still use IDs. Bundle names are not unique. We need to update assert to use new bundle variable IDs.

m1kola (commented):
Looks good to me!

P.S. Not submitting this as "Approve" review beucase I was heavily involved in this and it is not right to approve my own code .

PR #420, opened 2023-09-20
fix: do not include channel name or priority in bundle sorting during resolution

### LINKED ISSUE #419: Bug: Non-latest version installed in packages with multiple channels
What I did:
1. Install the operatorhubio `Catalog`
2. Create an `Operator` with just `spec.packageName: grafana-operator`

What happened:
- Grafana-operator v3.10.3 was resolved/installed

What I expected to happen:
- Grafana-operator v5.4.1 should have been resolved/installed

What I think is happening:
- We are sorting by channel priority and lexicographical name before sorting for version. Therefore bundles present in higher priority channels or in equal priority channels with lexicographically lower names are preferred.

What I think the fix is:
- Stop using channel for anything other than set membership checks. Specifically, remove [these lines](https://github.com/operator-framework/operator-controller/blob/4b9cfd94862dc486c02cf4482dff97f8429d007f/internal/resolution/util/sort/sort.go#L25-L29) (and related code/tests). It is not (yet at least) in the OLMv1 design for there to be channel priorities or for there to be any sorting related to channels. It is thus far meant to be purely a set membership attribute used to filter bundles via `spec.channel`.

joelanford:
# Description

#413 fixed the bug described in #419

This PR updates tests to ensure there are no regressions.

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/controllers/operator_controller_test.go
+++ b/internal/controllers/operator_controller_test.go
@@ -788,7 +788,7 @@ var _ = Describe("Operator Controller Test", func() {
 			BeforeEach(func() {
 				By("initializing cluster state")
 				pkgName = "prometheus"
-				pkgChan = "alpha"
+				pkgChan = "non-existent"
 				operator = &operatorsv1alpha1.Operator{
 					ObjectMeta: metav1.ObjectMeta{Name: opKey.Name},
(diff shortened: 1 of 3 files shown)


dtfranz (approved):
LGTM!

PR #423, opened 2023-09-20
Fixes semver range parsing for required packages

m1kola:
# Description

Moving fix from #413 into a separate PR.

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/catalogmetadata/types.go
+++ b/internal/catalogmetadata/types.go
@@ -31,7 +31,7 @@ type Channel struct {
 
 type PackageRequired struct {
 	property.PackageRequired
-	SemverRange *bsemver.Range `json:"-"`
+	SemverRange bsemver.Range `json:"-"`
 }
 
 type Bundle struct {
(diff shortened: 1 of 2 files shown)


internal/catalogmetadata/types.go
```diff
@@ -31,7 +31,7 @@ type Channel struct {
 
 type PackageRequired struct {
 	property.PackageRequired
-	SemverRange *bsemver.Range `json:"-"`
```
m1kola, line 34: I don't think we need a pointer to a function here.

internal/catalogmetadata/types.go
```diff
@@ -98,17 +98,17 @@ func (b *Bundle) loadRequiredPackages() error {
...
 			if err != nil {
 				return fmt.Errorf(
 					"error parsing bundle required package semver range for bundle %q (required package %q): %s",
 					b.Name,
-					requiredPackage.PackageName,
+					requiredPackages[i].PackageName,
 					err,
 				)
 			}
-			requiredPackage.SemverRange = &semverRange
```
m1kola, line 111: So the issue is that we were setting `SemverRange` on a copy and not on the instance from `requiredPackages`.

internal/catalogmetadata/types_test.go
```diff
@@ -123,6 +123,16 @@ func TestBundleRequiredPackages(t *testing.T) {
 	} {
 		t.Run(tt.name, func(t *testing.T) {
 			packages, err := tt.bundle.RequiredPackages()
+			assert.Len(t, tt.wantRequiredPackages, len(packages))
```
m1kola, line 126:
```suggestion
			assert.Len(t, packages, len(tt.wantRequiredPackages))
```

perdasilva (approved):
lgtm

PR #426, opened 2023-09-20
🌱 Support configuring plain bundle image url

awgreene:
This commit allows users to modify the plain bundle image url.


m1kola:
@awgreene could you please provide more background for this? E.g. what issue are we trying to solve?

perdasilva:
closing due to inactivity

PR #427, opened 2023-09-21
Fixes reading properties with the same type

### LINKED ISSUE #258: Ensure the FBC --> entity translation for bundle properties is not lossy

m1kola:
Some properties like `olm.package` only can appear in a bundle once and we currently handle this correctly.

However properties such as `olm.gvk` and `olm.gvk.required` can be present multiple times in the list. The current implementation overrides all the instances with the one which is the last in the list.

This modifies the code to be able to read in both cases.

Closes #258

# Description

## Reviewer Checklist

- [ ] API Go Documentation


internal/catalogmetadata/types.go
```diff
@@ -94,7 +94,7 @@ func (b *Bundle) loadRequiredPackages() error {
 	b.mu.Lock()
 	defer b.mu.Unlock()
 	if b.requiredPackages == nil {
-		requiredPackages, err := loadFromProps[[]PackageRequired](b, property.TypePackageRequired, false)
+		requiredPackages, err := loadOneFromProps[[]PackageRequired](b, property.TypePackageRequired, false)
```
joelanford, line 97: Is this `olm.package.required`? If so, there can be multiple of these.

joelanford (approved):

PR #438, opened 2023-10-05
Add ForceSemverUpgradeConstraints feature gate

### LINKED ISSUE #395: Add `ForceSemverUpgradeConstraints` feature gate

We need to implement a feature gate as [per RFC](https://docs.google.com/document/d/1dfiNr0kvP1nfS0f9xaGiLyfrJ_KcD38clYs4LYBKllk/edit#heading=h.u7sdl028f7k5):

> Need a feature gate to switch between two upgrade constraint semantics. When:
> * Disabled - operator-controller uses legacy semantics
>* Enabled ([default](https://docs.google.com/document/d/1dfiNr0kvP1nfS0f9xaGiLyfrJ_KcD38clYs4LYBKllk/edit#heading=h.cwjgvarjzbfe)) - operator-controller uses semver
>
> The feature gate can be enabled or disabled by passing the following flag to the manager binary:
>
> ```
> --feature-gates=ForceSemverUpgradeConstraints=true
> ```
>
> This functionality will be built on top of operator-controller’s [existing feature gate mechanisms](https://github.com/operator-framework/operator-controller/blob/4eba8f35591cfd4ba65bd85d651603f97e9eed78/pkg/features/features.go) which is based on the [k8s.io/component-base/featuregate](https://pkg.go.dev/k8s.io/component-base/featuregate) package.

m1kola:
# Description

Closes #395

## Reviewer Checklist

- [ ] API Go Documentation


pkg/features/features.go
```diff
@@ -6,13 +6,17 @@ import (
...
+	// Ex: SomeFeature featuregate.Feature = "SomeFeature"
+
+	ForceSemverUpgradeConstraints featuregate.Feature = "ForceSemverUpgradeConstraints"
 )
 
 var operatorControllerFeatureGates = map[featuregate.Feature]featuregate.FeatureSpec{
 	// Add new feature gate definitions
 	// Ex: SomeFeature: {...}
+
+	ForceSemverUpgradeConstraints: {Default: true, PreRelease: featuregate.Alpha},
```
everettraven, line 19: Non-blocking, just wanted to bring it up - in catalogd all the feature gates have `Default: false` and in the Deployment spec for the controller manager we set the feature gates we want defaulting to true as such through the args field. 

For reference:
- Catalogd feature gates: https://github.com/operator-framework/catalogd/blob/main/pkg/features/features.go
- Catalogd manager Deployment setting the feature gates: https://github.com/operator-framework/catalogd/blob/8a90d8d92450efdc8d53661561028e4f1db749bb/config/manager/manager.yaml#L79

I don't have a strong opinion here, just wanted to point out the way that it has been done in catalogd in case we wanted to have consistency here (which isn't a huge deal to have some variation IMO)

everettraven (approved):

pkg/features/features.go
m1kola: Good call! I also though about this and would like to here opinions.

We want this to be enabled by default accordingly to [RFC](https://docs.google.com/document/d/1dfiNr0kvP1nfS0f9xaGiLyfrJ_KcD38clYs4LYBKllk/edit). We can achieve this by adding the flag to manifests similar to catalogd as mentioned above.

The slight benefit of having default set here is that we won't have to remember to add flags when running locally to get "default" behaviour.

I think both ways are fine and I can do it catalogd-way.

everettraven: I think the catalogd approach is nice because it forces enabled by default feature gates to be specified in the deployment for the controller manager. This gives some insights as to what feature gates are available and enabled by simply looking at the deployment/pod YAML. From a user perspective this is nice because I don't have to peruse the documentation every time to see what feature gates are available and whether or not they are enabled by default (IIRC in catalogd we even included disabled by default flags in the manifests, when we had them, and explicitly set them to "False").

m1kola: I think I'm now leaning toward doing it catalogd-way. I'll put this PR back into draft state and update it tomorrow.

Still if someone has any input on this topic - please share in the meantime.

m1kola: Updated

Issue #270, opened 2023-06-21
Surface information about installed version in Operator CR


anik120:
When an operator is installed, there is no information about which version of the operator was installed: 

```bash 
$ $  kubectl get operator argocd-operator -o yaml 
apiVersion: operators.operatorframework.io/v1alpha1
kind: Operator
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"operators.operatorframework.io/v1alpha1","kind":"Operator","metadata":{"annotations":{},"name":"argocd-operator"},"spec":{"packageName":"argocd-operator"}}
  creationTimestamp: "2023-06-21T14:57:50Z"
  generation: 1
  name: argocd-operator
  resourceVersion: "10690"
  uid: 6e0c67a5-eb9c-41c6-a455-140b28714d34
spec:
  packageName: argocd-operator
status:
  conditions:
  - lastTransitionTime: "2023-06-21T14:57:51Z"
    message: resolved to "quay.io/operatorhubio/argocd-operator@sha256:1a9b3c8072f2d7f4d6528fa32905634d97b7b4c239ef9887e3fb821ff033fef6"
    observedGeneration: 1
    reason: Success
    status: "True"
    type: Resolved
  - lastTransitionTime: "2023-06-21T14:57:57Z"
    message: installed from "quay.io/operatorhubio/argocd-operator@sha256:1a9b3c8072f2d7f4d6528fa32905634d97b7b4c239ef9887e3fb821ff033fef6"
    observedGeneration: 1
    reason: Success
    status: "True"
    type: Installed
  installedBundleResource: quay.io/operatorhubio/argocd-operator@sha256:1a9b3c8072f2d7f4d6528fa32905634d97b7b4c239ef9887e3fb821ff033fef6
  resolvedBundleResource: quay.io/operatorhubio/argocd-operator@sha256:1a9b3c8072f2d7f4d6528fa32905634d97b7b4c239ef9887e3fb821ff033fef6
```

One has to confirm the version installed by cross referencing the image reference in the `installedBundleResource` field with a matching image reference for a `BundleMetadata` 

```
$  kubectl get bundlemetadata operatorhubio-argocd-operator.v0.6.0 -o yaml | grep image 
  image: quay.io/operatorhubio/argocd-operator@sha256:1a9b3c8072f2d7f4d6528fa32905634d97b7b4c239ef9887e3fb821ff033fef6
[...]

PR #442, opened 2023-10-06
Add annotations to `BundleDeployment`

m1kola:
# Description

This adds annotations to `BundleDeployment` objects created by operator-controller in response to `Operator` objects.

This makes it easier to filter bundles during resolution and makes us less dependant on image bundle format.

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/controllers/operator_controller.go
+++ b/internal/controllers/operator_controller.go
@@ -47,6 +47,7 @@ import (
 	"github.com/operator-framework/operator-controller/internal/catalogmetadata"
 	"github.com/operator-framework/operator-controller/internal/controllers/validators"
 	olmvariables "github.com/operator-framework/operator-controller/internal/resolution/variables"
+	"github.com/operator-framework/operator-controller/internal/resolution/variablesources"
 )
 
 // OperatorReconciler reconciles a Operator object
@@ -179,9 +180,15 @@ func (r *OperatorReconciler) reconcile(ctx context.Context, op *operatorsv1alpha
 		setInstalledStatusConditionFailed(&op.Status.Conditions, err.Error(), op.GetGeneration())
 		return ctrl.Result{}, err
 	}
+
 	// Ensure a BundleDeployment exists with its bundle source from the bundle
 	// image we just looked up in the solution.
-	dep := r.generateExpectedBundleDeployment(*op, bundle.Image, bundleProvisioner)
+	dep, err := r.generateExpectedBundleDeployment(*op, bundle, bundleProvisioner)
+	if err != nil {
+		op.Status.InstalledBundleResource = ""
+		setInstalledStatusConditionFailed(&op.Status.Conditions, err.Error(), op.GetGeneration())
+		return ctrl.Result{}, err
+	}
 	if err := r.ensureBundleDeployment(ctx, dep); err != nil {
 		// originally Reason: operatorsv1alpha1.ReasonInstallationFailed
 		op.Status.InstalledBundleResource = ""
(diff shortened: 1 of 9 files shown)


internal/resolution/variablesources/installed_package.go
```diff
@@ -26,18 +29,20 @@ func (r *InstalledPackageVariableSource) GetVariables(ctx context.Context) ([]de
...
 	if len(resultSet) == 0 {
 		return nil, r.notFoundError()
 	}
-
-	// TODO: fast follow - we should check whether we are already supporting the channel attribute in the operator spec.
-	//       if so, we should take the value from spec of the operator CR in the owner ref of the bundle deployment.
-	//       If that channel is set, we need to update the filter above to filter by channel as well.
-	sort.SliceStable(resultSet, func(i, j int) bool {
-		return catalogsort.ByVersion(resultSet[i], resultSet[j])
-	})
```
m1kola, line 40: I think we no longer need sorting by version since we filter bundles by bundle version amongst other things. I assume previously we were doing sorting beucase multiple bundles could in theory use the same image (.e.g some sort of fake version).

internal/resolution/variablesources/installed_package.go
```diff
@@ -26,18 +29,20 @@ func (r *InstalledPackageVariableSource) GetVariables(ctx context.Context) ([]de
...
+		catalogfilter.WithName(r.bundleName),
+		catalogfilter.InMastermindsSemverRange(vr),
+	))
 	if len(resultSet) == 0 {
 		return nil, r.notFoundError()
 	}
-
-	// TODO: fast follow - we should check whether we are already supporting the channel attribute in the operator spec.
-	//       if so, we should take the value from spec of the operator CR in the owner ref of the bundle deployment.
-	//       If that channel is set, we need to update the filter above to filter by channel as well.
```
m1kola, line 37: I believe this is no longer relevant given that we filter by package name, bundle name and bundle version.

internal/resolution/variablesources/bundle_deployment.go
```diff
@@ -36,20 +42,15 @@ func (o *BundleDeploymentVariableSource) GetVariables(ctx context.Context) ([]de
...
-			ips, err := NewInstalledPackageVariableSource(o.catalogClient, bundleDeployment.Spec.Template.Spec.Source.Image.Ref)
-			if err != nil {
-				return nil, err
-			}
-			variableSources = append(variableSources, ips)
+		pkgName := bundleDeployment.Annotations[AnnotationPackageName]
+		bundleName := bundleDeployment.Annotations[AnnotationBundleName]
+		bundleVersion := bundleDeployment.Annotations[AnnotationBundleVersion]
+		if pkgName == "" || bundleName == "" || bundleVersion == "" {
+			continue
```
m1kola, line 50: Do we need to care about bundle deployments created directly? These annotations will only be present on bundle deployments reconciled by operator-controller.

internal/controllers/operator_controller.go
```diff
@@ -264,18 +271,28 @@ func (r *OperatorReconciler) bundleFromSolution(solution *solver.Solution, packa
...
 	bd := &unstructured.Unstructured{Object: map[string]interface{}{
 		"apiVersion": rukpakv1alpha1.GroupVersion.String(),
 		"kind":       rukpakv1alpha1.BundleDeploymentKind,
 		"metadata": map[string]interface{}{
 			"name": o.GetName(),
+			"annotations": map[string]string{
+				variablesources.AnnotationPackageName:   bundle.Package,
+				variablesources.AnnotationBundleName:    bundle.Name,
+				variablesources.AnnotationBundleVersion: bundleVersion.String(),
+			},
```
joelanford, line 295: Should we make these labels (rather than annotations) so that we can use a label selector to populate our BD cache with _just_ BDs that we are managing?

internal/resolution/variablesources/bundle_deployment.go
joelanford: For now, I don't think we need to care, at least in this context.

I sorta think this context is: is there a BD _for this operator_ already? (And maybe that means we should have a label on the BD like `operators.operatorframework.io/operator-name` whose value is `Operator.metadata.name`?

Perhaps in the future, there's a separate feature that helps the resolver become aware of other BDs, but I don't think we need to do that now.

internal/resolution/variablesources/installed_package.go
joelanford: Got it. It looks like we check the result set for 0 bundles, but not more than 1. Can we add another check that errors if there is more than 1 bundle?

internal/controllers/operator_controller.go
```diff
@@ -264,18 +271,28 @@ func (r *OperatorReconciler) bundleFromSolution(solution *solver.Solution, packa
...
+		return nil, err
+	}
+
 	bd := &unstructured.Unstructured{Object: map[string]interface{}{
 		"apiVersion": rukpakv1alpha1.GroupVersion.String(),
 		"kind":       rukpakv1alpha1.BundleDeploymentKind,
 		"metadata": map[string]interface{}{
 			"name": o.GetName(),
+			"annotations": map[string]string{
+				variablesources.AnnotationPackageName:   bundle.Package,
```
stevekuznetsov, line 292: If this is a 1:1 relationship, why not an OwnerRef or similar?

m1kola: `bundle` here is not a kubernetes CR.

internal/controllers/operator_controller.go
m1kola: @joelanford I think we already do that but in a different way.

We set owner reference on a BD here:

https://github.com/operator-framework/operator-controller/blob/205a48778e07f96819381269cb652536c6f895e6/internal/controllers/operator_controller.go#L297-L306

And only reconcile owned objects:

https://github.com/operator-framework/operator-controller/blob/205a48778e07f96819381269cb652536c6f895e6/internal/controllers/operator_controller.go#L316

joelanford: Kubernetes does not support setting up informers to filter on the server-side using owner references though. So my suggestion is to use labels so that we can _also_ configure the cache to select only BDs that have the expected label set.

m1kola:
I've been thinking more about this:

1. In `main` we have `BundleDeploymentVariableSource` which looks into `.Spec.Template.Spec.Source.Image.Ref` to find an image we currently have installed. I think it should look at the actual state instead of a desired state. We already have `.Status.InstalledBundleResource` on `Operator` and can probably use this instead.
2. I think instead of adding labels/annotations into `BundleDeployment`s we should probably add this information into `Operator`'s status. This will also solve #270

I'll bring this up for a discussion and will create a new PR or repoen this one based on the outcome.

Closing for now.

Issue #377, opened 2023-08-31
[Epic] Initial upgrade support


joelanford:
### Design docs

- [Brief: Initial upgrade support](https://docs.google.com/document/d/1KIhV8oqr22Bp8lf370Aej2n-3uW_xGcTptONHHPRVYA/edit#heading=h.btd8kkom5km4)
- [RFC: Initial upgrade support](https://docs.google.com/document/d/1dfiNr0kvP1nfS0f9xaGiLyfrJ_KcD38clYs4LYBKllk/edit)

### Precursor and related items
- #231 
- #267
- #275

### Task List
- #268 
- #395
- #396
- #397
- #398
- #449
- #470

m1kola:
Added #449 since I realised that we do not cover `.spec.upgradeConstraintPolicy` in any other issue.

m1kola:
Added #470 for documentation.

PR #444, opened 2023-10-06
Add initial SemVer upgrade support

### LINKED ISSUE #398: Add semver upgrade edge generation support

Need to implement semver upgrade edge generation function as discussed in the [RFC](https://docs.google.com/document/d/1dfiNr0kvP1nfS0f9xaGiLyfrJ_KcD38clYs4LYBKllk/edit#heading=h.zfkgmgda4f9w).

> ### Semver
> Semver based implementation is expected to:
> * Satisfy the interface discussed above
> * Return a list of successor bundles based on semver. The list will not include bundle versions outside the major version of the installed bundle. (In [semver](https://semver.org/#if-even-the-tiniest-backward-incompatible-changes-to-the-public-api-require-a-major-version-bump-wont-i-end-up-at-version-4200-very-rapidly) a new major version is intended to indicate breaking changes, and we don’t want to default to that behavior. Future iterations may include knobs to extend this behavior.)
> * This mode will not allow downgrading.

As part of this issue we also need to make sure that `ForceSemverUpgradeConstraints` feature gate is respected:

> The solver needs to be able to switch between upgrade constraint semantics based on rules defined above (see Fig 1).

m1kola:
# Description

OLM will now use SemVer to determine next upgrade. In this iteration we only support upgrade within the same major versions: e.g 1.0.1 can be upgraded to 1.0.2 or 1.3.2, but not 2.0.0.

Closes #398

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,8 +20,8 @@ export XDG_DATA_HOME ?= /tmp/.local/share
 include .bingo/Variables.mk
 
 # ARTIFACT_PATH is the absolute path to the directory where the operator-controller e2e tests will store the artifacts
-# for example: ARTIFACT_PATH=/tmp/artifacts make test 
-export ARTIFACT_PATH ?= 
+# for example: ARTIFACT_PATH=/tmp/artifacts make test
+export ARTIFACT_PATH ?=
 
 OPERATOR_CONTROLLER_NAMESPACE ?= operator-controller-system
 KIND_CLUSTER_NAME ?= operator-controller
@@ -156,11 +156,19 @@ kind-load-test-artifacts: $(KIND) #EXHELP Load the e2e testdata container images
 	$(CONTAINER_RUNTIME) build $(TESTDATA_DIR)/bundles/registry-v1/prometheus-operator.v0.37.0 -t localhost/testdata/bundles/registry-v1/prometheus-operator:v0.37.0
 	$(CONTAINER_RUNTIME) build $(TESTDATA_DIR)/bundles/registry-v1/prometheus-operator.v0.47.0 -t localhost/testdata/bundles/registry-v1/prometheus-operator:v0.47.0
 	$(CONTAINER_RUNTIME) build $(TESTDATA_DIR)/bundles/registry-v1/prometheus-operator.v0.65.1 -t localhost/testdata/bundles/registry-v1/prometheus-operator:v0.65.1
+	$(CONTAINER_RUNTIME) tag localhost/testdata/bundles/registry-v1/prometheus-operator:v0.65.1 localhost/testdata/bundles/registry-v1/prometheus-operator:v1.0.0
+	$(CONTAINER_RUNTIME) tag localhost/testdata/bundles/registry-v1/prometheus-operator:v0.65.1 localhost/testdata/bundles/registry-v1/prometheus-operator:v1.0.1
+	$(CONTAINER_RUNTIME) tag localhost/testdata/bundles/registry-v1/prometheus-operator:v0.65.1 localhost/testdata/bundles/registry-v1/prometheus-operator:v1.2.0
+	$(CONTAINER_RUNTIME) tag localhost/testdata/bundles/registry-v1/prometheus-operator:v0.65.1 localhost/testdata/bundles/registry-v1/prometheus-operator:v2.0.0
 	$(CONTAINER_RUNTIME) build $(TESTDATA_DIR)/bundles/plain-v0/plain.v0.1.0 -t localhost/testdata/bundles/plain-v0/plain:v0.1.0
 	$(CONTAINER_RUNTIME) build $(TESTDATA_DIR)/catalogs -f $(TESTDATA_DIR)/catalogs/test-catalog.Dockerfile -t localhost/testdata/catalogs/test-catalog:e2e
 	$(KIND) load docker-image localhost/testdata/bundles/registry-v1/prometheus-operator:v0.37.0 --name $(KIND_CLUSTER_NAME)
 	$(KIND) load docker-image localhost/testdata/bundles/registry-v1/prometheus-operator:v0.47.0 --name $(KIND_CLUSTER_NAME)
 	$(KIND) load docker-image localhost/testdata/bundles/registry-v1/prometheus-operator:v0.65.1 --name $(KIND_CLUSTER_NAME)
+	$(KIND) load docker-image localhost/testdata/bundles/registry-v1/prometheus-operator:v1.0.0 --name $(KIND_CLUSTER_NAME)
+	$(KIND) load docker-image localhost/testdata/bundles/registry-v1/prometheus-operator:v1.0.1 --name $(KIND_CLUSTER_NAME)
+	$(KIND) load docker-image localhost/testdata/bundles/registry-v1/prometheus-operator:v1.2.0 --name $(KIND_CLUSTER_NAME)
+	$(KIND) load docker-image localhost/testdata/bundles/registry-v1/prometheus-operator:v2.0.0 --name $(KIND_CLUSTER_NAME)
 	$(KIND) load docker-image localhost/testdata/bundles/plain-v0/plain:v0.1.0 --name $(KIND_CLUSTER_NAME)
 	$(KIND) load docker-image localhost/testdata/catalogs/test-catalog:e2e --name $(KIND_CLUSTER_NAME)
 
--- a/internal/resolution/variablesources/installed_package.go
+++ b/internal/resolution/variablesources/installed_package.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"sort"
 
+	mmsemver "github.com/Masterminds/semver/v3"
 	"github.com/operator-framework/deppy/pkg/deppy"
 	"github.com/operator-framework/deppy/pkg/deppy/input"
 
@@ -12,6 +13,7 @@ import (
 	catalogfilter "github.com/operator-framework/operator-controller/internal/catalogmetadata/filter"
 	catalogsort "github.com/operator-framework/operator-controller/internal/catalogmetadata/sort"
 	"github.com/operator-framework/operator-controller/internal/resolution/variables"
+	"github.com/operator-framework/operator-controller/pkg/features"
 )
 
 var _ input.VariableSource = &InstalledPackageVariableSource{}
@@ -59,10 +61,15 @@ func (r *InstalledPackageVariableSource) notFoundError() error {
 }
 
 func NewInstalledPackageVariableSource(catalogClient BundleProvider, bundleImage string) (*InstalledPackageVariableSource, error) {
+	successors := legacySemanticsSuccessors
+	if features.OperatorControllerFeatureGate.Enabled(features.ForceSemverUpgradeConstraints) {
+		successors = semverSuccessors
+	}
+
 	return &InstalledPackageVariableSource{
 		catalogClient: catalogClient,
 		bundleImage:   bundleImage,
-		successors:    legacySemanticsSuccessors,
+		successors:    successors,
 	}, nil
 }
 
@@ -77,6 +84,31 @@ func legacySemanticsSuccessors(allBundles []*catalogmetadata.Bundle, installedBu
 	// find the bundles that replace the bundle provided
 	// TODO: this algorithm does not yet consider skips and skipRange
 	upgradeEdges := catalogfilter.Filter(allBundles, catalogfilter.Replaces(installedBundle.Name))
+	sort.SliceStable(upgradeEdges, func(i, j int) bool {
+		return catalogsort.ByVersion(upgradeEdges[i], upgradeEdges[j])
+	})
+
+	return upgradeEdges, nil
+}
+
+// semverSuccessors returns successors based on Semver.
+// Successors will not include versions outside the major version of the
+// installed bundle as major version is intended to indicate breaking changes.
+func semverSuccessors(allBundles []*catalogmetadata.Bundle, installedBundle *catalogmetadata.Bundle) ([]*catalogmetadata.Bundle, error) {
+	currentVersion, err := installedBundle.Version()
+	if err != nil {
+		return nil, err
+	}
+
+	// Based on current version create a caret range comparison constraint
+	// to allow only minor and patch version as successors and exclude current version.
+	constraintStr := fmt.Sprintf("^%s, != %s", currentVersion.String(), currentVersion.String())
+	wantedVersionRangeConstraint, err := mmsemver.NewConstraint(constraintStr)
+	if err != nil {
+		return nil, err
+	}
+
+	upgradeEdges := catalogfilter.Filter(allBundles, catalogfilter.InMastermindsSemverRange(wantedVersionRangeConstraint))
 	sort.SliceStable(upgradeEdges, func(i, j int) bool {
 		return catalogsort.ByVersion(upgradeEdges[i], upgradeEdges[j])
 	})
--- a/internal/resolution/variablesources/installed_package_test.go
+++ b/internal/resolution/variablesources/installed_package_test.go
@@ -121,6 +121,27 @@ func TestInstalledPackageVariableSource(t *testing.T) {
 	const bundleImage = "registry.io/repo/test-package@v2.0.0"
 	fakeCatalogClient := testutil.NewFakeCatalogClient(bundleList)
 
+	t.Run("with ForceSemverUpgradeConstraints feature gate enabled", func(t *testing.T) {
+		defer featuregatetesting.SetFeatureGateDuringTest(t, features.OperatorControllerFeatureGate, features.ForceSemverUpgradeConstraints, true)()
+
+		ipvs, err := variablesources.NewInstalledPackageVariableSource(&fakeCatalogClient, bundleImage)
+		require.NoError(t, err)
+
+		variables, err := ipvs.GetVariables(context.TODO())
+		require.NoError(t, err)
+		require.Len(t, variables, 1)
+		packageVariable, ok := variables[0].(*olmvariables.InstalledPackageVariable)
+		assert.True(t, ok)
+		assert.Equal(t, deppy.IdentifierFromString("installed package test-package"), packageVariable.Identifier())
+
+		// ensure bundles are in version order (high to low)
+		bundles := packageVariable.Bundles()
+		require.Len(t, bundles, 3)
+		assert.Equal(t, "test-package.v2.2.0", packageVariable.Bundles()[0].Name)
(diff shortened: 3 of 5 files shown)


Makefile
```diff
@@ -156,11 +156,19 @@ kind-load-test-artifacts: $(KIND) #EXHELP Load the e2e testdata container images
 	$(CONTAINER_RUNTIME) build $(TESTDATA_DIR)/bundles/registry-v1/prometheus-operator.v0.37.0 -t localhost/testdata/bundles/registry-v1/prometheus-operator:v0.37.0
 	$(CONTAINER_RUNTIME) build $(TESTDATA_DIR)/bundles/registry-v1/prometheus-operator.v0.47.0 -t localhost/testdata/bundles/registry-v1/prometheus-operator:v0.47.0
 	$(CONTAINER_RUNTIME) build $(TESTDATA_DIR)/bundles/registry-v1/prometheus-operator.v0.65.1 -t localhost/testdata/bundles/registry-v1/prometheus-operator:v0.65.1
+	$(CONTAINER_RUNTIME) build $(TESTDATA_DIR)/bundles/registry-v1/prometheus-operator.v0.65.1 -t localhost/testdata/bundles/registry-v1/prometheus-operator:v1.0.0
+	$(CONTAINER_RUNTIME) build $(TESTDATA_DIR)/bundles/registry-v1/prometheus-operator.v0.65.1 -t localhost/testdata/bundles/registry-v1/prometheus-operator:v1.0.1
+	$(CONTAINER_RUNTIME) build $(TESTDATA_DIR)/bundles/registry-v1/prometheus-operator.v0.65.1 -t localhost/testdata/bundles/registry-v1/prometheus-operator:v1.2.0
+	$(CONTAINER_RUNTIME) build $(TESTDATA_DIR)/bundles/registry-v1/prometheus-operator.v0.65.1 -t localhost/testdata/bundles/registry-v1/prometheus-operator:v2.0.0
```
m1kola, line 162: I'm cheating here: I'm using 0.65.1 manifests to create fake 1.0.0, 1.0.1, 1.2.0 and 2.0.0 versions.

I don't like it, but I also do not like adding almost 10mb of yamls into the repo when they do not make any difference for my test.

Makefile
```diff
@@ -104,7 +104,7 @@ test: manifests generate fmt vet test-unit test-e2e #HELP Run all tests.
 
 .PHONY: e2e
 FOCUS := $(if $(TEST),-v -focus "$(TEST)")
-E2E_FLAGS ?= ""
+E2E_FLAGS ?= --label-filter="!ForceSemverUpgradeConstraints=false"
```
m1kola, line 107: Created #447 for better switching between feature gate states.

For now: we have E2E tests for both true and false values. But for now we will be excluding tests for `ForceSemverUpgradeConstraints=false` feature gate state from run since default is `ForceSemverUpgradeConstraints=true`.

internal/resolution/variablesources/installed_package.go
```diff
@@ -83,3 +90,28 @@ func legacySemanticsSuccessors(allBundles []*catalogmetadata.Bundle, installedBu
...
+// Successors will not include versions outside the major version of the
+// installed bundle as major version is intended to indicate breaking changes.
+func semverSuccessors(allBundles []*catalogmetadata.Bundle, installedBundle *catalogmetadata.Bundle) ([]*catalogmetadata.Bundle, error) {
+	currentVersion, err := installedBundle.Version()
+	if err != nil {
+		return nil, err
+	}
+
+	// Based on current version create a caret range comparison constraint
+	// to allow only major and patch version as successors and exclude current version.
```
m1kola, line 104: Re: "exclude current version" see https://github.com/operator-framework/operator-controller/pull/444#discussion_r1350400240

m1kola (commented):
I think it is ready for review now

internal/resolution/variablesources/installed_package.go
```diff
@@ -5,13 +5,15 @@ import (
 	"fmt"
 	"sort"
 
+	mmsemver "github.com/Masterminds/semver/v3"
```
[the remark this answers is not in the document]
m1kola: Yes, we have places like this. Example:

https://github.com/operator-framework/operator-controller/blob/205a48778e07f96819381269cb652536c6f895e6/internal/catalogmetadata/filter/bundle_predicates.go#L4-L5

`github.com/Masterminds/semver` was chosen for `Operator` API so users can specify version ranges for upgrades. `github.com/blang/semver` is used in catalog properties (in `olm.package.required`).

Makefile
stevekuznetsov: Unless it's somehow very challenging, IMHO makes a lot of sense to only check in tests that run - is it a big lift to get working for this pR?

internal/resolution/variablesources/installed_package.go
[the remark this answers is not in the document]
stevekuznetsov: Yikes.

Makefile
m1kola: > Unless it's somehow very challenging, IMHO makes a lot of sense to only check in tests that run

Just to clarify: tests we no longer run in this PR exist in `main`. What I did:
* Moved existing tests around
* Labeled them
* Excluded from execution since default behaviour changed.

Didn't want to remove existing tests because I hope that we will eventually run them (#447).

> is it a big lift to get working for this pR?

I would like to keep this PR focused on semver and leave E2E.

But I'm happy to look into this as a separate effort. If you have any suggestions how to do it - could you please point into the right direction on #447?

internal/resolution/variablesources/installed_package.go
```diff
@@ -60,8 +61,65 @@ func (r *InstalledPackageVariableSource) notFoundError() error {
...
+
+	// Turn current version into a constraint
+	currentVersionConstraint, err := mmsemver.NewConstraint(currentVersion.String())
+	if err != nil {
+		return nil, err
+	}
+
+	upgradeEdges := catalogfilter.Filter(allBundles, catalogfilter.And(
+		catalogfilter.InMastermindsSemverRange(wantedVersionRangeConstraint),
+		catalogfilter.Not(catalogfilter.InMastermindsSemverRange(currentVersionConstraint)),
```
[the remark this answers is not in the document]
grokspawn: I like how it's clear in `GetVariables` that we get the slice of successors and then explicitly include the currently-installed, but at the end of the slice so that it is presumably the least-preferred of all options.

internal/resolution/variablesources/installed_package.go
grokspawn, line 104:
```suggestion
	// to allow only minor and patch version as successors and exclude current version.
```

internal/resolution/variablesources/installed_package.go
grokspawn, line 1: I was looking to validate that when the operator has set `UpgradeConstraintPolicy == Ignore` that there was no impact (and there wasn't ✔️), but then I looked in `legacySemanticsSuccessors` and saw that there also was no support for this.  
I also wasn't able to spot an issue in operator-controller which discussed it.  
Is there already an issue to use that strategy instead of either of the defined successor functions?

m1kola: @grokspawn good eye for details! :) Yesterday I created #449 for this. I forgot to create it initially when I was breaking RFC down into tickets, but we now have it tracked in issues.

Epic #377 contains all the related issues for this.

Makefile
```diff
@@ -20,8 +20,8 @@ export XDG_DATA_HOME ?= /tmp/.local/share
 include .bingo/Variables.mk
 
 # ARTIFACT_PATH is the absolute path to the directory where the operator-controller e2e tests will store the artifacts
-# for example: ARTIFACT_PATH=/tmp/artifacts make test 
-export ARTIFACT_PATH ?= 
+# for example: ARTIFACT_PATH=/tmp/artifacts make test
+export ARTIFACT_PATH ?=
```
tmshort, line 24: And yet, Makefiles are one of the few types of files where end-of-line whitespace may be significant!

Makefile
tmshort: Actually not a bad idea, to generate the test data from a template.

internal/resolution/variablesources/installed_package.go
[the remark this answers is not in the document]
tmshort: I added masterminds, and was hoping to drop blang, but there are ever-so-slight differences in acceptable syntax that we didn't get rid of blang because it processes and external, well-defined resource. We probably could get rid of blang if we convert the syntax.

Makefile
m1kola: My editor removed trailing spaces automatically here. But I believe it is not significant here.

Makefile
joelanford: What about using unit tests of the Reconcile function to test both feature gate states?

Then the e2e can simply test the default state of the feature gate.

joelanford: As an aside: I have a pretty strong opinion that we should test as much as possible in `Reconcile` tests so that our e2e's remain light and fast and relatively resistant to change.

m1kola: > What about using [unit tests of the Reconcile function](https://github.com/operator-framework/operator-controller/blob/main/internal/controllers/operator_controller_test.go) to test both feature gate states?
> 
> Then the e2e can simply test the default state of the feature gate.

We do that on the variable source level already [here](https://github.com/operator-framework/operator-controller/pull/444/files#diff-d617afce32874544d5987c6a05c4b55525d66d1150c18330e9e70352d7959698R125) (there is another test in this file for disabled state which is no in the diff).

internal/resolution/variablesources/installed_package.go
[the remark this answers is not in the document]
joelanford: I see. In that case, maybe we use a different `wantedVersionRangeConstraint` that excludes the currently installed version from the get-go?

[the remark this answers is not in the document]
joelanford: Oh sweet. I see that's what you already did in the most recent commits! 🚀

Makefile
tmshort: No, it's not... but it's just another Makefile-ism.

Makefile
joelanford: Use the `tag` subcommand instead of re-building? Caches probably make this _not that bad_ with a re-build, but re-tagging would definitely be faster.

Makefile
joelanford: My point is: rather than continuing to add more e2e tests, can we instead add more controller reconcile tests?

Is there a reason we think what's being tested in the new e2es from this PR can't instead be tested in controller reconcile tests?

Incidentally, it would solve this problem with inert tests because we can really easily enable/disable the gate in those tests.

m1kola (commented):
@joelanford please take another look. I updated `Makefile` to use tags instead of builds.

Also update the PR to address feedback on testing (see comment below).

internal/resolution/variablesources/installed_package.go
```diff
@@ -83,3 +90,28 @@ func legacySemanticsSuccessors(allBundles []*catalogmetadata.Bundle, installedBu
...
+// installed bundle as major version is intended to indicate breaking changes.
+func semverSuccessors(allBundles []*catalogmetadata.Bundle, installedBundle *catalogmetadata.Bundle) ([]*catalogmetadata.Bundle, error) {
+	currentVersion, err := installedBundle.Version()
+	if err != nil {
+		return nil, err
+	}
+
+	// Based on current version create a caret range comparison constraint
+	// to allow only minor and patch version as successors and exclude current version.
+	constraintStr := fmt.Sprintf("^%s, != %s", currentVersion.String(), currentVersion.String())
```
awgreene, line 105: Based on [this doc](https://docs.npmjs.com/cli/v6/using-npm/semver#caret-ranges-123-025-004), a caret range will prevent the left-most non-zero digit in the `[major, minor, patch]` tuple.
In other words, this allows:
>1. patch and minor updates for versions 1.0.0 and above
>2. patch updates for versions 0.X >=0.1.0
>3. no updates for versions 0.0.X.

What happens in cases 2 and 3? I think we'll need good documentation to address these cases.

Case 2 doesn't seem too concerning given this comment from the doc linked earlier:
>Many authors treat a 0.x version as if the x were the major "breaking-change" indicator.

Case 3 seems more concerning, I suspect that many people packaging an operator for the first time will use version `0.0.1` and fail to upgrade it.

internal/resolution/variablesources/installed_package_test.go
```diff
@@ -121,6 +121,27 @@ func TestInstalledPackageVariableSource(t *testing.T) {
...
+		packageVariable, ok := variables[0].(*olmvariables.InstalledPackageVariable)
+		assert.True(t, ok)
+		assert.Equal(t, deppy.IdentifierFromString("installed package test-package"), packageVariable.Identifier())
+
+		// ensure bundles are in version order (high to low)
+		bundles := packageVariable.Bundles()
+		require.Len(t, bundles, 3)
+		assert.Equal(t, "test-package.v2.2.0", packageVariable.Bundles()[0].Name)
+		assert.Equal(t, "test-package.v2.1.0", packageVariable.Bundles()[1].Name)
+		assert.Equal(t, "test-package.v2.0.0", packageVariable.Bundles()[2].Name)
```
awgreene, line 142: Is the caret inclusive?

test/e2e/install_test.go
```diff
@@ -180,48 +180,56 @@ var _ = Describe("Operator Install", func() {
...
-				g.Expect(c.Get(ctx, types.NamespacedName{Name: operator.Name}, operator)).To(Succeed())
-				cond := apimeta.FindStatusCondition(operator.Status.Conditions, operatorv1alpha1.TypeResolved)
-				g.Expect(cond).ToNot(BeNil())
-				g.Expect(cond.Reason).To(Equal(operatorv1alpha1.ReasonResolutionFailed))
-				g.Expect(cond.Message).To(MatchRegexp(`^constraints not satisfiable:.*; installed package prometheus requires at least one of.*0.47.0[^,]*,[^,]*0.37.0[^;]*;.*`))
-				g.Expect(operator.Status.ResolvedBundleResource).To(BeEmpty())
-			}).Should(Succeed())
+			It("does not allow to upgrade the Operator to a non-successor version", func() {
+				By("updating the Operator resource to a non-successor version")
+				// Semver only allows upgrades within major version at the moment.
```
awgreene, line 204: How are we enforcing SemVer upgrades in this test?

test/e2e/install_test.go
```diff
@@ -180,48 +180,56 @@ var _ = Describe("Operator Install", func() {
...
+				operator.Spec.Version = "1.2.0"
+				Expect(c.Update(ctx, operator)).To(Succeed())
+				By("eventually reporting a successful resolution and bundle path")
+				Eventually(func(g Gomega) {
+					g.Expect(c.Get(ctx, types.NamespacedName{Name: operator.Name}, operator)).To(Succeed())
+					cond := apimeta.FindStatusCondition(operator.Status.Conditions, operatorv1alpha1.TypeResolved)
+					g.Expect(cond).ToNot(BeNil())
+					g.Expect(cond.Reason).To(Equal(operatorv1alpha1.ReasonSuccess))
+					g.Expect(cond.Message).To(ContainSubstring("resolved to"))
+					g.Expect(operator.Status.ResolvedBundleResource).To(Equal("localhost/testdata/bundles/registry-v1/prometheus-operator:v1.2.0"))
```
awgreene, line 230: From above:
>// Test catalog has versions between the initial version and new version

This doesn't seem to be a new version.

test/e2e/install_test.go
```diff
@@ -180,48 +180,56 @@ var _ = Describe("Operator Install", func() {
...
+				// Semver only allows upgrades within major version at the moment.
+				operator.Spec.Version = "2.0.0"
+				Expect(c.Update(ctx, operator)).To(Succeed())
+				By("eventually reporting an unsatisfiable resolution")
+				Eventually(func(g Gomega) {
+					g.Expect(c.Get(ctx, types.NamespacedName{Name: operator.Name}, operator)).To(Succeed())
+					cond := apimeta.FindStatusCondition(operator.Status.Conditions, operatorv1alpha1.TypeResolved)
+					g.Expect(cond).ToNot(BeNil())
+					g.Expect(cond.Reason).To(Equal(operatorv1alpha1.ReasonResolutionFailed))
+					g.Expect(cond.Message).To(MatchRegexp(`^constraints not satisfiable:.*; installed package prometheus requires at least one of.*1.2.0[^;]*;.*`))
```
awgreene, line 213: What does this error message look like in practice?

awgreene (commented):
I have a couple of open questions regarding the user experience around `0.minor.patch` and `0.0.patch` releases, but overall this looks great.

test/e2e/install_test.go
```diff
@@ -180,48 +180,56 @@ var _ = Describe("Operator Install", func() {
...
-			By("eventually reporting a successful resolution")
-			Eventually(func(g Gomega) {
-				g.Expect(c.Get(ctx, types.NamespacedName{Name: operator.Name}, operator)).To(Succeed())
-				cond := apimeta.FindStatusCondition(operator.Status.Conditions, operatorv1alpha1.TypeResolved)
-				g.Expect(cond).ToNot(BeNil())
-				g.Expect(cond.Reason).To(Equal(operatorv1alpha1.ReasonSuccess))
-				g.Expect(cond.Message).To(ContainSubstring("resolved to"))
-				g.Expect(operator.Status.ResolvedBundleResource).ToNot(BeEmpty())
-			}).Should(Succeed())
+		When("resolving upgrade edges", func() {
```
awgreene, line 183: Could we add some `0.minor.patch` and `0.0.patch` version upgrades? Would be great to codify expectations.

m1kola (commented):
Updated the PR to get rid of `MatchRegexp`. Answered some questions.

internal/resolution/variablesources/installed_package.go
awgreene: Beauty!

test/e2e/install_test.go
awgreene: Ah I see, thanks!

test/e2e/install_test.go
awgreene: Ah I see, thanks!

awgreene (approved):
/approve

internal/resolution/variablesources/installed_package_test.go
awgreene: Got it, thanks!

test/e2e/install_test.go
m1kola: I ended up doing a separate PR here: https://github.com/operator-framework/operator-controller/pull/472

PR #448, opened 2023-10-09
Bump rukpak to 0.14.0

### LINKED ISSUE #446: Bump rukpak to v0.14.0
Rukpak v0.14.0 adds a new `Healthy` condition to BundleDeployments. It would be great to pull this version of Rukpak into Operator Controller so that users have more visibility into the state of their cluster.

Follow-ups that we should capture for after bumping the version:
1. Sync the BD healthy condition into Operator status, so that clients interacting solely with the Operator API have information about the health of the operator.
2. Consider BD health when determining whether an upgrade should proceed. If the BD for an Operator is not healthy, don't upgrade.
3. Add a knob in the Operator API that allows a cluster admin to proceed with an upgrade even if the BD is unhealthy.

See #434 for dependabot's rukpak version bump attempt. It looks like we have some tests that are overly particular about the BD conditions.

kevinrizza:
Modify tests to check for new Health condition

Resolves #446


test/e2e/install_test.go
```diff
@@ -123,9 +128,14 @@ var _ = Describe("Operator Install", func() {
...
-					g.Expect(bd.Status.Conditions[0].Reason).To(Equal("UnpackSuccessful"))
-					g.Expect(bd.Status.Conditions[1].Reason).To(Equal("InstallationSucceeded"))
+					for _, condition := range bd.Status.Conditions {
+						switch conditionType := condition.Type; conditionType {
+						case "HasValidBundle":
+							g.Expect(condition.Reason).To(Equal("UnpackSuccessful"))
+						case "Installed":
+							g.Expect(condition.Reason).To(Equal("InstallationSucceeded"))
+						}
+					}
```
joelanford, line 138: Sorry to be a bother, but if there are no conditions set, this will still pass, and I'm not sure that's what we want either?

It seems like we may want to do this:
```go
hasValidBundle := meta.FindStatusCondition(bd.Status.Conditions, rukpakv1alpha1.TypeHasValidBundle)
g.Expect(hasValidBundle).ToNot(BeNil())
g.Expect(hasValidBundle.Reason).To(Equal(rukpakv1alpha1.ReasonUnpackSuccessful))
```

And then a similar thing for the `Installed` condition.

joelanford (approved):

PR #450, opened 2023-10-10
Refactor legacy upgrade constraint semantics

m1kola:
# Description

* Make successors func swappable
* Update test to not use Ginkgo & prepare the test to be able to switch between different states of the `ForceSemverUpgradeConstraints` feature flag

## Reviewer Checklist

- [ ] API Go Documentation


internal/resolution/variablesources/installed_package.go
```diff
@@ -63,5 +62,24 @@ func NewInstalledPackageVariableSource(catalogClient BundleProvider, bundleImage
 	return &InstalledPackageVariableSource{
 		catalogClient: catalogClient,
 		bundleImage:   bundleImage,
+		successors:    legacySemanticsSuccessors,
```
m1kola, line 65: Context: this will be swapable based on feature gate. More to come in #444.

internal/resolution/variablesources/installed_package_test.go
```diff
@@ -3,116 +3,140 @@ package variablesources_test
...
+		assert.NoError(t, err)
 
-	It("should return the correct package variable", func() {
 		variables, err := ipvs.GetVariables(context.TODO())
-		Expect(err).NotTo(HaveOccurred())
-		Expect(variables).To(HaveLen(1))
-		reqPackageVar, ok := variables[0].(*olmvariables.InstalledPackageVariable)
-		Expect(ok).To(BeTrue())
-		Expect(reqPackageVar.Identifier()).To(Equal(deppy.IdentifierFromString("installed package test-package")))
+		assert.NoError(t, err)
```
joelanford, line 130: Might need to use the `require` package rather than the `assert` package if the test is unable to continue when an assertion fails.

joelanford (commented):
Looks good to me, after addressing my comment about use of `require` vs `assert` testify packages in the tests.

internal/resolution/variablesources/installed_package_test.go
m1kola: I had no idea `require` package existed :) Good to know. Updated.

PR #453, opened 2023-10-11
Reduce variable count

### LINKED ISSUE #414: Reduce number of resolution variables to one per bundle
Today one bundle in the catalog may result in multiple variables for the solver due to the fact that we create one variable per bundle occurrence in each channel. For example, If package `foo` has bundle `foo.v1` which occurs in channels `beta` and `stable`, we create two variables for the solver.

This is the case currently in `main` branch where we create entities (which later get translated into variables) [here](https://github.com/operator-framework/operator-controller/blob/3a8f2a000096828c40367baca478a19b7dbc632e/internal/resolution/entitysources/catalogdsource.go#L143-L147) and in the upcoming PR which replaces entities & entity sources (see [this discussion](https://github.com/operator-framework/operator-controller/pull/413#discussion_r1326532379)).

This issue is to look into whether we can reduce a number of variables we create to one per bundle.

This should be possible because we should be doing all the filtering of bundles by channel before we create a variable for the solver.

m1kola:
# Description

After this commit we will be creating one bundle per package. Previously we we creating one variable per bundle occurance in a channel which is currently unnecessary.

Closes #414

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/resolution/variables/bundle.go
+++ b/internal/resolution/variables/bundle.go
@@ -26,17 +26,17 @@ func (b *BundleVariable) Dependencies() []*catalogmetadata.Bundle {
 	return b.dependencies
 }
 
-func NewBundleVariable(id deppy.Identifier, bundle *catalogmetadata.Bundle, dependencies []*catalogmetadata.Bundle) *BundleVariable {
-	var dependencyIDs []deppy.Identifier
+func NewBundleVariable(bundle *catalogmetadata.Bundle, dependencies []*catalogmetadata.Bundle) *BundleVariable {
+	dependencyIDs := make([]deppy.Identifier, 0, len(dependencies))
 	for _, bundle := range dependencies {
-		dependencyIDs = append(dependencyIDs, BundleToBundleVariableIDs(bundle)...)
+		dependencyIDs = append(dependencyIDs, BundleVariableID(bundle))
 	}
 	var constraints []deppy.Constraint
 	if len(dependencyIDs) > 0 {
 		constraints = append(constraints, constraint.Dependency(dependencyIDs...))
 	}
 	return &BundleVariable{
-		SimpleVariable: input.NewSimpleVariable(id, constraints...),
+		SimpleVariable: input.NewSimpleVariable(BundleVariableID(bundle), constraints...),
 		bundle:         bundle,
 		dependencies:   dependencies,
 	}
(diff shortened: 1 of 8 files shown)


internal/resolution/variablesources/bundles_and_dependencies.go
```diff
@@ -63,25 +63,25 @@ func (b *BundlesAndDepsVariableSource) GetVariables(ctx context.Context) ([]depp
...
-			}
+		id := olmvariables.BundleVariableID(head)
 
-			// add bundle dependencies to queue for processing
-			bundleQueue = append(bundleQueue, dependencies...)
+		// ignore bundles that have already been processed
+		if _, ok := visited[id]; ok {
+			continue
+		}
+		visited[id] = struct{}{}
```
stevekuznetsov, line 72: Can we use `sets.New[deppy.Identifier]()` for clearer semantics on `visited` and `added`?

m1kola: AFAIK Go doesn't have sets in standard library and as a result using maps for sets is pretty common.

But we can do that. However I do not see any sets package in our codebase. What package is it?

m1kola: Anyway I think it should not be in this PR. Could you please create an issue or PR for this suggesting a pacakge to use?

stevekuznetsov: The k8s API machinery sets package is the one - we use it in basically every other package. Feel free to create the issue if you don't think you can do it here.

everettraven (approved):
These changes seem reasonable and look good to me, nice job on the variable reduction!

I'm approving, but would prefer waiting to merge on a second approval from someone who has better insight into the variable(sources) and resolution process as a whole to verify this doesn't have any cascading impacts.

internal/resolution/variablesources/bundles_and_dependencies.go
m1kola: I'm a fan of small PRs serve one porpuse. Don't want to mix different efforts into this PR.

Created #454 for this.

internal/resolution/variables/bundle.go
```diff
@@ -26,17 +26,17 @@ func (b *BundleVariable) Dependencies() []*catalogmetadata.Bundle {
 	return b.dependencies
 }
 
-func NewBundleVariable(id deppy.Identifier, bundle *catalogmetadata.Bundle, dependencies []*catalogmetadata.Bundle) *BundleVariable {
-	var dependencyIDs []deppy.Identifier
+func NewBundleVariable(bundle *catalogmetadata.Bundle, dependencies []*catalogmetadata.Bundle) *BundleVariable {
+	dependencyIDs := make([]deppy.Identifier, 0, len(dependencies))
 	for _, bundle := range dependencies {
```
joelanford, line 31: Nit: unrelated to this PR, but we can we change the loop variable to be called `dependency` instead of `bundle` so that it doesn't shadow the `bundle` variable passed in as a parameter?

joelanford (approved):
LGTM. I left a nit about a semi-unrelated fix. We can address it now or later.

internal/resolution/variables/bundle.go
m1kola: I'll submit a follow up PR straight after merging this one.

m1kola:
I experimented with t8c operator from https://operatorhub.io/: when installing without specifying version it currently creates 87 variables. With changes in this PR it creates 21 variables.

I haven't mesured the preformance (with a naked eye it is unnoticable), but at least it will give a shorter error message in case of resolution failure.

internal/resolution/variables/bundle.go
m1kola: https://github.com/operator-framework/operator-controller/pull/455

Issue #449, opened 2023-10-10
Respect `.spec.upgradeConstraintPolicy`


m1kola:
**Please read [RFC](https://docs.google.com/document/d/1dfiNr0kvP1nfS0f9xaGiLyfrJ_KcD38clYs4LYBKllk/edit) for better formatting and full scope.** The excerpt below is just to indicate the scope covered by this specific issue.

Need to update code to respect `.spec.upgradeConstraintPolicy` accordingly to the [RFC](https://docs.google.com/document/d/1dfiNr0kvP1nfS0f9xaGiLyfrJ_KcD38clYs4LYBKllk/edit):

> When `.spec.upgradeConstraintPolicy` is set to `Ignore` we do not apply upgrade constraints. This means an `Operator` can be transitioned to any available version of the package (even one older than the currently installed one). Note that, if set, other constraints on `Operator` still apply (e.g. `.spec.version` and `.spec.channel`).
>
> When `.spec.upgradeConstraintPolicy` is set to `Enforce` we use either semver or legacy semantics depending on the position of the `ForceSemverUpgradeConstraints` feature gate.

m1kola:
We probably need to do #437 first because with the current structure of variable sources it is tricky to get access to relevant `Operator`.

m1kola:
We need to update tests introduced in https://github.com/operator-framework/operator-controller/pull/472 to include a case where we make sure that it is possible to upgrade within major version zero when `.spec.upgradeConstraintPolicy` is set to `Ignore` as this is the only way to upgrade in these versions.

For example, upgrade from `0.0.1` to `0.0.2` is not possible without this.

m1kola:
I will either need #437 to be solved (PRs #498, #499, #500 and #501 to get merged) or #508 to get merged to be able to make progress with this.

PR #460, opened 2023-10-13
🌱 Reduce number of variable sources

### LINKED ISSUE #437: Reduce number of variable sources involved in resolution process
Today we have variable sources which create other variable sources which create variables. While this gives us very short and simple units - it makes harder to follow and creates a lot of boilerplate (e.g. need for support of nested variable soruces).

For example `OperatorVariableSource` creates `RequiredPackageVariableSource` which creates `RequiredPackageVariable`:

https://github.com/operator-framework/operator-controller/blob/6808a5020e579da6a5da0083bc066b1c1a63702d/internal/resolution/variablesources/operator.go#L42-L47

https://github.com/operator-framework/operator-controller/blob/6808a5020e579da6a5da0083bc066b1c1a63702d/internal/resolution/variablesources/required_package.go#L88-L90

`BundleDeploymentVariableSource` creates `InstalledPackageVariableSource` which creates `InstalledPackageVariable`:
https://github.com/operator-framework/operator-controller/blob/6808a5020e579da6a5da0083bc066b1c1a63702d/internal/resolution/variablesources/bundle_deployment.go#L47

https://github.com/operator-framework/operator-controller/blob/6808a5020e579da6a5da0083bc066b1c1a63702d/internal/resolution/variablesources/installed_package.go#L53-L55

m1kola:
# Description

This PR reduces number of variable sources to one.

Closes #437

## Reviewer Checklist

- [ ] API Go Documentation


cmd/resolutioncli/variable_source.go
```diff
@@ -17,28 +17,90 @@ limitations under the License.
...
+	if err != nil {
+		return nil, err
+	}
+
+	requiredPackages := []*olmvariables.RequiredPackageVariable{}
+	requiredPackage, err := variablesources.RequiredPackageVariable(allBundles, o.packageName, o.packageChannel, o.packageVersionRange)
+	if err != nil {
+		return nil, err
+	}
+	requiredPackages = append(requiredPackages, requiredPackage)
```
m1kola, line 75: They only real difference between `OfflineOLMVariableSource` and `OLMVariableSource` is this so far.

I don't like the current structure of this as it will definitevely drift apart from the main variable source.

internal/resolution/variablesources/olm.go
```diff
@@ -0,0 +1,344 @@
...
+
+		// find corresponding bundle for the installed content
+		resultSet := catalogfilter.Filter(allBundles, catalogfilter.WithBundleImage(bundleImage))
+		if len(resultSet) == 0 {
+			return nil, fmt.Errorf("bundleImage %q not found", bundleImage)
+		}
+
+		// TODO: fast follow - we should check whether we are already supporting the channel attribute in the operator spec.
+		//       if so, we should take the value from spec of the operator CR in the owner ref of the bundle deployment.
+		//       If that channel is set, we need to update the filter above to filter by channel as well.
```
m1kola, line 167: This is an old todo. I'm trying to move code around without significant refactoring.

But we have an issue related to this: https://github.com/operator-framework/operator-controller/issues/466

cmd/resolutioncli/variable_source.go
m1kola: I've been thinking about it and I think - yes, there is a chance that `OfflineOLMVariableSource` and `OLMVariableSource` will drift apart. But on the other hand - `OfflineOLMVariableSource` might not need everything `OLMVariableSource` does.

I'm inclined to leave it like this for now and change it depending on where we want to take the offline resolution.

internal/resolution/variablesources/installed_package.go
```diff
@@ -1,76 +1,74 @@
...
+
+		// find corresponding bundle for the installed content
+		resultSet := catalogfilter.Filter(allBundles, catalogfilter.WithBundleImage(bundleImage))
+		if len(resultSet) == 0 {
+			return nil, fmt.Errorf("bundleImage %q not found", bundleImage)
+		}
+
+		// TODO: fast follow - we should check whether we are already supporting the channel attribute in the operator spec.
+		//       if so, we should take the value from spec of the operator CR in the owner ref of the bundle deployment.
+		//       If that channel is set, we need to update the filter above to filter by channel as well.
```
m1kola, line 55: This is an old todo. I'm trying to move code around without significant refactoring.

But we have an issue related to this: https://github.com/operator-framework/operator-controller/issues/466

internal/resolution/variablesources/bundle.go
```diff
@@ -1,63 +1,34 @@
 package variablesources
```
m1kola, line 1: I initially had everything in this package in two files: `olm.go` and `olm_test.go`. While `olm.go` was is still manageable, navigating `olm_test.go` was a bit difficult. So I decided to split into multiple files.

It resulted in a more complicated diff (while the files themselves are straightforward).

stevekuznetsov:
Please separate your changes into commits that only move code from one place to another and bona-fide logic/implementation changes. Reviewing a +1200 LoC diff is otherwise not really possible :)

m1kola:
> Please separate your changes into commits that only move code from one place to another and bona-fide logic/implementation changes. Reviewing a +1200 LoC diff is otherwise not really possible :)

Moving wasn't my intention. I deleted a bunch of old code (6 or so variable sources) and added 5 new functions which re-use snippets of code from deleted variable soruces. I see that in the diff it shows like `variablesources/bundle.go` was renamed from `bundles_and_dependencies.go`, but I didn't move it. I think git shows it like this in the diff becuase of large re-used snippets.

@stevekuznetsov do you think it will make it easier for reviewers if I do something like this?

* Commit 1: delete old code
* Commit 2: add new stuff

It feels like it won't make big difference, but I'm happy to do that if it easier this way.

stevekuznetsov:
Generally what would be helpful is a distinction between code you _wrote_ or _edited_ as part of this work, versus code that existed prior and was relocated.

m1kola:
> Generally what would be helpful is a distinction between code you _wrote_ or _edited_ as part of this work, versus code that existed prior and was relocated.

Agreed. As a reviewer I love that and small diffs. I also love to have proper commit history for each line. As an author I try to make by diffs be like that, but in this case I'm struggling to come up with a way to achieve that.

For now I'm going to add comments to files and describe where snippets of code came from to make it a bit easier (similar to what I did for [this todo](https://github.com/operator-framework/operator-controller/pull/460#discussion_r1374350231)).

stevekuznetsov:
Good luck. I'll be clear - I really don't know that anyone can do a useful review of a diff this large. If you're looking for a rubber stamp, it's less important.

internal/resolution/variablesources/installed_package.go
```diff
@@ -1,76 +1,74 @@
...
-	return fmt.Errorf("bundleImage %q not found", r.bundleImage)
-}
-
-func NewInstalledPackageVariableSource(catalogClient BundleProvider, bundleImage string) (*InstalledPackageVariableSource, error) {
-	successors := legacySemanticsSuccessors
+// InstalledPackageVariables returns variables representing packages
+// already installed in the system.
+// Meaning that each BundleDeployment managed by operator-controller
+// has own variable.
+func InstalledPackageVariables(
```
m1kola, line 23: This re-uses code snippets from `InstalledPackageVariableSource` and `BundleDeploymentVariableSource`.

From `BundleDeploymentVariableSource` it is a few lines around `processed` set which were modified for new structure.

internal/resolution/variablesources/required_package.go
```diff
@@ -1,104 +1,53 @@
...
-	return func(r *RequiredPackageVariableSource) error {
-		if channelName != "" {
-			r.channelName = channelName
-			r.predicates = append(r.predicates, catalogfilter.InChannel(channelName))
-		}
-		return nil
+// RequiredPackageVariable returns a variable which represent
+// explicit requirement for a package from an user.
+// This is when an user explicitly asks "install this" via Operator API.
+func RequiredPackageVariable(allBundles []*catalogmetadata.Bundle, packageName, channelName, versionRange string) (*olmvariables.RequiredPackageVariable, error) {
```
m1kola, line 18: This re-uses a bits from `RequiredPackageVariableSource`. What was removed:

* `NewRequiredPackageVariableSource` & `RequiredPackageVariableSourceOption`. We just pass things as arguments into function without options abstraction
* Separate `notFoundError` function for formatting an error message

m1kola (commented):
> Good luck. I'll be clear - I really don't know that anyone can do a useful review of a diff this large. If you're looking for a rubber stamp, it's less important.

@stevekuznetsov this comment is not helpful. If you have any practical suggestions on how to split this or reduce the diff - please let me know.

I added some comments to this PR which will hopefully help navigate this change. I'll think more about how I can split/reduce this diff, but at the moment nothing comes to mind.

internal/resolution/variablesources/olm_variable_source.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+func NewOLMVariableSource(cl client.Client, catalogClient BundleProvider) *OLMVariableSource {
+	return &OLMVariableSource{
+		client:        cl,
+		catalogClient: catalogClient,
+	}
+}
+
+func (o *OLMVariableSource) GetVariables(ctx context.Context) ([]deppy.Variable, error) {
```
perdasilva, line 27: looks much nicer ^^ ty

internal/resolution/variablesources/olm_variable_source.go
```diff
@@ -0,0 +1,81 @@
...
+	if err := o.client.List(ctx, &operatorList); err != nil {
+		return nil, err
+	}
+
+	bundleDeploymentList := rukpakv1alpha1.BundleDeploymentList{}
+	if err := o.client.List(ctx, &bundleDeploymentList); err != nil {
+		return nil, err
+	}
+
+	allBundles, err := o.catalogClient.Bundles(ctx)
```
perdasilva, line 38: would it make sense to return some kind of wrapper type around the list of bundles (something like `snapshot`...?) and attach the filtering methods to that, rather than relying on utility functions to do it?

internal/resolution/variablesources/olm_variable_source.go
```diff
@@ -0,0 +1,81 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	requiredPackages := []*olmvariables.RequiredPackageVariable{}
+	for _, operator := range operatorList.Items {
+		requiredPackage, err := RequiredPackageVariable(allBundles, operator.Spec.PackageName, operator.Spec.Channel, operator.Spec.Version)
```
perdasilva, line 45: The *Variable function names feel almost like constructors here, which normally would be fine - but the allBundles being passed everywhere gives you a sense that those variables aren't really carrying around "all the bundles" and there's actually some non-trivial amount of work being done under the covers. Maybe even like `Make*Variable`.

internal/resolution/variablesources/installed_package.go
```diff
@@ -1,76 +1,74 @@
...
 		successors = semverSuccessors
 	}
 
-	return &InstalledPackageVariableSource{
-		catalogClient: catalogClient,
-		bundleImage:   bundleImage,
-		successors:    successors,
-	}, nil
+	result := make([]*olmvariables.InstalledPackageVariable, 0, len(bundleDeployments))
+	processed := sets.Set[string]{}
```
perdasilva, line 33: I'm not a big fan of using sets.Set outside of reconciler code - it creates a dependency between this package and random unrelated kube libs. Not a blocker, as this is all internal. But something to think about. Never understood why Go never bothered to provide standard impls of basic data structures but instead leaves the developer to constantly re-invent the wheel...=S

internal/resolution/variablesources/olm_variable_source.go
perdasilva: or maybe something like: `Make*Variable(...).From(allBundles)`...?

internal/resolution/variablesources/installed_package.go
stevekuznetsov: It's [250](https://github.com/kubernetes/apimachinery/blob/master/pkg/util/sets/set.go) lines of code - if you worry about the import we could just copy it - but FWIW the package doesn't pull in anything else from Kube.

cmd/resolutioncli/variable_source.go
joelanford: If this is the only difference, I would _highly_ suggest refactoring now to consolidate somehow.

I think it's all but guaranteed that other contributors will be unaware of this variable source since it is tucked away in a place that not many maintainers are even aware of. I, for one, would likely forget to come change this one if I was making a change in the variable source used by the controller.

internal/resolution/variablesources/installed_package.go
ncdc: We actively encourage the use of sets.Set anywhere/everywhere it's useful.

internal/resolution/variablesources/olm_variable_source.go
m1kola: I also would prefer to keep things simple.

internal/resolution/variablesources/installed_package.go
m1kola: > Go never bothered to provide standard impls of basic data structures but instead leaves the developer to constantly re-invent the wheel...=S

A map of empty structs is basically a set, so it kinda already exsits :) IMO map of empty structs is absolutely fine and is common in Go. But sets pacakge from k8s is ok too. I'm happy to do it eithe way.

internal/resolution/variablesources/installed_package.go
joelanford: Maybe I'm misunderstanding this TODO, but it also just seems like something we don't want to do?

The only purpose of the `Operator.spec.channel` is to restrict what bundles are possible candidates for the next resolution. Once a bundle is installed, it makes no difference which channel it "came" from. There is no concept like "this bundle deployment represents a bundle from a specific channel".

internal/resolution/variablesources/installed_package.go
joelanford: I think apimachinery is so ubiquitous in everything we do that it has no measurable impact to use it in a module that already has a dependency on it anyway

It also serves as a good opportunity to teach others about the fact that k8s has tons of helpful libraries scattered around in various places, with lots of usage and tests and community support/maintenance. We should be using _more_ not less, IMO.

cmd/resolutioncli/variable_source.go
m1kola: @joelanford yes, that was my concern as well. But the other side of the coin is that PoC code becomes harder to maintain. Having done a few refactoring of different complexity in this area (removal of entity sources most recently) - this PoC always stands in a way.

I was hoping that functions will become building blocks for varaible sources (each tested separately) and variable sources will be just combining them together as need. E.g. offline variable soruce will only contain funcs we need for offline resolution and controller variable source will contain more.

But yes, we can consolidate them.

internal/resolution/variablesources/installed_package.go
```diff
@@ -1,76 +1,74 @@
...
+	for _, bundleDeployment := range bundleDeployments {
```
joelanford, line 34: If I create a bundleDeployment directly and it happens to reference an image that's in a catalog, now OLM is going to start including it in resolution?

That seems incorrect. We should only be looking at installed bundles and successors for the Operator objects that exist.

internal/resolution/variablesources/installed_package.go
m1kola: > Maybe I'm misunderstanding this TODO, but it also just seems like something we don't want to do?

I just re-read it and now I think I misunderstood it: looks like it is not related to #466. But I still think that #466 is an issue that needs to be addressed (I replied ro you rcomment [here](https://github.com/operator-framework/operator-controller/issues/466#issuecomment-1785166959)).

I interpreted it like "we should do a beter job at finding the correct bundle for installed package and we could use channel for that... Maybe?".



> The only purpose of the `Operator.spec.channel` is to restrict what bundles are possible candidates for the next resolution. Once a bundle is installed, it makes no difference which channel it "came" from. There is no concept like "this bundle deployment represents a bundle from a specific channel".

Yes, that was my understanding too.

I think we should delete this comment to avoid further confusion, but keep #466 open.

@joelanford What do you think?

ncdc:
I would prefer we remove the VariableSource concept entirely and instead create a DeppySolver on the fly when you need it (i.e. not thread safe), and call something like `AddVariables(...)` 1 or more times when you want to give it inputs. I believe this will be easier to maintain, easier to test, easier to reason about, and it makes the set of inputs static.

I realize you have put a lot of time into this PR @m1kola, but I believe not merging this as-is and instead going in the direction proposed above will have significant long-term benefits.

internal/resolution/variablesources/olm_variable_source.go
```diff
@@ -0,0 +1,81 @@
...
+
+	installedPackages, err := InstalledPackageVariables(allBundles, bundleDeploymentList.Items)
+	if err != nil {
+		return nil, err
+	}
+
+	bundles, err := BundleVariables(allBundles, requiredPackages, installedPackages)
+	if err != nil {
+		return nil, err
+	}
```
joelanford, line 60: It feels off to me that we're using variables to compute more variables. For me at least, this structure makes it really hard to build intuition about what each variable is actually adding to the problem.

Without diving too deeply, it appears to me like we:
- Figure out what bundles are required based on the `Operator` objects that exist (`requiredPackages`)
- Figure out what bundles are required based on existing BDs and their successors (but this is for all BundleDeployments, not just BDs managed by Operators.  (`installedPackages`)
- For `requiredPackage` bundles and `installedPackage` bundles combined, build a set of `Bundle` variables that:
   1. include all `requiredPackage` and `installedPakcage` bundles (so `bundles` is a superset?)
   2. include all possible dependencies, recursively, based on the roots from (i)


Am I getting this correct? If so, what purpose do the `requiredPackages` and `installedPackages` variables actually serve for the resolver? If none (because `bundles` is a superset?), can we refactor such that `requiredPackages` and `installedPackages` are not actually variables and are instead just lists that we use to filter `allBundles` to `bundles`?

m1kola:
> I would prefer we remove the VariableSource concept entirely

@ncdc I had similar thoughts recently. We removed the concept of entity sources from deppy and I think we can remove the abstraction of varaible sources from it too. Deppy can accept a slice of inputs and it should be good enough.

I'll create an issue in Deppy for this so we can discuss in it.

> I realize you have put a lot of time into this PR @m1kola, but I believe not merging this as-is and instead going in the direction proposed above will have significant long-term benefits.

I think we should proceed with this PR. Splitting variable sources into separate functions will make the transition to deppy version without entity sources easier: we will be able to use the same functions which are not avare of variable source concept (and are tested separately). Also this change _I think_ will help me to complete the remaining bit of implementation for initial upgrade support (#449) more easier: I won't need to shatter the implementation between multiple variable soruces (I will be focusing on #449 before anything else most likely).

internal/resolution/variablesources/installed_package.go
joelanford: > A map of empty structs is basically a set, so it kinda already exsits

It is, but using maps directly requires more boilerplate (e.g. remembering to initialize a `nil` map) and the syntax can make it harder to tell that the purpose of the map is to act as a set.

internal/resolution/variablesources/olm_variable_source.go
m1kola: I added `Make` prefix. I'm not in favour of method chaining (at least not for at this point).

internal/resolution/variablesources/olm_variable_source.go
joelanford: > However on failed resolution error messages will be harder to understand: there will be no difference between a dependency of a dependency of a dependecny and a thing which user explicitly requested.

I'm not convinced that's true (or at least necessarily has to be the case). I'd imagine something like this is what is (or could be) fed into the resolver:
- OperatorVariable(Package: a, DependsOnOneOf(aCurrent, aSuccessor), Mandatory: true)
- BundleVariable(Bundle: aCurrent, DependsOnOneOf(bCandidate), Mandatory: false)
- BundleVariable(Bundle: aSuccessor, DependsOnOneOf(bCandidate), Mandatory: false)
- BundleVariable(Bundle: bCandidate, DependsOnOneOf(cCandidate), Mandatory: false)
- BundleVariable(Bundle: cCandidate, Mandatory: false)

Any resolution failure that involves `cCandidate` being in conflict will only be unsat in this case because of the mandatory Operator variable for package `a`. And the unsat error message would traverse this dependency chain, right?

m1kola: > Both `RequiredPackageVariable` and `InstalledPackageVariable` represent bundles so is `BundleVariable`

Just realised - this is incorrect. `RequiredPackageVariable` and `InstalledPackageVariable` represent packages which depend on specific bundles.

> I'd imagine something like this is what is (or could be) fed into the resolver

Do I understand correctly that it will be an equivalent of combining `RequiredPackageVariable` and `InstalledPackageVariable` together into `OperatorVariable`? At first glance it makes sense and might work, but we need to look into it in more depth.

I suggest that we do it as a separate effort: in this PR leave the structure of the resolution problem as it currently is in `main` (this PR is already large) and separately look into combining variables.

I'll split #437 into two issues: geting rid of variable sources and combining variables.

joelanford:
> I added some comments to this PR which will hopefully help navigate this change. I'll think more about how I can split/reduce this diff, but at the moment nothing comes to mind.

I do agree with @stevekuznetsov that this PR could have been made easier to review. I happen to be somewhat familiar with the state on `main`, so I was able to focus on the new code with my background knowledge of the old code.

One suggestion would be to have a commit that literally just takes the original code and moves it into the destination files; so no changes other than moving code. Then another commit that actually changes structures and function signatures, then finally any other commits with net-new changes.

m1kola:
> One suggestion would be to have a commit that literally just takes the original code and moves it into the destination files; so no changes other than moving code. Then another commit that actually changes structures and function signatures, then finally any other commits with net-new changes.

I tried to move `variablesources/crd_constraints.go` to `variablesources/bundle_uniqueness.go` as the latter uses a big chunk from from the former. I can probably do the same for `variablesources/bundles_and_dependencies.go` & `variablesources/bundle.go` (it appears in the diff like a move anyway). But it did not seem very helpful to me (probably because some variable sources were combined together).

At the moment I'm trying to split it into a train of small incremental PRs, but it is taking a lot time. Bear with me.

m1kola:
I think I managed to split into a train of smaller PRs. First two are ready for review:

- https://github.com/operator-framework/operator-controller/pull/495
- https://github.com/operator-framework/operator-controller/pull/497
- https://github.com/operator-framework/operator-controller/pull/498
- https://github.com/operator-framework/operator-controller/pull/499
- https://github.com/operator-framework/operator-controller/pull/500
- https://github.com/operator-framework/operator-controller/pull/501

I'll keep this PR open in draft for now just verify at the end that I did not miss something.

m1kola:
Closing this now as everyhing from this PR got into one of the above mentioned PRs in one form or the other.

Issue #459, opened 2023-10-13
Resolution error message is not deterministic


m1kola:
I noticed that we iterate over maps here:

https://github.com/operator-framework/operator-controller/blob/e001a0f6ffd62fc478148463ad9b39a7b2e404a3/internal/resolution/variablesources/crd_constraints.go#L66-L73

I believe this results in randomised order of variables.

Tests do not flag this because they use `ConsistOf` which ignores order:

https://github.com/operator-framework/operator-controller/blob/e001a0f6ffd62fc478148463ad9b39a7b2e404a3/internal/resolution/variablesources/crd_constraints_test.go#L285-L294

It likely contributes to what was reported in https://github.com/operator-framework/deppy/issues/142:

> It turns out the assertion is flaky because the content of the message changes from run to run, only in terms of the order in which the individual constraint messages appear. This is problematic when this error message is written to a kubernetes status object because it results in a reconcile -> update status -> reconcile -> update status loop until we get "lucky" and deppy returns the same message two reconciles in a row.

m1kola:
#461 gets rid of randomness in package uniqueness ordering.

I would still like to keep this issue open so we can get rid of this function:

https://github.com/operator-framework/operator-controller/blob/e001a0f6ffd62fc478148463ad9b39a7b2e404a3/internal/controllers/operator_controller.go#L467-L485

This will require:
* Change to error formating on deppy side (replace comma with semicolon)
* Additional tests for `OperatorReconciler` to make sure that ordering is stable

PR #461, opened 2023-10-16
Remove randomness in package uniqueness ordering

m1kola:
# Description

Partly addresses #459. I would still want to keep the issue open so we can get rid of `prettyUnsatMessage` function.

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/resolution/variablesources/crd_constraints.go
+++ b/internal/resolution/variablesources/crd_constraints.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/operator-framework/deppy/pkg/deppy"
 	"github.com/operator-framework/deppy/pkg/deppy/input"
+	"k8s.io/apimachinery/pkg/util/sets"
 
 	"github.com/operator-framework/operator-controller/internal/catalogmetadata"
 	olmvariables "github.com/operator-framework/operator-controller/internal/resolution/variables"
@@ -43,7 +44,9 @@ func (g *CRDUniquenessConstraintsVariableSource) GetVariables(ctx context.Contex
 	// todo(perdasilva): better handle cases where a provided gvk is not found
 	//                   not all packages will necessarily export a CRD
 
-	pkgToBundleMap := map[string]map[deppy.Identifier]struct{}{}
+	bundleIDs := sets.Set[deppy.Identifier]{}
+	packageOrder := []string{}
+	bundleOrder := map[string][]deppy.Identifier{}
 	for _, variable := range variables {
 		switch v := variable.(type) {
 		case *olmvariables.BundleVariable:
@@ -54,22 +57,22 @@ func (g *CRDUniquenessConstraintsVariableSource) GetVariables(ctx context.Contex
 				// get bundleID package and update map
 				packageName := bundle.Package
 
-				if _, ok := pkgToBundleMap[packageName]; !ok {
-					pkgToBundleMap[packageName] = map[deppy.Identifier]struct{}{}
+				if _, ok := bundleOrder[packageName]; !ok {
+					packageOrder = append(packageOrder, packageName)
 				}
-				pkgToBundleMap[packageName][id] = struct{}{}
+
+				if !bundleIDs.Has(id) {
+					bundleIDs.Insert(id)
+					bundleOrder[packageName] = append(bundleOrder[packageName], id)
+				}
 			}
 		}
 	}
 
 	// create global constraint variables
-	for packageName, bundleIDMap := range pkgToBundleMap {
-		var bundleIDs []deppy.Identifier
-		for bundleID := range bundleIDMap {
-			bundleIDs = append(bundleIDs, bundleID)
-		}
+	for _, packageName := range packageOrder {
 		varID := deppy.IdentifierFromString(fmt.Sprintf("%s package uniqueness", packageName))
-		variables = append(variables, olmvariables.NewBundleUniquenessVariable(varID, bundleIDs...))
+		variables = append(variables, olmvariables.NewBundleUniquenessVariable(varID, bundleOrder[packageName]...))
 	}
 
 	return variables, nil
--- a/internal/resolution/variablesources/crd_constraints_test.go
+++ b/internal/resolution/variablesources/crd_constraints_test.go
@@ -284,13 +284,13 @@ var _ = Describe("CRDUniquenessConstraintsVariableSource", func() {
 		}
 		// Note: As above, the 5 GVKs would appear here as GVK uniqueness constraints
 		// if GVK Uniqueness were being accounted for.
-		Expect(crdConstraintVariables).To(WithTransform(CollectGlobalConstraintVariableIDs, ConsistOf([]string{
-			"another-package package uniqueness",
-			"bar-package package uniqueness",
-			"test-package-2 package uniqueness",
+		Expect(crdConstraintVariables).To(WithTransform(CollectGlobalConstraintVariableIDs, Equal([]string{
 			"test-package package uniqueness",
 			"some-package package uniqueness",
 			"some-other-package package uniqueness",
+			"another-package package uniqueness",
+			"bar-package package uniqueness",
+			"test-package-2 package uniqueness",
 		})))
 	})
 


internal/resolution/variablesources/crd_constraints.go
```diff
@@ -43,7 +44,9 @@ func (g *CRDUniquenessConstraintsVariableSource) GetVariables(ctx context.Contex
 	// todo(perdasilva): better handle cases where a provided gvk is not found
 	//                   not all packages will necessarily export a CRD
 
-	pkgToBundleMap := map[string]map[deppy.Identifier]struct{}{}
```
stevekuznetsov, line 46: It might be more straightforward to use `map[string]sets.Set[deppy.Identifier]{}` and simply append a sorted list from the set. That would give a defined order to the output, whereas now it seems to mirror the order of the variables in the input - is the order of the variables somehow useful or do we just need something deterministic?

stevekuznetsov: > space will still be O(n) in both cases.

You can remove `bundleIDs` and `packageOrder` variables. Regardless of the memory efficiency, it's just easier to read since there are fewer things in play.

m1kola: I think going into logarithmic is not worth the readability improvement.

Also It seems to me that we will still need:
* A slice for package we need to perform ordering on it
* A slice or a map of slices for bundle ids so we can perform ordering on it
* A loop or two to convert a map into slices before sorting

Or am I missing something?

m1kola: > How large is `n`? I imagine the change in runtime has no practical implication whatsoever.

We don't know. It depends on a cluster setup (how many catalogs are out there, how many packages, etc).

> believe that would be sufficient, no?

I think it won't be sufficient since we also need to ensure that we do not have duplicate items in `packageOrder` which can happen (multiple packages depend on the package name).

> previously if a bundle ID existed for more than one package, it would be added to the list of both. 

Single bundle belongs to only one package (package is a property on a bundle in FBC). Bundle ID is unique. So I think it is not a issue.

> Now there's a global uniqueness constraint. Is that intentional?

Yes, this is exactly what we want to do here.

> Similarly, by tracking `bundleOrder` separately (as opposed to extracting the keys of `pkgToBundles`) there's space for errors if the two are out of sync. If you don't think the duplication of the source of truth there and some of the new edges are worth worrying about, that's totally acceptable.

There is always space for errors and regressions are inevitable. I don't think it is worth worrying about it to this extent.

stevekuznetsov: I won't belabor the point. Your PR was ostensibly to remove randomness in ordering, with a couple entirely unrelated side-effects. You claim those are all important and necessary. Great. If so, let's merge the PR. Please understand that the original description of this work did not imply that the other things you did here were useful or necessary :)

stevekuznetsov: > There is always space for errors and regressions are inevitable. 

I will mention, since this is the second time you've made such a comment recently .... yes, egregious human errors are always possible. Creating systems in our code that make them less likely is worthwhile. Duplicating sources of truth is a classic case to worry about. The approach that, since errors are inevitable, it's not worth bothering to make them rare, is weird.

m1kola: > Your PR was ostensibly to remove randomness in ordering, with a couple entirely unrelated side-effects.

> Please understand that the original description of this work did not imply that the other things you did here were useful or necessary :)

What other things or side effects? Is it about "Now there's a global uniqueness constraint" comment? It is not new - the code before my changes and after my changes ensures that there is global uniqueness constraint. The output is exactly the same as it was apart from stable order vs random order and it can be seen in tests.

> Creating systems in our code that make them less likely is worthwhile.

I absolutely agree with this point.

> will mention, since this is the second time you've made such a comment recently .... yes, egregious human errors are always possible. Creating systems in our code that make them less likely is worthwhile. Duplicating sources of truth is a classic case to worry about. The approach that, since errors are inevitable, it's not worth bothering to make them rare, is weird.

I however will also mention that on both occasions you failed to convince me that the approaches you suggest substantially better. In this occasion it is more clear than on the other. What you suggested:

* Performs worse
* Requires same amount of space
* Requires same amount of variables, if we want it to work the way it worked before changes, but fix randomness in ordering.

To me it seems that at the end it all boils down to personal preferences or beliefs.

> The approach that, since errors are inevitable, it's not worth bothering to make them rare, is weird.

Factoring code in a specific way is not the only thing which helps us make errors less likely to reach live systems / be merged into `main`. I think tests are move effective tool for this than some specific way of writing code (at least in this situation).

There is a balance to be found between writing nearly perfect code (whatever that means for each individual) and making progress. I believe in our situation with a greenfield project and high tolerance to bugs, we should be somewhere here on the scale:

```
perfect code |-------------| iteration speed
                     ^
```

stevekuznetsov: > the code before my changes and after my changes ensures that there is global uniqueness constraint. 

Can you point out what part of the previous code enforces the same thing that the `bundleIDs` variable does in your PR? It seems to me that the previous version of the code ensures with `pkgToBundleMap` that `id` is unique per `packageName`, but not globally that `id` is unique?

m1kola: > Can you point out what part of the previous code enforces the same thing that the `bundleIDs` variable does in your PR?

Sure, here:

https://github.com/operator-framework/operator-controller/blob/7161a7b62ec17dc36790455b03773a5f14e7671f/internal/resolution/variablesources/crd_constraints.go#L60

And as I explained [above](https://github.com/operator-framework/operator-controller/pull/461#discussion_r1360830868):

> Single bundle belongs to only one package (package is a property on a bundle in FBC). Bundle ID is unique.

So given the fact that it is 1:1 relationship between packages and bundle IDs are unique within the cluster - it is just a different way to skin a cat.

stevekuznetsov: Not sure what is being lost here. The permissible states are different now and the code is not equivalent. Your claim is that domain knowledge means that the global uniqueness is something you already have so it's fine. All I was trying to communicate is that without that domain knowledge it's really not obvious that the change is equivalent, warranted or necessary, since the stated intent of the PR was to enforce a deterministic ordering.

You wrote:

>  It is not new - the code before my changes and after my changes ensures that there is global uniqueness constraint. 

That is not true. The code before your change does no such thing.

stevekuznetsov: I saw a PR that claimed to make a deterministic ordering, but also did some other things. It was not clear that the other things were necessary or required. I suggested that a simpler approach might exist. You claimed that the other things the PR did were necessary. That's perfectly OK & reasonable.

> I don't understand. Didn't you just admint that the change is equivalent when you wrote this:

No. Given the extra information from the domain, it might be true that the output of the function is the same for the inputs you expect to see, but the code is literally not equivalent. The assumption you make (ID uniqueness) may well be true but is an implicit assumption and not encoded into the system. There are no tests for this type of input (either in production code, raising an error if an ID is not globally unique, or in test code for what the function does in that case). Changes to code that hinge on unwritten assumptions are the sort of things that can introduce bugs, which is why I look for them when reviewing.

m1kola: > No. Given the extra information from the domain, it might be true that the output of the function is the same for the inputs you expect to see, but the code is literally not equivalent. The assumption you make (ID uniqueness) may well be true but is an implicit assumption and not encoded into the system. There are no tests for this type of input (either in production code, raising an error if an ID is not globally unique, or in test code for what the function does in that case). Changes to code that hinge on unwritten assumptions are the sort of things that can introduce bugs, which is why I look for them when reviewing.

If FBC API ever changes to allow bundle to be part of multiple packages - we will have to reconsider multiple things including this piece of code (no mater before or after this change). So from this perspective - it makes no difference.

Non-unique variable IDs can cause huge problems for resolution (no mater before or after this change). This is something we should worry about, but IMO it should not happen on this level (maybe Deppy should reject multiple variables with the same IDs, if it doesn't already).

> I saw a PR that claimed to make a deterministic ordering, but also did some other things. It was not clear that the other things were necessary or required. I suggested that a simpler approach might exist.

I understand.

>  You claimed that the other things the PR did were necessary. That's perfectly OK & reasonable.

Are you ready to approve this PR at this point?

[^1]: I preveusly said "1:1 relationship between packages and bundles" which is technically incorrect, but it [was under](https://github.com/operator-framework/operator-controller/pull/461#discussion_r1361758657) the quote saying "Single bundle belongs to only one package". Hopefully that was not too confusing.

everettraven: >I think it won't be sufficient since we also need to ensure that we do not have duplicate items in packageOrder which can happen (multiple packages depend on the package name).

Just to verify, the current logic already ensures that packages and bundles are not duplicated in the resolver even if multiple catalogs provide the same information? For example, if I created a `Catalog` named `foo` that had an image ref to `someimage:foo` _and_ a `Catalog` named `bar` that had the exact same image ref of `someimage:foo` - what would happen in resolution if I asked to install a package that exists in the `someimage:foo` catalog image?

I'm definitely lacking domain knowledge regarding resolution, but I do think it is important to point out that catalogd doesn't perform any validation as to whether or not FBC content is repeated across multiple catalogs. As long as this is accounted for in resolution I think that should be fine though.

m1kola: > Just to verify, the current logic already ensures that packages and bundles are not duplicated in the resolver even if multiple catalogs provide the same information?

> I do think it is important to point out that catalogd doesn't perform any validation as to whether or not FBC content is repeated across multiple catalogs. As long as this is accounted for in resolution I think that should be fine though.

OLMv1 (at least at the moment) allows the same package to come from multiple catalogs. If you have a package `foo` in `catalog1` and in `catalog2` it will treat them as the same package and _I belive_ will allow upgrade of of `foo` from version listed in `catalog1` to version listed in `catalog2`.

However in this specific code we ensure that resolution only allows 1 bundle which belongs to pacakge `foo`. So even if you install `catalog1` and `catalog2` which is direct copy - we should be good.


> For example, if I created a `Catalog` named `foo` that had an image ref to `someimage:foo` _and_ a `Catalog` named `bar` that had the exact same image ref of `someimage:foo` - what would happen in resolution if I asked to install a package that exists in the `someimage:foo` catalog image?

Did you mean that `someimage:foo` is the ref of the catalog image? If yes - we should be good as we only allow one bundle per pacakge in resolution.

However I fear the case when there are pacakges `package1` and `package2` which have bundles which share the same **bundle image**. I think [this](https://github.com/operator-framework/operator-controller/blob/7161a7b62ec17dc36790455b03773a5f14e7671f/internal/resolution/variablesources/bundle_deployment.go#L43-L53) and [this](https://github.com/operator-framework/operator-controller/blob/7161a7b62ec17dc36790455b03773a5f14e7671f/internal/resolution/variablesources/installed_package.go#L31-L43) piece of code (they work together) might allow upgrading from `package1` to `package2` and I think we don't want that.

E.g. when catalogs (single catalog or two different catalogs) have something like this:

`package1` bundles:
* `image:tag1` - version `1.0.0`

`package2` bundles:
* `image:tag1` - version `1.0.0`
* `image:tag2` - version `1.0.1`

In this case, _I think_, when you install `package1` at version `1.0.0` and then upgrade - our [installed bundle](https://github.com/operator-framework/operator-controller/blob/7161a7b62ec17dc36790455b03773a5f14e7671f/internal/resolution/variablesources/installed_package.go#L43) might end up being the one from `package2` since we search bundes by image references. And installed bundle is driving successors.

I looked into how to mitigate this in https://github.com/operator-framework/operator-controller/pull/442, but it didn't feel like the right solutiion.

I want to get back to this.

P.S: Just noticed that we do not have an isue to track this - I'll create one shortly.

everettraven: >the same bundle image.

Is the bundle uniqueness based on the bundle image or the "name" of the bundle? would there be an issue if two separate packages specified a bundle with the same "name" but referenced different images?

m1kola: > Is the bundle uniqueness based on the bundle image or the "name" of the bundle?

Bundle image issue I mentioned above is the different part of the process (how we determine currently installed bundles for calculating possible successors). Sorry if I added confusion by mentioning unrelated thing: someone you wrote in the previous comment reminded me of this.

Here, where we create global uniqueness constraints for bundles, we are saying that "in our solution, bundle `foo` must have at most 1 bundle from the list of catalog-foo-bundlename.v1.0.0, catalog-foo-bundlename.v1.0.1, etc`. where things in the list are bundle variable IDs.

This brings us to this:

> would there be an issue if two separate packages specified a bundle with the same "name" but referenced different images?

It will not be an issue. For resolution we use variable IDs and they must be unique across whole cluster. When it comes to bundles - currently we construct variable ID based on catalog name, package name and bundle name:

https://github.com/operator-framework/operator-controller/blob/c9d387efad4b121e0a7435e4232e6d1ad0179b45/internal/resolution/variables/bundle.go#L63-L67

In OLMv0 (and until recently in OLMv1) we also had a channel as part of ID, but it is not necessary since we filter out unrelated bundles before we created a variable ID for them bundle and before we give them to the resolver.

PR #463, opened 2023-10-16
Update not found error message wording

### LINKED ISSUE #456: Address a code TODO about error message in `RequiredPackageVariableSource`
It looks like this can be addressed now since I believe version ranges already landed in `main`:

https://github.com/operator-framework/operator-controller/blob/c24fdbcdbec8ee5832633ffe45d0faadb2d2edb2/internal/resolution/variablesources/required_package.go#L93-L108

m1kola:
# Description

Unless I misunderstood it - [the todo](https://github.com/operator-framework/operator-controller/blob/c24fdbcdbec8ee5832633ffe45d0faadb2d2edb2/internal/resolution/variablesources/required_package.go#L93-L108) in code was about wording the not found error messages a bit differently so they work for version ranges as well as specific versions.

I think "matching version" sounds a good for both version ranges and concrete versions, but I'm open to any suggestions.

Fixes #456

## Reviewer Checklist

- [ ] API Go Documentation


internal/resolution/variablesources/required_package.go
```diff
@@ -91,15 +91,11 @@ func (r *RequiredPackageVariableSource) GetVariables(ctx context.Context) ([]dep
 }
 
 func (r *RequiredPackageVariableSource) notFoundError() error {
-	// TODO: update this error message when/if we decide to support version ranges as opposed to fixing the version
-	//  context: we originally wanted to support version ranges and take the highest version that satisfies the range
-	//  during the upstream call on the 2023-04-11 we decided to pin the version instead. But, we'll keep version range
-	//  support under the covers in case we decide to pivot back.
 	if r.versionRange != "" && r.channelName != "" {
-		return fmt.Errorf("package '%s' at version '%s' in channel '%s' not found", r.packageName, r.versionRange, r.channelName)
+		return fmt.Errorf("package '%s' matching version '%s' in channel '%s' not found", r.packageName, r.versionRange, r.channelName)
```
stevekuznetsov, line 95: nit: A more idiomatic English clause might be 

> no package matching ... found

m1kola: @stevekuznetsov Sounds better. Should we update other versions to match the same style?

* When there is only package name: `no package '%s' found`
* When there is package name and version range: `no package '%s' matching version '%s' found`
* When there is package name and channel: `no package '%s' in channel '%s' found`
* When there is package name, channel and version range: `no package '%s' matching version '%s' in channel '%s' found`

Or perhaps something like this for the last two involving channel:
* `no package '%s' found in channel '%s'`
* `no package '%s' matching version '%s' found in channel '%s'`

stevekuznetsov: I think those sound better, yes - your suggestion with the channel clause at the end is also good!

m1kola: Updated, please take another look

dtfranz (approved):
lgtm!

PR #469, opened 2023-10-18
`Operator` update tests with semver and legacy upgrade constraints

### LINKED ISSUE #468: Upgrade tests on controller reconcile level
This is a follow up for https://github.com/operator-framework/operator-controller/pull/444#discussion_r1362101972

We have E2E tests for upgrades for default behaviour, but we also want to test different states of feature gates on operator-controller.

Tests should include:

* Upgrade tests with semver upgrade constraints
* Upgrade tests with legacy OLMv0 upgrade constraints

m1kola:
Adds `OperatorReconciler` reconcile tests for `Operator` upgrades. Tests include:

* Upgrade tests with semver upgrade constraints
* Upgrade tests with legacy OLMv0 upgrade constraints

Closes #468

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/controllers/operator_controller_test.go
+++ b/internal/controllers/operator_controller_test.go
@@ -165,10 +165,10 @@ var _ = Describe("Operator Controller Test", func() {
 					Expect(bd.Spec.Template.Spec.ProvisionerClassName).To(Equal("core-rukpak-io-registry"))
 					Expect(bd.Spec.Template.Spec.Source.Type).To(Equal(rukpakv1alpha1.SourceTypeImage))
 					Expect(bd.Spec.Template.Spec.Source.Image).NotTo(BeNil())
-					Expect(bd.Spec.Template.Spec.Source.Image.Ref).To(Equal("quay.io/operatorhubio/prometheus@sha256:5b04c49d8d3eff6a338b56ec90bdf491d501fe301c9cdfb740e5bff6769a21ed"))
+					Expect(bd.Spec.Template.Spec.Source.Image.Ref).To(Equal("quay.io/operatorhubio/prometheus@fake2.0.0"))
 				})
 				It("sets the resolvedBundleResource status field", func() {
-					Expect(operator.Status.ResolvedBundleResource).To(Equal("quay.io/operatorhubio/prometheus@sha256:5b04c49d8d3eff6a338b56ec90bdf491d501fe301c9cdfb740e5bff6769a21ed"))
+					Expect(operator.Status.ResolvedBundleResource).To(Equal("quay.io/operatorhubio/prometheus@fake2.0.0"))
 				})
 				It("sets the InstalledBundleResource status field", func() {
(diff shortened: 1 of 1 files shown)


internal/controllers/operator_controller_test.go
```diff
@@ -1048,6 +1056,223 @@ func verifyConditionsInvariants(op *operatorsv1alpha1.Operator) {
...
+	require.NoError(t, err)
+	assert.NotNil(t, cfg)
+
+	sch = runtime.NewScheme()
+	require.NoError(t, operatorsv1alpha1.AddToScheme(sch))
+	require.NoError(t, rukpakv1alpha1.AddToScheme(sch))
+
+	cl, err = client.New(cfg, client.Options{Scheme: sch})
+	require.NoError(t, err)
+	assert.NotNil(t, cl)
```
m1kola, line 1079: Given that we want to get rid of ginkgo tests (see #189) I decided to add a new test for updates instead of expanding existing ginkgo based test (above in this file).

---

Now the issue is - I had to include all this boilerplate into the test and do clean up in each subtest (see `DeleteAllOf` in `defer`s below). For ginkgo we have [`BeforeSuite` here](https://github.com/operator-framework/operator-controller/blob/7d53835f38b8092476fb92b87da4be14567cb527/internal/controllers/suite_test.go#L56-L78) and `AfterSuite`.

Wondering if we can do it differently. Any suggestions?

I considered a few things:
* Adding this stuff into `init`, but I think cleanup will be tricky. We will have to clean up in each test/sub test?
* I was hoping to use controller-runtime's fake client instead, but it doesn't support `Patch` which is being used in `OperatorReconciler`.

m1kola: #473 was merged to address boilerplate related to `envtest`

internal/controllers/operator_controller_test.go
```diff
@@ -165,10 +170,10 @@ var _ = Describe("Operator Controller Test", func() {
 					Expect(bd.Spec.Template.Spec.ProvisionerClassName).To(Equal("core-rukpak-io-registry"))
 					Expect(bd.Spec.Template.Spec.Source.Type).To(Equal(rukpakv1alpha1.SourceTypeImage))
 					Expect(bd.Spec.Template.Spec.Source.Image).NotTo(BeNil())
-					Expect(bd.Spec.Template.Spec.Source.Image.Ref).To(Equal("quay.io/operatorhubio/prometheus@sha256:5b04c49d8d3eff6a338b56ec90bdf491d501fe301c9cdfb740e5bff6769a21ed"))
+					Expect(bd.Spec.Template.Spec.Source.Image.Ref).To(Equal("quay.io/operatorhubio/prometheus@fake2.0.0"))
```
m1kola, line 173: I changed test data (included more packages with higher version). There is a bunch of tests where we test installing packages without a specific version and the behaviour is to select the latest version. So I had to update the assertions to reflect the latest version.

dtfranz (approved):
lgtm!

PR #473, opened 2023-10-20
Refactor controllers test setup & teardown for reusability between Ginkgo and standard Go tests

m1kola:
# Description

This will enable both Ginkgo and standard Go tests to use the same setup & teardown.

This will help to get rid of some of the setup boilerplate in #469.

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/controllers/suite_test.go
+++ b/internal/controllers/suite_test.go
@@ -17,89 +17,77 @@ limitations under the License.
 package controllers_test
 
 import (
-	"context"
+	"fmt"
+	"os"
 	"path/filepath"
 	"testing"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
-	rukpakv1alpha1 "github.com/operator-framework/rukpak/api/v1alpha1"
-	"k8s.io/apimachinery/pkg/api/meta"
+
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/client-go/rest"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/envtest"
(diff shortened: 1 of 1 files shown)


internal/controllers/suite_test.go
```diff
@@ -18,88 +18,91 @@ package controllers_test
...
-// These tests use Ginkgo (BDD-style Go testing framework). Refer to
-// http://onsi.github.io/ginkgo/ to learn more about Ginkgo.
-
 var (
-	cfg     *rest.Config
-	cl      client.Client
-	sch     *runtime.Scheme
-	testEnv *envtest.Environment
+	cl  client.Client
+	sch *runtime.Scheme
```
m1kola, line 42: I removed `cfg` and `testEnv` from here. They were defined on the package level so that both `BeforeSuite` and `AfterSuite` could access them. Now that we do setup and teardown in single `TestMain` function - they can be function scoped.

internal/controllers/suite_test.go
```diff
@@ -17,89 +17,77 @@ limitations under the License.
...
-
-var _ = AfterSuite(func() {
-	var operators operatorsv1alpha1.OperatorList
-	var bundleDeployments rukpakv1alpha1.BundleDeploymentList
-
-	Expect(cl.List(context.Background(), &operators)).To(Succeed())
-	Expect(cl.List(context.Background(), &bundleDeployments)).To(Succeed())
-
-	Expect(namesFromList(&operators)).To(BeEmpty(), "operators left in the cluster")
-	Expect(namesFromList(&bundleDeployments)).To(BeEmpty(), "bundle deployments left in the cluster")
```
m1kola, line 89: I droped this bit. I do not see much value in failing after tests.

everettraven (approved):

internal/controllers/suite_test.go
everettraven: I think it was to ensure everything is cleaned up before stopping the running envtest process, but if envtest exits just fine without this cleanup I think its OK to remove

PR #475, opened 2023-10-20
(vendor) k8s + catalogd bump

anik120:
# Description

## Reviewer Checklist

- [ ] API Go Documentation


everettraven (approved):
Aside from the lint warning this looks good to me

m1kola:
Heads up, I think the `operator-developer-e2e` failure is legit, but not related to this PR. I think it is related to #444, but I'm not 100% sure (tests were green on #444 as well as some other PRs which came after it).

I'm investigating.

joelanford (approved):
Needs a go.mod conflict resolution, but looks good to me otherwise.

m1kola:
> Edit: It looks like some of the failures across our PRs were due to [golang/go#63684](https://github.com/golang/go/issues/63684) and I no longer see issues in other PRs. So it might be related to this PR after all?

Correction: `operator-developer-e2e` failures are not related to this PR after all (see https://github.com/operator-framework/operator-controller/issues/481 and PR for this - https://github.com/operator-framework/operator-controller/pull/488)

go.mod
```diff
@@ -8,49 +8,49 @@ require (
 	github.com/go-logr/logr v1.2.4
 	github.com/onsi/ginkgo/v2 v2.13.0
 	github.com/onsi/gomega v1.28.0
-	github.com/operator-framework/api v0.17.4-0.20230223191600-0131a6301e42
-	github.com/operator-framework/catalogd v0.7.0
+	github.com/operator-framework/api v0.17.7
```
m1kola, line 11: I see that you updated to 0.17.7. Does it make sense to update to 0.18.0?

anik120: This came in as a transitive dependency from operator-registry, but I'm happy to bump it to v0.18.0. Done [b10ddf6bf5](https://github.com/operator-framework/operator-controller/compare/1f2c1528b5a2696a867c502fc95222a181e34476..af4dfb7e5818909c84dbd8719d0f4bb10ddf6bf5)

grokspawn (approved):

m1kola (commented):
I went through about half of it. So far noticed only a minor thing.

test/operator-framework-e2e/setup.sh
```diff
@@ -204,8 +210,51 @@ cat <<EOF > "${TMP_ROOT}"/catalog/index.yaml
...
+kind: Job
+metadata:
+  name: kaniko
+  namespace: "${namespace}"
+spec:
+  template:
+    spec:
+      containers:
+      - name: kaniko
+        image: gcr.io/kaniko-project/executor:latest
```
m1kola, line 234: Could you please explain why we need to build image from within the cluster? Is it possible to avoid dependency on this? E.g. build locally + kind load?

everettraven: We have to build within the cluster because catalogd now relies on fetching information directly from an image registry. Unfortunately kind loading isn't sufficient because it doesn't load it onto a "real" registry that is accessible by catalogd. I went through various iterations (port-forward then push, followed KinD documentation on using a local registry, etc.) when making the change to catalogd and building and pushing within the cluster using kaniko was the solution that worked best for running the test on both mac and linux. When attempting the alternatives there was always a situation where it worked for mac or linux but not the other.

m1kola: Got it, thanks for the background!

test/operator-framework-e2e/setup.sh
```diff
@@ -18,31 +18,35 @@ The following environment variables are required for configuring this script:
 setup.sh also takes 5 arguments. 
 
 Usage:
-  setup.sh [OPERATOR_SDK] [CONTAINER_RUNTIME] [KUSTOMIZE] [KIND] [KIND_CLUSTER_NAME]
+  setup.sh [OPERATOR_SDK] [CONTAINER_RUNTIME] [KUSTOMIZE] [KIND] [KIND_CLUSTER_NAME] [NAMESPACE]
```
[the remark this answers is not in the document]
everettraven: Ah, does look like I forgot to increase the argument count there. I think it could be helpful to check for a non-empty namespace but I think requiring it to be an argument alleviates the need for that. You'd have to explicitly make it an empty value

PR #497, opened 2023-10-31
Part 1: Reduce number of variable sources. Bundle uniqueness

m1kola:
# Description

In this part I extract code related to bundle uniqueness from `CRDUniquenessConstraintsVariableSource` into a separate function.

`CRDUniquenessConstraintsVariableSource`  will be removed later in #501

## Reviewer Checklist

- [ ] API Go Documentation


internal/resolution/variablesources/bundle_uniqueness.go
```diff
@@ -4,15 +4,52 @@ import (
...
 	"github.com/operator-framework/deppy/pkg/deppy"
 	"github.com/operator-framework/deppy/pkg/deppy/input"
-	"k8s.io/apimachinery/pkg/util/sets"
 
 	"github.com/operator-framework/operator-controller/internal/catalogmetadata"
 	olmvariables "github.com/operator-framework/operator-controller/internal/resolution/variables"
 )
 
-var _ input.VariableSource = &CRDUniquenessConstraintsVariableSource{}
+// MakeBundleUniquenessVariables produces variables that constraint
```
stevekuznetsov, line 16: s/constraint/constraints/

internal/resolution/variablesources/bundle_uniqueness.go
```diff
@@ -4,15 +4,52 @@ import (
...
+// the solution to at most 1 bundle per package.
+// These variables guarantee that no two version of
```
stevekuznetsov, line 18: s/version/versions/

internal/resolution/variablesources/bundle_uniqueness_test.go
```diff
@@ -17,6 +20,255 @@ import (
 	"github.com/operator-framework/operator-controller/internal/resolution/variablesources"
 )
 
+func TestMakeBundleUniquenessVariables(t *testing.T) {
```
stevekuznetsov, line 23: Is this the most minimal test we can create? There's a *lot* of setup here and it's not entirely clear to me that all of it is necessary - for instance, if you need to test dependencies through GVK and package name, it makes sense to create minimal setups for each. It's hard to follow the independent threads here and the size of the test dataset makes it difficult to ascertain relationships between entities.

internal/resolution/variablesources/bundle_uniqueness_test.go
```diff
@@ -17,6 +20,255 @@ import (
...
+			Package: "unrelated-package-2",
+			Properties: []property.Property{
+				{Type: property.TypePackage, Value: json.RawMessage(`{"packageName": "unrelated-package-2", "version": "3.0.0"}`)},
+				{Type: property.TypeGVK, Value: json.RawMessage(`[{"group":"buz.io","kind":"Buz","version":"v1alpha1"}]`)},
+			}},
+			InChannels: []*catalogmetadata.Channel{&channel},
+		},
+	}
+
+	t.Run("convert bundle variables into create global uniqueness constraint variables", func(t *testing.T) {
```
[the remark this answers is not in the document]
m1kola: I think I'll keep this subtest. Eventually we will need uniqueness to be applied also based on CRDs. I would imagine it will be a separete subtest.

I believe it doesn't hurt.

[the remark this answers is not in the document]
ncdc: `t.Run` does not follow the same philosophy as Ginkgo's Describe/It blocks, FYI. We really only use it for table-style tests to denote which test case we're running.

internal/resolution/variablesources/bundle_uniqueness_test.go
m1kola: Hm. I  took the test data from the old test (it is still below, but I copied it over so it is easier to clean it up later in #501). But it looks like we have some test data related to GVK uniqueness in here which we probably don't need at the moment.

Tomorrow I'll look into whether we can reduce the test data.

Background: we had some dead code related to GVK uniqueness and some tests for it. Code got removed, but it looks like data was not cleaned up.

internal/resolution/variablesources/bundle_uniqueness_test.go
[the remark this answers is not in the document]
m1kola: @ncdc @stevekuznetsov I'm not convinced. `t.Run` has many uses. Yes, table-style tests is one of them, but you can also use it for creating hierarchical tests and separating test cases from setup/tear down code.

I like how it separates generic catalog data from test case data.

I think Go testing framework is not as prescriptive as you might think: I do not see any documentation that supports the point that I'm doing something ilegal here.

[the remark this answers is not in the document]
stevekuznetsov: GoDoc for it seems pretty straightforward:

>Run runs f as a subtest of t called name. It runs f in a separate goroutine and blocks until f returns or calls t.Parallel to become a parallel test.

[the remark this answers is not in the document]
ncdc: Blocking comment: please remove this usage of `t.Run`. It is not consistent with how we write go subtests.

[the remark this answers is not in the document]
m1kola: > Blocking comment: please remove this usage of `t.Run`. It is not consistent with how we write go subtests.

@ncdc consistency is a good point, but I don't think it applies in this case. We have a variety of approaches accross operator framework and in this repo (and that is ok IMO). We have table-style tests with `t.Run` as well as tests like this [^1][^2].

If by we you mean Go developers in general - Go codebase itself uses `t.Run` for both table-style tests [^3] and tests like this [^4][^5].

I believe this should not be a reson to block the PR.

> GoDoc for it seems pretty straightforward:
> 
> > Run runs f as a subtest of t called name. It runs f in a separate goroutine and blocks until f returns or calls t.Parallel to become a parallel test.

@stevekuznetsov It does not say that I'm doing something wrong here. I have a test with one subtest. It seems like a normal thing to do in go code (see examples from Go sources).

I think saying that it is not a "sub" since it is the only one is not correct. E.g. subprocess is still a "sub" even if it is the only one.

[^1]: https://github.com/operator-framework/operator-controller/blob/71564646b879c211016bfb96fcd1fa034df6ccb1/internal/catalogmetadata/cache/cache_test.go#L54
[^2]: https://github.com/operator-framework/operator-controller/blob/71564646b879c211016bfb96fcd1fa034df6ccb1/internal/catalogmetadata/client/client_test.go#L37
[^3]: https://github.com/golang/go/blob/e73e25b624c37a936bb42f50a11f56297a4cd637/src/syscall/fs_wasip1_test.go#L59
[^4]: https://github.com/golang/go/blob/e73e25b624c37a936bb42f50a11f56297a4cd637/src/regexp/all_test.go#L968
[^5]: https://github.com/golang/go/blob/e73e25b624c37a936bb42f50a11f56297a4cd637/src/cmd/go/internal/modindex/index_test.go#L77

internal/resolution/variablesources/bundle_uniqueness.go
```diff
@@ -4,15 +4,52 @@ import (
...
+		}
+	}
+
+	// create global constraint variables
+	for _, packageName := range packageOrder {
+		varID := deppy.IdentifierFromString(fmt.Sprintf("%s package uniqueness", packageName))
+		result = append(result, olmvariables.NewBundleUniquenessVariable(varID, bundleOrder[packageName]...))
+	}
+
+	return result, nil
```
m1kola, line 51: I think error can be removed

internal/resolution/variablesources/bundle_uniqueness_test.go
m1kola: @stevekuznetsov I think this was a really good catch! I had a closer look at the test data and there was quite a lot of junk related to GVK uniqueness which we do not enforce at the moment. Now we are down to 3 bundles.

I also added some comments to the test to explain why we need test data to meet certain requirements. Hopefully it will help future readers.

internal/resolution/variablesources/bundle_uniqueness_test.go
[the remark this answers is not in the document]
m1kola: I made the change requested because I want to make progress and I do not think this detail is important enough to spend more cycles on this.

However I do not want this to create a precedent for the future.

I would like us to collaborate instead of prescribing each other specific ways. Usually it is possible to reach some consensus based on technical facts.

We can also discuss and document specific decisions or even code them into tools. E.g. if we want subtests to be used only for table-style tests - either all of our tests must be consistent or it should be discussed with the community (like we did [here](https://github.com/operator-framework/operator-controller/discussions/168) when we decied to get rid of ginkgo) and ideally documented in some sort of style guide.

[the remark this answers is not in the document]
ncdc: Links 1 and 2 are not how we should be writing tests.

There is no reason for links 4 or 5 to use t.Run for a subtest.

When I review PRs, I have certain things I look for. This is one of them. Please do not use t.Run for unnecessary subtests.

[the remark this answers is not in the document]
m1kola: > Links 1 and 2 are not how we should be writing tests.

> There is no reason for links 4 or 5 to use t.Run for a subtest.

@ncdc Could you please explain why? I believe I demonstrated evidence that previous arguments are not valid. Is there a technical reason for that? Do we have a OF agreed document/style guide am not aware of?

Multiple people authored it and multiple people approved each of these change sets. So opinions on this differ.

> When I review PRs, I have certain things I look for. This is one of them. Please do not use t.Run for unnecessary subtests.

I would like us to think collaboratively and avoid things like "please do it this way because I said so".

Maybe we should adopt something like [this](https://google.github.io/eng-practices/review/reviewer/standard.html)? See [principles](https://google.github.io/eng-practices/review/reviewer/standard.html#principles) section:

> * Technical facts and data overrule opinions and personal preferences.
> 
> * On matters of style, the style guide is the absolute authority. Any purely style point (whitespace, etc.) that is not in the style guide is a matter of personal preference. The style should be consistent with what is there. If there is no previous style, accept the author’s.
> 
> * Aspects of software design are almost never a pure style issue or just a personal preference. They are based on underlying principles and should be weighed on those principles, not simply by personal opinion. Sometimes there are a few valid options. If the author can demonstrate (either through data or based on solid engineering principles) that several approaches are equally valid, then the reviewer should accept the preference of the author. Otherwise the choice is dictated by standard principles of software design.
> 
> * If no other rule applies, then the reviewer may ask the author to be consistent with what is in the current codebase, as long as that doesn’t worsen the overall code health of the system.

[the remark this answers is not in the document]
ncdc: I am happy to write it down. There is no precedent to use t.Run in this way. This is also an experiential thing... so sometimes "please do it this way because I said so" is backed by years of experience reading, writing, and reviewing code.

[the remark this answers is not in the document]
m1kola: > I am happy to write it down. 

@ncdc thanks. Google has very nice guidance for [reviewers](https://google.github.io/eng-practices/review/reviewer/) and [authors](https://google.github.io/eng-practices/review/developer/) as well as style guide for different languages. I think we can learn a lot from it and can probably borrow some stuff from it.

I personally prefer not to have formal documents like this but it looks like we won't be able to work productively if we continue discuss matters like this in each PR.

> There is no precedent to use t.Run in this way.

But [I provided](https://github.com/operator-framework/operator-controller/pull/497#discussion_r1378620490) evidence that it is used this way in our codebase and in golang/go codebase...

> This is also an experiential thing... so sometimes "please do it this way because I said so" is backed by years of experience reading, writing, and reviewing code.

Please do not dismiss years of my experience and experience of other authors like that.

I think it is not a good idea to evaluate arguments on the accomplishments/success of the person making the argument. Let's evaluate on the merits of the argument itself.

It will be very hard to build a welcoming community with approach like this.

---

Back to the PR itself.

@ncdc @stevekuznetsov given that the PR was updated with the requested change - are we good to go? Or do you have any other feedback?

m1kola:
Rebased on top of `main`. Need this for rest of the PRs in the series. No other changes apart from the rebase since yesterday.

internal/resolution/variablesources/bundle_uniqueness.go
```diff
@@ -4,15 +4,52 @@ import (
...
+// the same package are running at the same time.
+func MakeBundleUniquenessVariables(bundleVariables []*olmvariables.BundleVariable) []*olmvariables.BundleUniquenessVariable {
+	result := []*olmvariables.BundleUniquenessVariable{}
+
+	bundleIDs := sets.Set[deppy.Identifier]{}
+	packageOrder := []string{}
+	bundleOrder := map[string][]deppy.Identifier{}
+	for _, bundleVariable := range bundleVariables {
+		bundles := []*catalogmetadata.Bundle{bundleVariable.Bundle()}
+		bundles = append(bundles, bundleVariable.Dependencies()...)
```
joelanford, line 28: Nit: we know size of bundles will be `1+len(bundleVariable.Dependencies)`. Does it make sense to allocate at the right size first?

Alternately, do we really even _need_ a bundles slice? Seems like we could make the `for` loop's contents its own function, and then do something like:

```go
updateMaps := func(bundles ...*olmvariables.BundleVariable) {
	for _, bundle := range bundles {
		id := olmvariables.BundleVariableID(bundle)
	 	// get bundleID package and update map
 		packageName := bundle.Package

		if _, ok := bundleOrder[packageName]; !ok {
	 		packageOrder = append(packageOrder, packageName)
 		}

	 	if !bundleIDs.Has(id) {
 			bundleIDs.Insert(id)
	 		bundleOrder[packageName] = append(bundleOrder[packageName], id)
 		}
	}
}
updateMaps(bundleVariable.Bundle())
updateMaps(bundleVariable.Dependencies()...)

File: internal/resolution/variablesources/bundle_uniqueness_test.go [hunk 7]
```diff
@@ -17,6 +21,97 @@ import (
...
+		olmvariables.NewBundleVariable(
+			bundleSet["test-package.v1.0.1"],
+			[]*catalogmetadata.Bundle{
+				bundleSet["some-package.v1.0.0"],
+			},
+		),
+	}
+
+	variables := variablesources.MakeBundleUniquenessVariables(bundleVariables)
+	require.Len(t, variables, 2)
```
stevekuznetsov, line 88: Is it possible to simply create a slice of struct literals here and use `cmp.Diff()` for testing that the output is exactly as expected?

internal/resolution/variablesources/bundle_uniqueness.go
m1kola: > Nit: we know size of bundles will be `1+len(bundleVariable.Dependencies)`. Does it make sense to allocate at the right size first?

Oh, yeah! Let's allocate once!

I'll update to something like this shortly:

```go
bundles := make([]*catalogmetadata.Bundle, 0, 1+len(bundleVariable.Dependencies()))
bundles = append(bundles, bundleVariable.Bundle())
bundles = append(bundles, bundleVariable.Dependencies()...)
```

internal/resolution/variablesources/bundle_uniqueness_test.go
m1kola: I think go-cmp ignores unexported fields by default. I'll see if I can make it work with `AllowUnexported`.

m1kola: Updated with go-cmp & `AllowUnexported` option. Looks cleaner.

internal/resolution/variablesources/bundle_uniqueness.go
m1kola: Updated

m1kola: > Alternately, do we really even _need_ a bundles slice? Seems like we could make the `for` loop's contents its own function, and then do something like

I think there still be slices in the variadic arugments. They are just implicit. So I went with the single allocaiton option.

m1kola:
@joelanford @stevekuznetsov I addressed the latest comments. Please take another look.

m1kola:
Had to re-push due to `github.com/google/go-cmp` moving from inderect dependency to derect one (`go.mod` needed an update). No other changes apart from this.

stevekuznetsov (approved):

PR #498, opened 2023-10-31
Part 2: Reduce number of variable sources. Bundle variables.

m1kola:
# Description

In this part I extract code related to creating bundle variables from `BundlesAndDepsVariableSource` into a separate function.

`BundlesAndDepsVariableSource` will be removed later in #501

## Reviewer Checklist

- [ ] API Go Documentation


internal/resolution/variablesources/bundle_test.go
```diff
@@ -4,19 +4,344 @@ import (
...
 
 	"github.com/operator-framework/operator-controller/internal/catalogmetadata"
 	olmvariables "github.com/operator-framework/operator-controller/internal/resolution/variables"
 	"github.com/operator-framework/operator-controller/internal/resolution/variablesources"
 	testutil "github.com/operator-framework/operator-controller/test/util"
 )
 
+func TestMakeBundleVariables(t *testing.T) {
+	channel := catalogmetadata.Channel{Channel: declcfg.Channel{Name: "stable"}}
+	allBundles := []*catalogmetadata.Bundle{
```
m1kola, line 25: Similar to part one: https://github.com/operator-framework/operator-controller/pull/497#discussion_r1377929603

Need to check whether we can reduce the size of the test data.

m1kola: Right, I think we can get rid of a bunch of bundles here as many of them also seem to be related to GVK requirements which we do not care about here.

m1kola: Got rid of more than half of test data

internal/resolution/variablesources/bundle.go
```diff
@@ -2,62 +2,36 @@ package variablesources
...
-		variables = append(variables, inputVariables...)
-	}
-
-	// create bundle queue for dependency resolution
+func MakeBundleVariables(
+	allBundles []*catalogmetadata.Bundle,
+	requiredPackages []*olmvariables.RequiredPackageVariable,
+	installedPackages []*olmvariables.InstalledPackageVariable,
+) ([]*olmvariables.BundleVariable, error) {
 	var bundleQueue []*catalogmetadata.Bundle
```
joelanford, line 25: Allocate with an initial capacity:
```suggestion
	bundleQueue := make([]*catalogmetadata.Bundle, 0, len(requiredPackages) + len(installedPackages))
```

internal/resolution/variablesources/bundle.go
```diff
@@ -2,62 +2,36 @@ package variablesources
...
-
-	allBundles, err := b.catalogClient.Bundles(ctx)
-	if err != nil {
-		return nil, err
+	for _, variable := range installedPackages {
+		bundleQueue = append(bundleQueue, variable.Bundles()...)
 	}
 
 	// build bundle and dependency variables
+	var result []*olmvariables.BundleVariable
```
joelanford, line 34: We expect result to always need capacity greater than or equal to length of requiredPackages, right? And (for now at least), also installedPackages?
```suggestion
	result := make([]*olmvariables.BundleVariable, 0, len(requiredPackages) + len(installedPackages))
```

internal/resolution/variablesources/bundle.go
```diff
@@ -82,23 +56,25 @@ func (b *BundlesAndDepsVariableSource) GetVariables(ctx context.Context) ([]depp
...
 	requiredPackages, _ := bundle.RequiredPackages()
 	for _, requiredPackage := range requiredPackages {
-		packageDependencyBundles := catalogfilter.Filter(allBundles, catalogfilter.And(catalogfilter.WithPackageName(requiredPackage.PackageName), catalogfilter.InBlangSemverRange(requiredPackage.SemverRange)))
+		packageDependencyBundles := catalogfilter.Filter(allBundles, catalogfilter.And(
+			catalogfilter.WithPackageName(requiredPackage.PackageName),
+			catalogfilter.InBlangSemverRange(requiredPackage.SemverRange),
+		))
 		if len(packageDependencyBundles) == 0 {
-			return nil, fmt.Errorf("could not find package dependencies for bundle '%s'", bundle.Name)
+			return nil, errors.New("could not find package dependencies for bundle")
```
joelanford, line 77: This is probably a broader structural question, but shouldn't this case be propagated into the SAT solver so that we get an UNSAT answer, rather than what appears in the current code to be just an error building the problem (which means we never progress to even running the solver).

This may not be important right now, but it could become more important if we ever want to try to find sub-solutions for disjoint sub-problems.

internal/resolution/variablesources/bundle.go
```diff
@@ -117,3 +93,53 @@ func (b *BundlesAndDepsVariableSource) filterBundleDependencies(allBundles []*ca
...
+		if err != nil {
+			return nil, err
+		}
+		variables = append(variables, inputVariables...)
+	}
+
+	allBundles, err := b.catalogClient.Bundles(ctx)
+	if err != nil {
+		return nil, err
+	}
```
joelanford, line 123: Help me remember. Is this call eventually being removed by the time we get to the end of the PR chain? Just want to make sure we aren't ending up with clients sourcing input under the covers anywhere.

internal/resolution/variablesources/bundle_test.go
```diff
@@ -4,19 +4,196 @@ import (
...
+	"github.com/stretchr/testify/require"
 
 	"github.com/operator-framework/operator-controller/internal/catalogmetadata"
 	olmvariables "github.com/operator-framework/operator-controller/internal/resolution/variables"
 	"github.com/operator-framework/operator-controller/internal/resolution/variablesources"
 	testutil "github.com/operator-framework/operator-controller/test/util"
 )
 
+func TestMakeBundleVariables(t *testing.T) {
+	t.Run("valid dependencies", func(t *testing.T) {
```
joelanford, line 24: Since the `t.Run()`s here have nothing shared, the typical pattern I've seen is to split them into separate top-level `Test*` functions:

```go
func TestMakeBundleVariables_ValidDepedencies(t *testing.T) {}
func TestMakeBundleVariables_NonExistentDepedencies(t *testing.T) {}
```

internal/resolution/variablesources/bundle.go
joelanford: Here's how I handled this in my PoC: https://github.com/joelanford/operator-controller/blob/31082f42fa9fbe61a4c9155683d0956e2425b774/internal/resolution/v2/variablesources/operator.go#L59-L69

internal/resolution/variablesources/bundle.go
m1kola: I think here we should not do that becuase `bundleQueue` is for bundles on which `requiredPackages` and `installedPackages` variables depend and for dependencies of these bundles.

`len(requiredPackages) + len(installedPackages)` will not correspond to the amount of data we handle and the initial capacity will not be meaningful.

internal/resolution/variablesources/bundle.go
m1kola: This specific change to remove repetition: this error is in context of handling a specific bundle + the calling code adds an id of the bundle into the error message already.

> This is probably a broader structural question, but shouldn't this case be propagated into the SAT solver so that we get an UNSAT answer, rather than what appears in the current code to be just an error building the problem (which means we never progress to even running the solver).

I agree - this should be clear what went wrong during resolution form unsat error.

But can we handle it separately? I would like to avoid extending scope of this refactoring any further.

internal/resolution/variablesources/bundle.go
m1kola: Another reason is that - I think it is a bit confusing. When I see a `make` like this - it looks like we allocating a new bigger slice to accomadate both `requiredPackages` and `installedPackages`. But it is not the case.

internal/resolution/variablesources/bundle.go
m1kola: > We expect result to always need capacity greater than or equal to length of requiredPackages, right? And (for now at least), also installedPackages?

That is correct, but there is no direct relation between size of these two slices and the final slice (we go through dependencies recursively).

I think here it is better to let Go use default algorithm for expanding the slice.

Aslo https://github.com/operator-framework/operator-controller/pull/498#discussion_r1380172620 and the next comment in this thread. I think it is very similar

internal/resolution/variablesources/bundle.go
m1kola: > Help me remember. Is this call eventually being removed by the time we get to the end of the PR chain?

Correct, I'm cheating here. Passing all bundles trough current structure is going to be challenging.

This whole variable source will go away in the final part #501: https://github.com/operator-framework/operator-controller/pull/501/commits/27dd463595641d2068d8733d280d0af4f1b8f663#diff-26d40ef0af5660e5b7fd99da5952b39c645bef2ad6bf2ee6d8b7123bbdd2ff73

internal/resolution/variablesources/bundle_test.go
```diff
@@ -4,19 +4,196 @@ import (
...
+		// Each dependency must have a variable.
+		// Dependencies from the same package must be sorted by version
+		// with higher versions first.
+		expectedIDs := []deppy.Identifier{
+			"fake-catalog-first-level-dependency-first-level-dependency.v1.0.0",
+			"fake-catalog-second-level-dependency-second-level-dependency.v1.0.1",
+			"fake-catalog-second-level-dependency-second-level-dependency.v1.0.0",
+		}
+		actualIDs := collectVariableIDs(bundles)
+		assert.Equal(t, expectedIDs, actualIDs)
```
m1kola, line 144: Need to check how tricky it is going to be to do something similar to #497 with `go-cmp` and `AllowUnexported`.

internal/resolution/variablesources/bundle.go
joelanford: I didn't look closely enough to see if there's a minimum bound on the high-water mark of the `bundleQueue`. My general point is: "let's always pre-allocate if the minimum high-water mark is greater than 0".

If, in this case, that minimum high-water mark is 0, then I withdraw my suggestion.

joelanford: Even if the slice can end up being bigger than the minimum bound, it still saves on allocations and copies for all of the intermediate allocations/copies that we would have made otherwise.

internal/resolution/variablesources/bundle.go
joelanford: Yep similar question, similar response: https://github.com/operator-framework/operator-controller/pull/498#discussion_r1380445055

internal/resolution/variablesources/bundle.go
joelanford: Yes, I'm like 99% positive there are other places that return errors like this when we should be taking these situations and making them unsatisfiable variables.

Totally fine by me to find/replace all of those as a separate effort rather than play whack-a-mole.

m1kola: > Totally fine by me to find/replace all of those as a separate effort rather than play whack-a-mole.

I'll check if we already have an issue for this and will create one if we don't. It will probably be tomorrow.

Please let's keep this thread open for this follow up^.

internal/resolution/variablesources/bundle.go
m1kola: On this function's level - it is possible that it will be 0, but looking at the bigger picture  - I think will never reach this function without bundles for `requiredPackages` and `installedPackages`. Each item in these slices should have >= 1 bundle.

Updated.

internal/resolution/variablesources/bundle.go
m1kola: Updated: https://github.com/operator-framework/operator-controller/pull/498#discussion_r1380489544

internal/resolution/variablesources/bundle.go
m1kola: I did not find an issue for this. I need to run now, but I have a reminder to create an issue for this set for tomorrow.

m1kola:
@joelanford please take another look. I believe I address all your feedback.

internal/resolution/variablesources/bundle.go
m1kola: Created an issue for this: https://github.com/operator-framework/operator-controller/issues/507

internal/resolution/variablesources/bundle.go
m1kola: @joelanford I came up with a relatively small change so we can get rid of this: https://github.com/operator-framework/operator-controller/pull/508

This PR also enables me to progress with the last bit of implementation for initial upgrade support RFC (#449) without having to wait for part 2-5 of this refactorign to land.

internal/resolution/variablesources/bundle.go
```diff
@@ -2,62 +2,36 @@ package variablesources
...
-func NewBundlesAndDepsVariableSource(catalogClient BundleProvider, inputVariableSources ...input.VariableSource) *BundlesAndDepsVariableSource {
-	return &BundlesAndDepsVariableSource{
-		catalogClient:   catalogClient,
-		variableSources: inputVariableSources,
+func MakeBundleVariables(
+	allBundles []*catalogmetadata.Bundle,
+	requiredPackages []*olmvariables.RequiredPackageVariable,
+	installedPackages []*olmvariables.InstalledPackageVariable,
+) ([]*olmvariables.BundleVariable, error) {
+	bundleQueue := make([]*catalogmetadata.Bundle, 0, len(requiredPackages)+len(installedPackages))
```
stevekuznetsov, line 25: `len(requiredPackages)+len(installedPackages)` is an inexact approximation of the size you need, since each of those has a 1:N relationship to the underlying bundles? Personal preference, but `var bundleQueue []*catalogmetadata.Bundle` is fine unless benchmarks prove otherwise. Pointers are small.

internal/resolution/variablesources/bundle.go
```diff
@@ -2,62 +2,36 @@ package variablesources
...
+	for _, variable := range requiredPackages {
+		bundleQueue = append(bundleQueue, variable.Bundles()...)
 	}
-}
-
-func (b *BundlesAndDepsVariableSource) GetVariables(ctx context.Context) ([]deppy.Variable, error) {
```
stevekuznetsov, line 32: A lot of this diff seems like movement of code to later in the file - can we undo that to make the diff as minimal as we can? Hard to tell if you changed the code in `GetVariables()` or not.

internal/resolution/variablesources/bundle_test.go
```diff
@@ -374,7 +563,7 @@ var _ = Describe("BundlesAndDepsVariableSource", func() {
 		)
 		_, err := bdvs.GetVariables(context.TODO())
 		Expect(err).To(HaveOccurred())
-		Expect(err.Error()).To(ContainSubstring("could not determine dependencies for bundle with id 'fake-catalog-test-package-bundle-2': could not find package dependencies for bundle 'bundle-2'"))
+		Expect(err.Error()).To(ContainSubstring(`could not determine dependencies for bundle with id "fake-catalog-test-package-bundle-2": could not find package dependencies for bundle`))
```
stevekuznetsov, line 566: While a savvy user that has memorized how to disambiguate the bundle ID into the separate parts (let's hope nobody puts a dash in the catalog name, package name or bundle name ...), all other users would prefer the more explicit error message you removed. It might be better to simply fix the top-level wrapper here to break up the bundle ID into constituent parts (is the transform lossy?).

internal/resolution/variablesources/bundle_test.go
```diff
@@ -4,19 +4,208 @@ import (
...
+		),
+		olmvariables.NewBundleVariable(
+			bundleSet["second-level-dependency.v1.0.0"],
+			nil,
+		),
+	}
+	gocmpopts := []cmp.Option{
+		cmpopts.IgnoreUnexported(sync.RWMutex{}),
+		cmp.AllowUnexported(
+			sync.RWMutex{},
```
stevekuznetsov, line 161: I think we can be certain that the test - and the caller in this package - will never have any opinion on the unexported internal state of the mutex.

internal/resolution/variablesources/bundle.go
m1kola: That was initial version and is my preference too. We had a discussion about this here: https://github.com/operator-framework/operator-controller/pull/498#discussion_r1380044187

I'm happy to revert this back to `var bundleQueue []*catalogmetadata.Bundle`.

@joelanford any objections?

internal/resolution/variablesources/bundle.go
m1kola: I'm not sure that I can minimise the diff further. I took out the part out of `GetVariables()` and moved into `MakeBundleVariables`. Also I'm happy that the important bits (where we generate variables) is now in `MakeBundleVariables` and history of these lines will be preserved.

`BundlesAndDepsVariableSource` and its `GetVariables` is now just a bridge between old structure and new structure. It will be completely removed in #501.

internal/resolution/variablesources/bundle.go
joelanford: I think @stevekuznetsov and I have pretty much the exact same perspective, and we just fell on either side of the fence in terms of what to do.

In my mind, we _know_ `len(requiredPackages)+len(installedPackages)` is a strictly better approximation than `0`, so that's why I fell on that side.

I also prefaced my comment with "this may not matter, benchmarks would help", so I don't have a strong opinion.

Maybe the outcome of this should be a follow-on issue that says "let's write some benchmarks and have a way to measure how we're doing".

internal/resolution/variablesources/bundle_test.go
m1kola: I wanted to ignore it (I have `cmpopts.IgnoreUnexported(sync.RWMutex{})` for that reason above). This line is a junk I forgot to remove. I'll update shortly.

internal/resolution/variablesources/bundle_test.go
```diff
@@ -4,19 +4,208 @@ import (
...
+			olmvariables.BundleVariable{},
+			input.SimpleVariable{},
+			constraint.DependencyConstraint{},
+			catalogmetadata.Bundle{},
```
m1kola, line 165: I think we can put this into `IgnoreUnexported`. Unexported fields are internal state and exported fields will be enough for comparions.

internal/resolution/variablesources/bundle_test.go
joelanford: +1

Also, sort of related: I've noticed we don't really have a standard for how to build variable IDs. Another separate structural issue to follow-up on is to maybe write up a developer doc and/or write and ubiquitously use a helper function that can build/validate ID formats.

That could go a long way in making the bundle IDs more meaningful.

m1kola: I removed the part which was ambiguous: there could be multiple bundles with the same name from different catalogs/packages.

But I gree with you - this error message is not great. We need something more human readable. Good news is that we don't need to split bundle ID back into components: we have all the data in our structs.

I think what I'll do - I'll revert this change and deal with this in a separate PR.

internal/resolution/variablesources/bundle.go
stevekuznetsov: I think you might be able to - try moving the code at the bottom of the file to the top and see.

internal/resolution/variablesources/bundle.go
stevekuznetsov: My perspective is that a couple things have to be true before it's worth doing something like that:

a) the code in question is performance sensitive
b) we have a way to measure it
c) the naive performance is insufficient

I know loads of performance sensitive code where either _N_ is small and the few copies don't matter or the majority of time is taken on other work, so even a 99% speedup in alloc makes negligible/unmeasureably small impact on the overall runtime.

The reason _not_ to do this unless the prerequisites are met is that it's easy to do subtly incorrect things with this - especially when performance sensitive code uses pre-allocated length with `[i]` member access instead of `append`, so not asking readers to have the mental overhead unless we know we need to seems good.

Feel free to keep the code here as it is, as I don't feel super strongly on this, but overall IMHO we benefit from KISS.

m1kola:
Moveing back to draft for now becuase it needs a bit more work after #508 + address feedback.

internal/resolution/variablesources/bundle.go
m1kola: You are right. It produced a smaller diff and still preserves history for `MakeBundleVariables`. I'll push an update shortly.

internal/resolution/variablesources/bundle_test.go
m1kola: I think this will also spare us from using `IgnoreFields`.

m1kola:
@stevekuznetsov @joelanford rebased on top of #508 + addressed feedback. Please take another look

stevekuznetsov (approved):

Issue #443, opened 2023-10-06
Installation of an `Operator` fails when `BundleDeployment` with the same bundle already exists


m1kola:
**Steps to reproduce:**

1. Create a catalog on cluster: `kubectl apply -f config/samples/catalogd_operatorcatalog.yaml`
1. Create a `BundleDeployment`:
  ```yaml
  apiVersion: core.rukpak.io/v1alpha1
  kind: BundleDeployment
  metadata:
    name: operator-sample-manual
  spec:
    provisionerClassName: core-rukpak-io-plain
    template:
      metadata: {}
      spec:
        provisionerClassName: core-rukpak-io-registry
        source:
          image:
            ref: quay.io/operatorhubio/argocd-operator@sha256:1a9b3c8072f2d7f4d6528fa32905634d97b7b4c239ef9887e3fb821ff033fef6
          type: image
  ```
1. Create an `Operator` which resolves to the same bundle:
  ```yaml
  apiVersion: operators.operatorframework.io/v1alpha1
  kind: Operator
  metadata:
    labels:
      app.kubernetes.io/name: operator
      app.kubernetes.io/instance: operator-sample
      app.kubernetes.io/part-of: operator-controller
      app.kubernetes.io/managed-by: kustomize
      app.kubernetes.io/created-by: operator-controller
    name: operator-sample
  spec:
    packageName: argocd-operator
    version: 0.6.0
  ```

**Actual result:**
```yaml
apiVersion: operators.operatorframework.io/v1alpha1
kind: Operator
metadata:
  # ...
  name: operator-sample
spec:
# ...
status:
  conditions:
  - lastTransitionTime: "2023-10-06T12:26:51Z"
    message: resolved to "quay.io/operatorhubio/argocd-operator@sha256:1a9b3c8072f2d7f4d6528fa32905634d97b7b4c239ef9887e3fb821ff033fef6"
    observedGeneration: 1
    reason: Success
    status: "True"
    type: Resolved
  - lastTransitionTime: "2023-10-06T12:26:57Z"
    message: 'bundledeployment not ready: rendered manifests contain a resource that
      already exists. Unable to continue with install: Namespace "argocd-operator-system"
      in namespace "" exists and cannot be imported into the current release: invalid
      ownership metadata; annotation validation error: key "meta.helm.sh/release-name"
      must equal "operator-sample": current value is "operator-sample-manual"'
    observedGeneration: 1
    reason: InstallationFailed
    status: "False"
    type: Installed
[...]

PR #499, opened 2023-10-31
🌱 Part 3: Reduce number of variable sources. Installed packages

m1kola:
# Description

In this part I extract code related to creating installed package variables from `InstalledPackageVariableSource` and `BundleDeploymentVariableSource` into a separate function.

`InstalledPackageVariableSource` gets removed in this PR. `BundleDeploymentVariableSource` will be removed later in #501

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/resolution/variablesources/bundle_deployment.go
+++ b/internal/resolution/variablesources/bundle_deployment.go
@@ -3,8 +3,6 @@ package variablesources
 import (
 	"context"
 
-	"k8s.io/apimachinery/pkg/util/sets"
-
 	"github.com/operator-framework/deppy/pkg/deppy"
 	"github.com/operator-framework/deppy/pkg/deppy/input"
 	rukpakv1alpha1 "github.com/operator-framework/rukpak/api/v1alpha1"
@@ -34,21 +32,18 @@ func (o *BundleDeploymentVariableSource) GetVariables(ctx context.Context) ([]de
 		variableSources = append(variableSources, o.inputVariableSource)
 	}
 
-	processed := sets.Set[string]{}
-	for _, bundleDeployment := range o.bundleDeployments {
-		sourceImage := bundleDeployment.Spec.Template.Spec.Source.Image
-		if sourceImage != nil && sourceImage.Ref != "" {
-			if processed.Has(sourceImage.Ref) {
-				continue
-			}
-			processed.Insert(sourceImage.Ref)
-			ips, err := NewInstalledPackageVariableSource(o.allBundles, bundleDeployment.Spec.Template.Spec.Source.Image.Ref)
-			if err != nil {
-				return nil, err
-			}
-			variableSources = append(variableSources, ips)
-		}
+	variables, err := variableSources.GetVariables(ctx)
+	if err != nil {
+		return nil, err
 	}
 
-	return variableSources.GetVariables(ctx)
+	installedPackages, err := MakeInstalledPackageVariables(o.allBundles, o.bundleDeployments)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, v := range installedPackages {
+		variables = append(variables, v)
+	}
+	return variables, nil
 }
--- a/internal/resolution/variablesources/installed_package.go
+++ b/internal/resolution/variablesources/installed_package.go
@@ -1,71 +1,74 @@
 package variablesources
 
 import (
-	"context"
 	"fmt"
 	"sort"
 
 	mmsemver "github.com/Masterminds/semver/v3"
-	"github.com/operator-framework/deppy/pkg/deppy"
-	"github.com/operator-framework/deppy/pkg/deppy/input"
+	"k8s.io/apimachinery/pkg/util/sets"
+
+	rukpakv1alpha1 "github.com/operator-framework/rukpak/api/v1alpha1"
 
 	"github.com/operator-framework/operator-controller/internal/catalogmetadata"
 	catalogfilter "github.com/operator-framework/operator-controller/internal/catalogmetadata/filter"
 	catalogsort "github.com/operator-framework/operator-controller/internal/catalogmetadata/sort"
-	"github.com/operator-framework/operator-controller/internal/resolution/variables"
+	olmvariables "github.com/operator-framework/operator-controller/internal/resolution/variables"
 	"github.com/operator-framework/operator-controller/pkg/features"
 )
 
-var _ input.VariableSource = &InstalledPackageVariableSource{}
-
-type InstalledPackageVariableSource struct {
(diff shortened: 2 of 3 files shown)


m1kola:
Will keep in draft for now. I think "[FAIL] Operator Install when An operator is installed from an operator catalog [It] resolves again when a new catalog is available" failure is a legit one. From a quick look I do not see what is causing it.

Everything else seems to be a result of it (looks like some dependency between e2e tests).

m1kola:
> Will keep in draft for now. I think "[FAIL] Operator Install when An operator is installed from an operator catalog [It] resolves again when a new catalog is available" failure is a legit one. From a quick look I do not see what is causing it.

it was legit issue: I was doing things in incorrect order in `BundleDeploymentVariableSource`. Fixed now.

> Everything else seems to be a result of it (looks like some dependency between e2e tests).

Created an issue for this: https://github.com/operator-framework/operator-controller/issues/512

internal/resolution/variablesources/installed_package.go
```diff
@@ -1,71 +1,71 @@
...
+			continue
+		}
+		processed.Insert(sourceImage.Ref)
+
+		bundleImage := sourceImage.Ref
+
+		// find corresponding bundle for the installed content
+		resultSet := catalogfilter.Filter(allBundles, catalogfilter.WithBundleImage(bundleImage))
+		if len(resultSet) == 0 {
+			return nil, fmt.Errorf("bundleImage %q not found", bundleImage)
```
ncdc, line 50: When this error surfaces, the reader won't know what `bundleImage` is or necessarily where it's coming from. Would it be possible to revise the text to try to minimize confusion?

m1kola: This one is a bit tricky becuase this is a situaion when we have something installed on the cluster, but do not know what it is.

I imagine we can end up here with steps like this:

* User creates a catalog
* Installs a pacakge
* Removes the catalog

At this point we are unable to determine what bundle was installed.

For #449 I'll (likely) be modifying this function so that we will have a relevant `Operator` here. After this change we will probably be able to use `Operator`'s spec to add more context (e.g. package name).

But there are still things related to #449 I'm not sure about. For example, today we take all `BundleDeployment`s into account (even ones created directly). Doing the above will likely mean that we will only care about `BundleDeployment`s which we manage (ones owned by an `Operator`). So this is a bit in the air at the moment. Hopefully I'll get to it next week. Relevant discussions:
* https://github.com/operator-framework/operator-controller/pull/460#discussion_r1376252897
* https://github.com/operator-framework/operator-controller/issues/443

everettraven: Even though we don't have bundle or package information I think we have enough information on hand to provide a more clear error message that gives someone enough information to take some kind of remediation action if necessary. Please correct me if I'm wrong, but my assumption here is that this results in a failed resolution and prevents an Operator from successfully installing.

Maybe something like:
```suggestion
			return nil, fmt.Errorf("bundle with image %q not found in set of installable content but is installed via BundleDeployment %q", bundleImage, bundleDeployment.Name)
```

internal/resolution/variablesources/bundle_deployment.go
```diff
@@ -34,21 +32,18 @@ func (o *BundleDeploymentVariableSource) GetVariables(ctx context.Context) ([]de
...
-			variableSources = append(variableSources, ips)
-		}
+	variables, err := variableSources.GetVariables(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	installedPackages, err := MakeInstalledPackageVariables(o.allBundles, o.bundleDeployments)
+	if err != nil {
+		return nil, err
```
everettraven, line 42: Would it be beneficial to wrap these errors with more context before returning?

internal/resolution/variablesources/installed_package.go
m1kola: I just remembered that there was an idea to add labels to `BundleDeployment` to contain package name, bundle name, catalog name. It will help a bit, but we will still have an issue with BDs created derectly.

@everettraven yeah, it looks like having a BD's name in here will be helpful. I'll take a look into your suggestion on Monday.

---

I'll probably create a separate PR for this because most likely some tests will have to be updated as well.

Please let's keep this thread open until I come back to this on Monday.

internal/resolution/variablesources/bundle_deployment.go
m1kola: I think on this level we should just pass them down the call stack. In `MakeInstalledPackageVariables` we should be able to add a bit more context (e.g. specific bundle deployment name).

internal/resolution/variablesources/installed_package.go
m1kola: Created a separate PR for this: https://github.com/operator-framework/operator-controller/pull/514

It will need a rebase once we merge this one.

internal/resolution/variablesources/installed_package.go
```diff
@@ -1,71 +1,71 @@
...
 
-	return &InstalledPackageVariableSource{
-		allBundles:  allBundles,
-		bundleImage: bundleImage,
-		successors:  successors,
-	}, nil
+	result := make([]*olmvariables.InstalledPackageVariable, 0, len(bundleDeployments))
+	processed := sets.Set[string]{}
+	for _, bundleDeployment := range bundleDeployments {
+		sourceImage := bundleDeployment.Spec.Template.Spec.Source.Image
```
joelanford, line 35: This can panic, in theory, if `bundleDeployment.Spec.Template` is `nil` (not sure why that's even a pointer, to be honest)

ncdc: You want to make it a non-pointer in this PR?

m1kola: I'll add a `nil` check, but I would prefer not to increase the scope of this PR (it will require another PR to rukpak, rukpak release and a bump here).

internal/resolution/variablesources/bundle_deployment.go
m1kola: @everettraven I'm going to resolve this thread. Please re-open if you think we should change something here.

ncdc (approved):

ncdc:
@joelanford any other comments?

internal/resolution/variablesources/installed_package.go
m1kola: Check added. I created a PR for pointer here: https://github.com/operator-framework/rukpak/pull/758

joelanford (approved):

PR #500, opened 2023-10-31
🌱 Part 4: Reduce number of variable sources. Required packages

m1kola:
# Description

In this part I extract code related to creating required package variables from `RequiredPackageVariableSource` and `OperatorVariableSource` into a separate function.

`RequiredPackageVariableSource` gets removed in this PR. `OperatorVariableSource` will be removed later in #501

## Reviewer Checklist

- [ ] API Go Documentation


internal/controllers/operator_controller_test.go
```diff
@@ -83,7 +83,7 @@ var _ = Describe("Operator Controller Test", func() {
 				By("running reconcile")
 				res, err := reconciler.Reconcile(ctx, ctrl.Request{NamespacedName: opKey})
 				Expect(res).To(Equal(ctrl.Result{}))
-				Expect(err).To(MatchError(fmt.Sprintf("no package '%s' found", pkgName)))
```
m1kola, line 86: Since I had similar comments on previous parts of this refactoring & most of our errors use double quotation marks I updated the format here.

I'm happy to revert this and move into a separate PR if this is too distracting

joelanford: +1 on keeping this

internal/resolution/variablesources/required_package.go
```diff
@@ -1,99 +1,64 @@
...
+		if len(resultSet) == 0 {
+			if versionRange != "" && channelName != "" {
+				return nil, fmt.Errorf("no package %q matching version %q found in channel %q", packageName, versionRange, channelName)
+			}
+			if versionRange != "" {
+				return nil, fmt.Errorf("no package %q matching version %q found", packageName, versionRange)
+			}
+			if channelName != "" {
+				return nil, fmt.Errorf("no package %q found in channel %q", packageName, channelName)
+			}
```
joelanford, line 53: What do you think about building up the error message as we're building the predicates (since we're already making the necessary conditional checks as we go)?

m1kola: I think I'm not in favour of this approach because:
1. There are more combinations of errors than we have conditions for predicates. Unless I'm missing something, we will have to add more `if`s under existing predicate `if`s to get equivalent behaviour
2. Might be just me, but when I debug an error or read code I appreicate when codebase has unique errors and it is clear which condition yields the error. I found it quite hard to follow what was going on with errors in [this PoC](https://github.com/operator-framework/operator-controller/pull/331) which had similar approach.

dtfranz: I agree with the emphasis on readability here.

awgreene (approved):
Nice work @m1kola

m1kola:
For some reasons jobs in the merge queue did not get triggered. Probably will have to wait for it to timeout and requeue again.

PR #501, opened 2023-10-31
🌱 Part 5: Reduce number of variable sources. Final cleanup

### LINKED ISSUE #437: Reduce number of variable sources involved in resolution process

m1kola:
# Description

Spliting https://github.com/operator-framework/operator-controller/pull/460 into smaller chunks.

In this PR I replace controller variable source with an implementation containing a bunch of function calls (from parts 1-4). I also remove old variable sources and related tests

Closes #437

## Reviewer Checklist

- [ ] API Go Documentation


internal/controllers/variable_source_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+	require.Len(t, vars, 4)
+	assert.Equal(t, "required package packageA", vars[0].Identifier().String())
+	assert.IsType(t, &olmvariables.RequiredPackageVariable{}, vars[0])
+	assert.Equal(t, "installed package packageA", vars[1].Identifier().String())
+	assert.IsType(t, &olmvariables.InstalledPackageVariable{}, vars[1])
+	assert.Equal(t, "fake-catalog-packageA-packageA.v2.0.0", vars[2].Identifier().String())
+	assert.IsType(t, &olmvariables.BundleVariable{}, vars[2])
+	assert.Equal(t, "packageA package uniqueness", vars[3].Identifier().String())
+	assert.IsType(t, &olmvariables.BundleUniquenessVariable{}, vars[3])
```
m1kola, line 113: See if we can replace it with go-cmp similar to other PRs.

awgreene (approved):

internal/controllers/variable_source.go
```diff
@@ -64,21 +63,35 @@ func (v *VariableSource) GetVariables(ctx context.Context) ([]deppy.Variable, er
...
+	}
+
+	bundles, err := variablesources.MakeBundleVariables(allBundles, requiredPackages, installedPackages)
+	if err != nil {
+		return nil, err
+	}
+
+	bundleUniqueness := variablesources.MakeBundleUniquenessVariables(bundles)
+
+	result := []deppy.Variable{}
```
ncdc, line 83: Maybe take a look at https://freshman.tech/snippets/go/concatenate-slices/, specifically the section "Concatenating multiple slices at once", as a way to do this more efficiently?

m1kola: @ncdc the above works when we want to concatenate slices of the same type. Here we have slices of multiple different types all of which implement `deppy.Variable` interface and we need to create a slice of `deppy.Variable`s. I don't think it will work here.

I considered doing pre-allocation like this:

```go
result := make([]deppy.Variable, 0, len(requiredPackages)+len(installedPackages)+len(bundles)+len(bundleUniqueness))
```

But decided not to go with it. It still requires a loop with an `append`. Also It seems brittle: I can definitely see myself not updating the `make` call with a new variable when adding another loop. But it avoids extra allocations.

PR #503, opened 2023-10-31
refactor operator-framework-e2e

### LINKED ISSUE #349: `make operator-developer-e2e` fails with podman
`make operator-developer-e2e` builds and pushes images into local registry. Unfortunately it fails with `podman` when it is being used as a replacement for `docker`.

```
pinging container registry localhost:5001: Get "https://localhost:5001/v2/": http: server gave HTTP response to HTTPS client
```

This can be bypassed by setting `--tls-verify` to `false` in `podman push`. However, It looks like docker CLI doens't have this option.

### LINKED ISSUE #350: `make operator-developer-e2e` fails when container with name "local-registry" is already exists
```
docker run -d -p 5001:5000 --restart=always --name local-registry registry:2
Error: creating container storage: the container name "local-registry" is already in use by 7a2822e1ac9473f75a0f1e1e42ae0704b2d3123b18c49181a089e4624feea350. You have to remove that container to be able to reuse that name: that name is already in use
make: *** [Makefile:154: deploy-local-registry] Error 125
````

### LINKED ISSUE #482: operator-framework-e2e: we should verify eventual consistency between catalog unpacking and operator resolution
In the follow code, we create a `Catalog` and wait for it to be unpacked before we create the `Operator`: https://github.com/operator-framework/operator-controller/blob/a00a93ffb00c3a9fe894c1974e17400933bffebd/test/operator-framework-e2e/operator_framework_test.go#L334-L337

We should change this (and similar) tests in all e2es such that `Catalog` and `Operator` creation/update happens roughly simultaneously or in random orders (e.g. create/update `Operator` before creating/updating a `Catalog`).

If we make this change, the expectation is that everything continues to work as is (modulo some timeout tweaks, perhaps) because we have ostensibly built an eventually consistent set of controllers.

### PR DESCRIPTION
**Pull request by everettraven**
# Description
- Adds a script called `setup.sh` to perform all the setup logic for the coast-to-coast e2e test
    - This makes it easier to read and modify the setup logic in the future
- Removes obsolete Go code and updates the Ginkgo tests as necessary
- Removes obsolete Makefile targets and updates the Makefile as necessary

The combination of the changes here makes the test still generally do the same things of:
- Create a temp directory for building operators
- Use the `operator-sdk` to:
    - create two operators and build their images, one for the registry+v1 and one for the plain+v0 bundle formats
    - generate and build the bundle image for the registry+v1 bundle format
- For the plain+v0 operator, create and build the plain+v0 bundle image
- Creates a File-Based Catalog (FBC) image that contains both the operators
- Load the operator, bundle, and catalog images onto a KinD cluster
- Clean up the temp directories
- Run the Ginkgo test suite that:
    - Creates a `Catalog` that references a catalog image, either the one with the plain+v0 bundle or the one with the registry+v1 bundle
    - Creates an `Operator` that references a package, either the one with the plain+v0 bundle or the one with the registry+v1 bundle
    - Verifies that the `Operator` eventually has the condition `Installed` with a status of `True` and a reason of `Success`

## Motivation
- Makes it easier to read and modify the operator-framework-e2e test. Particularly it will help with making the e2e changes needed for #475 easier
- resolves #482 
- resolves #349
- resolves #350

## Open Questions
- The previous implementation included "upgrading" an operator by changing its `.spec.version` to be the "new" version. This implementation does not have that and instead focuses on the coast-to-coast of build, package, deploy. Is it necessary to include the "upgrading" of an operator as part of this e2e test?

## Reviewer Checklist

- [ ] API Go Documentation

### CODE REVIEW

File: test/operator-framework-e2e/operator_framework_test.go [hunk 1]
```diff
@@ -95,892 +46,98 @@ var _ = BeforeSuite(func() {
...
-	if containerRuntime == "" {
-		containerRuntime = "docker"
+// Since the CATALOG_IMG environment variable is used to configure the
+// tag of the catalog image that is built see if it is specified and use
+// it if it is. Otherwise use the default that is used in setup.sh
+func getCatalogRef() string {
+	if val, exist := os.LookupEnv("CATALOG_IMG"); exist {
+		return val
 	}
+	return "oc-opdev-e2e.operatorframework.io/catalog:e2e"
```
stevekuznetsov, line 62: This is loosely coupled to `setup.sh` - it might be better to simply enforce that `$CATALOG_IMG` is given as an input.

test/operator-framework-e2e/setup.sh
```diff
@@ -0,0 +1,193 @@
+#! /bin/bash
+
+set -e
```
stevekuznetsov, line 3:
```sh
set -o errexit
set -o nounset
set -o pipefail
```

a) you always want nounset, and pipefails
b) write for reading, not for brevity

test/operator-framework-e2e/setup.sh
```diff
@@ -0,0 +1,193 @@
...
+
+#################################################################
+# setup.sh is used to build operators using the operator-sdk and
+# build the image + bundle image, and create a FBC image for the
+# following bundle formats:
+# - registry+v1
+# - plain+v0
+# This script will ensure that all images built are loaded onto
+# a KinD cluster with the name specified in the arguments.
+# The following environment variables areused for configuring this script:
```
stevekuznetsov, line 13:
```suggestion
# The following environment variables are used for configuring this script:
```

test/operator-framework-e2e/setup.sh
```diff
@@ -0,0 +1,193 @@
...
+# - $REGISTRY_BUNDLE_IMG - the tag for the registry+v1 operator bundle image.
+#     Defaults to "oc-opdev-e2e.operatorframework.io/registry-bundle:v0.0.1"
+# - $PLAIN_CTRL_IMG - the tag for the plain+v0 operator image.
+#     Defaults to "oc-opdev-e2e.operatorframework.io/registry:v0.0.1"
+# - $PLAIN_BUNDLE_IMG - the tag for the plain+v0 operator bundle image.
+#     Defaults to "oc-opdev-e2e.operatorframework.io/plain-bundle:v0.0.1"
+# - $CATALOG_IMG - the tag for the catalog image that contains the plain+v0 and registry+v0 operator bundle.
+#     Defaults to "oc-opdev-e2e.operatorframework.io/catalog:e2e"
+# setup.sh also takes 4 arguments. Usage:
+# setup.sh [OPERATOR_SDK] [CONTAINER_RUNTIME] [KUSTOMIZE] [KIND] [KIND_CLUSTER_NAME]
```
stevekuznetsov, line 25: It would be nice to detect when the user has not passed the correct number of args and print this

test/operator-framework-e2e/setup.sh
```diff
@@ -0,0 +1,193 @@
...
+#################################################################
+
+########################################
+# Setup temp dir and local variables
+########################################
+
+# We're going to do file manipulation, so let's work in a temp dir
+TMP_ROOT="$(mktemp -p . -d 2>/dev/null || mktemp -d ./tmpdir.XXXXXXX)"
```
stevekuznetsov, line 33:
```suggestion
TMP_ROOT="$(mktemp -d)"
```

everettraven: changed this in https://github.com/operator-framework/operator-controller/pull/503/commits/cc277fcd2703a5aaab68e7802480fc8b6fc7412b

test/operator-framework-e2e/setup.sh
everettraven: Done in https://github.com/operator-framework/operator-controller/pull/503/commits/cc277fcd2703a5aaab68e7802480fc8b6fc7412b - includes validation on required environment variables

test/operator-framework-e2e/setup.sh
everettraven: https://github.com/operator-framework/operator-controller/pull/503/commits/cc277fcd2703a5aaab68e7802480fc8b6fc7412b

test/operator-framework-e2e/setup.sh
everettraven: https://github.com/operator-framework/operator-controller/pull/503/commits/cc277fcd2703a5aaab68e7802480fc8b6fc7412b

test/operator-framework-e2e/operator_framework_test.go
everettraven: Done as of https://github.com/operator-framework/operator-controller/pull/503/commits/cc277fcd2703a5aaab68e7802480fc8b6fc7412b

m1kola:
Correct me if I'm wrong, but it seems that both #349 and #350 are resolved by this since we no longer use `local-registry`.

test/operator-framework-e2e/operator_framework_test.go
```diff
@@ -83,904 +34,96 @@ func TestOperatorFramework(t *testing.T) {
...
-		By("creating an operator CR and verifying the operator operations")
-		namespace := fmt.Sprintf("%s-system", sdkInfo.projectName)
-		operator, err = createOperator(ctx, catalogDInfo.operatorName, operatorAction.installVersion)
-		Expect(err).ToNot(HaveOccurred())
-		checkOperatorOperationsSuccess(operator, catalogDInfo.operatorName, operatorAction.installVersion, bundleInfo.baseFolderPath, namespace)
-
-		By("upgrading an operator and verifying the operator operations")
-		operator, err = upgradeOperator(ctx, catalogDInfo.operatorName, operatorAction.upgradeVersion)
-		Expect(err).ToNot(HaveOccurred())
-		checkOperatorOperationsSuccess(operator, catalogDInfo.operatorName, operatorAction.upgradeVersion, bundleInfo.baseFolderPath, namespace)
```
m1kola, line 215: It looks like update to a new version of the operator we built is missing from the new test.
Same for operator deletion below.

What is our position on this? Leave it to the regular e2e to test?

everettraven:
> #350

Looks like it. I looked through the issues to try and find the ones this resolved and missed these. Will add them as resolved to the PR description, thanks!

m1kola (commented):
Looks good overal. Just one question

test/operator-framework-e2e/operator_framework_test.go
everettraven: I had this as an open question in the PR description:

>The previous implementation included "upgrading" an operator by changing its .spec.version to be the "new" version. This implementation does not have that and instead focuses on the coast-to-coast of build, package, deploy. Is it necessary to include the "upgrading" of an operator as part of this e2e test?

My thinking here is that we should have this test focus solely on the coast-to-coast of build, package, deploy with OLMv1. It seems unnecessary to test the deletion and manual upgrade by modifying the version. If we want to include the coast-to-coast of build, package, deploy, build+package new version, auto-upgrade I think that is reasonable but would prefer that be done in a follow-up to keep the scope of this PR to the simplification of this test.

everettraven: I also don't think there is anything wrong with adding _additional_ coast-to-coast e2e tests as those workflows are enabled

m1kola: Ah, sorry. I missed that. That makes sense.

m1kola (approved):

m1kola:
Re: `go-apidiff` - we removed a bunch of e2e helper functions which were exported. I think we can ignore this.

PR #509, opened 2023-11-03
🌱 Improve error message about missing bundles

m1kola:
# Description

Follow up for https://github.com/operator-framework/operator-controller/pull/498#discussion_r1381696321

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/resolution/variablesources/bundle.go
+++ b/internal/resolution/variablesources/bundle.go
@@ -94,7 +94,7 @@ func MakeBundleVariables(
 		// get bundle dependencies
 		dependencies, err := filterBundleDependencies(allBundles, head)
 		if err != nil {
-			return nil, fmt.Errorf("could not determine dependencies for bundle with id %q: %w", id, err)
+			return nil, fmt.Errorf("could not determine dependencies for bundle %q from package %q in catalog %q: %s", head.Name, head.Package, head.CatalogName, err)
 		}
 
 		// add bundle dependencies to queue for processing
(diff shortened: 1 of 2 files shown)


internal/resolution/variablesources/bundle.go
```diff
@@ -94,7 +95,7 @@ func MakeBundleVariables(
 		// get bundle dependencies
 		dependencies, err := filterBundleDependencies(allBundles, head)
 		if err != nil {
-			return nil, fmt.Errorf("could not determine dependencies for bundle with id %q: %w", id, err)
+			return nil, fmt.Errorf("could not determine dependencies for bundle %q from package %q from catalog %q: %s", head.Name, head.Package, head.CatalogName, err)
```
m1kola, line 98: I don't like the wording: doesn't feel natural. Any suggestions?

everettraven: A couple options I could think of:
- `could not determine dependencies for bundle %q from package %q in catalog %q`. This is only a slight variation but prevents the repetition of the word "from"
- Something that includes additional context after the error message like: `could not determine dependencies for bundle %q. catalog: %q, package: %q`

m1kola: Simply replacing "from" with "in" sounds much better, thanks! I updated the PR with this version.

internal/resolution/variablesources/bundle.go
```diff
@@ -119,7 +120,7 @@ func filterBundleDependencies(allBundles []*catalogmetadata.Bundle, bundle *cata
 			catalogfilter.InBlangSemverRange(requiredPackage.SemverRange),
 		))
 		if len(packageDependencyBundles) == 0 {
-			return nil, fmt.Errorf("could not find package dependencies for bundle %q", bundle.Name)
+			return nil, errors.New("could not find package dependencies for bundle")
```
joelanford, line 123: Can we add more context to this as well? For example:
```suggestion
			return nil, fmt.Errorf("no bundles found matching required package %q in range %q", requiredPackage.PackageName, requiredPackage.SemverRange)
```

everettraven (approved):
LGTM

PR #520, opened 2023-11-06
✨ Add upgrade constraint policy

### LINKED ISSUE #449: Respect `.spec.upgradeConstraintPolicy`

m1kola:
To do:
- [ ] Merge #499
- [ ] Rebase
- [ ] Complete the PR

# Description

Closes #449

## Reviewer Checklist

- [ ] API Go Documentation


internal/resolution/variablesources/installed_package.go
```diff
@@ -76,6 +84,18 @@ func MakeInstalledPackageVariables(
 // Must not return installed bundle as a successor
 type successorsFunc func(allBundles []*catalogmetadata.Bundle, installedBundle *catalogmetadata.Bundle) ([]*catalogmetadata.Bundle, error)
 
+func successorsFuncForOperator(operator operatorsv1alpha1.Operator) successorsFunc {
+	if operator.Spec.UpgradeConstraintPolicy == operatorsv1alpha1.UpgradeConstraintPolicyIgnore {
+		return ignoreConstraints
```
joelanford, line 89: Instead of building a new upgrade edge constraint that says "you can upgrade from this bundle to anything", can we just _not_ add the constraint for the this operator in the first place?

Essentially, the behavior in the `Ignore` case should be identical to the "not yet installed" case, which you handle in line 36-40. I'd propose modifying that if statement to also check for `operator.Spec.UpgradeConstraintPolicy == operatorsv1alpha1.UpgradeConstraintPolicyIgnore`

internal/resolution/variablesources/installed_package.go
```diff
@@ -22,16 +24,21 @@ import (
...
-	if features.OperatorControllerFeatureGate.Enabled(features.ForceSemverUpgradeConstraints) {
-		successors = semverSuccessors
-	}
+	ownerIDToBundleDeployment := mapOwnerIDToBundleDeployment(bundleDeployments)
 
-	result := make([]*olmvariables.InstalledPackageVariable, 0, len(bundleDeployments))
+	result := make([]*olmvariables.InstalledPackageVariable, 0, len(operators))
 	processed := sets.Set[string]{}
-	for _, bundleDeployment := range bundleDeployments {
+	for _, operator := range operators {
```
joelanford, line 34: 🎉 does this mean that with this PR, we will only be considering bundle deployments associated with operators?

m1kola: Yes, but there are still things I want to discuss about this change. I'm worried about conflicts between content installed directly via `BundleDeployment`s and content installed via `Operators`.

How do we make it safe? Maybe we should always fail resolution when there are `BundleDeployment`s which are not owned by an `Operator`? This way we can prevent OLM from breaking cluster when reconciling an  `Operator` and can propogate a clear message via unsat error. It will still be possible to mess cluster up when installing directly via `BundleDeployment` something that conflicts with existing content, but it is a more explicit way to break things and should be less of a surprise to users.

What do you think?

internal/resolution/variablesources/installed_package.go
m1kola: 🤦‍♂️ Totally forgot that we were going to go this way. Thanks for the reminder. I updated the PR (still in progress) with this appraoch.

internal/resolution/variablesources/installed_package.go
joelanford: I don't think we need to worry too much about this. An unmanaged BundleDeployment isn't all that different than an unmanaged CRD or unmanaged cluster role or whatever else that ultimately conflicts with something that comes from an Operator.

What I would expect to happen is:
- Resolution is not affected by the conflicting BundleDeployment
- If the conflicting BundleDeployment has the same name as the one that we want to create/manage, there should be an error along the lines of "couldn't create BD foo, already exists" or "couldn't update BD foo, not owned by Operator"
- The the conflict BundleDeployment does _not_ have the same name, operator-controller should create/update it, and then rukpak should report the conflict along the lines of "could not create CustomResourceDefinition `foos.acme.org`, already exists or not owned by BD", which would then bubble up on the `Installed` condition back to the `Operator`.

joelanford (approved):

internal/resolution/variablesources/installed_package.go
m1kola: > An unmanaged BundleDeployment isn't all that different than an unmanaged CRD or unmanaged cluster role or whatever else that ultimately conflicts with something that comes from an Operator

Ok, I think that is fair if we look at it from this angle. The only difference is that `BundleDeployment` and `Operator` are both coming from OLM and users might expect OLM to be aware of the bigger picture in OLM domain.

But I think we can ship it as you suggest and react based on feedback from community. I imagine adding a new variable to restrict resolution if there are `BundleDeployment` not owned by an `Operator` will be very easy to add.


> * If the conflicting BundleDeployment has the same name as the one that we want to create/manage, there should be an error along the lines of "couldn't create BD foo, already exists" or "couldn't update BD foo, not owned by Operator"
> 
> * The the conflict BundleDeployment does _not_ have the same name, operator-controller should create/update it, and then rukpak should report the conflict along the lines of "could not create CustomResourceDefinition `foos.acme.org`, already exists or not owned by BD", which would then bubble up on the `Installed` condition back to the `Operator`.

This is related to #443, right?

m1kola: > Somewhat outside the scope of this PR, but we may also now want to update our cache settings (and BD labelling, if necessary) to filter BDs by some label selector so that we're not caching _all_ BDs in the cluster, but instead just the ones we care about.

Created #521 for this

m1kola: > Somewhat outside the scope of this PR, but we may also now want to update our cache settings (and BD labelling, if necessary) to filter BDs by some label selector so that we're not caching _all_ BDs in the cluster, but instead just the ones we care about.

Created #521 for this

PR #524, opened 2023-11-08
🌱 Update installed package unit tests

m1kola:
# Description

This commit does two things.

First: It splits the `TestMakeInstalledPackageVariables` test into two tests:
* One for when `ForceSemverUpgradeConstraints` feature gate enabled (semver)
* One for when `ForceSemverUpgradeConstraints` feature gate disabled (legacy semantics)

Both tests are now table-style tests.

This is done mainly so that we can maintain test data for different semantics separately (e.g. no need to maintain channel entires for semver tests).

Second: Adds extra coverage for disabled state of the `ForceSemverUpgradeConstraints` feature gate. Previously we were not covering some cases for this state of the gate. Example of such case is when `UpgradeConstraintPolicy` field on `Operator` is set to `Ignore`.

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/resolution/variablesources/installed_package_test.go
+++ b/internal/resolution/variablesources/installed_package_test.go
@@ -4,24 +4,284 @@ import (
 	"encoding/json"
 	"testing"
 
-	"github.com/operator-framework/deppy/pkg/deppy"
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	"github.com/operator-framework/deppy/pkg/deppy/constraint"
+	"github.com/operator-framework/deppy/pkg/deppy/input"
 	"github.com/operator-framework/operator-registry/alpha/declcfg"
 	"github.com/operator-framework/operator-registry/alpha/property"
 	rukpakv1alpha1 "github.com/operator-framework/rukpak/api/v1alpha1"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/util/uuid"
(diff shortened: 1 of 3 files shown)


internal/resolution/variablesources/installed_package_test.go
```diff
@@ -4,392 +4,439 @@ import (
...
+			name:                    "UpgradeConstraintPolicy is set to Ignore",
+			upgradeConstraintPolicy: operatorsv1alpha1.UpgradeConstraintPolicyIgnore,
+			installedBundle:         bundleSet["test-package.v2.0.0"],
+			expectedResult:          []*olmvariables.InstalledPackageVariable{},
+		},
+		{
+			name:                    "no BundleDeployment for an Operator",
+			upgradeConstraintPolicy: operatorsv1alpha1.UpgradeConstraintPolicyEnforce,
+			expectedResult:          []*olmvariables.InstalledPackageVariable{},
+		},
```
awgreene, line 395: Could you explain these two tests?

m1kola: @awgreene sure. 

1. `UpgradeConstraintPolicy is set to Ignore` - When `UpgradeConstraintPolicy` is set to `Ignore` we basically saying that "there are no constraints on how to ugprade". This allows upgrades/downgrades to any version.
2. `no BundleDeployment for an Operator` - means that there is an `Operator` CR, but the requested package has not yet been installed and there is no `BundleDeployment`. In this case - we do not apply upgrade constraints as well (most likely we are reconciling the installation, not an upgrade/downgrade).

Please don't hesitate to to suggest alternative names for test cases if you think that we can communicate better the porpuse of these test cases.

awgreene (approved):
Looks good! Thanks @m1kola

PR #535, opened 2023-11-10
🐛 Resolve Deppy error check todo

m1kola:
# Description

We already have a version which contains the fix and can revert to a simpler check.

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/controllers/operator_controller.go
+++ b/internal/controllers/operator_controller.go
@@ -138,19 +138,13 @@ func (r *OperatorReconciler) reconcile(ctx context.Context, op *operatorsv1alpha
 		return ctrl.Result{}, err
 	}
 
-	// TODO: Checking for unsat is awkward using the current version of deppy.
-	//    This awkwardness has been fixed in an unreleased version of deppy.
-	//    When there is a new minor release of deppy, we can revisit this and
-	//    simplify this to a normal error check.
-	//    See https://github.com/operator-framework/deppy/issues/139.
-	unsat := deppy.NotSatisfiable{}
-	if ok := errors.As(solution.Error(), &unsat); ok && len(unsat) > 0 {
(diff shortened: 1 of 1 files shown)


internal/controllers/operator_controller.go
```diff
@@ -472,7 +466,12 @@ func operatorRequestsForCatalog(c client.Reader, logger logr.Logger) handler.Map
 // and joins them with a semicolon (rather than a comma, which the unsat.Error()
 // function does). This function also has the side effect of sorting the items
 // in the unsat slice.
-func prettyUnsatMessage(unsat deppy.NotSatisfiable) string {
+func prettyUnsatMessage(err error) string {
+	unsat := deppy.NotSatisfiable{}
+	if !errors.As(err, &unsat) {
+		panic("we must always receive deppy.NotSatisfiable here")
```
awgreene, line 472: I'm always hesitant about introducing panics in code, could you explain the benefit?

m1kola: `prettyUnsatMessage` is a workaround https://github.com/operator-framework/deppy/issues/142 (there is a comment about it above, not in the diff). It will go away and we will be doing something like this in controller code eventually:

```go
setResolvedStatusConditionFailed(&op.Status.Conditions, err.Error(), op.GetGeneration())
```

I was thinking failing laudly given that this is a temporary code and this condition should never happen unless we introduce a regression. But now I'm thinking about it again - yes, probably there is not much benefit in this and we can do just something like this: 

```go
unsat := deppy.NotSatisfiable{}
if !errors.As(err, &unsat) {
	return err.Error()
}
```

I'll update.

m1kola: Done

awgreene (approved):
Nice work!

PR #537, opened 2023-11-10
🐛 Wait for catalog deletion before recreation

### LINKED ISSUE #536: E2E flake - object is being deleted: catalogs.catalogd.operatorframework.io "test-catalog" already exists
I think this happens because we delete the test catalog here:

https://github.com/operator-framework/operator-controller/blob/f147e61c14dad352ab50d49dbf948d67a3315513/test/e2e/install_test.go#L158

And then without waiting we re-create the catalog with the same name here:

https://github.com/operator-framework/operator-controller/blob/f147e61c14dad352ab50d49dbf948d67a3315513/test/e2e/install_test.go#L173-L175

---

Example is [here](https://github.com/operator-framework/operator-controller/actions/runs/6826929016/job/18567917618):

```
  [FAILED] Unexpected error:
      <*errors.StatusError | 0xc00032fc20>: 
      object is being deleted: catalogs.catalogd.operatorframework.io "test-catalog" already exists
      {
          ErrStatus: {
              TypeMeta: {Kind: "", APIVersion: ""},
              ListMeta: {
                  SelfLink: "",
                  ResourceVersion: "",
                  Continue: "",
                  RemainingItemCount: nil,
              },
              Status: "Failure",
              Message: "object is being deleted: catalogs.catalogd.operatorframework.io \"test-catalog\" already exists",
              Reason: "AlreadyExists",
              Details: {
                  Name: "test-catalog",
                  Group: "catalogd.operatorframework.io",
                  Kind: "catalogs",
                  UID: "",
                  Causes: nil,
                  RetryAfterSeconds: 0,
              },
              Code: 409,
          },
      }
  occurred
  In [It] at: /home/runner/work/operator-controller/operator-controller/test/e2e/install_test.go:176 @ 11/10/23 15:56:52.44
```

m1kola:
# Description

Fixes #536

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/test/e2e/install_test.go
+++ b/test/e2e/install_test.go
@@ -154,8 +154,12 @@ var _ = Describe("Operator Install", func() {
 				PackageName: pkgName,
 			}
 
-			// Delete the catalog first
+			By("deleting the catalog first")
 			Expect(c.Delete(ctx, operatorCatalog)).To(Succeed())
+			Eventually(func(g Gomega) {
+				err := c.Get(ctx, types.NamespacedName{Name: operatorCatalog.Name}, &catalogd.Catalog{})
+				g.Expect(errors.IsNotFound(err)).To(BeTrue())
+			}).Should(Succeed())
 
 			By("creating the Operator resource")
 			Expect(c.Create(ctx, operator)).To(Succeed())


test/e2e/install_test.go
```diff
@@ -154,8 +154,12 @@ var _ = Describe("Operator Install", func() {
 				PackageName: pkgName,
 			}
 
-			// Delete the catalog first
+			By("deleting the catalog first")
 			Expect(c.Delete(ctx, operatorCatalog)).To(Succeed())
+			Eventually(func(g Gomega) {
+				err := c.Get(ctx, types.NamespacedName{Name: operatorCatalog.Name}, &catalogd.Catalog{})
+				g.Expect(errors.IsNotFound(err)).To(BeTrue())
+			}).Should(Succeed())
```
awgreene, line 162: We could retry on failed deletes until we get a 404, this allows us to recover from the initial delete call failure.
```suggestion
			Eventually(func(g Gomega) {
				err := c.Delete(ctx, operatorCatalog)
				g.Expect(errors.IsNotFound(err)).To(BeTrue())
			}).Should(Succeed())
```

awgreene: Bit of a nit btw.

awgreene: Could do
```
g.Expect(errors.IsNotFound(c.Delete(ctx, operatorCatalog))).To(BeTrue())
```

m1kola: I like that! We have one place related to catalog deletion. I'll update it too to be consistent.

ncdc: I don't think I'd do it this way. We expect a delete always to succeed. If you had a concern about it succeeding, you _could_ wrap the delete itself in an eventually (although I wouldn't).

This pattern of
1. Do an action that is always expected to succeed
2. Wait until result of action is eventually visible

is fairly standard for k8s e2e tests.

m1kola: That also makes sense.

@ncdc @awgreene given that it works as we want both ways and we already have a pattern of delete & wait elsewhere - I'll keep it as currently is.

PR #544, opened 2023-11-16
:book: Issue #407: Docs for version range support

### LINKED ISSUE #407: Documentation for support of version ranges
Let's document our newly added support for version ranges being specified in the Operator API's `spec.version` field.

michaelryanpeter:
# Description
Add docs explaining version range. Closes #407.

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/docs/drafts/version-ranges.md
+++ b/docs/drafts/version-ranges.md
@@ -1,0 +1,49 @@
+# Operator version ranges
+
+This document explains how to specify a version range to install or update an Operator with OLM 1.0.
+
+You define an Operator's target version in its custom resource (CR) file.
+The following list describes how OLM resolves an Operator's target version, and the resulting actions:
+
+Not specifying a version in the CR
+: Installs or updates the latest version of the Operator.
+Updates are applied automatically when they are published to the catalog.
+
+Specifying a version in the CR
+: Installs or updates the specified version.
+Updates are not applied automatically.
+If you want to update the Operator, you must manually edit the CR and apply the changes to the cluster.
+
+Specifying a channel
+: Installs or updates the latest version of the Operator in the channel.
+Updates are applied automatically when they are published to the specified channel.
+
+Specifying a version range in the CR
+: Installs or updates the latest version of the Operator within the version range.
+Updates that are within the specified range are automatically installed.
+Updates that are outside of the specified range are not installed.
+
+The `spec.version` field uses the Masterminds `semver` package to enable the version range functionality.
+
+OLM 1.0 does not support following methods for specifying a version range:
+
+* Using tags and labels are not supported.
+You must use semantic versioning to define a version range.
+* Using [hypen range comparisons](https://github.com/Masterminds/semver#hyphen-range-comparisons) are not supported.
+For example, the following range option is not supported:
+
+  ```yaml
+  version: 3.0 - 3.6
+  ```
+
+  To specify a range option, use a method similar to the following example:
+
+  ```yaml
+  version: >=3.0 <3.6
+  ```
+
+You can use the `x`, `X`, and `*` characters as wildcard characters in all comparison operations.
+
+For more information about parsing, sorting, checking, and comparing version constraints, see the [Masterminds SemVer README](https://github.com/Masterminds/semver#semver).
+
+## Example CRs that specify a version range


docs/drafts/version-ranges.md
```diff
@@ -0,0 +1,49 @@
+# Operator version ranges
+
+This document explains how to specify a version range to install or update an Operator with OLM 1.0.
+
+You define an Operator's target version in its custom resource (CR) file.
+The following list describes how OLM resolves an Operator's target version, and the resulting actions:
+
+Not specifying a version in the CR
```
michaelryanpeter, line 8: I was trying to get a definition list here. I think I messed up the formatting for that.

docs/drafts/version-ranges.md
```diff
@@ -0,0 +1,49 @@
...
+
+  ```yaml
+  version: >=3.0 <3.6
+  ```
+
+You can use the `x`, `X`, and `*` characters as wildcard characters in all comparison operations.
+
+For more information about parsing, sorting, checking, and comparing version constraints, see the [Masterminds SemVer README](https://github.com/Masterminds/semver#semver).
+
+## Example CRs that specify a version range
```
michaelryanpeter, line 49: Do we want to supply a few example CRs? Or could that wait until we flesh out the topic in the customer-facing documentation?

docs/drafts/version-ranges.md
```diff
@@ -0,0 +1,49 @@
...
+Updates that are within the specified range are automatically installed.
+Updates that are outside of the specified range are not installed.
+
+The `spec.version` field uses the Masterminds `semver` package to enable the version range functionality.
+
+OLM 1.0 does not support following methods for specifying a version range:
+
+* Using tags and labels are not supported.
+You must use semantic versioning to define a version range.
+* Using [hypen range comparisons](https://github.com/Masterminds/semver#hyphen-range-comparisons) are not supported.
```
joelanford, line 32: Not sure how others feel, but I think we should try to avoid referencing masterminds directly. We have defined a version range pattern that we validate in the API, so I think we should just document that. Today, it happens to be a subset of the valid masterminds patterns.

Ultimately, _that_ is what we'll have to support, even if masterminds makes other changes.

docs/drafts/version-ranges.md
```diff
@@ -0,0 +1,49 @@
...
+Updates are applied automatically when they are published to the catalog.
+
+Specifying a version in the CR
+: Installs or updates the specified version.
+Updates are not applied automatically.
+If you want to update the Operator, you must manually edit the CR and apply the changes to the cluster.
+
+Specifying a channel
+: Installs or updates the latest version of the Operator in the channel.
+Updates are applied automatically when they are published to the specified channel.
```
joelanford, line 19: I would expect this doc to be purely about `spec.version`, so I'd suggest striking this.

In a higher-level doc, I'd expect something that talks about the various `spec` fields and their interactions.

docs/drafts/version-ranges.md
```diff
@@ -0,0 +1,49 @@
+# Operator version ranges
+
+This document explains how to specify a version range to install or update an Operator with OLM 1.0.
+
+You define an Operator's target version in its custom resource (CR) file.
+The following list describes how OLM resolves an Operator's target version, and the resulting actions:
+
+Not specifying a version in the CR
+: Installs or updates the latest version of the Operator.
```
joelanford, line 9: > latest

Technically, it's the latest version _that satisfies resolution_. I think it's important to call that out pretty clearly somewhere, and possibly explain with an example.

For example:

In the catalog:
- packageA v1.0.0 exists and depends on packageB v1.2.0
- packageB v1.2.0 exists
- packageB v1.2.1 exists

On the cluster:
- Operator for `packageA` says `spec.version: 1.0.0`
- Operator for `packageB` says `spec.version: 1.2.x`

The version that will ultimately be chosen for `packageB` is 1.2.0 even though 1.2.1 is available, because that's what `packageA` requires.

docs/drafts/version-ranges.md
tmshort: So, this really doesn't look like a "list", more a a set of short paragraphs.

docs/drafts/version-ranges.md
tmshort: If we are not going to reference Masterminds directly, we should put in the supported operations in here.

docs/drafts/version-ranges.md
michaelryanpeter: Agreed. The formatting didn't turn out the way I wanted it to. I will take another approach.

docs/drafts/version-ranges.md
michaelryanpeter: For the moment, I linked to Deppy's README to explain resolution. WDYT of creating a follow up issue to port the OCP docs about dependency resolution upstream?

docs/drafts/version-ranges.md
michaelryanpeter: Removed.

docs/drafts/version-ranges.md
```diff
@@ -0,0 +1,104 @@
...
+
+| Operator | Definition                        |
+|----------|-----------------------------------|
+| `=`      | equal (not aliased to an operator |
+| `!=`     | not equal                         |
+| `>`      | greater than                      |
+| `<`      | less than                         |
+| `>=`     | greater than or equal to          |
+| `<=`     | less than or equal to             |
+
```
michaelryanpeter, line 41: I created https://github.com/operator-framework/operator-controller/issues/550 to further explain[ working with pre-releases](https://github.com/Masterminds/semver/blob/2f39fdc11c33c38e8b8b15b1f04334ba84e751f2/README.md#working-with-prerelease-versions)

docs/drafts/version-ranges.md
```diff
@@ -0,0 +1,93 @@
+# Operator version ranges
+
+This document explains how to specify a version range to install or update an Operator with OLM 1.0.
+
+You define a version range in an Operator's custom resource (CR) file.
+
+## Specifying a version range in the CR
```
michaelryanpeter, line 7: I removed the other sections on the `spec.version` field. They didn't make sense with the expanded content, and I think it will take a bit more work to incorporate the content into one doc on the `spec.version` field. I created a follow up issue here: https://github.com/operator-framework/operator-controller/issues/551

docs/drafts/version-ranges.md
```diff
@@ -0,0 +1,93 @@
+# Operator version ranges
```
michaelryanpeter, line 1: I created a follow up issue to add example CRs to the doc: https://github.com/operator-framework/operator-controller/issues/552

docs/drafts/version-ranges.md
```diff
@@ -0,0 +1,93 @@
...
+
+### Comparisons
+
+You define a version range by adding a comparison string to the `spec.version` field. A comparison string is composed of a list of comma or space separated values and one or more comparison operators. You can add an additional comparison string by including an OR (`||`) operator between the strings.
+
+#### Basic comparisons
+
+| Operator | Definition                        |
+|----------|-----------------------------------|
+| `=`      | equal (not aliased to an operator |
```
everettraven, line 24: Missing a closing parenthesis
```suggestion
| `=`      | equal (not aliased to an operator) |
```

everettraven (commented):
This doc looks great to me! I noticed one small thing:

docs/drafts/version-ranges.md
```diff
@@ -0,0 +1,93 @@
...
+| `!=`     | not equal                         |
+| `>`      | greater than                      |
+| `<`      | less than                         |
+| `>=`     | greater than or equal to          |
+| `<=`     | less than or equal to             |
+
+#### Range comparisons
+
+OLM 1.0 does not support hypen range comparisons.
```
tmshort, line 33:
```suggestion
OLM 1.0 does not support hyphen range comparisons.
```

michaelryanpeter: Addressed in https://github.com/operator-framework/operator-controller/commit/164bb28f70011347ebf4483700903a50adbcfc88

docs/drafts/version-ranges.md
michaelryanpeter: Addressed in https://github.com/operator-framework/operator-controller/commit/164bb28f70011347ebf4483700903a50adbcfc88

everettraven (approved):

Issue #322, opened 2023-08-07
Create a package for scheme setup


m1kola:
We already have multiple places where we setup scheme:
* https://github.com/operator-framework/operator-controller/blob/fa84c733ade0f71f33085a978ae0163f570d48b7/cmd/manager/main.go#L49-L53
* https://github.com/operator-framework/operator-controller/blob/fa84c733ade0f71f33085a978ae0163f570d48b7/cmd/resolutioncli/main.go#L62-L65
* https://github.com/operator-framework/operator-controller/blob/fa84c733ade0f71f33085a978ae0163f570d48b7/internal/controllers/suite_test.go#L70-L74
* Various tests

And more will be introduced (e.g #286 adds a new one, see `test/operator-framework-e2e/operator_framework_test.go`).

To avoid having this repeated setup and inconsistent schemas we should consider creating a scheme package in the project which will be setting up a scheme. In rest of the places we will just be able to import already constructed scheme.

Example:

* `kubectl` scheme - https://github.com/kubernetes/kubectl/blob/197123726db24c61aa0f78d1f0ba6e91a2ec2f35/pkg/scheme/install.go

grokspawn:
solved by #770

PR #545, opened 2023-11-16
:seedling: Remove ginkgo from e2e tests

### LINKED ISSUE #196: Update e2e package tests

tmshort:
Fixes #196

# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -105,16 +105,16 @@ test: manifests generate fmt vet test-unit test-e2e #HELP Run all tests.
 .PHONY: e2e
 FOCUS := $(if $(TEST),-v -focus "$(TEST)")
 E2E_FLAGS ?= ""
-e2e: $(GINKGO) #EXHELP Run the e2e tests.
-	$(GINKGO) --tags $(GO_BUILD_TAGS) $(E2E_FLAGS) -trace -progress $(FOCUS) test/e2e
+e2e: $(SETUP_ENVTEST) #EXHELP Run the e2e tests.
+	eval $$($(SETUP_ENVTEST) use -p env $(ENVTEST_VERSION)) && go test -tags $(GO_BUILD_TAGS) -v ./test/e2e/...
 
 export REG_PKG_NAME=registry-operator
 export PLAIN_PKG_NAME=plain-operator
 export CATALOG_IMG=${E2E_REGISTRY_NAME}.${E2E_REGISTRY_NAMESPACE}.svc:5000/test-catalog:e2e
 .PHONY: test-op-dev-e2e
-test-op-dev-e2e: $(GINKGO) $(OPERATOR_SDK) $(KUSTOMIZE) $(KIND) #HELP Run operator create, upgrade and delete tests.
+test-op-dev-e2e: $(SETUP_ENVTEST) $(OPERATOR_SDK) $(KUSTOMIZE) $(KIND) #HELP Run operator create, upgrade and delete tests.
 	test/operator-framework-e2e/setup.sh $(OPERATOR_SDK) $(CONTAINER_RUNTIME) $(KUSTOMIZE) $(KIND) $(KIND_CLUSTER_NAME) ${E2E_REGISTRY_NAMESPACE}
-	$(GINKGO) --tags $(GO_BUILD_TAGS) -trace -progress $(FOCUS) test/operator-framework-e2e
+	eval $$($(SETUP_ENVTEST) use -p env $(ENVTEST_VERSION)) && go test -tags $(GO_BUILD_TAGS) -v ./test/operator-framework-e2e/...
 
 .PHONY: test-unit
 ENVTEST_VERSION = $(shell go list -m k8s.io/client-go | cut -d" " -f2 | sed 's/^v0\.\([[:digit:]]\{1,\}\)\.[[:digit:]]\{1,\}$$/1.\1.x/')
--- a/test/e2e/install_test.go
+++ b/test/e2e/install_test.go
@@ -7,12 +7,13 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"testing"
 	"time"
 
-	. "github.com/onsi/ginkgo/v2"
-	. "github.com/onsi/gomega"
 	catalogd "github.com/operator-framework/catalogd/api/core/v1alpha1"
 	rukpakv1alpha1 "github.com/operator-framework/rukpak/api/v1alpha1"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"gopkg.in/yaml.v2"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -23,6 +24,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/rand"
 	kubeclient "k8s.io/client-go/kubernetes"
 	"k8s.io/utils/env"
+
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	operatorv1alpha1 "github.com/operator-framework/operator-controller/api/v1alpha1"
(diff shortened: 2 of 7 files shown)


Makefile
```diff
@@ -105,16 +105,16 @@ test: manifests generate fmt vet test-unit test-e2e #HELP Run all tests.
 .PHONY: e2e
 FOCUS := $(if $(TEST),-v -focus "$(TEST)")
 E2E_FLAGS ?= ""
-e2e: $(GINKGO) #EXHELP Run the e2e tests.
-	$(GINKGO) --tags $(GO_BUILD_TAGS) $(E2E_FLAGS) -trace -progress $(FOCUS) test/e2e
+e2e: $(SETUP_ENVTEST) #EXHELP Run the e2e tests.
```
stevekuznetsov, line 108: Thank you for making this self-contained, nobody has to remember how to use setup-envtest anymore!

Makefile
```diff
@@ -105,16 +105,16 @@ test: manifests generate fmt vet test-unit test-e2e #HELP Run all tests.
...
+	eval $$($(SETUP_ENVTEST) use -p env $(ENVTEST_VERSION)) && go test -tags $(GO_BUILD_TAGS) -v $(shell go list ./test/e2e)
```
stevekuznetsov, line 109: Any specific reason to use `$(shell go list ./test/e2e)` instead of `./test/e2e/...`?

test/operator-framework-e2e/operator_framework_test.go
```diff
@@ -6,125 +6,86 @@ import (
...
-	c   client.Client
-	ctx context.Context
-)
-
 func TestOperatorFramework(t *testing.T) {
-	RegisterFailHandler(Fail)
-	RunSpecs(t, "Operator Framework E2E Suite")
-}
+	t.Parallel()
+	cfg := ctrl.GetConfigOrDie()
```
stevekuznetsov, line 26: Isn't the `ctrl.GetConfigOrDie()` helper for in-cluster config when running in a Pod? Why is it in use here?

test/operator-framework-e2e/operator_framework_test.go
```diff
@@ -6,125 +6,86 @@ import (
...
 
-var _ = BeforeSuite(func() {
-	cfg = ctrl.GetConfigOrDie()
 	scheme := runtime.NewScheme()
-	ctx = context.Background()
 
-	err := catalogd.AddToScheme(scheme)
-	Expect(err).ToNot(HaveOccurred())
+	utilruntime.Must(catalogd.AddToScheme(scheme))
```
stevekuznetsov, line 30: If you have a `t *testing.T` present, you might as well use `t.Fatal()` - but, why are we creating so many Schemes? There could just as well be a package somewhere in this repo that exposes a global.

test/operator-framework-e2e/operator_framework_test.go
```diff
@@ -6,125 +6,86 @@ import (
...
-
-var _ = Describe("Operator Framework E2E", func() {
-	var catalog *catalogd.Catalog
-	BeforeEach(func() {
-		catalog = &catalogd.Catalog{
+	var operators = []*operatorv1alpha1.Operator{
+		{
 			ObjectMeta: metav1.ObjectMeta{
-				Name: "catalog" + rand.String(10),
+				Name: "plainv0",
```
stevekuznetsov, line 39: Why not prefer to use GenerateName or random names for parallel runs?

stevekuznetsov (commented):
Looks good! Just a couple questions

test/operator-framework-e2e/operator_framework_test.go
m1kola: We have #322. I think this is what you are suggesting here.

test/operator-framework-e2e/operator_framework_test.go
everettraven: I believe it will also build the configuration from the KUBECONFIG when run outside of a cluster

test/e2e/install_test.go
```diff
@@ -32,237 +34,297 @@ const (
...
+	operator := &operatorv1alpha1.Operator{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: operatorName,
+		},
+	}
+	return operator, operatorName, operatorCatalog
+}
+
+func testCleanup(t *testing.T, cat *catalogd.Catalog, operator *operatorv1alpha1.Operator) {
+	// getArtifactsOutout?
```
everettraven, line 55: Seems reasonable to place it here if all the tests end up calling this function. IIRC getting the artifacts should only happen for tests that have failed.

everettraven: I see further down it is just called directly. Would remove this comment if that is the route being taken

test/operator-framework-e2e/operator_framework_test.go
everettraven: Looks like these names are used in a testing log message to signal we are testing for specific bundle types

everettraven: Although, I do wonder if it would be more appropriate to create a different testing struct (similar to what we do for table tests) to have the test name that is running be more descriptive. Maybe something along the lines of:
```go
{
  name: "Operator references a plain+v0 bundle"
  operator: &operatorv1alpha1.Operator{...}
}
```

test/operator-framework-e2e/operator_framework_test.go
tmshort: Yes, this has a `t` so, I can use the that. But since there's an issue for schemes already, we can use that to update the schemes.

test/operator-framework-e2e/operator_framework_test.go
tmshort: These tests are already run in parallel.
Really no need to use random names when the names are already different, because as @everettraven points out, the name is used as part of the log output.

_IF_ this test were more complicated, I'd go with what @everettraven is suggesting, but there's only two test cases. If this were to increase to 5 or more, then, yes, a more comprehensive table test structure would be appropriate.

test/e2e/install_test.go
tmshort: I think the original test was a bit weird, in that there were multiple places that `getArtifactOutput` was called, and I cleaned it up to be run after every test, rather than at the end.

test/e2e/install_test.go
```diff
@@ -272,90 +333,95 @@ var _ = Describe("Operator Install", func() {
...
 	// Create the full artifact path
 	err = os.MkdirAll(artifactPath, 0755)
-	Expect(err).To(Not(HaveOccurred()))
+	require.NoError(t, err)
 
 	// Get all namespaces
 	namespaces := corev1.NamespaceList{}
-	if err := c.List(ctx, &namespaces); err != nil {
-		GinkgoWriter.Printf("Failed to list namespaces %w", err)
+	if err := c.List(context.Background(), &namespaces); err != nil {
```
everettraven, line 356: Would it be worth creating a single context to reuse in all these calls instead of using `context.Background()` in every call?

stevekuznetsov: Not sure it matters much since there will never be a context in test code

stevekuznetsov (approved):

test/operator-framework-e2e/operator_framework_test.go
stevekuznetsov: As long as you can run it locally that's fine

PR #555, opened 2023-11-28
⚠ bump rukpak to v0.16.0

joelanford:
# Description
Bumping rukpak dependency to v0.16.0. This includes changes to the way that rukpak names `Bundle` objects derived from `BundleDeployments` and the way that performs registry+v1 to plain conversion. This results in different names and labels for some objects, so it may be a breaking change for some users.

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,7 +10,7 @@ require (
 	github.com/operator-framework/catalogd v0.10.0
 	github.com/operator-framework/deppy v0.1.0
 	github.com/operator-framework/operator-registry v1.32.0
-	github.com/operator-framework/rukpak v0.15.0
+	github.com/operator-framework/rukpak v0.16.0
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.8.4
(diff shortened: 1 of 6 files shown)


m1kola:
> This results in different names and labels for some objects, so it may be a breaking change for some users.

Trying to understand this a bit better. Is bumping rukpak in operator-controller causing breaking changes? Or deploying a new version of rukpak into the cluster? I suspect it is the latter.

joelanford:
There are no breaking changes to the Operator API in this PR.

However, if there are existing Operator objects in a cluster with existing BundleDeployment objects, the names and labels of Bundles, Roles, RoleBindings, ClusterRoles, and ClusterRoleBindings will change, which will cause the rukpak provisioner recreate those objects with their new names/labels to account for those changes.

If anyone is depending on the old names or labels (they really shouldn't be since they include a hash value that changes as the BD template changes), they'll be broken.

m1kola (approved):

PR #559, opened 2023-11-29
📖 Draft documentation for initial upgrade support

### LINKED ISSUE #470: Documentation for semver upgrade edge constraints support
Need to create user-facing documentation for #377 which is based on [this RFC](https://docs.google.com/document/d/1dfiNr0kvP1nfS0f9xaGiLyfrJ_KcD38clYs4LYBKllk/edit).

---

What we should include:

* Cover `ForceSemverUpgradeConstraints` feature gate and mention default state (enabled)
* Explain how semver upgrades work
    * Extensively cover behaviour of major version zero (see semver spec) which behaves differently fromv versions from >=1.0.0.
    * Explain that setting `.spec.upgradeConstraintPolicy` to `Ignore` is a way for upgrading versions in major version zero.
* Explain legacy OLMv0 semantics or link to the existing docs
    * Might be worth mentioning that at this time we only support `Replaces` and do not support `Skips` and `SkipRange`
* Explain what `.spec.upgradeConstraintPolicy` on `Operator` is for
    * Might be worth adding a warning that this need to be used carefully and independent verification for upgrades is required
    * Potentially mention ability to downgrade packages

m1kola:
# Description

Draft document to cover initial upgrade support.

Closes #470

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/docs/drafts/upgrade-support.md
+++ b/docs/drafts/upgrade-support.md
@@ -1,0 +1,66 @@
+# Upgrade support
+
+This document explains how OLM 1.0 handles upgrades.
+
+OLM 1.0 introduces a simplified UX for package authors and package admins to implicitly define upgrade edges via [Semantic Versioning](https://semver.org/).
+
+It also introduces an API to enable independently verified upgrades and downgrades.
+
+## Upgrade constraint semantics
+
+As of Operator Controller release 0.8.0, OLM 1.0 supports the following upgrade constraint semantics:
+
+* [Semantic Versioning](https://semver.org/) (Semver)
+* The `replaces` directive from the [legacy OLM 0 semantics](https://olm.operatorframework.io/docs/concepts/olm-architecture/operator-catalog/creating-an-update-graph/#methods-for-specifying-updates)
+
+The Kubernetes manifests in this repo enable Semver support by default. Cluster admins can control which semantics to use by passing one of the following arguments to the `manager` binary:
+* `--feature-gates=ForceSemverUpgradeConstraints=true` - enable Semver
+* `--feature-gates=ForceSemverUpgradeConstraints=false` - disable Semver, use legacy semantics
+
+For example, to enable Semver update the `controller-manager` Deployment manifest to include the following argument:
+
+```yaml
+- command:
+  - /manager
+  args:
+  - --feature-gates=ForceSemverUpgradeConstraints=true
+  image: controller:latest
+```
+
+In a future release, it is planned to remove the `ForceSemverUpgradeConstraints` feature gate and allow package authors to specify upgrade constraint semantics at the catalog level.
+
+## Upgrades
+
+OLM supports Semver to provide a simplified way for package authors to define compatible upgrades. According to the Semver standard, releases within a major version (e.g. `>=1.0.0 <2.0.0`) must be compatible. As a result, package authors can publish a new package version following the Semver specification, and OLM assumes compatibility. Package authors do not have to explicitly define upgrade edges in the catalog.
+
+**Note:** Currently, OLM 1.0 does not support automatic upgrades to the next major version. You must manually verify and perform major version upgrades. For more information about major version upgrades, see [Manually verified upgrades and downgrades](#manually-verified-upgrades-and-downgrades).
+
+### Upgrades within the major version zero
+
+According to the Semver specification, a major version zero release is for initial development. It is assumed that breaking changes might be introduced at any time. As a result, the following special conditions apply to upgrades within a major version zero release:
+
+* You cannot automatically upgrade from one patch version to another when both major and minor versions are `0`. For example, automatic upgrades within the following version range are not allowed: `>= 0.0.1 <0.1.0`.
+* You cannot automatically upgrade from one minor version to another minor version within the major version zero. For example, no upgrades from `0.1.0` to `0.2.0`. However, you can upgrade from patch versions. For example, upgrades are possible in ranges `>= 0.1.0 <0.2.0`, `>= 0.2.0 <0.3.0`, `>= 0.3.0 <0.4.0`, and so on.
+
+You must verify and perform upgrades manually in cases where automatic upgrades are blocked.
+
+## Manually verified upgrades and downgrades
+
+**Warning:** If you want to force an upgrade manually, you must thoroughly verify the outcome before applying any changes to production workloads. Failure to test and verify the upgrade might lead to catastrophic consequences such as data loss.
+
+As a package admin, if you must upgrade or downgrade to version that might be incompatible with the currently installed version, you can set the `.spec.upgradeConstraintPolicy` field to `Ignore` on the relevant `Operator` resource.
+
+If you set the field to `Ignore`, no upgrade constraints are set on the package. As a result, you can change the version to any version available in the catalogs for a given package.
+
+Example `Operator` with `.spec.upgradeConstraintPolicy` field set to `Ignore`:
+
+```yaml
+apiVersion: operators.operatorframework.io/v1alpha1
+kind: Operator
+metadata:
+  name: operator-sample
+spec:
+  packageName: argocd-operator
+  version: 0.6.0
+  upgradeConstraintPolicy: Ignore
+```


docs/drafts/initial-upgrade-support.md
```diff
@@ -0,0 +1,43 @@
...
+As per Semver major version zero is for initial development and anything may change at any time. Given that there are some special case for upgrades within the major version zero:
+
+* It is not possible to automatically upgrade one patch versions to another when both major and minor versions are `0`. Meaning no automatic upgrades withing range `>= 0.0.1 <0.1.0`.
+* It is not possible to automatically upgrade from one minor version to another minor version within the major version zero. E.g. no upgrades from `0.1.0` to `0.2.0`. But one can upgrade from patch to patch version. E.g. upgrades are possible in ranges `>= 0.1.0 <0.2.0`, `>= 0.2.0 <0.3.0`, `>= 0.3.0 <0.4.0` and so on.
+
+To perform an upgrade in the cases mentioned above - the upgrade must be independently verified.
+
+## Independently verified upgrades and downgrades
+
+**Warning:** This option requires thorough manual verification of the desired operation prior to applying to production workloads. Depending on the package, using this option might lead to catastrophic consequences such as data loss.
```
michaelryanpeter, line 39:
```suggestion
**Warning:** If you want to force an upgrade manually, you must thoroughly verify the outcome before applying any changes to production workloads. Failure to test and verify the upgrade might lead to catastrophic consequences such as data loss.
```

docs/drafts/initial-upgrade-support.md
```diff
@@ -0,0 +1,43 @@
...
+
+In cases when package admins need to upgrade or downgrade to a version which is not guaranteed to be compatible they can choose to set `.spec.upgradeConstraintPolicy` to `Ignore` on the relevant `Operator` resource.
```
michaelryanpeter, line 41:
```suggestion
As a package admin, if you must upgrade or downgrade to version that might be incompatible with the currently installed version, you can set the `.spec.upgradeConstraintPolicy` field to `Ignore` on the relevant `Operator` resource.
```

docs/drafts/initial-upgrade-support.md
```diff
@@ -0,0 +1,43 @@
...
+
+This option means that no upgrade constraints will be set on the package and version can be changed to any version available in catalogs for a given package.
```
michaelryanpeter, line 43:
```suggestion
If you set the field to `Ignore`, no upgrade constraints are set on the package. As a result, you can change the version to any version available in the catalogs for a given package.
```

docs/drafts/initial-upgrade-support.md
```diff
@@ -0,0 +1,43 @@
...
+
+* [Semantic Versioning](https://semver.org/) (Semver)
+* [Legacy OLM 0 semantics](https://olm.operatorframework.io/docs/concepts/olm-architecture/operator-catalog/creating-an-update-graph/#methods-for-specifying-updates)
+
+**Note:** At the moment of writing legacy semantics support in OLM 1.0 is incomplete: only `replaces` directive is taken into account.
+
+Kubernetes manifests provided in this repo enable Semver support by default, but cluster admins can control which semantics to use by passing one of the arguments into `manager` binary:
+* `--feature-gates=ForceSemverUpgradeConstraints=true` - enable Semver
+* `--feature-gates=ForceSemverUpgradeConstraints=false` - disable Semver, use legacy semantics
+
```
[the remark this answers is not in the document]
m1kola: Added an example calls. Or should we include example manifests?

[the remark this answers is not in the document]
michaelryanpeter: I think including example manifests would be helpful. Because you can set the argument using either the CLI or by editing the manifest/yaml file?

And sorry for not mentioning it yesterday, but I am not familiar with the `manager` binary. Is that something new to the operator-controller or is it a third party tool? Do we need to introduce how to download the binary or include it as a prerequisite?

docs/drafts/upgrade-support.md
```diff
@@ -0,0 +1,54 @@
...
+
+This document explains how OLM 1.0 handles upgrades.
+
+OLM 1.0 introduces a simplified UX for package authors and package admins to implicitly define upgrade edges via [Semantic Versioning](https://semver.org/).
+
+It also introduces an API to enable independently verified upgrades and downgrades.
+
+## Upgrade constraint semantics
+
+As of Operator Controller version 0.x, OLM 1.0 supports the following upgrade constraint semantics:
```
michaelryanpeter, line 11: Just to clarify this comment yesterday, I was sure what version of the Operator Controller introduced this feature, so we could update it to something like: "Operator Controller version 0.11.0" or something.

docs/drafts/upgrade-support.md
```diff
@@ -0,0 +1,54 @@
...
+
+You must verify and perform upgrades manually in cases where automatic upgrades are blocked.
+
+## Manually verified upgrades and downgrades
+
+**Warning:** If you want to force an upgrade manually, you must thoroughly verify the outcome before applying any changes to production workloads. Failure to test and verify the upgrade might lead to catastrophic consequences such as data loss.
+
+As a package admin, if you must upgrade or downgrade to version that might be incompatible with the currently installed version, you can set the `.spec.upgradeConstraintPolicy` field to `Ignore` on the relevant `Operator` resource.
+
+If you set the field to `Ignore`, no upgrade constraints are set on the package. As a result, you can change the version to any version available in the catalogs for a given package.
```
michaelryanpeter, line 54: WDYT of adding an example manifest/CR at the end to show how to enable manual upgrades?

michaelryanpeter (commented):
Just a couple of small follow up questions and ideas that didn't occur to me on the first review. 

This is looking really good :rocket:

docs/drafts/initial-upgrade-support.md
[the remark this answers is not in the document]
m1kola: @michaelryanpeter the `manager` binary is operator-controller itself. This repo contains this manifest:

https://github.com/operator-framework/operator-controller/blob/6cb7029801558cd9e69fd4285468e600f45df7a6/config/manager/manager.yaml#L68-L72

But we later customise args like so:

https://github.com/operator-framework/operator-controller/blob/6cb7029801558cd9e69fd4285468e600f45df7a6/config/default/manager_auth_proxy_patch.yaml#L48-L53

Do you think we should have a full `Deployment` example for both states of the flag? Maybe under a spoiler? Or should we include only a part of the deployment manifest? E.g.

```yaml
- command:
  - /manager
  args:
  - --feature-gates=ForceSemverUpgradeConstraints=true
  image: controller:latest
```

docs/drafts/upgrade-support.md
m1kola: Ahh, missted that. I'll update that

docs/drafts/upgrade-support.md
m1kola: Good idea, I think it will be helpful.

docs/drafts/initial-upgrade-support.md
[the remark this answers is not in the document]
michaelryanpeter: I think it would be work explaining. I read the instructions like I needed to enter a CLI command to enable the command, something like `kubectl manager --feature-gates=<etc>`.

I think that at this stage of development, just an excerpt like your last example would work. Then when we stitch the version, version range, channel, and upgrade docs together, we can maybe add another formulary of real world example manifests.

docs/drafts/upgrade-support.md
michaelryanpeter: It is tricky, but I think you came up with an elegant solution.

docs/drafts/upgrade-support.md
```diff
@@ -0,0 +1,67 @@
...
+It also introduces an API to enable independently verified upgrades and downgrades.
+
+## Upgrade constraint semantics
+
+As of Operator Controller release 0.8.0, OLM 1.0 supports the following upgrade constraint semantics:
+
+* [Semantic Versioning](https://semver.org/) (Semver)
+* The `replaces` directive from the [legacy OLM 0 semantics](https://olm.operatorframework.io/docs/concepts/olm-architecture/operator-catalog/creating-an-update-graph/#methods-for-specifying-updates)
+
+The Kubernetes manifests in this repo enable Semver support by default. Cluster admins can control which semantics to use by passing one of the following arguments to the `manager` binary:
```
michaelryanpeter, line 16: This might be out of scope for this issue, but it might be worth explaining in a follow up issue why a cluster would want to deviate from the upgrade constraint semantics specified by the package author.

docs/drafts/upgrade-support.md
```diff
@@ -0,0 +1,67 @@
...
+
+```bash
+manager --feature-gates=ForceSemverUpgradeConstraints=true
+```
+
+And to disable it:
+
+```bash
+manager --feature-gates=ForceSemverUpgradeConstraints=false
+```
```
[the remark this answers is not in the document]
m1kola: @michaelryanpeter I think it is probably better to keep it somewhere in the beginning beucase this section gives an overview of both semantics which are currently supported. After this section we go into more details.

Also we plan to remove the feature gate in favour of package authors controlling which semantics to use on the catalog level eventually ([this RFC](https://docs.google.com/document/d/1dfiNr0kvP1nfS0f9xaGiLyfrJ_KcD38clYs4LYBKllk/edit) has some details about this). This is a temporary solution and cluster admins will not be controlling it at some point.

docs/drafts/upgrade-support.md
m1kola: I added the following:

> We plan to remove the `ForceSemverUpgradeConstraints` feature gate and enable package authors to specify which upgrade constraint semantics they wish to use on the catalog level.

docs/drafts/upgrade-support.md
```diff
@@ -0,0 +1,66 @@
...
+
+```yaml
+- command:
+  - /manager
+  args:
+  - --feature-gates=ForceSemverUpgradeConstraints=true
+  image: controller:latest
+```
+
+We plan to remove the `ForceSemverUpgradeConstraints` feature gate and enable package authors to specify which upgrade constraint semantics they wish to use on the catalog level.
```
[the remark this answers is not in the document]
m1kola: Makes sense. Updated.

michaelryanpeter (approved):
LGTM

docs/drafts/upgrade-support.md
```diff
@@ -0,0 +1,66 @@
...
+  args:
+  - --feature-gates=ForceSemverUpgradeConstraints=true
+  image: controller:latest
+```
+
+In a future release, it is planned to remove the `ForceSemverUpgradeConstraints` feature gate and allow package authors to specify upgrade constraint semantics at the catalog level.
+
+## Upgrades
+
+OLM supports Semver to provide a simplified way for package authors to define compatible upgrades. According to the Semver standard, releases within a major version (e.g. `>=1.0.0 <2.0.0`) must be compatible. As a result, package authors can publish a new package version following the Semver specification, and OLM assumes compatibility. Package authors do not have to explicitly define upgrade edges in the catalog.
```
tmshort, line 34: Nit: There are various levels of Semver compatibility, as specified by all three component versions. There's no mention of patch compatibility here. But, we don't want to get ourselves too deep in the weeks with explanations here, as we reference Semver above.

m1kola: @tmshort  Here we say that anything witing the major version is expected to be compatible. E.g. 1.0.3 is compatible with 1.3.0. So pacakge authors can release accordingly. I think it is not necessary to go into the details of semver spec itself here (e.g. patch versions are for bug fixes, minor versions are fore backward compatible features).

But I'm happy to add more info if you think we should stress some specific aspect of semver here. Let me know.

docs/drafts/upgrade-support.md
```diff
@@ -0,0 +1,66 @@
...
+
+**Note:** Currently, OLM 1.0 does not support automatic upgrades to the next major version. You must manually verify and perform major version upgrades. For more information about major version upgrades, see [Manually verified upgrades and downgrades](#manually-verified-upgrades-and-downgrades).
```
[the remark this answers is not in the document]
m1kola: PR: https://github.com/operator-framework/operator-controller/pull/562

PR #560, opened 2023-12-01
⚠️ Upgrade Deppy

m1kola:
# Description

New version contains breaking changes: Deppy no longer has variable source API. Instead it accepts a slice of variables as input.

## Reviewer Checklist

- [ ] API Go Documentation


everettraven (approved):
Overall looks good to me. One nit about potentially confusing naming of a field but I don't think that should block this PR

internal/controllers/operator_controller.go
```diff
@@ -44,11 +45,18 @@ import (
...
+type BundleProvider interface {
+	Bundles(ctx context.Context) ([]*catalogmetadata.Bundle, error)
+}
+
 // OperatorReconciler reconciles a Operator object
 type OperatorReconciler struct {
 	client.Client
-	Scheme   *runtime.Scheme
-	Resolver *solver.DeppySolver
+	CatalogClient BundleProvider
```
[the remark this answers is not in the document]
m1kola: I renamed the field to be `BundleProvider` to match `BundleProvider` interface. But I kept catalog client names in the client itself (at least for now). The client already fetches all the data from catalogs - It is just that we only provide a method to get bundles. This might change at some point.

internal/controllers/operator_controller.go
```diff
@@ -124,8 +132,19 @@ func (r *OperatorReconciler) reconcile(ctx context.Context, op *operatorsv1alpha
 		setResolvedStatusConditionUnknown(&op.Status.Conditions, "validation has not been attempted as spec is invalid", op.GetGeneration())
 		return ctrl.Result{}, nil
 	}
+
+	// gather variables for resolution
+	variables, err := r.variables(ctx)
```
tmshort, line 137: Ugh... multiple things named `variables`.

tmshort (changes requested):
The use of the name `variables` in multiple places might be a bit confusing... I'd suggest using `vars` in places where there's a function/method named `variables`. Or possibly renaming the `variables()` method. (Yes, golang is able to figure it out, but can we?)

internal/controllers/operator_controller.go
everettraven: Does `r.variables` need to be a function? Could we just do everything in that function here?

m1kola:
> Re ⚠️ - while deppy had breaking changes, users won't experience any breaking changes, will they?

Good point. I think for operator-controller users everythign should stay the same. Replaced with 🌱

internal/controllers/operator_controller.go
```diff
@@ -194,6 +213,27 @@ func (r *OperatorReconciler) reconcile(ctx context.Context, op *operatorsv1alpha
...
+	bundleDeploymentList := rukpakv1alpha1.BundleDeploymentList{}
+	if err := r.Client.List(ctx, &bundleDeploymentList); err != nil {
+		return nil, err
+	}
+	variables, err := GenerateVariables(allBundles, operatorList.Items, bundleDeploymentList.Items)
+	if err != nil {
+		return nil, err
+	}
+
+	return variables, nil
```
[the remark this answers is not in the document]
m1kola: Can rename it into `result` or `out` or `vars`.

internal/controllers/operator_controller.go
m1kola: I want to keep the function. If we do everythign here - I think readability will be decresed significantly (more stuff in `reconcile`) and we will have to handle each `if err != nil` branch separately (set conditions, etc).

I'm happy to rename `variables` into `vars` as Todd suggested. Will update the PR shortly.

tmshort (approved):
/lgtm

m1kola:
Quay failed with 502. Will try to merge again.

PR #574, opened 2024-01-08
:sparkles: (feat): add graph deprecation logic

### LINKED ISSUE #525: (Placeholder): Add deprecation information to bundles/bundle variables
>[!NOTE]
>This issue is a placeholder issue and is subject to change

As part of implementing [RFC: OLMv1 Graph Deprecation](https://docs.google.com/document/d/1CYGANKILfzfOoWtHvyg4luF7i-cC1NRa7_1N9xTLatw/edit?usp=sharing) deprecation information needs to be added to the bundles used to build bundle variables for resolution.

This will likely include updates to the following areas of code:
- Updating the types used to process FBC data: https://github.com/operator-framework/operator-controller/blob/main/internal/catalogmetadata/types.go
- Updating the logic to fetch catalog data from the catalogd http server to include parsing deprecation data: https://github.com/operator-framework/operator-controller/blob/main/internal/catalogmetadata/client/client.go
- Updating bundle variables to include deprecation information (I'm unsure where or what changes are necessary to make this happen)

Bundles should be considered deprecated if any of the following conditions are satisfied:
- The parent `olm.package` is deprecated
- The `olm.channel` being used for resolution is deprecated and contains an entry for the `olm.bundle` object
    - Note: This only applies if resolution has been restricted to use the deprecated channel OR the bundle only exists in that channel
- The `olm.bundle` object itself is deprecated

### LINKED ISSUE #526: (Placeholder): Add new status condition to convey deprecation status of `Operator`
>[!NOTE]
>This issue is a placeholder issue and is subject to change

As part of implementing [RFC: OLMv1 Graph Deprecation](https://docs.google.com/document/d/1CYGANKILfzfOoWtHvyg4luF7i-cC1NRa7_1N9xTLatw/edit?usp=sharing) an `Operator` resource should have a status condition that conveys deprecation status.

The RFC proposes these types for the new status condition:
```go
const (
	TypeDeprecated = "Deprecated"
	
	ReasonPackageDeprecated = "PackageDeprecated"
	ReasonChannelDeprecated = "ChannelDeprecated"
	ReasonBundleDeprecated = "BundleDeprecated"
)
```

### LINKED ISSUE #527: (Placeholder): When `ForceSemverUpgradeConstraints=false` prefer deprecated bundles less during install/upgrade
>[!NOTE]
>This issue is a placeholder issue and is subject to change

As part of implementing [RFC: OLMv1 Graph Deprecation](https://docs.google.com/document/d/1CYGANKILfzfOoWtHvyg4luF7i-cC1NRa7_1N9xTLatw/edit?usp=sharing) deprecated bundles should be preferred less than non-deprecated bundles during resolution for installation and upgrading.

The preference will follow the pattern [non-deprecated, ordered by highest → lowest version] [deprecated, ordered by highest → lowest version]

As an example, consider the following scenario:
```
v1.0.0 <-- Deprecated
v1.1.0 <-- Installed
v1.2.0 <-- Long-Term Support
v1.3.0 <-- Deprecated
v1.4.0 <-- Deprecated
v1.5.0 <-- Long-Term Support
```
The resulting list of successors, ordered by most preferred first, would be:
```
v1.5.0
v1.2.0
v1.4.0 <-- Deprecated
v1.3.0 <-- Deprecated
```

everettraven:
# Description
- Add new condition types and reasons to signal when an installed bundle for a `ClusterExtension` has deprecations associated with it
- Update the `catalogmetadata.Bundle` type to have fields to hold the deprecation information for a bundle
- Update the resolution logic to prefer bundles with deprecations less than bundles without
- Update the `ClusterExtension` reconciler logic to add the deprecation statuses if there are any associated with the installed bundle

# Motivation
- Implements [RFC: OLMv1 Graph Deprecation](https://docs.google.com/document/d/1CYGANKILfzfOoWtHvyg4luF7i-cC1NRa7_1N9xTLatw/edit?usp=sharing)
- fixes #525 
- fixes #526 
- fixes #527

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/catalogmetadata/client/client.go
+++ b/internal/catalogmetadata/client/client.go
@@ -57,6 +57,7 @@ func (c *Client) Bundles(ctx context.Context) ([]*catalogmetadata.Bundle, error)
 		}
 		channels := []*catalogmetadata.Channel{}
 		bundles := []*catalogmetadata.Bundle{}
+		deprecations := []*catalogmetadata.Deprecation{}
 
 		rc, err := c.fetcher.FetchCatalogContents(ctx, catalog.DeepCopy())
 		if err != nil {
@@ -81,6 +82,12 @@ func (c *Client) Bundles(ctx context.Context) ([]*catalogmetadata.Bundle, error)
 					return fmt.Errorf("error unmarshalling bundle from catalog metadata: %s", err)
 				}
 				bundles = append(bundles, &content)
+			case declcfg.SchemaDeprecation:
+				var content catalogmetadata.Deprecation
+				if err := json.Unmarshal(meta.Blob, &content); err != nil {
+					return fmt.Errorf("error unmarshalling deprecation from catalog metadata: %s", err)
+				}
+				deprecations = append(deprecations, &content)
 			}
 
 			return nil
@@ -89,7 +96,7 @@ func (c *Client) Bundles(ctx context.Context) ([]*catalogmetadata.Bundle, error)
 			return nil, fmt.Errorf("error processing response: %s", err)
 		}
 
-		bundles, err = PopulateExtraFields(catalog.Name, channels, bundles)
+		bundles, err = PopulateExtraFields(catalog.Name, channels, bundles, deprecations)
 		if err != nil {
 			return nil, err
 		}
@@ -100,7 +107,7 @@ func (c *Client) Bundles(ctx context.Context) ([]*catalogmetadata.Bundle, error)
 	return allBundles, nil
 }
 
-func PopulateExtraFields(catalogName string, channels []*catalogmetadata.Channel, bundles []*catalogmetadata.Bundle) ([]*catalogmetadata.Bundle, error) {
+func PopulateExtraFields(catalogName string, channels []*catalogmetadata.Channel, bundles []*catalogmetadata.Bundle, deprecations []*catalogmetadata.Deprecation) ([]*catalogmetadata.Bundle, error) {
 	bundlesMap := map[string]*catalogmetadata.Bundle{}
 	for i := range bundles {
 		bundleKey := fmt.Sprintf("%s-%s", bundles[i].Package, bundles[i].Name)
@@ -118,6 +125,36 @@ func PopulateExtraFields(catalogName string, channels []*catalogmetadata.Channel
 			}
 
 			bundle.InChannels = append(bundle.InChannels, ch)
+		}
+	}
+
+	// According to https://docs.google.com/document/d/1EzefSzoGZL2ipBt-eCQwqqNwlpOIt7wuwjG6_8ZCi5s/edit?usp=sharing
+	// the olm.deprecations FBC object is only valid when either 0 or 1 instances exist
+	// for any given package
+	deprecationMap := make(map[string]*catalogmetadata.Deprecation, len(deprecations))
+	for _, deprecation := range deprecations {
+		deprecationMap[deprecation.Package] = deprecation
+	}
+
+	for i := range bundles {
+		if dep, ok := deprecationMap[bundles[i].Package]; ok {
+			for _, entry := range dep.Entries {
+				switch entry.Reference.Schema {
+				case declcfg.SchemaPackage:
+					bundles[i].Deprecations = append(bundles[i].Deprecations, entry)
+				case declcfg.SchemaChannel:
+					for _, ch := range bundles[i].InChannels {
+						if ch.Name == entry.Reference.Name {
+							bundles[i].Deprecations = append(bundles[i].Deprecations, entry)
+							break
+						}
+					}
+				case declcfg.SchemaBundle:
+					if bundles[i].Name == entry.Reference.Name {
+						bundles[i].Deprecations = append(bundles[i].Deprecations, entry)
+					}
+				}
+			}
 		}
 	}
 
--- a/internal/catalogmetadata/types.go
+++ b/internal/catalogmetadata/types.go
@@ -18,7 +18,7 @@ const (
 )
 
 type Schemas interface {
-	Package | Bundle | Channel
+	Package | Bundle | Channel | Deprecation
 }
 
 type Package struct {
@@ -29,6 +29,10 @@ type Channel struct {
 	declcfg.Channel
 }
 
+type Deprecation struct {
+	declcfg.Deprecation
+}
+
 type PackageRequired struct {
 	property.PackageRequired
 	SemverRange bsemver.Range `json:"-"`
@@ -36,8 +40,9 @@ type PackageRequired struct {
 
 type Bundle struct {
 	declcfg.Bundle
-	CatalogName string
-	InChannels  []*Channel
+	CatalogName  string
+	InChannels   []*Channel
+	Deprecations []declcfg.DeprecationEntry
 
 	mu sync.RWMutex
 	// these properties are lazy loaded as they are requested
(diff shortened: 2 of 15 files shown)


internal/catalogmetadata/client/client.go
```diff
@@ -121,5 +128,31 @@ func PopulateExtraFields(catalogName string, channels []*catalogmetadata.Channel
 		}
 	}
 
+	// if package is deprecated, add deprecation to bundle
+	// if channel is deprecated, add deprecation to bundle
+	// if bundle is deprecated, add deprecation to bundle
```
everettraven, line 133: Leftover notes to self. Need to remove them.

internal/catalogmetadata/client/client.go
```diff
@@ -121,5 +128,31 @@ func PopulateExtraFields(catalogName string, channels []*catalogmetadata.Channel
...
+						}
+					case declcfg.SchemaBundle:
+						if bundles[i].Name == entry.Reference.Name {
+							bundles[i].Deprecations = append(bundles[i].Deprecations, entry)
+						}
+					}
+				}
+			}
+		}
+	}
```
everettraven, line 155: This is very deeply nested. I'll look into ways to have less nesting here.

internal/controllers/clusterextension_controller.go
```diff
@@ -267,6 +273,65 @@ func mapBDStatusToInstalledCondition(existingTypedBundleDeployment *rukpakv1alph
...
 }
 
+// setDeprecationStatus will set the appropriate deprecation statuses for a ClusterExtension
+// based on the provided bundle
+func setDeprecationStatus(ext *v1alpha1.ClusterExtension, bundle *catalogmetadata.Bundle) {
+	// unset conditions
+	meta.RemoveStatusCondition(&ext.Status.Conditions, v1alpha1.TypePackageDeprecated)
+	meta.RemoveStatusCondition(&ext.Status.Conditions, v1alpha1.TypeChannelDeprecated)
+	meta.RemoveStatusCondition(&ext.Status.Conditions, v1alpha1.TypeBundleDeprecated)
+	meta.RemoveStatusCondition(&ext.Status.Conditions, v1alpha1.TypeDeprecated)
```
everettraven, line 283: Is it more appropriate to set the conditions to false and set necessary fields to their empty values?

internal/catalogmetadata/client/client.go
everettraven: Linter doesn't seem to be upset about it, anyone feel strongly about this nesting such that I need to refactor it?

internal/controllers/clusterextension_controller.go
everettraven: Unit tests fail when the conditions are removed. made it so that these conditions are always present after a reconciliation

internal/resolution/variablesources/required_package_test.go
```diff
@@ -63,6 +63,25 @@ func TestMakeRequiredPackageVariables(t *testing.T) {
...
+			Deprecations: []declcfg.DeprecationEntry{
+				{
+					Reference: declcfg.PackageScopedReference{
+						Schema: declcfg.SchemaBundle,
+						Name:   "test-package.v4.0.0",
+					},
+					Message: "test-package.v4.0.0 has been deprecated",
+				},
+			},
+		},
```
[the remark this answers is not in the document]
everettraven: done in https://github.com/operator-framework/operator-controller/pull/574/commits/ef355d035fcc9f6a0b243aa0c9ae371300a7e983

internal/catalogmetadata/types.go
```diff
@@ -140,6 +145,10 @@ func (b *Bundle) propertiesByType(propType string) []*property.Property {
 	return b.propertiesMap[propType]
 }
 
+func (b *Bundle) HasDeprecation() bool {
```
dtfranz, line 148: Nit-picky but maybe:
```suggestion
func (b *Bundle) IsDeprecated() bool {
```
reads a little better to me.

everettraven: I can see that. Reason I went with `HasDeprecation` is more so because of the potentially confusing classification of "Deprecated" in this case. If a channel is deprecated, that doesn't necessarily mean _every_ bundle in that channel is deprecated. The `ClusterExtension` would be marked deprecated if it specifically asks for that channel but otherwise that deprecation notice wouldn't show up. Due to this I felt `HasDeprecation` was more like asking "Are there any deprecations that are associated with this bundle?" rather than "Is _this_ bundle deprecated?"

That being said, I'm not opposed to changing this. I do feel that `IsDeprecated()` could be a bit misleading without having a deeper understanding of how the deprecation concept works.

dtfranz: Ah good point, and as you noted I myself also misunderstood the concept here by thinking that all bundles under a deprecated channel were also considered deprecated. If that's not the case then I agree with `HasDeprecation()`.

internal/catalogmetadata/client/client.go
dtfranz: Could this work? It would at least make the runtime not be `num(bundles) * num(deprecations)`
```suggestion
	deprecationMap := make(map[string]*catalogmetadata.Deprecation, 0)
	for _, deprecation := range deprecations {
		deprecationMap[deprecation.Package] = deprecation
	}

	for i := range bundles {
		if deprecation, ok := deprecationMap[bundles[i].Package]; ok {
			for _, entry := range deprecation.Entries {
				switch entry.Reference.Schema {
				case declcfg.SchemaPackage:
					bundles[i].Deprecations = append(bundles[i].Deprecations, entry)
				case declcfg.SchemaChannel:
					for _, ch := range bundles[i].InChannels {
						if ch.Name == entry.Reference.Name {
							bundles[i].Deprecations = append(bundles[i].Deprecations, entry)
						}
					}
				case declcfg.SchemaBundle:
					if bundles[i].Name == entry.Reference.Name {
						bundles[i].Deprecations = append(bundles[i].Deprecations, entry)
					}
				}
			}
		}
	}
```
Probably fine to leave the channels and entries loops in there though.

dtfranz: Oh, but if you can have multiple deprecation entries with the same package name that wouldn't work, would it?

internal/catalogmetadata/types.go
stevekuznetsov: Sounds like a good place to add GoDoc :)

internal/catalogmetadata/client/client.go
everettraven: Updated in the latest force push (force pushed to squash commits)

internal/catalogmetadata/types.go
everettraven: https://github.com/operator-framework/operator-controller/pull/574/commits/a2ff3ff37ae6f532db7c425aa4798eed57537f06

everettraven: One thing to point out with this that I thought about while writing the GoDoc is that during resolution if the bundle is associated with _any_ deprecations it will be less preferred over other bundles even if it is an olm.channel deprecation that isn't being considered due to the specification of using a different channel.

For example, given the following set of bundles in a desired channel `stable`:
```
foo.v1.0.0
foo.v1.1.0
foo.v2.0.0
foo.v2.1.0
foo.v3.0.0 <-- associated with a deprecated channel named stable-v3 due to $reasons
```

The resulting resolution list in order of most preferred -> least preferred would be:
```
foo.v2.1.0
foo.v2.0.0
foo.v1.1.0
foo.v1.0.0
foo.v3.0.0
```

This feels like a scenario we may want to avoid by having a distinction between a bundle having explicit deprecation due to the package or itself being deprecated and the looser deprecation associated with the bundle being an entry within a deprecated channel

everettraven: cc @joelanford @grokspawn ^

everettraven: I've updated the implementation in https://github.com/operator-framework/operator-controller/pull/574/commits/9186d16a4353a45131e3e460cf6ffa9d5d3886f9 to avoid the scenario I mentioned above

dtfranz (commented):
You've taken care of my concerns here and also handled the channel deprecation case well in my opinion, thanks!
I'll hold off on explicit approval for now just to make sure that some of the other reviewers get a chance to look at things :)

stevekuznetsov (approved):

PR #577, opened 2024-01-12
:seedling: don't consider bundle deprecated if the package is deprecated

everettraven:
# Description
- Updates the `catalogmetadata.Bundle.IsDeprecated()` method to no longer return true if there is a package deprecation associated with the bundle
- Adds unit tests for the `catalogmetadata.Bundle.IsDeprecated()` method 

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/catalogmetadata/types.go
+++ b/internal/catalogmetadata/types.go
@@ -157,19 +157,15 @@ func (b *Bundle) HasDeprecation() bool {
 
 // IsDeprecated returns true if the bundle
 // has been explicitly deprecated. This can occur
-// in one of two ways:
-// - the olm.package the bundle belongs to has been deprecated
-// - the bundle itself has been deprecated
+// if the bundle itself has been deprecated.
 // this function does not take into consideration
-// olm.channel deprecations associated with the bundle
-// as a bundle can be present in multiple channels with
-// some channels being deprecated and some not.
+// olm.channel or olm.package deprecations associated
+// with the bundle as a bundle can be present in multiple
+// channels with some channels being deprecated and some not
+// Similarly if the package is deprecated then all the bundles
+// are deprecated and there should be no special treatment
 func (b *Bundle) IsDeprecated() bool {
 	for _, dep := range b.Deprecations {
-		if dep.Reference.Schema == declcfg.SchemaPackage && dep.Reference.Name == b.Package {
-			return true
-		}
-
 		if dep.Reference.Schema == declcfg.SchemaBundle && dep.Reference.Name == b.Name {
 			return true
 		}
--- a/internal/catalogmetadata/types.go
+++ b/internal/catalogmetadata/types.go
@@ -162,8 +162,8 @@ func (b *Bundle) HasDeprecation() bool {
 // olm.channel or olm.package deprecations associated
 // with the bundle as a bundle can be present in multiple
 // channels with some channels being deprecated and some not
-// Similarly if the package is deprecated then all the bundles
-// are deprecated and there should be no special treatment
+// Package deprecation does not carry the same meaning as an individual
+// bundle deprecation, so package deprecation is not considered.
 func (b *Bundle) IsDeprecated() bool {
 	for _, dep := range b.Deprecations {
 		if dep.Reference.Schema == declcfg.SchemaBundle && dep.Reference.Name == b.Name {
--- a/internal/catalogmetadata/types_test.go
+++ b/internal/catalogmetadata/types_test.go
@@ -229,3 +229,55 @@ func TestBundleHasDeprecation(t *testing.T) {
 		})
 	}
 }
+
+func TestBundleIsDeprecated(t *testing.T) {
+	for _, tt := range []struct {
+		name       string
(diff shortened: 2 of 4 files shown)


internal/catalogmetadata/types.go
```diff
@@ -157,19 +157,15 @@ func (b *Bundle) HasDeprecation() bool {
...
+// with the bundle as a bundle can be present in multiple
+// channels with some channels being deprecated and some not
+// Similarly if the package is deprecated then all the bundles
+// are deprecated and there should be no special treatment
 func (b *Bundle) IsDeprecated() bool {
 	for _, dep := range b.Deprecations {
-		if dep.Reference.Schema == declcfg.SchemaPackage && dep.Reference.Name == b.Package {
-			return true
-		}
-
```
everettraven, line 172: Removing this will actually mean that explicitly deprecated bundles will be preferred less than ones that aren't even if the entire package is deprecated - fixing

joelanford: Looking at this a little closer (sorry I didn't get the chance with the original PR), I'm a little concerned that we're putting package and channel level deprecations in the `Bundle` type.

I think I would have expected that we would put the package deprecation in the `Package` type, channel deprecation in the `Channel` type and bundle deprecation in the `Bundle` type.

And if we had that, we could use composition by embedding the deprecation type in each of the package/channel/bundle types and then use a shared `IsDeprecated` function that comes with that deprecation type.

Is it possible to do something like this?

everettraven: Probably, but my concern with doing that is that we would likely need to restructure resolution to provide all those types as results from the resolution. IIUC the resolution process currently will only return the `Bundle` type that contains all the necessary information for continuing with the install/upgrade process.

everettraven: Restructuring resolution to return variables for `Package`, `Channel`, and `Bundle` types is out of scope for this work IMO

joelanford: > Removing this will actually mean that explicitly deprecated bundles will be preferred less than ones that aren't even if the entire package is deprecated - fixing

One other question about this point. Think about this scenario:
- Package `foo` is deprecated
- Bundle `foo.v1.0.0` is not deprecated
- Bundle `foo.v1.0.1` is deprecated

I would expect `foo.v1.0.0` to be preferred over `foo.v1.0.1`. The `package` deprecation should not cascade to the bundles one way or the other.

everettraven: https://github.com/operator-framework/operator-controller/pull/577/commits/6fdc060493762e4d0e4c20577ffc415ac84645aa - reverts to where the check is removed and should satisfy the expectation you have.

everettraven: @joelanford Could you take another look when you have some time?

internal/catalogmetadata/types.go
```diff
@@ -157,19 +157,15 @@ func (b *Bundle) HasDeprecation() bool {
...
+// if the bundle itself has been deprecated.
 // this function does not take into consideration
-// olm.channel deprecations associated with the bundle
-// as a bundle can be present in multiple channels with
-// some channels being deprecated and some not.
+// olm.channel or olm.package deprecations associated
+// with the bundle as a bundle can be present in multiple
+// channels with some channels being deprecated and some not
+// Similarly if the package is deprecated then all the bundles
+// are deprecated and there should be no special treatment
```
joelanford, line 166:
```suggestion
// Package deprecation does not carry the same meaning as an individual
// bundle deprecation, so package deprecation is not considered.
```

everettraven: Good catch. Updated in https://github.com/operator-framework/operator-controller/pull/577/commits/17799afa25efe9dcf4578c2656fc804b2f6105cf

internal/catalogmetadata/types_test.go
```diff
@@ -229,3 +229,55 @@ func TestBundleHasDeprecation(t *testing.T) {
...
+							Schema: "olm.channel",
+							Name:   "foo",
+						},
+					},
+				},
+			},
+		},
+		{
+			name:   "has no deprecation entries, not deprecated",
+			bundle: &catalogmetadata.Bundle{},
```
joelanford, line 259: Possibly a silly question: Does this `catalogmetadata.Bundle{}` need to have the package and channel deprecations listed for them to show up in the ClusterExtension deprecation status conditions, or is that handled in a separate data structure?

If it is handled in a separate data strucutre, could we change `catalogmetadata.Bundle` to contain something like:

```go
package catalogmetadata

type Bundle struct {
	...
	deprecated bool
}


```

It seems unnecessary to store a list of `Deprecations` where there will always be 0 or 1 and the schema/name will always be `olm.bundle` and `bundle.Name`.

everettraven: I think this question is related to https://github.com/operator-framework/operator-controller/pull/577#discussion_r1452452552

joelanford (approved):

PR #583, opened 2024-01-22
:sparkles: Bump rukpak to 0.17.0

varshaprasad96:
This commit makes the following changes:
1. Bump Rukpak to use version 0.17.0
2. Modify the BundleDeployment spec accordingly to not specify a bundle template.
3. Introduce `watchNamespaces` field in the spec that accepts a list of namespaces to watch.
4. Fix unit and e2e tests accordingly.

# Description

## Reviewer Checklist

- [ ] API Go Documentation


docs/drafts/support-watchNamespaces.md
```diff
@@ -0,0 +1,24 @@
...
+
+1. OwnNamespace: This mode allows the operator to monitor and respond to events within its own deployment namespace.
+1. SingleNamespace: In this mode, the operator is set up to observe events in a single, specific namespace other than the one it is deployed in.
+1. MultiNamespace: This enables the operator to function across multiple specified namespaces.
+1. AllNamespaces: Under this mode, the operator is equipped to monitor events across all namespaces within the cluster.
+
+When creating a cluster extension for deploying an operator, users have the option to define a list of `watchNamespaces`. This list determines the specific namespaces within which they intend the operator to operate. The configuration of `watchNamespaces` must align with the InstallModes supported by the operator as specified by the bundle author. The supported configurations in the order of preference are as follows:
+
```
varshaprasad96, line 11: Using the word "Operators" for now. Not sure of the naming convention we want to follow. ("Extension" vs "Operators")

internal/controllers/clusterextension_controller.go
```diff
@@ -387,28 +387,29 @@ func (r *ClusterExtensionReconciler) generateExpectedBundleDeployment(o ocv1alph
...
+			"type": string(rukpakv1alpha2.SourceTypeImage),
+			"image": map[string]interface{}{
+				"ref": bundlePath,
+			},
+		},
+	}
+
+	if len(o.Spec.WatchNamespaces) > 0 {
+		spec["watchNamespaces"] = o.Spec.WatchNamespaces
+	}
```
m1kola, line 404: Is it worth adding test coverage for this?

test/e2e/install_test.go
```diff
@@ -390,25 +389,8 @@ func getArtifactsOutput(t *testing.T) {
...
-		bundleYaml, err := yaml.Marshal(bundle)
-		if err != nil {
-			fmt.Printf("Failed to marshal bundle: %v", err)
-			continue
-		}
-		if err := os.WriteFile(filepath.Join(artifactPath, bundle.Name+"-bundle.yaml"), bundleYaml, 0600); err != nil {
-			fmt.Printf("Failed to write bundle to file: %v", err)
-		}
-	}
-
```
m1kola, line 409: Could you pleae explain why we are removing this?

docs/drafts/support-watchNamespaces.md
```diff
@@ -0,0 +1,24 @@
+# Install Modes and WatchNamespaces in OMLv1
```
m1kola, line 1: My understanding is that this implements [this brief](https://docs.google.com/document/d/1EGcJUjEqyA6CVfRHP5EFzdAJZ8EZAv5oBU5wTJfjhso/edit). Is this correct? Did we have an RFC for this or just the brief?

m1kola (commented):
A few questions

api/v1alpha1/clusterextension_types.go
```diff
@@ -65,6 +65,11 @@ type ClusterExtensionSpec struct {
 	//
 	// Defines the policy for how to handle upgrade constraints
 	UpgradeConstraintPolicy UpgradeConstraintPolicy `json:"upgradeConstraintPolicy,omitempty"`
+
+	//+kubebuilder:Optional
+	//
+	// watchNamespaces indicates which namespaces the operator should watch.
```
joelanford, line 71: Nit: General GoDoc best practice is for the first word of GoDoc to be the exact name of the identifier (so in this case `WatchNamespaces`). Spot check of `kubectl explain pod.spec` seems to mostly-ish follow that pattern, although it is admittedly hit or miss.

```suggestion
	// WatchNamespaces indicates which namespaces the operator should watch.
```

Should we say something here about this being a convention that the extension needs to support?

This also reminds me. Does rukpak have any handling of this field for plain bundles? Or is this right now only something that works with registry+v1 bundles?

docs/drafts/support-watchNamespaces.md
joelanford: My approach for the rename was:
- If we're talking about this API or a bundle in general, those are extensions.
- If we're talking specifically about the "Operator" pattern, which is a specific type of extension, it is okay to use "operator" terminology.

Since we're adding a "watchNamespaces" field directly to the ClusterExtension API, I think we're saying "extensions are things that may be able to be configured to watch specific namespaces", so I don't think that watching namespaces is strictly an operator concept.

test/e2e/install_test.go
joelanford: There is no such thing as the Bundle API anymore.

api/v1alpha1/clusterextension_types.go
ncdc: While it's true that godoc docs should typically start with a capital letter, because these comments are used to generate openapi schema descriptions, the k8s convention is to start with a lowercase letter so it matches the field name in the schema.

docs/drafts/support-watchNamespaces.md
everettraven: I believe only the brief. I think @varshaprasad96 was already working in this area before the brief was "accepted"

api/v1alpha1/clusterextension_types.go
varshaprasad96: > This also reminds me. Does rukpak have any handling of this field for plain bundles? Or is this right now only something that works with registry+v1 bundles?

This works only for registryV1 for now, wherein we generate the RBAC based on CSV. For the other bundle formats, we just apply whatever is present in the bundle. We would have to discuss further on how evaluate the operators' RBAC and validate if specified targetNs is in sync with what the operator supports. (have created an issue in rukpak: https://github.com/operator-framework/rukpak/issues/815). Probably use SAR? We can follow up in the other issue. 

Given this is only for registryV1 for now I don't think we are there yet to say that all extensions need to support this feature imo. Thoughts?

everettraven, line 71: Since this is on the `ClusterExtension` API I'd change "operator" to "extension":
```suggestion
	// watchNamespaces indicates which namespaces the extension should watch.
```

docs/drafts/support-watchNamespaces.md
varshaprasad96: It does so partially - in the sense we support specifying watchNamespaces only for registryV1. We only have brief for this at the moment. 

Additionally, there were changes which were done in https://github.com/operator-framework/rukpak/pull/774 that have been pulled in this PR - which involves:
- Removing bundle API
- Modifying the BD spec accordingly
- Introducing `watchNamespaces` to support all other install strategies - connected to the brief you mentioned

test/e2e/install_test.go
```diff
@@ -104,18 +104,18 @@ func TestClusterExtensionInstallRegistry(t *testing.T) {
 		assert.Contains(ct, cond.Message, "installed from")
 		assert.NotEmpty(ct, clusterExtension.Status.InstalledBundleResource)
 
-		bd := rukpakv1alpha1.BundleDeployment{}
+		bd := rukpakv1alpha2.BundleDeployment{}
 		assert.NoError(ct, c.Get(context.Background(), types.NamespacedName{Name: clusterExtensionName}, &bd))
-		hasValidBundle := apimeta.FindStatusCondition(bd.Status.Conditions, rukpakv1alpha1.TypeHasValidBundle)
-		if !assert.NotNil(ct, hasValidBundle) {
+		isUnpackSuccessfull := apimeta.FindStatusCondition(bd.Status.Conditions, rukpakv1alpha2.TypeUnpacked)
```
everettraven, line 109: Nit:
```suggestion
		isUnpackSuccessful := apimeta.FindStatusCondition(bd.Status.Conditions, rukpakv1alpha2.TypeUnpacked)
```

api/v1alpha1/clusterextension_types.go
joelanford: Yeah, we need to think about how this would work for other bundle types. For now, would it be reasonable for the field description to explicitly call out that this configuration only applies to registry+v1 bundles?

docs/drafts/support-watchNamespaces.md
m1kola: > We only have brief for this at the moment.

> I believe only the brief. I think @varshaprasad96 was already working in this area before the brief was "accepted"

That is ok. The brief is very brief.

Thanks for the clarification both. Will resolve this thread.

internal/controllers/clusterextension_controller.go
varshaprasad96: I've added an unit test for the method for now. An test to support this particular feature is already in rukpak. I'll create an issue to follow up and add an e2e test with a catalog for this.

api/v1alpha1/clusterextension_types.go
varshaprasad96: Done.

docs/drafts/support-watchNamespaces.md
varshaprasad96: Modified to be "Extension" now.

everettraven (approved):
LGTM

m1kola (approved):

joelanford (approved):

PR #598, opened 2024-01-24
:sparkles: Initial support for Extension

### LINKED ISSUE #596: Add Extension API

tmshort:
Fix #596

Based on ClusterExtension

# Description

## Reviewer Checklist

- [ ] API Go Documentation


tmshort:
Looking for feedback on the initial definition of `Extension` before hooking everything up.

api/v1alpha1/extension_types.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// ExtensionSpec defines the desired state of Extension
+type ExtensionSpec struct {
+	// +optional
+	// Pause reconciliation on this Extension
+	Paused *bool `json:"paused,omitempty"`
```
joelanford, line 51: It is somewhat frowned upon to use bools in kubernetes APIs because there are many cases where a 3rd (or more) state creeps in.

Should we try to make this an enum-style string field with a oneOf validation?

api/v1alpha1/extension_types.go
```diff
@@ -0,0 +1,112 @@
...
+	//+kubebuilder:validation:MaxLength:=64
+	//+kubebuilder:validation:Pattern:=^[a-z0-9]+([\.-][a-z0-9]+)*$
+	// ServiceAccount name used to install this extension
+	ServiceAccountName string `json:"serviceAccountName,omitempty"`
+
+	//+kubebuilder:validation:MaxLength:=64
+	//+kubebuilder:validation:Pattern:=^[a-z0-9]+([\.-][a-z0-9]+)*$
+	// +optional
+	// Location of installation TBD??
+	DefaultNamespace string `json:"defaultNamespace,omitempty"`
```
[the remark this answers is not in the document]
everettraven: +1 to using the install namespace as the default namespace. I'm not sure if there has been more discussion on implementing it yet, but maybe this can be repurposed as `WatchNamespaces`?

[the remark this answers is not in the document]
ncdc: The `App` CR has defaultNamespace. I was thinking it would be helpful here, but I'm totally fine waiting to add that later.

api/v1alpha1/extension_types.go
```diff
@@ -0,0 +1,112 @@
...
+
+	// Defintion of package to be installed
+	Package ExtensionPackage `json:"package,omitempty"`
```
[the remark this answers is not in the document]
ncdc: Source is required. Only 1 of its field may be specified, and is required. We can do that with CEL validation.

api/v1alpha1/extension_types.go
```diff
@@ -0,0 +1,112 @@
...
+// ExtensionSpec defines the desired state of Extension
+type ExtensionSpec struct {
+	// +optional
+	// Pause reconciliation on this Extension
+	Paused *bool `json:"paused,omitempty"`
+
+	//+kubebuilder:validation:MaxLength:=64
+	//+kubebuilder:validation:Pattern:=^[a-z0-9]+([\.-][a-z0-9]+)*$
+	// ServiceAccount name used to install this extension
+	ServiceAccountName string `json:"serviceAccountName,omitempty"`
```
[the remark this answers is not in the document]
ncdc: Correct

api/v1alpha1/extension_types.go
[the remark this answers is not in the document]
ncdc: What are you proposing, Kevin?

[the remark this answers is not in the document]
kevinrizza: Maybe I'm missing the context of this, but if we kept this field wouldn't that mean that if it's set to something other than the namespace the extension cr is applied to, we will template the namespace of manifests in the bundle to whatever this field is set to? I'm saying I see a use case for that.

api/v1alpha1/extension_types.go
```diff
@@ -20,7 +20,17 @@ import (
...
 
-type ExtensionPackage struct {
+// Although this looks like a bool, there might
+type ExtensionPaused string
+
+const (
+	// Pause resolution of this Extension
+	ExtensionPausedTrue ExtensionPaused = "True"
+	// Peform resolution of this Extension
+	ExtensionPausedFalse ExtensionPaused = "False"
```
joelanford, line 30: Sorry, should have been more specific earlier. What about something with:
- `Paused` and `Active` enum values
- `Active` would be default.

In the future, this would give us space to easily introduce another meaningful enum value.

api/v1alpha1/extension_types.go
```diff
@@ -0,0 +1,136 @@
...
+
+	//+kubebuilder:validation:MaxLength:=48
+	//+kubebuilder:validation:Pattern:=^[a-z0-9]+([\.-][a-z0-9]+)*$
+	// Channel constraint definition
+	Channel string `json:"channel,omitempty"`
+}
+
+type ExtensionSourceDirect struct {
+	// URL of the direct source
+	URL string `json:"url"`
```
varshaprasad96, line 56: Can `ExtensionSourceDirect` be structured the same as `ImageSource` in Rukpak?

```suggestion
type ExtensionSourceDirect struct {
	Image *ImageSource `json:"image,omitempty"`
}

type ImageSource struct {
	// Ref contains the reference to a container image containing Bundle contents.
	Ref string `json:"ref"`
	// ImagePullSecretName contains the name of the image pull secret in the namespace that the provisioner is deployed.
	ImagePullSecretName string `json:"pullSecret,omitempty"`
}
```

Ideally it would be nice to have this same as the entire `BundleSource`: https://github.com/operator-framework/rukpak/blob/6a8a84c9aff05efaba7b05992704ad38462a7ee8/api/v1alpha2/bundle_types.go#L45. 

Just the URL seems ambiguous. Additionally, it would be nice to have consistency b/w what goes into Rukpak and the options that will be exposed here. With direct reference installations, we are anyway going to pass the values directly to BD API.

ncdc: Remember we're planning on using kapp-controller's App CR, not rukpak/BundleDeployment. But yes, we should see what fields to put in based on that.

varshaprasad96: Right! Can we remove `ExtensionSourceDirect` from spec for now and add it later when we are ready for its implementation? We will have a better idea on what it should look like at that point in time.

api/v1alpha1/extension_types.go
tmshort: The spec named the field `paused`. Having the field and a value be the same makes little sense. Do you have a suggestion for new name?

api/v1alpha1/extension_types.go
ncdc: Makes sense

api/v1alpha1/extension_types.go
ncdc: I really like paused true/false and I'm not sure I'm convinced that in the future we'd want to expand paused with more values... but some brainstorming if we want to:

managed: active/paused
management.state: active/paused
reconcile: active/paused
reconciliation: active/paused

api/v1alpha1/extension_types.go
tmshort: I can remove `source.direct.url` from the spec... I'm assuming we still want to keep `source.package` as-is?

api/v1alpha1/extension_types.go
tmshort: I wrote this as `resolution` rather than reconciliation, so thinking `managed` might be more appropriate.

api/v1alpha1/extension_types.go
varshaprasad96: 👍 it contains all the catalog fields as present now, so looks good to me.

api/v1alpha1/extension_types.go
```diff
@@ -0,0 +1,134 @@
...
+*/
+
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Although this looks like a bool, there might
+type ExtensionManaged string
```
everettraven, line 24: More of a nit, but this name feels like it could be a bit misleading without looking more into the type when writing some code with it. Maybe something like:
- `ExtensionManagedState`
- `ManagedState` (I prefer this as it is shorter but less "descriptive" so I could see an argument for the longer one)

api/v1alpha1/extension_types.go
```diff
@@ -0,0 +1,134 @@
...
+
+const (
+	// Pause resolution of this Extension
+	ExtensionPausedTrue ExtensionManaged = "Active"
+	// Peform resolution of this Extension
+	ExtensionPausedFalse ExtensionManaged = "Paused"
```
everettraven, line 30: The names of these don't align with what I think our convention has been in the past regarding Condition naming and I feel this falls into a similar boat. I think something like:
```suggestion
	ExtensionManagedActive ExtensionManaged = "Active"
	// Peform resolution of this Extension
	ExtensionManagedPaused ExtensionManaged = "Paused"
```

is more indicative of these constants values and will, IMO, ultimately result in a better experience when reading and writing code that uses them.

tmshort: Yeah, I missed some renames :(

tmshort:
Ignoring the build failures (which will be fixed) is this looking like something we want to proceed with?

api/v1alpha1/extension_types.go
joelanford: I don't have a really strong opinion here. What I want to avoid is a situation where there is some other knob that we decide we need to add later that isn't completely orthogonal to `spec.paused`. I can't think of anything off hand, though. This does seem fairly cut and dry.

Not exactly the same, but the OLMv0 CatalogSource API has some fields that become inert based on the value of some other fields. Perhaps the main issue there is that we didn't add validation at the time that would prevent setting a field if something else made it inert.

api/v1alpha1/extension_types.go
```diff
@@ -0,0 +1,133 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type ExtensionManagedState string
+
+const (
+	// Pause resolution of this Extension
+	ManagedStateActive ExtensionManagedState = "Active"
+	// Peform resolution of this Extension
+	ManagedStatePaused ExtensionManagedState = "Paused"
```
joelanford, line 29: For the GoDoc comments, I think we should change `resolution` to `reconciliation` (assumption is that the intent of `spec.managed: "Paused"` means: "Don't resolve, update the underlying App, touch the Extension status, nothing")

api/v1alpha1/extension_types.go
```diff
@@ -0,0 +1,133 @@
...
+	// TODO: Direct *ExtensionSourceDirect `json:"direct,omitempty"`
+}
+
+// ExtensionSpec defines the desired state of Extension
+type ExtensionSpec struct {
+	//+kubebuilder:validation:Enum:=Active;Paused
+	//+kubebuilder:default:=Active
+	//+kubebuilder:Optional
+	//
+	// Pause reconciliation on this Extension
```
joelanford, line 69: A few nits:
- Start GoDoc with the name of the field. In Kubernetes APIs, the convention is to use lowercase so that the OpenAPI description docs align with the json field name.
- Mention what the semantics of each of the enumerated values are.
- 
```suggestion
	// managed controls the management state of the extension. "Active" means this extension will be reconciled and "Paused" means this Extension will be ignored.
```

api/v1alpha1/extension_types.go
```diff
@@ -0,0 +1,133 @@
...
+	//
+	// ServiceAccount name used to install this extension
+	ServiceAccountName string `json:"serviceAccountName"`
+
+	//+kubebuilder:validation:MaxLength:=64
+	//+kubebuilder:validation:Pattern:=^[a-z0-9]+([\.-][a-z0-9]+)*$
+	//+kubebuilder:Optional
+	//
+	// Location of installation TBD??
+	DefaultNamespace string `json:"defaultNamespace,omitempty"`
```
joelanford, line 83: Remove for now?

api/v1alpha1/extension_types.go
```diff
@@ -0,0 +1,133 @@
...
+	//+kubebuilder:default:=Active
+	//+kubebuilder:Optional
+	//
+	// Pause reconciliation on this Extension
+	Managed ExtensionManagedState `json:"managed,omitempty"`
+
+	//+kubebuilder:validation:MaxLength:=64
+	//+kubebuilder:validation:Pattern:=^[a-z0-9]+([\.-][a-z0-9]+)*$
+	//
+	// ServiceAccount name used to install this extension
```
joelanford, line 75:
```suggestion
	// serviceAccountName is the name of a service account in the Extension's namespace that will be used to manage the installation and lifecycle of the extension.
```

api/v1alpha1/extension_types.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+// TODO: Implement ExtensionSourceDirect containing a URL or other reference mechanism
+
+type ExtensionSource struct {
+	// A source package defined by a name, version and/or channel
+	Package *ExtensionSourcePackage `json:"package,omitempty"`
+
+	// A direct package defined by a URL
+	// TODO: Direct *ExtensionSourceDirect `json:"direct,omitempty"`
```
joelanford, line 60: Remove for now?

api/v1alpha1/extension_types.go
```diff
@@ -0,0 +1,133 @@
...
+
+	// Source of Extension to be installed
+	Source ExtensionSource `json:"source"`
+
+	//+kubebuilder:validation:Enum:=Enforce;Ignore
+	//+kubebuilder:default:=Enforce
+	//+kubebuilder:Optional
+	//
+	// Defines the policy for how to handle upgrade constraints
+	UpgradeConstraintPolicy UpgradeConstraintPolicy `json:"upgradeConstraintPolicy,omitempty"`
```
joelanford, line 93: What is our perception of when this field will be applicable?

If we add a way to directly reference a bundle in the future, that means we are not using a catalog. However a catalog is where we (currently) encode all upgrade edge information. There is no upgrade edge information on individual bundles, and there is no implicit versioning scheme (like semver) today. Ergo, there is no way to implement upgrade constraint policies without a catalog.

Should we move this field into the `ExtensionSourcePackage` struct?

api/v1alpha1/extension_types.go
tmshort: The discussion above (@kevinrizza) was to keep it

api/v1alpha1/extension_types.go
tmshort: Remove more than as it currently is?

api/v1alpha1/extension_types.go
```diff
@@ -0,0 +1,133 @@
...
+	// Pause resolution of this Extension
+	ManagedStateActive ExtensionManagedState = "Active"
+	// Peform resolution of this Extension
+	ManagedStatePaused ExtensionManagedState = "Paused"
+)
+
+type ExtensionSourcePackage struct {
+	//+kubebuilder:validation:MaxLength:=48
+	//+kubebuilder:validation:Pattern:=^[a-z0-9]+(-[a-z0-9]+)*$
+	Name string `json:"name"`
```
varshaprasad96, line 35: Do we want to mark this required?

api/v1alpha1/extension_types.go
```diff
@@ -0,0 +1,133 @@
...
+	//+kubebuilder:Optional
+	//
+	// Pause reconciliation on this Extension
+	Managed ExtensionManagedState `json:"managed,omitempty"`
+
+	//+kubebuilder:validation:MaxLength:=64
+	//+kubebuilder:validation:Pattern:=^[a-z0-9]+([\.-][a-z0-9]+)*$
+	//
+	// ServiceAccount name used to install this extension
+	ServiceAccountName string `json:"serviceAccountName"`
```
varshaprasad96, line 76: This can be marked required too, I guess?

api/v1alpha1/extension_types.go
varshaprasad96: If we are planning to keep it, can we elaborate it further by mentioning that it is the default namespace where Extension resources are installed, and it defaults to the ns where the Extension is present.

Also do we need the max length validation for namespace? I think the allowed resource name limit of [253](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names) characters applies to ns too.

api/v1alpha1/extension_types.go
tmshort: It's not marked as `+kubebuilder:Optional`

api/v1alpha1/extension_types.go
tmshort: It's not marked as `+kubebuilder:Optional`
My initial tests have indicated that it's required.

api/v1alpha1/extension_types.go
joelanford: Just a nit and completely personal opinion. I prefer issues to describe TODOs over comments in code. In my experience, TODO comments rot more than issues do.

internal/controllers/clusterextension_controller.go
```diff
@@ -71,12 +63,12 @@ type ClusterExtensionReconciler struct {
 //+kubebuilder:rbac:groups=catalogd.operatorframework.io,resources=catalogmetadata,verbs=list;watch
 
 func (r *ClusterExtensionReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	l := log.FromContext(ctx).WithName("operator-controller")
+	l := log.FromContext(ctx).WithName("clusterextensionerator-controller")
```
joelanford, line 66:
```suggestion
	l := log.FromContext(ctx).WithName("clusterextension-controller")
```

joelanford (commented):
What's the thought on order of operations for fully switching over to Carvel? With this PR, we would have ClusterExtensions and Extensions both managing BundleDeployments.

And @varshaprasad96 is working on replacing BD with Carvel App.

Not sure what you all are thinking, but my hunch is that coding and review would be fastest with the following order:
1. Add Extension API with no controller
2. Varsha adds Extension controller with App support (essentially pretending like ClusterExtension API and controller do not exist).
3. We remove ClusterExtension API and controller.

I'm worried that we'll spend unnecessary time writing code, updating tests, reviewing, and discussing/handling edge cases if we approach this as keeping both or implementing Extension+BundleDeployment.

api/v1alpha1/extension_types.go
```diff
@@ -0,0 +1,169 @@
...
+	//
+	// serviceAccountName is he name of a service account in the Extension's namespace that will be used to manage the installation and lifecycle of the extension.
+	ServiceAccountName string `json:"serviceAccountName"`
+
+	//+kubebuilder:validation:MaxLength:=253
+	//+kubebuilder:validation:Pattern:=^[a-z0-9]+([\.-][a-z0-9]+)*$
+	//+kubebuilder:Optional
+	//
+	// defaultNamespace is the location of installation if different than the resource namespace
+	DefaultNamespace string `json:"defaultNamespace,omitempty"`
```
joelanford, line 93: I still vote that we drop this for now. We already have existing logic for namespace defaulting in OLMv0, which I think we would hardcode into the YTT template we use?

If/when we care to support other bundle types, we can revisit.

api/v1alpha1/extension_types.go
```diff
@@ -0,0 +1,169 @@
...
+
+	// source of Extension to be installed
+	Source ExtensionSource `json:"source"`
+
+	//+kubebuilder:Optional
+	//
+	// watchNamespaces indicates which namespaces the extension should watch.
+	// This feature is currently supported only with RegistryV1 bundles.
+	WatchNamespaces []string `json:"watchNamespaces,omitempty"`
```
joelanford, line 102: Let's drop `WatchNamespace` for now too. This is what I was proposing in our OLM WG sync this morning. I think we can add in a values field later based on this issue https://github.com/operator-framework/operator-controller/issues/612.

api/v1alpha1/extension_types.go
tmshort: This is currently unused in the code I have, so not a problem.

api/v1alpha1/extension_types.go
tmshort: This is currently supported by rukpak, but I ought to be able to get rid of it, and remove any appropriate tests.

internal/controllers/clusterextension_controller.go
tmshort: Well... that was a typo.

joelanford:
IMO, all of the changes that are made in this PR that make it possible to reuse code for both ClusterExtensions and Extensions is essentially adding an extra layer of abstraction that we won't want or need once we remove the ClusterExtension API.

Even just the time that folks will spend reviewing seems pretty steep (>5000 LoC change in this PR right now).

If we're going to have a Extension+BundleDeployment controller in place in this transition, it seems like it might be easier if we order things somewhat like this:
1. rename ClusterExtension to Extension and make it namespace-scoped, but otherwise no changes to the API. Also make whatever changes are necessary in the controller (if any) to account for this being namespace-scoped.
2. swap out Rukpak for Carvel
3. Update Extension API to match what we have here.

Any maybe 2 and 3 can be reversed?

varshaprasad96:
Adding some thoughts:

> 2. swap out Rukpak for Carvel

How long do we want to support Rukpak and Kapp controller side-by-side? Based on the previous discussions, as I understand we want kapp-ctrl to be under feature gate, and still support Rukpak before making a complete swap.

If that is going to be the case, the initial sequence of steps seem more apt - https://github.com/operator-framework/operator-controller/pull/598#pullrequestreview-1866147368, to avoid supporting an intermediate Extension+BD at all, which users can find confusing - a namespace scoped API supporting 2 appliers underneath that support different scoping. Though there would be some rework, it would keep things clearer in terms of 

a. ClusterExtension+Rukpak 
b. Extension+Kapp.
Once we are confident enough with (b), we remove (a) in entirety (along with any relevant tests). 

If there is going to be a complete immediate swap without needing to support both Rukpak and Kapp, we could do https://github.com/operator-framework/operator-controller/pull/598#issuecomment-1932042795. Would prefer supporting Carvel first with bare minimal specs in Extension API, so that we can break down and follow up with requirements to complete (3).

joelanford:
> a. ClusterExtension+Rukpak
> b. Extension+Kapp.
> Once we are confident enough with (b), we remove (a) in entirety (along with any relevant tests).

This also seems like a reasonable path forward without too much churn.

tmshort:
I seem to recall a feature gate possibility, too? Such that we can do either rukpak or kapp-controller. In such a case, I'm now leaning toward what @varshaprasad96 is saying: keep ClusterExtension + rukpak, and add Extension without it being _fully_ plumbed.

In which case, I can remove/revert some of the code and keep a partial controller in place to use with kapp-controller?

tmshort:
With @joelanford 's response just before mine, I think we're in general agreement?

tmshort:
This is about 1/3rd of the previous iteration.

api/v1alpha1/extension_types.go
tmshort: I've removed a number of TODOs from the code I'm adding, and it's specifically for kapp-controller

api/v1alpha1/extension_types.go
```diff
@@ -0,0 +1,140 @@
...
+
+const (
+	// Peform reconcilliation of this Extension
+	ManagedStateActive ExtensionManagedState = "Active"
+	// Pause reconcilliation of this Extension
+	ManagedStatePaused ExtensionManagedState = "Paused"
+)
+
+type ExtensionSourcePackage struct {
+	//+kubebuilder:validation:MaxLength:=48
```
everettraven, line 34: Why 48? 

I took a look at the operatorhubio catalog and some of the package names get to the high thirties with the highest I saw being a length 40 which makes me feel that putting an arbitrary limit to package name length being 48 might be insufficient. I'm also not aware of this length being enforced during the catalog validation process, but I could just not be aware of it.

For reference, this is the command I ran to check the operatorhubio catalog package name lengths:
```sh
opm render quay.io/operatorhubio/catalog:latest | jq -s '.[] | select(.schema == "olm.package") | .name | length'
```

internal/controllers/extension_controller.go
```diff
@@ -0,0 +1,203 @@
...
+		ext.Status.ResolvedBundleResource = ""
+		setResolvedStatusConditionFailed(&ext.Status.Conditions, err.Error(), ext.GetGeneration())
+
+		setDeprecationStatusesUnknown(&ext.Status.Conditions, "deprecation checks have not been attempted as resolution failed", ext.GetGeneration())
+		return ctrl.Result{}, err
+	}
+
+	// TODO: kapp-controller integration
+	// * lookup the bundle in the solution/selection that corresponds to the Extension's desired Source
+	// * set the status of the Extension based on the respective deployed application status conditions.
```
everettraven, line 150: Is the plan here to still use both rukpak and kapp control and have the paths guarded by feature gates?

api/v1alpha1/extension_types.go
tmshort: 48 is what the ClusterExtension API uses for these fields.
https://github.com/operator-framework/operator-controller/blob/2df4bceb4cbab606a5e9dd14490900a499d27354/api/v1alpha1/clusterextension_types.go#L43-L45

internal/resolution/variablesources/installed_package.go
```diff
@@ -23,7 +24,7 @@ import (
 // has own variable.
 func MakeInstalledPackageVariables(
 	allBundles []*catalogmetadata.Bundle,
-	clusterExtensions []ocv1alpha1.ClusterExtension,
+	clusterExtensions []internal.ExtensionInterface,
```
everettraven, line 27: `clusterExtensions` --> `extensions`?

tmshort:
ping @joelanford @varshaprasad96 ?

api/v1alpha1/extension_types.go
```diff
@@ -0,0 +1,140 @@
...
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+type ExtensionManagedState string
+
+const (
+	// Peform reconcilliation of this Extension
+	ManagedStateActive ExtensionManagedState = "Active"
```
perdasilva, line 28: would it be necessary to add a new print column to show the state when doing kubectl get?

internal/controllers/common_controller.go
```diff
@@ -0,0 +1,119 @@
...
+	apimeta.SetStatusCondition(conditions, metav1.Condition{
+		Type:               ocv1alpha1.TypeInstalled,
+		Status:             metav1.ConditionFalse,
+		Reason:             ocv1alpha1.ReasonInstallationFailed,
+		Message:            message,
+		ObservedGeneration: generation,
+	})
+}
+
+func setDeprecationStatusesUnknown(conditions *[]metav1.Condition, message string, generation int64) {
```
perdasilva, line 102: every other function got a comment, except for me!

tmshort: That's the way it was in the original code (I'm guessing).

api/v1alpha1/extension_types.go
joelanford: It is because Rukpak reserves some characters for its use so that it can use the full clusterextension name when it creates secondary objects for the clusterextension.

As much as possible, it is best to fully define, enforce, and err towards overly strict schemas because once we GA, we cannot relax them without a breaking change.

joelanford:
My overriding thought here is that we haven't actually discussed at all how or if resolution should happen for namespace-scoped Extensions, and I think we need to before we have a controller for the Extension API.

If we don't, I fear that we'll commit some code that sets some precedent that our future selves begin to internalize as the way things are done before we know if that's what we actually want.

Perhaps instead of doing resolution stuff in this PR, we could instead have a dummy controller that simply updates a status condition with something like Resolved=False, Installed=False with Reason=TODONotImplemented and message="not implemented" or something along those lines?

cmd/manager/main.go
```diff
@@ -123,6 +123,17 @@ func main() {
...
+		if err = (&controllers.ExtensionReconciler{
+			Client:         cl,
+			BundleProvider: catalogClient,
+			Scheme:         mgr.GetScheme(),
+			Resolver:       resolver,
+		}).SetupWithManager(mgr); err != nil {
+			setupLog.Error(err, "unable to create controller", "controller", "Extension")
+			os.Exit(1)
+		}
+	}
```
everettraven, line 136: One of the implications with not even adding a reconciler for the `Extension` API to the controller manager is that if this feature gate is disabled, creating an `Extension` resource will silently do nothing. From a UX perspective I think it would make sense to still add the reconciler but when the feature gate is disabled we simply add a status condition that states something along the lines of "Reconciliation skipped. feature gate for the Extension API is disabled. The Extension API is an alpha feature that is still in development. To enable reconciliation use the flag --feature-gates=EnableExtensionApi=true".

Doing this would at least signal to an end user that it is currently in a "do nothing" state and how to enable it.

everettraven: That being said, I'm also not sure of any API versioning constraints that come with adding a status condition for this. Definitely something to think about regarding the pros/cons

internal/controllers/extension_controller.go
```diff
@@ -0,0 +1,177 @@
...
+// Helper function to do the actual reconcile
+//
+// Today we always return ctrl.Result{} and an error.
+// But in the future we might update this function
+// to return different results (e.g. requeue).
+//
+//nolint:unparam
+func (r *ExtensionReconciler) reconcile(_ context.Context, ext *ocv1alpha1.Extension) (ctrl.Result, error) {
+	// Don't do anything if Paused
+	if ext.Spec.Managed == ocv1alpha1.ManagedStatePaused {
```
varshaprasad96, line 107: Can we add a log here stating no reconciliation done because the managedState is set to paused?

internal/controllers/extension_controller.go
```diff
@@ -0,0 +1,177 @@
...
+		For(&ocv1alpha1.Extension{}).
+		Watches(&catalogd.Catalog{},
+			handler.EnqueueRequestsFromMapFunc(extensionRequestsForCatalog(mgr.GetClient(), mgr.GetLogger()))).
+		Complete(r)
+}
+
+// Generate reconcile requests for all extensions affected by a catalog change
+func extensionRequestsForCatalog(c client.Reader, logger logr.Logger) handler.MapFunc {
+	return func(ctx context.Context, _ client.Object) []reconcile.Request {
+		// no way of associating an extension to a catalog so create reconcile requests for everything
```
varshaprasad96, line 159: Given we are not mapping catalog and extension, and creating reconcile requests for all extensions, I don't think we need this mapping func. Watching all catalogs should be enough (in the current state).

cmd/manager/main.go
tmshort: Yeah, I'm a bit torn. on where to put the feature gate.

internal/controllers/extension_controller.go
tmshort: I believe this is still necessary as the comment says, there's no way to indicate ownership due to the namespace scope of `Extension`. I can look, but this is how `ClusterExtension` does it.

tmshort: I think I figured out the correct method.

cmd/manager/main.go
varshaprasad96: The ideal scenario would be to not have the Extension CRD applied if the feature gate is disabled - but I don't think that's something which is possible here, since the CRDs are applied before the controller pod is created. Instead of setting up the reconciler by default and adding the overhead of watching Extension resources, what do you all think about adding a log/warning during startup, mentioning the value of the feature gate flag, and explicitly stating that - "feature gate flag for `ExtensionAPI` set to false, no action will be taken by operator controller if Extension object is created". That should to some extent satisfy the UI difficulties without having us to take more effort on it for now?

everettraven: @varshaprasad96 I think that assumes a user who might create the Extension resource would have an understanding to go and look at the logs of the operator-controller pod and explicitly find that singular log message.

Throwing a message on the status is more immediately visible IMO and to me doesn't feel like a bunch of overhead. I'm mainly approaching this from what I would want as a user and I would much prefer seeing a status be set than having to find the pod that is running the controller manager, get the logs, and search the logs for any indication as to why nothing happened

tmshort: @everettraven has a point, if your extension isn't running, you're going to look at your extension first, and not necessarily a pod to which you don't have log access to.

varshaprasad96: Some thoughts on why this suggestion:

> I think that assumes a user who might create the Extension resource would have an understanding to go and look at the logs of the operator-controller pod and explicitly find that singular log message.

I would expect that users who create Extension resources have the understanding of troubleshooting at this stage, which includes checking logs when something doesn't work as expected - this is particularly considering the fact that afaik there is no one trying to use v1 in production in near future, at least till we make kapp-ctrl switch -scheduled soon. 

That being said, I understand and completely agree that having just a log line - that too in an o-c pod that the user trying to install the package, ideally should not concern with - is not the perfect UX. This is to avoid - both the overhead of having an additional reconciler watch extension and other relevant resources on cluster which is not necessary, and us spending time on an API field that could turn out to be temporary in very near future. On a quick search upstream found an [example](https://github.com/Kong/kubernetes-ingress-controller/blob/d5cb0fa57d7d6f9360f1b8f58859c6b633cda1b6/internal/manager/controllerdef.go#L278-L279) of one of the optional controllers enabled only when the feature gate is present.

internal/controllers/extension_controller.go
joelanford: For follow-up, check in other popular Kubernetes projects to and see if there is a pattern where a Paused condition is set.

config/rbac/extension_viewer_role.yaml
```diff
@@ -0,0 +1,27 @@
...
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    app.kubernetes.io/name: clusterrole
+    app.kubernetes.io/instance: extension-viewer-role
+    app.kubernetes.io/component: rbac
+    app.kubernetes.io/created-by: operator-controller
+    app.kubernetes.io/part-of: operator-controller
+    app.kubernetes.io/managed-by: kustomize
```
joelanford, line 11: Nit: drop all these labels (here and in the editor role). These are setup by SDK/Kubebuilder, right? I don't think they provide any real value.

config/rbac/extension_editor_role.yaml
```diff
@@ -0,0 +1,31 @@
...
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - olm.operatorframework.io
+  resources:
+  - extensions/status
+  verbs:
+  - get
```
joelanford, line 31: Nit: Is `get` a valid verb on the CRD status subresource? I'm fairly sure it isn't and this rule accomplishes nothing. (Is this another SDK/Kubebuilder-ism?)

config/rbac/extension_viewer_role.yaml
```diff
@@ -0,0 +1,27 @@
...
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - olm.operatorframework.io
+  resources:
+  - extensions/status
+  verbs:
+  - get
```
joelanford, line 27: Same question as above about `get` on the status subresource.

config/rbac/role.yaml
```diff
@@ -51,3 +51,29 @@ rules:
...
+  resources:
+  - extensions/finalizers
+  verbs:
+  - update
+- apiGroups:
+  - olm.operatorframework.io
+  resources:
+  - extensions/status
+  verbs:
+  - get
```
joelanford, line 77: And here.

config/rbac/extension_viewer_role.yaml
tmshort: If I remove them, they might be re-added?

config/rbac/extension_editor_role.yaml
tmshort: Yes, I believe kubebuilder did this, and might put it back...

config/rbac/extension_viewer_role.yaml
joelanford: I'm guessing this is part of `operator-sdk create api`? So no they would not be re-added.

config/samples/olm_v1alpha1_extension.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: olm.operatorframework.io/v1alpha1
+kind: Extension
+metadata:
+  labels:
+    app.kubernetes.io/name: extension
+    app.kubernetes.io/instance: extension-sample
+    app.kubernetes.io/part-of: operator-controller
+    app.kubernetes.io/managed-by: kustomize
+    app.kubernetes.io/created-by: operator-controller
```
joelanford, line 9: Also remove these labels?

config/rbac/extension_viewer_role.yaml
tmshort: Not sure, but I'll try; there definitely a build process that regenerates _some_ of these types of things.

config/rbac/extension_viewer_role.yaml
joelanford, line 1: Do these actually get included in the kustomize output? It looks like not because they are not listed in the kustomization.yaml file (neither are the ClusterExtension roles).

config/rbac/extension_editor_role.yaml
tmshort: Why would kubebuilder put on an invalid verb for a resource such as this?

internal/controllers/validators/extension_validators.go
```diff
@@ -0,0 +1,58 @@
...
+	// we need to make a decision on whether we want to run all validators or stop at the first error. If the the former,
+	// we should consider how to present this to the user in a way that is easy to understand and fix.
+	// this issue is tracked here: https://github.com/operator-framework/operator-controller/issues/167
+	for _, validator := range validators {
+		if err := validator(e); err != nil {
+			return err
+		}
+	}
+	return nil
+}
```
joelanford, line 58: Ideally, we would not need reconcile-time validators at all because we instead have validation done via CRD schema validation.

I won't push to remove this validation code now, but I think we should capture a follow-up that removes it and then beefs up tests of the CRD validation (if they aren't already beefy) to make sure our CRD schema validation functions catch everything we expect. If anything slips through, it likely fails anyway somewhere else during reconciliation.

config/rbac/extension_editor_role.yaml
tmshort: `client.Status().Get()` does not seem to exist... so perhaps you can't get it.

internal/controllers/extension_controller_test.go
```diff
@@ -0,0 +1,210 @@
...
+		Namespace: fmt.Sprintf("namespace-%s", rand.String(8)),
+	}
+	namespace := &corev1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: extKey.Namespace,
+		},
+	}
+	require.NoError(t, cl.Create(ctx, namespace))
+
+	t.Log("By injecting creating a client with the bad clusterextension CR")
```
joelanford, line 138:
```suggestion
	t.Log("By injecting creating a client with the bad extension CR")
```

config/rbac/extension_viewer_role.yaml
tmshort: No idea, but they were generated... we can remove them later if deemed necessary?

internal/controllers/extension_controller_test.go
```diff
@@ -0,0 +1,210 @@
...
+	pkgName := fmt.Sprintf("exists-%s", rand.String(6))
+	extension := &ocv1alpha1.Extension{
+		ObjectMeta: metav1.ObjectMeta{Name: extKey.Name, Namespace: extKey.Namespace},
+		Spec: ocv1alpha1.ExtensionSpec{
+			Source: ocv1alpha1.ExtensionSource{
+				Package: &ocv1alpha1.ExtensionSourcePackage{
+					Name:    pkgName,
+					Version: "1.2.3-123abc_def", // bad semver that matches the regex on the CR validation
```
joelanford, line 146: This passes our CRD schema validation, but does not parse with the masterminds parser? If that's the case, how hard would it be to update the CRD pattern to disallow this?

tmshort: This was from cluster extension (originally operator), we might have to look at the original PR to understand why... my recollection was complexity of the regex, maybe?

tmshort: Looking further into the test, it purposefully bypasses CR validation, and exercises the validation code.

api/v1alpha1/clusterextension_types.go
```diff
@@ -158,3 +159,21 @@ type ClusterExtensionList struct {
...
+	return p
+}
+
+func (r *ClusterExtension) GetUID() types.UID {
+	return r.ObjectMeta.GetUID()
+}
+
+func (r *ClusterExtension) GetUpgradeConstraintPolicy() UpgradeConstraintPolicy {
+	return r.Spec.UpgradeConstraintPolicy
+}
```
joelanford, line 179: Drop this since UpgradeConstraintPolicy is now part of the package struct?

tmshort: The interface function is not being used, so I'm removing it

api/v1alpha1/clusterextension_types.go
```diff
@@ -158,3 +159,21 @@ type ClusterExtensionList struct {
...
 	SchemeBuilder.Register(&ClusterExtension{}, &ClusterExtensionList{})
 }
+
+func (r *ClusterExtension) GetPackageSpec() *ExtensionSourcePackage {
+	p := &ExtensionSourcePackage{}
+
+	p.Channel = r.Spec.Channel
+	p.Name = r.Spec.PackageName
+	p.Version = r.Spec.Version
+
```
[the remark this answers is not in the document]
tmshort: Not needed yet, we have no processing for it in Extension

config/rbac/extension_viewer_role.yaml
joelanford: Sure.

config/rbac/extension_editor_role.yaml
joelanford: I'll dig in kubebuilder code. But I bet just an uninformed addition to the generation that seemed like a good idea.

joelanford: Wow, it has been there for a very long time (5 years). I suspect it was overlooked originally and is otherwise completely inert so no one has ever noticed.

config/rbac/extension_viewer_role.yaml
joelanford: I guess I'd suggest adding them to the kustomization file?

(It looks like I have a few issues to open in Kubebuilder 😅 )

api/v1alpha1/clusterextension_types.go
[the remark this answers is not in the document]
joelanford: I was thinking just in terms of symmetry with what you'd get from `Extension.GetPackageSpec()`.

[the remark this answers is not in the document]
m1kola: Should we consider removing `GetPackageSpec` and `ExtensionInterface` in general? It looks like all the functions right now receive concrete types. Someday, and that day may never come, we will add an interface. But until that day, we should be fine without it.

m1kola (commented):
I left a few questions and suggestions.

Ask for the next time - can we please split large PRs into smaller ones? For this PR split could look something like this:
1. Refactor existing controllers to extract generic parts
2. Add a new `Extension` type & generate manifests for it
3. Introduce a new controller for `Extension`

It is easier to review smaller PRs.

P.S: I'm guilty of making large PRs myself, but I'm trying :)

internal/internal.go
```diff
@@ -0,0 +1,45 @@
...
+
+import (
+	"k8s.io/apimachinery/pkg/types"
+
+	ocv1alpha1 "github.com/operator-framework/operator-controller/api/v1alpha1"
+)
+
+type ExtensionInterface interface {
+	GetPackageSpec() *ocv1alpha1.ExtensionSourcePackage
+	GetUpgradeConstraintPolicy() ocv1alpha1.UpgradeConstraintPolicy
```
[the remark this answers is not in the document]
m1kola: Since I started review `GetUpgradeConstraintPolicy` was resolved. So resolving this thread. but I have a suggestion about the whole interface [here](https://github.com/operator-framework/operator-controller/pull/598#discussion_r1491036970).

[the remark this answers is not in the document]
tmshort: (already outdated!)

joelanford (approved):
I think this PR is still a bit large to be able to engage in high fidelity review, and I'm feeling a bit of review fatigue with all of the moving parts.

However, I would like to unblock further work, so perhaps we could merge what we have and then follow back up with smaller PRs that cover the following:

- Focused discussion and possibly PR about `spec.managed` staying as an enum to converting back to a bool. I know I suggested the enum, but the more I think about it, the more I agree with Andy that this feature is unlikely to ever expand beyond two values.
- Some cleanup from kubebuilder-generated stuff. I submitted some issues in upstream kubebuilder about it
  - https://github.com/kubernetes-sigs/kubebuilder/issues/3781
  - https://github.com/kubernetes-sigs/kubebuilder/issues/3782
  - https://github.com/kubernetes-sigs/kubebuilder/issues/3783
- Remove reconcile-time validations and tests. We should be able to have robust testing for our CRD validations that happen at admission time
- Remove unused fields from extension reconciler (e.g. BundleProvider, Scheme, etc.). We can add them when we need them.
- Remove ExtensionInterface and associated functions. Current plan is to remove ClusterExtension, so no need to introduce an abstraction that will very shortly obsolete when we do remove ClusterExtension. If plans change and ClusterExtension sticks around, we can re-assess the need for an abstraction.
- Mirror the paused/active spec value to the status.
- Make the `ExtensionSource` struct use the [union discriminator pattern](https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/1027-api-unions/README.md#discriminator-field), and add associated CRD schema validation.
- Refactor tests to be more table-driven.

tmshort:
@joelanford are we good to merge, with issues created for those items listed?

config/samples/olm_v1alpha1_extension.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: olm.operatorframework.io/v1alpha1
+kind: Extension
+metadata:
+  name: extension-sample
+  namespace: extension-namespace
+spec:
+  paused: false
+  serviceAccountName: extension-sa
+  defaultNamespace: extension-sample
+  package:
```
varshaprasad96, line 10: Was working based of this PR, realized that this sample need to be updated according to the recent changes made in the spec.

varshaprasad96 (commented):
@tmshort Just a small update comment on samples based on recent changes made in the PR.

config/samples/olm_v1alpha1_extension.yaml
varshaprasad96: Basically this:
```yaml
apiVersion: olm.operatorframework.io/v1alpha1
kind: Extension
metadata:
  labels:
    app.kubernetes.io/name: extension
    app.kubernetes.io/instance: extension-sample
    app.kubernetes.io/part-of: operator-controller
    app.kubernetes.io/managed-by: kustomize
    app.kubernetes.io/created-by: operator-controller
  name: extension-sample
  namespace: extension-sample
spec:
  paused: false
  serviceAccountName: extension-sa
  source:
    package:
      name: argcocd-operator
      version: <>
```

varshaprasad96: Not the labels, just the spec. Removal of default namespace, embedding package inside source and also renaming paused to managed.
I was trying to use this sample to test extensions with App CR, which is when realized that some of the fields are not upto date with the what is present in CRD.

joelanford:
You have the all clear from me to merge. We can also capture Varsha's sample update in a follow up as well as far as I'm concerned.

joelanford:
I've captured the follow-ups in the epic issue. Adding this one to the queue now.

PR #625, opened 2024-02-09
✨  Integrate App with Extension reconciler

varshaprasad96:
# Description
This PR introduces support for installing App CR with extension reconciler. 

Each of the TODOs below, will have a separate issue created:
- Add e2e tests for extension reconciler
- Remove the use of resolver, and instead work on an alternative that identifies if the same bundle is present in the namespace or not and if cluster scoped APIs already exist on cluster. 
- Improve status reporting based on App's statuses. 
- Introduce support for registryV1 bundles through YTT templating mechanism. 

## Reviewer Checklist

- [ ] API Go Documentation


varshaprasad96:
/hold

Based on top of #598

internal/controllers/extension_controller.go
```diff
@@ -0,0 +1,378 @@
...
+
+	// TODO: kapp-controller integration
+	// * lookup the bundle in the solution/selection that corresponds to the Extension's desired Source
+	// * set the status of the Extension based on the respective deployed application status conditions.
+
+	// Now we can set the Resolved Condition, and the resolvedBundleSource field to the bundle.Image value.
+	ext.Status.ResolvedBundleResource = bundle.Image
+	setResolvedStatusConditionSuccess(&ext.Status.Conditions, fmt.Sprintf("resolved to %q", bundle.Image), ext.GetGeneration())
+
+	// TODO: Question - Should we set the deprecation statuses after we have successfully resolved instead of after a successful installation?
```
everettraven, line 187: This is a question I had in the past that I don't think was fully answered. I ended up leaving it after a successful installation because if the installation fails and the previous installation is _not_ deprecated we shouldn't mark the Extension as deprecated. IMO once a deprecated extension bundle is installed that is when the deprecation status should be set.

cmd/manager/main.go
```diff
@@ -147,3 +162,23 @@ func main() {
 		os.Exit(1)
 	}
 }
+
+// HasKappApis checks whether the cluster has Kapp APIs installed in the cluster.
+// This does not guarantee that the controller is present to reconcile the App CRs.
+func HasKappApis(config *rest.Config) (bool, error) {
```
joelanford, line 168: Nit: Doesn't really matter in `main` package since nothing can import a main package, but I'd suggest making this unexported to avoid confusion.

```suggestion
func hasKappApis(config *rest.Config) (bool, error) {
```

internal/controllers/extension_controller.go
```diff
@@ -116,29 +137,101 @@ func (r *ExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alpha1.Ext
 		return ctrl.Result{}, nil
 	}
 
+	if !r.HasKappApis {
+		setInstalledStatusConditionFailed(&ext.Status.Conditions, errkappAPIUnavailable.Error(), ext.GetGeneration())
+
+		ext.Status.ResolvedBundleResource = ""
+		setResolvedStatusConditionUnknown(&ext.Status.Conditions, "kapp apis are unavailable", ext.GetGeneration())
```
joelanford, line 144: Technically, we _could_ resolve a bundle, figure out the correct deprecation condition values, and populate everything "normally". And then only the `Installed` condition would be `False` with this message.

At a minimum, I would suggest moving this below the `paused` check.

internal/controllers/extension_controller.go
```diff
@@ -156,10 +249,161 @@ func (r *ExtensionReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+}
+
+func (r *ExtensionReconciler) ensureApp(ctx context.Context, desiredApp *unstructured.Unstructured) error {
+	existingApp, err := r.existingAppUnstructured(ctx, desiredApp.GetName(), desiredApp.GetNamespace())
+	if client.IgnoreNotFound(err) != nil {
+		return err
+	}
+
+	// If the existing BD already has everything that the desired BD has, no need to contact the API server.
+	// Make sure the status of the existingBD from the server is as expected.
```
joelanford, line 383:
```suggestion
	// If the existing App already has everything that the desired App has, no need to contact the API server.
	// Make sure the status of the existingApp from the server is as expected.
```

internal/controllers/extension_controller.go
```diff
@@ -116,29 +137,101 @@ func (r *ExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alpha1.Ext
...
 	}
 
-	// TODO: kapp-controller integration
-	// * gather variables for resolution
-	// * perform resolution
-	// * lookup the bundle in the solution/selection that corresponds to the Extension's desired Source
-	// * set the status of the Extension based on the respective deployed application status conditions.
+	// TODO: Instead of using a SAT solver and generating variables based on installed and
+	// available bundles, work on a filtering mechanism to evaluate if the bundle and its
+	// APIs are installed on a cluster or not. Feature tracked in a separate issue.
```
joelanford, line 159: We don't need to evaluate if the bundle and its APIs are already installed on the cluster.

For now, the "resolution" without deppy would be:
1. Get package, channel, bundle info from catalogd for `ext.Spec.Source.Package.Name`
2. Set `candidates` to all the bundles in the package.
3. If `ext.Spec.Source.Package.Channel` is set, remove candidates unless they have an `entries[].name` in the named channel.
4. If `ext.Spec.Source.Package.Version` is set, remove candidates unless their version is in the specified range.
5. If an `App` already exists for the Extension, remove candidates that are not the existing installed bundle or a successor of the existing installed bundle. (Ideally this would also account for UpgradeConstraintPolicy and the ForceSemver feature gate).
6.  If no candidates remain, "resolution" fails
7. Sort the remaining candidates by semver, and pick the highest.

So this is all just "grab from the catalog(s) and filter based on what's in the spec and what is installed"

Feel free to capture in a follow-up!

internal/controllers/extension_controller.go
```diff
@@ -156,10 +249,161 @@ func (r *ExtensionReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		*desiredApp = *existingApp
+		return nil
+	}
+
+	return r.Client.Patch(ctx, desiredApp, client.Apply, client.ForceOwnership, client.FieldOwner("operator-controller"))
+}
+
+func (r *ExtensionReconciler) existingAppUnstructured(ctx context.Context, name, namespace string) (*unstructured.Unstructured, error) {
+	existingApp := &kappctrlv1alpha1.App{}
+	err := r.Client.Get(ctx, types.NamespacedName{Name: name, Namespace: namespace}, existingApp)
```
joelanford, line 394: I recall that the controller-runtime client has some potentially unexpected default behavior around unstructured objects (iirc, the unstructured client is uncached by default).

Looks like we setup a structured informer for Apps in like 252.  Is it possible to fetch structured and convert to unstructured without breaking the comparison logic? Alternatively, could we setup the watch to use unstructured Apps and configure the client to enable caching of unstructured Apps?

internal/controllers/extension_controller.go
```diff
@@ -182,3 +426,89 @@ func extensionRequestsForCatalog(c client.Reader, logger logr.Logger) handler.Ma
...
+		return nil, err
+	}
+
+	extensionList := ocv1alpha1.ExtensionList{}
+	if err := r.Client.List(ctx, &extensionList); err != nil {
+		return nil, err
+	}
+
+	// Filter apps belogining to current namespace only to create install variables?
+	// This would ignore cluster scoped APIs installed from other Apps.
```
joelanford, line 494: I think we're going to let kapp-controller/kapp logic enforce the "an object can only be owned by one extension rule" because kapp already has that rule for "an object can only be owned by one app"

joelanford (commented):
This PR adds a bunch of resolution code that I don't think we'll want in the long run.

Instead of bickering over it in this PR, what do you think about going ahead with a simple (as simple as you want it for now) filter/sort/pick highest approach?

I'd propose that in this PR, the simple filtering could just be "give me the highest version for the package is that is in the specified channel and range".

And then we could follow-up with a separate PR to handle the various upgrade concerns.

internal/controllers/extension_controller.go
varshaprasad96: Leaving it to be after successful installation for now.

internal/controllers/extension_controller.go
varshaprasad96: I've captured this for a follow up #654 and done basic filtering based on version, package and channel. Meanwhile, looking into the code, realized that we do have helpers in place to filter out catalog bundles based on relevant info from extension that currently deppy uses to create variables. We can re-use them to make filtering cleaner.

internal/controllers/extension_controller.go
varshaprasad96: I had intentionally put the check above resolution thinking that if applier is not in place to install bundle contents, then resolution is of no use. But what you said is also right. For now, I've put it below the `paused` check. Will revisit to also set the deprecation status after resolution instead of installation.

varshaprasad96:
Surprised that rukpak is breaking the tilt tests, which is not related to this PR at all.

internal/controllers/extension_controller.go
joelanford: Nevermind! I completely missed this being a structured Get converted to unstructured. All good.

ankitathomas (commented):
Apart from a couple of nits, looks good!

internal/controllers/extension_controller.go
```diff
@@ -156,10 +227,161 @@ func (r *ExtensionReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+				Status:             metav1.ConditionTrue,
+				Message:            deprecation.Message,
+				ObservedGeneration: ext.Generation,
+			})
+		}
+
+		deprecationMessages = append(deprecationMessages, deprecation.Message)
+	}
+
+	if len(deprecationMessages) > 0 {
```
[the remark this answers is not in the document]
varshaprasad96: This was taken from the ClusterExtension reconciler. But iiuc, each individual deprecation message correspond to a particular type, and then there is the overall `TypeDeprecated` condition that aggregates all the messages together (though I'm not particularly sure of the reason behind that).

internal/controllers/extension_controller.go
```diff
@@ -123,22 +147,69 @@ func (r *ExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alpha1.Ext
 		return ctrl.Result{}, nil
 	}
 
-	// TODO: kapp-controller integration
-	// * gather variables for resolution
-	// * perform resolution
-	// * lookup the bundle in the solution/selection that corresponds to the Extension's desired Source
-	// * set the status of the Extension based on the respective deployed application status conditions.
+	if !r.HasKappApis {
```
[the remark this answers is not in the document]
varshaprasad96: No, this is evaluated only once when the controller starts. This is intentional since we also start the informers for watching App resources when the controller starts. As of today, there is no way for controller-runtime to dynamically add watches for resource types (though I remember some work on stopping and starting controllers was on the way).

gallettilance (dismissed):
/lgtm

Seems like a great jumping off point! Just to sanity check I understood things properly. It looks like:

1. If the kapp apis are installed after olm is, olm won't be aware of them unless you restart the pod manually
2. We don't have a good way of knowing if the kapp apis will be reconciled
3. The only thing this extension reconciler does is
    - install plain bundles
    - map these manifests to an app cr
    - relay status of app cr back to extension api

Let me know if I misunderstood anything / if there's more I didn't see.

internal/controllers/extension_controller.go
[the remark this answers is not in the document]
joelanford: Reasoning here was that different clients will have different concerns.

One set just cares about: "is this deprecated or not?"
Another set might care about the specific kind of deprecation.

So we decided to just include the individuals and a rollup.

internal/controllers/extension_controller.go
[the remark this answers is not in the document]
joelanford: It is possible to add watches for resource types after the fact, but I would call it overkill.

In fact, I was going to suggest that maybe we don't even need this check since we have a feature gate. If someone enables the feature gate without installing kapp-controller, that a somewhat obvious configuration issue.

internal/controllers/extension_controller.go
```diff
@@ -123,22 +147,69 @@ func (r *ExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alpha1.Ext
...
+	}
+
+	// TODO: Improve the resolution logic.
+	bundle, err := r.resolve(ctx, *ext)
+	if err != nil {
+		ext.Status.InstalledBundleResource = ""
+		setInstalledStatusConditionUnknown(&ext.Status.Conditions, "installation has not been attempted as resolution failed", ext.GetGeneration())
+		ext.Status.ResolvedBundleResource = ""
+		setResolvedStatusConditionFailed(&ext.Status.Conditions, err.Error(), ext.GetGeneration())
+		setDeprecationStatusesUnknown(&ext.Status.Conditions, "deprecation checks have not been attempted as resolution failed", ext.GetGeneration())
```
joelanford, line 167: Might not be possible based on the way bundle provider works, but in theory we should be able to determine package and channel deprecation status even if bundle selection fails.

I think #599 helps address this, though I haven't reviewed that yet.

joelanford: It would be good to capture this as a non-blocking follow-up improvement.

internal/controllers/extension_controller.go
```diff
@@ -18,30 +18,54 @@ package controllers
...
+	BundleProvider BundleProvider
+	Resolver       *solver.Solver
+	HasKappApis    bool
 }
 
+var errkappAPIUnavailable = errors.New("kapp-controller apis unavailable on cluster")
+
 //+kubebuilder:rbac:groups=olm.operatorframework.io,resources=extensions,verbs=get;list;watch;create;update;patch;delete
-//+kubebuilder:rbac:groups=olm.operatorframework.io,resources=extensions/status,verbs=update;patch
+//+kubebuilder:rbac:groups=olm.operatorframework.io,resources=extensions/status,verbs=get;update;patch
```
joelanford, line 66: There's no such thing as the `get` verb on the `/status` subresource. I removed this in one of the Extension API introduction followups. I'm guessing this was an easy thing to miss in the rebase.

internal/controllers/extension_controller.go
```diff
@@ -18,30 +18,54 @@ package controllers
...
+	"github.com/operator-framework/operator-controller/internal/catalogmetadata"
+	catalogfilter "github.com/operator-framework/operator-controller/internal/catalogmetadata/filter"
 	"github.com/operator-framework/operator-controller/pkg/features"
 )
 
 // ExtensionReconciler reconciles a Extension object
 type ExtensionReconciler struct {
 	client.Client
+	BundleProvider BundleProvider
+	Resolver       *solver.Solver
```
joelanford, line 59: Delete the `Resolver` field now that you are filtering directly?

internal/controllers/extension_controller.go
```diff
@@ -123,22 +147,69 @@ func (r *ExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alpha1.Ext
...
 
-	// TODO: kapp-controller integration
-	// * gather variables for resolution
-	// * perform resolution
-	// * lookup the bundle in the solution/selection that corresponds to the Extension's desired Source
-	// * set the status of the Extension based on the respective deployed application status conditions.
+	if !r.HasKappApis {
+		setInstalledStatusConditionFailed(&ext.Status.Conditions, errkappAPIUnavailable.Error(), ext.GetGeneration())
+
+		ext.Status.ResolvedBundleResource = ""
```
joelanford, line 153: Also set `ext.Status.InstalledBundleResource` to empty string?

internal/controllers/extension_controller.go
```diff
@@ -123,22 +147,69 @@ func (r *ExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alpha1.Ext
...
+		setInstalledStatusConditionUnknown(&ext.Status.Conditions, fmt.Sprintf("bundle type %s not supported currently", mediaType), ext.GetGeneration())
+		setDeprecationStatusesUnknown(&ext.Status.Conditions, "deprecation checks have not been attempted as installation has failed", ext.GetGeneration())
+		return ctrl.Result{}, nil
+	}
+
+	app := r.GenerateExpectedApp(*ext, bundle.Image)
+	if err := r.ensureApp(ctx, app); err != nil {
+		// originally Reason: ocv1alpha1.ReasonInstallationFailed
+		ext.Status.InstalledBundleResource = ""
+		setInstalledStatusConditionFailed(&ext.Status.Conditions, err.Error(), ext.GetGeneration())
```
joelanford, line 197: What if a previous `App` _is_ installed? It seems like we should be able to keep this status reflecting the truth of what's actually installed? This can be a follow-up.

internal/controllers/extension_controller.go
```diff
@@ -156,10 +227,161 @@ func (r *ExtensionReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&ocv1alpha1.Extension{}).
+		Owns(&kappctrlv1alpha1.App{}).
 		Watches(&catalogd.Catalog{}, handler.EnqueueRequestsFromMapFunc(extensionRequestsForCatalog(mgr.GetClient(), mgr.GetLogger()))).
 		Complete(r)
 }
 
+// TODO: follow up with mapping of all the available App statuses: https://github.com/carvel-dev/kapp-controller/blob/855063edee53315811a13ee8d5df1431ba258ede/pkg/apis/kappctrl/v1alpha1/status.go#L28-L35
+// mapAppStatusToInstalledCondition currently maps only the installed condition.
+func mapAppStatusToInstalledCondition(existingApp *kappctrlv1alpha1.App, ext *ocv1alpha1.Extension, bundleImage string) {
```
joelanford, line 237: We may want another `Extension` condition type as well. I could see a `Progressing` type that is populated based on App conditions when the App is not `ReconcileSucceeded`.

This could be part of the followup you mention in the TODO comment.

internal/controllers/extension_controller.go
```diff
@@ -156,10 +227,161 @@ func (r *ExtensionReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+			&ext.Status.Conditions,
+			appReady.Message,
+			ext.GetGeneration(),
+		)
+		return
+	}
+
+	// InstalledBundleResource this should be converted into a slice as App allows fetching
+	// from multiple sources.
+	ext.Status.InstalledBundleResource = bundleImage
```
joelanford, line 257: This is something we should definitely discuss more. There might be a few ways to handle this. Perhaps `InstalledBundleResource` is not the correct thing to capture anyway.

It could be that the correct thing to capture is something a bit more general (e.g. `.name` of the `olm.bundle` blob in the catalog).

This is the kind of thing that would be a good discussion topic for our WG sync or office hours.

internal/controllers/extension_controller.go
varshaprasad96: Yes, this is a part of #655, we need additional condition types to track - Reconciling, ReconcileFailed, Deleting, DeletingFailed statuses as provided in the App. Created a separate issue to discuss more about them.

internal/controllers/extension_controller.go
varshaprasad96: Right, I'll create a separate issue to follow this up and dig deeper into the use of `InstalledBundledResource` in general. It's definitely good to have, but I'm curious on how it is being consumed externally. Captured in: https://github.com/operator-framework/operator-controller/issues/672

internal/controllers/extension_controller.go
varshaprasad96: If existing app with the same spec exists, we will be returning with a nil error after doing the equality check: https://github.com/operator-framework/operator-controller/pull/625/files#diff-c04822d94b6cb8ff190859213d75fa6cd5c8b7c5284aaa2cf93cd504f727f5a9R363. After which we map the status of App we return from `ensureApp` - which is going to be either the existing one or the patched one.

internal/controllers/extension_controller.go
varshaprasad96: My bad! Yeah, missed it.

internal/controllers/extension_controller.go
joelanford: But if we have a new spec (e.g. we're doing an upgrade), and we patch the `App`, does the `App` status have a dual indication of:
- Still have this old thing active
- Progressing on rolling out the new thing

It would be nice, for example, that if something failed during one of (fetch, template, preflight) we could tell Extension users that the previous version is still installed, but this upgrade needs some attention.

joelanford (approved):
Awesome PR @varshaprasad96! Thanks!

Did you ever figure out what was going on with that tilt job? I wonder if there's something that's just broken about it because of the passage of time and some non-pinned configuration of something?

internal/controllers/extension_controller.go
varshaprasad96: > we could tell Extension users that the previous version is still installed, but this upgrade needs some attention.

Upgrade scenario needs to be verified with kapp-controller. I think the App status still reflects "Reconciling", I don't see anything specific related to it in their status. Will create a follow up to look into it.

varshaprasad96:
> Did you ever figure out what was going on with that tilt job? I wonder if there's something that's just broken about it because of the passage of time and some non-pinned configuration of something?

https://github.com/operator-framework/rukpak/pull/830 should fix it. Though I _think_ we need a new release, because `Tiltfile` is configured to pull in a release (from go mod) rather than the latest commit from main.

Issue #655, opened 2024-02-27
(Follow up): Integrate Kapp: Improve status reporting in Extension


varshaprasad96:
App CR currently defines an extensive list of statuses which are mentioned here: https://github.com/carvel-dev/kapp-controller/blob/3cb96a5ba384d4be2e580d9ca07c28ff9eee17f1/pkg/apis/kappctrl/v1alpha1/status.go#L28-L35. 

Extension needs to translate these conditions into its status, similar to how it is done here: https://github.com/operator-framework/operator-controller/blob/d51d908b1d999935642c0fa36178aa92fdaf10a8/internal/controllers/clusterextension_controller.go#L221. #625 currently maps only the reconcile succeeded status.

PR #683, opened 2024-03-07
✨improve Extension status conditions

ankitathomas:
# Description

Decouple the extension status conditions and introduce a new Progressing condition.

See https://github.com/operator-framework/operator-controller/issues/655

The installed condition now indicates whether the last installed App was successfully applied. 

The progressing condition indicates whether resolution, installation, upgrade or app reconciliation is in flight, with the reason and message on the condition providing more information.

## Reviewer Checklist

- [ ] API Go Documentation


internal/controllers/common_controller.go
```diff
@@ -118,3 +118,20 @@ func setDeprecationStatusesUnknown(conditions *[]metav1.Condition, message strin
 		})
 	}
 }
+
+// setInstalledStatusConditionFailed sets the installed status condition to failed.
+func setDeletingStatusConditionFailed(conditions *[]metav1.Condition, message string, generation int64) {
+	apimeta.SetStatusCondition(conditions, metav1.Condition{
+		Type:               ocv1alpha1.TypeDeleting,
+		Status:             metav1.ConditionFalse,
```
joelanford, line 126: This one is tricky. I think this combo of `TypeDeleting` and `ConditionFalse` would be interpreted as "this extension is not in the deleting state"

And I suppose that "DeletionFailed" could mean "I'm no longer actively in the deleting state because the deletion attempt failed".

But what happens at this point? Does `kapp-controller` keep retry-ing the deletion?

api/v1alpha1/clusterextension_types.go
```diff
@@ -77,6 +77,7 @@ const (
 	// TODO(user): add more Types, here and into init()
 	TypeInstalled = "Installed"
 	TypeResolved  = "Resolved"
+	TypeDeleting  = "Deleting"
```
joelanford, line 80: Could we have a more generic "Progressing" status instead? That way, whether we are in the process of installing, upgrading or deleting, we can set `Progressing=True` (and conversely `Progressing=False` would mean we're at a steady state).

internal/controllers/common_controller.go
varshaprasad96: Digging into the codebase (haven't verified yet by trying it out), looks like it doesn't. If the deletion fails, the process is still marked as completed (https://github.com/carvel-dev/kapp-controller/blob/d89101eb78b2c26a73a7afd8e8ae3111a6731dde/cli/pkg/kctrl/cmd/app/app_tailer.go#L52) and the watches are closed (https://github.com/carvel-dev/kapp-controller/blob/d89101eb78b2c26a73a7afd8e8ae3111a6731dde/cli/pkg/kctrl/cmd/app/app_tailer.go#L58). 

Going a bit further and looking into the PackageInstall reconciler to see what happens - looks like if deletion fails, and the App's status is conveys the same - the packageInstall just updates its status: https://github.com/carvel-dev/kapp-controller/blob/d89101eb78b2c26a73a7afd8e8ae3111a6731dde/pkg/packageinstall/packageinstall.go#L408-L428. 

This makes sense, because the reconcile period for the App CR is set by the user (and defaulted to 30sec if it's unset). 

This is why they seem to have the reconcileFailed condition which is set when Deleting is set to false.

api/v1alpha1/clusterextension_types.go
ankitathomas: Do we need a separate condition for this? I think we could achieve the same thing with `Installed`, with `Installed=True` being our steady state and `False` or `Unknown` for other states, with the `Reason` giving more details.

varshaprasad96: I'm worried if this may cause confusion - what would it mean when: Install is either fail/succeeded, with Progressing - True. How does the user know what action is being taken next? We would have to ensure that when Progressing is set to true, then other status are moved to Unknown.

internal/controllers/common_controller.go
ankitathomas: `kapp-controller` does seem to retry deletion on Apps indefinitely (essentially just a `kapp delete`).

I think you're right about the TypeDeleting being confusing, it reads more like another state of the Install/Reconcile/Progressing set of conditions. A single condition looks a lot cleaner.

api/v1alpha1/clusterextension_types.go
varshaprasad96: Wondering if we can instead re-name this condition to capture just the `Delete` action similar to how we handle others. 

1. `TypeDelete` Fail - means App's `kcv1alpha1.DeleteFailed` is True, `kcv1alpha1.Deleting` is false.
2. `TypeDelete` Unknown - means App's `kcv1alpha1.Deleting` is true or any other combination which we are unsure of what it represents. 

Looking into how Kapp is handling their status references - https://github.com/carvel-dev/kapp-controller/blob/d89101eb78b2c26a73a7afd8e8ae3111a6731dde/cli/pkg/kctrl/cmd/app/app.go#L39-L64 - All they are trying to map is Reconcile Fail/Succeeded/Progressing, Delete Progressing/Failed. So I'm hoping that the above should capture what we want.

internal/controllers/common_controller.go
ankitathomas: From what I can tell, `kapp-controller` does seem to keep retrying the App delete, though I'm not sure if there's anything to limit the retries: https://github.com/carvel-dev/kapp-controller/blob/d89101eb78b2c26a73a7afd8e8ae3111a6731dde/pkg/app/app_reconcile.go#L36

api/v1alpha1/clusterextension_types.go
joelanford: > I'm worried if this may cause confusion - what would it mean when: Install is either fail/succeeded, with Progressing - True. How does the user know what action is being taken next? 

We can use the `Reason` and `Message` to get into more specifics.

Taking a stab at enumerating various states:

<table>
<tr>
 <td><b>State</td>
 <td><b>Progressing</td>
 <td><b>Installed</td>
</tr>

<tr>
 <td>Initial install is in progress</td>
 <td>True<br>Reason=Installing<br>Message="installing $bundleID"</td>
 <td>False<br>Reason=NothingInstalled</td>
</tr>

<tr>
 <td>Initial install is failing with a non-terminal error (i.e. retries might resolve the error in the future)</td>
 <td>True<br>Reason=RetryingInstallation<br>Message="retrying installation of $bundleID due to error: $error"</td>
 <td>False<br>Reason=NothingInstalled</td>
</tr>

<tr>
 <td>Initial install failed with a terminal error (i.e. retries will never resolve in the future without some external change)</td>
 <td>False<br>Reason=InstallFailed<br>Message="failed installation of $bundleID: $error</td>
 <td>False<br>Reason=NothingInstalled</td>
</tr>

<tr>
 <td>Install failed, deletion is in progress</td>
 <td>True<br>Reason= Uninstalling <br>Message="uninstalling failed installation of $bundleID"</td>
 <td>False<br>Reason=NothingInstalled</td>
</tr>

<tr>
 <td>Initial install failed, deletion is failing</td>
 <td>True<br>Reason=RetryingUninstall<br>Message="retrying uninstall of $bundleID due to error: $error"</td>
 <td>False<br>Reason=NothingInstalled</td>
</tr>

<tr>
 <td>Initial install failed, deletion failed</td>
 <td>False<br>Reason=UninstallFailed<br>Message="failed uninstall of $bundleID: $error>"</td>
 <td>False<br>Reason=NothingInstalled</td>
</tr>

<tr>
 <td>Install/upgrade succeeded, steady state</td>
 <td>False<br>Reason=NothingToDo</td>
 <td>True<br>Reason=InstallSucceeded<br>Message="successfully installed $bundleID"</td>
</tr>

<tr>
 <td>Install succeeded, upgrade is in progress</td>
 <td>True<br>Reason=Upgrading<br>Message="upgrading to $newBundleID</td>
 <td>True<br>Reason=InstallSucceeded<br>Message="successfully installed $oldBundleID"</td>
</tr>

<tr>
 <td>Install succeeded, upgrade is failing with a non-terminal error</td>
 <td>True<br>Reason=RetryingUpgrade<br>Message="retrying upgrade of $newBundleID due to error: $error"</td>
 <td>True<br>Reason=InstallSucceeded<br>Message="successfully installed $oldBundleID"</td>
</tr>

<tr>
 <td>Install succeeded, upgrade failed with a terminal error</td>
 <td>False<br>Reason=UpgradeFailed<br>Message="failed upgrade to $newBundleID: $error"</td>
 <td>True<br>Reason=InstallSucceeded<br>Message="successfully installed $oldBundleID"</td>
</tr>

<tr>
 <td>Install succeeded, deletion is in progress</td>
 <td>True<br>Reason=Uninstalling<br>Message="uninstalling $bundleID"</td>
 <td>True<br>Reason=InstallSucceeded<br>Message="successfully installed $bundleID"</td>
</tr>

<tr>
 <td>Install succeeded, deletion is failing with a non-terminal error</td>
 <td>True<br>Reason=RetryingUninstall<br>Message="retrying uninstall of $bundleID due to error: $error"</td>
 <td>True<br>Reason=InstallSucceeded<br>Message="successfully installed $bundleID"</td>
</tr>

<tr>
 <td>Install succeeded, deletion is failing with a terminal error</td>
 <td>False<br>Reason=UninstallFailed<br>Message="failed uninstall of $bundleID: $error"</td>
 <td>True<br>Reason=InstallSucceeded<br>Message="successfully installed $bundleID"</td>
</tr>
</table>



Notice how `Progressing` and `Installed` are actually orthogonal. `Installed` does not depend on `Progressing` or vice-versa.

> We would have to ensure that when Progressing is set to true, then other statuses are moved to Unknown.

I don't think so. My idea and goal here is actually that a Progressing condition gives us a place to keep some status about what we're doing so that we don't have to wipe out some status about what we did before. 

For example, if someone installs v1, and then tries to upgrade to v2, but v2 fails the permissions preflight check, the actual state of the system is:
- v1 is still doing its thing untouched
- v2 failed the upgrade preflight check (which might resolve in the future if the admin reconfigures the permissions of the Extension/App's service account).

It would be really nice to be able to say "v1 is still installed and running, but here's why we can't upgrade to v2."

ankitathomas: Wouldn't a progressing/failing update/delete possibly impact the currently installed extension? If the idea is to have a separate condition specific to preflight checks, we may have to add more support for that separately from the `kapp-controller` App conditions.

joelanford: It might impact it, yes. Hopefully `App`'s status would tell us enough about the failure to help us know if the current installation was impacted or not.

This is probably a good discussion topic with the Carvel folks.

internal/controllers/common_controller.go
```diff
@@ -55,51 +88,106 @@ func setInstalledStatusConditionUnknown(conditions *[]metav1.Condition, message
...
-// setResolvedStatusConditionFailed sets the resolved status condition to failed.
-func setResolvedStatusConditionFailed(conditions *[]metav1.Condition, message string, generation int64) {
+// setProgressingStatusConditionFalse sets the progressing status condition to false, awaiting next reconcile.
+func setProgressingStatusConditionFalse(conditions *[]metav1.Condition, message string, generation int64) {
 	apimeta.SetStatusCondition(conditions, metav1.Condition{
-		Type:               ocv1alpha1.TypeResolved,
+		Type:               ocv1alpha1.TypeProgressing,
 		Status:             metav1.ConditionFalse,
-		Reason:             ocv1alpha1.ReasonResolutionFailed,
+		Reason:             ocv1alpha1.ReasonPending,
```
joelanford, line 96: If we're not progressing, doesn't that mean we're at steady state with nothing to do? `Pending` seems like it would indicate that we _are_ progressing and waiting on some action to occur.

internal/controllers/extension_controller.go
```diff
@@ -148,32 +147,35 @@ func (r *ExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alpha1.Ext
...
 	}
 
 	// Now we can set the Resolved Condition, and the resolvedBundleSource field to the bundle.Image value.
 	ext.Status.ResolvedBundleResource = bundle.Image
-	setResolvedStatusConditionSuccess(&ext.Status.Conditions, fmt.Sprintf("resolved to %q", bundle.Image), ext.GetGeneration())
 
 	mediaType, err := bundle.MediaType()
 	if err != nil {
-		setInstalledStatusConditionFailed(&ext.Status.Conditions, err.Error(), ext.GetGeneration())
+		if ext.Status.InstalledBundleResource == "" {
```
joelanford, line 173: Don't read our own status here to determine if there is an installed bundle. Instead, we need to `Get` the `App` that would be managed by this `Extension`.

If it does not exist:
- installedBundleResource = ""
- installed condition = False

If we encountered some other error:
- don't touch installedBundleResource?
- installed condition = Unknown

If it does exist:
- installedBundleResource = <the image ref on the app>
- installed condition = true

internal/controllers/extension_controller.go
```diff
@@ -148,32 +147,35 @@ func (r *ExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alpha1.Ext
...
 		setDeprecationStatusesUnknown(&ext.Status.Conditions, "kapp apis are unavailable", ext.GetGeneration())
 		return ctrl.Result{}, errkappAPIUnavailable
 	}
 
 	// TODO: Improve the resolution logic.
 	bundle, err := r.resolve(ctx, *ext)
 	if err != nil {
-		ext.Status.InstalledBundleResource = ""
-		setInstalledStatusConditionUnknown(&ext.Status.Conditions, "installation has not been attempted as resolution failed", ext.GetGeneration())
+		if ext.Status.InstalledBundleResource == "" {
```
joelanford, line 159: Same comment here about not reading our own status.

internal/controllers/extension_controller.go
```diff
@@ -148,32 +147,35 @@ func (r *ExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alpha1.Ext
...
+			setInstalledStatusConditionUnknown(&ext.Status.Conditions, "installation has not been attempted as resolution failed", ext.GetGeneration())
+		}
 		ext.Status.ResolvedBundleResource = ""
-		setResolvedStatusConditionFailed(&ext.Status.Conditions, err.Error(), ext.GetGeneration())
+		setProgressingStatusConditionResolutionFailed(&ext.Status.Conditions, "resolution failed: "+err.Error(), ext.GetGeneration())
 		setDeprecationStatusesUnknown(&ext.Status.Conditions, "deprecation checks have not been attempted as resolution failed", ext.GetGeneration())
```
joelanford, line 164: I know you didn't touch this part of the code, but we should be able to determine deprecation status:
- for package and channel if we have a catalog
- for bundle if we have an installed bundle

Can you make a follow-up issue to improve this?

internal/controllers/extension_controller.go
```diff
@@ -148,32 +147,35 @@ func (r *ExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alpha1.Ext
...
 
 	mediaType, err := bundle.MediaType()
 	if err != nil {
-		setInstalledStatusConditionFailed(&ext.Status.Conditions, err.Error(), ext.GetGeneration())
+		if ext.Status.InstalledBundleResource == "" {
+			// Set the TypeInstalled condition to Unknown to indicate that the resolution
+			// hasn't been attempted yet, due to the spec being invalid.
+			setInstalledStatusConditionUnknown(&ext.Status.Conditions, err.Error(), ext.GetGeneration())
+		}
+		setProgressingStatusConditionUnsupportedOrInvalidBundle(&ext.Status.Conditions, "failed to read bundle mediaType: "+err.Error(), ext.GetGeneration())
```
joelanford, line 178: It is generally prefered to use `fmt.Sprintf` over string concatenation via `+`.

```suggestion
		setProgressingStatusConditionUnsupportedOrInvalidBundle(&ext.Status.Conditions, fmt.Sprintf("failed to read bundle mediaType: %v", err), ext.GetGeneration())
```

internal/controllers/extension_controller.go
```diff
@@ -183,8 +185,10 @@ func (r *ExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alpha1.Ext
 	if mediaType != catalogmetadata.MediaTypePlain {
 		// Set the TypeInstalled condition to Unknown to indicate that the resolution
 		// hasn't been attempted yet, due to the spec being invalid.
-		ext.Status.InstalledBundleResource = ""
-		setInstalledStatusConditionUnknown(&ext.Status.Conditions, fmt.Sprintf("bundle type %s not supported currently", mediaType), ext.GetGeneration())
+		if ext.Status.InstalledBundleResource == "" {
+			setInstalledStatusConditionUnknown(&ext.Status.Conditions, fmt.Sprintf("bundle type %s not supported currently", mediaType), ext.GetGeneration())
+		}
```
joelanford, line 190: I would only be expecting changes to the `InstalledBundleResource` and the `Installed` condition based on `Get`-ing the existing `App`.

Resolution and Installation are orthogonal. A failure computing a resolution shouldn't impact our ability to assess the installed state and vice versa.

internal/controllers/extension_controller.go
```diff
@@ -194,6 +198,8 @@ func (r *ExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alpha1.Ext
 		// originally Reason: ocv1alpha1.ReasonInstallationFailed
 		ext.Status.InstalledBundleResource = ""
 		setInstalledStatusConditionFailed(&ext.Status.Conditions, err.Error(), ext.GetGeneration())
+		setProgressingStatusConditionFalse(&ext.Status.Conditions, "app install failed", ext.GetGeneration())
```
joelanford, line 201: This seems like a case where I'd say `Progressing` is still `True` because we encountered a failure but are going to retry.

internal/controllers/extension_controller.go
```diff
@@ -203,11 +209,15 @@ func (r *ExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alpha1.Ext
 		// originally Reason: ocv1alpha1.ReasonInstallationStatusUnknown
 		ext.Status.InstalledBundleResource = ""
 		setInstalledStatusConditionUnknown(&ext.Status.Conditions, err.Error(), ext.GetGeneration())
+		setProgressingStatusConditionFalse(&ext.Status.Conditions, "app install failed", ext.GetGeneration())
```
joelanford, line 212: Same comment here as above. A rule of thumb is: if we're going to end up returning an error from `Reconcile`, that means we're going to retry, which means we're still progressing.

internal/controllers/extension_controller.go
```diff
@@ -231,30 +241,61 @@ func (r *ExtensionReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
-// TODO: follow up with mapping of all the available App statuses: https://github.com/carvel-dev/kapp-controller/blob/855063edee53315811a13ee8d5df1431ba258ede/pkg/apis/kappctrl/v1alpha1/status.go#L28-L35
-// mapAppStatusToInstalledCondition currently maps only the installed condition.
-func mapAppStatusToInstalledCondition(existingApp *kappctrlv1alpha1.App, ext *ocv1alpha1.Extension, bundleImage string) {
-	appReady := findStatusCondition(existingApp.Status.GenericStatus.Conditions, kappctrlv1alpha1.ReconcileSucceeded)
-	if appReady == nil {
-		ext.Status.InstalledBundleResource = ""
-		setInstalledStatusConditionUnknown(&ext.Status.Conditions, "install status unknown", ext.Generation)
+// mapAppStatusToCondition maps the reconciling/deleting App conditions to the installed/deleting conditions on the Extension.
+func MapAppStatusToCondition(existingApp *kappctrlv1alpha1.App, ext *ocv1alpha1.Extension) {
+	// Note: App.Status.Inspect errors are never surfaced to App conditions, so are currently ignored when determining App status.
```
varshaprasad96, line 246: iirc, also `Status.Inspect` appears if the knob to inspect resources is set to true in the spec. We would have to probably cover this aspect better with the health checks.

internal/controllers/common_controller.go
everettraven: +1. If it is assumed that when `Progressing` is `False` we are in a stable state maybe `Stable` as the reason is sufficient?

internal/controllers/extension_controller.go
```diff
@@ -231,30 +241,61 @@ func (r *ExtensionReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
-			&ext.Status.Conditions,
-			appReady.Message,
-			ext.GetGeneration(),
-		)
+	if existingApp == nil {
+		setInstalledStatusConditionUnknown(&ext.Status.Conditions, message, ext.Generation)
 		return
 	}
+	message = existingApp.Status.FriendlyDescription
+	if strings.Contains(message, "Error (see .status.usefulErrorMessage for details)") || len(message) == 0 {
```
everettraven, line 257: Nit: Switch the positioning of these conditionals. Switching them makes it so that `strings.Contains` won't be called if the length of the message is already `0`.

internal/controllers/extension_controller.go
```diff
@@ -129,10 +129,9 @@ func (r *ExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alpha1.Ext
 		// hasn't been attempted yet, due to the spec being invalid.
 		ext.Status.InstalledBundleResource = ""
 		setInstalledStatusConditionUnknown(&ext.Status.Conditions, "extension feature is disabled", ext.GetGeneration())
-		// Set the TypeResolved condition to Unknown to indicate that the resolution
-		// hasn't been attempted yet, due to the spec being invalid.
+
 		ext.Status.ResolvedBundleResource = ""
-		setResolvedStatusConditionUnknown(&ext.Status.Conditions, "extension feature is disabled", ext.GetGeneration())
+		setProgressingStatusConditionUnknown(&ext.Status.Conditions, "extension feature is disabled", ext.GetGeneration())
```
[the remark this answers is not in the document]
joelanford: +1, but as I mentioned on our call, I'm also personally okay with following the existing pattern and then working in a separate followup to reorganize the code around the fact that the various condition types can be calculated and determined in isolation from each other.

internal/controllers/extension_controller.go
```diff
@@ -231,30 +241,61 @@ func (r *ExtensionReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	}
+	for _, cond := range orderedAppStatuses {
+		if c := findStatusCondition(existingApp.Status.GenericStatus.Conditions, cond); c != nil && c.Status == corev1.ConditionTrue {
+			if len(message) == 0 {
+				message = c.Message
+			}
+			appStatusMapFn[cond](&ext.Status.Conditions, message, ext.Generation)
+			return
+		}
+	}
```
everettraven, line 283: Instead of looping through these `orderedAppStatuses` and then doing a loop through the conditions to find them, could we just loop through all the App conditions, check if there is a status mapping function, and if there is one call it?

internal/controllers/extension_controller.go
```diff
@@ -162,21 +164,29 @@ func (r *ExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alpha1.Ext
...
+
 	mediaType, err := bundle.MediaType()
 	if err != nil {
-		setInstalledStatusConditionFailed(&ext.Status.Conditions, err.Error(), ext.GetGeneration())
-		setDeprecationStatusesUnknown(&ext.Status.Conditions, "deprecation checks have not been attempted as installation has failed", ext.GetGeneration())
+		if c := apimeta.FindStatusCondition(ext.Status.Conditions, ocv1alpha1.TypeInstalled); c == nil {
+			ext.Status.InstalledBundle = nil
+			setInstalledStatusConditionFailed(&ext.Status.Conditions, fmt.Sprintf("failed to read bundle mediaType: %v", err), ext.GetGeneration())
+		}
+		setProgressingStatusConditionFailed(&ext.Status.Conditions, fmt.Sprintf("failed to read bundle mediaType: %v", err), ext.GetGeneration())
```
varshaprasad96, line 176: A little confusion here: `setProgressingStatusConditionFailed` is setting Progressing too true. As discussed [here](https://github.com/operator-framework/operator-controller/pull/683#discussion_r1540999472), if there is an error and we are not able to proceed further, then though an error, we are in a stable state with no other attempt.

internal/controllers/extension_controller.go
```diff
@@ -162,21 +164,29 @@ func (r *ExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alpha1.Ext
...
-		ext.Status.InstalledBundle = nil
-		setInstalledStatusConditionUnknown(&ext.Status.Conditions, fmt.Sprintf("bundle type %s not supported currently", mediaType), ext.GetGeneration())
-		setDeprecationStatusesUnknown(&ext.Status.Conditions, "deprecation checks have not been attempted as installation has failed", ext.GetGeneration())
+		if c := apimeta.FindStatusCondition(ext.Status.Conditions, ocv1alpha1.TypeInstalled); c == nil {
+			// Set the TypeInstalled condition to Failed to indicate that the resolution
+			// hasn't been attempted yet, due to the spec being invalid.
+			ext.Status.InstalledBundle = nil
+			setInstalledStatusConditionFailed(&ext.Status.Conditions, fmt.Sprintf("bundle type %s not supported currently", mediaType), ext.GetGeneration())
+		}
+		setProgressingStatusConditionFailed(&ext.Status.Conditions, fmt.Sprintf("bundle type %s not supported currently", mediaType), ext.GetGeneration())
```
varshaprasad96, line 189: The same question as before: Shouldn't progressing condition be false in case of an error indicating that we are in a stable state of failure.

internal/controllers/extension_controller.go
ankitathomas: That makes sense, updated the PR so that we're only ever progressing if we're potentially unstable.

internal/controllers/common_controller.go
```diff
@@ -117,3 +106,36 @@ func setDeprecationStatusesUnknown(conditions *[]metav1.Condition, message strin
 		})
 	}
 }
+
+// setProgressingStatusConditionSuccess sets the progressing status condition to false for a successful install or upgrade.
+func setProgressingStatusConditionSuccess(conditions *[]metav1.Condition, message string, generation int64) {
+	apimeta.SetStatusCondition(conditions, metav1.Condition{
+		Type:               ocv1alpha1.TypeProgressing,
+		Status:             metav1.ConditionFalse,
+		Reason:             ocv1alpha1.ReasonSuccess,
```
joelanford, line 115: Nit: something like `ReasonReachedDesiredIntent` seems more apt for the `False` status of the progressing condition. WDYT?

joelanford (approved):
Nice job @ankitathomas! Just one nit. Once you decide on that, let's merge it!

internal/controllers/extension_controller.go
```diff
@@ -162,35 +164,47 @@ func (r *ExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alpha1.Ext
 	ext.Status.ResolvedBundle = bundleMetadataFor(bundle)
 	setResolvedStatusConditionSuccess(&ext.Status.Conditions, fmt.Sprintf("resolved to %q", bundle.Image), ext.GetGeneration())
 
+	// Populate the deprecation status using the resolved bundle
+	SetDeprecationStatusInExtension(ext, bundle)
+
```
everettraven, line 169: Based on the current state of the logic in determining deprecation status, and without increasing scope to include modifications to it, I think this shouldn't be called until installation has succeeded.

There is some nuance where we theoretically could set the package and/or channel deprecation status after resolution but IMO any changes to the existing deprecation logic is out of scope of this PR.

everettraven (approved):
Overall looks good to me. I had one nit but I'm OK merging this without that being addressed since we are also shortly planning on doing a "soft removal" of the Extension API while we focus on the ClusterExtension API

internal/controllers/extension_controller.go
```diff
@@ -228,28 +248,37 @@ func (r *ExtensionReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+				message = c.Message
+			}
+			appStatusMapFn[cond](&ext.Status.Conditions, fmt.Sprintf("App %s: %s", c.Type, message), ext.Generation)
+			return
+		}
+	}
+	if len(message) == 0 {
+		message = "waiting for app"
+	}
+	setProgressingStatusConditionProgressing(&ext.Status.Conditions, message, ext.Generation)
```
everettraven, line 281: I'm not a huge fan of looping through all the status conditions 5 times but since we are moving away from building on top of the App CR and this is an optimization that could be made in the future I'm fine with it for now

everettraven (approved):

PR #689, opened 2024-03-11
✨ Admission webhook to restrict usage of kapp-controller APIs

m1kola:
# Description

We are not prepared to support the full surface area of the Carvel APIs at the moment and want to put admission webhook in place to make it harder to use Carvel APIs directly when installed together with OLM.

The plan is to only accept create, update and delete operations from operator-controller and kapp-controller service accounts. This will apply to all resources in the following groups:
* `internal.packaging.carvel.dev`
* `kappctrl.k14s.io`
* `packaging.carvel.dev`

## Reviewer Checklist

- [ ] API Go Documentation


m1kola:
We decided to not restrict use of Carvel APIs upstream for now. Closing

PR #690, opened 2024-03-11
✨ [Add] Unit tests for Extension reconciler

varshaprasad96:
# Description
This PR:
- Removes the additional `HasKappAPI` test, since the feature is anyway protected using featureGate. This also helps in running tests easily.
- Adds unit tests for Extension controller. 

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/controllers/extension_controller_test.go
+++ b/internal/controllers/extension_controller_test.go
@@ -2,21 +2,33 @@ package controllers_test
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	kappctrlv1alpha1 "github.com/vmware-tanzu/carvel-kapp-controller/pkg/apis/kappctrl/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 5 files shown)


joelanford (commented):
It seems like there may be an opportunity to collapse some of the test code because it looks like a decent chunk is repetitive between different tests.

You might consider a table-driven set of tests, where a test case could contain arbitrary setup and assertion functions that could be implemented per test case.

internal/controllers/extension_controller_test.go
```diff
@@ -92,6 +86,931 @@ func TestExtensionReconcile(t *testing.T) {
 	}
 }
 
+func TestExtensionNonExistentPackage(t *testing.T) {
+	defer featuregatetesting.SetFeatureGateDuringTest(t, features.OperatorControllerFeatureGate, features.EnableExtensionAPI, true)()
```
dtfranz, line 90: These new tests looked like they had similar enough structure to potentially be done in [this kind of format](https://github.com/operator-framework/operator-controller/blob/38da6fc3c70c60d4da570333a3e5fc005192d44f/internal/catalogmetadata/types_test.go#L18). Seems to me like that might be more maintainable?

dtfranz: Oops, just realized this is a duplicate comment of Joe's [here](https://github.com/operator-framework/operator-controller/pull/690#pullrequestreview-1931627893).

perdasilva:
closing due to inactivity - please re-open if we still need this =D

Issue #612, opened 2024-02-01
Add extensible `values` field to Extension spec


oceanc80:
In scope:
- Registry+v1 bundles only
- watchNamespaces
- Stretch: `subscription.spec.config`

joelanford:
I implemented a proof of concept as part of https://github.com/operator-framework/operator-controller/pull/1418. We should revisit that PR when we have time to pick this issue up.

perdasilva:
replaced by #1750

Issue #593, opened 2024-01-23
[epic] Support registry+v1 bundle installation in SingleNamespace and OwnNamespace install modes


ncdc:
- [x] #1751
- [x] #1752
- [x] #1753
- [x] #1750
- [x] #612
- [x] #613
- [x] #614
- [x] #2268 
- [x] #2283
- [x] #2292

oceanc80:
Add ytt config
- This should use the same method as used in OLM v0 (annotation key @joelanford to provide specific key)
- Look into if watchNamespaces can be passed as parameters to ytt

LalatenduMohanty:
- We should figure-out the use-cases we want to solve with this epic.

joelanford:
I implemented a PoC that includes parameterization related to registry+v1 watch namespaces. See #1418.

oceanc80:
@thetechnick Do you know if the status of the remaining open issues here is correct?

Issue #381, opened 2023-08-31
[epic] ClusterExtension parameters passed through to templating engine


ncdc:
Define a way for the content author to specify what parameters are user-facing, along with a way to plumb them through from the ClusterExtension spec the templating engine

Example use case: I publish a bundle for my Widget app, and the only knobs I want users to be able to tweak are replica count and logging verbosity.

This either needs to accommodate v0 subscription config settings (e.g. env vars, node selector, etc.), or we need a different mechanism for that.

joelanford:
See https://github.com/operator-framework/operator-controller/issues/612

In order to keep watch namespace configuration out of the Extension API as a first class field, I think we should define our own schema/values that account for watch namespaces and subscription config, and put that in scope for the initial registry+v1 bundle support in Extension+App

tmshort:
This should also consider proxy information that is determine dynamically via environment.

joelanford:
I implemented a proof of concept as part of https://github.com/operator-framework/operator-controller/pull/1418. We should revisit that PR when we have time to pick this issue up.

anik120:
Fyi we have an RFC now for this: https://docs.google.com/document/d/18O4qBvu5I4WIJgo5KU1opyUKcrfgk64xsI3tyXxmVEU/edit?tab=t.0#heading=h.x3tfh25grvnv

the work for which has started https://github.com/operator-framework/operator-controller/pull/2454. 

The upcoming work is not being tracked in issues but I'm hoping to remember to come back here and update when the work is completed and the feature is available for anyone who's interested in this feature.

grokspawn:
Related to #2482

anik120:
Resolved by https://github.com/operator-framework/operator-controller/pull/2482

PR #692, opened 2024-03-12
📖 Add OLMv1 Overview doc

joelanford:
# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,22 @@
 # operator-controller
 The operator-controller is the central component of Operator Lifecycle Manager (OLM) v1. It extends Kubernetes with an API through which users can install extensions.
 
-## Description
-OLM v1 is the follow-up to OLM v0, located [here](https://github.com/operator-framework/operator-lifecycle-manager). It consists of four different components, including this one, which are as follows:
+
+## Mission
+
+OLM’s purpose is to provide APIs, controllers, and tooling that support the packaging, distribution, and lifecycling of Kubernetes extensions. It aims to align with Kubernetes designs and user assumptions, provide secure, high-quality, and predictable user experiences centered around declarative GitOps concepts, and to give cluster admins the minimal necessary controls to build their desired cluster architectures and to have ultimate control.
+
+## Overview
+
+OLM v1 is the follow-up to OLM v0, located [here](https://github.com/operator-framework/operator-lifecycle-manager).
+
+OLM v1 consists of four different components, including this one, which are as follows:
 * operator-controller
 * [deppy](https://github.com/operator-framework/deppy)
 * [rukpak](https://github.com/operator-framework/rukpak)
 * [catalogd](https://github.com/operator-framework/catalogd)
+
+For a more complete overview of OLM v1 and how it differs from OLM v0, see our [overview](./docs/olmv1_overview.md).
 
 ## Getting Started
 You’ll need a Kubernetes cluster to run against. You can use [KIND](https://sigs.k8s.io/kind) to get a local cluster for testing, or run against a remote cluster.
--- a/docs/olmv1_overview.md
+++ b/docs/olmv1_overview.md
@@ -1,0 +1,243 @@
+# OLMv1 Overview
+
+## What won't OLMv1 do that OLMv0 did?
+
+TL;DR: OLMv1 cannot feasibly support multi-tenancy or any feature that assumes multi-tenancy. All multi-tenancy features end up falling over because of the global API system of Kubernetes. While this short conclusion may be unsatisfying, the reasons are complex and intertwined.
+
+### Historical Context
+
+Nearly every active contributor in the Operator Framework project contributed to design explorations and prototypes over an entire year. For each of these design explorations, there are complex webs of features and assumptions that are necessary to understand the context that ultimately led to a conclusion of infeasibility.
+
+Here is a sampling of some of the ideas we explored:
+- [[WIP] OLM v1's approach to multi-tenancy](https://docs.google.com/document/d/1xTu7XadmqD61imJisjnP9A6k38_fiZQ8ThvZSDYszog/edit#heading=h.m19itc78n5rw)
+- [OLMv1 Multi-tenancy Brainstorming](https://docs.google.com/document/d/1ihFuJR9YS_GWW4_p3qjXu3WjvK0NIPIkt0qGixirQO8/edit#heading=h.vy9860qq1j01)
+
+### Watched namespaces cannot be configured in a first-class API
+
+OLMv1 will not have a first-class API for configuring the namespaces that a controller will watch.
+
+Kubernetes APIs are global. Kubernetes is designed with the assumption that a controller WILL reconcile an object no matter where it is in the cluster.
+
+However, Kubernetes does not assume that a controller will be successful when it reconciles an object.
+
+The Kubernetes design assumptions are:
+- CRDs and their controllers are trusted cluster extensions.
+- If an object for an API exists a controller WILL reconcile it, no matter where it is in the cluster.
+
+OLMv1 will make the same assumption that Kubernetes does and that users of Kubernetes APIs do. That is: If a user has RBAC to create an object in the cluster, they can expect that a controller exists that will reconcile that object. If this assumption does not hold, it will be considered a configuration issue, not an OLMv1 bug.
+
+This means that it is a best practice to implement and configure controllers to have cluster-wide permission to read and update the status of their primary APIs. It does not mean that a controller needs cluster-wide access to read/write secondary APIs. If a controller can update the status of its primary APIs, it can tell users when it lacks permission to act on secondary APIs.
+
+### Dependencies based on watched namespaces
+
+Since there will be no first-class support for configuration of watched namespaces, OLMv1 cannot resolve dependencies among bundles based on where controllers are watching.
+
+However, not all bundle constraints are based on dependencies among bundles from different packages. OLMv1 will be able to support constraints against cluster state. For example, OLMv1 could support a “kubernetesVersionRange” constraint that blocks installation of a bundle if the current kubernetes cluster version does not fall into the specified range.
+
+#### Background
+
+For packages that specify API-based dependencies, OLMv0’s dependency checker knows which controllers are watching which namespaces. While OLMv1 will have awareness of which APIs are present on a cluster (via the discovery API), it will not know which namespaces are being watched for reconciliation of those APIs. Therefore dependency resolution based solely on API availability would only work in cases where controllers are configured to watch all namespaces.
+
+For packages that specify package-based dependencies, OLMv0’s dependency checker again knows which controllers are watching which namespaces. This case is challenging for a variety of reasons:
+1. How would a dependency resolver know which extensions were installed (let alone which extensions were watching which namespaces)? If a user is running the resolver, they would be blind to an installed extension that is watching their namespace if they don’t have permission to list extensions in the installation namespace. If a controller is running the resolver, then it might leak information to a user about installed extensions that the user is not otherwise entitled to know.
+2. Even if (1) could be overcome, the lack of awareness of watched namespaces means that the resolver would have to make assumptions. If only one controller is installed, is it watching the right set of namespaces to meet the constraint? If multiple controllers are installed, are any of them watching the right set of namespaces? Without knowing the watched namespaces of the parent and child controllers, a correct dependency resolver implementation is not possible to implement.
+
+Note that regardless of the ability of OLMv1 to perform dependency resolution (now or in the future), OLMv1 will not automatically install a missing dependency when a user requests an operator. The primary reasoning is that OLMv1 will err on the side of predictability and cluster-administrator awareness.
+
+### "Watch namespace"-aware operator discoverability
+
+When operators add APIs to a cluster, these APIs are globally visible. As stated before, there is an assumption in this design that a controller will reconcile an object of that API anywhere it exists in the cluster.
+
+Therefore, the API discoverability story boils down to answering this question for the user: “What APIs do I have access to in a given namespace?” Fortunately, built-in APIs exist to answer this question: Kubernetes Discovery, SelfSubjectRulesReview (SSRR), and SelfSubjectAccessReview (SSAR).
+
+However, helping users discover which actual controllers will reconcile those APIs is not possible unless OLMv1 knows which namespaces those controllers are watching.
+
+Any solution here would be unaware of where a controller is actually watching and could only know “is there a controller installed that provides an implementation of this API?”. However even knowledge of a controller installation is not certain. Any user can use the discovery, SSRR, and SSAR. Not all users can list all Extensions (see [User discovery of “available” APIs](#user-discovery-of-available-apis)).
+
+## What does this mean for multi-tenancy?
+
+The multi-tenancy promises that OLMv0 made were false promises. Kubernetes is not multi-tenant with respect to management of APIs (because APIs are global). Any promise that OLMv0 has around multi-tenancy evaporates when true tenant isolation attempts are made, and any attempt to fix a broken promise is actually just a bandaid on an already broken assumption.
+
+So where do we go from here? There are multiple solutions that do not involve OLM implementing full multi-tenancy support, some or all of which can be explored.
+1. Customers transition to a control plane per tenant
+2. Extension authors update their operators to support customers’ multi-tenancy use cases
+3. Extension authors with “simple” lifecycling concerns transition to other packaging and deployment strategies (e.g. helm charts)
+
+### Single-tenant control planes
+
+One choice for customers would be to adopt low-overhead single-tenant control planes in which every tenant can have full control over their APIs and controllers and be truly isolated (at the control plane layer at least) from other tenants. With this option, the things OLMv1 cannot do (listed above) are irrelevant, because the purpose of all of those features is to support multi-tenant control planes in OLM.
+
+The [Kubernetes multi-tenancy docs](https://kubernetes.io/docs/concepts/security/multi-tenancy/#virtual-control-plane-per-tenant) contain a good overview of the options in this space. Kubernetes vendors may also have their own virtual control plane implementations.
+
+### Shift multi-tenant responsibility to operators
+
+There is a set of operators that both (a) provide fully namespace-scoped workload-style operands and that (b) provide a large amount of value to their users for advanced features like backup and migration. For these operators, the Operator Framework program would suggest that they shift toward supporting multi-tenancy directly. That would involve:
+1. Taking extreme care to avoid API breaking changes.
+2. Supporting multiple versions of their operands in a single version of the operator (if required by users in multi-tenant clusters).
+3. Maintaining support for versioned operands for the same period of time that the operator is supported for a given cluster version.
+4. Completely avoiding global configuration. Each tenant should be able to provide their configuration separately.
+
+### Operator authors ship controllers outside of OLM
+
+Some projects have been successful delivering and supporting their operator on Kubernetes, but outside of OLM, for example with helm-packaged operators. On this path, individual layered project teams have more flexibility in solving lifecycling problems for their users because they are unencumbered by OLM’s opinions. However the tradeoff is that those project teams and their users take on responsibility and accountability for safe upgrades, automation, and multi-tenant architectures. With OLMv1 no longer attempting to support multi-tenancy in a first-class way, these tradeoffs change and project teams may decide that a different approach is necessary.
+
+This path does not necessarily mean a scattering of content in various places. It would still be possible to provide customers with a marketplace of content (e.g. see https://artifacthub.io/).
+
+### Authors of "simple" operators ship their workload without an operator
+
+Another direction is to just stop shipping content via an operator. The operator pattern is particularly well-suited for applications that require complex upgrade logic, that need to convert declarative intent to imperative action, or that require sophisticated health monitoring logic and feedback. But a sizable portion of the OperatorHub catalog contain operators that are not actually taking advantage of the benefits of the operator pattern and are instead a simple wrapper around the workload, which is where the real value is.
+
+Using the [Operator Capability Levels](https://sdk.operatorframework.io/docs/overview/operator-capabilities/) as a rubric, operators that fall into Level 1 and some that fall into Level 2 are not making full use of the operator pattern. If content authors had the choice to ship their content without also shipping an operator that performs simple installation and upgrades, many supporting these Level 1 and Level 2 operators might make that choice to decrease their overall maintenance and support burden while losing very little in terms of value to their customers.
+
+## What will OLM do that a generic package manager doesn't?
+
+OLM will provide multiple features that are absent in generic package managers. Some items listed below are already implemented, while others may be planned for the future.
+
+### Upgrade controls
+
+An operator author can control the upgrade flow by specifying supported upgrade paths from one version to another. Or, they could use semantic versioning (semver) - this is fully supported by OLM too.
+
+A user can see the author’s supplied upgrade information.
+
+While these features might seem standard in package managers, they are fairly unique in the Kubernetes ecosystem. Helm, for example, doesn’t have any features that help users stay on supported upgrade paths.
+
+### On-cluster component for automated upgrades, health monitoring
+
+OLM automatically upgrades an operator to the latest acceptable matching version whenever a new matching version appears in a catalog, assuming the user has enabled this for their operator.
+
+OLM constantly monitors the state of all on-cluster resources for all the operators it manages, reporting the health in aggregate on each operator.
+
+### CRD Upgrade Safety Checks
+
+Before OLM upgrades a CRD, OLM performs a set of safety checks to identify any changes that potentially would have negative impacts, such as:
+- data loss
+- incompatible schema changes
+
+These checks may not be a guarantee that an upgrade is safe; instead, they are intended to provide an early warning sign for identifiable incompatibilities. False positives (OLMv1 claims a breaking change when there is none) and false negatives (a breaking change makes it through the check without being caught) are possible, at least while the OLMv1 team iterates on this feature.
+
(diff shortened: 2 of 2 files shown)


docs/olmv1_overview.md
```diff
@@ -0,0 +1,239 @@
...
+
+So where do we go from here? There are multiple solutions that do not involve OLM implementing full multi-tenancy support, some or all of which can be explored.
+1. Customers transition to a control plane per tenant
+2. Extension authors update their operators to support customers’ multi-tenancy use cases
+3. Extension authors with “simple” lifecycling concerns transition to other packaging and deployment strategies (e.g. helm charts)
+
+### Single-tenant control planes
+
+One choice for customers would be to adopt low-overhead single-tenant control planes (e.g. hypershift?) in which every tenant can have full control over their APIs and controllers and be truly isolated (at the control plane layer at least) from other tenants. With this option, the things OLMv1 cannot do (listed above) are irrelevant, because the purpose of all of those features is to support multi-tenant control planes in OLM.
```
everettraven, line 66: The [Kubernetes docs on multi-tenancy](https://kubernetes.io/docs/concepts/security/multi-tenancy/#virtual-control-plane-per-tenant) also mention the `Cluster` API, Kamaji, and vcluster as options for creating virtual control planes per tenant for tenant isolation. It might be worth linking to the kubernetes docs and/or some of the projects that are specifically designed for addressing this (hypershift can be included in this but shouldn't be the only example IMO)

everettraven: I think including links to some of these other projects trying to solve the multi-tenancy problem will help get the point across that using something specifically for enabling multi-tenancy is a better solution than trying to force multi-tenancy support into OLMv1

docs/olmv1_overview.md
```diff
@@ -0,0 +1,239 @@
...
+
+Another direction is to just stop shipping content via an operator. The operator pattern is particularly well-suited for applications that require complex upgrade logic, that need to convert declarative intent to imperative action, or that require sophisticated health monitoring logic and feedback. But a sizable portion of the OperatorHub catalog contain operators that are not actually taking advantage of the benefits of the operator pattern and are instead a simple wrapper around the workload, which is where the real value is.
+
+Using the [Operator Capability Levels](https://sdk.operatorframework.io/docs/overview/operator-capabilities/) as a rubric, operators that fall into Level 1 and some that fall into Level 2 are not making full use of the operator pattern. If content authors had the choice to ship their content without also shipping an operator that performs simple installation and upgrades, many supporting these Level 1 and Level 2 operators might make that choice to decrease their overall maintenance and support burden while losing very little in terms of value to their customers.
+
+## What will OLM doo that a generic package manager doesn't?
```
everettraven, line 88:
```suggestion
## What will OLM do that a generic package manager doesn't?
```

everettraven (approved):
Overall looks good to me. Had a couple comments, but nothing that I think is worth really holding this PR for.

docs/olmv1_overview.md
```diff
@@ -0,0 +1,239 @@
...
+
+Nearly every engineer in the Operator Framework group contributed to design explorations and prototypes over an entire year. For each of these design explorations, there are complex webs of features and assumptions that are necessary to understand the context that ultimately led to a conclusion of infeasibility that led us to today’s conclusion.
+
+Here is a sampling of some of the ideas we explored:
+- [[WIP] OLM v1's approach to multi-tenancy](https://docs.google.com/document/d/1xTu7XadmqD61imJisjnP9A6k38_fiZQ8ThvZSDYszog/edit#heading=h.m19itc78n5rw)
+- [OLMv1 Multi-tenancy Brainstorming](https://docs.google.com/document/d/1ihFuJR9YS_GWW4_p3qjXu3WjvK0NIPIkt0qGixirQO8/edit#heading=h.vy9860qq1j01)
+
+### Watched namespaces cannot be configured in a first-class API
+
+OLMv1 will not have a first-class API for configuring the namespaces that a controller will watch.
```
griffindvs, line 15: How is "first-class API" being defined? Does this mean that OLM will not provide this capability itself, but this could be provided by someone else?

docs/olmv1_overview.md
```diff
@@ -0,0 +1,239 @@
...
+
+Kubernetes APIs are global. Kubernetes is designed with the assumption that a controller WILL reconcile an object no matter where it is in the cluster.
+
+However, Kubernetes does not assume that a controller will be successful when it reconciles an object.
+
+The Kubernetes design assumptions are:
+- CRDs and their controllers are trusted cluster extensions.
```
griffindvs, line 22: I understand that a CRD (the API) is global, but it is not clear to me why a controller cannot be namespaced. If the controller is running in a namespace, with a service account that has RBAC limited to the namespace, and only reconciles CRs within that namespace -- I'm not sure I would consider that a cluster extension?

docs/olmv1_overview.md
```diff
@@ -0,0 +1,239 @@
...
+- If an object for an API exists a controller WILL reconcile it, no matter where it is in the cluster.
```
griffindvs, line 23: If a controller is running with a service account with namespaced RBAC, would the controller even see a CR created in another namespace?

docs/olmv1_overview.md
```diff
@@ -0,0 +1,239 @@
...
+
+OLMv1 will make the same assumption that Kubernetes does and that users of Kubernetes APIs do. That is: If a user has RBAC to create an object in the cluster, they can expect that a controller exists that will reconcile that object. If this assumption does not hold, it will be considered a configuration issue, not an OLMv1 bug.
+
+This means that it is a best practice to implement and configure controllers to have cluster-wide permission to read and update the status of their primary APIs. It does not mean that a controller needs cluster-wide access to read/write secondary APIs. If a controller can update the status of its primary APIs, it can tell users when it lacks permission to act on secondary APIs.
```
griffindvs, line 27: I wonder if it would be useful to challenge this. Although APIs are inherently cluster-scoped in Kubernetes, can we achieve some level of multi-tenancy by splitting the APIs from the controllers?

Can a cluster admin make an API available -- with no controller installed with cluster-wide RBAC? Can a namespace admin then install a controller in their namespace with limited RBAC to reconcile CRs in that namespace? This may require allowing for the installation of individual components of a bundle.

docs/olmv1_overview.md
```diff
@@ -0,0 +1,239 @@
...
+
+### Dependencies based on watched namespaces
+
+Since there will be no first-class support for configuration of watched namespaces, OLMv1 cannot resolve dependencies among bundles based on where controllers are watching.
```
griffindvs, line 31: If we have the following scenario:

* Customer goes to install Operator A at the cluster scope (CRD + controller with cluster-wide RBAC)
* Operator A has a dependency on Operator B (A will be creating CRs from B and expects B to reconcile them)

I'm not sure why the absence of a watched namespaces concept prevents this dependency fulfillment.

OLMv1, given its limited RBAC, can likely only see that CRD B is not present. If it was present, OLMv1 couldn't see if controller B is available and reconciling.

OLMv1 however, if its RBAC was limited to a single namespace, could create a CR for B and see if controller B picks it up and sets some status field. This would give OLMv1 the information it needs about whether controller B is running and reconciling at the cluster scope (which as presented is the recommend -- possibly only? -- install mode).

docs/olmv1_overview.md
```diff
@@ -0,0 +1,239 @@
...
+2. Even if (1) could be overcome, the lack of awareness of watched namespaces means that the resolver would have to make assumptions. If only one controller is installed, is it watching the right set of namespaces to meet the constraint? If multiple controllers are installed, are any of them watching the right set of namespaces? Without knowing the watched namespaces of the parent and child controllers, a correct dependency resolver implementation is not possible to implement.
+
+Note that regardless of the ability of OLMv1 to perform dependency resolution (now or in the future), OLMv1 will not automatically install a missing dependency when a user requests an operator. The primary reasoning is that OLMv1 will err on the side of predictability and cluster-administrator awareness.
```
griffindvs, line 43: Given it was properly conveyed via the UI (or logging in a CLI), I think you can achieve administrator awareness while still fulfilling dependencies.

docs/olmv1_overview.md
```diff
@@ -0,0 +1,239 @@
...
+2. OLMv0’s logic for choosing a dependency among multiple contenders is confusing and error-prone, and an administrator’s ability to have fine-grained control of upgrades is essentially limited to building and deploying tailor-made catalogs.
+3. OLMv0 automatically installs dependencies. The only way for an administrator to avoid this OLMv0 functionality is to fully understand the dependency tree in advance and to then install dependencies from the leaves to the root so that OLMv0 always detects that dependencies are already met. If OLMv0 installs a dependency for you, it does not uninstall it when it is no longer depended upon.
+
+OLMv1 will not provide dependency resolution among packages in the catalog (see [Dependencies based on watched namespaces](#dependencies-based-on-watched-namespaces))
+
+OLMv1 will provide constraint checking based on available cluster state. Constraint checking will be limited to checking whether the existing constraints are met. If so, install proceeds. If not, unmet constraints will be reported and the install/upgrade waits until constraints are met.
```
griffindvs, line 214: In OLMv0 this seems to exist via `nativeAPIs` defined in CSVs. The way this is built today, if you try to install an operator via the UI and the `nativeAPI` requirements are not fulfilled, the failure is not made apparent to the user with the missing dependencies. The installation appears to just be stuck pending. I'd suggest we make this more visible and easily debuggable in OLMv1.

docs/olmv1_overview.md
everettraven: My understanding is that not having a "first-class API" means that there will be no field on the APIs introduced by OLMv1 to set that information explicitly. There is nothing stopping the author of a controller from providing configuration options to users for this or hardcoding the namespaces it should watch.

I don't think we have really talked about this, but I could imagine the APIs introduced by OLMv1 having a kind of "pass through" field where users can use to set some arbitrary values on the manifests deployed by OLM that doesn't prevent this (e.g setting env vars on a deployment).

docs/olmv1_overview.md
everettraven: I could be wrong, but in this case I am interpreting "cluster extension" as referring to it literally being something that extends the functionality of the Kubernetes cluster. I don't think there is anything in Kubernetes preventing a controller from running with reconciliation scoped only to a namespace.

docs/olmv1_overview.md
everettraven: No. My understanding is that the design assumption we are stating here is that Kubernetes assumes that there is _a_ controller _somewhere_ that will reconcile an object for an API no matter where it is in the cluster. As far as I am aware this doesn't necessarily mean it _has_ to be a single instance of a controller. I do think in general it makes more sense to use a single controller over many namespace specific controllers though.

docs/olmv1_overview.md
everettraven: I don't think we are going to intentionally prevent any of those possible approaches to trying to achieve multi-tenancy but I don't think we are going to intentionally support them either. A user absolutely should be able to do all of those things by creating individual bundles that way, but going that route fights the design of Kubernetes and comes with a host of it's own issues, making it difficult for OLM to be able to effectively handle the automatic life cycling of the controllers. Taking this approach, it will be entirely on the user to understand the impacts of making the decision of using this approach and why OLM may not behave as expected in this case. 

For example, I would not see it as a bug if automatic upgrades for a controller were to fail and require manual intervention in this scenario. I would also not see it as a bug if OLM successfully installed multiple controllers reconcile the same resource in the same namespace in this scenario.

docs/olmv1_overview.md
```diff
@@ -0,0 +1,239 @@
+# OLMv1 Overview
+
+## What won't OLMv1 do that OLMv0 did?
+
+TL;DR: OLMv1 cannot feasibly support multi-tenancy or any feature that assumes multi-tenancy. All multi-tenancy features end up falling over because of the global API system of Kubernetes. While this short conclusion may be unsatisfying, the reasons are complex and intertwined.
+
+Nearly every engineer in the Operator Framework group contributed to design explorations and prototypes over an entire year. For each of these design explorations, there are complex webs of features and assumptions that are necessary to understand the context that ultimately led to a conclusion of infeasibility that led us to today’s conclusion.
```
grokspawn, line 7:
```suggestion
Nearly every engineer in the Operator Framework group contributed to design explorations and prototypes over an entire year. For each of these design explorations, there are complex webs of features and assumptions that are necessary to understand the context that ultimately led to a conclusion of infeasibility.
```

docs/olmv1_overview.md
acornett21: I read `cluster extension` as things that extend the functionally of the cluster, further then what is shipped with a default install of Kubernetes. This has no baring on the scope that an operator author chooses to be implemented within a controller. Even if the controller in question is _just_ namespace scoped, it still `extended` a k8s clusters functionality.

docs/olmv1_overview.md
joelanford: Yep, I'll update this. Thanks! Missed the hypershift callout in my RH-specific scrub.

docs/olmv1_overview.md
joelanford: +1 Bryce. Some of this is already captured in issues:
- https://github.com/operator-framework/operator-controller/issues/381
- https://github.com/operator-framework/operator-controller/issues/612
- https://github.com/operator-framework/operator-controller/issues/593

docs/olmv1_overview.md
joelanford: The fact that the APIs are global means that there isn't true multi-tenancy of controllers either. All of the controllers for that global API MUST agree on the single API they will all use. Therefore tenants will be limited by the choices made by other tenants when it comes to lifecycling controllers.

As Bryce said, OLMv1 will not get in the way of multiple controller installations, but it also won't help de-conflict between them.

docs/olmv1_overview.md
joelanford: > Customer goes to install Operator A at the cluster scope (CRD + controller with cluster-wide RBAC)

OLMv1 will be completely unaware of the scoping configuration of Operator A. It doesn't know if Operator A is watching the entire cluster or is watching just a subset.

The fact that an opaque configuration applied to the bundle results in cluster-wide RBAC for the service account tied to a deployment is _maybe_ a good enough proxy for watch namespace. But you are correct that a dependency resolver would _also_ need awareness of the RBAC in use by any dependents, which is not available to OLM and may not be available to a user.

> OLMv1 however, if its RBAC was limited to a single namespace, could create a CR for B and see if controller B picks it up and sets some status field. This would give OLMv1 the information it needs about whether controller B is running and reconciling at the cluster scope (which as presented is the recommend -- possibly only? -- install mode).

This seems fragile and complex and lots could go wrong. Off the top of my head:
- OLM would have to evaluate the schema of each CRD and be able to produce a valid object
- This assumes that all objects have a status
- Operators may have admission webhooks that reject creates unless arbitrary conditions are met.
- Creating a CR may have major implications on the operations of a cluster, and would like incur costs.
- The fact that a CR for B can be created in a particular namespace provides no signal about whether a controller would reconcile B in another namespace.

The most likely dependency resolver implementation given the constraints is probably:
- Client-based
- Limited to cluster admins who can see RBAC and Extensions cluster-wide.
- Required to use RBAC as a proxy for watch namespaces, which may result in assumptions that a controller is watching a namespace, even if it is not (i.e. it has RBAC, but for whatever reason isn't actually watching there)

docs/olmv1_overview.md
```diff
@@ -0,0 +1,239 @@
...
+
+Using the [Operator Capability Levels](https://sdk.operatorframework.io/docs/overview/operator-capabilities/) as a rubric, operators that fall into Level 1 and some that fall into Level 2 are not making full use of the operator pattern. If content authors had the choice to ship their content without also shipping an operator that performs simple installation and upgrades, many supporting these Level 1 and Level 2 operators might make that choice to decrease their overall maintenance and support burden while losing very little in terms of value to their customers.
+
+## What will OLM doo that a generic package manager doesn't?
+
+OLM will provide multiple features that are absent in generic package managers. Some items listed below are already implemented, while others are most likely planned for the future.
```
tmshort, line 90: Remove "are most likely". Prefer simply "are", or "may be".

docs/olmv1_overview.md
joelanford: @tmshort This is somewhat nuanced, and I want to make sure that the overall message is clear.

Namespace watching is not _provided_ by OLMv1 at all. However, OLMv1 _will_ support arbitrary configuration with schemas provided by extension authors and values that must match those schemas provided by extension admins. If authors decide to include knobs in their configuration schemas that control scoping, that's fine. But it isn't anything OLMv1 knows about or can build features around.

I call this out with a bit more detail in the `Approach -> Don't Fight Kubernetes` section further down.

There is one and only one exception. In order to maintain backward compatibility with registry+v1 bundles, the OLM maintainers will define the parameterization schema for registry+v1 bundles and will include the ability to define the watched namespaces. But again, this is particular to this bundle format and is not something that the broader OLM system will have awareness of.

docs/olmv1_overview.md
joelanford: Yes, there are a few rough edges of OLMv0 like this. Another is `minKubeVersion` in the CSV.  The goal is to get all of the constraint-related information into the catalog where it can be evaluated before pulling, extracting, and applying bundle contents.

README.md
```diff
@@ -1,13 +1,23 @@
...
+
+OLM’s purpose is to provide APIs, controllers, and tooling that support the packaging, distribution, and lifecycling of Kubernetes extensions. It aims to align with Kubernetes designs and user assumptions, provide secure, high-quality, and predictable user experiences centered around declarative GitOps concepts, and to give cluster admins the minimal necessary controls to build their desired cluster architectures and to have ultimate control.
+
+## Overview
+
+OLM v1 is the follow-up to OLM v0, located [here](https://github.com/operator-framework/operator-lifecycle-manager).
+
+It consists of four different components, including this one, which are as follows:
 * operator-controller
 * [deppy](https://github.com/operator-framework/deppy)
```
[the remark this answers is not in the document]
joelanford: It is still in use by the ClusterExtension controller. As is rukpak. Let's leave those in until we actually stop using them?

docs/olmv1_overview.md
joelanford: Predictability is definitely the harder thing to achieve though. All sorts of inputs go into which dependency is chosen. In other package managers, there is almost always an imperative flow where an admin has a chance to review the chosen dependencies before they are installed.

If it was possible to overcome the difficulty of building a dependency resolver without awareness of controller scope (that's a big if, and not something we're pursuing), then a client-based resolver that presents the user with the chosen set of Extensions to install and lets them decide how to proceed would be the best of both worlds.

With OLMv1's focus on GitOps friendliness and security posture, we have decided not to pursue a controller-based dependency resolver/installer.

And again, this is all fairly moot because we are not pursuing a dependency resolver. However, part of the beauty of this design is that it lends itself more to extensibility. A third party should be able to implement a dependency resolver over the APIs provided by core OLMv1:
- Catalog contents are available to clients
- Catalog metadata is extensible, so third parties could include their own dependency metadata (e.g. about "requires", "provides", "conflicts", etc).
- Extension API is available to clients

durera (commented):
To sum up my thoughts on reviewing this:

OLM v1 feels like a wasted opportunity in the form presented and I really hoped to see more focus around the previously discussed separation of API (cluster-scoped) from controller (not necessarily cluster-scoped) and the impact on lifecycle that such a split would necessitate.

I thought/hoped that OLM v1 would be trying to provide a solution for that challenging problem, rather than what appears to me to be a view of "that's someone else's problem to solve, we are going to design OLM to not get in the way of whoever tries to solve that"

In a model where the CRD and the controller(s) that implement that CRD are distinct entities I don't think something that wants to be a lifecycle manager can just "opt out" of such a key aspect that drives lifecycle events as the relationship between CRD and controller(s).

I think that the discussion around tenancy in the cluster has got in the way a little here perhaps, because this is not about multi-tenancy really, and - unlike multi-tenancy - this seperation of CRD from controller is something that is a natural fit for Kubernetes, certainly a more natural fit then having them bound together as it is today IMO.

OLM v1 should be the glue that binds and manages the CRD and it's controllers, that's what I'd expect of a lifecycle manager for cluster extensions.  Put the discussion about tenancy aside, this isn't about being able to isolate one namespace from changes in another (which as had been said many times really isn't possible due to the nature of Kubernetes), but as a cluster administrator I should be able to install an API in my cluster, and then control which namespaces can use that API.

- Manage a Kubernetes API extension (CRD)
- Manage a single cluster-scoped controller for an installed API extension
- Manage one or more namespace-scoped controllers for an installed API extension

I'd expect OLM to be the thing that can manage the extension as a whole including:
- Ensuring either use of multiple controllers with namespace scopes or a single cluster scoped controller, and preventing a mix or the two
- Managing the relationship between controller and CRD, and e.g. preventing controller updates happening which would create incompatibiity with the CRD version or at least flagging them as moving into a warning state if the controller and the CRD are not compatible
- Moving from a model where the CRD and controller are delivered in one package to offering distinct packages for CRDs and controllers, as well as a convenience bundle for both perhaps

The customers with large clusters that I have worked with in the last 2 years do not want to be using cluster-scoped controllers, and they are not seeking to address tenancy concerns through the use of multiple namespace-scoped controllers; they see namespace-scoped controllers as a way to enable the extensions on a namespace-by-namespace basis.  Yes, the extension is installed to the cluster, but it's only been enabled for us in namespace1,2, & 3.  

They are seeking to limit the exposure/impact of introducing a new API in their large clusters rather than the ability to independently operate in namespace1 and namespace2 through some form on tenancy that guarantees nothing done in namespace1 will impact what's happening in namespace2.  

This is what I think the next evolution of OLM should be, transition to a first-class data model and lifecycle built around the seperation of CRD from controller.

**Example: Strimzi and Red Hat AMQ Streams**
Today use of these two major Kafka operators is problematic, if you install both in a cluster things go bad for you because both respond to the same API (`kafka.kafka.strimzi.io`).

I would expect/hope for OLM v1 to address this with first class support for something like this:
- Install the `kafka.strimzi.io` API extension to say "we support Kafka in this cluster" (no controller(s) included in this action)
- Install the Strimzi controller in `namespace1` (no change to the API included in this action)
- Install the AMQ Streams controller in `namespace2` (no change to the API included in this action, it can not break anything in `namespace1`)
- Update the `kafka.strimzi.io` API extension ... at this point we are performing a cluster-scoped operation that may impact all namespaces with a controller (`namespace1` and `namespace2`)
- Update the AMQ Streams controller in `namespace2` (no need to worry about impacting `namespace1`)

To me, this is what I think about I hear people talking about tenancy, because this is what the customers I work with are seeking, and it's a perfect fit for Kubernetes if we just broke apart the delivery mechanism for CRDs and controllers.

I would liken this change in approach to the difference it made when file-based catalogs came around and removed the channel graph information from the operator bundles.  It never made sense that an individual operator bundle had to know "what channel will I be in", or "what's the default channel of the package I belong to" and once we were able to define that relationship in the correct place (in the catalog itself) it was a game-changer for managing operator packages.  I feel that the same thing needs to happen for CRD/Controllers as part of the evolution of OLM.

docs/olmv1_overview.md
```diff
@@ -0,0 +1,243 @@
...
+### Watched namespaces cannot be configured in a first-class API
+
+OLMv1 will not have a first-class API for configuring the namespaces that a controller will watch.
+
+Kubernetes APIs are global. Kubernetes is designed with the assumption that a controller WILL reconcile an object no matter where it is in the cluster.
+
+However, Kubernetes does not assume that a controller will be successful when it reconciles an object.
+
+The Kubernetes design assumptions are:
+- CRDs and their controllers are trusted cluster extensions.
```
pgodowski, line 24: The example which is counter argument is the Ingress API, which has reference to the ingress class name, which may or may not having a running controller on the cluster.

pgodowski (changes requested):

docs/olmv1_overview.md
```diff
@@ -0,0 +1,243 @@
...
+2. Even if (1) could be overcome, the lack of awareness of watched namespaces means that the resolver would have to make assumptions. If only one controller is installed, is it watching the right set of namespaces to meet the constraint? If multiple controllers are installed, are any of them watching the right set of namespaces? Without knowing the watched namespaces of the parent and child controllers, a correct dependency resolver implementation is not possible to implement.
+
+Note that regardless of the ability of OLMv1 to perform dependency resolution (now or in the future), OLMv1 will not automatically install a missing dependency when a user requests an operator. The primary reasoning is that OLMv1 will err on the side of predictability and cluster-administrator awareness.
+
+### "Watch namespace"-aware operator discoverability
+
+When operators add APIs to a cluster, these APIs are globally visible. As stated before, there is an assumption in this design that a controller will reconcile an object of that API anywhere it exists in the cluster.
```
pgodowski, line 49: Such assumption is dangereous. Controllers might have privileged to read and modify Secrets, to properly manage the operands. Cluster admins will not allow global access to the all Secrets on the cluster, they will allow such access only to the selected namespaces only.

Example: lets assume we are installing Kafka operator at the cluster scope, which will have Secrets get/update permission. Cluster admins will not let Kafka operator access to namespaces running some other workloads where the confidential/sensitive data is stored as Secrets.

Therefore, controllers must be provided a way to restrict their access. It could be done as such that RBAC is granted only in selected namespaces and it is up to the controller to somehow know which namespaces to watch. But then, it begs for some API to understand what is such scope of controller, or at least some though being given what is the controller developer best practices how to handle scope discovery.

pgodowski (approved):

grokspawn (approved):
RH maintainers had a series of meetings last week and determined some updates/clarifications needed here, but it would be much easier to discern the updates in a separate PR, so let's merge this and do a follow-up to capture updates.  

Reviewers:  please interpret this as an attempt to ensure that your comment is interpreted in the updated context

PR #695, opened 2024-03-13
🌱 Move kube-rbac-proxy into base manager yaml

m1kola:
# Description

Having `kube-rbac-proxy` in a separate patch makes working with manager args confusing. We either have to include all args into that patch (e.g. like `--feature-gates=ForceSemverUpgradeConstraints=true` was added) or have a separate patch to override args.

Since we have always have `kube-rbac-proxy` enabled anyway - why not move it into the base manager manifest?

## Reviewer Checklist

- [ ] API Go Documentation


config/manager/manager.yaml
```diff
@@ -28,26 +28,22 @@ spec:
...
+                  operator: In
+                  values:
+                    - amd64
+                    - arm64
+                    - ppc64le
+                    - s390x
+                - key: kubernetes.io/os
+                  operator: In
+                  values:
+                    - linux
```
m1kola, line 46: This is not new: we already have affinity (see deleted `config/default/manager_auth_proxy_patch.yaml`).

everettraven (approved):
Seems reasonable to me, thanks @m1kola !

PR #699, opened 2024-03-14
:sparkles: Add field to Extension API to skip crd safety checks

### LINKED ISSUE #658: Add configuration to the `Extension` API for disabling CRD upgrade safety validations (should be enabled by default)
A new, optional, field should be introduced to the `Extension` API spec for enabling/disabling the CRD Upgrade Safety preflight checks. The default state should be enabled.

everettraven:
# Description
Adds an optional `Extension.spec.skipCRDUpgradeSafetyChecks` field that when set to `true` will be used to disable the CRD upgrade safety checks when attempting to install an extension.

No logic included in this PR as the CRD Upgrade Safety preflight check hasn't been fully implemented in carvel-dev/kapp yet.

I figured it doesn't hurt to go ahead and add the field and have it do nothing for now, but I'm also happy for this PR to be put on hold until the underlying logic to enable/disable the check is able to be put in place.

fixes #658 

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/api/v1alpha1/extension_types.go
+++ b/api/v1alpha1/extension_types.go
@@ -82,6 +82,11 @@ type ExtensionSpec struct {
 
 	// source of Extension to be installed
 	Source ExtensionSource `json:"source"`
+
+	//+kubebuilder:Optional
+	//
+	// skipCRDUpgradeSafetyCheck specifies whether or not the CRD upgrade safety checks should be skipped when attempting to install the extension
+	SkipCRDUpgradeSafetyCheck bool `json:"skipCRDUpgradeSafetyCheck,omitempty"`
 }
 
 // ExtensionStatus defines the observed state of Extension
--- a/Makefile
+++ b/Makefile
@@ -95,7 +95,7 @@ generate: $(CONTROLLER_GEN) #EXHELP Generate code containing DeepCopy, DeepCopyI
 	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
 
 .PHONY: verify
-verify: tidy fmt vet generate #HELP Verify all generated code is up-to-date.
+verify: tidy fmt vet generate manifests #HELP Verify all generated code is up-to-date.
 	git diff --exit-code
 
 .PHONY: fmt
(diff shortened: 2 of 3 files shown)


perdasilva (approved):
lgtm

api/v1alpha1/extension_types.go
```diff
@@ -82,6 +82,11 @@ type ExtensionSpec struct {
 
 	// source of Extension to be installed
 	Source ExtensionSource `json:"source"`
+
+	//+kubebuilder:Optional
+	//
+	// skipCRDUpgradeSafetyCheck specifies whether or not the CRD upgrade safety checks should be skipped when attempting to install the extension
+	SkipCRDUpgradeSafetyCheck bool `json:"skipCRDUpgradeSafetyCheck,omitempty"`
```
grokspawn, line 89: I think in the past we identified that bool fields are problematic over time since quite frequently an additional value would slip in, so for maximum flexibility we've been using enum/string for this role, like [so](https://github.com/operator-framework/operator-controller/blob/main/config/crd/bases/olm.operatorframework.io_clusterextensions.yaml#L46-L52). 
Is there any reason not to do so here?

everettraven: I don't disagree, but I recall us having also gone back and forth on our feelings on that. I'm happy to make a change to this, but I do think we should make sure we come up with a consistent approach as a project and that we all agree to prevent flip-flopping.

joelanford: IMO, let's go ahead and get this merged to unblock further development. We'll separately plan for a full API review once we have all of the MVP functionality complete, and do any necessary refactoring at that point.

joelanford (dismissed):

joelanford:
I'm surprised I don't see one of:
- a failing sanity test
- an updated CRD YAML

everettraven:
> I'm surprised I don't see one of:
> 
> - a failing sanity test
> 
> - an updated CRD YAML 

That's a good point, I totally forgot to run `make generate` to update the CRD YAML

everettraven:
@joelanford I just pushed some updates to fix what you pointed out was missing. Great catch!

joelanford (approved):

PR #704, opened 2024-03-18
🌱 Update e2e action to reduce flakes

everettraven:
# Description
- Removes the `-k` option from the `make` command run to execute our e2e tests in CI. The `-k` flag would force `make` to continue running a target even when a dependency of the target has failed leading to flakes. Removing the flag means that the e2e tests will fail if any of the `test-e2e` target dependencies fail.

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -28,7 +28,7 @@ jobs:
           # after failing tests.
           # With -k flag make will continue the build, but will return non-zero
           # exit code in case of any errors.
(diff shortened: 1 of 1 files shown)


grokspawn (approved):
I imagine that the longer-term strategy we need to consider here is where our ci workflow would execute a script with robust error trapping instead of simply a stack of dependencies like we do today.  
But we'll cross that bridge when/if we get to it.

PR #705, opened 2024-03-19
🌱 Source Makefile kapp version from go.mod

dtfranz:
Grabs the kapp version from what we are currently using in go.mod in order to ensure parity.


ankitathomas (approved):
/lgtm

PR #711, opened 2024-03-22
✨ Filter out bundle versions lower than installed

### LINKED ISSUE #654: (Follow up): Integrate Kapp: Improve resolution logic in extension reconciler
#625 currently performs very basic filtering/sorting as a part of its resolution logic. This issue is intended to track work on improving the resolution logic without the use of deppy. More details can be found on the comment: https://github.com/operator-framework/operator-controller/pull/625#discussion_r1504489670

dtfranz:
# Description
Adds an annotation to the App created for each extension indicating the bundle version, which can then be used on future reconciles to filter out older versions. This can be overridden by setting the Extension's `UpgradeConstraintPolicy` to `Ignore`.

Closes #654 

This PR borrows a lot of stuff from @varshaprasad96 's PR #690 so I've added her as co-author :)

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -18,7 +18,6 @@ package main
 
 import (
 	"flag"
-	"fmt"
 	"net/http"
 	"os"
 	"time"
@@ -28,13 +27,10 @@ import (
 	rukpakv1alpha2 "github.com/operator-framework/rukpak/api/v1alpha2"
 	"github.com/spf13/pflag"
 	"go.uber.org/zap/zapcore"
-	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
-	"k8s.io/client-go/discovery"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
-	"k8s.io/client-go/rest"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
@@ -131,16 +127,9 @@ func main() {
 		os.Exit(1)
 	}
 
-	hasKappApis, err := hasKappApis(mgr.GetConfig())
-	if err != nil {
-		setupLog.Error(err, "unable to evaluate if App needs to be created")
-		os.Exit(1)
-	}
-
 	if err = (&controllers.ExtensionReconciler{
 		Client:         cl,
 		BundleProvider: catalogClient,
-		HasKappApis:    hasKappApis,
 	}).SetupWithManager(mgr); err != nil {
 		setupLog.Error(err, "unable to create controller", "controller", "Extension")
 		os.Exit(1)
@@ -161,28 +150,4 @@ func main() {
 		setupLog.Error(err, "problem running manager")
 		os.Exit(1)
 	}
-}
-
-// hasKappApis checks whether the cluster has Kapp APIs installed in the cluster.
-// This does not guarantee that the controller is present to reconcile the App CRs.
-func hasKappApis(config *rest.Config) (bool, error) {
-	discoveryClient, err := discovery.NewDiscoveryClientForConfig(config)
-	if err != nil {
-		return false, fmt.Errorf("creating discovery client: %v", err)
-	}
-	apiResourceList, err := discoveryClient.ServerResourcesForGroupVersion(carvelv1alpha1.SchemeGroupVersion.String())
-	if err != nil && !errors.IsNotFound(err) {
-		return false, fmt.Errorf("listing resource APIs: %v", err)
-	}
-
-	if apiResourceList == nil {
-		return false, nil
-	}
-
-	for _, resource := range apiResourceList.APIResources {
-		if resource.Kind == "App" {
-			return true, nil
-		}
-	}
-	return false, nil
 }
--- a/internal/catalogmetadata/filter/bundle_predicates.go
+++ b/internal/catalogmetadata/filter/bundle_predicates.go
@@ -31,6 +31,19 @@ func InMastermindsSemverRange(semverRange *mmsemver.Constraints) Predicate[catal
 	}
 }
 
+func HigherBundleVersion(currentVersion *bsemver.Version) Predicate[catalogmetadata.Bundle] {
+	return func(bundle *catalogmetadata.Bundle) bool {
+		if currentVersion == nil {
+			return false
+		}
+		bundleVersion, err := bundle.Version()
+		if err != nil {
+			return false
+		}
+		return bundleVersion.GTE(*currentVersion)
+	}
+}
+
 func InBlangSemverRange(semverRange bsemver.Range) Predicate[catalogmetadata.Bundle] {
 	return func(bundle *catalogmetadata.Bundle) bool {
 		bundleVersion, err := bundle.Version()
--- a/internal/catalogmetadata/filter/bundle_predicates_test.go
+++ b/internal/catalogmetadata/filter/bundle_predicates_test.go
@@ -2,6 +2,7 @@ package filter_test
 
 import (
 	"encoding/json"
+	"fmt"
 	"testing"
 
 	mmsemver "github.com/Masterminds/semver/v3"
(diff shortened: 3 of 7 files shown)


internal/controllers/extension_controller.go
```diff
@@ -482,19 +481,38 @@ func (r *ExtensionReconciler) resolve(ctx context.Context, extension ocv1alpha1.
 		predicates = append(predicates, catalogfilter.InMastermindsSemverRange(vr))
 	}
 
+	// Do not include versions older than currently installed unless UpgradeConstraintPolicy = 'Ignore'
+	var installedVersion string
+	if extension.Spec.Source.Package.UpgradeConstraintPolicy != ocv1alpha1.UpgradeConstraintPolicyIgnore {
```
dtfranz, line 486: I'm not seeing an obvious way to reduce the nesting further without creating a `func`, which feels unnecessary just to get past the linter. What do you all think?

internal/controllers/extension_controller.go
```diff
@@ -55,10 +56,11 @@ import (
 type ExtensionReconciler struct {
 	client.Client
 	BundleProvider BundleProvider
-	HasKappApis    bool
 }
 
-var errkappAPIUnavailable = errors.New("kapp-controller apis unavailable on cluster")
+var (
+	bundleVersionKey = "olm.bundle-version"
```
dtfranz, line 62: I feel like I might be missing a best-practice for this sort of thing; let me know if there's a better way!

joelanford: I would suggest `olm.operatorframework.io/bundleVersion`

internal/controllers/extension_controller.go
joelanford: I actually would suggest a separate function, I think. I'd expect you to be able to outdent the `else if` contents because you'll have the ability to return early if there is an error or the annotation doesn't exist.

internal/catalogmetadata/filter/bundle_predicates.go
```diff
@@ -31,6 +31,19 @@ func InMastermindsSemverRange(semverRange *mmsemver.Constraints) Predicate[catal
 	}
 }
 
+func HigherBundleVersion(currentVersion *bsemver.Version) Predicate[catalogmetadata.Bundle] {
```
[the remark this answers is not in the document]
joelanford: Actually, I'm not sure ClusterExtension handles skips and skipRange for the second case. But it needs to there too.

cmd/manager/main.go
```diff
@@ -162,27 +151,3 @@ func main() {
...
-		return false, nil
-	}
-
-	for _, resource := range apiResourceList.APIResources {
-		if resource.Kind == "App" {
-			return true, nil
-		}
-	}
-	return false, nil
-}
```
tmshort, line 188: Why?

dtfranz: This change was taken from #690 , in which Varsha explained that the check is not needed due to the feature gate.

tmshort: So, why is it in yours? 😕

internal/controllers/extension_controller.go
dtfranz: I avoided returning early here for annotation failure or semver parsing because to me this filtering felt non-critical. The `resultSet` at the end will be sorted with the highest version at the top anyway, so besides performance improvements what does removing older versions from the `resultSet` give us?

dtfranz: To answer my own question a little bit, I can imagine if an installed bundle is removed from a catalog for some reason, we would want to give an error here rather than have this resolve to an older version of the bundle and cause a downgrade.

internal/catalogmetadata/filter/bundle_predicates.go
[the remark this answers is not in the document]
dtfranz: @joelanford Coming back to this after a v0 distraction and had a follow-up question here: for the second case, are we looking for bundles that do all three (replaces **and** skips **and** skipRange) or any one of the three?

cmd/manager/main.go
dtfranz: That PR is undergoing a refactor to combine the tests into a test loop, which is likely going to take some time given its relative priority. This particular change is valuable though since it made doing these tests much easier.

internal/catalogmetadata/filter/bundle_predicates.go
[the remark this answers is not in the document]
joelanford: Any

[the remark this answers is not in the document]
dtfranz: Thanks @joelanford ! I was just about finished adding the feature gating and replaces/skips/skipsRange then realized it was going to make the PR significantly larger (mostly due to tests). How do you feel about me doing that as a follow-up?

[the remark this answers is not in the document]
joelanford: +1 on wrapping this and capturing a follow-up issue.

I've implemented this for ClusterExtension here: #743 

Given the recent change roadmap updates, I think we should try to wrap up work on Extension API for now, and capture a "current state of the world" of the Extension API and behavior.

internal/catalogmetadata/filter/bundle_predicates.go
```diff
@@ -31,6 +31,19 @@ func InMastermindsSemverRange(semverRange *mmsemver.Constraints) Predicate[catal
...
+	return func(bundle *catalogmetadata.Bundle) bool {
+		if currentVersion == nil {
+			return false
+		}
+		bundleVersion, err := bundle.Version()
```
perdasilva, line 39: Not for this PR, just a though. I wonder if we should validate the bundle version on creation/ingress rather than on egress (i.e. bundle.Version() shouldn't return an error). So, a bundle object is always validated and correct.

internal/catalogmetadata/filter/bundle_predicates_test.go
```diff
@@ -62,6 +62,53 @@ func TestInMastermindsSemverRange(t *testing.T) {
 	assert.False(t, f(b3))
 }
 
+func TestHigherBundleVersion(t *testing.T) {
+	b1 := &catalogmetadata.Bundle{Bundle: declcfg.Bundle{
+		Properties: []property.Property{
+			{
+				Type:  property.TypePackage,
+				Value: json.RawMessage(`{"packageName": "package1", "version": "1.0.0"}`),
```
perdasilva, line 70: Oh, I see. I guess we'd need some sort of factory is we want to validate on creation...

internal/catalogmetadata/filter/bundle_predicates_test.go
```diff
@@ -62,6 +62,53 @@ func TestInMastermindsSemverRange(t *testing.T) {
 	assert.False(t, f(b3))
 }
 
+func TestHigherBundleVersion(t *testing.T) {
+	b1 := &catalogmetadata.Bundle{Bundle: declcfg.Bundle{
```
perdasilva, line 66: is it worth tabulating this test?

internal/controllers/extension_controller.go
```diff
@@ -189,7 +179,12 @@ func (r *ExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alpha1.Ext
 		return ctrl.Result{}, nil
 	}
 
-	app := r.GenerateExpectedApp(*ext, bundle.Image)
+	bundleVersion, err := bundle.Version()
+	if err != nil {
+		l.Info("could not determine bundle version", "name", ext.GetName(), "namespace", ext.GetNamespace())
+	}
+
+	app := r.GenerateExpectedApp(*ext, bundle.Image, bundleVersion.String())
```
perdasilva, line 187: will we get an NPE on `bundleVersion.String()` here if `bundle.Version()` returned an error?

perdasilva: Also, should we just pass the `bundle` to `GenerateExpectedApp` and grab the version inside? Simplifies the function signature and makes the bundle available to the function in case other bundle attributes are needed.

internal/controllers/extension_controller.go
```diff
@@ -454,6 +452,23 @@ func (r *ExtensionReconciler) GenerateExpectedApp(o ocv1alpha1.Extension, bundle
 	return app
 }
 
+func (r *ExtensionReconciler) getInstalledVersion(ctx context.Context, extNN types.NamespacedName) (*bsemver.Version, error) {
```
perdasilva, line 455: `extNN` is a little unclear to me (ah! NamespacedName). Would `extNamespacedName` be too verbose? Or maybe just `namespaceName`?

internal/controllers/extension_controller_test.go
```diff
@@ -92,6 +98,152 @@ func TestExtensionReconcile(t *testing.T) {
...
+	ctx := context.Background()
+
+	testCases := []struct {
+		name                    string
+		packageName             string
+		packageVersion          string
+		upgradeConstraintPolicy ocv1alpha1.UpgradeConstraintPolicy
+		wantErr                 error
+		existingApp             *carvelv1alpha1.App
+		assert                  func(*testing.T, ctrl.Result, client.Client, types.NamespacedName)
```
perdasilva, line 112: It seems the only difference in the `assert` function implementation for each case is where it expects a True or False condition, and the Message. Would it be better to update the test case struct to have the expected condition and message and move the assertions to loop? That way if something changes we don't need to do it for each test case.

internal/controllers/extension_controller_test.go
```diff
@@ -92,6 +98,152 @@ func TestExtensionReconcile(t *testing.T) {
...
+				assert.Equal(t, ctrl.Result{}, res)
+
+				condition := apimeta.FindStatusCondition(ext.Status.Conditions, ocv1alpha1.TypeResolved)
+				assert.NotNil(t, condition)
+				assert.Equal(t, metav1.ConditionTrue, condition.Status)
+				assert.Equal(t, `resolved to "quay.io/operatorhubio/prometheus@sha256:3e281e587de3d03011440685fc4fb782672beab044c1ebadc42788ce05a21c35"`, condition.Message)
+			},
+		},
+		{
+			name:           "existing App of higher version than available",
```
perdasilva, line 158:
```suggestion
			name:           "existing App of higher version than requested",
```

internal/catalogmetadata/filter/bundle_predicates.go
joelanford: I've had this thought too. If it didn't break any assumptions, it would be nice to get something into the `declcfg.Bundle` type for this.

dtfranz: Yeah that makes great sense and also would make it a lot easier to grab.

internal/controllers/extension_controller.go
dtfranz: +1, will add

internal/controllers/extension_controller_test.go
dtfranz: Updated

internal/controllers/extension_controller_test.go
dtfranz: Updated

internal/controllers/extension_controller.go
dtfranz: Updated

dtfranz:
Thank you for reviewing @perdasilva , I've updated per your comments. Please take another look when you can and let me know if you see anything else!

internal/catalogmetadata/filter/bundle_predicates.go
perdasilva: I'll create a discussion upstream - see what ppl think - or would a ticket suffice?

perdasilva (approved):
chef's kiss! Thanks mate!!

PR #756, opened 2024-04-16
:sparkles: Starting Helm POC - pull in some rukpak

tmshort:
# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/controllers/clusterextension_controller.go
+++ b/internal/controllers/clusterextension_controller.go
@@ -50,6 +50,7 @@ import (
 	"github.com/operator-framework/operator-controller/internal/catalogmetadata"
 	catalogfilter "github.com/operator-framework/operator-controller/internal/catalogmetadata/filter"
 	catalogsort "github.com/operator-framework/operator-controller/internal/catalogmetadata/sort"
+	rukpakapi "github.com/operator-framework/operator-controller/internal/rukpak/api"
 )
 
 // ClusterExtensionReconciler reconciles a ClusterExtension object
@@ -135,6 +136,12 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
 	// Unpack contents into a fs based on the bundle.
 	// Considering only image source.
 
+	// Generate a BundleSource, and then pass this and the ClusterExtension to Unpack
+	// TODO:
+	// bs := r.GenerateExpectedBundleSource(*ext, bundle.Image)
+	// unpacker := NewDefaultUnpacker(msg, namespace, unpackImage)
+	// unpacker..Unpack(bs, ext)
+
 	// set the status of the cluster extension based on the respective bundle deployment status conditions.
 	return ctrl.Result{}, nil
 }
@@ -218,6 +225,15 @@ func SetDeprecationStatus(ext *ocv1alpha1.ClusterExtension, bundle *catalogmetad
 			Message:            strings.Join(deprecationMessages, ";"),
 			ObservedGeneration: ext.Generation,
 		})
+	}
+}
+
+func (r *ClusterExtensionReconciler) GenerateExpectedBundleSource(o ocv1alpha1.ClusterExtension, bundlePath string) *rukpakapi.BundleSource {
+	return &rukpakapi.BundleSource{
+		Type: rukpakapi.SourceTypeImage,
+		Image: rukpakapi.ImageSource{
+			Ref: bundlePath,
+		},
 	}
 }
 
--- a/internal/rukpak/api/bundle_types.go
+++ b/internal/rukpak/api/bundle_types.go
@@ -1,0 +1,51 @@
+/*
+Copyright 2021.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package rukpakapi
+
+type SourceType string
+
+const (
+	SourceTypeImage SourceType = "image"
+
+	TypeUnpacked = "Unpacked"
+
+	ReasonUnpackPending             = "UnpackPending"
+	ReasonUnpacking                 = "Unpacking"
+	ReasonUnpackSuccessful          = "UnpackSuccessful"
+	ReasonUnpackFailed              = "UnpackFailed"
+	ReasonProcessingFinalizerFailed = "ProcessingFinalizerFailed"
+
+	PhasePending   = "Pending"
+	PhaseUnpacking = "Unpacking"
+	PhaseFailing   = "Failing"
+	PhaseUnpacked  = "Unpacked"
+)
+
+type BundleSource struct {
+	// Type defines the kind of Bundle content being sourced.
+	Type SourceType `json:"type"`
+	// Image is the bundle image that backs the content of this bundle.
+	Image ImageSource `json:"image,omitempty"`
+	// Git is the git repository that backs the content of this Bundle.
+}
+
+type ImageSource struct {
+	// Ref contains the reference to a container image containing Bundle contents.
+	Ref string `json:"ref"`
+	// ImagePullSecretName contains the name of the image pull secret in the namespace that the provisioner is deployed.
+	ImagePullSecretName string `json:"pullSecret,omitempty"`
+}
--- a/internal/rukpak/source/unpacker.go
+++ b/internal/rukpak/source/unpacker.go
@@ -1,0 +1,110 @@
+package source
+
+import (
+	"context"
+	"fmt"
+	"io/fs"
+
+	"k8s.io/client-go/kubernetes"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+
+	clusterextension "github.com/operator-framework/operator-controller/api/v1alpha1"
+	rukpakapi "github.com/operator-framework/operator-controller/internal/rukpak/api"
+)
+
+// Unpacker unpacks bundle content, either synchronously or asynchronously and
+// returns a Result, which conveys information about the progress of unpacking
+// the bundle content.
+//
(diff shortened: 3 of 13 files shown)


internal/rukpak/api/bundle_types.go
```diff
@@ -0,0 +1,123 @@
...
+	// Refer to https://kubernetes.io/docs/concepts/configuration/secret/#ssh-authentication-secrets
+	Secret corev1.LocalObjectReference `json:"secret,omitempty"`
+	// InsecureSkipVerify controls whether a client verifies the server's certificate chain and host name. If InsecureSkipVerify
+	// is true, the clone operation will accept any certificate presented by the server and any host name in that
+	// certificate. In this mode, TLS is susceptible to machine-in-the-middle attacks unless custom verification is
+	// used. This should be used only for testing.
+	InsecureSkipVerify bool `json:"insecureSkipVerify,omitempty"`
+}
+
+type ProvisionerID string
```
varshaprasad96, line 123: We don't need Provisioners any more, since we will have just one controller handling unpacking, and installing for all kinds of bundles. Also, for now, we are going to consider only registryV1 formats.

internal/rukpak/api/zz_generated.deepcopy.go
```diff
@@ -0,0 +1,151 @@
...
+func (in *HTTPSource) DeepCopy() *HTTPSource {
+	if in == nil {
+		return nil
+	}
+	out := new(HTTPSource)
+	in.DeepCopyInto(out)
+	return out
+}
+
+// DeepCopyInto is an autogenerated deepcopy function, copying the receiver, writing into out. in must be non-nil.
```
varshaprasad96, line 138: We need not have BundleSource as a separate API which means as mentioned in the previous comment we can remove these deep copy methods.

internal/rukpak/provisioner/helm/helm.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/operator-framework/operator-controller/internal/rukpak/util"
+	rukpakv1alpha2 "github.com/operator-framework/rukpak/api/v1alpha2"
+)
+
+const (
+	// ProvisionerID is the unique helm provisioner ID
+	ProvisionerID = "core-rukpak-io-helm"
+)
+
+func HandleBundleDeployment(_ context.Context, fsys fs.FS, bd *rukpakv1alpha2.BundleDeployment) (*chart.Chart, chartutil.Values, error) {
```
varshaprasad96, line 24: This bit of code is going to change since we don't have BundleDeployment anymore. Plus looks like this is using helm based bundles (from the provisionerID). We need to port the registryV1 piece of implementation, which would be:
1. Converting registryV1 to plain bundles: the conversion code which is called here: https://github.com/operator-framework/rukpak/blob/882714a00b5d99fdfc1b65992206c4e90edfd696/internal/provisioner/registry/registry.go#L22
2. The logic where we create a chart from the manifest objects: https://github.com/operator-framework/rukpak/blob/882714a00b5d99fdfc1b65992206c4e90edfd696/internal/provisioner/plain/plain.go#L28

This would need some amount of refinement in terms of creating the chart from unpacked bundle contents from the file system. Can we remove this from this PR, and work on this separately along with the helm install logic?

internal/rukpak/provisioner/plain/plain.go
```diff
@@ -0,0 +1,102 @@
...
+)
+
+const (
+	// ProvisionerID is the unique plain provisioner ID
+	ProvisionerID = "core-rukpak-io-plain"
+
+	manifestsDir = "manifests"
+)
+
+func HandleBundleDeployment(_ context.Context, fsys fs.FS, _ *rukpakv1alpha2.BundleDeployment) (*chart.Chart, chartutil.Values, error) {
```
varshaprasad96, line 27: As mentioned in the previous comment, this is going to change as we no longer have BundleDeployment. What do you think about removing the Handler implementations from this PR and working on them in follow ups?

internal/rukpak/provisioner/registry/registry.go
```diff
@@ -0,0 +1,27 @@
...
+	"github.com/operator-framework/operator-controller/internal/rukpak/provisioner/plain"
+	rukpakv1alpha2 "github.com/operator-framework/rukpak/api/v1alpha2"
+)
+
+const (
+	// ProvisionerID is the unique registry provisioner ID
+	ProvisionerID = "core-rukpak-io-registry"
+)
+
+func HandleBundleDeployment(ctx context.Context, fsys fs.FS, bd *rukpakv1alpha2.BundleDeployment) (*chart.Chart, chartutil.Values, error) {
```
varshaprasad96, line 21: The same as before in terms of removing BundleDeployment related handlers.

internal/rukpak/source/configmaps.go
```diff
@@ -0,0 +1,85 @@
+package source
+
+import (
```
varshaprasad96, line 3: We can have just the `internal/rukpak/source/image.go` for now, so that we figure out the unpacker for just one source for now. All the other implementations in `internal/rukpak/source/` can be removed for now

internal/rukpak/source/unpacker.go
```diff
@@ -0,0 +1,132 @@
...
+	// ResolvedSource is a reproducible view of a Bundle's Source.
+	// When possible, source implementations should return a ResolvedSource
+	// that pins the Source such that future fetches of the bundle content can
+	// be guaranteed to fetch the exact same bundle content as the original
+	// unpack.
+	//
+	// For example, resolved image sources should reference a container image
+	// digest rather than an image tag, and git sources should reference a
+	// commit hash rather than a branch or tag.
+	ResolvedSource *rukpakapi.BundleSource
```
varshaprasad96, line 50: I'm wondering if we need the ResolvedSource here? What if we directly populate the ClusterExtension status with the details of resolved bundle?

internal/rukpak/util/util.go
```diff
@@ -0,0 +1,300 @@
...
+)
+
+var (
+	// ErrMaxGeneratedLimit is the error returned by the BundleDeployment controller
+	// when the configured maximum number of Bundles that match a label selector
+	// has been reached.
+	ErrMaxGeneratedLimit = errors.New("reached the maximum generated Bundle limit")
+)
+
+func BundleDeploymentProvisionerFilter(provisionerClassName string) predicate.Predicate {
```
varshaprasad96, line 40: We are not using Provisioners here, so I think all the methods in this file can be removed and would not be useful.

tmshort:
General note: it was _just easy_ to pull in more code than necessary, rather than trying to figure out exactly what was needed.

internal/rukpak/provisioner/helm/helm.go
tmshort: Yes, this code was just pulled in, it hasn't been updated. I expect that things will be iterated and removed.

internal/controllers/clusterextension_controller.go
```diff
@@ -221,6 +228,15 @@ func SetDeprecationStatus(ext *ocv1alpha1.ClusterExtension, bundle *catalogmetad
 	}
 }
 
+func (r *ClusterExtensionReconciler) GenerateExpectedBundleSource(o ocv1alpha1.ClusterExtension, bundlePath string) *rukpakapi.BundleSource {
+	return &rukpakapi.BundleSource{
```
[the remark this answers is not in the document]
tmshort: So, this API is not a k8s API, it's just a struct as you point out. It was convenient to use. However, it's named `rukpakapi` since it's part of the interface to the pulled-in rukpak code.

internal/rukpak/source/unpacker.go
tmshort: Also, I was trying to avoid changing k8s APIs

internal/controllers/clusterextension_controller.go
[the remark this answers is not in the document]
varshaprasad96: It had deepycopy methods, so thought this was intended to be a separate API. The naming is all good 👍

tmshort:
There's still quite possibly too much code in here.

internal/rukpak/api/bundle_types.go
```diff
@@ -0,0 +1,54 @@
...
+
+package rukpakapi
+
+type SourceType string
+
+const (
+	SourceTypeImage      SourceType = "image"
+	SourceTypeGit        SourceType = "git"
+	SourceTypeConfigMaps SourceType = "configMaps"
+	SourceTypeHTTP       SourceType = "http"
```
varshaprasad96, line 25: Nit: cleanups
```suggestion
```

internal/rukpak/source/unpacker.go
```diff
@@ -0,0 +1,120 @@
...
+	if err != nil {
+		return nil, err
+	}
+	httpTransport := http.DefaultTransport.(*http.Transport).Clone()
+	if httpTransport.TLSClientConfig == nil {
+		httpTransport.TLSClientConfig = &tls.Config{
+			MinVersion: tls.VersionTLS12,
+		}
+	}
+	httpTransport.TLSClientConfig.RootCAs = rootCAs
```
varshaprasad96, line 111: I think we don't need the TLS client config or the rootCAs. IIRC it was used only in http source.

internal/rukpak/util/hash.go
```diff
@@ -0,0 +1,39 @@
...
+	"crypto/sha256"
+	"encoding/json"
+	"fmt"
+	"math/big"
+)
+
+// DeepHashObject writes specified object to hash using the spew library
+// which follows pointers and prints actual values of the nested objects
+// ensuring the hash does not change when a pointer changes.
+func DeepHashObject(obj interface{}) (string, error) {
```
varshaprasad96, line 13: TODO (follow up): We can have this for now, but just wondering if we need to ever use it anywhere to hash the entire spec of the object. The linter would complain in the future anyway, we can take care as the design proceeds.

varshaprasad96 (commented):
Just a nit, the other changes look good!

Are you going to plug in the unpacker in the reconciler in this PR, or as a follow up?

internal/rukpak/source/unpacker.go
tmshort: Hmmm... even in the original code, it doesn't appear used. It's a clone (copy?) and isn't even referenced after this. Odd...

varshaprasad96 (approved):
The CI may probably fail, since this is a PoC not going to worry about that.

/lgtm
/approve

Issue #458, opened 2023-10-13
Support for non-image bundle types


m1kola:
There are several places where we rely on bundle being an image. With the plan to support more types (such as git repos) we need to refactor operator controller to not assume image type.

Here are a few places where this is happening at the moment:

https://github.com/operator-framework/operator-controller/blob/e001a0f6ffd62fc478148463ad9b39a7b2e404a3/internal/controllers/operator_controller.go#L287-L291

https://github.com/operator-framework/operator-controller/blob/e001a0f6ffd62fc478148463ad9b39a7b2e404a3/internal/resolution/variablesources/bundle_deployment.go#L41

joelanford:
Yeah, I think this sounds correct. I know that @varshaprasad96's rukpak v1alpha2 work is going to change this to an extent.

Also, I think we're headed in a direction where the catalog contains a `BundleDeployment.spec` (or template of one) that would potentially simplify this. Operator controller would no longer have to derive a BD. It would just _have_ it from the catalog.

PR #758, opened 2024-04-17
⚠️ Remove Deppy solver

m1kola:
# Description

An attempt to remove Deppy and replace it with a simple catalog filtering. This also removes resolution CLI.

With this change we:
* No longer check/solving bundle dependencies
* No longer check bundle uniqueness on the cluster

## Reviewer Checklist

- [ ] API Go Documentation


tmshort:
Did you want to create a separate branch for this? e.g. `rm-deppy-poc`?

m1kola:
> Did you want to create a separate branch for this? e.g. `rm-deppy-poc`?
No, just accidentally removed the remote branch and GH closed the PR.

tmshort:
But, if this is a PoC, you don't want it merged...?

internal/controllers/successors.go
```diff
@@ -0,0 +1,47 @@
...
+
+	mmsemver "github.com/Masterminds/semver/v3"
+
+	"github.com/operator-framework/operator-controller/internal/catalogmetadata"
+	catalogfilter "github.com/operator-framework/operator-controller/internal/catalogmetadata/filter"
+)
+
+// successorsPredicateFunc returns a predicate to find successors
+// for a bundle. Predicate must not include the current version.
+type successorsPredicateFunc func(bundle *catalogmetadata.Bundle) (catalogfilter.Predicate[catalogmetadata.Bundle], error)
```
m1kola, line 14: This stuff can potentially be reused in the `Extension` controller.

internal/controllers/clusterextension_controller.go
```diff
@@ -202,21 +179,110 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
...
+		if err != nil {
+			return nil, fmt.Errorf("invalid version range %q: %w", versionRange, err)
+		}
+		predicates = append(predicates, catalogfilter.InMastermindsSemverRange(vr))
+	}
+
+	// Do not include bundle versions older than currently installed unless UpgradeConstraintPolicy = 'Ignore'
+	var installedVersion string
+	if ext.Spec.UpgradeConstraintPolicy != ocv1alpha1.UpgradeConstraintPolicyIgnore {
+		bd := &rukpakv1alpha2.BundleDeployment{}
```
m1kola, line 216: I dumped a lot in here. Will split into a separate function if we decide to go with it.

internal/controllers/clusterextension_controller_test.go
```diff
@@ -608,57 +608,6 @@ func TestClusterExtensionExpectedBundleDeployment(t *testing.T) {
 	require.NoError(t, cl.DeleteAllOf(ctx, &rukpakv1alpha2.BundleDeployment{}))
 }
 
-func TestClusterExtensionDuplicatePackage(t *testing.T) {
```
m1kola, line 611: Fails because we no longer check uniqueness. Do we want to deal with it in some other way without a solver? E.g. iterating over all extensions and checking their packages on each reconcile?

internal/controllers/clusterextension_controller.go
```diff
@@ -202,21 +179,110 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
...
+		err := r.Client.Get(ctx, types.NamespacedName{Name: ext.GetName()}, bd)
+		if client.IgnoreNotFound(err) != nil {
+			return nil, err
+		}
+
+		if !apierrors.IsNotFound(err) && bd.Spec.Source.Image != nil && bd.Spec.Source.Image.Ref != "" {
```
varshaprasad96, line 222: `client.IgnoreNotFound` is going to ignore 404 NotFound and return if there is any other error. We need not again check `!apierrors.IsNotFound(err)` - as that's dealing with 404 NotFound errors too. Am I missing something?

internal/controllers/clusterextension_controller.go
```diff
@@ -202,21 +179,110 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
...
+
+			sort.SliceStable(installedBundleResultSet, func(i, j int) bool {
+				return catalogsort.ByVersion(installedBundleResultSet[i], installedBundleResultSet[j])
+			})
+			installedBundle := installedBundleResultSet[0]
+			installedBundleVersion, err := installedBundle.Version()
+			if err != nil {
+				return nil, err
+			}
+			installedVersion = installedBundleVersion.String()
```
varshaprasad96, line 241: Can we have this in a separate method, that returns the installedVersion, so that its easier to change when we have a helm integration?

Also, to understand this better - To find the installed version in the cluster, we are looking at the spec of the bundle deployment. Since the spec is the specified (desired) component rather than what exists on cluster, do we need to look at the status rather?

There was a similar discussion related to this for the extension controller, and since we can't read from our own status, it was suggested to add an annotation containing the installed bundle Version into the object. Which is what is implemented [here](https://github.com/operator-framework/operator-controller/blob/25af3b02ab8e2737ec982628dccdaca38f0dc618/internal/controllers/extension_controller.go#L468) and is discussed in [slack](https://redhat-internal.slack.com/archives/C06KP34REFJ/p1710352584410229).

internal/controllers/successors.go
varshaprasad96, line 14: Can we also have tests for this? Probably a follow up is also good?

m1kola: Yes, we have tests cases in [here](https://github.com/operator-framework/operator-controller/blob/25af3b02ab8e2737ec982628dccdaca38f0dc618/internal/resolution/variablesources/installed_package_test.go). I'll have tests for the final version.

Needs a bit of work since I can't just copy-paste due to changed changes to the code for new context + these tests were not direct (they were testing `variablesources.MakeInstalledPackageVariables`). But for sure I want to cover this.

internal/controllers/clusterextension_controller.go
m1kola: You are right. I think this check should not be here. Will remove in the next update.

internal/controllers/clusterextension_controller.go
m1kola: > Can we have this in a separate method, that returns the installedVersion, so that its easier to change when we have a helm integration?

Yes, that's the plan for the final version.

> Also, to understand this better - To find the installed version in the cluster, we are looking at the spec of the bundle deployment. Since the spec is the specified (desired) component rather than what exists on cluster, do we need to look at the status rather?

Your understanding is correct. We had a similar discussion when we were initially implementing this and I can't remember why we decided to do it this way. I also can't find the relevant thread anymore (was a while ago on Kube slack).

But I think it is ok in this context: here were are reconciling `ClusterExtension` and are reading from `BundleDeployment`'s spec. The actual cluster state is not important here as we are trying to retrieve the bundle version from the previous reconciliation (e.g. previous desired state).

I suggest that for now we do the same approach as in main (based on spec) to reduce the blast radius. I also expect this bit to change once we start accounting for non-image content.

internal/controllers/clusterextension_controller.go
```diff
@@ -202,21 +178,133 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
...
+	}
+
+	sort.SliceStable(resultSet, func(i, j int) bool {
+		return catalogsort.ByVersion(resultSet[i], resultSet[j])
+	})
+
+	return resultSet[0], nil
+}
+
+func upgradePredicate(installedBundle *catalogmetadata.Bundle) (catalogfilter.Predicate[catalogmetadata.Bundle], error) {
```
m1kola, line 282: Can probably this helper into `internal/controllers/successors.go` so it can be re-used in `Extension`.

tmshort:
Looks reasonable at first pass.

internal/controllers/clusterextension_controller.go
varshaprasad96: Fair, since this is based on BD and we are going to modify it soon it's better we don't spend much time on this. I'm good to merge with what we have now.

internal/controllers/clusterextension_controller.go
```diff
@@ -202,21 +178,133 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
...
+
+	resultSet := catalogfilter.Filter(allBundles, catalogfilter.And(predicates...))
+
+	var upgradeErrorPrefix string
+	if installedBundle != nil {
+		installedBundleVersion, err := installedBundle.Version()
+		if err != nil {
+			return nil, err
+		}
+		upgradeErrorPrefix = fmt.Sprintf("error upgrading from currently installed version %q: ", installedBundleVersion.String())
```
varshaprasad96, line 229: Nit: `upgradeErrorPrefix` can be defined as a separate error type globally, and can then be wrapped/formatted to include the bundle version, along with other respective error info like version range, or channel name in the condition. This way it's easier to check for this error with `errors.Is` for the client.

varshaprasad96 (approved):
@m1kola The changes look good, just one nit. 

Thanks for cleaning up the solver code and fixing cluster extension to work without it!

internal/controllers/clusterextension_controller.go
m1kola: Wrapping errors makes them part of the API and I believe we currently do not have a use case for type checking. I suggest to leave errors opaque and add types, if/when we need them.

internal/controllers/clusterextension_controller.go
```diff
@@ -202,21 +178,133 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
...
+	successorsPredicate, err := successors(installedBundle)
+	if err != nil {
 		return nil, err
 	}
 
-	return GenerateVariables(allBundles, clusterExtensionList.Items, bundleDeploymentList.Items)
+	// We need either successors or current version (no upgrade)
+	return catalogfilter.Or(
+		successorsPredicate,
+		catalogfilter.InMastermindsSemverRange(installedVersionConstraint),
```
m1kola, line 306: This needs to be something like this:

```go
catalogfilter.And(
	catalogfilter.WithPackageName(installedBundle.Package),
	catalogfilter.InMastermindsSemverRange(installedVersionConstraint),
)
```

Otherwise we will include bundles which match `installedVersionConstraint`, but not from the same package.

internal/controllers/successors_test.go
```diff
@@ -1,7 +1,8 @@
-package variablesources_test
+package controllers_test
```
m1kola, line 1: It might be easier to review this file with "hide whitespace" option (`-w` in `git diff`).

m1kola:
Converted back to draft while we are looking into solving package uniqueness for `ClusterExtension`.

I had a chat with @joelanford about package uniqueness and we decided that it is best to not allow `ClusterExtension` to install the same package multiple times.

We will look into whether we can achieve that with `ValidatingAdmissionPolicy`. In the meantime I'm putting this on hold.

The plan is:

* Introduce validating admission for uniqueness
* Merge this PR (without uniqueness on the controller level)

m1kola:
#774 and #785 are now merged. I rebased the PR to solve conflicts. Should be ready for review now.

varshaprasad96 (approved):
The resolver removals and version filter additions look good to me! 
/lgtm

internal/controllers/clusterextension_controller.go
```diff
@@ -202,21 +177,105 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
...
-func (r *ClusterExtensionReconciler) variables(ctx context.Context) ([]deppy.Variable, error) {
+func (r *ClusterExtensionReconciler) resolve(ctx context.Context, ext *ocv1alpha1.ClusterExtension) (*catalogmetadata.Bundle, error) {
 	allBundles, err := r.BundleProvider.Bundles(ctx)
 	if err != nil {
 		return nil, err
 	}
-	clusterExtensionList := ocv1alpha1.ClusterExtensionList{}
-	if err := r.Client.List(ctx, &clusterExtensionList); err != nil {
+
+	installedBundle, err := r.installedBundle(ctx, allBundles, ext)
```
joelanford, line 186: At some point, I know we talked about adding some labels or annotations to the BD for the installed package name and version and using _that_ in order to derive successors.

Looking on the removed code in the installed_package variable source, I see that we never made that change. So I think it's fine to stick with that same logic like we are in this PR. But I also think we should capture a follow-up issue (if we don't already have one), to implement the installed bundle detection that way.

Somewhat related: I don't think we should have to find the existing installed bundle in the catalog in order to be able to upgrade from it.

joelanford (approved):
Left one comment about capturing a follow-up. Otherwise LGTM to merge this as is.

internal/controllers/clusterextension_controller.go
m1kola: > At some point, I know we talked about adding some labels or annotations to the BD for the installed package name and version and using _that_ in order to derive successors.
>
> Looking on the removed code in the installed_package variable source, I see that we never made that change. So I think it's fine to stick with that same logic like we are in this PR. But I also think we should capture a follow-up issue (if we don't already have one), to implement the installed bundle detection that way.

@joelanford I can't find the thread in upstream slack, but I think we ended up deciding against using labels and in favour of digests which were supposed to support non-image sources. We have [this issue](https://github.com/operator-framework/operator-controller/issues/458) for non-image bundle types support and the idea with digests captured [here](https://github.com/operator-framework/operator-controller/issues/458#issuecomment-1761199892).

> Somewhat related: I don't think we should have to find the existing installed bundle in the catalog in order to be able to upgrade from it.

I might be missing something, but in order to support both server and legacy semantics we seem to need the following data:
* Package name (for both server and legacy)
* Current version (for both server and legacy)
* Bundle name (for legacy)

Now that we have #679 we _might_ be able to get this info from `ClusterExtensions`'s status, but we probably don't want to to read from `ClusterExtensions`'s status while reconciling `ClusterExtensions`.

Even if we store this data in labels and use on upgrades - it will be an equivalent of looking at `ClusterExtensions`'s status.

grokspawn:
> Left one comment about capturing a follow-up. Otherwise LGTM to merge this as is.

I'd merge this, but I'd like to make sure that a follow-up need is captured somewhere first.  @m1kola is there an issue/pr for this yet?  Could you create one (if not already existing) and mention it here?

m1kola:
> I'd merge this, but I'd like to make sure that a follow-up need is captured somewhere first. @m1kola is there an issue/pr for this yet? Could you create one (if not already existing) and mention it here?

@grokspawn we already have something that captures this. See https://github.com/operator-framework/operator-controller/pull/758#discussion_r1584658741 for more details.

internal/controllers/clusterextension_controller.go
joelanford: Agreed we should not read from `ClusterExtension` status. I think @varshaprasad96 is planning to include that information as labels on the Helm release secret. Prior to the helm integration landing, we could include that information as BD labels or annotations.

I could have sworn we added BD annotations for that info already, but maybe not.

m1kola: @joelanford I think you did it in your PoC a while ago and I had a draft PR which we then closed in favour of [the digest idea](https://github.com/operator-framework/operator-controller/issues/458#issuecomment-1761199892), but we didn't get to implement the the digest idea.

PR #759, opened 2024-04-17
⚠️ (Partially) fix upgrade behaviour in `Extension`

m1kola:
# Description

With the current code there are a few issues:

1. `catalogfilter.HigherBundleVersion` is not really required as we already have more generic `catalogfilter.InMastermindsSemverRange`.
1. I think `catalogfilter.HigherBundleVersion` allows upgrades from one major version to another major version, etc. This is not the same behaviour as in `ClusterExtension`
1. We do not seem to support legacy `replaces` here.

This PR solves the first two, but the third still needs to be addressed.

How it currently works in `ClusterExtension`: https://github.com/operator-framework/operator-controller/blob/52cb7eb1a650731f2e16253a7f0f0ba25ea3af70/docs/drafts/upgrade-support.md

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/controllers/extension_controller.go
+++ b/internal/controllers/extension_controller.go
@@ -512,7 +512,14 @@ func (r *ExtensionReconciler) resolve(ctx context.Context, extension ocv1alpha1.
 		}
 		if installedVersionSemver != nil {
 			installedVersion = installedVersionSemver.String()
-			predicates = append(predicates, catalogfilter.HigherBundleVersion(installedVersionSemver))
+
+			// Based on installed version create a caret range comparison constraint
+			// to allow only minor and patch version as successors.
+			wantedVersionRangeConstraint, err := mmsemver.NewConstraint(fmt.Sprintf("^%s", installedVersion))
+			if err != nil {
+				return nil, err
+			}
+			predicates = append(predicates, catalogfilter.InMastermindsSemverRange(wantedVersionRangeConstraint))
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


internal/controllers/extension_controller_test.go
```diff
@@ -204,7 +204,7 @@ func TestExtensionResolve(t *testing.T) {
 			extNN := types.NamespacedName{Name: ext.GetName(), Namespace: ext.GetNamespace()}
 
 			res, reconcileErr := reconciler.Reconcile(ctx, ctrl.Request{NamespacedName: extNN})
-			assert.Equal(t, reconcileErr, tc.wantErr)
+			assert.Equal(t, tc.wantErr, reconcileErr)
```
m1kola, line 207: Should be expected and then actual.

internal/controllers/extension_controller.go
```diff
@@ -512,7 +512,14 @@ func (r *ExtensionReconciler) resolve(ctx context.Context, extension ocv1alpha1.
 		}
 		if installedVersionSemver != nil {
 			installedVersion = installedVersionSemver.String()
-			predicates = append(predicates, catalogfilter.HigherBundleVersion(installedVersionSemver))
```
m1kola, line 515: With the current code there are a few issues:

1. `catalogfilter.HigherBundleVersion` is not really required as we already have more generic `catalogfilter.InMastermindsSemverRange`.
1. I think `catalogfilter.HigherBundleVersion` allows upgrades from one major version to another major version, etc. This is not the same behaviour as in `ClusterExtension`
1. We do not seem to support legacy `replaces` here.

This PR solves first two. Third still need to be addressed.

How it currently works in `ClusterExtension`: https://github.com/operator-framework/operator-controller/blob/52cb7eb1a650731f2e16253a7f0f0ba25ea3af70/docs/drafts/upgrade-support.md

tmshort (approved):
/lgtm

PR #763, opened 2024-04-17
✨ Set up right watches and all labels to postrenderer

varshaprasad96:
I don't think there is need to list helm releases, to find metadata. The helm install code, in the `getReleaseState` method does comparison of the templated and existing chart and finds if changes are needed. 

Additionally, based on the [documentation](https://helm.sh/docs/topics/advanced/#storage-backends), helm by default creates and manages secrets for each release, and if we want pass any additional sensitive data to the chart, we need to do so while templating. Which means, we can use the default release which the helm creates and store bundle version and name through annotations in the chart. 

# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/controllers/clusterextension_controller.go
+++ b/internal/controllers/clusterextension_controller.go
@@ -35,6 +35,7 @@ import (
 	"helm.sh/helm/v3/pkg/postrender"
 	"helm.sh/helm/v3/pkg/release"
 	"helm.sh/helm/v3/pkg/storage/driver"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
@@ -49,6 +50,7 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 	crcontroller "sigs.k8s.io/controller-runtime/pkg/controller"
 	crhandler "sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/log"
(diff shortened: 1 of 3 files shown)


internal/rukpak/util/labels.go
```diff
@@ -1,6 +1,10 @@
...
 
 const (
-	CoreOwnerKindKey = "core.rukpak.io/owner-kind"
-	CoreOwnerNameKey = "core.rukpak.io/owner-name"
+	CoreOwnerKindKey          = "core.clusterextension.io/owner-kind"
+	CoreOwnerNameKey          = "core.clusterextension.io/owner-name"
+	ResolvedBundleCatalogID   = "core.clusterextension.io/catalog-id"
+	ResolvedbundlePackageName = "core.clusterextension.io/package-name"
+	ResolvedbundleName        = "core.clusterextension.io/bundle-name"
+	ResolvedbundleVersion     = "core.clusterextension.io/bundle-version"
```
tmshort, line 9: I don't believe we "own" the `clusterextension.io` domain, but this is a PoC...

internal/controllers/clusterextension_controller.go
```diff
@@ -460,6 +471,58 @@ func (r *ClusterExtensionReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return nil
 }
 
+func mapOwneeToOwnerHandler(cl client.Client, log logr.Logger, owner client.Object) crhandler.EventHandler {
```
tmshort, line 474: It took me a second to "get" this name. Is "ownee" even a word? :)

tmshort (approved):

internal/rukpak/util/labels.go
joelanford: +1 these should definitely be updated before 1.0.0

I'd suggest `olm.operatorframework.io/*`

joelanford: Also a few nits on naming:
1. CamelCase words in the identifier (e.g. `Resolvedbundle` should be `ResolvedBundle`)
2. Consistently use `Key` suffix to all of the identifiers to clarify that these are to be used for label/annotation keys.
3. Drop `Core` prefix.

Overall, align the identifier name with the value of the constant. e.g.
   - `"core.clusterextension.io/catalog-id"` would map to `CatalogIDKey`
   - `"core.clusterextension.io/package-name"` would map to `PackageNameKey`
   - etc.

internal/controllers/clusterextension_controller.go
```diff
@@ -223,8 +231,11 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
 
 	post := &postrenderer{
 		labels: map[string]string{
-			util.CoreOwnerKindKey: ocv1alpha1.ClusterExtensionKind,
-			util.CoreOwnerNameKey: ext.GetName(),
+			util.CoreOwnerKindKey:          ocv1alpha1.ClusterExtensionKind,
+			util.CoreOwnerNameKey:          ext.GetName(),
+			util.ResolvedbundleName:        bundle.Name,
+			util.ResolvedbundlePackageName: bundle.Package,
+			util.ResolvedbundleVersion:     bundleVersion.String(),
```
joelanford, line 238: No catalogID here? Can we drop that constant?

internal/rukpak/util/labels.go
tmshort: This is a PoC branch, so it's not going into `main` as-is.

tmshort: https://github.com/operator-framework/operator-controller/pull/765

internal/controllers/clusterextension_controller.go
tmshort: https://github.com/operator-framework/operator-controller/pull/765

Issue #723, opened 2024-03-28
Support opt-in of orphaning on delete


varshaprasad96:
This is issue is to allow the bundle resources remain in cluster, intentionally when the respective owner API is set of deletion.

With respect to Extension API, that uses App CR under the hood - this is possible by setting `noopDelete: true` in the spec of App CR while creation. (ref: https://carvel.dev/kapp-controller/docs/v0.50.x/app-spec/)

PR #772, opened 2024-04-22
:sparkles: Add noopdelete option to extension api

gallettilance:
# Description

Solves #723 

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/api/v1alpha1/extension_types.go
+++ b/api/v1alpha1/extension_types.go
@@ -74,6 +74,12 @@ type ExtensionSpec struct {
 	// paused controls the management state of the extension. If the extension is paused, it will be ignored by the extension controller.
 	Paused bool `json:"paused,omitempty"`
 
+	//+kubebuilder:default:=false
+	//+kubebuilder:Optional
+	//
+	// NoopDelete controls what is done with the associated resources. If set to true they will not be deleted.
+	NoopDelete bool `json:"noopdelete,omitempty"`
+
 	//+kubebuilder:validation:MaxLength:=253
 	//+kubebuilder:validation:Pattern:=^[a-z0-9]+([\.-][a-z0-9]+)*$
 	//
--- a/internal/controllers/extension_controller.go
+++ b/internal/controllers/extension_controller.go
@@ -430,6 +430,7 @@ func (r *ExtensionReconciler) GenerateExpectedApp(o ocv1alpha1.Extension, bundle
 				"kapp": map[string]interface{}{},
 			},
 		},
+		"noopdelete": o.Spec.NoopDelete,
 	}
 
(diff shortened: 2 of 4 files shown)


internal/controllers/extension_controller.go
```diff
@@ -432,6 +432,10 @@ func (r *ExtensionReconciler) GenerateExpectedApp(o ocv1alpha1.Extension, bundle
 		},
 	}
 
+	if o.Spec.NoopDelete {
```
joelanford, line 435: No need for a conditional here, right?

grokspawn:
I ✨ the PR title to pass the pr-title workflow, based on PR contents.  This repo enforces naming along these lines: 
    * 🐛 (:bug:, patch/bug fix)
    * 📖 (:book:, docs)
    * 🌱 (:seedling:, other)
    
   If that's not correct, please choose the option that best fits.

tmshort:
`make verify` needs to be run before commiting.

joelanford:
I'm curious if there's a reason we're adding a new feature to the extension API, given that we're planning to focus on the cluster extension API and temporarily remove this API.

Should we just make this change directly on the ClusterExtension instead?

api/v1alpha1/extension_types.go
```diff
@@ -74,6 +74,9 @@ type ExtensionSpec struct {
 	// paused controls the management state of the extension. If the extension is paused, it will be ignored by the extension controller.
 	Paused bool `json:"paused,omitempty"`
 
+	// NoopDelete controls what is done with the associated resources. If set to true they will not be deleted.
+	NoopDelete bool `json:"noopdelete,omitempty"`
```
varshaprasad96, line 78: Nit: We could add validation to this with KB markers, setting it to optional, so we don't worry about the default value, then directly assign it to App's spec.

varshaprasad96 (commented):
Looks good! /lgtm from my end, after addressing Joe's review. 

Regarding surfacing this on the status - looking deeper, it doesn't seem necessary imo. We don't specify anything explicitly for paused state either. Also this info may be useful only when the user tries to delete the relevant extension, at which point we are not going to persist the object itself. Just cross-checked, looks like [PackageInstall](https://github.com/carvel-dev/kapp-controller/blob/a8f75d9c13baa2f63c43a1aa6da845458091629f/pkg/apis/packaging/v1alpha1/package_install.go#L103) doesn't have it too.

api/v1alpha1/extension_types.go
gallettilance: done :)

joelanford:
Close or update/rebase for `ClusterExtension`?

PR #774, opened 2024-04-22
✨ Add `ValidatingAdmissionPolicy` to enforce package uniqueness across `ClusterExtension`

m1kola:
# Description

To enforce package uniqueness across `ClusterExtension`.

We currently have this enforcement on the operator-controller level, but with Deppy removal in #758 we want to shift it into API server.

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/config/admission/admission.yaml
+++ b/config/admission/admission.yaml
@@ -1,0 +1,37 @@
+apiVersion: admissionregistration.k8s.io/v1beta1
+kind: ValidatingAdmissionPolicy
+metadata:
+  name: "clusterextensions-package-uniqueness"
+spec:
+  failurePolicy: Fail
+  paramKind:
+    apiVersion: olm.operatorframework.io/v1alpha1
+    kind: ClusterExtension
+  matchConstraints:
+    resourceRules:
+    - apiGroups:   ["olm.operatorframework.io"]
+      apiVersions: ["v1alpha1"]
+      operations:  ["CREATE", "UPDATE"]
+      resources:   ["clusterextensions"]
+  matchConditions:
+    # Only apply the policy when the request operation is CREATE
+    # or when the package is being changed
+    - name: 'only-create-or-package-change'
+      expression: request.operation == 'CREATE' || oldObject.spec.packageName != object.spec.packageName
+  validations:
+    - expression: object.spec.packageName != params.spec.packageName
+      messageExpression: "'Package \"' + string(object.spec.packageName) + '\" is already installed via ClusterExtension \"' +  string(params.metadata.name) + '\"'"
+      reason: Invalid
+
+---
+
+apiVersion: admissionregistration.k8s.io/v1beta1
+kind: ValidatingAdmissionPolicyBinding
+metadata:
+  name: "clusterextensions-package-uniqueness-binding"
+spec:
+  policyName: "clusterextensions-package-uniqueness"
+  validationActions: [Deny]
+  paramRef:
+    parameterNotFoundAction: Allow
+    selector: {}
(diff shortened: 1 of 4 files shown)


config/admission/admission.yaml
```diff
@@ -0,0 +1,32 @@
...
+      reason: Invalid
+
+---
+
+apiVersion: admissionregistration.k8s.io/v1beta1
+kind: ValidatingAdmissionPolicyBinding
+metadata:
+  name: "clusterextensions-package-uniqueness-binding.olm.operatorframework.io"
+spec:
+  policyName: "clusterextensions-package-uniqueness.olm.operatorframework.io"
```
m1kola, line 28: I'll keep it in draft for now because `kustomize` currently doesn't handle this (even in the latest versions it seems).

I'll report an issue and see what else I can do to make it work

m1kola: Replacing this:
https://github.com/operator-framework/operator-controller/blob/652e4bca63f9e0a04973b5a1ef7ec03ca0cf330b/config/default/kustomization.yaml#L4-L9

With something like this might work:

```yaml
transformers:
- |-
  apiVersion: builtin
  kind: PrefixSuffixTransformer
  metadata:
    name: prefix-all-names
  prefix: operator-controller-
  fieldSpecs:
    - path: metadata/name
    - group: admissionregistration.k8s.io
      kind: ValidatingAdmissionPolicyBinding
      path: spec/policyName
```

m1kola: Reported: https://github.com/kubernetes-sigs/kustomize/issues/5674

Also updated the PR with a workaround.

m1kola:
Keeping this in draft for now. It seems like tilt job needs to be updated in addition to https://github.com/operator-framework/operator-controller/pull/771

config/admission/admission.yaml
joelanford: That workaround looks like it worked!

Another way to do this that I've seen is to use `nameReference` in a kustomize configuration file (explained fairly well here: https://www.innoq.com/en/blog/2022/07/advanced-kustomize-features/#supportcrds)

m1kola: Nice! I think `nameReference` looks slightly better. I'll use it instead.

config/admission/admission.yaml
```diff
@@ -0,0 +1,32 @@
...
+    - expression: "object.spec.packageName != params.spec.packageName"
+      messageExpression: "'Package \"' + string(object.spec.packageName) + '\" is already installed via ClusterExtension \"' +  string(params.metadata.name) + '\"'"
+      reason: Invalid
+
+---
+
+apiVersion: admissionregistration.k8s.io/v1beta1
+kind: ValidatingAdmissionPolicyBinding
+metadata:
+  name: "clusterextensions-package-uniqueness-binding.olm.operatorframework.io"
```
m1kola, line 26: After kustomize it ends up being `operator-controller-clusterextensions-package-uniqueness-binding.olm.operatorframework.io`. I think I'll drop `.olm.operatorframework.io` part to make it a bit shorter.

Same with `ValidatingAdmissionPolicy`

config/admission/admission.yaml
```diff
@@ -0,0 +1,32 @@
...
+    apiVersion: olm.operatorframework.io/v1alpha1
+    kind: ClusterExtension
+  matchConstraints:
+    resourceRules:
+    - apiGroups:   ["olm.operatorframework.io"]
+      apiVersions: ["v1alpha1"]
+      operations:  ["CREATE", "UPDATE"]
+      resources:   ["clusterextensions"]
+  validations:
+    - expression: "object.spec.packageName != params.spec.packageName"
```
m1kola, line 17: Now need to sort out editing of the existing objects

Issue #734, opened 2024-04-04
[epic] ClusterExtension supports simple registry+v1 bundles only


joelanford:
### Components
- [x] [Brief](https://docs.google.com/document/d/1Bwhv00KtFfB5JQ_11JAN00YnyehzGPIyYb3mdnc9GxM/edit?usp=drive_link)
- [x] #779
- [x] #815
- [x] #816

m1kola:
This is currently blocked by #846: to meet the requirement about supported install modes we need to bump rukpak so we can benefit from [this fix](https://github.com/operator-framework/rukpak/pull/892). We can not bump rukpak while work on helm integration is still ongoing.

m1kola:
Closing this epic: all relevant issues are done.

PR #781, opened 2024-04-24
⚠️ Disallow bundles with dependencies and constraints

### LINKED ISSUE #779: Fail installation attempts for registry+v1 bundles with expressed dependencies

m1kola:
# Description

Draft for the changes proposed in [this doc](https://docs.google.com/document/d/1Bwhv00KtFfB5JQ_11JAN00YnyehzGPIyYb3mdnc9GxM/edit): disallowing registry+v1 bundles with dependencies.

Closes #779

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/controllers/clusterextension_controller.go
+++ b/internal/controllers/clusterextension_controller.go
@@ -31,6 +31,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/utils/ptr"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -40,6 +41,7 @@ import (
 
 	catalogd "github.com/operator-framework/catalogd/api/core/v1alpha1"
 	"github.com/operator-framework/operator-registry/alpha/declcfg"
+	"github.com/operator-framework/operator-registry/alpha/property"
 	rukpakv1alpha2 "github.com/operator-framework/rukpak/api/v1alpha2"
 
 	ocv1alpha1 "github.com/operator-framework/operator-controller/api/v1alpha1"
@@ -140,6 +142,13 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
 		setDeprecationStatusesUnknown(&ext.Status.Conditions, "deprecation checks have not been attempted as installation has failed", ext.GetGeneration())
 		return ctrl.Result{}, err
 	}
+
+	if err := r.validateBundle(bundle); err != nil {
+		setInstalledStatusConditionFailed(&ext.Status.Conditions, err.Error(), ext.GetGeneration())
+		setDeprecationStatusesUnknown(&ext.Status.Conditions, "deprecation checks have not been attempted as installation has failed", ext.GetGeneration())
+		return ctrl.Result{}, err
+	}
+
 	bundleProvisioner, err := mapBundleMediaTypeToBundleProvisioner(mediaType)
 	if err != nil {
 		setInstalledStatusConditionFailed(&ext.Status.Conditions, err.Error(), ext.GetGeneration())
@@ -276,6 +285,25 @@ func (r *ClusterExtensionReconciler) installedBundle(ctx context.Context, allBun
 	})
 
 	return resultSet[0], nil
+}
+
+func (r *ClusterExtensionReconciler) validateBundle(bundle *catalogmetadata.Bundle) error {
+	unsupportedProps := sets.New(
+		property.TypePackageRequired,
+		property.TypeGVKRequired,
+		property.TypeConstraint,
+	)
+	for i := range bundle.Properties {
+		if unsupportedProps.Has(bundle.Properties[i].Type) {
+			return fmt.Errorf(
+				"bundle %q has a dependency declared via property %q which is currently not supported",
+				bundle.Name,
+				bundle.Properties[i].Type,
+			)
+		}
+	}
+
+	return nil
 }
 
 func mapBDStatusToInstalledCondition(existingTypedBundleDeployment *rukpakv1alpha2.BundleDeployment, ext *ocv1alpha1.ClusterExtension, bundle *catalogmetadata.Bundle) {
--- a/internal/controllers/clusterextension_registryv1_validation_test.go
+++ b/internal/controllers/clusterextension_registryv1_validation_test.go
@@ -1,0 +1,135 @@
+package controllers_test
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	apimeta "k8s.io/apimachinery/pkg/api/meta"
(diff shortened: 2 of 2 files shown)


internal/controllers/clusterextension_controller.go
```diff
@@ -219,6 +227,33 @@ func (r *ClusterExtensionReconciler) variables(ctx context.Context) ([]deppy.Var
...
+	if err != nil {
+		return fmt.Errorf("unable to determine type of bundle %q: %s", bundle.Name, err)
+	}
+
+	if mediaType != catalogmetadata.MediaTypeRegistry {
+		// We only validate registry+v1 bundles
+		return nil
+	}
+
+	dependencyPropertyTypes := []string{property.TypePackageRequired, property.TypeGVKRequired, "olm.constraint"}
```
m1kola, line 241: Find a constant for `"olm.constraint"`.

internal/controllers/clusterextension_controller.go
```diff
@@ -165,6 +166,13 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
 		setDeprecationStatusesUnknown(&ext.Status.Conditions, "deprecation checks have not been attempted as installation has failed", ext.GetGeneration())
 		return ctrl.Result{}, err
 	}
+
+	if err := r.validateBundle(bundle); err != nil {
```
m1kola, line 170: 1. Need tests
2. Need to find a real world example of `"olm.constraint"` usage.

grokspawn:
closes https://github.com/operator-framework/operator-controller/issues/734

internal/controllers/clusterextension_controller.go
```diff
@@ -219,6 +227,33 @@ func (r *ClusterExtensionReconciler) variables(ctx context.Context) ([]deppy.Var
...
+		// We only validate registry+v1 bundles
+		return nil
+	}
+
+	dependencyPropertyTypes := []string{property.TypePackageRequired, property.TypeGVKRequired, "olm.constraint"}
+	for i := range bundle.Properties {
+		for _, propType := range dependencyPropertyTypes {
+			if bundle.Properties[i].Type == propType {
+				return fmt.Errorf(
+					"bundle %q has one a dependency declared via one of the properties: %s",
```
m1kola, line 246:
```suggestion
					"bundle %q has a dependency declared via one of the properties: %s",
```

internal/controllers/clusterextension_controller.go
```diff
@@ -278,6 +286,36 @@ func (r *ClusterExtensionReconciler) installedBundle(ctx context.Context, allBun
...
+func (r *ClusterExtensionReconciler) validateBundle(bundle *catalogmetadata.Bundle) error {
+	mediaType, err := bundle.MediaType()
+	if err != nil {
+		return fmt.Errorf("unable to determine type of bundle %q: %s", bundle.Name, err)
+	}
+
+	if mediaType != catalogmetadata.MediaTypeRegistry {
+		// We only validate registry+v1 bundles
+		return nil
+	}
```
m1kola, line 298: Wondering if we actually want to have this check... Maybe we should just disallow any bundles that express dependencies?

m1kola: > Is it possible for a simple bundle to express dependencies?

Do you mean `plain+v0`? I'm not 100% sure, but I think they can have all the FBC properties.

grokspawn: I thought `plain+v0` were bare manifests, so would not present with an `olm.bundle` and properties like FBC, right?

m1kola: The bundle itself is bare manifests, but the catalog will still have metadata. For example, take a look at [this demo](https://github.com/operator-framework/operator-controller/blob/main/docs/demos/coast-to-coast-q4-2023.md#create-an-initial-file-based-catalog-fbc-image).

Or am I missing something?

grokspawn: You're good.   I forgot that there is still FBC associated with it, and that the v1-only `olm.bundle.mediatype` property is exercised to determine (and all my time up to now has been in v0 catalogs). 

In that case, you've talked me into it: I'd support either expanding this to remove the conditional and add tests, or handling plain+v0 bundles in a follow-up PR.

everettraven: My understanding is the same as @m1kola 

The bundle media type in this case is, IMO, unnecessary to consider and should instead assume that the input is valid FBC and therefore evaluate only the FBC contents for validation

everettraven: Although if I am misunderstanding the intention of this PR and the goal is to enforce that only `registry+v1` bundles will ever be allowed then adding a constraint to the validation to reject any other bundle media type

internal/controllers/clusterextension_controller.go
```diff
@@ -278,6 +286,36 @@ func (r *ClusterExtensionReconciler) installedBundle(ctx context.Context, allBun
...
+
+	for i := range bundle.Properties {
+		for _, propType := range []string{
+			property.TypePackageRequired,
+			property.TypeGVKRequired,
+			property.TypeConstraint,
+		} {
```
everettraven, line 305: Could we instead make this a map and see if the bundle property type exists in the map we return the error?

This might be a premature and low-value optimization, but it reduces the time complexity from O(n*m) to O(n) (where n is the number of bundle properties and m is the unsupported properties) and is, IMO, easier to read

everettraven: This also might be a good use of the `sets` package since we really only care about the key (property type) and not the value in the map.

internal/controllers/clusterextension_controller.go
everettraven: Another thing to note, the way the media type property was designed was such that registry+v1 bundles would not need to be changed. If the media type property does not exist it is assumed it is a registry+v1 bundle

internal/controllers/clusterextension_controller.go
m1kola: @everettraven  I'm not a big fan of sets since they are just very slim wrapper around a no value map `map[T]struct{}{}`, but happy to switch to sets if people find it more clean. I'll do it in the next update.

> This might be a premature and low-value optimization, but it reduces the time complexity from O(n*m) to O(n)

Despite the nested loop this is considered O(n) since number of iterations in the inner loop is always 3 and does not depend on the input. Same as the following will be O(n):

```go
for i := range bundle.Properties {
	if bundle.Properties[i].Type == property.TypePackageRequired {
		return // ...
	}
	if bundle.Properties[i].Type == property.TypeGVKRequired {
		return // ...
	}
	if bundle.Properties[i].Type == property.TypeConstraint {
		return // ...
	}
}
```

internal/controllers/clusterextension_controller.go
m1kola: After [this discussion](https://redhat-internal.slack.com/archives/C06KP34REFJ/p1714659939965849) in Slack we decided to get rid of the media type check. If any bundle installed via `ClusterExtension` be it `registry+v1` or `plain+v0` or whether else expresses any of the properties - we fail.

internal/controllers/clusterextension_controller.go
everettraven: Fair enough on the time complexity. I guess I was approaching it from the perspective that the unsupported properties is an arbitrary value so while that is the case for this iteration it could shrink/grow as much or as little as we see fit. Using a map ensures that it is always an O(1) lookup.

everettraven (approved):

PR #790, opened 2024-04-26
🐛 Use bin-dir override on setup_envtest

trgeiger:
Rather than change XDG_DATA_HOME, since many distributions do not explicitly define XDG_DATA_HOME but rather fallback to the default, use a new optional variable SETUP_ENVTEST_BIN_DIR to override the setup_envtest binary directory.

Setting XDG_DATA_HOME in the Makefile was breaking the kind cluster setup on Fedora which just uses the default fallback value for XDG_DATA_HOME.

# Description

## Reviewer Checklist

- [ ] API Go Documentation


joelanford (approved):

joelanford:
Thanks!

PR #791, opened 2024-04-26
🌱 Use server side apply to update test resource

m1kola:
# Description

Deals with a flake such as this:

```
go test -tags upstream -v ./test/e2e/...
=== RUN   TestClusterExtensionPackageUniqueness
    cluster_extension_admission_test.go:30: create first resource
    cluster_extension_admission_test.go:42: create second resource with the same package as the first resource
    cluster_extension_admission_test.go:54: create second resource with different package
    cluster_extension_admission_test.go:66: update second resource with package which already exists on the cluster
    cluster_extension_admission_test.go:69: 
        	Error Trace:	/home/runner/work/operator-controller/operator-controller/test/e2e/cluster_extension_admission_test.go:69
        	Error:      	Error "Operation cannot be fulfilled on clusterextensions.olm.operatorframework.io \"test-extension-74hdj\": the object has been modified; please apply your changes to the latest version and try again" does not contain "Package \"package1\" is already installed via ClusterExtension \"test-extension-first\""
        	Test:       	TestClusterExtensionPackageUniqueness
--- FAIL: TestClusterExtensionPackageUniqueness (2.03s)
```

Was introduced in https://github.com/operator-framework/operator-controller/pull/785.

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/test/e2e/cluster_extension_admission_test.go
+++ b/test/e2e/cluster_extension_admission_test.go
@@ -9,12 +9,14 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	ocv1alpha1 "github.com/operator-framework/operator-controller/api/v1alpha1"
 )
 
 func TestClusterExtensionPackageUniqueness(t *testing.T) {
 	ctx := context.Background()
+	fieldOwner := client.FieldOwner("operator-controller-e2e")
 
 	deleteClusterExtension := func(clusterExtension *ocv1alpha1.ClusterExtension) {
 		require.NoError(t, c.Delete(ctx, clusterExtension))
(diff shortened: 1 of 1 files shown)


test/e2e/cluster_extension_admission_test.go
```diff
@@ -64,12 +66,33 @@ func TestClusterExtensionPackageUniqueness(t *testing.T) {
...
 	t.Log("update second resource with package which already exists on the cluster")
-	clusterExtension2.Spec.PackageName = firstResourcePackageName
-	err = c.Update(ctx, clusterExtension2)
+	intent := &ocv1alpha1.ClusterExtension{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: ocv1alpha1.GroupVersion.String(),
+			Kind:       "ClusterExtension",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: clusterExtension2.Name,
```
perdasilva, line 75: what should be the value of `clusterExtension2.Name` at this point? I think it might be "" because in its definition above uses `GenerateName` (or am I missing something?)

m1kola: After `c.Create(ctx, clusterExtension2)` we would have `Name` populated with the real one from the server.

test/e2e/cluster_extension_admission_test.go
```diff
@@ -9,12 +9,14 @@ import (
...
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	ocv1alpha1 "github.com/operator-framework/operator-controller/api/v1alpha1"
 )
 
 func TestClusterExtensionPackageUniqueness(t *testing.T) {
 	ctx := context.Background()
+	fieldOwner := client.FieldOwner("operator-controller-e2e")
```
perdasilva, line 19: I think using SSA is valid and maybe we should have a wider discussion about how best to use it in e2e tests. But, we shouldn't be seeing conflicts of the type in the issue coming out of our test code. If we do it's because we forgot to get the latest from the server after a create or update. SSA is nice because it makes things feel more natural for e2e tests. Do we need to respect any conventions when using SSA? Should we switch all mutating interactions in e2e over to SSA?

test/e2e/cluster_extension_admission_test.go
perdasilva: Right! I was thinking of `c.Create` as non-mutating. But, of course, it makes sense to update the input, since the api server will anyway return the resource on create.

test/e2e/cluster_extension_admission_test.go
```diff
@@ -64,12 +66,33 @@ func TestClusterExtensionPackageUniqueness(t *testing.T) {
 	defer deleteClusterExtension(clusterExtension2)
 
 	t.Log("update second resource with package which already exists on the cluster")
-	clusterExtension2.Spec.PackageName = firstResourcePackageName
```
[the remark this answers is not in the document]
perdasilva: At the time of writing I didn't factor in the changes coming from the reconciler :face_palm:

perdasilva (approved):
lgtm - on a general note we may want to consider best practices (if any) around SSA that we should respect in the e2es

PR #794, opened 2024-04-26
🐛 More needed for podman workaround in kind-load (take 2)

bentito:
The commit adding the podman workaround for the kind-load target seems to need a bit more work:

It needs to have the tag name set before image-archive is saved

# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -153,9 +153,13 @@ e2e-coverage:
 kind-load: $(KIND) #EXHELP Loads the currently constructed image onto the cluster.
 ifeq ($(CONTAINER_RUNTIME),podman)
 	@echo "Using Podman"
-	podman save $(IMG) -o $(IMG).tar
-	$(KIND) load image-archive $(IMG).tar --name $(KIND_CLUSTER_NAME)
-	rm $(IMG).tar
+	@DEPLOY_TEMPLATE_IMG_NAME=quay.io/operator-framework/operator-controller:devel; \
+	TMP_FILE=temp_image.tar; \
+	podman tag $(IMG) $$DEPLOY_TEMPLATE_IMG_NAME; \
+	echo "Saving image to temporary file $$TMP_FILE"; \
+	podman save $$DEPLOY_TEMPLATE_IMG_NAME -o $$TMP_FILE; \
+	$(KIND) load image-archive $$TMP_FILE --name $(KIND_CLUSTER_NAME); \
+	rm $$TMP_FILE
 else
 	@echo "Using Docker"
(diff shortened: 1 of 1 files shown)


joelanford (commented):
Meta-comment:

It is really difficult to support every permutation of developer environment, especially if we don't centralize these sorts of task definitions. Whatever we decide here would need to be duplicated in basically all of the active repos and then maintained.

If we don't test each configuration in CI, then they are extremely prone to rot, because the only one guaranteed to work is the one used in CI. If we _do_ test each configuration, then CI becomes more expensive AND it is hard for a developer making this kind of change to fix something about an environment they don't have setup locally.

IMO, if at all possible, it is better for the repo to be very opinionated on the environment and to have developers adapt to the repo instead of vice versa

Even Mac vs Linux is hard enough (and we don't explicitly test Mac in CI). With podman support, the matrix becomes linux/docker, linux/podman, mac/docker, mac/podman.

m1kola:
Could you please explain the issue in more detail? I'm using podman on Mac and do not observe issues with our `Makefile`. But to be fair - instead of passing `CONTAINER_RUNTIME` into `make` I have a symlink from `docker` to `podman` for convenience.

bentito:
> Meta-comment:
> 
> It is really difficult to support every permutation of developer environment, especially if we don't centralize these sorts of task definitions. Whatever we decide here would need to be duplicated in basically all of the active repos and then maintained.
> 
> If we don't test each configuration in CI, then they are extremely prone to rot, because the only one guaranteed to work is the one used in CI. If we _do_ test each configuration, then CI becomes more expensive AND it is hard for a developer making this kind of change to fix something about an environment they don't have setup locally.
> 
> IMO, if at all possible, it is better for the repo to be very opinionated on the environment and to have developers adapt to the repo instead of vice versa
> 
> Even Mac vs Linux is hard enough (and we don't explicitly test Mac in CI). With podman support, the matrix becomes linux/docker, linux/podman, mac/docker, mac/podman.

I'd argue our time is more expensive than CI time, most of us have Macs for local dev and I feel like we should be trying to use Podman if its diffs from Docker don't get in our way too badly. Also, CI likely is closer to Podman than Docker internally anyway because CRI-O is using buildah under the hood. If we can't add this conditional to make Podman easy and cheaper for devs to use, I guess I'll work out how to auto-apply/delete before commit these kind of things so I can personally work better, but they're in conditionals, does it matter if the podman version is found to have rotted later? It can get fixed if someone is motivated and all can benefit in practice.

PR #803, opened 2024-04-29
🐛 Sanitize temporary filename for podman image load

trgeiger:
# Description
The podman-specific kind-load commands in the Makefile currently write the image archive to a temporary file named $IM but that by default includes slashes and a colon. My changes replace any slashes or colons with hyphens to prevent any errors trying to write that file.

Also, the naming doesn't really matter, we can just make it be a hard-coded "tmp.tar" or something.

Also, also, I saw the closed PR from earlier today and the discussion about not getting too into the weeds on podman/docker-specific configuration in the Makefile, but the current setup will cause that file-write error for anyone who does use CONTAINER_RUNTIME=podman with our default env variables, so I figured that was worth a quick fix.

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -154,15 +154,7 @@ e2e-coverage:
 
 .PHONY: kind-load
 kind-load: $(KIND) #EXHELP Loads the currently constructed image onto the cluster.
-ifeq ($(CONTAINER_RUNTIME),podman)
-	@echo "Using Podman"
-	podman save $(IMG) -o $(IMG).tar
-	$(KIND) load image-archive $(IMG).tar --name $(KIND_CLUSTER_NAME)
-	rm $(IMG).tar
-else
-	@echo "Using Docker"
-	$(KIND) load docker-image $(IMG) --name $(KIND_CLUSTER_NAME)
-endif
+	$(CONTAINER_RUNTIME) save $(IMG) | $(KIND) load image-archive /dev/stdin --name $(KIND_CLUSTER_NAME)
 
 kind-deploy: export MANIFEST="./operator-controller.yaml"
 kind-deploy: manifests $(KUSTOMIZE) #EXHELP Install controller and dependencies onto the kind cluster.


Makefile
```diff
@@ -156,9 +156,9 @@ e2e-coverage:
 kind-load: $(KIND) #EXHELP Loads the currently constructed image onto the cluster.
 ifeq ($(CONTAINER_RUNTIME),podman)
 	@echo "Using Podman"
-	podman save $(IMG) -o $(IMG).tar
-	$(KIND) load image-archive $(IMG).tar --name $(KIND_CLUSTER_NAME)
-	rm $(IMG).tar
+	$(CONTAINER_RUNTIME) save $(IMG) -o /tmp/$(shell echo $(IMG) | sed 's/[/:]/-/g').tar
+	$(KIND) load image-archive /tmp/$(shell echo $(IMG) | sed 's/[/:]/-/g').tar --name $(KIND_CLUSTER_NAME)
+	rm /tmp/$(shell echo $(IMG) | sed 's/[/:]/-/g').tar
```
m1kola, line 161: Would it make more sense to use `mktemp`?

trgeiger: Just with the default randomly generated filename to simplify all of this?

tmshort: Or even just a shell variable for the filename.

trgeiger: Ahh, yeah good idea lol

Makefile
```diff
@@ -156,9 +156,9 @@ e2e-coverage:
 kind-load: $(KIND) #EXHELP Loads the currently constructed image onto the cluster.
 ifeq ($(CONTAINER_RUNTIME),podman)
```
joelanford, line 157: Probably out of scope for what this PR is trying to accomplish (and I asked in the other now-closed PR), but would be possible to use this podman procedure for `docker` as well? That way we don't have to maintain two different implementations of kind-load.

trgeiger: I would be find doing that, I can just remove the ifeq bits and leave only the first block with the temp archive. That should always work regardless of if the user is using podman, docker, or podman with the "podman-docker" alias package.

Makefile
```diff
@@ -154,15 +154,10 @@ e2e-coverage:
...
-	$(KIND) load image-archive $(IMG).tar --name $(KIND_CLUSTER_NAME)
-	rm $(IMG).tar
-else
-	@echo "Using Docker"
-	$(KIND) load docker-image $(IMG) --name $(KIND_CLUSTER_NAME)
-endif
+	TMP_FILE=$(shell mktemp -t operator-controller-image-archive-XXXX.tar) ;\
+	$(CONTAINER_RUNTIME) save $(IMG) -o $$TMP_FILE ;\
+	$(KIND) load image-archive $$TMP_FILE --name $(KIND_CLUSTER_NAME) ;\
+	rm $$TMP_FILE
```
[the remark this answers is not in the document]
trgeiger: If you or someone else has docker available, would be nice just to double check this works as expected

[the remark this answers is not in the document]
trgeiger: oh but the CI probably uses docker doesn't it?

joelanford (approved):

PR #805, opened 2024-04-30
:warning: Remove install references to rukpak

tmshort:
https://issues.redhat.com/browse/OPRUN-3294

# Description

## Reviewer Checklist

- [ ] API Go Documentation


Makefile
```diff
@@ -167,7 +166,7 @@ endif
 kind-deploy: export MANIFEST="./operator-controller.yaml"
 kind-deploy: manifests $(KUSTOMIZE) #EXHELP Install controller and dependencies onto the kind cluster.
 	$(KUSTOMIZE) build $(KUSTOMIZE_BUILD_DIR) > operator-controller.yaml
-	envsubst '$$CATALOGD_VERSION,$$CERT_MGR_VERSION,$$KAPP_VERSION,$$RUKPAK_VERSION,$$MANIFEST' < scripts/install.tpl.sh | bash -s
+	envsubst '$$CATALOGD_VERSION,$$CERT_MGR_VERSION,$$KAPP_VERSION,$$MANIFEST' < scripts/install.tpl.sh | bash -s
```
varshaprasad96, line 169: Now that we are not using Rukpak, do we need cert-manager? I am not sure why Rukpak needed cert-manager (would have to check), but if we are not using it here we can remove it too.

varshaprasad96 (approved):
Just one question on cert-manager, we can do it follow up. 

/lgtm
/approve

Makefile
everettraven: I'd recommend keeping it (can comment it out if desired) since we are slated to add HTTPS encryption to catalogd for v1.0.0 and that will require certificate management

everettraven: I also see an argument for removing it until then, but my thinking is that with us moving so quickly on stuff it would probably reduce long term effort by keeping it in place, even if it is unused temporarily.

varshaprasad96: Okay, let's just keep it for now. We can add a comment and revisit this later (when we get to merging this to main).

PR #810, opened 2024-05-01
⚠ ClusterExtension: add installNamespace, remove watchNamespaces

joelanford:
# Description

In `ClusterExtension`:
- Add `spec.installNamespace` as a required field. This will be used to configure the context of the client that templates and applies the bundle contents. This will also be necessary for the introduction of a `spec.serviceAccount` field that operator-controller will use to perform all interactions with the apiserver for that ClusterExtension.
- Remove `spec.watchNamespaces` and hardcode all registry+v1 based bundles to watch all namespaces (or fail if they don't support AllNamespaces mode)

## Reviewer Checklist

- [ ] API Go Documentation


joelanford:
/hold

I want to add a few more tests that are specific to the use of the new InstallNamespace field.

internal/controllers/clusterextension_admission_test.go
```diff
@@ -186,6 +190,52 @@ func TestClusterExtensionAdmissionChannel(t *testing.T) {
...
+	testCases := []struct {
+		name             string
+		installNamespace string
+		errMsg           string
+	}{
+		{"just alphanumeric", "justalphanumberic1", ""},
+		{"hypen-separated", "hyphenated-name", ""},
+		{"no install namespace", "", regexMismatchError},
+		{"dot-separated", "dotted.name", regexMismatchError},
+		{"includes version", "channel-has-version-1.0.1", regexMismatchError},
```
everettraven, line 206: We already check for dot separated, I don't think we need to explicitly check for "includes version" for install namespaces

```suggestion
```

everettraven (approved):
Looks good. One minor nit, but IMO not something that is worth holding this PR on and will leave up to you if you'd like to address.

internal/controllers/clusterextension_admission_test.go
joelanford: Yep good point, this was a copy/paste job from the channel name tests, and this one does not seem relevant here. I'll remove.

everettraven (approved):

PR #811, opened 2024-05-01
✨ Helm poc unpack image

dtfranz:
Builds and loads the unpacker binary into the operator-controller image so we don't need to use the rukpak image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,14 +1,21 @@
 # Note: This dockerfile does not build the binaries 
 # required and is intended to be built only with the
 # 'make build' or 'make release' targets.
+# Stage 1:
+FROM gcr.io/distroless/static:debug-nonroot AS builder
+
+# Stage 2: 
 FROM gcr.io/distroless/static:nonroot
+
+# Grab the cp binary so we can cp the unpack
+# binary to a shared volume in the bundle image
+COPY --from=builder /busybox/cp /cp
 
 WORKDIR /
 
 COPY manager manager
+COPY unpack unpack
 
 EXPOSE 8080
 
 USER 65532:65532
-
-ENTRYPOINT ["/manager"]
--- a/Makefile
+++ b/Makefile
@@ -209,7 +209,10 @@ export GO_BUILD_GCFLAGS  ?= all=-trimpath=${PWD}
 export GO_BUILD_TAGS     ?= upstream
 export GO_BUILD_FLAGS    ?=
 
-BUILDCMD = go build $(GO_BUILD_FLAGS) -tags '$(GO_BUILD_TAGS)' -ldflags '$(GO_BUILD_LDFLAGS)' -gcflags '$(GO_BUILD_GCFLAGS)' -asmflags '$(GO_BUILD_ASMFLAGS)' -o $(BUILDBIN)/manager ./cmd/manager
+BINARIES=manager unpack
+
+$(BINARIES):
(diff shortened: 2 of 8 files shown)


config/manager/manager.yaml
```diff
@@ -55,6 +55,8 @@ spec:
         - "--health-probe-bind-address=:8081"
         - "--metrics-bind-address=127.0.0.1:8080"
         - "--leader-elect"
+        # TODO Need a way to match the container image to the unpack image
+        - "--unpack-image=quay.io/operator-framework/operator-controller:devel"
```
varshaprasad96: How about:
```suggestion
        - "--unpack-image=quay.io/operator-framework/operator-controller-unpack:devel"
```

dtfranz: Are you suggesting this change in order to more explicitly indicate that this string is being replaced, or in order to use a separate image? 
If the former, the latest commit should take care of that, but if the latter, I think it's better to have the `unpack` binary in the same image as the `manager` so that we don't have to build, push, and manage two separate images. Rukpak also used this strategy FWIW.

varshaprasad96: Please ignore - Just realised from the Dockerfile that operator controller and unpack image are going to be the same.

Makefile
```diff
@@ -203,22 +203,24 @@ export GO_BUILD_GCFLAGS  ?= all=-trimpath=${PWD}
...
-	$(BUILDCMD)
+go-build-local: $(BINARIES)
 
 .PHONY: build-linux go-build-linux
 build-linux: build-deps go-build-linux #EXHELP Build manager binary for GOOS=linux and local GOARCH.
 go-build-linux: BUILDBIN = bin/linux
-go-build-linux:
-	GOOS=linux $(BUILDCMD)
+go-build-linux: GOOS=linux
+go-build-linux: $(BINARIES)
```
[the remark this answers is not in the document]
dtfranz: Just so we don't need to have duplicate build commands for `manager` and `unpack`. It's also a clean way to have the build commands run builds for each binary without needing to specify the list of binaries more than once.

config/manager/kustomization.yaml
```diff
@@ -6,3 +6,23 @@ images:
 - name: controller
   newName: quay.io/operator-framework/operator-controller
   newTag: devel
+replacements:
```
varshaprasad96, line 9: This is nice! I think in rukpak we had to manually edit the image in manager.

Dockerfile
```diff
@@ -6,9 +6,8 @@ FROM gcr.io/distroless/static:nonroot
 WORKDIR /
 
 COPY manager manager
+COPY unpack unpack
 
 EXPOSE 8080
 
 USER 65532:65532
-
-ENTRYPOINT ["/manager"]
```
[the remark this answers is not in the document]
varshaprasad96: If it's set in the deployment then let's just remove it.

varshaprasad96 (commented):
The changes look good! Thanks @dtfranz

Before merging, could you please test this and confirm if it works as expected locally as there are no tests in place for now.

tmshort:
You need to do a `go mod tidy`

tmshort:
You need to add an emoji to the start of the PR.
```
Please prefix the title of this PR with one of the following icons:

    * 🐛 (:bug:, patch/bug fix)
    * 📖 (:book:, docs)
    * 🌱 (:seedling:, other)

**Comment by tmshort** [comment]
Also, seeing this error during `make verify` in CI (but not local):
```
/bin/sh: 1: [[: not found
/bin/sh: 1: /home/runner: Permission denied

PR #812, opened 2024-05-01
:bug: [Fix] Constant reconciles due to Catalog's poll interval

varshaprasad96:
# Description

Due to the polling feature in the Catalog, the `lastPolledInterval` in the Catalog's status gets updated. Because of this there is a reconcile triggered every time the catalog is polled, even though there is no change in the resolved image ref.

This causes unnecessary continuous reconciles in the CE reconciler. To overcome this, a Predicate is added to accept update events only when there is a change in Resolved image reference in catalog's status.

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/controllers/clusterextension_controller.go
+++ b/internal/controllers/clusterextension_controller.go
@@ -32,8 +32,10 @@ import (
 	"k8s.io/utils/ptr"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	catalogd "github.com/operator-framework/catalogd/api/core/v1alpha1"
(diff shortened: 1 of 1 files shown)


internal/controllers/clusterextension_controller.go
```diff
@@ -409,6 +411,17 @@ func (r *ClusterExtensionReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 			handler.EnqueueRequestsFromMapFunc(clusterExtensionRequestsForCatalog(mgr.GetClient(), mgr.GetLogger()))).
+		WithEventFilter(predicate.Funcs{
+			UpdateFunc: func(ue event.UpdateEvent) bool {
+				oldObject, isOldCatalog := ue.ObjectOld.(*catalogd.Catalog)
+				newObject, isNewCatalog := ue.ObjectNew.(*catalogd.Catalog)
+
+				if !isOldCatalog || !isNewCatalog {
+					return true
+				}
+				return oldObject.Status.ResolvedSource.Image.Ref != newObject.Status.ResolvedSource.Image.Ref
```
joelanford, line 422: `.Status.ResolvedSource` and `.Status.ResolveSource.Image` are both pointers.

I assume we need to handle cases where either of those is `nil` for both the old and new objects?

internal/controllers/clusterextension_controller.go
```diff
@@ -409,6 +411,23 @@ func (r *ClusterExtensionReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+				oldObject, isOldCatalog := ue.ObjectOld.(*catalogd.Catalog)
+				newObject, isNewCatalog := ue.ObjectNew.(*catalogd.Catalog)
+
+				if !isOldCatalog || !isNewCatalog {
+					return true
+				}
+
+				if oldObject.Status.ResolvedSource != nil && newObject.Status.ResolvedSource != nil {
+					if oldObject.Status.ResolvedSource.Image != nil && newObject.Status.ResolvedSource.Image != nil {
+						return oldObject.Status.ResolvedSource.Image.Ref != newObject.Status.ResolvedSource.Image.Ref
```
everettraven, line 425: IIRC the `.status.resolvedSource.image.resolvedRef` is probably what we want here. See https://github.com/operator-framework/catalogd/blob/29870e83a0dd4b1d99d69796a3d4c61887737f0d/api/core/v1alpha1/catalog_types.go#L120-L121

tmshort (approved):
/lgtm
/approve

internal/controllers/clusterextension_controller.go
varshaprasad96: Thanks for pointing out @everettraven! Didn't know that existed.

PR #813, opened 2024-05-02
🌱 Makefile: fix issues with repeated evaluations, other minor improvements

joelanford:
# Description

I was experiencing extreme delays in running `make` for any target so finally decided to run it through `strace` to see what I could find. It turns out `make` was evaluating almost all of our variables many many times due to the use of `=` and `?=` instead of `:=`.

See https://www.gnu.org/software/make/manual/html_node/Flavors.html#Flavors for docs about the various flavors of variables.

While I was at it, I also cleaned up and improved a few other things:
- Auto-detect docker, fallback to podman, fail if neither are present
- Remove GO_BUILD_TAGS. The original intention was to have a way to inject extra or different code in other distributions of operator-controller. We've seen Red Hat's OCP distribution use carry patches in a fork instead.
- Remove unused or minimally useful variables. e.g. GOBIN, TESTDATA_DIR

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,15 +2,25 @@
 # Configuration Variables #
 ###########################
 # Image URL to use all building/pushing image targets
-export IMAGE_REPO ?= quay.io/operator-framework/operator-controller
-export IMAGE_TAG ?= devel
-export CERT_MGR_VERSION ?= v1.9.0
-export CATALOGD_VERSION ?= $(shell go list -mod=mod -m -f "{{.Version}}" github.com/operator-framework/catalogd)
-export KAPP_VERSION ?= $(shell go list -mod=mod -m -f "{{.Version}}" github.com/vmware-tanzu/carvel-kapp-controller)
-export RUKPAK_VERSION=$(shell go list -mod=mod -m -f "{{.Version}}" github.com/operator-framework/rukpak)
-export WAIT_TIMEOUT ?= 60s
-IMG?=$(IMAGE_REPO):$(IMAGE_TAG)
-TESTDATA_DIR := testdata
+ifeq ($(origin IMAGE_REPO), undefined)
+IMAGE_REPO := quay.io/operator-framework/operator-controller
+endif
+export IMAGE_REPO
+
+ifeq ($(origin IMAGE_TAG), undefined)
+IMAGE_TAG := devel
+endif
+export IMAGE_TAG
+
+IMG := $(IMAGE_REPO):$(IMAGE_TAG)
+
+
+# Define dependency versions (use go.mod if we also use Go code from dependency)
+export CERT_MGR_VERSION := v1.9.0
+export CATALOGD_VERSION := $(shell go list -mod=mod -m -f "{{.Version}}" github.com/operator-framework/catalogd)
+export KAPP_VERSION     := $(shell go list -mod=mod -m -f "{{.Version}}" github.com/vmware-tanzu/carvel-kapp-controller)
+export RUKPAK_VERSION   := $(shell go list -mod=mod -m -f "{{.Version}}" github.com/operator-framework/rukpak)
+export WAIT_TIMEOUT     := 60s
 
 # By default setup-envtest will write to $XDG_DATA_HOME, or $HOME/.local/share if that is not defined.
 # If $HOME is not set, we need to specify a binary directory to prevent an error in setup-envtest.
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -23,34 +33,28 @@ endif
 # bingo manages consistent tooling versions for things like kind, kustomize, etc.
 include .bingo/Variables.mk
 
-# ARTIFACT_PATH is the absolute path to the directory where the operator-controller e2e tests will store the artifacts
-# for example: ARTIFACT_PATH=/tmp/artifacts make test
```
bentito, line 27: losing the useless export seems good, but do we want to keep this bit of dev knowledge around somehow?

bentito:
/lgtm

Just had the one comment about removing a comment.

Makefile
```diff
@@ -2,15 +2,25 @@
...
+IMAGE_TAG := devel
+endif
+export IMAGE_TAG
+
+IMG := $(IMAGE_REPO):$(IMAGE_TAG)
+
+
+# Define dependency versions (use go.mod if we also use Go code from dependency)
+export CERT_MGR_VERSION := v1.9.0
+export CATALOGD_VERSION := $(shell go list -mod=mod -m -f "{{.Version}}" github.com/operator-framework/catalogd)
```
everettraven, line 20: How does this new variable assignment impact doing something like:
```sh
make run CATALOGD_VERSION=vX.Y.Z
```

IIUC using the `?=` variable assignment would only expand the variables if they were not already defined via something like the above example. Is overriding the values still possible with this new variable assignment?

Makefile
joelanford: There's nothing in the Makefile that uses it.

Just this: https://github.com/operator-framework/operator-controller/blob/f22ce65c3cfd7f548709a4730f5f83095847ff8b/.github/workflows/e2e.yaml#L31
And this: https://github.com/operator-framework/operator-controller/blob/f22ce65c3cfd7f548709a4730f5f83095847ff8b/test/e2e/cluster_extension_install_test.go#L380

Maybe we could just leave this comment in the Makefile next to the `e2e` target though?

Makefile
```diff
@@ -2,15 +2,25 @@
...
-export IMAGE_REPO ?= quay.io/operator-framework/operator-controller
-export IMAGE_TAG ?= devel
-export CERT_MGR_VERSION ?= v1.9.0
-export CATALOGD_VERSION ?= $(shell go list -mod=mod -m -f "{{.Version}}" github.com/operator-framework/catalogd)
-export KAPP_VERSION ?= $(shell go list -mod=mod -m -f "{{.Version}}" github.com/vmware-tanzu/carvel-kapp-controller)
-export RUKPAK_VERSION=$(shell go list -mod=mod -m -f "{{.Version}}" github.com/operator-framework/rukpak)
-export WAIT_TIMEOUT ?= 60s
-IMG?=$(IMAGE_REPO):$(IMAGE_TAG)
-TESTDATA_DIR := testdata
+ifeq (, $(IMAGE_REPO))
```
grokspawn, line 5: These questions are all for the changes to former `?=` statements.  

The gnu recommendation is to have `ifeq(test, expected)`, e.g.

```Makefile
ifeq ($(IMAGE_REPO),)
```

but AFAIK this should be fine.  It should be noted that this will trigger even if the test var is empty, which in this case would overwrite with the default value for the arg.  I can think of no case where we would want the variable to be defined-but-empty, so it's probably fine.
 
Any reason that this is superior to the conditional variable example:

```Makefile
 ifeq ($(origin IMAGE_REPO), undefined)
  IMAGE_REPO = blah
endif
```
?

Finally, this is supposed to be the semantic equivalent of `?=` and it seems reasonable to expect no performance benefit from this particular change.  If that's not the case, how big a performance impact are we seeing?

Makefile
joelanford: If we want the variables to be overridable, they would need this sort of construction:
```
ifeq (,$(VAR))
VAR := $(shell echo "Hello")
endif
```

Does the catalogd version need to be configurable like that? If so, I can change it. Let me know if there are other variables in the same boat that you see with the non-overridable assignment.

Makefile
joelanford: This is not great
```
ifeq ($(origin IMAGE_REPO), undefined)
  IMAGE_REPO = $(shell long-running.sh)
endif
```

This is better
```
ifeq ($(origin IMAGE_REPO), undefined)
  IMAGE_REPO := $(shell long-running.sh)
endif
```

In the first case (which is equivalent to `?=`), `long-running-sh` runs every time `$(IMAGE_REPO)` is referenced.
In the second case, `long-running.sh` only runs once.

joelanford: I'll switch to this style though. I just saw that in the docs as well
```
ifeq ($(origin IMAGE_REPO), undefined)
  IMAGE_REPO := $(shell long-running.sh)
endif
```

everettraven (approved):

Makefile
```diff
@@ -114,38 +118,43 @@ test: manifests generate fmt vet test-unit test-e2e #HELP Run all tests.
...
 e2e: $(SETUP_ENVTEST) #EXHELP Run the e2e tests.
-	go test -tags $(GO_BUILD_TAGS) -v ./test/e2e/...
+	go test -v ./test/e2e/...
 
-export REG_PKG_NAME=registry-operator
-export PLAIN_PKG_NAME=plain-operator
-export CATALOG_IMG=${E2E_REGISTRY_NAME}.${E2E_REGISTRY_NAMESPACE}.svc:5000/test-catalog:e2e
+E2E_REGISTRY_NAME      := docker-registry
+E2E_REGISTRY_NAMESPACE := operator-controller-e2e
+export REG_PKG_NAME   := registry-operator
```
m1kola, line 125: @joelanford I found a misalignment here ;) 

I would prefer if we could avoid formatting like this as it is hard to maintain this style. This also leads to larger diffs: for example, [here](https://github.com/operator-framework/operator-controller/pull/798/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52) I wanted to only remove one line but I had to change another for it to not look weird.

m1kola: https://github.com/operator-framework/operator-controller/pull/827

PR #819, opened 2024-05-02
🐛 Demonstrate that admission cannot ensure field uniqueness.

benluddy:
# Description

The only field you can practically guarantee uniqueness on is .metadata.name because it is used to construct the etcd key, and etcd does offer consistency. API servers can't because they can handle more than one concurrent request for the same resource and both can be past admission but before storage. There's also latency between creation and an update to the binding's param informer. The problems gets worse in HA setups when you have multiple kube-apiserver instances.

This test nearly always fails on my machine against a kind cluster set up with `make run`:

```
    cluster_extension_admission_test.go:164: duplicate package name: pkg-x (47 duplicates)
--- FAIL: TestClusterExtensionPackageUniquenessConsistency (0.13s)
```

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/test/e2e/cluster_extension_admission_test.go
+++ b/test/e2e/cluster_extension_admission_test.go
@@ -3,15 +3,19 @@ package e2e
 import (
 	"context"
 	"fmt"
+	"net/http"
+	"sync"
 	"testing"
 
 	"github.com/stretchr/testify/require"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	ocv1alpha1 "github.com/operator-framework/operator-controller/api/v1alpha1"
+	"github.com/operator-framework/operator-controller/pkg/scheme"
 )
 
 func TestClusterExtensionPackageUniqueness(t *testing.T) {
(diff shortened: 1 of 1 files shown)


dinhxuanvu:
@joelanford @m1kola FYI

benluddy:
https://github.com/operator-framework/operator-controller/actions/runs/8927081787/job/24519632591?pr=819#step:4:393

```
 === RUN   TestClusterExtensionPackageUniquenessConsistency
    cluster_extension_admission_test.go:163: duplicate package name: pkg-x (18 duplicates)
--- FAIL: TestClusterExtensionPackageUniquenessConsistency (0.13s)
```

joelanford:
Thanks! And also, 😭.

As our resident apiserver experts, do you have any suggestions about how to handle package uniqueness. Should we keep these best effort admission checks and then fallback to a reconciler-based check? Even then, it seems like we'd need to have some synchorization going on to avoid similar problems detecting dups in a reconciler.

benluddy:
If you must guarantee that at most one ClusterExtension resource can have a certain package name, the only practical option is to validate that ClusterExtension's metadata.name is the same as spec.packageName. If you do that, duplicate name conflicts come from the etcd level and will be consistent.

The prior art on this that comes to mind is name validation for CustomResourceDefinitions and the singleton pattern (validating that metadata.name matches some constant string).

joelanford:
I think that's what OCP platform-operators does. I need to see if all package names validate as `metadata.name` though.

m1kola:
Thanks for the demo and explanation. At some point I was wondering about issues of this sort. My bad for not digging deeper.

varshaprasad96:
Please hold this from merging till https://github.com/operator-framework/operator-controller/pull/846 gets in. Thanks!

Issue #735, opened 2024-04-04
[epic] Remove Extension API and controllers from main branch (until after v1 release)


joelanford:
The intent is to remove this API and its controllers for the v1.0.0 release. Once we've released 1.0.0, we plan to re-introduce the Extension API and controllers.

everettraven:
@joelanford This epic is marked as "Needs Docs". Is this still the case or can this be closed out?

joelanford:
Not sure! We just need to comb through docs and make sure we don't have anything in there about the old Extension API.

everettraven:
I don't see anything about the old Extension API in the docs. Closing this issue and moving the epic to "Done". Thanks!

Issue #747, opened 2024-04-05
[epic] Port relevant Extension reconciler logic over to ClusterExtension


joelanford:
We have been working a lot on fine tuning the logic of the Extension controller recently. The major themes have been:
- Make the updates of various status fields more orthogonal
   - resolution results
   - prorgressing state
   - status of the currently installed bundle
   - deprecation status of the currently installed bundle and the specified package and channel
- Removal of deppy due to the lack of a need for a SAT solver.
- Make room in the ClusterExtension API for other techniques for sourcing bundles (i.e. enable a future where bundles don't have to be sourced from a catalog)

Once we wrap that work up, let's port the relevant bits over to the ClusterExtension controller.

m1kola:
Deppy removal PR: https://github.com/operator-framework/operator-controller/pull/758

gallettilance:
Extension API fields in ClusterExtension API #842

ankitathomas:
https://github.com/operator-framework/operator-controller/pull/846 includes the status field updates

ankitathomas:
Closing this out, follow-up work for sourcing bundles from non-catalog sources tracked in https://github.com/operator-framework/operator-controller/issues/906:  [epic #903](https://github.com/operator-framework/operator-controller/issues/903)

PR #820, opened 2024-05-02
⚠ REMOVE extension api from main branch

acmenezes:
# Description

The intent is to remove this API and its controllers for the v1.0.0 release. Once we've released 1.0.0, we plan to re-introduce the Extension API and controllers.

@joelanford, please take a look and tell me if that's what you had in mind. I did a flat removal of all the references to the api as well. Didn't try to circumvent or redesign around any errors. Any suggestions or changes are welcomed. Saving this as a draft PR to be safe.

My question, and I think it's an important one, would we pull the main branch's code to some secondary branch as a reference for after v1.0.0 before merging this if we plan to reintroduce it?

Solves #735 

## Reviewer Checklist

- [ ] API Go Documentation


joelanford:
I don't think we need to save the code off. The git history is there, so we can just come back to the commit prior to this PR if necessary.

But I suspect we don't even do that. I think it is more likely that we'll essentially copy `ClusterExtension` back over to `Extension` and make the few tweaks necessary for it to work as a namespace-scoped API. And we'll do a similar thing for the controller.

api/v1alpha1/clusterextension_types.go
```diff
@@ -121,11 +121,6 @@ func init() {
 
 // ClusterExtensionStatus defines the observed state of ClusterExtension
 type ClusterExtensionStatus struct {
-	// +optional
-	InstalledBundle *BundleMetadata `json:"installedBundle,omitempty"`
-	// +optional
-	ResolvedBundle *BundleMetadata `json:"resolvedBundle,omitempty"`
-
```
joelanford, line 128: These should stay. If `BundleMetadata` is defined in `extension_types.go`, let's move those definitions to this file.

config/rbac/role.yaml
```diff
@@ -29,17 +29,6 @@ rules:
   - patch
   - update
   - watch
-- apiGroups:
-  - kappctrl.k14s.io
```
joelanford, line 33: We should remove the kapp dependency (I think you'll see breadcrumbs in `Makefile` and the install script template).

internal/controllers/clusterextension_controller.go
```diff
@@ -121,17 +121,14 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
 	// Lookup the bundle that corresponds to the ClusterExtension's desired package.
 	bundle, err := r.resolve(ctx, ext)
 	if err != nil {
-		ext.Status.InstalledBundle = nil
```
joelanford, line 124: Re: keeping this fields in ClusterExtension, we also want to keep the ClusterExtension reconciler logic intact.

internal/controllers/common_controller.go
```diff
@@ -110,9 +110,7 @@ func setDeprecationStatusesUnknown(conditions *[]metav1.Condition, message strin
 // setProgressingStatusConditionSuccess sets the progressing status condition to false for a successful install or upgrade.
 func setProgressingStatusConditionSuccess(conditions *[]metav1.Condition, message string, generation int64) {
 	apimeta.SetStatusCondition(conditions, metav1.Condition{
-		Type:               ocv1alpha1.TypeProgressing,
```
joelanford, line 113: If ClusterExtension controller is calling these functions, let's make sure they remain unchanged. That again might mean moving types from `extension_types.go` into `clusterextension_types.go`.

api/v1alpha1/clusterextension_types.go
acmenezes: Done.

config/rbac/role.yaml
acmenezes: Cool. Thanks I will check on that one yet.

internal/controllers/clusterextension_controller.go
acmenezes: Ok. Ack. Keeping it intact.

internal/controllers/common_controller.go
acmenezes: I saw. I needed to move some of those over the other. I'm getting some test errors. Now I need to work through them. Thanks for your assistance @joelanford ! I appreciate it.

everettraven (commented):
IMO the lint failures can be addressed via a `//nolint:unused` comment for now and be updated in #747 when they are used again.

This should help keep the scope of this PR smaller

config/rbac/role.yaml
acmenezes: Hi @joelanford , I finally removed kapp from Makefile, install scripts and also from the scheme package I put on the last PR. I guess that should cover it. Let me know.

api/v1alpha1/clusterextension_types.go
```diff
@@ -106,6 +115,7 @@ func init() {
 		TypePackageDeprecated,
 		TypeChannelDeprecated,
 		TypeBundleDeprecated,
+		TypeProgressing,
```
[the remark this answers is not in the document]
acmenezes: Hi @everettraven , actually it will fail another test and ask me to add it to the list. What I did was adding that to the list but then it requires me to actually set progressing status in the ClusterExtension controller. That part I'm honestly not sure if I did right. I appreciate some eyes on this. On the other hand if you really want to keep the type out of the list of conditions sets then we may have to change some tests. Happy to go either way. I count on your directions here.

acmenezes:
> IMO the lint failures can be addressed via a `//nolint:unused` comment for now and be updated in #747 when they are used again.
> 
> This should help keep the scope of this PR smaller

@everettraven good idea. I just added the comments there. I think no more lint issues are found. Thanks.

api/v1alpha1/clusterextension_types.go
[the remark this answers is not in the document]
everettraven: Oh that is interesting. I didn't realize there was a test that we had the ensured those constants are added to the conditionsets. 

I'm not sure this is really a valuable test for us at the moment. I'm not sure what the best path forward here is personally, maybe:
- Merge #842 
- Rebase this PR on ^, so it essentially becomes shifting the logic around but the clusterextension_controller already uses everything so we don't have to fight with CI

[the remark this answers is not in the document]
acmenezes: Makes sense. I think we can talk about it on our meeting on Tuesday.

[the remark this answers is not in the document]
joelanford: > actually it will fail another test and ask me to add it to the list.

What was the specific test failure? I think @everettraven is correct that we should be able to let this status condition type be completely removed since it is currently only present in the Extension API.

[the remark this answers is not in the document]
everettraven: @joelanford I think it is these test cases: https://github.com/operator-framework/operator-controller/blob/e9acbd319c59e3805e5795a5f60a3fff9f24ccf9/api/v1alpha1/clusterextension_types_test.go#L18-L63

internal/controllers/common_controller.go
everettraven: I totally misinterpreted this interaction. We can remove all these functions that have to do with setting the `Progressing` condition and _also_ remove the condition type. 

The condition type and functionality can be added back as part of #842

api/v1alpha1/clusterextension_types.go
```diff
@@ -25,6 +25,15 @@ import (
...
 
 const (
+
+	// TypeProgressing indicates whether operator-controller is
+	// reconciling, installing, updating or deleting an extension.
+	TypeProgressing = "Progressing"
+
+	ReasonProgressing                = "Progressing"
+	ReasonFailedToReachDesiredIntent = "FailedToReachDesiredIntent"
+	ReasonReachedDesiredIntent       = "ReachedDesiredIntent"
```
varshaprasad96, line 35: _IIRC_ these were related to upgrades and being used in extension-controller. These can be removed too.

varshaprasad96 (changes requested):
@acmenezes There is also a Kapp CRD in testdata: https://github.com/operator-framework/operator-controller/blob/b82812ea96a44eccc4ac4bd9bbfe8b023271130d/testdata/crds/kappctrl.k14s.io_app.yaml, could you also remove this please? Thanks!

varshaprasad96 (approved):
/lgtm
/approve

PR #821, opened 2024-05-02
:sparkles: Use rukpak as a library

tmshort:
# Description

## Reviewer Checklist

- [ ] API Go Documentation


cmd/manager/main.go
```diff
@@ -127,7 +128,8 @@ func main() {
 		systemNamespace = util.PodNamespace()
 	}
 
-	unpacker, err := source.NewDefaultUnpacker(mgr, systemNamespace, unpackImage)
+	var rootCAs *x509.CertPool
+	unpacker, err := source.NewDefaultUnpacker(mgr, systemNamespace, unpackImage, rootCAs)
```
varshaprasad96, line 132: Can we explicitly pass a `nil` instead? Passing an empty variable is very deceiving imo. We aren't using the http source, so setting up the `rootCAs` to nil shouldn't matter at least for now.

varshaprasad96 (commented):
Just one comment, rest looks good!

cmd/manager/main.go
joelanford: +1 (I would also make a comment to that effect)

tmshort: It's basically the same, but it's indicating what the argument is (i.e. an `x509.CertPool`) rather than some random `nil`... Not sure I care either way, but this at least indicates what we're _not_ doing.

joelanford: Maybe a compromise:
```suggestion
	unpacker, err := source.NewDefaultUnpacker(mgr, systemNamespace, unpackImage, (*x509.CertPool)(nil))
```

tmshort: I'm putting in Joe's suggestion, that way it's still clear that we're not providing a CertPool.

PR #831, opened 2024-05-03
✨ OPRUN-3293 Fixes several unit tests - not cache ones 

bentito:
# Description

Fixes several ClusterExtension controller related unit tests. Not ones related to cache not being started though.

## Reviewer Checklist

- [ ] API Go Documentation
- [X] Comprehensive Commit Messages

### DIFF
--- a/internal/controllers/clusterextension_controller.go
+++ b/internal/controllers/clusterextension_controller.go
@@ -177,6 +177,34 @@ func checkForUnexpectedFieldChange(a, b ocv1alpha1.ClusterExtension) bool {
 	return !equality.Semantic.DeepEqual(a, b)
 }
 
+func (r *ClusterExtensionReconciler) handleResolutionErrors(ext *ocv1alpha1.ClusterExtension, err error) (ctrl.Result, error) {
+	var aggErrs utilerrors.Aggregate
+	if errors.As(err, &aggErrs) {
+		for _, err := range aggErrs.Errors() {
+			errorMessage := err.Error()
+			if strings.Contains(errorMessage, "no package") {
+				// Handle no package found errors, potentially setting status conditions
+				setResolvedStatusConditionFailed(&ext.Status.Conditions, errorMessage, ext.Generation)
+				ensureAllConditionsWithReason(ext, "ResolutionFailed", errorMessage)
+			} else if strings.Contains(errorMessage, "invalid version range") {
+				// Handle invalid version range errors, potentially setting status conditions
+				setResolvedStatusConditionFailed(&ext.Status.Conditions, errorMessage, ext.Generation)
+				ensureAllConditionsWithReason(ext, "ResolutionFailed", errorMessage)
+			} else {
+				// General error handling
+				setResolvedStatusConditionFailed(&ext.Status.Conditions, errorMessage, ext.Generation)
+				ensureAllConditionsWithReason(ext, "InstallationStatusUnknown", "")
+			}
+		}
+	} else {
+		// If the error is not an aggregate, handle it as a general error
+		errorMessage := err.Error()
+		setResolvedStatusConditionFailed(&ext.Status.Conditions, errorMessage, ext.Generation)
+		ensureAllConditionsWithReason(ext, "InstallationStatusUnknown", "")
+	}
+	return ctrl.Result{}, err
+}
+
 // Helper function to do the actual reconcile
 //
 // Today we always return ctrl.Result{} and an error.
@@ -190,33 +218,7 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
 	bundle, err := r.resolve(ctx, *ext)
 	if err != nil {
 		l.V(1).Info("bundle resolve error", "error", err)
-
-		if aggErrs, ok := err.(utilerrors.Aggregate); ok {
-			for _, err := range aggErrs.Errors() {
-				if errMsg := err.Error(); strings.Contains(errMsg, "no package") {
-					// Handle no package found errors, potentially setting status conditions
-					errorMessage := errMsg
-					setResolvedStatusConditionFailed(&ext.Status.Conditions, errorMessage, ext.Generation)
-					ensureAllConditionsWithReason(ext, "ResolutionFailed", errorMessage)
-				} else if strings.Contains(errMsg, "invalid version range") {
-					// Handle invalid version range errors, potentially setting status conditions
-					errorMessage := errMsg
-					setResolvedStatusConditionFailed(&ext.Status.Conditions, errorMessage, ext.Generation)
-					ensureAllConditionsWithReason(ext, "ResolutionFailed", errorMessage)
-				} else {
-					// General error handling
-					errorMessage := err.Error()
-					setResolvedStatusConditionFailed(&ext.Status.Conditions, errorMessage, ext.Generation)
-					ensureAllConditionsWithReason(ext, "InstallationStatusUnknown", "")
-				}
-			}
-		} else {
-			// If the error is not an aggregate, handle it as a general error
-			errorMessage := err.Error()
-			setResolvedStatusConditionFailed(&ext.Status.Conditions, errorMessage, ext.Generation)
-			ensureAllConditionsWithReason(ext, "InstallationStatusUnknown", "")
-		}
-		return ctrl.Result{}, err
+		return r.handleResolutionErrors(ext, err)
 	}
 
 	bundleVersion, err := bundle.Version()
--- a/internal/controllers/clusterextension_controller.go
+++ b/internal/controllers/clusterextension_controller.go
@@ -25,6 +25,7 @@ import (
 	"sort"
 	"strings"
 	"sync"
+	"time"
 
 	mmsemver "github.com/Masterminds/semver/v3"
 	bsemver "github.com/blang/semver/v4"
@@ -59,6 +60,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
+	"github.com/operator-framework/api/pkg/operators/v1alpha1"
 	catalogd "github.com/operator-framework/catalogd/api/core/v1alpha1"
 	helmclient "github.com/operator-framework/helm-operator-plugins/pkg/client"
 	"github.com/operator-framework/operator-registry/alpha/declcfg"
@@ -72,6 +74,7 @@ import (
 	"github.com/operator-framework/operator-controller/internal/catalogmetadata"
 	catalogfilter "github.com/operator-framework/operator-controller/internal/catalogmetadata/filter"
 	catalogsort "github.com/operator-framework/operator-controller/internal/catalogmetadata/sort"
+	"github.com/operator-framework/operator-controller/internal/conditionsets"
 	"github.com/operator-framework/operator-controller/internal/handler"
 	"github.com/operator-framework/operator-controller/internal/labels"
 	"github.com/operator-framework/operator-controller/internal/packageerrors"
@@ -111,14 +114,13 @@ func (r *ClusterExtensionReconciler) Reconcile(ctx context.Context, req ctrl.Req
 
 	var existingExt = &ocv1alpha1.ClusterExtension{}
 	if err := r.Client.Get(ctx, req.NamespacedName, existingExt); err != nil {
-		return ctrl.Result{}, client.IgnoreNotFound(err)
+		return ctrl.Result{}, utilerrors.NewAggregate([]error{client.IgnoreNotFound(err), nil})
 	}
 
 	reconciledExt := existingExt.DeepCopy()
 	res, reconcileErr := r.reconcile(ctx, reconciledExt)
-	if reconcileErr != nil {
-		return ctrl.Result{}, reconcileErr
-	}
+
+	var updateErrors []error
 
 	// Do checks before any Update()s, as Update() may modify the resource structure!
 	updateStatus := !equality.Semantic.DeepEqual(existingExt.Status, reconciledExt.Status)
@@ -127,7 +129,7 @@ func (r *ClusterExtensionReconciler) Reconcile(ctx context.Context, req ctrl.Req
 
 	if updateStatus {
 		if updateErr := r.Client.Status().Update(ctx, reconciledExt); updateErr != nil {
-			return res, utilerrors.NewAggregate([]error{reconcileErr, updateErr})
+			updateErrors = append(updateErrors, updateErr)
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


internal/controllers/clusterextension_controller.go
```diff
@@ -157,15 +178,31 @@ func checkForUnexpectedFieldChange(a, b ocv1alpha1.ClusterExtension) bool {
...
 	bundle, err := r.resolve(ctx, *ext)
-	if err != nil {
-		// set right statuses
-		ext.Status.ResolvedBundle = nil
-		ext.Status.InstalledBundle = nil
+	if err != nil || bundle == nil {
+		l.V(1).Info("bundle resolve error: %v\n", err)
+
+		errorMessage := err.Error()
+		packageNotFoundPattern := regexp.MustCompile(`^no package "([^"]+)" found$`)
```
bentito, line 188: likely we should get more of a type wrapping the error than do this

internal/controllers/clusterextension_controller_test.go
```diff
@@ -62,7 +64,13 @@ func TestClusterExtensionNonExistentPackage(t *testing.T) {
 	t.Log("By running reconcile")
 	res, err := reconciler.Reconcile(ctx, ctrl.Request{NamespacedName: extKey})
 	require.Equal(t, ctrl.Result{}, res)
-	require.EqualError(t, err, packageerrors.GenerateError(pkgName).Error())
+	require.EqualError(t,
```
bentito, line 67: needed change b/c the errors are aggregated now.

internal/controllers/clusterextension_controller_test.go
```diff
@@ -122,12 +135,10 @@ func TestClusterExtensionNonExistentVersion(t *testing.T) {
 	cond = apimeta.FindStatusCondition(clusterExtension.Status.Conditions, ocv1alpha1.TypeInstalled)
 	require.NotNil(t, cond)
 	require.Equal(t, metav1.ConditionUnknown, cond.Status)
-	require.Equal(t, ocv1alpha1.ReasonInstallationStatusUnknown, cond.Reason)
-	require.Equal(t, "installation has not been attempted due to failure to gather data for resolution", cond.Message)
+	//require.Equal(t, "installation has not been attempted due to failure to gather data for resolution", cond.Message) // ??
```
bentito, line 138: this seemed very specific and not set by a corresponding set in the codebase elsewhere, so I think we can live without it in the test.

internal/controllers/clusterextension_controller.go
```diff
@@ -139,7 +140,27 @@ func (r *ClusterExtensionReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		}
 	}
 
-	return res, reconcileErr
+	return res, utilerrors.NewAggregate([]error{reconcileErr, nil})
```
varshaprasad96, line 143: Same - why this?

internal/controllers/clusterextension_controller.go
```diff
@@ -109,14 +113,11 @@ func (r *ClusterExtensionReconciler) Reconcile(ctx context.Context, req ctrl.Req
 
 	var existingExt = &ocv1alpha1.ClusterExtension{}
 	if err := r.Client.Get(ctx, req.NamespacedName, existingExt); err != nil {
-		return ctrl.Result{}, client.IgnoreNotFound(err)
+		return ctrl.Result{}, utilerrors.NewAggregate([]error{client.IgnoreNotFound(err), nil})
 	}
 
 	reconciledExt := existingExt.DeepCopy()
 	res, reconcileErr := r.reconcile(ctx, reconciledExt)
-	if reconcileErr != nil {
```
varshaprasad96, line 117: It's better to exit early that do it after status changes. The below code assumes that reconcile was successful, so I'd revert this change.

internal/controllers/clusterextension_controller.go
varshaprasad96: Can we improve the errors on the resolution instead of this? That would probably be easier, given the resolution code is in the same file.

internal/controllers/clusterextension_controller_test.go
varshaprasad96: If I'm guessing right, I think this was to capture error while building dippy variables for resolution. We can remove this.

internal/controllers/clusterextension_controller.go
```diff
@@ -116,152 +178,195 @@ func checkForUnexpectedFieldChange(a, b ocv1alpha1.ClusterExtension) bool {
...
+		packageNotFoundPattern := regexp.MustCompile(`^no package "([^"]+)" found$`)
+		versionNotFoundPattern := regexp.MustCompile(`^no package "([^"]+)" matching version "([^"]+)" found$`)
+
+		switch {
+		case packageNotFoundPattern.MatchString(errorMessage):
+			setResolvedStatusConditionFailed(&ext.Status.Conditions, errorMessage, ext.Generation)
+			ensureAllConditionsWithReason(ext, "ResolutionFailed", errorMessage)
+
+		case versionNotFoundPattern.MatchString(errorMessage):
+			setInstalledStatusConditionUnknown(&ext.Status.Conditions, errorMessage, ext.Generation)
```
varshaprasad96, line 197: Also, `versionNotFound` alone - has the installedStatus set to unknown, whereas others don't. Can we keep it consistent?

internal/controllers/clusterextension_controller.go
bentito: yeah we should, mostly wanted to see if custom casing was the way to go

bentito: seems like if the errs are changes to be different we can case on them much more reliably

internal/controllers/clusterextension_controller_test.go
bentito: changed in c6d6f078

internal/controllers/clusterextension_controller.go
```diff
@@ -109,14 +113,11 @@ func (r *ClusterExtensionReconciler) Reconcile(ctx context.Context, req ctrl.Req
 
 	var existingExt = &ocv1alpha1.ClusterExtension{}
 	if err := r.Client.Get(ctx, req.NamespacedName, existingExt); err != nil {
-		return ctrl.Result{}, client.IgnoreNotFound(err)
+		return ctrl.Result{}, utilerrors.NewAggregate([]error{client.IgnoreNotFound(err), nil})
```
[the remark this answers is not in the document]
bentito: due to previously introduced changed to aggregate errors.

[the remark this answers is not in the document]
bentito: before these changes, `Reconcile()` had one prior usage of the the non-aggregate error return, so it was mixed. Not being able to parse the aggregate form in the tests was a one cause of failure in some tests.

internal/controllers/clusterextension_controller.go
bentito: same reason as above.

internal/controllers/clusterextension_controller.go
bentito: we can exit early but I'll need to look at making sure the right stuff is still set in the failure scenario -- was a cause of test failure.

internal/controllers/clusterextension_controller.go
bentito: Well we inconsistently test for it then:

https://github.com/operator-framework/operator-controller/blob/c6d6f07881bf8aa6a8de0dfaf5a48a5968c1a7ca/internal/controllers/clusterextension_controller_test.go#L137

Want it to be Unknown for Version missing (TestClusterExtensionNonExistentVersion) but don't when it's the Package missing (TestClusterExtensionNonExistentPackage). Should it always be Unknown?

internal/controllers/clusterextension_controller.go
bentito: fixed in a53c47fe

bentito:
Remaining ClusterExtension controller unit test fails should now be limited to:

- something messed up in code with "cache not started";
- rukpak refs needing to be pulled from test code

tmshort:
Can you rebase this onto the latest helm-poc branch?

internal/controllers/clusterextension_controller.go
```diff
@@ -609,16 +663,17 @@ func (r *ClusterExtensionReconciler) resolve(ctx context.Context, clusterExtensi
 			// to allow only minor and patch version as successors.
 			wantedVersionRangeConstraint, err := mmsemver.NewConstraint(fmt.Sprintf("^%s", installedVersion))
 			if err != nil {
-				return nil, err
+				return nil, OtherResolveError, err
 			}
 			predicates = append(predicates, catalogfilter.InMastermindsSemverRange(wantedVersionRangeConstraint))
+
```
tmshort, line 669: Lint doesn't like this new blank line

internal/controllers/clusterextension_controller.go
```diff
@@ -568,10 +610,22 @@ func clusterExtensionRequestsForCatalog(c client.Reader, logger logr.Logger) crh
...
+	PackageNotFoundError ResolveErrorType = iota
+	VersionNotFoundError
+	OtherResolveError
+)
+
+func (r ResolveErrorType) String() string {
+	return [...]string{"PackageNotFound", "VersionNotFound", "Other"}[r]
+}
+
+func (r *ClusterExtensionReconciler) resolve(ctx context.Context, clusterExtension ocv1alpha1.ClusterExtension) (*catalogmetadata.Bundle, ResolveErrorType, error) {
```
tmshort, line 625: Why are we adding another return value that always ends up being the same value? in other words, it does not appear that `PackageNotFoundError` nor "VersionNotFoundError` are used.
Shouldn't these be rolled up into the `error` return?
Is this to avoid aggregate errors? If so, there's got to be a better way.

tmshort: i.e. the linter is complaining:
```
(*ClusterExtensionReconciler).resolve - result 1 (github.com/operator-framework/operator-controller/internal/controllers.ResolveErrorType) is always 2 (unparam)
```

tmshort:
Yeah, you may need to cherry-pick your commit to the updated helm-poc branch... not sure rebase will be smart enough due to the squash.

internal/controllers/clusterextension_controller_test.go
```diff
@@ -821,7 +821,7 @@ func TestClusterExtensionVersionNoChannel(t *testing.T) {
 	cond = apimeta.FindStatusCondition(clusterExtension.Status.Conditions, ocv1alpha1.TypeInstalled)
 
 	require.NotNil(t, cond)
-	require.Equal(t, metav1.ConditionUnknown, cond.Status)
+	require.Equal(t, metav1.ConditionFalse, cond.Status)
```
varshaprasad96, line 824: This looks good to me! We are anyway going to change the installation conditions, so marking install as False when there is error is fine for now.

internal/controllers/clusterextension_controller.go
tmshort: So, the linter is still complaining about this. And I'm not sure this is the right solution, here. It also seems that you are rewriting `resolve()` because you moved it further down in the file - diff is having difficulties. Can you put it back where it was originally?

internal/controllers/clusterextension_controller.go
```diff
@@ -330,68 +370,6 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
...
-
-	sort.SliceStable(resultSet, func(i, j int) bool {
-		return catalogsort.ByVersion(resultSet[i], resultSet[j])
-	})
-	sort.SliceStable(resultSet, func(i, j int) bool {
-		return catalogsort.ByDeprecated(resultSet[i], resultSet[j])
-	})
-	return resultSet[0], nil
-}
-
```
tmshort, line 394: i.e. git thinks you're deleting the function, when you're really changing it below. Can you move the `resolve()` function back where it was defined? It will minimize the diff, and make it wasier to us to see what changed.

internal/controllers/clusterextension_controller.go
```diff
@@ -541,23 +519,6 @@ func (r *ClusterExtensionReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
-				}
-
-				if oldObject.Status.ResolvedSource != nil && newObject.Status.ResolvedSource != nil {
-					if oldObject.Status.ResolvedSource.Image != nil && newObject.Status.ResolvedSource.Image != nil {
-						return oldObject.Status.ResolvedSource.Image.ResolvedRef != newObject.Status.ResolvedSource.Image.ResolvedRef
-					}
-				}
-				return true
-			},
-		}).
```
tmshort, line 560: Why is this removed?

internal/controllers/clusterextension_controller.go
bentito: done

internal/controllers/clusterextension_controller.go
bentito: mistake, it's back

internal/controllers/clusterextension_controller.go
```diff
@@ -330,16 +372,28 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
...
+	PackageNotFoundError ResolveErrorType = iota
+	VersionNotFoundError
+	OtherResolveError
+)
+
+func (r ResolveErrorType) String() string {
+	return [...]string{"PackageNotFound", "VersionNotFound", "Other"}[r]
+}
+
+func (r *ClusterExtensionReconciler) resolve(ctx context.Context, clusterExtension ocv1alpha1.ClusterExtension) (*catalogmetadata.Bundle, ResolveErrorType, error) {
```
tmshort, line 387: To minimize changes, can you rename `clusterExtension` to `ext`?

tmshort: And I still don't understand the need for `ResolveErrorType`, lint is complaining about it.

bentito: Refactor done.

lint is complaining b/c nothing uses the package not found err type. I'd like to work it out so this is used and we can report properly that package is non existent after resolve. And then I think the extra errorType will make sense.

tmshort: Perhaps use aggregate errors instead?, so no new return argument is needed?
e.g. 
```
var PackageNotFoundError = fmt.Errorf("package not found")
var VersionNotFoundError = fmt.Errorf("version not found")
var OtherResolveError = fmt.Errorf("other resolve error")

...

    // in resolve()
    return nil, utilerrors.NewAggregate([]error{OtherResolveError, err})

...
    // in reconcile():
    switch {
        case agg.Is(OtherResolveError): 
            ....
        case agg.Is(VersionNotFoundError):
            ...
        case agg.Is(OtherResolveError):
            ...
        default:
            ...
    }

bentito: think latest commit refactors out the added type and moves it all to use the aggregate error.

tmshort (approved):
/lgtm
/approve

internal/controllers/clusterextension_controller.go
bentito: added return value removed.

PR #842, opened 2024-05-09
✨ remove extension-specific type checks

ankitathomas:


api/v1alpha1/clusterextension_types.go
```diff
@@ -18,53 +18,35 @@ package v1alpha1
...
+	Source ClusterExtensionSource `json:"source"`
 
-	//+kubebuilder:validation:Enum:=Enforce;Ignore
-	//+kubebuilder:default:=Enforce
 	//+kubebuilder:Optional
 	//
-	// Defines the policy for how to handle upgrade constraints
-	UpgradeConstraintPolicy UpgradeConstraintPolicy `json:"upgradeConstraintPolicy,omitempty"`
+	// skipCRDUpgradeSafetyCheck specifies whether the CRD upgrade safety checks should be skipped when attempting to install the ClusterExtension
+	SkipCRDUpgradeSafetyCheck bool `json:"skipCRDUpgradeSafetyCheck,omitempty"`
```
everettraven, line 49: I think @rashmigottipati is working on some changes that are going to be more extensible for future preflight checks. I think it would be okay to keep this field out of these changes

api/v1alpha1/common_types.go
everettraven, line 1: Is there a particular reason for these fields to have been moved from the `clusterextension_types.go` file to this one?

api/v1alpha1/clusterextension_types.go
everettraven: Oh, I see - this PR still has the extension type. I forgot https://github.com/operator-framework/operator-controller/pull/820 has not merged yet :)

api/v1alpha1/common_types.go
ankitathomas: Mostly just readability.

everettraven: Oh, I see - this PR still has the extension type. I forgot https://github.com/operator-framework/operator-controller/pull/820 has not merged yet :)

api/v1alpha1/clusterextension_types.go
everettraven: Ignore the previous comment. I accidentally commented this on the wrong review thread.

config/admission/admission.yaml
everettraven, line 1: Just a heads up, I think this is being removed in #850 due to the findings in #819

internal/controllers/clusterextension_controller.go
```diff
@@ -119,14 +119,25 @@ func checkForUnexpectedFieldChange(a, b ocv1alpha1.ClusterExtension) bool {
...
 //nolint:unparam
 func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alpha1.ClusterExtension) (ctrl.Result, error) {
+	l := log.FromContext(ctx).WithName("operator-controller")
+
+	// Don't do anything if Paused
+	ext.Status.Paused = ext.Spec.Paused
+	if ext.Spec.Paused {
+		l.Info("resource is paused", "name", ext.GetName(), "namespace", ext.GetNamespace())
+		return ctrl.Result{}, nil
+	}
```
everettraven, line 129: I know that in theory we shouldn't do anything, but from a UX perspective I think it might be worth setting some form of status indicating that "no work was done" because it is paused.

What do you think of something like setting `Progressing` to `False` with a reason `Paused`?

everettraven: I'm also OK with this being done in a follow-up since the scope of this PR is to make the two APIs consistent

everettraven (approved):
A couple comments relating to existing or future PRs that are modifying some of the same things as a heads up and a suggestion for UX on paused state so users don't have to check the logs to verify that a resource is in the paused state.

Since the scope of this PR is to bring the `Extension` and `ClusterExtension` APIs to parity I think that it is OK to merge this without addressing any of my comments and implementing my pause suggestion as a follow-up.

Great work @ankitathomas !

api/v1alpha1/clusterextension_types.go
```diff
@@ -18,53 +18,35 @@ package v1alpha1
...
+// ClusterExtensionSource defines the source for this ClusterExtension, right now, only a package is supported.
+type ClusterExtensionSource struct {
+	//+kubebuilder:validation:Enum:=package
+	//+kubebuilder:validation:Required
+	// SourceType is the discriminator for the source type
+	SourceType string `json:"sourceType"`
+
+	// Package defines a reference for a bundle in a catalog defined by a name and a version and/or channel
+	Package *SourcePackage `json:"package,omitempty"`
+}
```
varshaprasad96, line 34: Heads up - this change is going have us modify the helm-poc branch.
cc: @bentito @dtfranz @tmshort

internal/controllers/clusterextension_controller.go
```diff
@@ -119,14 +119,25 @@ func checkForUnexpectedFieldChange(a, b ocv1alpha1.ClusterExtension) bool {
...
+
 	// Lookup the bundle that corresponds to the ClusterExtension's desired package.
 	bundle, err := r.resolve(ctx, ext)
 	if err != nil {
-		ext.Status.InstalledBundle = nil
-		setInstalledStatusConditionUnknown(&ext.Status.Conditions, "installation has not been attempted as resolution failed", ext.GetGeneration())
+		if c := apimeta.FindStatusCondition(ext.Status.Conditions, ocv1alpha1.TypeInstalled); c == nil {
```
varshaprasad96, line 134: Do we need this additional check for finding status condition? _IIRC_ Installed status is added by default. I see that we are not doing this for bundle validation or unpack, so to maintain consistency, it's better to rather remove this check here, and explicitly set installed to nil regardless.

internal/controllers/clusterextension_controller.go
```diff
@@ -155,6 +167,7 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
 		ext.Status.InstalledBundle = nil
 		setInstalledStatusConditionFailed(&ext.Status.Conditions, err.Error(), ext.GetGeneration())
 		setDeprecationStatusesUnknown(&ext.Status.Conditions, "deprecation checks have not been attempted as installation has failed", ext.GetGeneration())
+		setProgressingStatusConditionProgressing(&ext.Status.Conditions, "installation failed", ext.GetGeneration())
```
varshaprasad96, line 170: If installed is explicitly set to failed, and we are not able to find if the bundle deployment exists on cluster or not, I think this should be set to `False` instead, since there is no other state of improvement after this. Was there a specific reason to make it true instead?

internal/controllers/clusterextension_controller.go
```diff
@@ -163,16 +176,19 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
 	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(dep.UnstructuredContent(), existingTypedBundleDeployment); err != nil {
 		// originally Reason: ocv1alpha1.ReasonInstallationStatusUnknown
 		ext.Status.InstalledBundle = nil
-		setInstalledStatusConditionUnknown(&ext.Status.Conditions, err.Error(), ext.GetGeneration())
+		setInstalledStatusConditionFailed(&ext.Status.Conditions, err.Error(), ext.GetGeneration())
 		setDeprecationStatusesUnknown(&ext.Status.Conditions, "deprecation checks have not been attempted as installation has failed", ext.GetGeneration())
+		setProgressingStatusConditionProgressing(&ext.Status.Conditions, "installation failed", ext.GetGeneration())
```
varshaprasad96, line 181: The same comment as above, why set Progressing as true and not as false when we are returning an error here and know that we cannot proceed with install?

varshaprasad96: Thinking about it, my previous comment may with respect to rukpak API may not be relevant when we get Helm stuff in, but this particular case would still hold true.

internal/controllers/clusterextension_controller.go
ankitathomas: We only need the status check if we want to preserve the installed condition from the currently installed thing in case of a resolution failure to indicate that the installed version is healthy even if there's an error on the Extension/ClusterExtension. We probably don't need this check once we have a separate healthy condition the users could look at instead.

internal/controllers/clusterextension_controller.go
ankitathomas: I thought of the progressing condition as more of an indication of  'an install/upgrade is in progress, the currently installed bundle may be unstable'. Since it's stuck progressing here, it should be an indicator that the current installation is possibly incomplete/broken.

varshaprasad96:
Please hold this from merging till https://github.com/operator-framework/operator-controller/pull/846 gets in. Thanks!

grokspawn (approved):

Issue #745, opened 2024-04-05
Add configuration to the `ClusterExtension` API for disabling CRD Upgrade Safety validations


everettraven:
A new, optional, field should be introduced to the `ClusterExtension` API spec for enabling/disabling the CRD Upgrade Safety preflight checks. The default state should be enabled.

rashmigottipati:
PR created in rukpak: https://github.com/operator-framework/rukpak/pull/902

rashmigottipati:
https://github.com/operator-framework/rukpak/pull/902 merged. This added the CRD upgrade safety preflight configuration API into rukpak.

Work to transplant this logic into operator controller is currently blocked by https://github.com/operator-framework/operator-controller/pull/846 and will be tracked by https://github.com/operator-framework/operator-controller/issues/877

PR #845, opened 2024-05-10
✨ Add optional field to ClusterExtension API to skip CRD upgrade safety checks

rashmigottipati:
# Description

Adds an optional ClusterExtension.spec.skipCRDUpgradeSafetyChecks field that when set to true will be used to disable the CRD upgrade safety checks when attempting to install a cluster extension.

No logic included in this PR as the CRD Upgrade Safety preflight check as there isn't anything yet to enable/disable. So logic to enable/disable will be added in a future PR once everything is in place. 

fixes https://github.com/operator-framework/operator-controller/issues/745

## Reviewer Checklist

- [ ] API Go Documentation


api/v1alpha1/clusterextension_types.go
```diff
@@ -73,6 +73,10 @@ type ClusterExtensionSpec struct {
 	// the bundle may contain resources that are cluster-scoped or that are
 	// installed in a different namespace. This namespace is expected to exist.
 	InstallNamespace string `json:"installNamespace"`
+
+	//+kubebuilder:Optional
+	// skipCRDUpgradeSafetyCheck specifies whether or not the CRD upgrade safety checks should be skipped when attempting to install the cluster extension.
+	SkipCRDUpgradeSafetyCheck bool `json:"skipCRDUpgradeSafetyCheck,omitempty"`
```
everettraven, line 79: What do you think of making a more generic and extensible field here for when we go to add additional preflight checks in the future rather than a field targeted at a single preflight check?

everettraven: I'm thinking maybe something like:
```go
type ClusterExtensionSpec {
...
    Preflight PreflightConfig `json:"preflight,omitempty"`
...
}

type PreflightConfig struct {
    CRDUpgradeSafety CRDUpgradeSafetyPreflightConfig
}

    Skip bool
}
```

With ^ we should be able to add additional fields / functionality to both the CRD Upgrade Safety preflight configuration (like enabling/disabling specific validations) as well as adding entirely new preflight checks to the more general `PreflightConfig` without any breaking changes or adding a bunch of root fields.

tmshort: Alternatively, avoiding a `bool` type, and making it an enum (@joelanford has made this suggestion to me before).

rashmigottipati: Addressed in [latest commit ](https://github.com/operator-framework/operator-controller/pull/845/commits/30bf1c649a88825d181468590011effe5e7ef5a7)

api/v1alpha1/clusterextension_types.go
```diff
@@ -73,8 +73,32 @@ type ClusterExtensionSpec struct {
...
 
+// PreflightConfig holds the config for the preflight checks. Currently, this is implemented for the
+// CRDUpgradeSafety preflight check and can be extended in the future to add other preflight checks.
+type PreflightConfig struct {
+	CRDUpgradeSafety CRDUpgradeSafetyPreflightConfig `json:"crdUpgradeSafety,omitempty"`
+}
+
+// CRDUpgradeSafetyPreflightConfig is a custom type to hold value for the CRDUpgradeSafetyPreflightConfig
+// that enables or disables the CRD Upgrade Safety validations.
+type CRDUpgradeSafetyPreflightConfig string
```
everettraven, line 90: We may want to make this a struct so we can expand upon it in the future by adding additional optional fields. I'm fine with the enum approach over a boolean and it can be the first field in this new struct

api/v1alpha1/clusterextension_types.go
```diff
@@ -73,8 +73,32 @@ type ClusterExtensionSpec struct {
...
+
+const (
+	// CRDUpgradeSafetyPreflightConfigApply represents the default state for the
+	// CRD Upgrade Safety validations by setting it to "Apply".
+	CRDUpgradeSafetyPreflightConfigApply CRDUpgradeSafetyPreflightConfig = "Apply"
+
+	// CRDUpgradeSafetyPreflightConfigSkip represents the option for the
+	// CRD Upgrade Safety validations to be skipped by setting it to "Skip".
+	CRDUpgradeSafetyPreflightConfigSkip CRDUpgradeSafetyPreflightConfig = "Skip"
```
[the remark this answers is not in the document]
rashmigottipati: Works for me. 
Addressed in [commit](https://github.com/operator-framework/operator-controller/pull/845/commits/e35ad160f1dbd3459bcebf0a20c0f109b49f4982)

api/v1alpha1/clusterextension_types.go
rashmigottipati: Agreed 
Updated in [commit](https://github.com/operator-framework/operator-controller/pull/845/commits/e35ad160f1dbd3459bcebf0a20c0f109b49f4982)

api/v1alpha1/clusterextension_types.go
```diff
@@ -73,8 +73,37 @@ type ClusterExtensionSpec struct {
...
+}
+
+// CRDUpgradeSafetyPreflightConfig is a custom struct that holds the necessary configuration values to
+// enable or disable the CRD Upgrade Safety validations. Currently, this holds CRDUpgradeSafetyPreflightConfigType
+// that sets the CRD Upgrade Safety validations to either Enabled/Disabled.
+type CRDUpgradeSafetyPreflightConfig struct {
+	CRDUpgradeSafetyPreflightConfigType CRDUpgradeSafetyPreflightConfigType `json:"crdUpgradeSafetyPreflightConfigType,omitempty"`
 }
 
+type CRDUpgradeSafetyPreflightConfigType string
```
everettraven, line 95: (Optional) Nit: Maybe `CRDUpgradeSafetyPreflightMode`?

api/v1alpha1/clusterextension_types.go
```diff
@@ -73,8 +73,37 @@ type ClusterExtensionSpec struct {
...
+// CRDUpgradeSafety preflight check and can be extended in the future to add other preflight checks.
+type PreflightConfig struct {
+	CRDUpgradeSafety CRDUpgradeSafetyPreflightConfig `json:"crdUpgradeSafety,omitempty"`
+}
+
+// CRDUpgradeSafetyPreflightConfig is a custom struct that holds the necessary configuration values to
+// enable or disable the CRD Upgrade Safety validations. Currently, this holds CRDUpgradeSafetyPreflightConfigType
+// that sets the CRD Upgrade Safety validations to either Enabled/Disabled.
+type CRDUpgradeSafetyPreflightConfig struct {
+	CRDUpgradeSafetyPreflightConfigType CRDUpgradeSafetyPreflightConfigType `json:"crdUpgradeSafetyPreflightConfigType,omitempty"`
```
everettraven, line 92: This name feels very long. Maybe something like `mode` would be nicer to type?

everettraven: I think we should also add a default here and have it be set to `Enabled`

everettraven: Can use https://github.com/operator-framework/operator-controller/pull/845/files#diff-8e9c0eb955489f1a05a95a75a25842a010fdc02029ce06aae7d2163b895c17cbR62-R64 as inspiration for validations and defaults

rashmigottipati: Yeah, updated it to `Mode` and set the default to `Enabled`

https://github.com/operator-framework/operator-controller/pull/845/commits/c4dd90710fb62661560e8972ae87aea92f8f76c6

api/v1alpha1/clusterextension_types.go
```diff
@@ -73,8 +73,41 @@ type ClusterExtensionSpec struct {
...
+
+const (
+	// CRDUpgradeSafetyPreflightConfigEnabled represents the default state for the
+	// CRD Upgrade Safety validations by setting it to "Enabled".
+	CRDUpgradeSafetyPreflightConfigEnabled Mode = "Enabled"
+
+	// CRDUpgradeSafetyPreflightConfigDisabled represents the option for the
+	// CRD Upgrade Safety validations to be disabled by setting it to "Disabled".
+	CRDUpgradeSafetyPreflightConfigDisabled Mode = "Disabled"
+)
```
everettraven, line 109: Instead of this being a more generic type called `Mode`, what do you think about making this type more unique to each preflight check we add? For this case maybe something like:
```suggestion
type CRDUpgradeSafetyMode string

const (
	// CRDUpgradeSafetyModeEnabled represents the default state for the
	// CRD Upgrade Safety validations by setting it to "Enabled".
	CRDUpgradeSafetyModeEnabled Mode = "Enabled"

	// CRDUpgradeSafetyModeDisabled represents the option for the
	// CRD Upgrade Safety validations to be disabled by setting it to "Disabled".
	CRDUpgradeSafetyModeDisabled Mode = "Disabled"
)
```

My thinking here is that this more easily allows us to manage/maintain a unique set of modes for each preflight check. I could totally be wrong here though and this is just a suggestion - let me know what you think!

Maybe @joelanford also has some opinions here?

rashmigottipati: I intentionally had it as `Mode` as it's going to be a property within the CRD Upgrade Safety field so it felt like it may be redundant to have it as something like `CRDUpgradeSafetyMode`.
So for example, let's say we add another preflight check called `ClusterHealthValidation`, below is how the `ClusterExtension` yaml would look like:

```              
              preflight:
                properties:
                  clusterHealthValidation:
                    properties:
                      mode:
                        default: Enabled
                        enum:
                        - Enabled
                        - Disabled
                        type: string
                    type: object
                  crdUpgradeSafety:
                    properties:
                      mode:
                        default: Enabled
                        enum:
                        - Enabled
                        - Disabled
                        type: string
                    type: object
                type: object
```

But if it helps to differentiate easily and visually, then I think changing it to `CRDUpgradeSafetyMode` to be more specific would be better. My opinion is that it may be a bit redundant, but I am open to changing it if you feel it's better that way.

Also, I like the shortened names: `CRDUpgradeSafetyModeEnabled` and `CRDUpgradeSafetyModeDisabled`. :+1:

everettraven: I'm fine with a single `Mode` type as long as there is an easy way to differentiate which preflight checks can use which modes. If we are going with a single type, maybe something like `PreflightMode` is a bit more descriptive?

rashmigottipati: Addressed in [latest commit](https://github.com/operator-framework/operator-controller/pull/845/commits/224d1393d317567db4c1f17208d31dd444772080) based on your original suggestion as I like it better and it wouldn't hurt to be specific :)

api/v1alpha1/clusterextension_types.go
```diff
@@ -73,8 +73,41 @@ type ClusterExtensionSpec struct {
...
+
+// CRDUpgradeSafetyPreflightConfig is a custom struct that holds the necessary configuration values to
+// enable or disable the CRD Upgrade Safety validations. Currently, this holds Mode
+// that sets the CRD Upgrade Safety validations to either Enabled/Disabled.
+// By default, this is set to "Enabled".
+type CRDUpgradeSafetyPreflightConfig struct {
+	//+kubebuilder:validation:Enum:=Enabled;Disabled
+	//+kubebuilder:default:=Enabled
+	//+kubebuilder:Optional
+	CRDUpgradeSafetyMode CRDUpgradeSafetyMode `json:"crdUpgradeSafetyMode,omitempty"`
```
everettraven, line 96: I think for the field name I'd still call this mode otherwise you have to define it in yaml like:
```yaml
...
preflight:
  crdUpgradeSafety:
    crdUpgradeSafetyMode: Disabled
...
```

which, IMO, feels redundant as opposed to:
```yaml
...
preflight:
  crdUpgradeSafety:
    mode: Disabled
...
```

rashmigottipati: Ah my bad, based on [this comment](https://github.com/operator-framework/operator-controller/pull/845#discussion_r1600201812), I had assumed that you were suggesting to change the field name in the struct and also the json tag to `Mode`. 
Just realized you were suggesting to change the field name in the struct alone. 

Fixed it now in [commit](https://github.com/operator-framework/operator-controller/pull/845/commits/fffc98a0a57f9d860aab3a2347a36bd8edfe28bd)

api/v1alpha1/clusterextension_types.go
```diff
@@ -73,8 +73,41 @@ type ClusterExtensionSpec struct {
...
+
+// CRDUpgradeSafetyPreflightConfig is a custom struct that holds the necessary configuration values to
+// enable or disable the CRD Upgrade Safety validations. Currently, this holds Mode
+// that sets the CRD Upgrade Safety validations to either Enabled/Disabled.
+// By default, this is set to "Enabled".
+type CRDUpgradeSafetyPreflightConfig struct {
+	//+kubebuilder:validation:Enum:=Enabled;Disabled
+	//+kubebuilder:default:=Enabled
+	//+kubebuilder:Optional
+	CRDUpgradeSafetyMode CRDUpgradeSafetyMode `json:"mode,omitempty"`
```
everettraven, line 96: nit:
```suggestion
	Mode CRDUpgradeSafetyMode `json:"mode,omitempty"`
```

everettraven (approved):
One minor nit, but not worth blocking this PR from merging IMO. Thanks @rashmigottipati !

api/v1alpha1/clusterextension_types.go
```diff
@@ -73,8 +73,41 @@ type ClusterExtensionSpec struct {
 	// the bundle may contain resources that are cluster-scoped or that are
 	// installed in a different namespace. This namespace is expected to exist.
 	InstallNamespace string `json:"installNamespace"`
+
+	//+kubebuilder:Optional
+	// Preflight defines the strategy for the preflight checks (i.e. as of now, the CRD upgrade safety checks) to be applied or skipped when attempting to install the cluster extension.
+	Preflight PreflightConfig `json:"preflight,omitempty"`
```
joelanford, line 79: If a field is optional (i.e. with `omitempty`), it should also be a pointer.

api/v1alpha1/clusterextension_types.go
```diff
@@ -73,8 +73,41 @@ type ClusterExtensionSpec struct {
...
+}
+
+// PreflightConfig holds the config for the preflight checks. Currently, this is implemented for the
+// CRDUpgradeSafety preflight check and can be extended in the future to add other preflight checks.
+type PreflightConfig struct {
+	CRDUpgradeSafety CRDUpgradeSafetyPreflightConfig `json:"crdUpgradeSafety,omitempty"`
```
joelanford, line 85: Same comment as above: re using pointer if field is optional.

varshaprasad96:
Please hold this from merging till #846 gets in. Thanks!

joelanford:
@rashmigottipati should we close this? Am I correct in assuming that #979 makes this obsolete?

rashmigottipati:
@joelanford yes, https://github.com/operator-framework/operator-controller/pull/979 makes this obsolete. Will close this

PR #846, opened 2024-05-10
✨ Modify ClusterExtension to use Helm under the hood to apply contents into the cluster

bentito:
# Description

ClusterExtension reconciler needs to perform the following actions:

[See this [doc](https://docs.google.com/document/d/1sPNHsmsSVVSPzFqMHsksUSLHvGM9zzGpvtG_Q4E04qI/) for more detail if possible.]

1. **Based on the cluster Extension spec, get the list of bundles from the catalog and perform resolution.**
   - The resolution here need not use dep.py, it can be simple filtering and sorting as done in extension controller currently.
     1. **Ref:** [Extension controller code](https://github.com/operator-framework/operator-controller/blob/49be998193d8dc1bd4a9fe09ad8fc10cccfb4a42/internal/controllers/extension_controller.go#L151)
     2. **Resolution:** [Extension controller resolution](https://github.com/operator-framework/operator-controller/blob/49be998193d8dc1bd4a9fe09ad8fc10cccfb4a42/internal/controllers/extension_controller.go#L480-L542)
     3. The resolution step involves fetching the installed version. Since there are no APIs (like BundleDeployment or App) present in clusters that reference the installed bundle, we need to investigate on listing the installed helm charts or use helm secrets. This also means that when the helm chart is created, we need to store metadata that uniquely identifies an extension.

2. **Unpack the contents of the resolved bundle. For this iteration we can cover only registryV1 bundles from an image source.**
   - Unpacker interface reference: [Unpacker interface](https://github.com/operator-framework/rkupak/blob/6cc42b9f1e7f432c8da29948e0dc47d62ec7c428/internal/source/unpacker.go#L31-L33)
   - Image unpacker: [Image image unpacker](https://github.com/operator-framework/rkupak/blob/6cc42b9f1e7f432c8da29948e0dc47d62ec7c428/internal/source/image.go#L37)
   - Instead of the BundleDeployment API we need to pass in the cluster extension directly.

3. **Conversion of registryV1 to plain bundle.**
   - Conversion logic: [Convert registryV1](https://github.com/operator-framework/rkupak/blob/6cc42b9f1e7f432c8da29948e0dc47d62ec7c428/internal/convert/registryv1.go#L44)

4. **Creating a Helm chart from the resources present in plain bundle.**
   - [Plain bundle provisioner](https://github.com/operator-framework/rkupak/blob/6cc42b9f1e7f432c8da29948e0dc47d62ec7c428/internal/provisioner/plain.go#L28-L36)

5. **Installing helm chart onto the cluster.**
   - [Helm chart deployer](https://github.com/operator-framework/rkupak/blob/6cc42b9f1e7f432c8da29948e0dc47d62ec7c428/internal/controllers/bundledeployment/bundledeployment.go#L325-L422)
   - We need to be able to inject metadata into the chart that will be required for resolution as mentioned in 1.(iii).
     - Currently, we run a post render hook, that adds labels on the objects created by the chart: [Post render hook details](https://github.com/operator-framework/rkupak/blob/6cc42b9f1e7f432c8da29948e0dc47d62ec7c428/internal/controllers/bundledeployment/bundledeployment.go#L318-L323). Will need to figure out on how to store this info in the helm secret or on the helm chart directory.

## Reviewer Checklist

- [ ] API Go Documentation


bentito:
@tmshort thanks for deconflicting main.go 👍

cmd/unpack/main.go
```diff
@@ -0,0 +1,121 @@
...
+	)
+	cmd := &cobra.Command{
+		Use:  "unpack",
+		Args: cobra.ExactArgs(0),
+		RunE: func(cmd *cobra.Command, _ []string) error {
+			if opConVersion {
+				// TODO
+				//fmt.Println(version.String())
+				os.Exit(0)
+			}
```
acornett21, line 40: Should a proper logger be used here?

CONTRIBUTING.md
```diff
@@ -19,9 +19,8 @@ The user experience captured in the OLM V1 PRD introduces many requirements that
 - [The Operator-Controller project](https://github.com/operator-framework/operator-controller/), which is the top level component allowing users to specify operators they'd like to install.
 - [The Catalogd project](https://github.com/operator-framework/catalogd/), which hosts operator content and helps users discover installable content.
 - [The Deppy project](https://github.com/operator-framework/deppy/), which enables the operator-controller to identify valid installs and upgrades from the list of installable content provided via the Catalogd project.
-- [The RukPak project](https://github.com/operator-framework/rukpak/), which facilitates the installation of operators.
```
everettraven, line 22: This technically still has a dependency on rukpak as a library right? It might be worth it to still mention rukpak since this is the case.

.goreleaser.yml
```diff
@@ -16,37 +16,50 @@ builds:
     - arm64
     - ppc64le
     - s390x
+  - id: unpack
```
everettraven, line 19: Could we create a follow-up issue to switch from the pod based unpacking method to the direct image registry client approach? Using it would eliminate the need to do all this effort to include this `unpack` binary in the images.

everettraven: I think it would be _ideal_ to use it as part of this PR, but I don't want to make you have to rewrite a bunch of logic since we have a lot of future work held on getting this helm work in.

cmd/manager/main.go
```diff
@@ -75,12 +110,33 @@ func main() {
...
+	}
+
+	dependentRequirement, err := k8slabels.NewRequirement(labels.OwnerKindKey, selection.In, []string{v1alpha1.ClusterExtensionKind})
+	if err != nil {
+		setupLog.Error(err, "unable to create dependent label selector for cache")
+		os.Exit(1)
+	}
+	dependentSelector := k8slabels.NewSelector().Add(*dependentRequirement)
+
+	fmt.Println("set up manager")
```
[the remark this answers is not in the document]
everettraven: If this is intended to stick around +1 to using the logger. I could also see this being a debugging statement that was missed in a cleanup phase.

.goreleaser.yml
bentito: Yep, created the issue here: https://github.com/operator-framework/operator-controller/issues/853

cmd/unpack/main.go
bentito: yep, done in: f5555202

cmd/manager/main.go
```diff
@@ -101,21 +157,58 @@ func main() {
...
 		os.Exit(1)
 	}
+
+	// if err = (&controllers.ExtensionReconciler{
+	// 	Client:         cl,
+	// 	BundleProvider: catalogClient,
+	// }).SetupWithManager(mgr); err != nil {
+	// 	setupLog.Error(err, "unable to create controller", "controller", "Extension")
+	// 	os.Exit(1)
+	// }
```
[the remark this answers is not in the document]
bentito: yep, done in f5555202

CONTRIBUTING.md
bentito: @varshaprasad96 any thoughts on restoring this mention?

varshaprasad96: Rukpak till now has been mentioned as a separate entity/API controller to manage installation. I would prefer removing this as users should not be concerned where our helpers reside. We are not intending to support the APIs provided in Rukpak separately, so having it not mentioned is good.

api/v1alpha1/clusterextension_types.go
```diff
@@ -77,15 +82,20 @@ type ClusterExtensionSpec struct {
 
 const (
 	// TODO(user): add more Types, here and into init()
-	TypeInstalled = "Installed"
-	TypeResolved  = "Resolved"
+	TypeInstalled      = "Installed"
+	TypeResolved       = "Resolved"
+	TypeHasValidBundle = "HasValidBundle"
+	TypeHealthy        = "Healthy"
```
varshaprasad96, line 88: Introducing health checks becomes tricky when we have to deal with adding service account. And Health checks are not a part of v1 cut. So based on discussions, we should be good to remove `TypeHealthy` for now, and add it later when the time comes (as we have a clear architecture designed for SA inclusion in spec).

varshaprasad96: Addressed by https://github.com/operator-framework/operator-controller/pull/854, which would merge into this branch soon!

test/e2e/cluster_extension_install_test.go
```diff
@@ -184,10 +183,11 @@ func TestClusterExtensionInstallReResolvesWhenNewCatalog(t *testing.T) {
 }
 
 func TestClusterExtensionBlockInstallNonSuccessorVersion(t *testing.T) {
+	t.Skip("Skipping tests related to upgrades")
```
m1kola, line 186: Why?

api/v1alpha1/clusterextension_types.go
```diff
@@ -95,13 +105,18 @@ const (
 	ReasonResolutionUnknown         = "ResolutionUnknown"
 	ReasonSuccess                   = "Success"
 	ReasonDeprecated                = "Deprecated"
+	ReasonErrorGettingReleaseState  = "ErrorGettingReleaseState"
+	ReasonUpgradeFailed             = "UpgradeFailed"
+	ReasonCreateDynamicWatchFailed  = "CreateDynamicWatchFailed"
```
m1kola, line 110: Do we need reasons like `ErrorGettingClient`, `CreateDynamicWatchFailed`, etc? Do clients care about this?

internal/controllers/clusterextension_controller.go
```diff
@@ -251,100 +451,7 @@ func (r *ClusterExtensionReconciler) resolve(ctx context.Context, ext *ocv1alpha
...
-			return fmt.Errorf(
-				"bundle %q has a dependency declared via property %q which is currently not supported",
-				bundle.Name,
-				bundle.Properties[i].Type,
-			)
-		}
-	}
-
-	return nil
-}
```
m1kola, line 300: I see it was removed but so far I do not see an equivalent of this. This is probably why you had to skip the test in `internal/controllers/clusterextension_registryv1_validation_test.go` (see my other comment there).

m1kola (commented):
I haven't reviewed the controller itself properly and some other bits. Some comments so far.

---

I think this PR too big to be adequately reviewed. Is there a way we can split this into multiple smaller PRs? I might be missing something, but it looks like unpacker is fairly independent and can be a separate PR. Then changes to goreleaser to make it buildable can be a separate PR. A bunch of helper functions can be moved into a separate PR, I think.

api/v1alpha1/clusterextension_types.go
bentito: I find the matrix of possibilities a bit overwhelming. And I think we should embed the matrix in the repo docs. But on the other hand, the many reasons basically represent different things to log about which overall is good for debugging and getting better field bug reports.

joelanford: Logging and Condition reasons are different concerns. I agree with @m1kola here. Condition types are reasons are part of our public API. We should keep those as minimal as we possibly can.

Reasons are generally for software clients.
Messages are for humans. If there isn't a use case in mind for some other software needing to distinguish between reasons of a condition, we should not have explicit reasons for them. Just put the details in a message for a human to assess.

CONTRIBUTING.md
everettraven: This is in the `CONTRIBUTING.md` where those interested in contributing would go. I think it is reasonable to mention rukpak if we are intending for libraries/helpers to be maintained there long term.

internal/packageerrors/packageerrors.go
```diff
@@ -0,0 +1,35 @@
...
+		versionError = fmt.Sprintf(" matching version %q", versionRange)
+	}
+	if channelName != "" {
+		channelError = fmt.Sprintf(" in channel %q", channelName)
+	}
+	if installedVersion != "" {
+		existingVersionError = fmt.Sprintf(" which upgrades currently installed version %q", installedVersion)
+	}
+	return fmt.Errorf("no package %q%s%s%s found", packageName, versionError, channelError, existingVersionError)
+}
```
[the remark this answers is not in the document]
everettraven: Big +1 here. When it comes to the formatting of the code I agree that we should be more explicit and prioritize the experience of being able to pinpoint the where something went wrong based on the error returned.

Dockerfile
```diff
@@ -1,14 +1,21 @@
...
 # 'make build' or 'make release' targets.
+# Stage 1:
+FROM gcr.io/distroless/static:debug-nonroot AS builder
+
+# Stage 2: 
 FROM gcr.io/distroless/static:nonroot
 
+# Grab the cp binary so we can cp the unpack
+# binary to a shared volume in the bundle image
+COPY --from=builder /busybox/cp /cp
```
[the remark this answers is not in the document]
dtfranz: The comment above was intended to explain but maybe needs more detail? `cp` is needed to copy the `unpack` binary to the bundle image. This happens in the `initContainer` for the bundle unpacker pod (this is how rukpak does it).

cmd/manager/main.go
```diff
@@ -17,47 +17,82 @@ limitations under the License.
...
+		cachePath                   string
+		operatorControllerVersion   bool
+		httpExternalAddr            string
+		systemNamespace             string
+		unpackImage                 string
+		provisionerStorageDirectory string
 	)
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
+	flag.StringVar(&httpExternalAddr, "http-external-address", "http://localhost:8080", "The external address at which the http server is reachable.")
```
everettraven, line 87: Do we need an http server here? IIUC this was present in rukpak so clients could query the http server and retrieved a tarball of the unpacked bundle contents. Is there a particular use case where this is needed in operator-controller?

config/manager/kustomization.yaml
```diff
@@ -6,3 +6,23 @@ images:
 - name: controller
   newName: quay.io/operator-framework/operator-controller
   newTag: devel
+replacements:
+- source: # replaces UNPACK_IMAGE in manager.yaml with image set by kustomize above
```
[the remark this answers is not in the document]
dtfranz: I prefer doing it this way because it ensures that `unpack-image` will always equal the image for `manager` without needing to write out the `quay` image more than once. I feel it's more maintainable if we ever change that url or do some custom deployment; we wouldn't have to change it in multiple places. 

Also, IIRC we wanted to eventually remove the `sed "s/:devel/:$(VERSION)/g"` since it's a bit brute-forcey, and if we did then these two image strings would still match without any need to do anything else.

Dockerfile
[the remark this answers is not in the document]
bentito: Expanded the comment in [ff0c3aa](https://github.com/operator-framework/operator-controller/pull/846/commits/ff0c3aa6886e0af8b48dec5ae2e01e601f2f0184)

cmd/unpack/main.go
everettraven, line 1: I know it was marked as a follow up item in #853 - but what do folks think of adopting the direct image registry client approach as part of this PR instead of a follow up? I think making use of it would significantly reduce the size of this PR by no longer needing to have this `unpack` binary and all the changes that came with it.

everettraven: On the flip side, I do think it could _possibly_ add complexity for the e2e set up do to needing to load all the bundles used for e2e tests onto the on cluster image registry

cmd/manager/main.go
```diff
@@ -101,9 +157,46 @@ func main() {
...
+		os.Exit(1)
+	}
+
+	acg, err := helmclient.NewActionClientGetter(cfgGetter)
+	if err != nil {
+		setupLog.Error(err, "unable to create helm client")
+		os.Exit(1)
+	}
+
+	unpacker, err := source.NewDefaultUnpacker(mgr, systemNamespace, unpackImage, (*x509.CertPool)(nil))
```
everettraven, line 174: IIRC the default unpacker has more source types than we are likely concerned with for now. Does it make sense to create a slimmed down unpacker that only has the image source for unpacking?

cmd/manager/main.go
```diff
@@ -101,9 +157,46 @@ func main() {
...
+	storageURL, err := url.Parse(fmt.Sprintf("%s/bundles/", httpExternalAddr))
+	if err != nil {
+		setupLog.Error(err, "unable to parse bundle content server URL")
+		os.Exit(1)
+	}
+
+	localStorage := &storage.LocalDirectory{
+		RootDirectory: provisionerStorageDirectory,
+		URL:           *storageURL,
+	}
```
everettraven, line 189: Instead of having a storage method that stands up an HTTP server, I think it could make more sense to either:
1. Always unpack and use the returned contents (could be optimized with a cache in the future - especially since a direct image registry client was recently added to the rukpak library that uses a cache and there are plans to switch to this in a follow-up)
2. Still have a `Storage` interface, but write an implementation that doesn't stand up an HTTP server and essentially acts as a "cache" for the unpacked bundle contents.

config/rbac/role.yaml
```diff
@@ -4,6 +4,12 @@ kind: ClusterRole
 metadata:
   name: manager-role
 rules:
+- apiGroups:
+  - '*'
+  resources:
+  - '*'
+  verbs:
+  - '*'
```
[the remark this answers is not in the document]
varshaprasad96: The operator controller needs to watch all the bundle objects and reconcileaccordingly. Though not ideal, but these permissions are required. Thishas been taken from rukpak, and an issue was created before to discuss. Added a comment in yaml. Going to keep this as is for now, to discuss in follow up. This should also be taken care by us using service accounts in the future.

internal/catalogmetadata/types.go
```diff
@@ -73,7 +73,7 @@ func (b *Bundle) loadPackage() error {
 	if b.semVersion == nil {
 		semVer, err := bsemver.Parse(b.bundlePackage.Version)
 		if err != nil {
-			return fmt.Errorf("could not parse semver %q for bundle '%s': %s", b.bundlePackage.Version, b.Name, err)
+			return fmt.Errorf("could not parse semver %q for bundle '%s': %s", b.bundlePackage.Version, b.Bundle.Name, err)
```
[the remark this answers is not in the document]
varshaprasad96: This got in because of a rebase probably. Undoing it.

internal/controllers/clusterextension_controller_test.go
```diff
@@ -120,19 +119,17 @@ func TestClusterExtensionNonExistentVersion(t *testing.T) {
...
-	require.Equal(t, "installation has not been attempted as resolution failed", cond.Message)
+	require.Equal(t, metav1.ConditionFalse, cond.Status)
 
 	verifyInvariants(ctx, t, reconciler.Client, clusterExtension)
 	require.NoError(t, cl.DeleteAllOf(ctx, &ocv1alpha1.ClusterExtension{}))
-	require.NoError(t, cl.DeleteAllOf(ctx, &rukpakv1alpha2.BundleDeployment{}))
 }
 
 func TestClusterExtensionBundleDeploymentDoesNotExist(t *testing.T) {
+	t.Skip("Skip and replace with e2e for BundleDeployment")
```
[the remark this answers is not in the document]
varshaprasad96: The BD API would be gone, but it was intentional to keep this for now rather than removing it because we want to test a similar scenario with ClusterExtension. This test may not even be relevant and just go away, but keeping the tests as is for now rather than removing them which we will handle in follow up.

internal/controllers/clusterextension_controller.go
varshaprasad96: This relates to the comment made here: https://github.com/operator-framework/operator-controller/pull/758/files#r1586496341. We have instead used labels on the helm release to get the installed version rather than fetching it from the bundle itself - which is what `getInstalledVersion` (https://github.com/operator-framework/operator-controller/blob/ff0c3aa6886e0af8b48dec5ae2e01e601f2f0184/internal/controllers/clusterextension_controller.go#L686) does here. There is no `catalogmetadata.Bundle` returned anymore. 

Are we missing anything?

everettraven: I think @m1kola is referring to the removal of the bundle validation logic. This should still be in place and used somewhere to validate that this is a "simple" registry+v1 bundle.

varshaprasad96: Ah, it got missed in the rebase. A follow up commit is in progress to include it.

api/v1alpha1/clusterextension_types.go
varshaprasad96: There could be argument both ways, as I see clients wanting to know why a bundle which was supposed to be watched not triggering a reconcile, or why an installed version is unavailable. These were taken in from Rukpak. 

But for now, have removed it from operator-controller, and using the ones in Rukpak (which we don't intend to support as an API).

internal/controllers/clusterextension_registryv1_validation_test.go
```diff
@@ -97,9 +96,9 @@ func TestClusterExtensionRegistryV1DisallowDependencies(t *testing.T) {
 		},
 	} {
 		t.Run(tt.name, func(t *testing.T) {
+			t.Skip("Skip and replace with e2e for BundleDeployment")
```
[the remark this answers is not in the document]
varshaprasad96: Reply in https://github.com/operator-framework/operator-controller/pull/846#discussion_r1601378742

cmd/manager/main.go
varshaprasad96: This was being used to expose the unpacked contents, which was then added to status as a `contentURL`. Removing this for now, since we are not adding the contentURL in the status anymore. (ref: https://github.com/operator-framework/operator-controller/pull/826#discussion_r1588621788)

cmd/unpack/main.go
varshaprasad96: We did discuss this, and lean towards doing it in a follow up. Introducing that in here, is going to have us modify the existing e2e's again. The major purpose of the PR is to get Helm running under the hood to install contents, and that is something we have achieved here. I'm not sure if having it modified again is going to reduce diffs to a greater extent - the only major change would in Makefile, and unpack build code which is a direct copy from rukpak.

cmd/manager/main.go
varshaprasad96: The other unpacker sources are going to be removed from Rukpak too. So the change would be in there. This was brought up before, but we then decided to do the bare minimal changes in Rukpak and o-c and not introduce a new unpacker. We can create a follow up to handle this in Rukpak.

cmd/manager/main.go
varshaprasad96: Removed the need of having an HTTP endpoint for now. We now unpack contents and store in `/var/cache/bundles` and load it from there. But this does relate to the comment [here](https://github.com/operator-framework/rukpak/pull/874#discussion_r1592945966) where we can have the direct image registry client write in a common location, no longer requiring a storage dir.

config/rbac/role.yaml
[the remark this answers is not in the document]
varshaprasad96: Added a comment above the cluster_extension reconciler to explain this.

internal/controllers/clusterextension_controller.go
varshaprasad96: Re-added https://github.com/operator-framework/operator-controller/pull/846/files#diff-a54f79b762054ccdc31e4011d28cce22af8f54d43e1fddc75745ce042f177f7eR227

internal/controllers/clusterextension_controller.go
```diff
@@ -112,82 +179,232 @@ func checkForUnexpectedFieldChange(a, b ocv1alpha1.ClusterExtension) bool {
...
+				setResolvedStatusConditionFailed(&ext.Status.Conditions, errorMessage, ext.Generation)
+				ensureAllConditionsWithReason(ext, "ResolutionFailed", errorMessage)
+			} else if strings.Contains(errorMessage, "invalid version range") {
+				// Handle invalid version range errors, potentially setting status conditions
+				setResolvedStatusConditionFailed(&ext.Status.Conditions, errorMessage, ext.Generation)
+				ensureAllConditionsWithReason(ext, "ResolutionFailed", errorMessage)
+			} else {
+				// General error handling
+				setResolvedStatusConditionFailed(&ext.Status.Conditions, errorMessage, ext.Generation)
+				ensureAllConditionsWithReason(ext, "InstallationStatusUnknown", "")
```
varshaprasad96, line 198: @bentito Do we still need this func? Can we set the resolution error directly and test the returned string - as we did before?

internal/controllers/clusterextension_controller.go
```diff
@@ -112,82 +179,232 @@ func checkForUnexpectedFieldChange(a, b ocv1alpha1.ClusterExtension) bool {
...
+			Reason:  ocv1alpha1.ReasonBundleLoadFailed,
+			Message: err.Error(),
+		})
 		return ctrl.Result{}, err
 	}
 
-	// Assume all bundles are registry+v1 for now, since that's all we support.
-	// If the bundle is not a registry+v1 bundle, the conversion will fail.
-	bundleProvisioner := "core-rukpak-io-registry"
+	chrt, values, err := r.Handler.Handle(ctx, bundleFS, ext)
```
varshaprasad96, line 293: Follow up: https://github.com/operator-framework/operator-controller/issues/867

cmd/manager/main.go
```diff
@@ -101,9 +155,40 @@ func main() {
...
+		os.Exit(1)
+	}
+
+	unpacker, err := source.NewDefaultUnpacker(mgr, systemNamespace, unpackImage, (*x509.CertPool)(nil))
+	if err != nil {
+		setupLog.Error(err, "unable to create unpacker")
+		os.Exit(1)
+	}
+
+	localStorage := &storage.LocalDirectory{
```
varshaprasad96, line 178: Follow up: https://github.com/operator-framework/operator-controller/issues/869

internal/controllers/clusterextension_controller.go
m1kola: If we do need it - would be great to replace `"ResolutionFailed"` and `"InstallationStatusUnknown"` with constants from `api/v1alpha1/clusterextension_types.go`.

internal/controllers/clusterextension_controller.go
```diff
@@ -112,82 +179,232 @@ func checkForUnexpectedFieldChange(a, b ocv1alpha1.ClusterExtension) bool {
 	return !equality.Semantic.DeepEqual(a, b)
 }
 
+func (r *ClusterExtensionReconciler) handleResolutionErrors(ext *ocv1alpha1.ClusterExtension, err error) (ctrl.Result, error) {
+	var aggErrs utilerrors.Aggregate
+	if errors.As(err, &aggErrs) {
+		for _, err := range aggErrs.Errors() {
+			errorMessage := err.Error()
+			if strings.Contains(errorMessage, "no package") {
```
m1kola, line 187: Relying on the error message (here and below) does not sound like a good idea to me.

internal/controllers/clusterextension_controller.go
```diff
@@ -112,82 +179,232 @@ func checkForUnexpectedFieldChange(a, b ocv1alpha1.ClusterExtension) bool {
...
-	bundle, err := r.resolve(ctx, ext)
+	l := log.FromContext(ctx).WithName("operator-controller")
+	// run resolution
+	bundle, err := r.resolve(ctx, *ext)
 	if err != nil {
-		ext.Status.InstalledBundle = nil
-		setInstalledStatusConditionUnknown(&ext.Status.Conditions, "installation has not been attempted as resolution failed", ext.GetGeneration())
-		ext.Status.ResolvedBundle = nil
-		setResolvedStatusConditionFailed(&ext.Status.Conditions, err.Error(), ext.GetGeneration())
+		l.V(1).Info("bundle resolve error", "error", err)
```
m1kola, line 231: Nit: I think this logging can be dropped. I believe errors returned from `Reconcile` are being logged by controller-runtime anyway.

internal/controllers/clusterextension_controller.go
m1kola: Also - why unknown here and below? We know that resolution failed.

In general - it looks like we can get rid of this and `ensureAllConditionsWithReason`. I imagine that we can just do something like that:

```go
	if err != nil {
		setResolvedStatusConditionFailed(&ext.Status.Conditions, err.Error(), ext.Generation)
		// setInstalledStatusConditionFailed, etc
		return ctrl.Result{}, err
	}
```

internal/controllers/clusterextension_controller.go
```diff
@@ -112,82 +179,232 @@ func checkForUnexpectedFieldChange(a, b ocv1alpha1.ClusterExtension) bool {
...
+		return r.handleResolutionErrors(ext, err)
+	}
 
-		setDeprecationStatusesUnknown(&ext.Status.Conditions, "deprecation checks have not been attempted as resolution failed", ext.GetGeneration())
+	if err := r.validateBundle(bundle); err != nil {
+		setInstalledStatusConditionFailed(&ext.Status.Conditions, err.Error(), ext.GetGeneration())
+		setDeprecationStatusesUnknown(&ext.Status.Conditions, "deprecation checks have not been attempted as installation has failed", ext.GetGeneration())
```
m1kola, line 237: Do we only need to set deprecation here? It seems to me that we need to set it under `bundle.Version()` and probably other places below.

internal/labels/labels.go
```diff
@@ -0,0 +1,12 @@
+package labels
+
+const (
+	OwnerKindKey = "olm.operatorframework.io/owner-kind"
+	OwnerNameKey = "olm.operatorframework.io/owner-name"
+
+	// Helm Secret annotations use the regex `(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?`
+	// to validate labels. Open to any suggestions, keeping this for now.
```
m1kola, line 8: Could someone please provide more context for this?

m1kola (commented):
Another set of comments. Will come back to this on Monday.

Note for myself: I was looking at `internal/controllers/clusterextension_controller.go` and paused my "linear" pass at around `bundleVersion, err := bundle.Version()`.

dtfranz:
FYI for reviewers: I've merged a new commit to the branch which re-enables all e2e tests; they should be passing now. If you like, you can view the standalone diff [here](https://github.com/operator-framework/operator-controller/pull/846/commits/53577db9b98362a820f43ad23d877cd4daa06262).

api/v1alpha1/clusterextension_types.go
```diff
@@ -111,13 +121,15 @@ func init() {
...
-		ReasonInstallationStatusUnknown,
-		ReasonInvalidSpec,
 		ReasonSuccess,
 		ReasonDeprecated,
+		ReasonUpgradeFailed,
+		ReasonBundleLoadFailed,
+		ReasonErrorGettingClient,
+		// TODO: this reason is not being used in the reconciler, it will be removed
+		// when we fix the tests. Avoiding removal here, to reduce diffs.
+		ReasonInstallationStatusUnknown,
```
everettraven, line 132: Now that there has been a commit to fix the tests can this reason be removed?

internal/controllers/clusterextension_controller.go
```diff
@@ -17,78 +17,121 @@ limitations under the License.
...
 func (r *ClusterExtensionReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	l := log.FromContext(ctx).WithName("operator-controller")
 	l.V(1).Info("starting")
 	defer l.V(1).Info("ending")
 
 	var existingExt = &ocv1alpha1.ClusterExtension{}
-	if err := r.Get(ctx, req.NamespacedName, existingExt); err != nil {
-		return ctrl.Result{}, client.IgnoreNotFound(err)
+	if err := r.Client.Get(ctx, req.NamespacedName, existingExt); err != nil {
+		return ctrl.Result{}, utilerrors.NewAggregate([]error{client.IgnoreNotFound(err), nil})
```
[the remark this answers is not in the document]
everettraven: +1. Not sure why this was changed from simply returning `client.IgnoreNoteFound(err)`

internal/controllers/clusterextension_controller.go
```diff
@@ -97,12 +140,36 @@ func (r *ClusterExtensionReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
+			updateErrors = append(updateErrors, updateErr)
 		}
 	}
 
-	return res, reconcileErr
+	if reconcileErr != nil {
+		updateErrors = append(updateErrors, reconcileErr)
+	}
+
+	return res, utilerrors.NewAggregate(updateErrors)
```
everettraven, line 152: Optional Nit: Go's std lib `errors` package now has an [`errors.Join()`](https://github.com/operator-framework/helm-operator-plugins/pull/337) function that joins errors into a single error similar to this Aggregate error from the k8s util/errors package. Where possible I'd personally prefer we use the std lib than another library (even if it is a k8s library)

internal/controllers/clusterextension_controller.go
```diff
@@ -17,78 +17,121 @@ limitations under the License.
...
+	var updateErrors []error
+
 	// Do checks before any Update()s, as Update() may modify the resource structure!
 	updateStatus := !equality.Semantic.DeepEqual(existingExt.Status, reconciledExt.Status)
 	updateFinalizers := !equality.Semantic.DeepEqual(existingExt.Finalizers, reconciledExt.Finalizers)
 	unexpectedFieldsChanged := checkForUnexpectedFieldChange(*existingExt, *reconciledExt)
 
 	if updateStatus {
-		if updateErr := r.Status().Update(ctx, reconciledExt); updateErr != nil {
-			return res, utilerrors.NewAggregate([]error{reconcileErr, updateErr})
```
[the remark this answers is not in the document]
everettraven: Looks like we do still return an aggregated error at the end if any errors occur throughout the process. I can't say for certain, but I imagine this is useful in ensuring we update as much as we can before returning any errors. (i.e if we don't have permissions to update the status subresource but we _do_ have permissions to update the finalizers subresource then we should still update the finalizers even if we fail to update the status)

internal/controllers/clusterextension_controller.go
everettraven: +1, if we need to perform functionality dependent on the error that occurred we should build custom error types that are returned by the method/function being called and passed into this function as a parameter

internal/controllers/clusterextension_controller.go
```diff
@@ -112,88 +179,244 @@ func checkForUnexpectedFieldChange(a, b ocv1alpha1.ClusterExtension) bool {
...
+		setDeprecationStatusesUnknown(&ext.Status.Conditions, "deprecation checks have not been attempted as installation has failed", ext.GetGeneration())
+		return ctrl.Result{}, err
+	}
 
-		setDeprecationStatusesUnknown(&ext.Status.Conditions, "deprecation checks have not been attempted as resolution failed", ext.GetGeneration())
+	bundleVersion, err := bundle.Version()
+	if err != nil {
+		ext.Status.ResolvedBundle = nil
+		ext.Status.InstalledBundle = nil
+		setResolvedStatusConditionFailed(&ext.Status.Conditions, fmt.Sprintf("%s:%v", ocv1alpha1.ReasonResolutionFailed, err), ext.GetGeneration())
```
everettraven, line 246: Nit: just put the error message here. The condition in the status should already have the appropriate reason, no need to repeat it in the message.

internal/controllers/clusterextension_controller.go
```diff
@@ -112,88 +179,244 @@ func checkForUnexpectedFieldChange(a, b ocv1alpha1.ClusterExtension) bool {
...
+		updateStatusUnpackPending(&ext.Status, unpackResult)
+		// There must be a limit to number of entries if status is stuck at
+		// unpack pending.
+		return ctrl.Result{}, nil
+	case rukpaksource.StateUnpacking:
+		updateStatusUnpacking(&ext.Status, unpackResult)
+		return ctrl.Result{}, nil
+	case rukpaksource.StateUnpacked:
+		// TODO: Add finalizer to clean the stored bundles, after https://github.com/operator-framework/rukpak/pull/897
+		// merges.
```
everettraven, line 274: Note for other reviewers, this looks like it will be addressed as a follow-up. Looks like https://github.com/operator-framework/operator-controller/issues/868 is the corresponding issue to track this.

@varshaprasad96 (and others) please feel free to correct me if I am incorrect.

internal/controllers/clusterextension_controller.go
```diff
@@ -112,88 +179,244 @@ func checkForUnexpectedFieldChange(a, b ocv1alpha1.ClusterExtension) bool {
...
+	unpackResult, err := r.Unpacker.Unpack(ctx, bd)
+	if err != nil {
+		return ctrl.Result{}, updateStatusUnpackFailing(&ext.Status, err)
+	}
+
+	switch unpackResult.State {
+	case rukpaksource.StatePending:
+		updateStatusUnpackPending(&ext.Status, unpackResult)
+		// There must be a limit to number of entries if status is stuck at
+		// unpack pending.
```
everettraven, line 267: Clarification: Is this a statement or something that needs to be done in the future?

internal/controllers/clusterextension_controller.go
```diff
@@ -112,88 +179,244 @@ func checkForUnexpectedFieldChange(a, b ocv1alpha1.ClusterExtension) bool {
...
-		ext.Status.InstalledBundle = nil
-		setInstalledStatusConditionUnknown(&ext.Status.Conditions, err.Error(), ext.GetGeneration())
-		setDeprecationStatusesUnknown(&ext.Status.Conditions, "deprecation checks have not been attempted as installation has failed", ext.GetGeneration())
+	post := &postrenderer{
+		labels: map[string]string{
+			labels.OwnerKindKey:     ocv1alpha1.ClusterExtensionKind,
+			labels.OwnerNameKey:     ext.GetName(),
+			labels.BundleNameKey:    bundle.Name,
+			labels.PackageNameKey:   bundle.Package,
+			labels.BundleVersionKey: bundleVersion.String(),
```
everettraven, line 318: We should include the catalog in which the bundle was resolved from as well (and support catalog filtering during resolution). There are no guarantees that a bundle is only present/served from a single catalog on the cluster. For example, on OpenShift the Red Hat Operator Catalog and Community Operator Catalog both have some of the same entries (`loki-operator` is one that I found relatively quickly). 

I'm fine with this being done as a follow up if we don't already have support for filtering bundles by catalog name during resolution, but if we do we should ensure we track that information as part of the labels as added by this postrenderer as well IMO.

everettraven: It does not look like there is currently logic to filter resolution to a particular catalog. Not sure it is really in scope for the Helm integration work, but we should probably have an issue tracking this somewhere

dtfranz: That's a good point, thanks Bryce! I'll create a follow-up issue for this.

dtfranz: #873

internal/controllers/clusterextension_registryv1_validation_test.go
```diff
@@ -46,6 +46,9 @@ func TestClusterExtensionRegistryV1DisallowDependencies(t *testing.T) {
 				},
 				CatalogName: "fake-catalog",
 			},
+			// Skipping the happy path, since it requires us to mock unpacker, store and the
+			// entire installation. This should be handled in an e2e instead.
+			skip: true,
```
everettraven, line 51: IMO this is ideally possible, but if that can't be done without a significant refactor then this can be un-skipped in the future when it is possible to unit test this.

internal/controllers/clusterextension_controller_test.go
```diff
@@ -931,51 +497,15 @@ func verifyConditionsInvariants(t *testing.T, ext *ocv1alpha1.ClusterExtension)
...
-		require.Equal(t, tt.clusterExtension.Spec.InstallNamespace, resultBundleDeployment.Spec.InstallNamespace)
-	}
-}
-
 func TestClusterExtensionUpgrade(t *testing.T) {
 	cl, reconciler := newClientAndReconciler(t)
 	ctx := context.Background()
 
 	t.Run("semver upgrade constraints enforcement of upgrades within major version", func(t *testing.T) {
+		t.Skip("Skip and replace with e2e for BundleDeployment")
```
everettraven, line 505: Ideally our controller should be easily unit testable. I am fine with it being a follow up for now, but I don't think we should be replacing unit tests with e2e tests

Makefile
```diff
@@ -162,10 +161,16 @@ e2e-coverage:
...
+kind-deploy: export MANIFEST="./operator-controller.yaml"
 kind-deploy: manifests $(KUSTOMIZE) #EXHELP Install controller and dependencies onto the kind cluster.
 	$(KUSTOMIZE) build $(KUSTOMIZE_BUILD_DIR) > operator-controller.yaml
-	envsubst '$$CATALOGD_VERSION,$$CERT_MGR_VERSION,$$RUKPAK_VERSION,$$MANIFEST' < scripts/install.tpl.sh | bash -s
+	envsubst '$$CATALOGD_VERSION,$$CERT_MGR_VERSION,$$MANIFEST' < scripts/install.tpl.sh | bash -s
+
+.PHONY: kind-redeploy
+kind-redeploy: generate docker-build kind-load kind-deploy #EXHELP Redeploy newly built executables
+	kubectl delete pod -l control-plane=controller-manager -n $(OPERATOR_CONTROLLER_NAMESPACE)
+	envsubst '$$CATALOGD_VERSION,$$CERT_MGR_VERSION,$$MANIFEST' < scripts/install.tpl.sh | bash -s
```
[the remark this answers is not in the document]
dtfranz: Yes, it's for dev workflow, recommended by @joelanford

internal/controllers/clusterextension_registryv1_validation_test.go
bentito: @everettraven , and others, please see my PR (https://github.com/bentito/operator-controller/pull/1) into `helm-poc` branch. This proposes changes to one unit test to allow it to be unskipped (we'd be able to unskip several more if this seems like a reasonable way to go). I tried to keep the mocking level as low as possible, but that necessitates some not doing some of the test. Part of the test still avoided by the later conditional are BundleDeployment related, which I think could just be removed for now, and other parts are the invariant and condition checking stuff -- stuff that we don't set b/c the mocks currently avoid those code paths to limit amount of mocking.

test/e2e/cluster_extension_install_test.go
```diff
@@ -94,6 +94,18 @@ func TestClusterExtensionInstallRegistry(t *testing.T) {
...
+	t.Log("By eventually reporting a successful unpacked")
+	require.EventuallyWithT(t, func(ct *assert.CollectT) {
+		assert.NoError(ct, c.Get(context.Background(), types.NamespacedName{Name: clusterExtension.Name}, clusterExtension))
+		cond := apimeta.FindStatusCondition(clusterExtension.Status.Conditions, rukpakv1alpha2.TypeUnpacked)
+		if !assert.NotNil(ct, cond) {
+			return
+		}
+		assert.Equal(ct, metav1.ConditionTrue, cond.Status)
+		assert.Equal(ct, rukpakv1alpha2.ReasonUnpackSuccessful, cond.Reason)
+		assert.Contains(ct, cond.Message, "Successfully unpacked")
```
m1kola, line 106: Where did `"Successfully unpacked"` come from? I do not see it in operator-controller codebase. I would suggest only assert on strings which come from the code we control.

test/e2e/cluster_extension_install_test.go
```diff
@@ -154,7 +153,7 @@ func TestClusterExtensionInstallReResolvesWhenNewCatalog(t *testing.T) {
 		}
 		assert.Equal(ct, metav1.ConditionFalse, cond.Status)
 		assert.Equal(ct, ocv1alpha1.ReasonResolutionFailed, cond.Reason)
-		assert.Equal(ct, fmt.Sprintf("no package %q found", pkgName), cond.Message)
+		assert.Contains(ct, cond.Message, fmt.Sprintf("no package %q found", pkgName))
```
m1kola, line 156: Nit: I think this can be reverted back to `assert.Equal`. `assert.Contains` doesn't seem necessary here.

test/e2e/cluster_extension_registryV1_limitations_test.go
```diff
@@ -2,60 +2,36 @@ package e2e
...
+	clusterExtension.Spec = ocv1alpha1.ClusterExtensionSpec{
+		PackageName:      "package-with-webhooks",
+		Version:          "1.0.0",
+		InstallNamespace: "default",
 	}
-	err = c.Create(ctx, clusterExtension)
-	defer deleteClusterExtension(clusterExtension)
-	require.NoError(t, err)
+	require.NoError(t, c.Create(ctx, clusterExtension))
+	defer testCleanup(t, catalog, clusterExtension)
```
m1kola, line 26: If there is an error on line 25 `testCleanup` will not be called. It is fine in CI, but this can be annoying locally if you try to re-use kind cluster for repeated test execution.

test/e2e/cluster_extension_registryV1_limitations_test.go
```diff
@@ -2,60 +2,36 @@ package e2e
...
+	defer getArtifactsOutput(t)
 
 	require.EventuallyWithT(t, func(ct *assert.CollectT) {
 		assert.NoError(ct, c.Get(ctx, types.NamespacedName{Name: clusterExtension.Name}, clusterExtension))
-
 		cond := apimeta.FindStatusCondition(clusterExtension.Status.Conditions, ocv1alpha1.TypeInstalled)
-		assert.NotNil(ct, cond)
+		if !assert.NotNil(ct, cond) {
+			return
+		}
```
m1kola, line 34: If you want to stop asserting (which makes sense here) - `require.NotNil` will give the same effect.

In general, functions in `github.com/stretchr/testify/require` are wrappers around functions in `github.com/stretchr/testify/assert` and but they fails the test early.

```suggestion
		require.NotNil(ct, cond)
```

internal/controllers/clusterextension_status.go
```diff
@@ -0,0 +1,72 @@
...
+	meta.SetStatusCondition(&status.Conditions, metav1.Condition{
+		Type:    rukpakv1alpha2.TypeUnpacked,
+		Status:  metav1.ConditionFalse,
+		Reason:  rukpakv1alpha2.ReasonUnpackFailed,
+		Message: err.Error(),
+	})
+	return err
+}
+
+// TODO: verify if we need to update the installBundle status or leave it as is.
```
m1kola, line 41: Is this (and next one) something that needs to be addressed now?

internal/controllers/clusterextension_status.go
```diff
@@ -0,0 +1,72 @@
...
+func updateStatusUnpackPending(status *ocv1alpha1.ClusterExtensionStatus, result *source.Result) {
+	status.ResolvedBundle = nil
+	status.InstalledBundle = nil
```
m1kola, line 44: I think we should not be setting `status.ResolvedBundle` to `nil` here and in `updateStatusUnpacking`. 

As far as I see we call both `updateStatusUnpackPending` and `updateStatusUnpacking` after we successfully "resolved" the bundle. If it is the case - IMO we should keep the info about resolved bundle in status.

What do you think?

m1kola (commented):
Few more comments. I think this time I finished going through all the files. I'll try to do a couple of more passes with a clear mind at some point later because I think I most likely missed something.

internal/controllers/clusterextension_controller.go
tmshort: Looking at the `errors.Join()` API, it does not use error aggregation (maybe not a big deal), but instead specifically states "wrap", but not can not be unwrapped via `errors.Unwrap()`? It also doesn't accept an array of `error`s, but only a list, so it's actually not very flexible... I think you'd actually have to call `errors.Join()` at every point.

tmshort: To clarify, errors returned by `errors.Join()` have an `Unwrap() []error` function, not `Unwrap() error`‽‽

everettraven: >specifically states "wrap", but can not be unwrapped via errors.Unwrap()?

Correct, it has it's own unwrap method that returns the slice of errors.

>It also doesn't accept an array of errors, but only a list, so it's actually not very flexible... I think you'd actually have to call errors.Join() at every point.

In this case you should be able to do:
```go
return res, errors.Join(updateErrors...)
```

---

Totally optional nit to address and I am fine with it staying as an Aggregate error. Just thought that it might be nice to suggest using the new function in the standard library :)

tmshort:
So, after a discussion, we decided to _hopefully_ make it easier on all, to reduce churn on this PR.

We'd like to not make many changes to this PR _unless absolutely necessary_.

We will be creating follow up PRs for specific issues (e.g. errors fixes #878)

internal/controllers/clusterextension_controller.go
```diff
@@ -112,82 +179,232 @@ func checkForUnexpectedFieldChange(a, b ocv1alpha1.ClusterExtension) bool {
...
+2. Validate: Ensure that the bundle returned from the resolution for install meets our requirements.
+3. Unpack: Unpack the contents from the bundle and store in a localdir in the pod.
+4. Install: The process of installing involves:
+4.1 Converting the CSV in the bundle into a set of plain k8s objects.
+4.2 Generating a chart from k8s objects.
+4.3 Apply the release on cluster.
+*/
 //nolint:unparam
 func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alpha1.ClusterExtension) (ctrl.Result, error) {
-	// Lookup the bundle that corresponds to the ClusterExtension's desired package.
```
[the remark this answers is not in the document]
tmshort: See #879

internal/controllers/clusterextension_controller.go
tmshort: Error handling will be resolved in PR #878.

api/v1alpha1/clusterextension_types.go
```diff
@@ -77,31 +82,36 @@ type ClusterExtensionSpec struct {
 
 const (
 	// TODO(user): add more Types, here and into init()
-	TypeInstalled = "Installed"
-	TypeResolved  = "Resolved"
+	TypeInstalled      = "Installed"
+	TypeResolved       = "Resolved"
+	TypeHasValidBundle = "HasValidBundle"
```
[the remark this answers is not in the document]
tmshort: This came over from rukpak, so we want to keep these statuses.
Also, the more information we can provide to the user/customer about what's happening the better.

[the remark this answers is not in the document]
tmshort: These came over from rukpak. We can handle them in a follow up if we don't want to expose them. See #880

internal/controllers/clusterextension_controller.go
```diff
@@ -112,82 +179,232 @@ func checkForUnexpectedFieldChange(a, b ocv1alpha1.ClusterExtension) bool {
...
+		unstructuredObj := &unstructured.Unstructured{Object: uMap}
+		if err := func() error {
+			r.dynamicWatchMutex.Lock()
+			defer r.dynamicWatchMutex.Unlock()
+
+			_, isWatched := r.dynamicWatchGVKs[unstructuredObj.GroupVersionKind()]
+			if !isWatched {
+				if err := r.controller.Watch(
+					source.Kind(r.cache, unstructuredObj),
+					crhandler.EnqueueRequestForOwner(r.Scheme, r.Client.RESTMapper(), ext, crhandler.OnlyControllerOwner()),
```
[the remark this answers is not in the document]
tmshort: Yes, likely.
OWNER: @varshaprasad96 too investigate/update.

internal/controllers/clusterextension_controller.go
```diff
@@ -489,53 +595,72 @@ func (r *ClusterExtensionReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
-		Complete(r)
+		Watches(&corev1.Pod{}, mapOwneeToOwnerHandler(mgr.GetClient(), mgr.GetLogger(), &ocv1alpha1.ClusterExtension{})).
+		Build(r)
 
 	if err != nil {
 		return err
 	}
+	r.controller = controller
+	r.cache = mgr.GetCache()
+	r.dynamicWatchGVKs = map[schema.GroupVersionKind]struct{}{}
```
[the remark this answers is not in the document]
tmshort: See #881 (we may get around to doing this)

internal/catalogmetadata/types.go
```diff
@@ -86,15 +86,15 @@ func (b *Bundle) loadRequiredPackages() error {
 	if b.requiredPackages == nil {
 		requiredPackages, err := loadFromProps[PackageRequired](b, property.TypePackageRequired, false)
 		if err != nil {
-			return fmt.Errorf("error determining bundle required packages for bundle %q: %s", b.Name, err)
+			return fmt.Errorf("error determining bundle required packages for bundle %q: %s", b.Bundle.Name, err)
```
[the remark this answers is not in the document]
tmshort: The linter didn't like this, so this was the resolution.

internal/controllers/clusterextension_controller.go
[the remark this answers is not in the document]
tmshort: See #878

internal/controllers/clusterextension_controller.go
```diff
@@ -112,82 +179,232 @@ func checkForUnexpectedFieldChange(a, b ocv1alpha1.ClusterExtension) bool {
...
+func (r *ClusterExtensionReconciler) resolve(ctx context.Context, ext ocv1alpha1.ClusterExtension) (*catalogmetadata.Bundle, error) {
 	allBundles, err := r.BundleProvider.Bundles(ctx)
 	if err != nil {
-		return nil, err
-	}
-
-	installedBundle, err := r.installedBundle(ctx, allBundles, ext)
-	if err != nil {
-		return nil, err
+		return nil, utilerrors.NewAggregate([]error{fmt.Errorf("error fetching bundles: %w", err)})
```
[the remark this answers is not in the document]
tmshort: See #878

internal/controllers/clusterextension_controller.go
tmshort: See #878

internal/controllers/clusterextension_controller.go
tmshort: Will delete
OWNER: @varshaprasad96

internal/controllers/clusterextension_controller.go
tmshort: Setting conditions should be handled in #842

internal/labels/labels.go
tmshort: There is a limitation on helm annotations. We'll add additional context.
OWNER: @varshaprasad96

api/v1alpha1/clusterextension_types.go
tmshort: We'll remove this and the TODO.
OWNER: @varshaprasad96

internal/controllers/clusterextension_controller.go
tmshort: We'll handle errors in #878.

internal/controllers/clusterextension_controller.go
tmshort: See #882

internal/controllers/clusterextension_registryv1_validation_test.go
bentito: Agree totally on making reconcile more modular and testable. With @varshaprasad96 commits on the #874 PR, I think we've got relevant parts of the unit tests that can be unskipped

tmshort: See #874

internal/controllers/clusterextension_controller_test.go
tmshort: See #874.

tmshort: See #879 for making `reconcile()` more testable, but this test will no longer be skipped after #874 is merged.

internal/controllers/clusterextension_controller.go
```diff
@@ -557,6 +688,137 @@ func clusterExtensionRequestsForCatalog(c client.Reader, logger logr.Logger) han
...
+	if apierrors.IsNotFound(err) {
+		return true
+	}
+
+	// TODO: improve NoKindMatchError matching
+	//   An error that is bubbled up from the k8s.io/cli-runtime library
+	//   does not wrap meta.NoKindMatchError, so we need to fallback to
+	//   the use of string comparisons for now.
+	return strings.Contains(err.Error(), "no matches for kind")
+}
```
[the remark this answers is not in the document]
tmshort: We'll fix this in #878.

internal/controllers/clusterextension_controller.go
```diff
@@ -489,53 +601,72 @@ func (r *ClusterExtensionReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		if err := cl.Get(ctx, oi.key, owner); client.IgnoreNotFound(err) != nil {
+			log.Info("map ownee to owner: get owner",
+				"ownee", client.ObjectKeyFromObject(obj),
+				"owneeKind", owneeGVK,
+				"owner", oi.key,
+				"ownerKind", oi.gvk,
+				"error", err.Error(),
+			)
+			return nil
+		}
```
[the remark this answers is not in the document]
tmshort: This was something done in rukpak, to log the owner of the `oi.key` object.
See https://github.com/operator-framework/rukpak/blob/58305404ab23a4117acb2726c515f44beb6a3b7f/pkg/util/util.go#L93-L102

internal/controllers/clusterextension_controller.go
```diff
@@ -557,6 +688,137 @@ func clusterExtensionRequestsForCatalog(c client.Reader, logger logr.Logger) han
...
+	}
+
+	// find corresponding bundle for the installed content
+	resultSet := catalogfilter.Filter(allBundles, catalogfilter.And(
+		catalogfilter.WithPackageName(release.Labels[labels.PackageNameKey]),
+		catalogfilter.WithBundleName(release.Labels[labels.BundleNameKey]),
+		catalogfilter.InMastermindsSemverRange(vr),
+	))
+	if len(resultSet) == 0 {
+		return nil, fmt.Errorf("bundle %q for package %q not found in available catalogs but is currently installed via helm chart %q in namespace %q", release.Labels[labels.BundleNameKey], ext.Spec.PackageName, release.Name, release.Namespace)
```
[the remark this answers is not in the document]
tmshort: We'll remove the "helm" reference here.
OWNER: @varshaprasad96

test/e2e/cluster_extension_install_test.go
tmshort: This is a case where we're pulling a rukpak API value that should only be used in BundleDeployment. We'll need to update the reason and message checks.
OWNER: @bentito See PR #874

test/e2e/cluster_extension_registryV1_limitations_test.go
tmshort: We'll fix this to be the right pattern.
OWNER: @dtfranz

test/e2e/cluster_extension_registryV1_limitations_test.go
tmshort: This pattern is used in numerous e2e tests.
This is inside an `EventuallyWithT` and using `require` will cause the test to fail prematurely. Using `assert` is the proper mechanism within `EventuallyWithT`.

internal/controllers/clusterextension_status.go
tmshort: See #842

internal/controllers/clusterextension_status.go
tmshort: See #874

test/e2e/cluster_extension_registryV1_limitations_test.go
dtfranz: Fixing in #883

dtfranz: Done!

api/v1alpha1/clusterextension_types.go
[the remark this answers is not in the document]
m1kola: But rukpak was operating on a different CRD. I don't think we should be expanding ClusterExtension API surface. Especially because nothing seems to rely on `TypeHasValidBundle` condition type.

internal/catalogmetadata/types.go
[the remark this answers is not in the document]
m1kola: Are you sure it was this and not something else? We have embedded structs all over the repo referenced like this (e.g. anything using `metav1.ObjectMeta`).

I also just tried `git checkout upstream/main -- internal/catalogmetadata/types.go` and `make lint` - all clear.

internal/controllers/clusterextension_controller.go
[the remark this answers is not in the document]
m1kola: Re-opening. #878 doesn't seem to be relevant.

internal/controllers/clusterextension_controller.go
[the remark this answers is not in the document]
m1kola: Reopening. In rukpak there was [this check](https://github.com/operator-framework/rukpak/blob/58305404ab23a4117acb2726c515f44beb6a3b7f/pkg/util/util.go#L103-L105) for the owner's provisioner class and we do not have it here. So this seems unnecessary.

Also logging only happens in case of an error on a get (anything other than 404):
* Most likely `owner` will be an empty struct anyway on error and logging fields will be empty strings as result
* If we do not query owner here - we do not have to log an issue here. I would expect 404 (or anything else) to be handled on reconcile anyway.

internal/controllers/clusterextension_controller.go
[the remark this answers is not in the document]
varshaprasad96: Reconciler implements client interface. So both of them (r.Scheme and r.Client.Scheme()) are the same and shouldn't matter. Going to make rest mapper to be `r.RESTMapper()` instead for consistency.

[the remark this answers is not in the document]
tmshort: This might have been changed due to the linter...

internal/controllers/clusterextension_controller.go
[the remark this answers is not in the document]
tmshort: #878 is on hold until this PR merges, and #878 will deal with _all_ the error returns. So, any discussions regarding error returns can be done on that PR when it's updated. i.e. we're deferring error handling fixups (to make it easier to review with less churn).

internal/controllers/clusterextension_controller.go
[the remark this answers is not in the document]
tmshort: https://github.com/operator-framework/operator-controller/pull/878 is on hold until this PR merges, and https://github.com/operator-framework/operator-controller/pull/878 will deal with all the error returns. So, any discussions regarding error returns can be done on that PR when it's updated. i.e. we're deferring error handling fixups (to make it easier to review with less churn).

internal/catalogmetadata/types.go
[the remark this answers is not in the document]
tmshort: But that doesn't take into account all the changes... but we'll look at it.
OWNER: @varshaprasad96

internal/controllers/clusterextension_controller.go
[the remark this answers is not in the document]
varshaprasad96: I see the usefulness of this check and logging of the owner in case the obj that caused reconciliation is not existing on cluster. 

If owner is not found - we skip. If not, it cannot be empty, because in line 616 we first query the owner's GVK. Before adding a reconcile request, this is ensuring that the object is not being modified b/w the time we get the owner info (in line 616), to when we add the reconcile request to the queue. If there is any other error apart from 404, it means that the object has been updated/ partially modified and we return without queuing. By doing this we are ensuring that the owner is current before putting in a request for reconcile.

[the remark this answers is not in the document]
varshaprasad96: I would keep this check as is given it's not just for logging.

internal/controllers/clusterextension_controller.go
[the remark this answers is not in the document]
m1kola: @varshaprasad96 I was thinking more the other way around: it looks like the only reasons schema gets passed into the reconciler struct is because it is required on this line. So you can get rid of this extra field on the reconciler if you get the schema from the client.

But I would say it is a nit and a very low priority.

api/v1alpha1/clusterextension_types.go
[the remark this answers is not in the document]
everettraven: I'm all for rethinking what information we are presenting in the status, but I agree with @tmshort that it would be fine to handle as a follow up to limit continued churn on this PR. 

This is the initial implementation taking a lot of inspiration from rukpak, let's create an issue to capture that this needs reconsideration.

internal/controllers/clusterextension_controller.go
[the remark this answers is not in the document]
m1kola: > #878 is on hold until this PR merges, and #878 will deal with _all_ the error returns. So, any discussions regarding error returns can be done on that PR when it's updated. i.e. we're deferring error handling fixups (to make it easier to review with less churn). #878 will be relevant when complete.

But... There is a bug in this line: when object is 404 we just need to return `nil` in the error. With this change - we will always be returning a non-nil error. Unless I'm missing something, this will lead to a constant re-queuing of 404 objects.

Also for some reason there is an extra `nil` in the slice of errors:

```go
return ctrl.Result{}, utilerrors.NewAggregate([]error{client.IgnoreNotFound(err), nil})
//                                                                                ^
```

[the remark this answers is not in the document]
everettraven: >There is a bug in this line: when object is 404 we just need to return nil in the error. With this change - we will always be returning a non-nil error. Unless I'm missing something, this will lead to a constant re-queuing of 404 objects.

My understanding is that this will not always return a non-nil error. If `client.IgnoreNotFound(err)` returns `nil` then all the errors in the list passed to `utilerrors.NewAggregate` are `nil` and `utilerrors.NewAggregate` will return a `nil` error.

Ref: https://github.com/kubernetes/apimachinery/blob/37988e577e160bf7749437141fee1d6b8ca15dd6/pkg/util/errors/errors.go#L46-L61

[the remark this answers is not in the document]
everettraven: I agree with @tmshort that we should move discussion to the new PR tailored specifically to address all error issues as a follow-up to this PR.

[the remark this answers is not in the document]
m1kola: You are right. I did not look into what `NewAggregate` does and assumed it just packs a slice of errors into own type.

internal/controllers/clusterextension_controller.go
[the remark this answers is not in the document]
varshaprasad96: Fixed in https://github.com/operator-framework/operator-controller/pull/887. Both scheme and RESTMapper are from client, so a field on reconciler is no longer required either.

internal/catalogmetadata/types.go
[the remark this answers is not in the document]
varshaprasad96: Fixed in https://github.com/operator-framework/operator-controller/pull/887. Passes linter locally

internal/controllers/clusterextension_controller.go
varshaprasad96: Fixed in https://github.com/operator-framework/operator-controller/pull/887

internal/labels/labels.go
varshaprasad96: Fixed in https://github.com/operator-framework/operator-controller/pull/887. 

Tldr: This comment is to highlight that the current label key is not in the usual format with `_` and `/` which we follow for Owner annotations. Shouldn't be an issue at all, since it's just a label format.

api/v1alpha1/clusterextension_types.go
varshaprasad96: Removed TODO in https://github.com/operator-framework/operator-controller/pull/887. If its not needed after fixing unit-test, we will address in #874

internal/controllers/clusterextension_controller.go
varshaprasad96: Fixed in https://github.com/operator-framework/operator-controller/pull/887

internal/controllers/clusterextension_controller.go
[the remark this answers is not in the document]
varshaprasad96: Fixed in https://github.com/operator-framework/operator-controller/pull/887

internal/controllers/clusterextension_controller.go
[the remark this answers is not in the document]
varshaprasad96: Resolving this conversation for now.

varshaprasad96:
The only part remaining to be fixed in this PR are the unit tests - #874 is in progress to handle the same.

.goreleaser.yml
```diff
@@ -16,37 +16,50 @@ builds:
...
+      - ppc64le
+      - s390x
 dockers:
 - image_templates:
   - "{{ .Env.IMAGE_REPO }}:{{ .Env.IMAGE_TAG }}-amd64"
   dockerfile: Dockerfile
   goos: linux
   goarch: amd64
-  use: buildx
+  use: docker
```
everettraven, line 38: Why are we switching from `buildx` to `docker` here?

everettraven: For reference, goreleaser's documentation states that there are 2 ways to do multi-platform images, using `ko` or `buildx`: https://goreleaser.com/cookbooks/multi-platform-docker-images/

IIRC we set the use of `buildx` so we could publish multi-platform images by creating these separate platform manifests and publishing them under a single tag

joelanford: I _think_ this was done in order to get `alias docker=podman` to make it possible to run `make release` with just podman installed and have it work. But if that breaks multi-arch builds, I agree that we have to change it back.

Bryce, here's your original PR that made multi-arch work: https://github.com/operator-framework/operator-controller/pull/186

tmshort:
@everettraven @m1kola @acornett21 PTAL
We've rebased the PR and updated the unit tests, and will be going forward with this. Other significant changes to be done have been noted in various issues.

.goreleaser.yml
bentito: It was done to allow for _local_ alias of podman and I think it got inadvertently committed. Good catch. Reverted it in 2402bb67

.goreleaser.yml
```diff
@@ -30,23 +43,23 @@ dockers:
   dockerfile: Dockerfile
   goos: linux
   goarch: arm64
-  use: buildx
+  use: docker
```
everettraven, line 46: I think @bentito mentioned this was supposed to be reverted

internal/controllers/clusterextension_controller_test.go
```diff
@@ -63,7 +64,7 @@ func TestClusterExtensionNonExistentPackage(t *testing.T) {
 	t.Log("By running reconcile")
 	res, err := reconciler.Reconcile(ctx, ctrl.Request{NamespacedName: extKey})
 	require.Equal(t, ctrl.Result{}, res)
-	require.EqualError(t, err, fmt.Sprintf("no package %q found", pkgName))
+	require.EqualError(t, err, fmt.Errorf("no package \"%v\" found", pkgName).Error())
```
everettraven, line 67: Nit: Why was `fmt.Sprintf()` insufficient here?

internal/controllers/clusterextension_controller.go
```diff
@@ -557,6 +676,137 @@ func clusterExtensionRequestsForCatalog(c client.Reader, logger logr.Logger) han
...
+	}
+	if desiredRelease.Manifest != currentRelease.Manifest ||
+		currentRelease.Info.Status == release.StatusFailed ||
+		currentRelease.Info.Status == release.StatusSuperseded {
+		return currentRelease, stateNeedsUpgrade, nil
+	}
+	return currentRelease, stateUnchanged, nil
+}
+
+var GetInstalledbundle = func(ctx context.Context, acg helmclient.ActionClientGetter, allBundles []*catalogmetadata.Bundle, ext *ocv1alpha1.ClusterExtension) (*catalogmetadata.Bundle, error) {
```
everettraven, line 712: Why did this go from a method to a global variable (package scoped)? Doing this opens us up to bugs if this variable is somehow overwritten. I would highly recommend avoiding this.

everettraven: Instead of using a global variable, I would recommend creating a new interface like:

```go
type InstalledBundleGetter interface {
  GetInstalledBundle(...) (...)
}
```

and make use of it via a field in the reconciler like:
```go
type Reconciler struct {
  ...
  InstalledBundleGetter InstalledBundleGetter
  ...
}

func (r *Reconciler) reconcile(...) (...) {
  ...
  installedBundle, err := r.InstalledBundleGetter.GetInstalledBundle(...)
  ...
}
```

This has the ability to be mocked for testing while also preventing potential issues with using global variables

internal/controllers/clusterextension_registryv1_validation_test.go
```diff
@@ -99,13 +100,21 @@ func TestClusterExtensionRegistryV1DisallowDependencies(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			defer func() {
 				require.NoError(t, cl.DeleteAllOf(ctx, &ocv1alpha1.ClusterExtension{}))
-				require.NoError(t, cl.DeleteAllOf(ctx, &rukpakv1alpha2.BundleDeployment{}))
 			}()
-
 			fakeCatalogClient := testutil.NewFakeCatalogClient([]*catalogmetadata.Bundle{tt.bundle})
+			mockUnpacker := unpacker.(*MockUnpacker)
+			// Verify if mockUnpacker is correctly set
+			t.Log("MockUnpacker set to:", mockUnpacker)
```
everettraven, line 107: Nit: Is this necessary?

tmshort: @bentito is looking at these few nits, and will push up a small change.

internal/controllers/clusterextension_controller.go
everettraven: Coming back around to this, something like:
```go
type Reconciler struct {
  GetInstalledBundleFunc func(...)(...)
}
```
_might_ be even better due to it being an explicit type that I don't believe would be allowed to be set as `nil`, alleviating the need to do a potential validation check for a nil reference

varshaprasad96: Though not the best design for now, since it's internal and has no chances of external users breaking it - I'd suggest to do it in a follow up. Especially because it'll require a lot more bits of controller to be changed. An even better way would be for us to instead modify the way we have designed unit tests to mock better.

everettraven: And I think ^ would result in less work to change while getting the benefits of no longer using a global variable

everettraven: If you really prefer this being done in a follow up I won't stop this from going in, but this is something that I would think we want fixed very quickly. I don't think this is a good practice, even if it is in an internal package.

.goreleaser.yml
bentito: fully reverted now

internal/controllers/clusterextension_controller_test.go
bentito: reverted to sprintf everywhere now

internal/controllers/clusterextension_registryv1_validation_test.go
bentito: eliminated now

internal/controllers/clusterextension_controller.go
bentito: well I'm part-way through with the swap out to interface, just doing the mocks needed in the tests now. If you are okay with it in a soon to be added PR, can we can get an approve on this PR and merge?

everettraven (approved):

PR #870, opened 2024-05-17
🌱 Helm POC e2e fixes

dtfranz:
Fixes the e2e tests in helm-poc branch.

### DIFF
--- a/internal/catalogmetadata/filter/bundle_predicates.go
+++ b/internal/catalogmetadata/filter/bundle_predicates.go
@@ -60,6 +60,12 @@ func WithBundleImage(bundleImage string) Predicate[catalogmetadata.Bundle] {
 	}
 }
 
+func WithBundleName(bundleName string) Predicate[catalogmetadata.Bundle] {
+	return func(bundle *catalogmetadata.Bundle) bool {
+		return bundle.Name == bundleName
+	}
+}
+
 func LegacySuccessor(installedBundle *catalogmetadata.Bundle) Predicate[catalogmetadata.Bundle] {
 	isSuccessor := func(candidateBundleEntry declcfg.ChannelEntry) bool {
 		if candidateBundleEntry.Replaces == installedBundle.Name {
--- a/internal/controllers/clusterextension_controller.go
+++ b/internal/controllers/clusterextension_controller.go
@@ -204,6 +204,7 @@ func (r *ClusterExtensionReconciler) handleResolutionErrors(ext *ocv1alpha1.Clus
 		setResolvedStatusConditionFailed(&ext.Status.Conditions, errorMessage, ext.Generation)
 		ensureAllConditionsWithReason(ext, "InstallationStatusUnknown", "")
 	}
+	ext.Status.ResolvedBundle = nil
 	return ctrl.Result{}, err
 }
 
@@ -411,6 +412,11 @@ func (r *ClusterExtensionReconciler) resolve(ctx context.Context, ext ocv1alpha1
 	channelName := ext.Spec.Channel
 	versionRange := ext.Spec.Version
 
+	installedBundle, err := r.installedBundle(ctx, allBundles, &ext)
+	if err != nil {
+		return nil, err
+	}
+
 	predicates := []catalogfilter.Predicate[catalogmetadata.Bundle]{
 		catalogfilter.WithPackageName(packageName),
 	}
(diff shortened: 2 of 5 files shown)


internal/catalogmetadata/filter/bundle_predicates.go
```diff
@@ -60,6 +60,22 @@ func WithBundleImage(bundleImage string) Predicate[catalogmetadata.Bundle] {
 	}
 }
 
+func WithBundleName(bundleName string) Predicate[catalogmetadata.Bundle] {
+	return func(bundle *catalogmetadata.Bundle) bool {
+		return bundle.Name == bundleName
+	}
+}
+
+func WithBundleVersion(bundleVersion string) Predicate[catalogmetadata.Bundle] {
```
dtfranz, line 69: This isn't needed at all, I'll remove it and use an existing in semver range predicate that checks for version equality.

internal/controllers/clusterextension_controller.go
```diff
@@ -411,6 +411,11 @@ func (r *ClusterExtensionReconciler) resolve(ctx context.Context, ext ocv1alpha1
 	channelName := ext.Spec.Channel
 	versionRange := ext.Spec.Version
 
+	installedBundle, err := r.installedBundle(ctx, allBundles, &ext)
```
dtfranz, line 414: This func now in large part resembles `main`. Our feature branch was missing the call to `SuccessorsPredicate` which contains the feature gate switching for semver, which was the primary cause of the e2e failures.

internal/controllers/clusterextension_controller.go
```diff
@@ -682,32 +687,43 @@ func clusterExtensionRequestsForCatalog(c client.Reader, logger logr.Logger) crh
...
+	vr, err := mmsemver.NewConstraint(release.Labels[labels.BundleVersionKey])
+	if err != nil {
+		return nil, err
 	}
 
-	existingVersionSemver, err := bsemver.New(existingVersion)
-	if err != nil {
-		return nil, fmt.Errorf("could not determine bundle version for the chart %q: %w", release.Name, err)
+	// TODO (dfranz) Are the PackageName + BundleName + BundleVersion enough to
+	// be 100% sure we have the correct bundle?
```
dtfranz, line 711: Please see `TODO` here.

test/e2e/cluster_extension_install_test.go
```diff
@@ -224,12 +223,12 @@ func TestClusterExtensionBlockInstallNonSuccessorVersion(t *testing.T) {
 		}
 		assert.Equal(ct, ocv1alpha1.ReasonResolutionFailed, cond.Reason)
 		assert.Equal(ct, "error upgrading from currently installed version \"1.0.0\": no package \"prometheus\" matching version \"1.2.0\" found", cond.Message)
-		assert.Empty(ct, clusterExtension.Status.ResolvedBundle)
+		// TODO (dfranz) do we still need to be clearing this status field when this error is reached?
```
dtfranz, line 226: Another `TODO` here. The `ResolvedBundle` status is not being cleared; I'm not sure if it was intentional or if it just got dropped in a rebase. It's easy to add back in, though.

test/e2e/cluster_extension_registryV1_limitations_test.go
```diff
@@ -2,60 +2,36 @@ package e2e
...
+	require.NoError(t, c.Create(ctx, clusterExtension))
+	defer testCleanup(t, catalog, clusterExtension)
+	defer getArtifactsOutput(t)
 
 	require.EventuallyWithT(t, func(ct *assert.CollectT) {
 		assert.NoError(ct, c.Get(ctx, types.NamespacedName{Name: clusterExtension.Name}, clusterExtension))
-
 		cond := apimeta.FindStatusCondition(clusterExtension.Status.Conditions, ocv1alpha1.TypeInstalled)
-		assert.NotNil(ct, cond)
+		if !assert.NotNil(ct, cond) {
```
dtfranz, line 32: All the other e2e tests use this. Without it, the `Eventually` will check the following `asserts` and panic if `cond` is `nil`.

internal/controllers/clusterextension_controller.go
varshaprasad96: I think so, the catalog is clusters scoped - I don't think having same bundle in multiple catalogs makes sense.

varshaprasad96: @m1kola could you please confirm if that's right?

test/e2e/cluster_extension_install_test.go
varshaprasad96: We can clear the `ResolvedBundle` status for now, to make sure there is no functionality being changed.

But there is a nuance here with the upgrade case, where we would probably not want to clear `Installed` during upgrade (though clearing resolved seems right). But this nuance will be handled in #842 - where we can discuss more on status conditions.

varshaprasad96:
@dtfranz Great work!! all e2e are passing 🎉 !

test/e2e/cluster_extension_install_test.go
dtfranz: Added this check back in and clearing `ResolvedBundle` field when we call `handleResolutionErrors`.

internal/controllers/clusterextension_controller.go
varshaprasad96: The code has been modified as before, @dtfranz can we resolve this then?

PR #874, opened 2024-05-20
🌱 Proposed way to unskip some tests

bentito:
# Description

TestClusterExtensionChannelVersionExists mostly restored here.

This is a proposed way to change the other skipped tests as well.

Items skipped by the last conditional are:

- Bundle Deployment related (can likely just be removed);
- Invariant or other conditions on code paths we're avoiding due to a desire to lessen the mocking.

## Reviewer Checklist

- [ ] API Go Documentation


internal/controllers/clusterextension_controller_test.go
```diff
@@ -229,22 +245,27 @@ func TestClusterExtensionChannelVersionExists(t *testing.T) {
...
+	// we're skipping based on same loadError since Status.Condition setting path is skipped and bundle deployment is absent
+	// also invariants will be different
+	if err != nil && err.Error() != utilerrors.NewAggregate([]error{loadError}).Error() {
+		cond = apimeta.FindStatusCondition(clusterExtension.Status.Conditions, ocv1alpha1.TypeInstalled)
+		require.NotNil(t, cond)
+		require.Equal(t, metav1.ConditionUnknown, cond.Status)
+		require.Equal(t, ocv1alpha1.ReasonInstallationStatusUnknown, cond.Reason)
+		require.Equal(t, "bundledeployment status is unknown", cond.Message)
+
+		t.Log("By fetching the bundled deployment")
```
bentito, line 258: This code chunk could be removed I think as with helm-poc all the Bundle Deployment stuff is removed?

internal/controllers/clusterextension_controller_test.go
```diff
@@ -229,22 +245,27 @@ func TestClusterExtensionChannelVersionExists(t *testing.T) {
...
+		bd := &rukpakv1alpha2.BundleDeployment{}
+		require.NoError(t, cl.Get(ctx, types.NamespacedName{Name: extKey.Name}, bd))
+		require.Equal(t, "core-rukpak-io-registry", bd.Spec.ProvisionerClassName)
+		require.Equal(t, installNamespace, bd.Spec.InstallNamespace)
+		require.Equal(t, rukpakv1alpha2.SourceTypeImage, bd.Spec.Source.Type)
+		require.NotNil(t, bd.Spec.Source.Image)
+		require.Equal(t, "quay.io/operatorhubio/prometheus@fake1.0.0", bd.Spec.Source.Image.Ref)
+
+		verifyInvariants(ctx, t, reconciler.Client, clusterExtension)
```
bentito, line 267: By returning an error on `Storage.Load()` intentionally we avoided a lot of mocking, but also code paths that would have set these exactly as expected. Also, the subsequent checking triggered here is kind of brittle.

internal/controllers/clusterextension_controller_test.go
```diff
@@ -229,22 +245,27 @@ func TestClusterExtensionChannelVersionExists(t *testing.T) {
...
-	require.Equal(t, installNamespace, bd.Spec.InstallNamespace)
-	require.Equal(t, rukpakv1alpha2.SourceTypeImage, bd.Spec.Source.Type)
-	require.NotNil(t, bd.Spec.Source.Image)
-	require.Equal(t, "quay.io/operatorhubio/prometheus@fake1.0.0", bd.Spec.Source.Image.Ref)
 
-	verifyInvariants(ctx, t, reconciler.Client, clusterExtension)
+	// we're skipping based on same loadError since Status.Condition setting path is skipped and bundle deployment is absent
+	// also invariants will be different
+	if err != nil && err.Error() != utilerrors.NewAggregate([]error{loadError}).Error() {
+		cond = apimeta.FindStatusCondition(clusterExtension.Status.Conditions, ocv1alpha1.TypeInstalled)
```
bentito, line 252: Do we need to nail down that these conditions have been set?

internal/controllers/clusterextension_controller_test.go
```diff
@@ -229,22 +245,27 @@ func TestClusterExtensionChannelVersionExists(t *testing.T) {
...
+		require.NotNil(t, cond)
+		require.Equal(t, metav1.ConditionUnknown, cond.Status)
+		require.Equal(t, ocv1alpha1.ReasonInstallationStatusUnknown, cond.Reason)
+		require.Equal(t, "bundledeployment status is unknown", cond.Message)
+
+		t.Log("By fetching the bundled deployment")
+		bd := &rukpakv1alpha2.BundleDeployment{}
+		require.NoError(t, cl.Get(ctx, types.NamespacedName{Name: extKey.Name}, bd))
```
tmshort, line 260: I thought we weren't using BundleDeployments anymore, so why should we expect this to work?

internal/controllers/clusterextension_controller_test.go
tmshort: Yes (see my comment below... I didn't see this)

internal/controllers/clusterextension_controller_test.go
```diff
@@ -186,8 +190,17 @@ func TestClusterExtensionBundleDeploymentDoesNotExist(t *testing.T) {
 }
 
 func TestClusterExtensionChannelVersionExists(t *testing.T) {
-	t.Skip("Skip and replace with e2e for BundleDeployment")
 	cl, reconciler := newClientAndReconciler(t)
+	mockUnpacker := unp.(*MockUnpacker)
```
tmshort, line 194: See comment re: `unp` declaration, as this seems like it might be a local variable, but it's not.

internal/controllers/suite_test.go
```diff
@@ -56,14 +103,16 @@ func newClientAndReconciler(t *testing.T) (client.Client, *controllers.ClusterEx
...
 	}
 	return cl, reconciler
 }
 
 var (
 	cfg *rest.Config
 	acg helmclient.ActionClientGetter
-	unp source.Unpacker
+	unp source.Unpacker // Interface, will be initialized as a mock in TestMain
+	sto storage.Storage
```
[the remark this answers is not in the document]
bentito: refactored, good point, ty

internal/controllers/clusterextension_controller_test.go
bentito: refactored now

internal/controllers/clusterextension_controller_test.go
bentito: I removed the BundleDeployment chunk (that was being avoided anyway)

internal/controllers/clusterextension_controller_test.go
bentito: yep, rm'ed now

internal/controllers/clusterextension_controller_test.go
bentito: same comment as below: @joelanford wdyt? there is a lot more mocking to do if we need to reach the code that sets these exactly as the current test(s) want.

everettraven: IMO, we should be able to only modify the behavior associated with the removal of the `BundleDeployment` API and the status condition that should be present on the `ClusterExtension` API. Other than those changes I would expect that, if our unit tests are well written, the tests should remain as is and function appropriately. 

That being said, it does seem like in order to get our controller logic more easily unit testable it would require a fairly significant refactor of the `reconcile()` function in which more abstractions are created to mock out and maybe that is better suited as a follow up done ASAP after the initial PR merges. IIRC consensus from the meeting where we walked through the main PR was that e2e tests are a must and should only be skipped with verifiably good reason and unit tests while not ideal to be skipped may be acceptable to un-skipped in a follow up with the abstractions.

In my head the very high-level flow seems to be:
- Resolution to get Bundle for ClusterExtension
- Unpack Bundle image
- Convert bundle contents to Helm Chart
- Apply Helm Chart

I would expect that there is some abstraction for each of these steps that can be mocked, making our controller much easier to unit test

internal/controllers/clusterextension_controller_test.go
```diff
@@ -214,7 +227,10 @@ func TestClusterExtensionChannelVersionExists(t *testing.T) {
 	t.Log("By running reconcile")
 	res, err := reconciler.Reconcile(ctx, ctrl.Request{NamespacedName: extKey})
 	require.Equal(t, ctrl.Result{}, res)
-	require.NoError(t, err)
+	// Asserting that only loadError occurred
+	if assert.Error(t, err) {
```
varshaprasad96, line 231: What's the value of returning an error aggregate? We can discuss in our sync on the error types.

internal/controllers/clusterextension_controller_test.go
```diff
@@ -229,22 +245,18 @@ func TestClusterExtensionChannelVersionExists(t *testing.T) {
...
-	require.NoError(t, cl.Get(ctx, types.NamespacedName{Name: extKey.Name}, bd))
-	require.Equal(t, "core-rukpak-io-registry", bd.Spec.ProvisionerClassName)
-	require.Equal(t, installNamespace, bd.Spec.InstallNamespace)
-	require.Equal(t, rukpakv1alpha2.SourceTypeImage, bd.Spec.Source.Type)
-	require.NotNil(t, bd.Spec.Source.Image)
-	require.Equal(t, "quay.io/operatorhubio/prometheus@fake1.0.0", bd.Spec.Source.Image.Ref)
+	// we're skipping based on same loadError since Status.Condition setting path is skipped
+	// and invariants will be different
+	if err != nil && err.Error() != utilerrors.NewAggregate([]error{loadError}).Error() {
+		cond = apimeta.FindStatusCondition(clusterExtension.Status.Conditions, ocv1alpha1.TypeInstalled)
```
varshaprasad96, line 252: @bentito I think the whole test that `skipinvariants` is not necessary hereon - I remember chatting to @joelanford about this. Once we are able to remove it, I don't think we need to even mock the storage. The installationStatusUnknown here was because there was no rukpak running under the hood, and the firs reconcile would always return an unknown status because unpack was still in progress. 

So the test here should be to verify only till the state is pending. The objective of this unit test ended to ensure that resolve was successful and a bundle deployment was created with the installation status unknown. With our changes this translates to - Resolve status set to the right value and unpack status to be pending - which is what the first reconcile is suppose to return.

varshaprasad96: This should solve all the unit follow up unit tests. Especially because these tests never ran rukpak previously - which means in current scenario all we need to test is till the unpack state is set to pending successfully.

internal/controllers/clusterextension_registryv1_validation_test.go
```diff
@@ -112,7 +112,7 @@ func TestClusterExtensionRegistryV1DisallowDependencies(t *testing.T) {
 			reconciler := &controllers.ClusterExtensionReconciler{
 				Client:             cl,
 				BundleProvider:     &fakeCatalogClient,
-				ActionClientGetter: acg,
+				ActionClientGetter: helmClientGetter,
```
varshaprasad96, line 115: We probably don't need this anymore, because we are not reaching the install stage. We are just exiting after unpacked. I can try running locally and update here.

internal/controllers/suite_test.go
```diff
@@ -74,22 +123,20 @@ func TestMain(m *testing.M) {
...
 
 	rm := meta.NewDefaultRESTMapper(nil)
-	cfgGetter, err := helmclient.NewActionConfigGetter(cfg, rm)
+	cfgGetter, err := helmclient.NewActionConfigGetter(config, rm)
 	utilruntime.Must(err)
-	acg, err = helmclient.NewActionClientGetter(cfgGetter)
+	helmClientGetter, err = helmclient.NewActionClientGetter(cfgGetter)
 	utilruntime.Must(err)
 
-	mgr, err := manager.New(cfg, manager.Options{})
```
varshaprasad96, line 89: Why remove manager? This seems weird. 

Also most of the diff here is in the naming - for ex (cfg to config, act to helmClientGetter). Its fine, but in order to keep the diffs low, can we revert them as before?

varshaprasad96:
@bentito: https://github.com/operator-framework/operator-controller/commit/d649164d840340ae475ba434caf4af2fb33ac801 - I tried making changes to your branch, but GH doesn't let me create a PR against your repo :swear With this change, we don't need to mock the loader itself. We use the mock unpacker to return `UnpackPending` and test relevant conditions which is equivalent on what the test did previously.

internal/controllers/clusterextension_controller_test.go
varshaprasad96: https://github.com/operator-framework/operator-controller/commit/d649164d840340ae475ba434caf4af2fb33ac801 - is what I meant. Doing this in the rest of the tests should solve all the cases. We shouldn't be needing to test loader or storer itself as that is something which these tests didn't do either - as they never ran rukpak.

internal/controllers/clusterextension_controller_test.go
```diff
@@ -214,7 +221,7 @@ func TestClusterExtensionChannelVersionExists(t *testing.T) {
 	t.Log("By running reconcile")
 	res, err := reconciler.Reconcile(ctx, ctrl.Request{NamespacedName: extKey})
 	require.Equal(t, ctrl.Result{}, res)
-	require.NoError(t, err)
+	require.NoError(t, err, nil)
```
joelanford, line 224: What does this change do?

internal/controllers/clusterextension_controller_test.go
```diff
@@ -224,27 +231,18 @@ func TestClusterExtensionChannelVersionExists(t *testing.T) {
...
+	require.Equal(t, ocv1alpha1.ReasonSuccess, resolvedCond.Reason)
+	require.Equal(t, "resolved to \"quay.io/operatorhubio/prometheus@fake1.0.0\"", resolvedCond.Message)
+
+	t.Log("By checking the expected unpacked conditions")
+	unpackedCond := apimeta.FindStatusCondition(clusterExtension.Status.Conditions, rukpakv1alpha2.TypeUnpacked)
+	require.NotNil(t, unpackedCond)
+	require.Equal(t, metav1.ConditionFalse, unpackedCond.Status)
+	require.Equal(t, rukpakv1alpha2.ReasonUnpackPending, unpackedCond.Reason)
 
-	verifyInvariants(ctx, t, reconciler.Client, clusterExtension)
```
joelanford, line 247: Curious why we're losing this piece of the test?

internal/controllers/clusterextension_registryv1_validation_test.go
joelanford: I was just about to comment about this in the main PR.

Not now, but in a quick follow-up, we should definitely add unit tests that cover cases where `Reconcile` is called after a bundle is successfully unpacked. In those cases, we will progress to installation and need an actionClientGetter, right?

internal/controllers/suite_test.go
```diff
@@ -17,31 +17,78 @@ limitations under the License.
...
 	"github.com/operator-framework/rukpak/pkg/source"
-	"github.com/operator-framework/rukpak/pkg/util"
+	"github.com/operator-framework/rukpak/pkg/storage"
 
 	"github.com/operator-framework/operator-controller/internal/controllers"
 	"github.com/operator-framework/operator-controller/pkg/scheme"
 	testutil "github.com/operator-framework/operator-controller/test/util"
 )
 
+// MockUnpacker is a mock of Unpacker interface
```
joelanford, line 46: Nit: move the mocks to a separate package and/or file to make them easier to find?

internal/controllers/suite_test.go
joelanford: Do we actually use the manager anywhere? I wouldn't expect to need a manager in a unit test.

internal/controllers/clusterextension_controller_test.go
joelanford: There's quite a bit of code path beyond `storage.Load` that we will need to cover pretty quickly. I don't think it needs to be covered _immediately_ in the initial PR that merges, but I think we'll want to beef up coverage for cases where `Load` returns non-errors.

internal/controllers/clusterextension_controller_test.go
joelanford: +1 to Bryce's comment. And then we would also want unit tests for each of the actual real implementations of those abstractions.

But also, the verifyInvariants test is there to make sure that no matter when we return from `reconcile` we have formed an opinion on every Condition type that we support. It's okay if we don't progress all the way through all the code paths, but this sounds like a good example of this test doing its job. It means that the paths that we _do_ test don't actually make sure that all of the conditions are initialized.

internal/controllers/clusterextension_controller_test.go
everettraven: Looks like `verifyInvariants` was removed from a few of the tests. I wasn't aware of what that function does so I went exploring. After learning what it's used for I agree that it should remain because IIUC we should be setting all of our status conditions on every reconcile.

After taking a look at the rest of the PR though, adding `verifyInvariants` back would be my only requested change (with the assumption that we currently expect some tests to be failing and will address them in another way).

internal/controllers/clusterextension_controller_test.go
bentito: not sure either, reverted.

internal/controllers/clusterextension_controller_test.go
varshaprasad96: We have way more condition types and reasons now - so setting all of the conditions whenever the reconciler exists is an issue by itself. We are working on fixing this and brining `verifyInvariants` back for now. But we would strongly urge for us to evaluate the whole design of what conditions we want. https://github.com/operator-framework/operator-controller/pull/842 deals with some of it, but have also created an issue to track it.

test/e2e/cluster_extension_install_test.go
```diff
@@ -207,6 +207,7 @@ func TestClusterExtensionBlockInstallNonSuccessorVersion(t *testing.T) {
 		assert.Equal(ct, ocv1alpha1.ReasonSuccess, cond.Reason)
 		assert.Contains(ct, cond.Message, "resolved to")
 		assert.Equal(ct, &ocv1alpha1.BundleMetadata{Name: "prometheus-operator.1.0.0", Version: "1.0.0"}, clusterExtension.Status.ResolvedBundle)
+		assert.Equal(ct, &ocv1alpha1.BundleMetadata{Name: "prometheus-operator.1.0.0", Version: "1.0.0"}, clusterExtension.Status.InstalledBundle)
```
dtfranz, line 210: If we don't actually check that this version has been installed, we may run into a scenario where we've resolved to `1.0.0` but haven't installed it yet. Then, if the user immediately changes the `ClusterExtension` to `1.2.0`, we're essentially installing `1.2.0` from scratch, and the non-successor limitation doesn't apply in that case. I don't personally think this is a bug.

This check is not necessary on `main` because when we finish `resolve` the `BundleDeployment` is immediately created with the installed version flags present, regardless of whether or not the `1.0.0` operator has finished installing yet. I think this is actually less desirable compared to the above scenario; for instance if the user quickly changed their mind about which version they wanted they wouldn't be able to install `1.2.0` even though `1.0.0` hasn't even installed yet.

varshaprasad96 (approved):
/lgtm

internal/controllers/clusterextension_registryv1_validation_test.go
varshaprasad96: https://github.com/operator-framework/operator-controller/issues/879

internal/controllers/clusterextension_controller_test.go
varshaprasad96: > verifyInvariants test is there to make sure that no matter when we return from reconcile we have formed an opinion on every Condition type that we support.

This is a mess given we have 8 condition types. This is fixed for now, but there is a follow up created to re-evaluate the condition types. Resolving this conversation for now.

internal/controllers/common_controller.go
```diff
@@ -95,3 +120,46 @@ func setDeprecationStatusesUnknown(conditions *[]metav1.Condition, message strin
...
+		Message: err.Error(),
+	})
+	return err
+}
+
+// TODO: verify if we need to update the installBundle status or leave it as is.
+func updateStatusUnpackPending(status *ocv1alpha1.ClusterExtensionStatus, result *source.Result, generation int64) {
+	status.InstalledBundle = nil
+	apimeta.SetStatusCondition(&status.Conditions, metav1.Condition{
+		Type:               rukpakv1alpha2.TypeUnpacked,
```
varshaprasad96, line 139: Same for the other two unpack status conditions below. 
```suggestion
		Type:               ocv1alpha1.TypeUnpacked,
```

tmshort: This is a status condition, didn't we say that another PR will deal with that stuff? (Although the above suggestion is correct.)

internal/controllers/clusterextension_controller_test.go
bentito: `verifyInvarients` is restored here and everywhere now.

internal/controllers/clusterextension_controller_test.go
bentito: agree. also `verifyInvariants` is restored here.

internal/controllers/suite_test.go
tmshort: Added as #895.

internal/controllers/common_controller.go
tmshort: #896

PR #878, opened 2024-05-21
:sparkles: Use error type rather than strings

tmshort:
# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/controllers/clusterextension_controller.go
+++ b/internal/controllers/clusterextension_controller.go
@@ -38,14 +38,12 @@ import (
 	"helm.sh/helm/v3/pkg/storage/driver"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
-	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
 	apimachyaml "k8s.io/apimachinery/pkg/util/yaml"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -79,6 +77,17 @@ import (
 	"github.com/operator-framework/operator-controller/internal/handler"
 	"github.com/operator-framework/operator-controller/internal/labels"
 )
+
+// ResolutionError type
+// If a more specific error type needs to be distinguished,
+// add another type here
+type ResolutionError struct {
+	message string
+}
+
+func (e ResolutionError) Error() string {
+	return e.message
+}
 
 // ClusterExtensionReconciler reconciles a ClusterExtension object
 type ClusterExtensionReconciler struct {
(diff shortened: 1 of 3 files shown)


internal/controllers/clusterextension_controller.go
```diff
@@ -184,11 +215,11 @@ func (r *ClusterExtensionReconciler) handleResolutionErrors(ext *ocv1alpha1.Clus
 	if errors.As(err, &aggErrs) {
 		for _, err := range aggErrs.Errors() {
 			errorMessage := err.Error()
-			if strings.Contains(errorMessage, "no package") {
+			if _, ok := err.(NoPackageError); ok {
```
everettraven, line 218: As the linting error says attached to this line, I think we should be using `errors.As()` just like was done above on line 215 for detecting it as the aggregated error type

tmshort: Hmmm... this shouldn't be an aggregated error type, but you are correct.

everettraven: Sorry, I was referring to how on line 215 we are using `errors.As()` to ensure the error is an `Aggregated` error. Not that this particular line should be checking for the same type.

tmshort: I knew what you meant. :)

internal/controllers/clusterextension_controller.go
```diff
@@ -184,11 +213,11 @@ func (r *ClusterExtensionReconciler) handleResolutionErrors(ext *ocv1alpha1.Clus
 	if errors.As(err, &aggErrs) {
 		for _, err := range aggErrs.Errors() {
 			errorMessage := err.Error()
-			if strings.Contains(errorMessage, "no package") {
+			if errors.As(err, &NoPackageError{}) {
```
everettraven, line 216: Are you able to verify that this works as expected? IIRC, the last time I did something like this I discovered it did not work as expected and had to do it this way: https://github.com/operator-framework/catalogd/blob/fa4a29bae42a5725d932b1bf6c31381f185e99f7/pkg/controllers/core/clustercatalog_controller.go#L79-L82

joelanford: Semi-off topic, but @everettraven, that recoverable error thing is a feature built directly into controller-runtime since 0.15: https://github.com/kubernetes-sigs/controller-runtime/blob/735b6073bb253c0449bfcf6641855dcf2118bb15/pkg/reconcile/reconcile.go#L105-L109

tmshort:
@m1kola (github doesn't support comment threads), I do believe we need to distinguish between Resolution errors, which likely involve errors in the `ClusterExtension` resource, vs. other errors that can occur on the cluster. In the first case, the user may need to fix it, and can likely do it themselves. In the other case, it may be a temporal error on the cluster, or something else that the user cannot handle.

m1kola:
> @m1kola (github doesn't support comment threads), I do believe we need to distinguish between Resolution errors, which likely involve errors in the `ClusterExtension` resource, vs. other errors that can occur on the cluster. In the first case, the user may need to fix it, and can likely do it themselves. In the other case, it may be a temporal error on the cluster, or something else that the user cannot handle.

@tmshort `handleResolutionErrors` only handles resolution errors (only errors returned from `r.resolve(ctx, *ext)`).

How I see it:
* If something is wrong with `ClusterExtension` spec (e.g. we can't find a package) - we want to make sure that the error is very human readable and appears on the condition. Human can take an action based on the error message.
* If it is something temporal (e.g. we timeout fetching bundles from catalogs) - then we should also put errors into the condition. Ideally adding some context for troubleshooting (e.g. not just "request timed out", but "error fetching bundles: request timed out", etc). Eventually temporal errors will be resolved during one of the attempts to reconcile.
* If it is something permanent - same as temporal. I believe at the moment we do not have any permanent/terminal conditions. So every error which appears to be "permanent" is logically just a temporal issue which has not been resolved yet.


In all the these cases we can fail resolution condition and write error into the condition message. I don't think that setting reason to `InstallationStatusUnknown` aids UX.

There is a chance that I'm missing something (e.g. maybe you expect programmatic clients to take advantage of a distinct combination of reasons and take some action?). Happy to discuss it further here or jump on a quick call if you like.

tmshort:
@m1kola 

> `handleResolutionErrors` only handles resolution errors (only errors returned from `r.resolve(ctx, *ext)`).

Non-resolution (e.g. timeout) errors can occur in `r.resolve()`, and thus any kind of errors are handled by `handleResolutionErrors`. Would it make more sense to remove the function (since it's a lot smaller now), and put the functionality into the main function? 

This is just trying to clean up error processing. It's not meant to clean up the status conditions (but it has to do _something_ with status conditions). It looks as though #842 changed title, so it's doing something different now, issue #880 is meant to clean up the status conditions. The hope is that we can have smaller changes (which is why I'm trying to keep this small).

tmshort:
This PR is focused on removing the string checks for error types, and all the aggregated errors that were being used. It was _not_ meant to fix all the problems with status conditions.

m1kola (approved):
Love removing code. Looks good to me.

PR #886, opened 2024-05-22
:seedling: Helm poc remove bd get artifacts

dtfranz:
When tests fail, gatherArtifacts is run to collect cluster state. This PR removes gathering of BundleDeployments since we don't use them anymore and it's generating additional misleading error messages.

This is minor and can be merged after `helm-poc` branch to reduce diffs in that PR.


varshaprasad96 (approved):
/lgtm
/approve

PR #888, opened 2024-05-23
⚠ add TLS overlay for Catalogd v0.13.0 web server TLS

trgeiger:
# Description
Updates the Catalogd dependency to v0.14.0 which uses TLS for the web server and olmv1-system for the namespace
- shifts manifests structure to use an overlay, similar to catalogd and rukpak
- installs Catalogd to the operator-controller-system namespace so operator-controller can access its tls.crt
- adds a flag to the manager specifying the ca cert to use for TLS authentication
- change default namespace to olmv1-system

## Reviewer Checklist

- [ ] API Go Documentation


cmd/manager/main.go
```diff
@@ -80,9 +82,11 @@ func main() {
 		systemNamespace             string
 		unpackImage                 string
 		provisionerStorageDirectory string
+		tlsCert                     string
 	)
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
+	flag.StringVar(&tlsCert, "tls-cert", "", "The TLS certificate to use for verifying HTTPS connections to the Catalogd web server.")
```
joelanford, line 89: It would be better to name this `--ca-cert` or similar to convey its intent better.

`--tls-cert` would give me the impression that this is used as a server cert.

cmd/manager/main.go
```diff
@@ -152,8 +156,27 @@ func main() {
 		os.Exit(1)
 	}
 
+	httpClient := &http.Client{Timeout: 10 * time.Second}
+
+	if tlsCert != "" {
+		cert, err := os.ReadFile(tlsCert)
+		if err != nil {
+			log.Fatalf("Failed to read certificate file: %v", err)
+		}
```
joelanford, line 165: We probably need to anticipate rotation of this cert as well.

cmd/manager/main.go
```diff
@@ -152,8 +156,27 @@ func main() {
...
+		caCertPool.AppendCertsFromPEM(cert)
+		tlsConfig := &tls.Config{
+			RootCAs:    caCertPool,
+			MinVersion: tls.VersionTLS12,
+		}
+		tlsTransport := &http.Transport{
+			TLSClientConfig: tlsConfig,
+		}
+		httpClient.Transport = tlsTransport
+	}
```
joelanford, line 176: Could this chunk of code be carved out into a helper function that builds an `httpClient`?

config/base/manager/manager.yaml
```diff
@@ -114,6 +114,6 @@ spec:
       terminationGracePeriodSeconds: 10
       volumes:
         - name: cache
-          emptyDir: {}
+          emptyDir: {} 
         - name: bundle-cache
-          emptyDir: {}
+          emptyDir: {}
```
joelanford, line 119: Extraneous whitespace changes?

config/base/default/kustomization.yaml
joelanford, line 1: Do we still need the `default` kustomization?

config/overlays/tls/kustomization.yaml
```diff
@@ -0,0 +1,27 @@
...
+# field above.
+namePrefix: operator-controller-
+
+# the following config is for teaching kustomize how to do var substitution
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- ../../base/crd
+- ../../base/rbac
+- ../../base/manager
```
joelanford, line 17: If we are able to drop the `./config/base/default` kustomization, we could add a `./config/base` kustomization and then here just use `../../base`.

config/overlays/tls/patches/manager_namespace_label.yaml
```diff
@@ -0,0 +1,3 @@
+- op: add
+  path: /metadata/labels/trust
+  value: "enabled"
```
joelanford, line 3: Is this a trust-manager leftover?

config/overlays/tls/patches/manager_deployment_cert.yaml
```diff
@@ -0,0 +1,9 @@
+- op: add
+  path: /spec/template/spec/volumes/-
+  value: {"name":"ca-certificate", "secret":{"secretName":"catalogd-catalogserver-cert", "optional": false, "items": [{"key": "tls.crt", "path": "tls.crt"}]}}
```
joelanford, line 3: Do you know if the file contents in the volume mount change if the secret contents change?

config/overlays/tls/patches/manager_namespace_label.yaml
trgeiger: yep, thanks

config/overlays/tls/patches/manager_deployment_cert.yaml
trgeiger: I just verified that the contents of the mounted file /var/certs/tls.crt on the filesystem do change after manually rotating the certificate.

cmd/manager/main.go
trgeiger: I've been trying to figure out a solution for this, but I'm hitting a wall. It looks like this is an issue others have encountered before and there's not a native mechanism for reloading rootCAs in go: https://github.com/golang/go/issues/35887#issuecomment-1798581909

joelanford: Oh that looks like a minefield. I suppose for now we can punt on handling CA rotation, and revisit if it becomes a practical issue. Maybe just make a separate issue for this?

config/overlays/tls/patches/manager_deployment_cert.yaml
joelanford: That's good. That's step 1 of being able to rotate the CA cert. But it seems moot for now since there's no easy in-process way to rotate them with an existing server/listener.

This is a good detail to capture in the "let's handle CA cert rotation" issue though.

config/base/e2e/kustomization.yaml
joelanford, line 1: I think I would put `e2e` in the overlays section since it is not actually part of the base. WDYT?

trgeiger: Yeah that makes sense to me

cmd/manager/main.go
trgeiger: That sounds good to me, might be a good contribution to make to Go at some point

trgeiger:
Barring any other feedback, the only remaining issue is deciding how we want to handle installing Catalogd to the operator-controller namespace for the purposes of Tilt. We could add some custom logic to one of the tiltfiles, but we could also change Catalogd's default namespace to just be operator-controller-system if we don't have any compelling reasons to keep it in its own namespace. @everettraven would appreciate your thoughts when you're back in the office.

cmd/manager/main.go
```diff
@@ -80,9 +79,11 @@ func main() {
 		systemNamespace             string
 		unpackImage                 string
 		provisionerStorageDirectory string
+		caCert                      string
 	)
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
+	flag.StringVar(&caCert, "ca-cert", "", "The TLS certificate to use for verifying HTTPS connections to the Catalogd web server.")
```
m1kola, line 86: Wondering if we should make it more generic: not only about catalogd. E.g. what if we want to pull bundles from a registry with a self-signed certs?

See https://github.com/operator-framework/operator-controller/pull/905#discussion_r1625934766 for example.

cc @varshaprasad96

scripts/install.tpl.sh
```diff
@@ -35,8 +35,8 @@ function kubectl_wait() {
 kubectl apply -f "https://github.com/cert-manager/cert-manager/releases/download/${cert_mgr_version}/cert-manager.yaml"
 kubectl_wait "cert-manager" "deployment/cert-manager-webhook" "60s"
 
-kubectl apply -f "https://github.com/operator-framework/catalogd/releases/download/${catalogd_version}/catalogd.yaml"
-kubectl_wait "catalogd-system" "deployment/catalogd-controller-manager" "60s"
+curl -L https://github.com/operator-framework/catalogd/releases/download/${catalogd_version}/catalogd.yaml | sed s/catalogd-system/operator-controller-system/g | kubectl apply -f -
```
everettraven, line 38: I think this is fine temporarily, but if we are going to use a common namespace I'd like to see if there is a way we could use something like kustomize to have a more declarative and consistent override than using `sed`.

I wonder if we could make use of the remote targets functionality in Kustomize to help us achieve this and it be a bit more robust (I think we could make sure it _only_ changes the namespace): https://github.com/kubernetes-sigs/kustomize/blob/master/examples/remoteBuild.md

trgeiger: Good point. However, if we do decide to make all OLMv1 components just install into the same namespace by default, then we wouldn't need to do this anymore.

everettraven (dismissed):

everettraven (approved):

PR #905, opened 2024-06-03
:warning: Bump rukpak

varshaprasad96:
# Description

## Reviewer Checklist

- [ ] API Go Documentation


Dockerfile
```diff
@@ -14,7 +14,6 @@ COPY --from=builder /busybox/cp /cp
 WORKDIR /
 
 COPY manager manager
-COPY unpack unpack
```
dtfranz, line 17: You should also be able to remove the staged build components and copying of `cp` binary above. Also be sure to `rm -rf` the cmd/unpack :smiley:

Makefile
```diff
@@ -220,7 +227,7 @@ export GO_BUILD_FLAGS :=
 export GO_BUILD_LDFLAGS := -s -w \
     -X '$(VERSION_PATH).version=$(VERSION)' \
 
-BINARIES=manager unpack
+BINARIES=manager
 
 $(BINARIES):
 	go build $(GO_BUILD_FLAGS) -tags '$(GO_BUILD_TAGS)' -ldflags '$(GO_BUILD_LDFLAGS)' -gcflags '$(GO_BUILD_GCFLAGS)' -asmflags '$(GO_BUILD_ASMFLAGS)' -o $(BUILDBIN)/$@ ./cmd/$@
```
dtfranz, line 233: We can probably just go back to:
```suggestion
BUILDCMD = go build $(GO_BUILD_FLAGS) -tags '$(GO_BUILD_TAGS)' -ldflags '$(GO_BUILD_LDFLAGS)' -gcflags '$(GO_BUILD_GCFLAGS)' -asmflags '$(GO_BUILD_ASMFLAGS)' -o $(BUILDBIN)/manager ./cmd/manager
```
And then below:
```
go-build-local:
	$(BUILDCMD)
```
and:
```
go-build-linux:
	GOOS=linux $(BUILDCMD)
```
since we don't have more than one binary to build anymore.

testdata/bundles/registry-v1/package-with-webhooks.v1.0.0/Dockerfile
```diff
@@ -12,4 +12,4 @@ LABEL operators.operatorframework.io.metrics.project_layout=unknown
 
 # Copy files to locations specified by labels.
 COPY manifests /manifests/
-COPY metadata /metadata/
+COPY metadata /metadata/
```
bentito, line 15: nit: I think we lost a trialing CR here

internal/controllers/clusterextension_controller.go
```diff
@@ -580,7 +586,8 @@ func (r *ClusterExtensionReconciler) generateBundleDeploymentForUnpack(bundlePat
 			Source: rukpakv1alpha2.BundleSource{
 				Type: rukpakv1alpha2.SourceTypeImage,
 				Image: &rukpakv1alpha2.ImageSource{
-					Ref: bundlePath,
+					Ref:                   bundlePath,
+					InsecureSkipTLSVerify: true,
```
bentito, line 590: Do we really want to hard code skipping this? Maybe this should come in from a flag as an option?

Dockerfile
tmshort: Do we need to, if we don't build it?

internal/controllers/clusterextension_controller.go
tmshort: Perhaps. Right now, it's necessary., unless we get CAs.

testdata/catalogs/test-catalog/catalog.yaml
```diff
@@ -74,7 +74,7 @@ entries:
 schema: olm.bundle
 name: package-with-webhooks.1.0.0
 package: package-with-webhooks
-image: localhost/testdata/bundles/registry-v1/package-with-webhooks:v1.0.0
+image: docker-registry.operator-controller-e2e.svc.cluster.local:5000/bundles/registry-v1/prometheus-with-webhooks.v1.0.0
```
dtfranz, line 77: Should this be a `:` instead of a `.` in front of version?
```suggestion
image: docker-registry.operator-controller-e2e.svc.cluster.local:5000/bundles/registry-v1/prometheus-with-webhooks:v1.0.0
```

internal/controllers/clusterextension_controller.go
```diff
@@ -390,12 +390,18 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
...
+					source.Kind(
+						r.cache, unstructuredObj, crhandler.TypedEnqueueRequestForOwner[*unstructured.Unstructured](
+							r.Scheme(),
+							r.RESTMapper(),
+							ext,
+							crhandler.OnlyControllerOwner(),
+						),
+						helmpredicate.DependentPredicateFuncs[*unstructured.Unstructured](),
+					),
+				); err != nil {
```
tmshort, line 402: I have a slightly different version of this in #912:
https://github.com/operator-framework/operator-controller/pull/912/files#diff-a54f79b762054ccdc31e4011d28cce22af8f54d43e1fddc75745ce042f177f7e

tmshort: To reduce the noise in this PR, I'm going to suggest the k8s upgrade happen in #912. But, there's a bit of different code in how this section is fixed up, and I'm not sure which one is better.

varshaprasad96: The difference in #912 is that you're setting the watches on objects directly, whereas this one is first modifying to an unstructured type. One benefit of using unstructured is being able to work with dynamic clients. I don't know the exact reason on why it was done this way, but the change to unstructured preserves the previous functionality.

joelanford: I wrote this originally I think. My memory is fuzzy, but I suspect unstructured was used because we need to support arbitrary objects. If a structured type is used, it needs to be added to the scheme. But we can't know how to add arbitrary types to a scheme. Hence, the use of unstructured.

This code has been around for like 5 years though, so circumstances and assumptions may have changed.

@tmshort I think the best way to test your change would be to craft a bundle that has an arbitrary CR in it, install that bundle, and then see if this dynamic watch setup is successful.

varshaprasad96: During install, any resource present in the bundle has the owner refs set to the cluster extension.

tmshort: It appears I can't attempt to use this version, because it requires an update to`helmpredicate`, which is in rukpak.

tmshort: I would prefer we keep the `main` version of this code, tweaked as necessary.

cmd/manager/main.go
```diff
@@ -89,7 +90,7 @@ func main() {
 	flag.StringVar(&cachePath, "cache-path", "/var/cache", "The local directory path used for filesystem based caching")
 	flag.BoolVar(&operatorControllerVersion, "version", false, "Prints operator-controller version information")
 	flag.StringVar(&systemNamespace, "system-namespace", "", "Configures the namespace that gets used to deploy system resources.")
-	flag.StringVar(&unpackImage, "unpack-image", defaultUnpackImage, "Configures the container image that gets used to unpack Bundle contents.")
+	flag.StringVar(&unpackCacheDir, "unpack-cache-dir", "/var/cache/unpack", "Configures the directory that gets used to unpack and cache Bundle contents.")
```
joelanford, line 93: Could this just be a hardcoded sub-directory of `cachePath`?

tmshort: Possibly

internal/controllers/clusterextension_controller.go
joelanford: What fails without this?

tmshort: Looks to have been used for debug and is not necessary.

go.mod
```diff
@@ -11,10 +11,9 @@ require (
 	github.com/google/go-cmp v0.6.0
 	github.com/operator-framework/api v0.25.0
 	github.com/operator-framework/catalogd v0.12.0
-	github.com/operator-framework/helm-operator-plugins v0.2.1
+	github.com/operator-framework/helm-operator-plugins v0.2.2-0.20240520180534-f463c36fedf9
 	github.com/operator-framework/operator-registry v1.43.1
-	github.com/operator-framework/rukpak v0.20.1-0.20240510194410-99faf1c1763f
-	github.com/spf13/cobra v1.8.0
+	github.com/operator-framework/rukpak v0.21.2-0.20240606055133-ac1802e9c1e8
```
m1kola, line 16: Should we cut a new release for this? Our latest release is v0.21.1 and https://github.com/operator-framework/rukpak/commit/ac1802e9c1e8 is not in it.

Makefile
```diff
@@ -122,8 +122,11 @@ e2e: #EXHELP Run the e2e tests.
 
 E2E_REGISTRY_NAME := docker-registry
 E2E_REGISTRY_NAMESPACE := operator-controller-e2e
+DNS_NAME           := $(E2E_REGISTRY_NAME).$(E2E_REGISTRY_NAMESPACE).svc.cluster.local
```
m1kola, line 125:
```suggestion
DNS_NAME := $(E2E_REGISTRY_NAME).$(E2E_REGISTRY_NAMESPACE).svc.cluster.local
```

Makefile
```diff
@@ -122,8 +122,11 @@ e2e: #EXHELP Run the e2e tests.
...
+
 export REG_PKG_NAME := registry-operator
 export CATALOG_IMG := $(E2E_REGISTRY_NAME).$(E2E_REGISTRY_NAMESPACE).svc:5000/test-catalog:e2e
+export REGISTRY_ROOT := $(E2E_REGISTRY_NAME).$(E2E_REGISTRY_NAMESPACE).svc:5000
```
m1kola, line 129: Is there a reason why we need both `DNS_NAME` and `REGISTRY_ROOT`?

internal/controllers/clusterextension_controller.go
```diff
@@ -257,10 +257,10 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
 			setStatusUnpackFailed(ext, err.Error())
 			return ctrl.Result{}, err
 		}
-		setStatusUnpacked(ext, unpackResult.Message)
+		setStatusUnpacked(ext, fmt.Sprintf("unpack successful: %v", unpackResult.Message))
 	default:
-		setStatusUnpackFailed(ext, err.Error())
-		return ctrl.Result{}, err
+		setStatusUnpackFailed(ext, fmt.Sprintf("unpack successful: %v", unpackResult.Message))
```
m1kola, line 262: Seems like "unpack successful" is not meant to be here

internal/controllers/clusterextension_controller.go
m1kola: Unresolving as this needs attention.

m1kola:
Also a question about ⚠️. Why is it a breaking change? Because we change condition messages?

testdata/catalogs/test-catalog/catalog.yaml
```diff
@@ -74,7 +74,7 @@ entries:
 schema: olm.bundle
 name: package-with-webhooks.1.0.0
 package: package-with-webhooks
-image: localhost/testdata/bundles/registry-v1/package-with-webhooks:v1.0.0
+image: docker-registry.operator-controller-e2e.svc.cluster.local:5000/bundles/registry-v1/prometheus-with-webhooks:v1.0.0
```
m1kola, line 77: Why did `package-with-webhooks` get renamed to `prometheus-with-webhooks` here?

The package itself in catalog still called `package-with-webhooks`. And we are building this image from `testdata/bundles/registry-v1/package-with-webhooks.v1.0.0`.

internal/controllers/clusterextension_controller.go
varshaprasad96: This should be a knob, and we should expose it an API, as it very much depends on where catalog is present, which in-turn depends on where v1 is running. 

If my understanding is correct - there is a draft PR which I have opened to discuss the change where we introduce this in a separate `ClusterExtensionSource`. 

cc: @tmshort @joelanford

internal/controllers/clusterextension_controller.go
varshaprasad96: Ah! Good catch!

internal/controllers/clusterextension_controller.go
m1kola: Is it required for rukpak bump? I was able to successfully run e2e in this PR without this change.

varshaprasad96: It does fail e2es in CI

m1kola: That's weird. What is the difference between `make test-e2e` in CI and locally?

varshaprasad96: But I think there may be a work around for this while setting up the remote registry... looking into it

Makefile
```diff
@@ -199,6 +202,12 @@ kind-load-test-artifacts: $(KIND) #EXHELP Load the e2e testdata container images
 	$(CONTAINER_RUNTIME) build testdata/bundles/registry-v1/package-with-webhooks.v1.0.0 -t  localhost/testdata/bundles/registry-v1/package-with-webhooks:v1.0.0
 	$(KIND) load docker-image localhost/testdata/bundles/registry-v1/package-with-webhooks:v1.0.0 --name $(KIND_CLUSTER_NAME)
 
+registry-load-bundles: ## Load selected e2e testdata container images created in kind-load-bundles into registry
```
m1kola, line 205: Given this - do we still need `kind-load-test-artifacts`? It seems like we can get rid of it.

Makefile
tmshort: `CATALOG_IMG` is used in both the `Makefile` and the `setup.sh` script for the dev-ext-e2e, where as `REGISTRY_ROOT` is only needed by `setup.sh`. I suppose one can be defined by the other.

internal/controllers/clusterextension_controller.go
tmshort: I was wondering the same, but I believe it might negatively impact the tests... but we should evaluate.

tmshort:
>Also a question about ⚠️. Why is it a breaking change? Because we change condition messages?

The PR guidance says:

>    * ⚠ (:warning:, major/breaking change)

Since rukpak's API updates were incompatible, I consider this a major (but mot necessarily a breaking) change.

Makefile
tmshort: You might be right.

internal/controllers/clusterextension_controller.go
m1kola: @varshaprasad96 makes sense. I figured out why I got a "pass" locally: because my test results got cached locally after a successful run with `InsecureSkipTLSVerify` set to `true`:

```
PASS
ok  	github.com/operator-framework/operator-controller/test/e2e	(cached)
```

Anyway - I feel a bit uneasy about changing the code to make tests work. Especially when it comes to TLS. You mentioned that there is a potentially a work around/fix for this. If we can fix that on test setup level - that would be great.

tmshort:
>Unresolving as this needs attention.

Unfortunately, there is a behavior difference between local e2e/dev-ext-e2e, and the CI. The CI seems to need `InsecureSkipTLSVerify: true`, and I'm not sure why.

go.mod
m1kola: As far as I see `main` currently points at https://github.com/operator-framework/rukpak/commit/ac1802e9c1e8. So it is as recent as it can be as of right now :)

testdata/bundles/registry-v1/build-push-e2e-bundle.sh
```diff
@@ -0,0 +1,81 @@
...
+tgz="${bundle_dir}/manifests.tgz"
+tar czf "${tgz}" -C "${bundle_dir}/" manifests metadata
+kubectl create configmap -n "${namespace}" --from-file="${tgz}" operator-controller-${bundle_name}.manifests
+rm "${tgz}"
+
+kubectl apply -f - << EOF
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: "kaniko-${bundle_name}"
```
m1kola, line 43: Because we have dots in bundle names, this results in:

```
Warning: metadata.name: this is used in Pod names and hostnames, which can result in surprising behavior; a DNS label is recommended: [must not contain dots]
```

It is just a warning, but a bit annoying.

Makefile
tmshort: ..._and_ it's used inside the ext-dev-e2e golang files...

m1kola: Are you talking about `CATALOG_IMG`?

I'm asking about `DNS_NAME` and `REGISTRY_ROOT` which seem to be virtually the same thing and both are just introduced in this PR.

Differences:
* `DNS_NAME` for some reason uses `cluster.local` DNS when `REGISTRY_ROOT` uses `svc`
* `DNS_NAME` does not include port, but all the usages of `DNS_NAME` (see below in `Makefile`) then add the port.

testdata/bundles/registry-v1/build-push-e2e-bundle.sh
tmshort: Fixed.

Makefile
tmshort: Ah... yeah, that can be cleaned up.

go.mod
tmshort: Ok, I did update that. ;)

internal/controllers/clusterextension_controller.go
varshaprasad96: Based on the discussion on slack, adding an annotation on CE to allow users to set `insecureSkipTLSVerify`. Will create a follow up, for an API design to include this in the spec.

Makefile
m1kola: I think this is now resolved.

go.mod
m1kola: Just to clarify what I was getting at: it looks like we are currently pinning rukpak to a specific commit (probably because of `go get github.com/operator-framework/rukpak@latest`).

Can we switch back to using tags as with catalogd, operator-registry, etc?

We used to use tags (see 6060534838ef1450bbe26c71d671a4c942fd80f3, for example), but it looks like with helm integration we merged non-release dependency.

internal/controllers/clusterextension_controller.go
m1kola: I didn't manage to find the discussion on slack you are referring to. Could you please share a link?

> Looking into the setup for spinning up the local registry, we do not configure any certificates so this seems expected.

For unrelated reason I was looking at the registry. It looks like we actually create certs for the registry with cert-manager, but they are self-signed and operator-controller knows nothing about their CA so it can't verify. We might want to look into #888 and see if we can make it more generic so that operator-controller is configurable with CAs.

m1kola (commented):
There are still some comment threads which are open. Apart from that - looks good to me.

internal/controllers/clusterextension_controller.go
varshaprasad96: @m1kola  This is the thread: https://redhat-internal.slack.com/archives/C06KP34REFJ/p1717787814237379

tl; dr; Yes, the set-up does involve creating certs using cert-manager (I missed looking at it previously). It involves creating certs and having them in secrets (https://github.com/operator-framework/operator-controller/blob/5687a69155e3c6e279ea0c304c9d83e5c11369c1/test/tools/image-registry.sh#L82-L89) which is then specified in the registry pod. However, the operator-controller pod currently does not have a volume to mount those secrets and all the deployments in this case are on different namespaces making it difficult to share. On exploring various options to go about this, the fix for now was to introduce an API change allowing users to pass an option to specify tls configuration. This is a little complicated to do in this PR, as we would have to discuss how the API would look like if we add it to the spec- to make it compatible with future options (like direct bundle install). Hence we decided to keep this PR moving and to have a super minimalistic change to allow users skip tls verification through an annotation.

m1kola: Thanks for the link and context.

Just to clarify - when I mentioned #888 and looking into how we can make operator-controller be aware of CAs, etc - I did not mean that we need to do it in this PR.

It is more about the follow up work - as an option.

go.mod
tmshort: We would need a tag; there is no tag at the tip of rukpak. I asked for one in the channel.

internal/controllers/clusterextension_controller.go
tmshort: We discussed adding a method to get CAs into operator-controller, and they are all _too big_ to be done here. It will require some additional thought, and it requires modifying the tests to do things in a proper order.
Specifically, we need to have any cert-manager secrets created _before_ operator controller, and right now, operator-controller is created before any of the e2e infrastructure, which includes the docker-registry and corresponding Secret.
So, we are adding an annotation to make this configurable on a per-ClusterExtension basis, and will subsequently update either the CE API, the operator-controller pod spec, or both.

go.mod
tmshort: There's now `v0.22.0`

testdata/catalogs/test-catalog/catalog.yaml
tmshort: Fixed

tmshort:
Made a couple (small) commits to _hopefully_ address the last of the comments... @m1kola ?

internal/controllers/clusterextension_controller.go
varshaprasad96: Yes. I was about to push the change. It should be:
```
	default:
		setStatusUnpackFailed(ext, "unexpected unpack status")
		// We previously exit with a failed status if error is not nil.
		return ctrl.Result{}, fmt.Errorf("unexpected unpack status: %v", unpackResult.Message)
```

Ran this fix locally and e2es work. Will push the change after a rebase.

varshaprasad96: Addressed by https://github.com/operator-framework/operator-controller/pull/905/commits/6bc072fa5a44f2884c753b4809c0bb16aceb1f6c

varshaprasad96:
@m1kola looks like all the comments have been addressed. Could you please provide a lgtm if all is well.

everettraven (approved):

tmshort:
```
GitHub Actions has encountered an internal error when running your job.

m1kola (approved):

Issue #737, opened 2024-04-04
[epic] ClusterExtension uses service account provided in spec to manage content


joelanford:
> Follow least privilege principle and reduce confused deputy problems by limiting the scope of OLM v1’s permissions. Instead, have users provide a ServiceAccount with the necessary permissions to manage an extension.

For more information on this feature, please see the following documents:
- [brief](https://docs.google.com/document/d/1hYgVF8aj5QPewV-WbGchsBX4BBH2G7te0qF_z7vuK-Q/)
- [rfc](https://docs.google.com/document/d/1LcjggSjrad12i4F9reeCW6tawPCbuAEww0yWcoUlsD4/edit?usp=sharing)

## Tasks
- [x] #971
- [x] #972
- [x] #973
- [x] #974
- [x] #975
- [x] #976
- [x] #977

joelanford:
I've done some hacking today to get support into helm-operator-plugins and rukpak for specifying and using a service account https://github.com/operator-framework/rukpak/pull/857

I know we're looking to eliminate the separate rukpak controller's but perhaps these changes are useful in some way if we reuse or vendor helm-operator-plugins and/or rukpak code.

varshaprasad96:
Related https://github.com/operator-framework/operator-controller/issues/840

joelanford:
By the way, I think this particular feature is going to be a fairly complex change that will require some upfront design.

@skattoju if you want to pick this up, can we do a kickoff meeting to discuss the high level goals and talk through some of the implications? And I think we'll want to put together a design doc for this one once we figure out what the plan is.

The issue that @varshaprasad96 linked has a bunch of the context.

skattoju:
Definitely! I'll set something up 👍

skattoju:
i have started a PoC i am still testing to see if it works..
https://github.com/skattoju/operator-controller/tree/sa_from_spec_poc

joelanford:
@everettraven just checking to see if this epic is still in correct state ("Needs Docs")? I see a doc task in the description that is closed and there are no open issues. Can we close this out?

everettraven:
@joelanford looks like the PR for that docs issue just merged today, should be good to close it out!

PR #907, opened 2024-06-04
🌱 Use SA from spec when installing bundles

skattoju:
# Description
Adds a ServiceAccountName field to the spec and sets up the action client getter to provide an action client that uses a token based on the service account provided in the spec. Related to #737 

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/api/v1alpha1/clusterextension_types.go
+++ b/api/v1alpha1/clusterextension_types.go
@@ -78,6 +78,12 @@ type ClusterExtensionSpec struct {
 	// the bundle may contain resources that are cluster-scoped or that are
 	// installed in a different namespace. This namespace is expected to exist.
 	InstallNamespace string `json:"installNamespace"`
+
+	//+kubebuilder:validation:Pattern:=^[a-z0-9]([-a-z0-9.]*[a-z0-9])?$
+	//+kubebuilder:validation:MaxLength:=253
+	// ServiceAccountName is the name of the ServiceAccount to use to manage the resources in the bundle.
+	// The service account is expected to exist in the InstallNamespace.
+	ServiceAccountName string `json:"serviceAccountName"`
 }
 
 const (
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -28,10 +28,16 @@ import (
 	"go.uber.org/zap/zapcore"
 	k8slabels "k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/selection"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
+	"k8s.io/client-go/rest"
 	ctrl "sigs.k8s.io/controller-runtime"
 	crcache "sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
 	crfinalizer "sigs.k8s.io/controller-runtime/pkg/finalizer"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
@@ -44,6 +50,7 @@ import (
 	"github.com/operator-framework/rukpak/pkg/storage"
 
 	ocv1alpha1 "github.com/operator-framework/operator-controller/api/v1alpha1"
+	"github.com/operator-framework/operator-controller/internal/authentication"
 	"github.com/operator-framework/operator-controller/internal/catalogmetadata/cache"
 	catalogclient "github.com/operator-framework/operator-controller/internal/catalogmetadata/client"
 	"github.com/operator-framework/operator-controller/internal/controllers"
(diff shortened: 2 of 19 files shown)


varshaprasad96:
/ok-to-test 

(not sure if this label is accepted)

internal/controllers/clusterextension_admission_test.go
```diff
@@ -222,11 +226,13 @@ func TestClusterExtensionAdmissionInstallNamespace(t *testing.T) {
 			t.Parallel()
 			cl := newClient(t)
 			err := cl.Create(context.Background(), buildClusterExtension(ocv1alpha1.ClusterExtensionSpec{
-				PackageName:      "package",
-				InstallNamespace: tc.installNamespace,
+				PackageName:        "package",
+				InstallNamespace:   tc.installNamespace,
+				ServiceAccountName: tc.serviceAccountName,
```
joelanford, line 231: This test is just for the `spec.installNamespace` validation. Should we hardcode `ServiceAccountName` to something we know will be valid?

That would avoid maintainer/contributor confusion. Folks might see that the service account name is varied as part of this test and conclude that the service account name somehow plays into the installNamespace validatation.

We also want to test side-effects from varying service account names.

internal/controllers/clusterextension_admission_test.go
```diff
@@ -222,11 +226,13 @@ func TestClusterExtensionAdmissionInstallNamespace(t *testing.T) {
...
 			}))
 			if tc.errMsg == "" {
 				require.NoError(t, err, "unexpected error for installNamespace %q: %w", tc.installNamespace, err)
+				require.NoError(t, err, "unexpected error for serviceAccountName %q: %w", tc.serviceAccountName, err)
```
joelanford, line 235: I think we should put the serviceAccount validations into an isolated test.

testdata/rbac/test-rbac.yaml
```diff
@@ -0,0 +1,29 @@
...
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  namespace: default
+  name: test-clusterrole
+rules:
+- apiGroups: ["*"]
+  resources: ["*"]
+  verbs: ["*"]
```
joelanford, line 15: This is exactly where I would start (and I know this is still WIP), but before we merge, we should update these rules and potentially create separate sets of rules/serviceaccounts for each test bundle.

That way, we can actually prove to ourselves the what a minimum rule set is for a particular bundle. And we'll want to set a good example and _not_ give service accounts cluster admin.

cmd/manager/main.go
```diff
@@ -155,14 +160,44 @@ func main() {
...
+			return nil, err
+		}
+		cfg.BearerToken = token
+		return cfg, nil
+	}
+
+	cfgGetter, err := helmclient.NewActionConfigGetter(mgr.GetConfig(), mgr.GetRESTMapper(),
+		helmclient.ClientNamespaceMapper(nsMapper),
+		helmclient.StorageNamespaceMapper(nsMapper),
+		helmclient.RestConfigMapper(rcm),
```
joelanford, line 195: With this setup, we should also remove the `*/*/*` kubebuilder rbac marker in the clusterextension_controller.go file.

api/v1alpha1/clusterextension_types.go
```diff
@@ -78,6 +78,12 @@ type ClusterExtensionSpec struct {
 	// the bundle may contain resources that are cluster-scoped or that are
 	// installed in a different namespace. This namespace is expected to exist.
 	InstallNamespace string `json:"installNamespace"`
+
+	//+kubebuilder:validation:Pattern:=^[a-z0-9]([-a-z0-9.]*[a-z0-9])?$
+	//+kubebuilder:validation:MaxLength:=253
+	// ServiceAccountName is the name of the ServiceAccount to use to manage the resources in the bundle.
+	// The service account is expected to exist in the InstallNamespace.
+	ServiceAccountName string `json:"serviceAccountName"`
```
everettraven, line 86: Optional Nit: Using https://pkg.go.dev/k8s.io/api/core/v1#LocalObjectReference could be nice here as it is meant for local object references like we are doing here.

testdata/rbac/test-rbac.yaml
varshaprasad96: Not related to this work at all: but thinking about it, to make a better UX it would be nice to have some tooling around generating rbac for bundle contents (could be complicated!)

internal/controllers/clusterextension_admission_test.go
skattoju: agreed

internal/controllers/clusterextension_admission_test.go
skattoju: makes sense 👍

cmd/manager/main.go
skattoju: we would still need list and watch on all objects right ?
 `//+kubebuilder:rbac:groups=*,resources=*,verbs=list;watch`

testdata/rbac/test-rbac.yaml
skattoju: Yeah seems like it will be a lot of trial and error to create the RBAC. Extracting the RBAC from the csv could be a starting point.

cmd/manager/main.go
varshaprasad96: The initial reason for using such a broad RBAC was the uncertainty regarding the objects the bundle would contain. Consequently, the controller was granted admin permissions to monitor and perform CRUD operations on all GVKs.

However, in this scenario, the user (specifically the cluster admin) provides a SA that the controller uses to install contents. It's not necessary for the SA to have admin permissions. Instead, the SA can be granted restricted permissions, limited to only the GVKs present in the bundle. It's expected that the entity providing this SA is aware of the specific permissions needed to install and manage the bundle.

This way we minimise the scope of the SA's access, ensuring it only has the necessary permissions to interact with the relevant GRs.

api/v1alpha1/clusterextension_types.go
skattoju: not sure what is meant here, is there an example you can point me to ?

cmd/manager/main.go
skattoju: True for installing the bundles the SA provided in the spec will be used but the controller also establishes watches on the bundle content and this is done with the controller service account so it needs to be able to watch arbitrary objects.

everettraven:
@skattoju I'm going to close this draft PR since we are going to split the implementation into individual PRs.

Thank you a ton for all the effort you put into this PoC that helped drive the discussion and design process!

PR #909, opened 2024-06-04
:warning: Update to go 1.22

tmshort:
Fix issues with the extension-developer-e2e and operator-sdk

# Description

## Reviewer Checklist

- [ ] API Go Documentation


test/extension-developer-e2e/setup.sh
```diff
@@ -75,11 +75,14 @@ reg_pkg_name="${REG_PKG_NAME}"
 (
   cd "${REG_DIR}" && \
   $operator_sdk init --domain="${DOMAIN}" && \
+  sed -i -e 's/CONTROLLER_TOOLS_VERSION ?= v0.14.0/CONTROLLER_TOOLS_VERSION ?= v0.15.0/' Makefile && \
```
joelanford, line 78: Delete this line? Also while we're at it, could you add some comments about this `sed` command and the `export OPERATOR_SDK` line?

Something like these:
-  `# controller-gen v0.13.0 (scaffolded by operator-sdk) panics when run with go 1.22, so pin to a more recent version. NOTE this is a rough edge that users will experience`
- `# The Makefile in the project scaffolded by operator-sdk uses an SDK binary in the path if it is present. Override to ensure we use the same version that we scaffolded with. NOTE: this is a rough edge that users will experience`

joelanford: (sorry, didn't see/think of this when we we're debugging it)

varshaprasad96:
@tmshort the last commit in [this](https://github.com/operator-framework/operator-controller/pull/905/commits/1f9c68ec7a801421cb1ef8aeb4169ab9ea6ed81e) PR was trying to do the same, without bumping go for now. 

We are bumping controller-gen for generating oc manifests here, but keeping controller-runtime and other related deps like helm-operator plugins the same. Its better to keep them tied together, so we can either bump it all in this or do an immediate follow up for the same?

.bingo/Variables.mk
```diff
@@ -23,11 +23,11 @@ $(BINGO): $(BINGO_DIR)/bingo.mod
 	@echo "(re)installing $(GOBIN)/bingo-v0.8.0"
 	@cd $(BINGO_DIR) && GOWORK=off $(GO) build -mod=mod -modfile=bingo.mod -o=$(GOBIN)/bingo-v0.8.0 "github.com/bwplotka/bingo"
 
-CONTROLLER_GEN := $(GOBIN)/controller-gen-v0.14.0
+CONTROLLER_GEN := $(GOBIN)/controller-gen-v0.15.0
```
varshaprasad96, line 26: We are using an updated controller gen to generate and test manifests for operator controller, but keeping controller-runtime tied to a previous version. We can either bump controller-gen for o-c in a follow up with others, or do it all here. I would suggest the former, to keep this PR confined to fixing operator sdk with go 1.22 bump.

tmshort: Bumping controller-runtime starts a cascade effect of including k8s v0.30.0, so, yes, I'd prefer to do it later.

tmshort:
Bumping things like operator-controller, etc, creates an avalanche of other changes. To get things completed and done faster (i.e. reviews), it's easier to do smaller, incremental changes. We'll get there, it's just doing it in one big chunk is probably not a good idea.

tmshort:
> @tmshort the last commit in [this](https://github.com/operator-framework/operator-controller/pull/905/commits/1f9c68ec7a801421cb1ef8aeb4169ab9ea6ed81e) PR was trying to do the same, without bumping go for now.

We're still having trouble updating rukpak... that change is quite big, and should be done "as it's own thing". If we can make that PR simpler by doing some of that work here, so much the better.

.bingo/Variables.mk
varshaprasad96: Sounds good, then we don't need the changes in `.bingo/`. The only change would for controller-gen would be in the set up script that gets used for scaffolding the operator using SDK.

tmshort:
The _executable_ for controller-gen has been updated consistently, so we're not running different versions of the same tool.

test/extension-developer-e2e/setup.sh
tmshort: I'll make sure it still works without this...

.bingo/Variables.mk
tmshort: The changes in bingo are necessary to install the the correct version of controller-gen, and to install a version of bingo that is compatible with go 1.22.

varshaprasad96 (approved):
/lgtm
/approve

PR #916, opened 2024-06-08
🐛 Fix: installed bundle provider no longer requires catalog

joelanford:
The installed bundle provider, which is necessary for safely handling package upgrades, was too strict. It required the currently installed bundle to exist in the catalog in order to "find" the installed bundle.

This is problematic in several  situations:
- If a user receives a catalog update that no longer contains their installed bundle (perhaps its was removed from the catalog for security or policy reasons)
- If a user is trying to transition to a different catalog that provides that package
- If a bundle was directly installed, and the user is trying to have a catalog-backed ClusterExtension adopt it.

This change simply returns the name and version of the installed bundle, and makes some minor changes in the successors functions to handle the simplified installed bundle metadata.

# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/catalogmetadata/filter/bundle_predicates.go
+++ b/internal/catalogmetadata/filter/bundle_predicates.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/operator-framework/operator-registry/alpha/declcfg"
 
+	ocv1alpha1 "github.com/operator-framework/operator-controller/api/v1alpha1"
 	"github.com/operator-framework/operator-controller/internal/catalogmetadata"
 )
 
@@ -66,7 +67,7 @@ func WithBundleName(bundleName string) Predicate[catalogmetadata.Bundle] {
 	}
 }
 
-func LegacySuccessor(installedBundle *catalogmetadata.Bundle) Predicate[catalogmetadata.Bundle] {
+func LegacySuccessor(installedBundle *ocv1alpha1.BundleMetadata) Predicate[catalogmetadata.Bundle] {
 	isSuccessor := func(candidateBundleEntry declcfg.ChannelEntry) bool {
 		if candidateBundleEntry.Replaces == installedBundle.Name {
 			return true
@@ -77,9 +78,9 @@ func LegacySuccessor(installedBundle *catalogmetadata.Bundle) Predicate[catalogm
 			}
 		}
 		if candidateBundleEntry.SkipRange != "" {
-			installedBundleVersion, _ := installedBundle.Version()
-			skipRange, _ := bsemver.ParseRange(candidateBundleEntry.SkipRange)
-			if installedBundleVersion != nil && skipRange != nil && skipRange(*installedBundleVersion) {
+			installedBundleVersion, vErr := bsemver.Parse(installedBundle.Version)
+			skipRange, srErr := bsemver.ParseRange(candidateBundleEntry.SkipRange)
+			if vErr == nil && srErr == nil && skipRange(installedBundleVersion) {
 				return true
 			}
 		}
--- a/internal/controllers/clusterextension_controller.go
+++ b/internal/controllers/clusterextension_controller.go
@@ -94,7 +94,7 @@ type ClusterExtensionReconciler struct {
 }
 
 type InstalledBundleGetter interface {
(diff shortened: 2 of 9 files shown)


trgeiger:
Sounds like a great change, nicely simplifies some sections of code. LGTM after a quick look, but can look more closely later

internal/catalogmetadata/filter/bundle_predicates.go
```diff
@@ -77,9 +78,9 @@ func LegacySuccessor(installedBundle *catalogmetadata.Bundle) Predicate[catalogm
 			}
 		}
 		if candidateBundleEntry.SkipRange != "" {
-			installedBundleVersion, _ := installedBundle.Version()
-			skipRange, _ := bsemver.ParseRange(candidateBundleEntry.SkipRange)
-			if installedBundleVersion != nil && skipRange != nil && skipRange(*installedBundleVersion) {
+			installedBundleVersion, vErr := bsemver.Parse(installedBundle.Version)
+			skipRange, srErr := bsemver.ParseRange(candidateBundleEntry.SkipRange)
+			if vErr == nil && srErr == nil && skipRange(installedBundleVersion) {
```
joelanford, line 83: Both the before and after code just silently ignore edges if the installed bundle version or the skipRange cannot be parsed.

I'm tempted to actually fail here if the bundle version or skip range fail to parse.

Thoughts?

internal/controllers/clusterextension_controller.go
```diff
@@ -392,7 +392,7 @@ func (r *ClusterExtensionReconciler) resolve(ctx context.Context, ext ocv1alpha1
 	}
 
 	if ext.Spec.UpgradeConstraintPolicy != ocv1alpha1.UpgradeConstraintPolicyIgnore && installedBundle != nil {
-		upgradePredicate, err := SuccessorsPredicate(installedBundle)
+		upgradePredicate, err := SuccessorsPredicate(ext.Spec.PackageName, installedBundle)
```
joelanford, line 395: Could be covered separately/later, but I wonder if the successors predicate should actually care about the package name?

I could envision a separate predicate that gets to decide which package names are acceptable to be upgraded from, where the default would be what it is now (i.e. "spec.packageName").

Something like that may make it easier for us if we wanted to help users do an upgrade between bundles that come from a different package.

joelanford: But perhaps `upgradeConstraintPolicy: Ignore` is enough of an escape hatch for that situation....

internal/catalogmetadata/filter/bundle_predicates.go
everettraven: What is the impact to non-semver versions if we fail here? Do we know how many packages use versioning schemes incompatible with semver in the commonly used catalogs?

internal/controllers/clusterextension_controller.go
everettraven: +1 on covering later. I think there is some additional nuance here of deciding the ordering of bundles between different packages. How would you decide between a newer version of a bundle in the same package versus a different one?

I think for now the current escape hatch should work?

everettraven (approved):
Changes look good to me!

internal/catalogmetadata/filter/bundle_predicates.go
joelanford: Pretty sure FBC requires valid semver and `opm validate` checks both places. So in theory, this should never fail, and if it does, it is because someone has invalid FBC.

Which is why I think I'd rather just fail here. If we ignore it, it's this sorta weird fence-straddling position of accepting it, but then not doing anything with it.

everettraven: If there is validation that the versions must be valid semver then I am fine with failing here

joelanford: Looks like it would be a significant change to bubble errors up through the predicate API. Probably better to handle this in a follow-up. I think we should do one of the following:
1. Update the predicate API to return `(bool, error)`. This is a big change because _lots_ of stuff implements the predicate interface.
2. Update the bundle-related predicates to use a types where the bundle version and channel skipRanges are pre-parsed. That way we eliminate the calls inside the predicate that could return an error.

PR #924, opened 2024-06-12
⚠️ Only allow bundles with `AllNamespaces` install mode

### LINKED ISSUE #816: Tests to confirm that registry+v1 bundles with with install mode other than `AllNamespaces` are not allowed
The [relevant check](https://github.com/operator-framework/rukpak/blob/882714a00b5d99fdfc1b65992206c4e90edfd696/internal/convert/registryv1.go#L197-L199) is currently in rukpak, but I'm planning to add a test in operator-controller. This will help to ensure that we do not break the checks in the process of moving rukpak functionality into operator controller.

m1kola:
# Description

TBD

Closes #816

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/handler/registry.go
+++ b/internal/handler/registry.go
@@ -10,6 +10,7 @@ import (
 
 	"helm.sh/helm/v3/pkg/chart"
 	"helm.sh/helm/v3/pkg/chartutil"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/yaml"
 
@@ -24,7 +25,7 @@ const (
 )
 
 func HandleClusterExtension(_ context.Context, fsys fs.FS, ext *ocv1alpha1.ClusterExtension) (*chart.Chart, chartutil.Values, error) {
-	plainFS, err := convert.RegistryV1ToPlain(fsys, ext.Spec.InstallNamespace, nil)
+	plainFS, err := convert.RegistryV1ToPlain(fsys, ext.Spec.InstallNamespace, []string{metav1.NamespaceAll})
 	if err != nil {
 		return nil, nil, fmt.Errorf("convert registry+v1 bundle to plain+v0 bundle: %v", err)
 	}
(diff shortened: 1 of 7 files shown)


test/e2e/cluster_extension_registryV1_limitations_test.go
```diff
@@ -37,3 +37,29 @@ func TestClusterExtensionPackagesWithWebhooksAreNotAllowed(t *testing.T) {
 		assert.Equal(ct, &ocv1alpha1.BundleMetadata{Name: "package-with-webhooks.1.0.0", Version: "1.0.0"}, clusterExtension.Status.ResolvedBundle)
 	}, pollDuration, pollInterval)
 }
+
+func TestClusterExtensionPackagesWithUnsupportedInstallModesAreNotAllowed(t *testing.T) {
```
joelanford, line 41: Is there a way to put this in a unit test instead?

m1kola: @joelanford I started working on this (and another test above `TestClusterExtensionPackagesWithWebhooksAreNotAllowed`) when rukpak was a separate thing running in the cluster with unpacker and all that complicated setup.

The idea was that we have an e2e test which helps us verify current limitations. I was hoping that this test will also help with the helm/rukpak integration into operator-controller. But it got blocked by helm/rukpak integration :)

I do not think that we can test it on `ClusterExtension ` reconciler: rukpak stuff is mocked there.

We can probably test it on [this level](https://github.com/operator-framework/operator-controller/blob/66a063b37f9faa1dae3ff17ad808410023a77f26/internal/handler/registry.go#L26-L40) with a fake FS, but at this point I'm not sure how useful it will be.

Now that we use rukpak as a library - we should probably treat it as such: assume that this stuff if well tested on the rukpak side and do not double test on operator controller.

What do you think?

joelanford: > Now that we use rukpak as a library - we should probably treat it as such: assume that this stuff if well tested on the rukpak side and do not double test on operator controller.

I think with @varshaprasad96 's change in #928, where the actual call to convert the bundle happens fully in rukpak, I agree.

internal/handler/registry.go
```diff
@@ -24,7 +25,7 @@ const (
 )
 
 func HandleClusterExtension(_ context.Context, fsys fs.FS, ext *ocv1alpha1.ClusterExtension) (*chart.Chart, chartutil.Values, error) {
-	plainFS, err := convert.RegistryV1ToPlain(fsys, ext.Spec.InstallNamespace, nil)
+	plainFS, err := convert.RegistryV1ToPlain(fsys, ext.Spec.InstallNamespace, []string{metav1.NamespaceAll})
```
m1kola, line 28: It looks like we no longer needs this fix because we no longer have it in operator-controller since #928

PR #928, opened 2024-06-12
✨  Cleanups after bumping Rukpak

varshaprasad96:
This PR makes the following changes:
1. Removes the use of a separate handler, given rukpak has now externalized its implementation.
2. Removes the unpacking status, which is no longer required with the use of direct registry client.

# Description

## Reviewer Checklist

- [ ] API Go Documentation


api/v1alpha1/clusterextension_types.go
```diff
@@ -110,7 +110,6 @@ const (
 	ReasonUnpackPending = "UnpackPending"
 	ReasonUnpackSuccess = "UnpackSuccess"
 	ReasonUnpackFailed  = "UnpackFailed"
-	ReasonUnpacking     = "Unpacking"
```
varshaprasad96, line 113: https://github.com/operator-framework/operator-controller/issues/929 - explanation on its removal.

joelanford (dismissed):

varshaprasad96:
Looks like https://github.com/operator-framework/operator-controller/pull/916 is causing merge conflicts when added to the queue. Will rebase this after #916 gets merged.

internal/controllers/clusterextension_controller.go
```diff
@@ -274,7 +269,7 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
 		return ctrl.Result{}, err
 	}
 
-	chrt, values, err := r.Handler.Handle(ctx, bundleFS, ext)
+	chrt, values, err := r.Handler.Handle(ctx, bundleFS, bd)
```
bentito, line 272: I assume `bd` for bundleDeployment, but why change and why just this one?

internal/controllers/common_controller.go
```diff
@@ -152,18 +152,6 @@ func setStatusUnpackPending(ext *ocv1alpha1.ClusterExtension, message string) {
...
-	apimeta.SetStatusCondition(&ext.Status.Conditions, metav1.Condition{
-		Type:               ocv1alpha1.TypeUnpacked,
-		Status:             metav1.ConditionFalse,
-		Reason:             ocv1alpha1.ReasonUnpacking,
-		Message:            message,
-		ObservedGeneration: ext.GetGeneration(),
-	})
-}
-
 func setStatusUnpacked(ext *ocv1alpha1.ClusterExtension, message string) {
```
bentito, line 155: I think the others, `setStatusUnpack[Pending|Failed]` can go away too

internal/controllers/clusterextension_controller.go
varshaprasad96: This change was made so that we can use the existing handler code in Rukpak and remove the duplicate implementation which we had here (the PR deletes files under handler/). This also makes sure that all the conversion logic from registryV1 to plain is in one place - which is well tested in rukpak, and we have less chances of erroring out.

api/v1alpha1/clusterextension_types.go
```diff
@@ -110,7 +110,6 @@ const (
 	ReasonUnpackPending = "UnpackPending"
 	ReasonUnpackSuccess = "UnpackSuccess"
 	ReasonUnpackFailed  = "UnpackFailed"
```
bentito, line 112: I think `ReasonUnpack[Pending|Failed|Success]` can go away too?

internal/controllers/common_controller.go
varshaprasad96: Technically yes, `unpackPending` can go away (we still need failed when it returns an error). But there is a caveat - which I have explained in detail in here: https://github.com/operator-framework/operator-controller/issues/929. It would break our tests, and we would have to continue implementing mock loader and storer. I wanted to keep that as a separate task as it needs considerable effort and not comprise on the tests we have currently - which is why have removed the `unpacking` state which is not being used anywhere.

api/v1alpha1/clusterextension_types.go
```diff
@@ -142,7 +141,6 @@ func init() {
 		ReasonHasValidBundleUnknown,
 		ReasonUnpackPending,
 		ReasonUnpackSuccess,
```
bentito, line 143: here too: I think `ReasonUnpack[Pending|Failed|Success]` can go away too?

api/v1alpha1/clusterextension_types.go
varshaprasad96: Success and Failed are still needed, so that we set the right status when we unpack [successfully](https://github.com/operator-framework/operator-controller/blob/3fa711f9ac986a0269cf500c62b8b6a495076070/internal/controllers/clusterextension_controller.go#L259) and [error](https://github.com/operator-framework/operator-controller/blob/3fa711f9ac986a0269cf500c62b8b6a495076070/internal/controllers/clusterextension_controller.go#L239) out while unpacking.

api/v1alpha1/clusterextension_types.go
varshaprasad96: Explained in https://github.com/operator-framework/operator-controller/pull/928#discussion_r1638192288. Success and Failed are still needed. Pending can be removed, but then there is an issue with the tests which is explained in #929

api/v1alpha1/clusterextension_types.go
bentito: 👍

joelanford (approved):

internal/controllers/clusterextension_controller.go
bentito: ah I get it, thanks!

api/v1alpha1/clusterextension_types.go
bentito: 👍

api/v1alpha1/clusterextension_types.go
bentito: 👍

internal/controllers/common_controller.go
bentito: yep seems fine for a followup on `unpackPending` 👍

everettraven (approved):
Changes look good to me and seems reasonable to have #929 as a follow up.

PR #937, opened 2024-06-14
🐛 Fix rbac generation in `Makefile`

m1kola:
# Description

It looks like #888 we moved rbac manifests, but did not update controller-gen output path. This PR should fix this.

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,7 +51,7 @@ CONTAINER_RUNTIME := docker
 else ifneq (, $(shell command -v podman 2>/dev/null))
 CONTAINER_RUNTIME := podman
 else
-$(warning Could not find docker or podman in path! This may result in targets requiring a container runtime failing!) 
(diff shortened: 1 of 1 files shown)


trgeiger:
Thanks for catching this, LGTM

grokspawn:
should we also .gitignore config/rbac since it is generated?

m1kola:
@grokspawn nothing should write into `config/rbac` anymore. Not sure that ignoring it makes a lot of sense.

varshaprasad96 (approved):
Agree with @m1kola. We aren't adding other generated files to gitignore, and should keep it the same for rbac.

/lgtm
/approve

grokspawn:
> @grokspawn nothing should write into `config/rbac` anymore. Not sure that ignoring it makes a lot of sense.

Ah!  I had leftover artifacts there after running the test for you this morning, and presumed that path was still used.  If vestigial, ignore!

PR #938, opened 2024-06-14
🌱 mac setup steps

grokspawn:
# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,29 @@ For a more complete overview of OLM v1 and how it differs from OLM v0, see our [
 
 ## Getting Started
 You’ll need a Kubernetes cluster to run against. You can use [KIND](https://sigs.k8s.io/kind) to get a local cluster for testing, or run against a remote cluster.
-**Note:** Your controller will automatically use the current context in your kubeconfig file (i.e. whatever cluster `kubectl cluster-info` shows).
+
+> [!NOTE] 
+> Your controller will automatically use the current context in your kubeconfig file (i.e. whatever cluster `kubectl cluster-info` shows).
+
+### Additional setup on Machintosh computers
+On Macintosh computers some additional setup is necessary to install and configure compatible tooling.
+
+#### Install Homebrew and tools
+Follow the instructions to [installing Homebrew](https://docs.brew.sh/Installation) and then execute the following to install tools:
+
+```sh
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -23,7 +23,29 @@ For a more complete overview of OLM v1 and how it differs from OLM v0, see our [
...
+
+```sh
+brew install bash gnu-tar gsed
+```
+
+#### Configure your shell
+Modify your login shell's `PATH` to prefer the new tools over those in the existing environment.  This example should work either with `zsh` (in $HOME/.zshrc) or `bash` (in $HOME/.bashrc):
+
+```sh
+for bindir in `find $(brew --prefix)/opt -type d -follow -name gnubin -print`
```
m1kola, line 44: I normally try to avoid running commands in `.zshrc` because it can slow down creating a new session significantly. Maybe it is not that big of a deal with `brew --prefix` and `find`, but I had this issue with `npm`.

If we can avoid this - that would be great. Here is my `.zshrc`:
```zsh
# coreutils
PATH="/opt/homebrew/opt/coreutils/libexec/gnubin:$PATH"

# gnu-tar
PATH="/opt/homebrew/opt/gnu-tar/libexec/gnubin:$PATH"
```

Not sure if `coreutils` are necessary for OLMv1.

m1kola: But we might still need to use `brew --prefix`. I think it it is different on Intel and ARM Macs unfortunately.

varshaprasad96 (approved):
/lgtm
/approve

README.md
grokspawn: I've been doing this for quite awhile never encountered new session timing issues, though I appreciate the concern.  The additional advantage of the approach is that if we identify we need additional tooling, sourcing the resource file will automagically pick it up.  

npm is known for sometimes performing package maintenance at awkward times, and I definitely wouldn't want to include a function in shell resources which invoked it.  I see no harm in a targeted path-walk though.

As far as I know we only need bash, gnu-tar, and gsed, unless you can think of a reason that we need coreutils...

m1kola:
```
find $(brew --prefix)/opt -type d -follow -name gnubin -print  0,01s user 0,10s system 96% cpu 0,112 total
```
Ok, it is not that bad.

Issue #879, opened 2024-05-21
Improve unit testing of `reconcile()` function.


tmshort:
As discussed in the meeting earlier today, I'd like to see us move to a model where this functions is a bit easier to unit test by creating logical abstractions where they make sense and mocking them during unit testing.

Totally fine with this being a follow-up, just commenting to track the recommendation :)

_Originally posted by @everettraven in https://github.com/operator-framework/operator-controller/pull/846#discussion_r1604028873_

joelanford:
This was closed by #1068 and #1033

PR #941, opened 2024-06-14
✨ remove unnecessary flag, optimize catalog watch handler, stop watching non-existent unpack pods

joelanford:
# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/controllers/clusterextension_controller.go
+++ b/internal/controllers/clusterextension_controller.go
@@ -36,7 +36,6 @@ import (
 	"helm.sh/helm/v3/pkg/postrender"
 	"helm.sh/helm/v3/pkg/release"
 	"helm.sh/helm/v3/pkg/storage/driver"
-	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -48,9 +47,9 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 	crcontroller "sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/event"
+	crfinalizer "sigs.k8s.io/controller-runtime/pkg/finalizer"
 	crhandler "sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
@@ -80,7 +79,6 @@ import (
 // ClusterExtensionReconciler reconciles a ClusterExtension object
 type ClusterExtensionReconciler struct {
 	client.Client
-	ReleaseNamespace      string
 	BundleProvider        BundleProvider
 	Unpacker              rukpaksource.Unpacker
 	ActionClientGetter    helmclient.ActionClientGetter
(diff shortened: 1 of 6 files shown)


internal/controllers/clusterextension_controller.go
```diff
@@ -574,7 +572,6 @@ func (r *ClusterExtensionReconciler) SetupWithManager(mgr ctrl.Manager) error {
 				return true
 			},
 		}).
-		Watches(&corev1.Pod{}, mapOwneeToOwnerHandler(mgr.GetClient(), mgr.GetLogger(), &ocv1alpha1.ClusterExtension{})).
```
varshaprasad96, line 577: Since we're on it - we should also be able to remove the rbac too, to watch pods: https://github.com/operator-framework/operator-controller/blob/cb63023e916dc302548ce841c7e5540303ad2f2b/internal/controllers/clusterextension_controller.go#L105-L107.

If not in this PR, it can be done in follow up too

joelanford: I can do it now.

joelanford: Pushed a change. I removed the pods stuff, and changed the configmaps RBAC (what was that for?) to secrets since the helm stuff needs to manage release secrets.

It's all sorta moot though because:
- We give ourselves `*/*/*` in the next line.
- We'll eventually drop the secrets permission here because we'll expect the SA to have the necessary permission to handle their own release secret.

varshaprasad96: > and changed the configmaps RBAC (what was that for?)

I _think_ it may have gotten ported over from Extension, where Kapp needed CRUD permissions for configmaps.

varshaprasad96 (dismissed):
/lgtm
/approve

tmshort:
Looks like 'make verify' or equivalent is needed.
/hold

joelanford:
Oops forgot to re-gen RBAC. Will do!

internal/controllers/clusterextension_registryv1_validation_test.go
```diff
@@ -116,6 +117,7 @@ func TestClusterExtensionRegistryV1DisallowDependencies(t *testing.T) {
 				ActionClientGetter:    helmClientGetter,
 				Unpacker:              unpacker,
 				InstalledBundleGetter: mockInstalledBundleGetter,
+				Finalizers:            crfinalizer.NewFinalizers(),
```
joelanford, line 120: It's a bit disconcerting to me that so much of the `ClusterExtensionReconciler` can be constructed _so_ differently than the real one can. We should discuss further, but off the top of my head, I wonder if we can use the exact same reconciler as is configured in `main.go`.
1. For catalogd we can use httptest to setup a mock catalogd server and use the _real_ ClusterExtension http client.
2. An image registry is an http server. I wonder if there is an in-process implementation that would be easy to run for test purposes.
3. Use `t.TempDir()` subdirectories for any necessary local storage and caching that happens during reconcile. 
4. For the kubernetes cluster, we use envtest.

If we had that setup, we wouldn't actually need to configure the ClusterExtension reconciler any differently.

joelanford: More immediately, we should have a unit tests that make sure finalizers are actually being handled properly during reconcile.

varshaprasad96: > It's a bit disconcerting to me that so much of the ClusterExtensionReconciler can be constructed so differently than the real one can.

This is definitely up for discussion and needs separate set of effort. The whole point of these unit tests is to make sure that the reconciler code path traversed during various instances is as expected. Which is why we ensure that the mock unpacker returns quickly as expected. The code coverage for unpacking and storage has separate unit tests (currently in rukpak).

I'm not sure if we immediately need to focus on this - rather if possible try finishing up https://github.com/operator-framework/operator-controller/issues/879 as immediately as possible (even if it requires us to mock some of the components) to ensure we don't skip any nuances in the reconciler code.

varshaprasad96 (approved):
/lgtm
/approve

tmshort (approved):

PR #951, opened 2024-06-17
🌱 Add test for fetching bundles

m1kola:
# Description

Cover what happens when we fail to fetch bundles from catalog.

This should solve flaky test coverage. It looks like in E2E we sometimes are hitting this condition and sometimes we don't which results in coverage fluctuation:

Screenshot 2024-06-17 at 16 37 28

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/controllers/clusterextension_controller_test.go
+++ b/internal/controllers/clusterextension_controller_test.go
@@ -3,6 +3,7 @@ package controllers_test
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"testing"
 
@@ -28,6 +29,7 @@ import (
 	"github.com/operator-framework/operator-controller/internal/conditionsets"
 	"github.com/operator-framework/operator-controller/internal/controllers"
 	"github.com/operator-framework/operator-controller/pkg/features"
+	testutil "github.com/operator-framework/operator-controller/test/util"
 )
 
 // Describe: ClusterExtension Controller Test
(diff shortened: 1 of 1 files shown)


internal/controllers/clusterextension_controller_test.go
```diff
@@ -1445,3 +1447,48 @@ var testBundleList = []*catalogmetadata.Bundle{
...
+			PackageName:      "prometheus",
+			InstallNamespace: "default",
+		},
+	}
+	err := cl.Create(ctx, clusterExtension)
+	defer func() {
+		require.NoError(t, cl.Delete(ctx, &ocv1alpha1.ClusterExtension{
+			ObjectMeta: metav1.ObjectMeta{Name: extKey.Name},
+		}))
+	}()
```
tmshort, line 1471: Should this `defer` be _after_ the test for success (below)?

m1kola: @tmshort as far as I understand `require.NoError` exists immediately. I think if we `defer` after it - there is a chance that we will not clean up.

On the other hand the chance of `Create` returning non-nil error and creating the resources are very slim (Can It happen at all? E.g. transport error after creation?)

m1kola: Updated

joelanford:
> It looks like in E2E we sometimes are hitting this condition and sometimes we don't which results in coverage fluctuation

Adding more test coverage is good, but it seems pretty bad that other e2e tests are non-deterministic in this way. Did you dive into this enough to understand why the flakiness is happening?

m1kola:
@joelanford Here is what _I think_ happens:

1. We create a catalog.
2. We run a test.
3. We sometimes hit "error fetching bundles" and sometimes not. When we do hit it - it gets coverage. 
4. E2Es wait for a certain condition and which gives the controller a chance to requeue the reconcile request. Eventually reconcile succeeds.

I think error happens because in tests we create catalogs and we do not wait for them to become ready: we create `ClusterExtension` immediately after creating a `ClusterCatalog`. So most likely `ClusterExtension` controller attempts to reconcile while catalog is still unpacking/not ready.

We can look into adding waits after creating `ClusterCatalog`, but IMO it is good that we do not have these artificial waits: we should be able to reconcile once catalog becomes ready.

tmshort (approved):
/lgtm

PR #956, opened 2024-06-18
:sparkles: Certificate support for image registry

### LINKED ISSUE #921: Remove `bundle.connection.config/insecureSkipTLSVerify` annotation
The `bundle.connection.config/insecureSkipTLSVerify` annotation was added to ClusterExtensions to allow TLS connections to local docker registries to succeed. This should be removed before v1.0.0.

It should be replaced with an API update to support skip TLS verify and/or a certificate secret/configmap/something.

There may also be a need to update the deplpoyment of the operator-controller to support certificates.

tmshort:
Updates the ClusterExtension API to support a certificate to retreive a bundle from an image registry.

Fixes #921 

# Description

## Reviewer Checklist

- [ ] API Go Documentation


go.mod
```diff
@@ -13,7 +13,7 @@ require (
 	github.com/operator-framework/catalogd v0.14.0
 	github.com/operator-framework/helm-operator-plugins v0.2.2-0.20240520180534-f463c36fedf9
 	github.com/operator-framework/operator-registry v1.43.1
-	github.com/operator-framework/rukpak v0.23.1
+	github.com/operator-framework/rukpak v0.23.2-0.20240618133950-e1d8b0e32344
```
m1kola, line 16: Could you please push a new tag of rukpak and pin it to a tag? Let me know if you need help with it (e.g. due to permission).

api/v1alpha1/clusterextension_types.go
```diff
@@ -78,6 +97,10 @@ type ClusterExtensionSpec struct {
 	// the bundle may contain resources that are cluster-scoped or that are
 	// installed in a different namespace. This namespace is expected to exist.
 	InstallNamespace string `json:"installNamespace"`
+
+	//+optional
+	// RegistryTLS defines the connection parameters to retrieve an image from a registry
+	RegistryTLS *ClusterExtensionTLS `json:"registryTLS,omitempty"`
```
m1kola, line 103: It feels like it does not belong to `ClusterExtension`. At least in the use case when we install packages from a catalog it feels like specifying this every time we create a `ClusterExtension` is excessive.

We need to think about UX for the following use cases:
1. We install extensions from a catalog. I think in this case we probably don't want users to specify certs in `ClusterExtension` every time they install something from a catalog. Feels like some configuration should be done on `ClusterCatalog` level once and re-used.
2. We install something directly (like future support for direct helm installation).

internal/controllers/clusterextension_controller.go
```diff
@@ -550,21 +552,54 @@ func (r *ClusterExtensionReconciler) generateBundleDeploymentForUnpack(bundlePat
...
 	}
-	value, ok := ce.Annotations[bundleConnectionAnnotation]
-	if !ok {
-		return false
+	return ce.Spec.RegistryTLS.InsecureSkipTLSVerify
+}
+
+func (r *ClusterExtensionReconciler) getCertificateData(ctx context.Context, ce *ocv1alpha1.ClusterExtension) (string, error) {
+	if ce == nil || ce.Spec.RegistryTLS == nil || ce.Spec.RegistryTLS.CertificateSecretRef == nil {
+		return "", nil
```
m1kola, line 571: No error and no cert here?

api/v1alpha1/clusterextension_types.go
```diff
@@ -43,6 +43,25 @@ const (
 	UpgradeConstraintPolicyIgnore UpgradeConstraintPolicy = "Ignore"
 )
 
+// Similar to NamespacedName, but with json
+type ClusterExtensionSecretRef struct {
+	// Name of the secret
+	Name string `json:"name"`
+	// Namespace of the secret
+	Namespace string `json:"namespace,omitempty"`
```
varshaprasad96, line 51: Should we force this to be the install namespace?

internal/controllers/clusterextension_controller.go
```diff
@@ -550,21 +552,54 @@ func (r *ClusterExtensionReconciler) generateBundleDeploymentForUnpack(bundlePat
...
+	}
+	secretName := getNamespacedName(*ce.Spec.RegistryTLS.CertificateSecretRef)
+	var secret = &corev1.Secret{}
+	if err := r.Reader.Get(ctx, secretName, secret); err != nil {
+		return "", fmt.Errorf("unable to get secret %v: %w", secretName, err)
+	}
```
m1kola, line 577: It looks like `r.Reader` is a live client (no cache). How often are we going to be hitting API server with this? Every reconcile?

cmd/manager/main.go
```diff
@@ -210,6 +209,7 @@ func main() {
 
 	if err = (&controllers.ClusterExtensionReconciler{
 		Client:                cl,
+		Reader:                mgr.GetAPIReader(),
```
varshaprasad96, line 212: Was this an intentional decision to use an API only reader to make sure an updated certificate is being used?

I'm just wondering if it's required, because we would be using operator-controller's client configured with manager, instead of using the client with SA. (Not a change for now, if just wondering if an API only reader is required specifically instead of a cache backed one)

api/v1alpha1/clusterextension_types.go
everettraven, line 51: Should we allow for specification of the namespace here? I'm wondering if using the install namespace as a point where all other resources referenced should exist is more reasonable?

internal/controllers/clusterextension_controller.go
```diff
@@ -550,21 +552,54 @@ func (r *ClusterExtensionReconciler) generateBundleDeploymentForUnpack(bundlePat
...
+	if secret.Type != corev1.SecretTypeTLS {
+		return "", fmt.Errorf("invalid type in secret %v: %v", secretName, secret.Type)
+	}
+	var certs []string
+	// Get any 'ca.crt'
+	data, ok := secret.Data[corev1.ServiceAccountRootCAKey]
+	if ok && len(data) > 0 {
+		certs = append(certs, string(data[:]))
```
m1kola, line 585: `string(data[:])` makes sense for byte arrays, but here you can just do `string(data)` since `data` is already a slice.

Is there another reason for this slicing?

internal/controllers/clusterextension_controller.go
```diff
@@ -550,21 +552,54 @@ func (r *ClusterExtensionReconciler) generateBundleDeploymentForUnpack(bundlePat
...
+	}
+	// Get any 'tls.crt'
+	data, ok = secret.Data[corev1.TLSCertKey]
+	if ok && len(data) > 0 {
+		certs = append(certs, string(data[:]))
```
m1kola, line 590: Same as above.

api/v1alpha1/clusterextension_types.go
tmshort: The secret is associated with the image registry, not the installation of the bundle itself. This secret is independent of the installation namespace.

cmd/manager/main.go
tmshort: Yes; if we use the non-Reader interface, it requires adding Secrets to the cache, which we really don't want to do. This was suggested by @joelanford

internal/controllers/clusterextension_controller.go
tmshort: The certificate is optional, if it's not present, we don't use it.

api/v1alpha1/clusterextension_types.go
tmshort: Yes, this was the intent.  This is the certificate used to verify the connection to the _image-registry_ from which we get the bundle. The certificate can't be part of the catalog, as it may expire, and it could be an external repository or other location as you point out.

HOWEVER, this is _optional_, as the operator-controller will load the system certificates, and if the image registry's certificate is signed by a system certificate, then it's not necessary.

internal/controllers/clusterextension_controller.go
tmshort: Correct. We don't want to cache _all_ the Secrets of the system.

IT should be accessed only when we actually attempt to install.

internal/controllers/clusterextension_controller.go
tmshort: You're right, it's probably not necessary.

api/v1alpha1/clusterextension_types.go
tmshort: I actually disagree, the registry and it's secret are likely in a separate namespace from the installation namespace.

tmshort: To be clear, it's not required on every installation.

If the docker-registry's certificate is signed by a system CA, this option is not required. This requires additional setup (i.e. cert-manager/trust-manager/deployment updates).

api/v1alpha1/clusterextension_types.go
m1kola: The more I think about this - the more I think we should not have this on `ClusterExtension`. Another reason - we do not specify specific catalog in `ClusterExtension`. If on `ClusterExtension` you say you want to install `argocd-operator` - it can come from any catalog installed on the cluster: it might come from a catalog which uses an image registry where certs are issued by a public CA, but it might also come from a catalog which uses an image registry with a self-signed cert. We don't know on the `ClusterExtension` level.

I can totally see that some users are not going to know what to specific in `registryTLS` because they don't know how `argocd-operator` will resolve.

> This is the certificate used to verify the connection to the _image-registry_ from which we get the bundle.

I understand that. But whoever configures `ClusterCatalog` probably knows better about what image registries the catalog uses. So they can provide a ref to a relevant CA on `ClusterCatalog` level.

> The certificate can't be part of the catalog, as it may expire, and it could be an external repository or other location as you point out.

Not sure that I'm following this.

api/v1alpha1/clusterextension_types.go
m1kola: > To be clear, it's not required on every installation.

You are right. But it is required on every installation if you use a registry with self-signed certs (like we do in our own tests: see `addCertToSpec` in this PR). I would expect it to be quite a common pattern.

api/v1alpha1/clusterextension_types.go
tmshort: If the certificate is part of the catalog metadata, it could easily become stale - especially if a separate image-registry is used. And as you point out, the argocd could come from any catalog, and subsequently any docker-registry.

Again, this is _optional_, and is intended for those specific docker-registries which may be private (e.g. our e2e docker-registries) or for which don't use a system certificate.

_This should really be considered an exception, not the rule._

@joelanford, I believe it was your idea to add a certificate to the ClusterExtension, do you have an additional response?

api/v1alpha1/clusterextension_types.go
tmshort: I would actually expect it to be an exception, not a rule. The operator-controller deployment can be modified to include CAs used by the cluster.

api/v1alpha1/clusterextension_types.go
m1kola: > If the certificate is part of the catalog metadata

I'm not suggesting to put a certificate into catalog metadata. I'm suggesting to _potentially_ do what you did for `ClusterExtension`, but for `catalogd`'s `ClusterCatalog` API instead.

My idea is that people using `ClusterExtension` should not be forced to specify CA every time they install something on a cluster where self-signed certs are used. The cluster admins should be able to specify refs to secrets containing required CAs once (when adding a catalog to their cluster) and `ClusterExtension` API can remain the same.

What I suggest might also not be the right approach, but it is clear to me that we as a group need to spend some more time on the design for this.

> This should really be considered an exception, not the rule.

If we provide this API - we and our users will be stuck with it for long time. We will have to support it and users will be forced to use it.

I agree with you - it won't affect people who use image registry with certs signed by public CAs. But it will result in a terrible user experience for people who install something on a cluster with self-signed certs. For them I see at least the following UX issues:

1. They will have to specify `registryTLS` for every `ClusterExtension`.
2. It is not clear what CA to specify on `ClusterExtension`. You have to know in advance how your package & version are going to resolve (is it going to use a registry with self-signed certs?). I think it is going to be more of a pain if people using `ClusterExtension` are not the same people configuring catalogs on the cluster.

IMO - we should either not support this use case at all or design so that we provide nice UX for all the use cases.

Let me know if you want to discuss this on a call. We should probably invite @joelanford as the author of the original idea.

tmshort: I spoke with Joe Tuesday night and we are going to abandon the addition to the API _for now_ and re-evaluate it later. This PR will still remove the annotations and fix the e2e to properly use certificates (and not use `InsecureSkipTLSVerify`).

m1kola: Sounds good!

> This PR will still remove the annotations and fix the e2e to properly use certificates

Could you please create a new PR for this instead? This one has quite a lot comments about the original `registryTLS` implementation. I think it will be easier to review in a fresh PR.

tmshort:
See #960 for a different approach

PR #960, opened 2024-06-20
:sparkles: Certificate support for image registry

tmshort:
Fixes: #921

Remove the InsecureSkipTLSVerify annotations.

* Create a ClusterIssuer CA (via cert-manager) that is used by OLMv1 e2e
* Update the operator controller to specify a cert directory, rather than a single file.
* Use this directory for catalogd and image-registries
* Update the deployment to reference CAs appropriately

# Description

## Reviewer Checklist

- [ ] API Go Documentation


tmshort:
To consolidate `catalogd`'s certificate would require modifying catalogd; it would also allow catalogd to be placed back into its own namespace.

config/overlays/e2e/kustomization.yaml
```diff
@@ -1,7 +1,7 @@
 namespace: olmv1-system
 
 resources:
-- ../../overlays/tls
+- ../../overlays/certs
```
m1kola, line 4: Are we only using it in our E2Es? Should we put it inside `config/overlays/e2e`?

internal/controllers/clusterextension_controller.go
```diff
@@ -549,24 +551,14 @@ func (r *ClusterExtensionReconciler) generateBundleDeploymentForUnpack(bundlePat
 				Type: rukpakv1alpha2.SourceTypeImage,
 				Image: &rukpakv1alpha2.ImageSource{
 					Ref:                   bundlePath,
-					InsecureSkipTLSVerify: isInsecureSkipTLSVerifySet(ce),
+					InsecureSkipTLSVerify: false,
```
m1kola, line 554: Shall we just drop this field altogether? Empty value is `false` anyway.

config/overlays/certs/kustomization.yaml
```diff
@@ -0,0 +1,8 @@
+namespace: olmv1-system
+
+resources:
+- ../../overlays/tls
```
m1kola, line 4: Do I understand correctly that `config/overlays/certs` only exists for E2Es? If yes, would it make sense to put this into `config/overlays/e2e`?

cmd/manager/main.go
```diff
@@ -79,11 +79,11 @@ func main() {
 		cachePath                 string
 		operatorControllerVersion bool
 		systemNamespace           string
-		caCert                    string
+		caCertDir                 string
 	)
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
-	flag.StringVar(&caCert, "ca-cert", "", "The TLS certificate to use for verifying HTTPS connections to the Catalogd web server.")
+	flag.StringVar(&caCertDir, "ca-cert", "", "The directory of TLS certificate to use for verifying HTTPS connections to the Catalogd and docker-registry web servers.")
```
m1kola, line 86: 1. Should we rename the `ca-cert` flag itself as well?
2. Do I understand correctly that we are changing this to a path to a dir so we can handle multiple CAs? E.g. `/var/certs` will contain CAs of catalogd and OLM generated one?

config/overlays/e2e/kustomization.yaml
tmshort: There are 3 different categories needed:
1.base
2. ext-dev-e2e
3. e2e

And they build upon each other (e2e adds coverage that isn't part of ext-dev-e2e). i could clean this up (I think there are now 4 categories), but I was aiming for minimal changes.

config/overlays/certs/kustomization.yaml
tmshort: The `e2e` has coverage that's not needed for the ext-dev-e2e, but the ext-dev-e2e needs certs as well.

cmd/manager/main.go
tmshort: Yes it's a directory and can handle multiple files., although I had thought I could also support specifying a single filename if desired.

config/overlays/e2e/kustomization.yaml
m1kola: @tmshort I currently only see only two overlays:
* tls (default)
* e2e (tls + coverage)

I think we should try to keep the number overlays to a minimum. If `overlays/certs` only introduced for the purpose of e2e tests - then IMO we should include this stuff into e2e.

> e2e adds coverage that isn't part of ext-dev-e2e

Gathering coverage from ext-dev-e2e will be beneficial. If absence of coverage in ext-dev-e2e makes it harder to add certs into existing setup - let's unify ext-dev-e2e and e2e (in a separate PR ideally) and then come back to this.

What do you think?

config/overlays/certs/kustomization.yaml
m1kola: It looks like I commented about the same thing twice in different places. Let's continue here: https://github.com/operator-framework/operator-controller/pull/960/files#r1648592694

Closing this thread.

cmd/manager/main.go
m1kola: Ack on the second point.

Back to first point: do you want to keep the flag name as `ca-cert` or rename to something like `ca-dir` or `ca-cert-dir` or something else?

config/overlays/e2e/kustomization.yaml
tmshort: I'm reducing the number of overlays. Anything beyond that is definitely new PR territory.

cmd/manager/main.go
tmshort: I'm ambivalent... but it should probably contain either `dir` or be plural...

tmshort:
I can't get kustomize to use a different directory, so I'm applying the yaml for the certificates directly from a file, which also means I can apply them for Tilt from the same source.

cmd/manager/main.go
joelanford: My vote `ca-certs-dir`.
- `ca-certs` plural, because the directory can contain multiple certificate files
- `dir` because we're asking for a directory (and only a directory, IMO)

internal/httputil/httputil.go
```diff
@@ -5,30 +5,60 @@ import (
...
-			return nil, err
+			return err
 		}
-		caCertPool := x509.NewCertPool()
-		caCertPool.AppendCertsFromPEM(cert)
-		tlsConfig := &tls.Config{
-			RootCAs:    caCertPool,
-			MinVersion: tls.VersionTLS12,
+		if info.IsDir() {
+			return nil
```
joelanford, line 24: Is the intention to recurse through all subdirectories? I'm tempted to suggest that we stick to finding files in a single named directory without recursing, but I'm curious if you have a specific reason.

To avoid recursing, we can return `filepath.SkipDir` here.
```suggestion
			return filepath.SkipDir
```

internal/httputil/httputil.go
```diff
@@ -5,30 +5,60 @@ import (
...
+	"strings"
 	"time"
 )
 
-func BuildHTTPClient(caCert string) (*http.Client, error) {
-	httpClient := &http.Client{Timeout: 10 * time.Second}
-
-	if caCert != "" {
-		// tlsFileWatcher, err := certwatcher.New(caCert, "")
+func LoadCerts(caDir string) (string, error) {
```
joelanford, line 13: Seems like it would be a nicer abstraction if this was:
```suggestion
func AddCertificatesFromDirectory(certPool *x509.CertPool, caDir string) error { ... }
```
In which case, we could add each file's content, collect errors reading/adding each file, and then tell callers which specific files failed for which reasons.

internal/httputil/httputil.go
tmshort: No, it's intended to skip that directory. I didn't see this in any examples.

internal/httputil/httputil.go
tmshort: This is not compatible with the current rukpak API, which expects a string of PEM. Maybe when rukpak is fully integrated into operator-controller? I agree that this would be better for a larger directory. There does not appear to be a way to extract certificates/PEM from a CertPool, so it can't be done with this signature, yet.

scripts/install.tpl.sh
```diff
@@ -35,6 +35,8 @@ function kubectl_wait() {
 kubectl apply -f "https://github.com/cert-manager/cert-manager/releases/download/${cert_mgr_version}/cert-manager.yaml"
 kubectl_wait "cert-manager" "deployment/cert-manager-webhook" "60s"
 
+kubectl apply -f testdata/certs/issuers.yaml
+
```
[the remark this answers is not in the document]
tmshort: I tried numerous ways to get the secret for the `ClusterIssuser` into the `cert-manager` namespace, but despite trying various `transformations:`, I couldn't get **kustomize** to do it. Using a separate file at least avoids duplication among the two e2e tests. But I can always do it via a HEREDOC.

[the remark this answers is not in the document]
tmshort: i.e. if you can figure out how to make kustomize work the way we want, I'd much prefer to do it that way!

internal/httputil/httputil.go
joelanford: Oh I was just looking at `LoadCerts` being called in `BuildHTTPClient`, which turns around and consumes the `certs` string by immediately adding to the `caCertPool`.

And that made me think we could pass the `caCertPool` into this function and call `AppendCertsFromPEM` for each file's contents. Then, when my suggested function signature returns, you end up with `caCertPool` having everything in `caDir` added to it.

Seems like I'm missing something though?

internal/httputil/httputil.go
tmshort: This appears to skip `"."` as a directory, meaning, it skips the current directory... so the e2es fail.

internal/httputil/httputil.go
tmshort: It's also used when creating a BundleDeployment for a rukpak API.
https://github.com/operator-framework/operator-controller/blob/ed4e40eaca385c8bd4d1d7e8329c9f9c76013d95/internal/controllers/clusterextension_controller.go#L252

So, we need to change the rukpak API to accept a caPool first, then we can do as you suggest.

internal/httputil/httputil.go
tmshort: OK, found one example, and the example has a comparison with a directory name, so we'd need to do that to properly use `filepathSkipDir`; not sure it's worth it?

config/overlays/tls/kustomization.yaml
```diff
@@ -12,9 +12,11 @@ apiVersion: kustomize.config.k8s.io/v1beta1
...
+- resources/manager_cert.yaml
 
 patches:
 - target:
     kind: Deployment
     name: controller-manager
-  path: patches/manager_deployment_cert.yaml
\ No newline at end of file
+  path: patches/manager_deployment_cert.yaml
+- path: patches/manager_cert_patch.yaml
```
m1kola, line 22: `manager_deployment_cert.yaml` and `manager_cert_patch.yaml` seem to patch the same deployment and same fields but in different ways. I think it would be good to standardise.

config/overlays/tls/resources/manager_cert.yaml
```diff
@@ -0,0 +1,16 @@
...
+  dnsNames:
+    - operator-controller.olmv1-system.svc
+    - operator-controller.olmv1-system.svc.cluster.local
+  privateKey:
+    algorithm: ECDSA
+    size: 256
+  issuerRef:
+    name: olmv1-ca
+    kind: ClusterIssuer
+    group: cert-manager.io
```
m1kola, line 16: This introduces a dependency on `ClusterIssuer` which is only being created In `install.tpl.sh`/in Tilt.

This means it is no longer possible to do this to install:

```
kubectl apply -f https://github.com/operator-framework/catalogd/releases/download/$CATALOGD_VERSION/catalogd.yaml
kubectl apply -f https://github.com/operator-framework/operator-controller/releases/download/$VERSION/operator-controller.yaml
```

You have to checkout the repo run the install script.

Are we ok with that?

cmd/manager/main.go
```diff
@@ -79,11 +79,11 @@ func main() {
 		cachePath                 string
 		operatorControllerVersion bool
 		systemNamespace           string
-		caCert                    string
+		caCertDir                 string
 	)
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
-	flag.StringVar(&caCert, "ca-cert", "", "The TLS certificate to use for verifying HTTPS connections to the Catalogd web server.")
+	flag.StringVar(&caCertDir, "ca-cert-dir", "", "The directory of TLS certificate to use for verifying HTTPS connections to the Catalogd and docker-registry web servers.")
```
m1kola, line 86: I'm +1 to what Joe suggested [here](https://github.com/operator-framework/operator-controller/pull/960#discussion_r1649511251):

> My vote `ca-certs-dir`.
> 
> * `ca-certs` plural, because the directory can contain multiple certificate files
> * `dir` because we're asking for a directory (and only a directory, IMO)

config/overlays/tls/resources/manager_cert.yaml
m1kola: Also about related changes in `scripts/install.tpl.sh` where we create CA, etc.

As far as I understand - we need to create CA and patch deployment of the manger only for E2E setups. If it is the case - can we put this in `config/overlays/e2e/`?

I would like to avoid having to create all this on every cluster just because our E2E setup requires it.

internal/httputil/httputil.go
joelanford: Maybe switch over to `os.ReadDir()`, which makes the intent more clear anyways?

config/overlays/tls/resources/manager_cert.yaml
tmshort: We need cert-manager regardless, AFAICT, as catalogd now creates certificates. So, the method listed above no longer works (or at least it shouldn't).
I tried to put this into the e2e overlay, but could not get the namespace to work as expected. Please see https://github.com/operator-framework/operator-controller/pull/960#discussion_r1649513635; I have been unable to figure out how to apply resources into the `cert-manager` namespace. i.e. `kustomize` seems to be locking me into a single namespace...

tmshort:
TL;DR: The challenge I've run into w/ adding the certificates is being able to put them into the cert-manager namespace for the `ClusterIssuer`. I tried various `kustomize` options and `transformations` and none of them seemed to work. I would love to get rid of the HEREDOC and yaml used to create the certificates. If anyone has any ideas, I'm all ears.

m1kola:
A number of my comments in this PR were based on the understanding that this is only about making our E2E setup work with TLS. After chatting with @joelanford and @tmshort I got more context.

The primary objective of the PR to get rid of `InsecureSkipTLSVerify` and use TLS in E2E, but it also prepares the ground for using single `ClusterIssuer` for all OLMv1 components. E.g. we can use it for catalogd too.

Explanation from @joelanford:
> 1. create a single ClusterIssuer in the standard install
> 2. any OLMv1 component that needs a certificate gets it from there, no matter which namespace that component is in.
> 3. any OLMv1 component that needs a CA can create a certificate (if it doesn't already have one), and that cert secret will contain the CA.
> 
> That setup would mean that we could move catalogd back to its own namespace if we wanted to, and it would mean that the e2e registry would be automatically trusted by the standard setup because operator-controller would already have the CA.

internal/httputil/httputil.go
tmshort: will look

everettraven:
Follow ups should be captured by this epic: https://github.com/operator-framework/operator-controller/issues/967

cmd/manager/main.go
tmshort: Done

tmshort:
> The changes themselves look fine to me, but I noticed the e2es are failing. Any insights as to why?

I made a change that passed due to caching issues... I had to fix it up and it should be ok now.

everettraven (approved):

testdata/certs/issuers.yaml
joelanford, line 1: I don't love that this is duplicated here and in the HEREDOC, but I also recognize that this setup is temporary. Can we just make sure that our follow-ups include a callout to remove this duplication?

tmshort: See https://github.com/operator-framework/operator-controller/issues/967#issuecomment-2187205871

PR #963, opened 2024-06-21
🐛 fix: reduce unnecessary caching and remove predicate from dynamic watches

joelanford:
Our cache configuration resulted in watching all cluster-scoped resources for each cluster-scoped resource that shows up in a ClusterExtension. Now, we only watch:
- All ClusterExtensions
- All ClusterCatalogs
- All objects in the systemNamespace
- All other objects that specifically match the dependentPredicate

This commit also removes the dependent predicate so that we reconcile for any change to managed resources. This is important to ensure that future features (like health checking) can account for any change to managed operator objects.

# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/controllers/clusterextension_controller.go
+++ b/internal/controllers/clusterextension_controller.go
@@ -45,6 +45,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 	apimachyaml "k8s.io/apimachinery/pkg/util/yaml"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	crcontroller "sigs.k8s.io/controller-runtime/pkg/controller"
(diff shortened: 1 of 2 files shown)


internal/controllers/clusterextension_controller.go
```diff
@@ -572,24 +571,24 @@ func (r *ClusterExtensionReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
-		WithEventFilter(predicate.Funcs{
-			UpdateFunc: func(ue event.UpdateEvent) bool {
-				oldObject, isOldCatalog := ue.ObjectOld.(*catalogd.ClusterCatalog)
-				newObject, isNewCatalog := ue.ObjectNew.(*catalogd.ClusterCatalog)
-
-				if !isOldCatalog || !isNewCatalog {
-					return true
-				}
+			crhandler.EnqueueRequestsFromMapFunc(clusterExtensionRequestsForCatalog(mgr.GetClient(), mgr.GetLogger())),
+			builder.WithPredicates(predicate.Funcs{
```
perdasilva, line 575: a tab walks into a bar sits down. The barkeep asks him, why don't space?

perdasilva (approved):
lgtm

internal/controllers/clusterextension_controller.go
joelanford: 😄 Yeah, the formatting looks weird because the predicate needs to go with the ClusterCatalog watch, and it means it all gets indented by an extra tab.

PR #964, opened 2024-06-21
🐛 fix: remove unnecessary bundle cache volume

joelanford:
# Description

When I removed the flag that controlled the bundle cache root directory in #941, I missed the fact that we also had a volumeMount for it, which is a child of our broader `/var/cache` volume emptyDir volume.

This commit eliminates the unnecessary extra volume and instead just creates the `/var/cache/bundles` directory in the existing volume.

This is important to eliminate the potential for issues when attempting to rename files across filesystem boundaries using `os.Rename` (e.g. if we ever needed to do something like  rename `/var/cache/tmp/foobar.txt` to `/var/cache/bundles/foobar.txt`)

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -196,8 +196,13 @@ func main() {
 		os.Exit(1)
 	}
 
+	localStorageRoot := filepath.Join(cachePath, "bundles")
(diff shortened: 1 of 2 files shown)


perdasilva (approved):
lgtm

Issue #740, opened 2024-04-04
[epic] Review and finalize v1 APIs for OperatorController


joelanford:

acornett21:
- Depends-on: operator-framework/catalogd/issues/239

joelanford:
Related: #1088

PR #965, opened 2024-06-21
🐛 incremental improvement in catalogmetadata cache/client performance

### LINKED ISSUE #914: catalogmetadata client adds extreme latency to reconciliation
I've narrowed this latency down to the `Bundles()` call that is made for every single reconcile.

I added logging to the `catalogmetadata.Client.Bundles()` call to report start time and end time, and the function consistently takes around 30s, and it pegs a CPU.

This seems like a bug we should treat with a fairly high priority.

### LINKED ISSUE #948: Bug: cached catalog contents are not cleaned up when catalog is deleted.

joelanford:
# Description

This partially resolves #914 by doing two things:
1. Exploding the `all.json` response into separate directories organized by package
2. Updating the BundleProvider to require a package name to be provided in the request, which results in a much smaller set of FBC that needs to be re-parsed into catalogmetadata objects on each reconcile.

This is still not completely resolved though because step (1) still happens synchronously during the reconciliation of a ClusterExtension. In order to finish the improvement (and to complete #948), we need to implement a controller that watches and reconciles ClusterCatalog changes so that it can create (and delete) caches for operator-controller. 

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -135,7 +135,7 @@ test-ext-dev-e2e: $(OPERATOR_SDK) $(KUSTOMIZE) $(KIND) #HELP Run extension creat
 ENVTEST_VERSION := $(shell go list -m k8s.io/client-go | cut -d" " -f2 | sed 's/^v0\.\([[:digit:]]\{1,\}\)\.[[:digit:]]\{1,\}$$/1.\1.x/')
 UNIT_TEST_DIRS := $(shell go list ./... | grep -v /test/)
 test-unit: $(SETUP_ENVTEST) #HELP Run the unit tests
-	eval $$($(SETUP_ENVTEST) use -p env $(ENVTEST_VERSION) $(SETUP_ENVTEST_BIN_DIR_OVERRIDE)) && go test -count=1 -short $(UNIT_TEST_DIRS) -coverprofile cover.out
+	eval $$($(SETUP_ENVTEST) use -p env $(ENVTEST_VERSION) $(SETUP_ENVTEST_BIN_DIR_OVERRIDE)) && CGO_ENABLED=1 go test -count=1 -race -short $(UNIT_TEST_DIRS) -coverprofile cover.out
 
 image-registry: ## Setup in-cluster image registry
 	./test/tools/image-registry.sh $(E2E_REGISTRY_NAMESPACE) $(E2E_REGISTRY_NAME)
--- a/internal/catalogmetadata/cache/cache.go
+++ b/internal/catalogmetadata/cache/cache.go
@@ -3,13 +3,14 @@ package cache
 import (
 	"context"
 	"fmt"
-	"io"
+	"io/fs"
 	"net/http"
 	"os"
 	"path/filepath"
 	"sync"
 
 	catalogd "github.com/operator-framework/catalogd/api/core/v1alpha1"
+	"github.com/operator-framework/operator-registry/alpha/declcfg"
 
 	"github.com/operator-framework/operator-controller/internal/catalogmetadata/client"
 )
@@ -66,7 +67,7 @@ type filesystemCache struct {
 // resources that have been successfully reconciled, unpacked, and are being served.
 // These requirements help ensure that we can rely on status conditions to determine
 // when to issue a request to update the cached Catalog contents.
-func (fsc *filesystemCache) FetchCatalogContents(ctx context.Context, catalog *catalogd.ClusterCatalog) (io.ReadCloser, error) {
+func (fsc *filesystemCache) FetchCatalogContents(ctx context.Context, catalog *catalogd.ClusterCatalog) (fs.FS, error) {
 	if catalog == nil {
 		return nil, fmt.Errorf("error: provided catalog must be non-nil")
 	}
@@ -80,25 +81,23 @@ func (fsc *filesystemCache) FetchCatalogContents(ctx context.Context, catalog *c
 	}
 
 	cacheDir := filepath.Join(fsc.cachePath, catalog.Name)
-	cacheFilePath := filepath.Join(cacheDir, "data.json")
-
 	fsc.mutex.RLock()
 	if data, ok := fsc.cacheDataByCatalogName[catalog.Name]; ok {
 		if catalog.Status.ResolvedSource.Image.ResolvedRef == data.ResolvedRef {
 			fsc.mutex.RUnlock()
-			return os.Open(cacheFilePath)
+			return os.DirFS(cacheDir), nil
 		}
 	}
 	fsc.mutex.RUnlock()
 
 	req, err := http.NewRequestWithContext(ctx, http.MethodGet, catalog.Status.ContentURL, nil)
 	if err != nil {
-		return nil, fmt.Errorf("error forming request: %s", err)
+		return nil, fmt.Errorf("error forming request: %v", err)
 	}
 
 	resp, err := fsc.client.Do(req)
 	if err != nil {
-		return nil, fmt.Errorf("error performing request: %s", err)
+		return nil, fmt.Errorf("error performing request: %v", err)
 	}
 	defer resp.Body.Close()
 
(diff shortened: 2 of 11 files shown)


internal/catalogmetadata/client/client.go
```diff
@@ -43,66 +47,93 @@ type Client struct {
...
 		return nil, err
 	}
+
+	var errs []error
 	for _, catalog := range catalogList.Items {
-		// if the catalog has not been successfully unpacked, skip it
+		// if the catalog has not been successfully unpacked, report an error. This ensures that our
+		// reconciles are deterministic and wait for all desired catalogs to be ready.
 		if !meta.IsStatusConditionPresentAndEqual(catalog.Status.Conditions, catalogd.TypeUnpacked, metav1.ConditionTrue) {
+			errs = append(errs, fmt.Errorf("catalog %q is not unpacked", catalog.Name))
```
joelanford, line 63: This is a change in semantics. If we ignore catalogs that are not unpacked, users could run into issues if the same package exists in two different catalogs and they are expecting a specific result.

This is similar to the behavior in OLMv0, and there have been some complaints about it. For example, "I want to install CNV, I don't care if the community catalog is having issues". In order to solve this problem, I think we need to have a knob on the ClusterExtension API to allow users to select which ClusterCatalogs to resolve from (likely using a label selector)

internal/catalogmetadata/client/client_test.go
```diff
@@ -24,30 +23,30 @@ import (
 )
 
 func TestClient(t *testing.T) {
-	t.Run("Bundles", func(t *testing.T) {
```
joelanford, line 27: I removed this outer/unnecessary `t.Run` wrapper (mainly because it caused my IDE to not be able to find the individual test cases to run/debug separately).

In order to review changes in this test, I advise ignoring whitespace.

internal/catalogmetadata/cache/cache_test.go
```diff
@@ -50,203 +56,196 @@ const (
 	}`
 )
 
-func TestCache(t *testing.T) {
-	t.Run("FetchCatalogContents", func(t *testing.T) {
```
joelanford, line 54: Similar to my other comment, I removed this unnecessary `t.Run` wrapper. When reviewing, it's much easier to see actual changes if you ignore whitespace.

internal/testutil/fake_catalog_client.go
joelanford, line 1: Note that I moved this into `internal`. I noticed go-apidiff complain that I changed the `Bundles()` function signature realized we shouldn't have a test utility in our public API, especially since it returns a slice of objects that are also in an internal package.

internal/catalogmetadata/cache/cache.go
```diff
@@ -117,34 +116,49 @@ func (fsc *filesystemCache) FetchCatalogContents(ctx context.Context, catalog *c
...
 
-	if err = os.MkdirAll(cacheDir, os.ModePerm); err != nil {
-		return nil, fmt.Errorf("error creating cache directory for Catalog %q: %s", catalog.Name, err)
-	}
-
-	file, err := os.Create(cacheFilePath)
+	tmpDir, err := os.MkdirTemp(fsc.cachePath, fmt.Sprintf(".%s-", catalog.Name))
 	if err != nil {
-		return nil, fmt.Errorf("error creating cache file for Catalog %q: %s", catalog.Name, err)
+		return nil, fmt.Errorf("error creating temporary directory to unpack catalog metadata: %s", err)
```
grokspawn, line 125:
```suggestion
		return nil, fmt.Errorf("error creating temporary directory to unpack catalog metadata: %v", err)
```

Makefile
```diff
@@ -135,7 +135,7 @@ test-ext-dev-e2e: $(OPERATOR_SDK) $(KUSTOMIZE) $(KIND) #HELP Run extension creat
 ENVTEST_VERSION := $(shell go list -m k8s.io/client-go | cut -d" " -f2 | sed 's/^v0\.\([[:digit:]]\{1,\}\)\.[[:digit:]]\{1,\}$$/1.\1.x/')
 UNIT_TEST_DIRS := $(shell go list ./... | grep -v /test/)
 test-unit: $(SETUP_ENVTEST) #HELP Run the unit tests
-	eval $$($(SETUP_ENVTEST) use -p env $(ENVTEST_VERSION) $(SETUP_ENVTEST_BIN_DIR_OVERRIDE)) && go test -count=1 -short $(UNIT_TEST_DIRS) -coverprofile cover.out
+	eval $$($(SETUP_ENVTEST) use -p env $(ENVTEST_VERSION) $(SETUP_ENVTEST_BIN_DIR_OVERRIDE)) && CGO_ENABLED=1 go test -count=1 -race -short $(UNIT_TEST_DIRS) -coverprofile cover.out
```
joelanford, line 138: CGO is required for the race detector. TIL

perdasilva: yeah - got caught on that one rejiggering the v0 Makefile

internal/catalogmetadata/cache/cache.go
```diff
@@ -117,34 +116,49 @@ func (fsc *filesystemCache) FetchCatalogContents(ctx context.Context, catalog *c
...
-	if err = file.Sync(); err != nil {
-		return nil, fmt.Errorf("error syncing contents to cache file for Catalog %q: %s", catalog.Name, err)
+	if err := os.RemoveAll(cacheDir); err != nil {
+		return nil, fmt.Errorf("error removing old cache directory: %v", err)
 	}
-
-	if _, err = file.Seek(0, 0); err != nil {
-		return nil, fmt.Errorf("error resetting offset for cache file reader for Catalog %q: %s", catalog.Name, err)
+	if err := os.Rename(tmpDir, cacheDir); err != nil {
+		return nil, fmt.Errorf("error moving temporary directory to cache directory: %v", err)
```
perdasilva, line 156: curious to know what happens if fail here? Just run of the mill expbackoff? what happens to resolution during that period? Same thing?

joelanford: Yes to both.

internal/catalogmetadata/client/client.go
everettraven: Do we have this captured as an issue somewhere or do you anticipate this falling under the umbrella of https://github.com/operator-framework/operator-controller/issues/740?

internal/catalogmetadata/client/client.go
```diff
@@ -43,66 +47,93 @@ type Client struct {
...
-		defer rc.Close()
 
-		err = declcfg.WalkMetasReader(rc, func(meta *declcfg.Meta, err error) error {
+		var (
+			channelsMu     sync.Mutex
+			bundlesMu      sync.Mutex
+			deprecationsMu sync.Mutex
+		)
+
+		if err := declcfg.WalkMetasFS(ctx, packageFS, func(_ string, meta *declcfg.Meta, err error) error {
```
everettraven, line 88: Is this function now guaranteed to run concurrently? Is there any guarantees on the same `*declcfg.Meta` object not being processed more than once?

internal/catalogmetadata/client/client.go
```diff
@@ -128,6 +159,10 @@ func PopulateExtraFields(catalogName string, channels []*catalogmetadata.Channel
 		}
 	}
 
+	for _, bundle := range bundles {
+		slices.SortFunc(bundle.InChannels, func(a, b *catalogmetadata.Channel) int { return cmp.Compare(a.Name, b.Name) })
+	}
```
everettraven, line 164: Could you elaborate on the purpose of this?

internal/catalogmetadata/client/client.go
joelanford: It runs concurrently by default (based on number of CPUs Go thinks are available). There is an exactly once guarantee for an individual `*declcfg.Meta` object.

internal/catalogmetadata/client/client.go
joelanford: Because of the concurrency introduced in `WalkMetasFS` the order that channels are added processed is non-deterministic. That non-determinism manifests as non-deterministic ordering of `bundle.InChannels`. So this function gets us back to a deterministic order.

It doesn't _really_ matter except that tests become more annoying to implement if the order isn't deterministic.

everettraven: Would you mind adding a comment just highlighting the why?

internal/catalogmetadata/client/client.go
joelanford: Ultimately, we're parsing `n` files at a time (where `n` is number of goroutines). You may notice that the cache write metas into their own files. This is on purpose to make as many files as possible so that as many goroutines as possible (up to `n`, at least) have work to do.

internal/catalogmetadata/client/client.go
joelanford: https://github.com/operator-framework/operator-controller/compare/0661a996c07672f58b91a4e18ac35ba00710fd43..3df60bbed3d17645aec5eae832f884bb51a7b9a9

internal/catalogmetadata/client/client.go
joelanford: I am going to make a separate issue for this. Just wanted to get consensus on this approach before I did.

joelanford: https://github.com/operator-framework/operator-controller/issues/968

everettraven (approved):

PR #966, opened 2024-06-22
🐛 fix: make installNamespace immutable

### LINKED ISSUE #961: Possible bug: Do we properly handle mutations of `spec.installNamespace`?
I haven't had a chance to attempt a reproduction yet, but looking at our code, I think we have a few problems if installNamespace changes on a ClusterExtension that has a successful installation.
1. We will not find the installed version in the previous install namespace, so our reconciler will treat it like a new install (assuming a helm release in the new install namespace doesn't already exist)
2. We will not do anything with the helm release in the previous install namespace, which essentially means orphaning it.
3. I don't know for sure, but there may be issues transferring any shared objects between the old and new release manifests.

Options include:
1. Make the `spec.installNamespace` field immutable. This has implications on users that want to relocate a cluster extension. We would likely need `spec.deletionPolicy: orphan` to make this tolerable to those users.
1. Stop using helm releases. Use `helm template` instead? This has implications on support for helm hooks, but perhaps we detect and reject manifests that use hooks. This option would eliminate the problem described in #923 as well.

Any other options folks can think of?

joelanford:
# Description

Make `spec.installNamespace` immutable.

Fixes #961 

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/api/v1alpha1/clusterextension_types.go
+++ b/api/v1alpha1/clusterextension_types.go
@@ -73,6 +73,7 @@ type ClusterExtensionSpec struct {
 
 	//+kubebuilder:validation:Pattern:=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
 	//+kubebuilder:validation:MaxLength:=63
+	//+kubebuilder:validation:XValidation:rule="self == oldSelf",message="installNamespace is immutable"
 	//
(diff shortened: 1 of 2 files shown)


perdasilva (approved):
lgtm

Issue #967, opened 2024-06-24
[epic] Improve kustomize layouts across OLMv1 components


everettraven:
Recently we've done some work related to TLS support that has uncovered some difficulty in navigating the existing layout of manifests and utilization of kustomize to compose the V1 components and install them as a "single entity". Some of the struggles we've encountered are:
- Difficulty in sharing CAs for TLS certificates across namespaces
- Shared namespaces (switched to this pattern to resolve ^) causes some clashing related to labels and selectors, especially for metrics
- Separate component repositories and installation steps results in potentially clashing resources (i.e shared namespace, cert-manager resources, etc.)

To mitigate this, it was discussed offline between a few of the maintainers and determined that it would be worth the effort to re-think our manifest layouts and how we compose the OLMv1 components for installation. The goal of this epic is to capture the need for this work and serve as an area of discussion / planning. 

- [x] #1000
- [x] https://github.com/operator-framework/catalogd/issues/295

tmshort:
As part of this, the consolidation of the ClusterIsser and dependencies need to be consoilidated. Currently, it is being applied via a HEREDOC for the e2e's and a separate YAML file for Tide.

Getting these resources into the kustomize config will fix this issue.

tmshort:
RFC: https://docs.google.com/document/d/1w1IGgE5M40jFIrSuY74E8PnSQn_ihWnjdoFedl6lp8c/edit?usp=sharing

tmshort:
See: https://github.com/operator-framework/operator-controller/pull/970

tmshort:
This is basically done.

PR #969, opened 2024-06-24
:bug:  Fixing the labels for better  metrics collection

LalatenduMohanty:
As the label selector used for both catalogd and operator-controller metrics services is "control-plane: controller-manager". Hence changing the labels in both operator-controller and catalogd to make sure we do not overlap.

# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/config/base/manager/manager.yaml
+++ b/config/base/manager/manager.yaml
@@ -2,7 +2,6 @@ apiVersion: v1
 kind: Namespace
 metadata:
   labels:
-    control-plane: controller-manager
     pod-security.kubernetes.io/enforce: restricted
     pod-security.kubernetes.io/enforce-version: latest
   name: system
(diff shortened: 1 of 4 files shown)


config/base/manager/manager.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 kind: Namespace
 metadata:
   labels:
-    control-plane: controller-manager
+    control-plane: operator-controller-controller-manager
```
joelanford, line 5: I think I would delete this label from the namespace. `catalogd` contributes this same namespace in it's manifest, which means we'll end up with conflicts or silent overwrites.

This is related to #967

tmshort: So, this is a minor change that can be done fairly quickly.

tmshort (commented):
Do we qnt to have a repetitive label value? Would `operator-conroller-manager` make more sense?

config/base/manager/manager.yaml
LalatenduMohanty: [catalogd uses](https://github.com/operator-framework/catalogd/pull/292/files) `catalogd-controller-manager` and not `operator-controller-controller-manager`. So I am little confused here.

everettraven: @LalatenduMohanty I think what @joelanford is saying is that both catalogd and operator-controller would create this particular namespace based on recent manifest changes across both projects

tmshort:
Where are we with this?

tmshort:
Looks to be still failing tests. Have we tried using prometheus with it?

joelanford (approved):

Issue #877, opened 2024-05-21
Transplant pre-flight logic introduced in rukpak to operator-controller


everettraven:
Since we are in progress of integrating the Helm based content management into operator-controller that takes heavy inspiration from rukpak and are initially integrating the CRD Upgrade Safety preflight logic into rukpak (to continue making progress while work is blocked on the Helm integration merging) we will need to transplant the appropriate logic from rukpak to operator-controller when it is unblocked.

PR #979, opened 2024-06-25
✨ Add CRD Upgrade Safety Preflight check to the ClusterExtension controller

rashmigottipati:
# Description

- Adds a new preflight field to the ClusterExtensionSpec 
- Updates the `ClusterExtension` controller to be configured with preflight checks and run them if they're enabled

Fixes https://github.com/operator-framework/operator-controller/issues/877

## Reviewer Checklist

- [ ] API Go Documentation


internal/controllers/clusterextension_controller.go
```diff
@@ -306,12 +323,29 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
...
 	}
 
-	rel, state, err := r.getReleaseState(ac, ext, chrt, values, post)
+	rel, desiredRel, state, err := r.getReleaseState(ac, ext, chrt, values, post)
 	if err != nil {
 		setInstalledStatusConditionFailed(ext, fmt.Sprintf("%s:%v", ocv1alpha1.ReasonErrorGettingReleaseState, err))
 		return ctrl.Result{}, err
 	}
 
+	for _, preflight := range r.Preflights {
```
everettraven, line 332: Looks like we still need to add logic here to skip a preflight check if it has been disabled in the ClusterExtension spec configuration

rashmigottipati: Are you referring to a top level check to not run all of the preflight checks if the check has been disabled in the ClusterExtension spec configuration? 

As of now, the way we have configured it in the ClusterExtension spec, we can disable all or none, and I don't think it's possible to skip individual preflight checks selectively, atleast based on the existing field we have in the API. Please correct me if I am wrong though.

everettraven: We added to the `ClusterExtension` API a method to disable the CRD Upgrade Safety preflight check specifically by creating a `ClusterExtension` that looks something like:
```yaml
apiVersion: olm.operatorframework.io/v1alpha1
kind: ClusterExtension
metadata:
  name: clusterextension-sample
spec:
  installNamespace: default
  packageName: argocd-operator
  version: 0.6.0
  preflight:
    crdUpgradeSafety:
      disabled: true
```

If we identify that the `ClusterExtension` is configured in this way we should ensure we are not running the preflight check if it is the CRD Upgrade Safety preflight check.

I am thinking we could maybe do a check like:
```go
...
if _, ok := preflight.(*crdupgradesafety.Preflight); ok && ext.Spec.Preflight.CRDUpgradeSafety.Disabled {
  // Skip this preflight check because it is of type *crdupgradesafety.Preflight and the CRD Upgrade Safety
  // preflight check has been disabled
  continue
}
...
```
at the top of the for loop

everettraven: Alternatively to that if statement, a switch statement could be utilized and is likely more easily extended than repeating if statements for each preflight check that can be disabled

everettraven: To clarify, when I say preflight check in this context I'm meaning the top level CRD Upgrade Safety preflight check and am not referring to the individual validations that are executed when running the CRD Upgrade Safety preflight check

everettraven (approved):
Changes LGTM - thanks @rashmigottipati !

Would love to see a demo of this functioning in the next community call!

internal/controllers/clusterextension_controller.go
```diff
@@ -306,12 +323,34 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
...
+		if _, ok := preflight.(*crdupgradesafety.Preflight); ok && ext.Spec.Preflight.CRDUpgradeSafety.Disabled {
+			// Skip this preflight check because it is of type *crdupgradesafety.Preflight and the CRD Upgrade Safety
+			// preflight check has been disabled
+			continue
+		}
```
everettraven, line 337: I did some investigation into the CI failures and it looks like it was due to a nil pointer reference. I forgot in my initial recommendation that the `Preflight` and `CRDUpgradeSafety` fields are pointers so I think this should actually be:
```go
if ext.Spec.Preflight != nil {
	if ext.Spec.Preflight.CRDUpgradeSafety != nil {
		if _, ok := preflight.(*crdupgradesafety.Preflight); ok && ext.Spec.Preflight.CRDUpgradeSafety.Disabled {
			// Skip this preflight check because it is of type *crdupgradesafety.Preflight and the CRD Upgrade Safety
			// preflight check has been disabled
			continue
		}
	}
}
```

rashmigottipati: Thanks :grinning: updated the PR

everettraven (approved):

internal/controllers/clusterextension_controller.go
```diff
@@ -304,12 +321,36 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
...
-	rel, state, err := r.getReleaseState(ac, ext, chrt, values, post)
+	rel, desiredRel, state, err := r.getReleaseState(ac, ext, chrt, values, post)
 	if err != nil {
 		setInstalledStatusConditionFailed(ext, fmt.Sprintf("%s:%v", ocv1alpha1.ReasonErrorGettingReleaseState, err))
 		return ctrl.Result{}, err
 	}
 
+	for _, preflight := range r.Preflights {
+		if ext.Spec.Preflight != nil {
+			if _, ok := preflight.(*crdupgradesafety.Preflight); ok && ext.Spec.Preflight.CRDUpgradeSafety.Disabled {
```
everettraven, line 332: Even though it is a required field, the `ext.Spec.Preflight.CRDUpgradeSafety` is a pointer right? Should we add a nil check for this field as well as an extra safeguard to ensure we don't hit a nil pointer error on that field?

rashmigottipati: That's a good idea, updated in [latest commit](https://github.com/operator-framework/operator-controller/pull/979/commits/06914d241169413f7cd299dba11e3208328792bd)

everettraven (approved):

Issue #975, opened 2024-06-25
Wire up dynamic caching layer to manage installed content


everettraven:
Following #974 , we need to wire up the `ClusterExtension` reconciler to utilize the dynamic cache for establishing watches on the resources to be managed for a particular `ClusterExtension`.

**Acceptance Criteria**:
- Utilize the dynamic cache implementation created in #974 for establishing watches on resources to be managed for a given `ClusterExtension`
- When ClusterExtensions are deleted, their managed content caches are stopped and removed
- Updates to unit+e2e tests as necessary
- Addition of an e2e test that validates a manual change to a managed resource is reconciled and reverted

PR #1001, opened 2024-07-01
:sparkles: WIP: Create a Go pkg with a managed content caching layer

### LINKED ISSUE #974: Implement a dynamic caching layer that establishes watches utilizing a ServiceAccount
As part of managing installed content and reconciling any changes made to the managed content, we need to configure watches on the resources managed for a given `ClusterExtension`. After #971, given a `ClusterExtension` and the content to be managed, we can configure clients for establishing watches using the `ServiceAccount` specified in the `ClusterExtension`.

While the exact implementation may vary, the currently proposed approach is as follows:

For each ClusterExtension:
- Uses a similar approach as the helm-operator-plugins [client.RestConfigMapper](https://pkg.go.dev/github.com/operator-framework/helm-operator-plugins@v0.2.1/pkg/client#RestConfigMapper) to generate a [rest.Config](https://pkg.go.dev/k8s.io/client-go/rest#Config) that uses a token from the ServiceAccount provided in ClusterExtension.Spec.ServiceAccount.Name
    - In order to support authentication token refreshing during a watch, we will need to do some further customization to the rest.Config resource. In order to achieve this, we will implement a custom [http.RoundTripper](https://pkg.go.dev/net/http#RoundTripper) that will set the Authorization HTTP header to a valid token retrieved from the “TokenGetter” object. This is a similar approach that is taken by [client-go to utilize a BearerTokenFile](https://github.com/kubernetes/client-go/blob/b9309ac26b168965a720c26e3f5329e563ae0c92/transport/round_trippers.go#L304-L318). The custom http.RoundTripper we create could be a lighter weight wrapper around the default used by client-go that can be configured via the rest.Config.WrapTransport option that uses a [transport.WrapperFunc](https://pkg.go.dev/k8s.io/client-go@v0.30.2/transport#WrapperFunc).
- Configures a new controller-runtime [cache.Cache](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/cache#Cache) with the generated [rest.Config](https://pkg.go.dev/k8s.io/client-go/rest#Config)
    - Note: It may be nice to have a lightweight abstraction on top to simplify the logic within the ClusterExtension reconciler for creating new caches and establishing watches for the managed resources
- Uses the [existing approach](https://github.com/operator-framework/operator-controller/blob/208e584ab6be82d11eb0955b0553adbd2244a9fc/internal/controllers/clusterextension_controller.go#L353-L355) for establishing watches with the [controller.Controller](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.18.4/pkg/controller#Controller) Watch() method
- Starts the cache
- When a ClusterExtension is deleted, it’s cache is stopped and removed

A mermaid graph to visualize the logical flow:
```mermaid
graph TD
    A(ClusterExtension)
    B(Content To Manage)
    C(Managed Content Cache)
    D(RestConfigMapper)
    E(rest.Config)
    F(cache.Cache)
    G(controller.Watch)
    J(ClusterExtension Reconciler)
    L(Watch)

    A -- Provided To --> C
    B -- Provided To --> C
    C -- Fetches Rest Config --> D
    D -- Creates for ServiceAccount --> E
    C -- Creates --> F
    E -- Provided To --> F
    F -- Provided To --> G
    G -- Creates --> L
    L -- Events Trigger --> J
    C -- Establishes Watch With --> G
```

**Acceptance Criteria**:
- A new Go library/type that provided a `ClusterExtension` and Managed Content can establish watches on managed resources using the `ServiceAccount` specified in the `ClusterExtension` spec
    - Will require #971 be completed first. Likely could utilize the implementation of a generic token fetcher from #972 
- Unit tests

everettraven:
# Description
- Adds a new internal package `internal/contentmanager` that has an implementation of a dynamic caching layer for managing watches on content managed by a `ClusterExtension`

Fixes #974 

## Reviewer Checklist

- [ ] API Go Documentation


internal/contentmanager/contentmanager.go
```diff
@@ -0,0 +1,143 @@
...
+
+func (i *instance) ManageContent(ctx context.Context, ctrl controller.Controller, ce *v1alpha1.ClusterExtension, objs []client.Object) error {
+	cfg, err := i.rcm(ctx, ce, i.baseCfg)
+	if err != nil {
+		return fmt.Errorf("getting rest.Config for ClusterExtension %q: %w", ce.Name, err)
+	}
+
+	// TODO: add a http.RoundTripper to the config to ensure it is always using an up
+	// to date authentication token for the ServiceAccount token provided in the ClusterExtension.
+	// Maybe this should be handled by the RestConfigMapper
```
everettraven, line 65: We are thinking this should indeed be done as part of the RestConfigMapper implementation so that everything utilizing it can take advantage of automated token rotation

internal/contentmanager/contentmanager.go
```diff
@@ -0,0 +1,143 @@
...
+					i.mapper,
+					ce,
+				),
+				nil,
+			),
+		)
+		if err != nil {
+			return fmt.Errorf("creating watch for ClusterExtension %q managed resource %s: %w", ce.Name, obj.GetObjectKind().GroupVersionKind(), err)
+		}
+	}
```
everettraven, line 120: I was looking into the controller-runtime code base and since we are using the `ClusterExtension` controller that will already be running I think we need to start the cache _before_ establishing these watches or else this will fail to start the sources successfully.

internal/contentmanager/contentmanager.go
```diff
@@ -0,0 +1,143 @@
...
+		return fmt.Errorf("getting rest.Config for ClusterExtension %q: %w", ce.Name, err)
+	}
+
+	// TODO: add a http.RoundTripper to the config to ensure it is always using an up
+	// to date authentication token for the ServiceAccount token provided in the ClusterExtension.
+	// Maybe this should be handled by the RestConfigMapper
+
+	// Assumptions: all objects received by the function will have the Object metadata specfically,
+	// ApiVersion and Kind set. Failure to which the code will panic when adding the types to the scheme
+	scheme := runtime.NewScheme()
```
everettraven, line 69: When I was playing around with integrating this logic into the controller, I ran into an issue that it couldn't properly configure a handler for the ClusterExtension resource as an owner because the API didn't exist in the scheme. I had to add the following code:
```go
err = v1alpha1.AddToScheme(scheme)
if err != nil {
  return fmt.Errorf("adding ClusterExtension APIs to scheme: %w", err)
}
```

internal/contentmanager/contentmanager.go
```diff
@@ -0,0 +1,143 @@
...
+		err = ctrl.Watch(
+			source.Kind(
+				c,
+				obj,
+				handler.TypedEnqueueRequestForOwner[client.Object](
+					scheme,
+					i.mapper,
+					ce,
+				),
+				nil,
```
everettraven, line 114: This `nil` being passed in here results in a nil pointer dereference. I don't recall why it was put there but it isn't needed and should be removed.

internal/contentmanager/contentmanager.go
```diff
@@ -0,0 +1,143 @@
...
+	if data, ok := i.extensionCaches[ce.GetName()]; ok {
+		data.Cancel()
+	}
+
+	ctx, cancel := context.WithCancel(ctx)
+	go c.Start(ctx)
+	i.extensionCaches[ce.Name] = extensionCacheData{
+		Cache:  c,
+		Cancel: cancel,
+	}
```
everettraven, line 131: As an initial pass, we are thinking to just blow away and recreate a new cache every time this is called. This is definitely not ideal but gives us a starting point for integration and making sure the core logic functions appropriately. We will create a follow up issue to track optimizing the implementation such that the same cache is used to add and remove informers as necessary based on changes to the set of resources managed.

internal/contentmanager/contentmanager.go
everettraven: After playing around with integration this seems to be OK as is

internal/contentmanager/contentmanager.go
```diff
@@ -0,0 +1,203 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	"github.com/operator-framework/operator-controller/api/v1alpha1"
+	oclabels "github.com/operator-framework/operator-controller/internal/labels"
+)
+
+type ContentManager interface {
```
everettraven, line 23: Typically I wouldn't define the interface where it is implemented, but we wanted to have a good starting point without having to wire everything together (scoped for #975) and created this interface based on what we think would be needed by the ClusterExtensionReconciler.

This interface and code will likely get moved around a bit when we wire everything up in #975

internal/contentmanager/contentmanager.go
```diff
@@ -0,0 +1,203 @@
...
+		}
+	}
+
+	// TODO: Instead of stopping the existing cache and replacing it every time
+	// we should stop the informers that are no longer required
+	// and create any new ones as necessary. To keep the initial pass
+	// simple, we are going to keep this as is and optimize in a follow-up.
+	// Doing this in a follow-up gives us the opportunity to verify that this functions
+	// as expected when wired up in the ClusterExtension reconciler before going too deep
+	// in optimizations.
```
everettraven, line 177: This will be done in a follow-up

internal/contentmanager/contentmanager.go
tmshort: I was about to note that it hasn't been wired up.

internal/contentmanager/contentmanager.go
```diff
@@ -0,0 +1,203 @@
...
+	// The ClusterExtension types must be added to the scheme since its
+	// going to be used to establish watches that trigger reconciliation
+	// of the owning ClusterExtension
+	if err := v1alpha1.AddToScheme(scheme); err != nil {
+		return nil, fmt.Errorf("adding operator controller APIs to scheme: %w", err)
+	}
+
+	for _, obj := range objs {
+		gvk := obj.GetObjectKind().GroupVersionKind()
+		listKind := obj.GetObjectKind().GroupVersionKind().Kind + "List"
```
tmshort, line 75:
```suggestion
		listKind := gvk.Kind + "List"
```

internal/contentmanager/contentmanager.go
```diff
@@ -0,0 +1,203 @@
...
+
+		// If the Kind or Version is not set in an object's GroupVersionKind
+		// attempting to add it to the runtime.Scheme will result in a panic.
+		// To avoid panics, we are doing the validation and returning early
+		// with an error if any objects are provided with a missing Kind or Version
+		// field
+		if gvk.Kind == "" {
```
tmshort, line 82: Should this check be before setting `listKind`?

internal/contentmanager/contentmanager.go
```diff
@@ -0,0 +1,203 @@
...
+// to ensure it is only caching and reacting to content that belongs to the ClusterExtension.
+// For each client.Object provided, a new source.Kind is created and used in a call to the Watch() method
+// of the provided controller.Controller to establish new watches for the managed resources.
+func (i *instance) ManageContent(ctx context.Context, ctrl controller.Controller, ce *v1alpha1.ClusterExtension, objs []client.Object) error {
+	// Return nil if the objs slice is empty before any further action
+	if len(objs) == 0 || ce == nil || ctrl == nil {
+		return nil
+	}
+
+	cfg, err := i.rcm(ctx, ce, i.baseCfg)
```
tmshort, line 127: Should this be the updated `ctx`?
i.e. the following from L182 should be here?
```
   ctx, cancel := context.WithCancel(ctx)

**Review by tmshort — COMMENTED** [review]
Just a few comments.

File: internal/contentmanager/contentmanager.go (hunk 6 above)
Comment (joelanford) [reply internal/contentmanager/contentmanager.go:23]: > Typically I wouldn't define the interface where it is implemented

I know this was on advice from Andy, but I haven't gotten used to this pattern at all and I think we may be taking it to an extreme. Looking at the Go standard library, locations of interfaces are fairly intuitive (e.g. `io.Reader` lives in the `io` package where there are lots of methods that accept or return it. `json.Marshaler` lives in the `encoding/json` package, where JSON marshaling/unmarshaling implementations exist.

To me, it is super-unituitive for our `controllers` package to be the place where the InstalledBundleGetter and Preflight interfaces are defined. https://github.com/operator-framework/operator-controller/blob/8947c3de2799b7bd765a4a586b81fd82ea668e2e/internal/controllers/clusterextension_controller.go#L103-L120

File: internal/contentmanager/contentmanager.go [hunk 11]
```diff
@@ -0,0 +1,203 @@
...
+
+	"github.com/operator-framework/operator-controller/api/v1alpha1"
+	oclabels "github.com/operator-framework/operator-controller/internal/labels"
+)
+
+type ContentManager interface {
+	// ManageContent will:
+	// 1. Create a new controller-runtime cache.Cache belonging to the provided ClusterExtension
+	// 2. For each object provided:
+	//   A. Use the provided controller.Controller to establish a watch for the resource
```
joelanford, line 27: This description seems implementation-specific, not interface-generic. Can this be re-written to describe the general purpose of the function and details about contracts that implementations are expected to uphold?

internal/contentmanager/contentmanager.go
```diff
@@ -0,0 +1,203 @@
...
+	if data, ok := i.extensionCaches[ce.GetName()]; ok {
+		data.Cancel()
+	}
+
+	ctx, cancel := context.WithCancel(ctx)
+	go c.Start(ctx) //nolint:errcheck
+	i.extensionCaches[ce.Name] = extensionCacheData{
+		Cache:  c,
+		Cancel: cancel,
+	}
```
joelanford, line 187: Map read and write should be protected by a mutex.

internal/contentmanager/contentmanager.go
```diff
@@ -0,0 +1,203 @@
...
+	"github.com/operator-framework/operator-controller/api/v1alpha1"
+	oclabels "github.com/operator-framework/operator-controller/internal/labels"
+)
+
+type ContentManager interface {
+	// ManageContent will:
+	// 1. Create a new controller-runtime cache.Cache belonging to the provided ClusterExtension
+	// 2. For each object provided:
+	//   A. Use the provided controller.Controller to establish a watch for the resource
+	ManageContent(context.Context, controller.Controller, *v1alpha1.ClusterExtension, []client.Object) error
```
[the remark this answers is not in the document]
joelanford: Or simply `Watch` and `Unwatch`?

Or what if the `ManageContent` equivalent returned an object that had a `Close()` method that could later be called?

internal/contentmanager/contentmanager.go
```diff
@@ -0,0 +1,203 @@
...
+		Scheme:               scheme,
+		DefaultLabelSelector: tgtLabels.AsSelector(),
+	})
+	if err != nil {
+		return fmt.Errorf("creating cache for ClusterExtension %q: %w", ce.Name, err)
+	}
+
+	for _, obj := range objs {
+		err = ctrl.Watch(
+			source.Kind(
```
joelanford, line 156: I know I asked similar questions to what I'm about to ask in the Brief/RFC, but now with a new slant.

We're cancelling the cache after we're done with it, but the `source.Kind`s stick around forever? Is this a memory leak?

internal/contentmanager/contentmanager.go
```diff
@@ -0,0 +1,203 @@
...
+	// simple, we are going to keep this as is and optimize in a follow-up.
+	// Doing this in a follow-up gives us the opportunity to verify that this functions
+	// as expected when wired up in the ClusterExtension reconciler before going too deep
+	// in optimizations.
+	if data, ok := i.extensionCaches[ce.GetName()]; ok {
+		data.Cancel()
+	}
+
+	ctx, cancel := context.WithCancel(ctx)
+	go c.Start(ctx) //nolint:errcheck
```
joelanford, line 183: No error check? What happens when `Start()` returns an error and we don't check it?

internal/contentmanager/contentmanager.go
```diff
@@ -0,0 +1,203 @@
...
+// RemoveManagedContent will stop the cache for the provided ClusterExtension
+// stopping any watches on managed content
+func (i *instance) RemoveManagedContent(ce *v1alpha1.ClusterExtension) {
+	if ce == nil {
+		return
+	}
+
+	if data, ok := i.extensionCaches[ce.GetName()]; ok {
+		data.Cancel()
+		delete(i.extensionCaches, ce.GetName())
```
joelanford, line 201: Another place we need a mutex to protect the map key deletion.

internal/contentmanager/contentmanager.go
everettraven: I think it is more so that the interfaces are defined where they are _used_. That being said, this isn't a hill I'm willing to die on, but it is a pattern I've come to appreciate.

internal/contentmanager/contentmanager.go
everettraven: Actually no, this wouldn't be a memory leak. I just dug a bit more into the controller-runtime internal logic and it looks like sources are temporary:
- When calling `ctrl.Watch()` the source is only stored if the controller has not yet been started, otherwise it's Start function is called and no reference is stored
    - Ref: https://github.com/kubernetes-sigs/controller-runtime/blob/a0c9fd9d3f310f48155ce985366b21914675fbea/pkg/internal/controller/controller.go#L117-L131
- When calling `ctrl.Start()` all sources stored to be started are started and then no longer stored as references so they can be garbage collected when no longer in use
    - Ref: https://github.com/kubernetes-sigs/controller-runtime/blob/a0c9fd9d3f310f48155ce985366b21914675fbea/pkg/internal/controller/controller.go#L171-L212

internal/contentmanager/contentmanager.go
```diff
@@ -0,0 +1,203 @@
...
+	// and create any new ones as necessary. To keep the initial pass
+	// simple, we are going to keep this as is and optimize in a follow-up.
+	// Doing this in a follow-up gives us the opportunity to verify that this functions
+	// as expected when wired up in the ClusterExtension reconciler before going too deep
+	// in optimizations.
+	if data, ok := i.extensionCaches[ce.GetName()]; ok {
+		data.Cancel()
+	}
+
+	ctx, cancel := context.WithCancel(ctx)
```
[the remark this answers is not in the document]
everettraven: See https://github.com/operator-framework/operator-controller/pull/1001#discussion_r1672182549

internal/contentmanager/contentmanager.go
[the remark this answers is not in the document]
everettraven: It should now be changed to `Watch` and `Unwatch`

internal/contentmanager/contentmanager.go
everettraven: Updated the names of the functions and description. Let me know what you think!

internal/contentmanager/contentmanager.go
everettraven: I was scratching my head on how to resolve this problem and couldn't come up with anything because the `Start()` method is blocking. The only thing I could think of is _maybe_ something like:
```go
errChan := make(chan error)
go func(){
  err := c.Start(ctx)
  if err != nil {
    errChan <- err
  }
}()
// handle errors down here
```
But I'm not sure what would happen if the function that created the channel had already exited.

I decided to go the route of not doing an error check here, but waiting for the cache to sync as the next step. If the cache could not sync we cancel the context and return an error. Otherwise the cache has started and we should be good to go.

I'm open to suggestions for better ways to handle this, so if you have some ideas here I'm all ears!

internal/contentmanager/contentmanager.go
everettraven: I could do a new context if folks felt more comfortable with that, but in playing around with moving it I got a notification on my text editor about having to call the `cancel()` function every time we return. Seems like all code paths should use the cancel function or you risk a "context leak".

If we find that we need to move it later it should be a small lift

internal/contentmanager/contentmanager.go
tmshort: You'll end up blocking on the channel, because you don't know exactly when the error will come in, if at all. You should have the error checking, but it should call `something.Cancel()` on error. This may mean you need to put the `Start()` below the addition to `i.extensionCaches[]` below. The `Cancel()` function will remove the item from the caches. This way, `Cancel()` is called regardless.

tmshort:
```suggestion
	i.extensionCaches[ce.Name] := extensionCacheData{
		Cache:  c,
		Cancel: cancel,
	}
	 = dataCache
	go func(){
	    err := c.Start(ctx)
	    if err != nil {
	        cancel()
	    }
	}()
```

everettraven: I may need a synchronous chat to better understand how this should be handled. I haven't had a scenario like this before and my brain is failing me on understanding the approach you are suggesting @tmshort

internal/contentmanager/contentmanager.go
joelanford: `Watch()` is a non-blocking function with respect to the cache running, so it makes sense to use the context passed in from the caller everywhere _except_ for the cache's context that needs to be cancelled.

I do think the cache's context should be derived from `context.Background()` though. It's lifecycle is separate from the lifecycle of the passed in context (which is the context passed into `Reconcile`, which theoretically can be cancelled as soon as `Reconcile` returns). And we wouldn't want that canceling our cache's context.

everettraven: Sounds reasonable to me. Will update it to be derived from `context.Background()`

everettraven: Should be done in https://github.com/operator-framework/operator-controller/pull/1001/commits/0999fb28b4cfb39e0c3804b0ea4683119a8e285c

internal/contentmanager/contentmanager.go
joelanford: That looks promising. Thanks!

I followed that path of `source.Kind.Start()` as well. It looks like that function:
1. Get's an informer from the cache.
2. Adds the specified event hander to it, but notably does _not_ capture the registration struct, which can be used to remove the event handler from the informer later.  Maybe this isn't important to us though because we are probably not adding/removing event handlers from informers. We would be adding/removing entire informers.
3. We stop the entire cache when we're done with it, which should stop all of the informers and event handlers, and free up any referenced resources.

internal/contentmanager/contentmanager.go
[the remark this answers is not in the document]
joelanford: Based on my last comment up there, I'd suggest:
1. Name the new `ctx` something different (e.g. `cacheCtx`) to avoid shadowing the original `ctx`
2. Use `context.Background()` as the base context, not `ctx`.

internal/contentmanager/contentmanager.go
```diff
@@ -0,0 +1,214 @@
...
+
+// Unwatch will stop the cache for the provided ClusterExtension
+// stopping any watches on managed content
+func (i *instance) Unwatch(ce *v1alpha1.ClusterExtension) {
+	if ce == nil {
+		return
+	}
+
+	i.mu.Lock()
+	if data, ok := i.extensionCaches[ce.GetName()]; ok {
```
joelanford, line 209: Nit: better variable name than `data`? Maybe `extensionCache`?

internal/contentmanager/contentmanager.go
joelanford: If we hooked this up as is, what would end up triggering a cache teardown and re-creation? Would it be just when we call `actionclient.Upgrade`? Or would we do it even after `actionclient.Reconcile`?

I think if it is just for an upgrade, we can probably get by because that would only happen when the resolved bundle changes or (eventually when we support it) the user-supplied configuration changes.

everettraven: Hooked up as is, I would assume we only call this whenever the identified state is one of `Install` or `Upgrade`

everettraven: To clarify, by identified state I mean the same state that is used in a few other places in the `reconcile` function of the ClusterExtensionReconciler:
https://github.com/operator-framework/operator-controller/blob/db5a75c0e1e7b78ef98f2a29c2bc5550a509a6d1/internal/controllers/clusterextension_controller.go#L341-L354

tmshort (approved):
/lgtm

PR #1003, opened 2024-07-02
✨ Add upgrade E2E

### LINKED ISSUE #856: [epic] Implement upgrade tests for OperatorController
In order to have confidence that feature additions and bug fixes are reflected after an upgrade, we should add a suite of upgrade tests that:
1. Install the most recent public release.
2. Create objects that sufficiently exercise our API surface and behaviors.
3. Upgrade in place from the most recent public release to a build from the current commit.
4. Make assertions that prove that changes from the current commit are reflected.

m1kola:
# Description

Testing upgrade from the latest release to the current commit.

Fixes #856

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -159,6 +159,24 @@ extension-developer-e2e: KUSTOMIZE_BUILD_DIR := config/overlays/cert-manager
 extension-developer-e2e: KIND_CLUSTER_NAME := operator-controller-ext-dev-e2e  #EXHELP Run extension-developer e2e on local kind cluster
 extension-developer-e2e: run image-registry test-ext-dev-e2e kind-clean
 
+.PHONY: run-latest-release
+run-latest-release:
+	curl -L -s https://github.com/operator-framework/operator-controller/releases/latest/download/install.sh | bash -s
+
+.PHONY: pre-upgrade-setup
+pre-upgrade-setup:
+	./hack/pre-upgrade-setup.sh $(CATALOG_IMG) $(TEST_CLUSTER_CATALOG_NAME) $(TEST_CLUSTER_EXTENSION_NAME)
+
+.PHONY: post-upgrade-checks
+post-upgrade-checks:
+	go test -count=1 -v ./test/upgrade-e2e/...
+
+.PHONY: test-upgrade-e2e
+test-upgrade-e2e: KIND_CLUSTER_NAME := operator-controller-upgrade-e2e
+test-upgrade-e2e: export TEST_CLUSTER_CATALOG_NAME := test-catalog
+test-upgrade-e2e: export TEST_CLUSTER_EXTENSION_NAME := test-package
+test-upgrade-e2e: kind-cluster run-latest-release image-registry build-push-e2e-catalog registry-load-bundles pre-upgrade-setup docker-build kind-load kind-deploy post-upgrade-checks kind-clean #HELP Run upgrade e2e tests on a local kind cluster
+
 .PHONY: e2e-coverage
 e2e-coverage:
 	COVERAGE_OUTPUT=./e2e-cover.out ./hack/e2e-coverage.sh
--- a/hack/pre-upgrade-setup.sh
+++ b/hack/pre-upgrade-setup.sh
@@ -1,0 +1,51 @@
+#!/bin/bash
+
+set -euo pipefail
+
+help="pre-upgrade-setup.sh is used to create some basic resources
+which will later be used in upgrade testing.
+
+Usage:
+  post-upgrade-checks.sh [TEST_CATALOG_IMG] [TEST_CATALOG_NAME] [TEST_CLUSTER_EXTENSION_NAME]
+"
+
+if [[ "$#" -ne 3 ]]; then
+  echo "Illegal number of arguments passed"
+  echo "${help}"
+  exit 1
+fi
+
+TEST_CATALOG_IMG=$1
+TEST_CLUSTER_CATALOG_NAME=$2
+TEST_CLUSTER_EXTENSION_NAME=$3
+
+kubectl apply -f - << EOF
+apiVersion: catalogd.operatorframework.io/v1alpha1
+kind: ClusterCatalog
+metadata:
+  name: ${TEST_CLUSTER_CATALOG_NAME}
+spec:
+  source:
+    type: image
+    image:
+      ref: ${TEST_CATALOG_IMG}
+      pollInterval: 24h
+      insecureSkipTLSVerify: true
+EOF
+
+
+kubectl apply -f - << EOF
+apiVersion: olm.operatorframework.io/v1alpha1
+kind: ClusterExtension
+metadata:
+  name: ${TEST_CLUSTER_EXTENSION_NAME}
+spec:
+  installNamespace: default
+  packageName: prometheus
+  version: 1.0.0
+  serviceAccount:
+    name: default
+EOF
+
+kubectl wait --for=condition=Unpacked --timeout=60s ClusterCatalog $TEST_CLUSTER_CATALOG_NAME
+kubectl wait --for=condition=Installed --timeout=60s ClusterExtension $TEST_CLUSTER_EXTENSION_NAME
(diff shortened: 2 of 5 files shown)


tmshort:
My general question is... do we want to use `bash` scripts for this, or do we want to use golang?

Makefile
```diff
@@ -158,6 +158,24 @@ test-e2e: run image-registry build-push-e2e-catalog registry-load-bundles e2e e2
...
+.PHONY: run-latest-release
+run-latest-release:
+	curl -L -s https://github.com/operator-framework/operator-controller/releases/latest/download/install.sh | bash -s
+
+.PHONY: pre-upgrade-setup
+pre-upgrade-setup:
+	./hack/pre-upgrade-setup.sh $(CATALOG_IMG) $(TEST_CLUSTER_CATALOG_NAME) $(TEST_CLUSTER_EXTENSION_NAME)
+
+.PHONY: post-upgrade-checks
+post-upgrade-checks:
```
joelanford, line 170: Can we also run the standard e2e after an upgrade as well?

hack/post-upgrade-checks.sh
```diff
@@ -0,0 +1,22 @@
...
+  echo "${help}"
+  exit 1
+fi
+
+TEST_CLUSTER_CATALOG_NAME=$1
+TEST_CLUSTER_EXTENSION_NAME=$2
+
+kubectl wait --for=condition=Available --timeout=60s -n olmv1-system deployment --all
+kubectl wait --for=condition=Unpacked --timeout=60s ClusterCatalog $TEST_CLUSTER_CATALOG_NAME
+kubectl wait --for=condition=Installed --timeout=60s ClusterExtension $TEST_CLUSTER_EXTENSION_NAME
```
joelanford, line 22: +1 on checking that the existing ClusterCatalog and ClusterExtension remain Unpacked and Installed. But two questions:
1. My instinct is that these waits might not actually check anything. We've already waited for them to be Unpacked and Installed prior to the upgrade, so they will still be Unpacked and Installed after the upgrade _at least_ until catalogd and operator-controller have reconciled them. Seems like we have a race condition here where these commands will execute before the upgraded reconcilers have finished reconciling them.
2. I think we should probably do a few extra tasks post-upgrade
   a. Change the ClusterExtension to specify a broader version range so that it finds and upgrades to a new version from the existing catalog
   b. Add a new ClusterExtension that installs another package from the existing ClusterCatalog


Also, potentially out-of-scope in this PR, but in the catalogd-specific upgrade tests, I think we'll want a scenario where the catalog image referenced by a ClusterCatalog receives an update after the upgrade, which would ensure that our polling logic doesn't break. That scenario may be relevant in operator-controller because we expect catalog updates to trigger reconciles (and potentially upgrade) ClusterExtension objects.

joelanford:
> Do we want to use bash scripts for this, or do we want to use golang?

This is a good question. I don't necessarily think its an either-or thing. I could see something like:
```sh
sh <installPreviousRelease>
sh <setupPreUpgradeState>
sh <upgradeIt>
go test ./test/upgrade-e2e/...
```

joelanford:
Another interesting upgrade question: At what point do we need to actually define an upgrade process? An upgrade is generally:
1. Create objects that are net new in the new release
2. Update objects that are changed from the old release to the new release
3. Delete objects that were present in the old release, but not the new release

We don't do step 3 right now, right? Is doing step 3 a prereq for valid upgrade testing?

m1kola:
> Another interesting upgrade question: At what point do we need to actually define an upgrade process? An upgrade is generally:
> 
> 1. Create objects that are net new in the new release
> 2. Update objects that are changed from the old release to the new release
> 3. Delete objects that were present in the old release, but not the new release
> 
> We don't do step 3 right now, right? Is doing step 3 a prereq for valid upgrade testing?

We were talking about this with @ankitathomas. It sounds a bit like we are testing something that doesn't exist at this moment (the upgrade process).

We decided not to open a can of worms for now and assume that the upgrade process is just:
* For operator-controller - run the install script on top of what already exists
* For catalogd - apply manifests on top of what already exists

This should be enough short-term: it should signal when someone adds something breaking to manifests/install script.

But long term we probably want to define upgrade process and probably create tool for that. E.g. we will likely need to some migration tool where it is possible to clean up old in-cluster objects from previous releases. We can use OpenShift's Cluster Version Operator and [how it handles deletions](https://github.com/openshift/enhancements/blob/9ea87a8b393a5051c7e54dd1b414ed184a574414/dev-guide/cluster-version-operator/dev/object-deletion.md).

But IMO - this deserves its own epic. What do you think? Should we put this on hold and define upgrade process & create necessary tools first? Or should we proceed with this naive approach of just applying things on top?



> > Do we want to use bash scripts for this, or do we want to use golang?
> 
> This is a good question. I don't necessarily think its an either-or thing. I could see something like:
> 
> ```shell
> sh <installPreviousRelease>
> sh <setupPreUpgradeState>
> sh <upgradeIt>
> go test ./test/upgrade-e2e/...
> ```

I decided to keep bash in the draft for now. We can switch to Go when/if we decide to do something more sophisticated here.

hack/post-upgrade-checks.sh
m1kola: > Are you suggesting to use "a" to trigger reconciliation? A workaround to my above question?

Note to self: `lastTransitionTime` won't change even if we bump a version. But we can check `observedGeneration`.

m1kola: I updated the scripts to initially install version `1.0.0` and then upgrade to `1.0.1` & added a check to make sure that the reconciliation actually works by checking installed contain again & its `observedGeneration`.

joelanford: I do think it would be better if we could
1. Wait for reconcile of the pre-upgrade state of the ClusterExtension before we modify it.
2. Make some assertions
3. Change the ClusterExtension
4. Make more assertions

To do (1), we can check `operator-controller` logs, which should have enough info to key off of.

joelanford:
Already have an epic for proper upgrade support. I agree that this epic should not be blocked on that one.

I think naive approach is good enough for now, but I suspect there could be some extreme edge cases where leaving around old manifests causes the upgrade tests to pass when cleaning them up would have made them fail. As an example: a PR that deletes our issuer `Certificate` from our manifest with no other changes.

m1kola:
> I suspect there could be some cases where leaving around old manifests causes the upgrade tests to pass when cleaning them up would have made them fail. As an example: a PR that deletes our issuer `Certificate` from our manifest with no other changes.

Yes, there will be cases like that, but in this specific example, I think, we will catch the issue in the E2E job because a clean setup will fail.

hack/post-upgrade-checks.sh
m1kola: @joelanford could you please take a another look? Should be nearly there.

joelanford (commented):
Anything else you're thinking of doing before taking this out of draft? At a quick glance, I think this is looking really good!

m1kola:
> Anything else you're thinking of doing before taking this out of draft? At a quick glance, I think this is looking really good!

@joelanford I just wanted to take another fresh look in the morning and clean some things up. Nothing major: adjusting test logs for consistency, a bit of comment changes and moving variables around. [Here](https://github.com/operator-framework/operator-controller/compare/fe28a5bd28022d36b7bb6feb7896893ef79967ae..2cf7d2e3892410aca66fbec59a336061b99696fc) is the diff between the version you looked at and the latest state.

Should be good now. Marking as ready for review.

hack/pre-upgrade-setup.sh
tmshort, line 1: Would it make sense to put this under the `test/tools` directory? There are a few other testing scripts there, we might want to consolidate?

tmshort (approved):
/lgtm
Just one question on the location of a file... it doesn't necessarily stop this from being merged.

hack/pre-upgrade-setup.sh
tmshort: This could be done later (if we desire)

PR #1004, opened 2024-07-02
🌱 Remove `kind-redeploy` make target

m1kola:
# Description

This target is broken because `OPERATOR_CONTROLLER_NAMESPACE` does not exist.

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -172,11 +172,6 @@ kind-deploy: manifests $(KUSTOMIZE) #EXHELP Install controller and dependencies 
 	$(KUSTOMIZE) build $(KUSTOMIZE_BUILD_DIR) > operator-controller.yaml
 	envsubst '$$CATALOGD_VERSION,$$CERT_MGR_VERSION,$$MANIFEST' < scripts/install.tpl.sh | bash -s
 
-.PHONY: kind-redeploy
-kind-redeploy: generate docker-build kind-load kind-deploy #EXHELP Redeploy newly built executables
-	kubectl delete pod -l control-plane=controller-manager -n $(OPERATOR_CONTROLLER_NAMESPACE)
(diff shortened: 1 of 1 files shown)


joelanford:
I noticed this was broken recently as well. But I know @tmshort added it fairly recently, so maybe worth getting his opinion on this PR.

bentito:
> I noticed this was broken recently as well. But I know @tmshort added it fairly recently, so maybe worth getting his opinion on this PR.

Uh.. it looks like the line got sucked out the airlock in b08a1e92 Should we just put `OPERATOR_CONTROLLER_NAMESPACE ?= operator-controller-system` and fix the target?

m1kola:
> > I noticed this was broken recently as well. But I know @tmshort added it fairly recently, so maybe worth getting his opinion on this PR.
> 
> Uh.. it looks like the line got sucked out the airlock in [b08a1e9](https://github.com/operator-framework/operator-controller/commit/b08a1e927ee37d17de998264fc50ce927790b574) Should we just put `OPERATOR_CONTROLLER_NAMESPACE ?= operator-controller-system` and fix the target?

I'm not opposed to doing this, but my preference is to remove it. IMO our `Makefile` is getting a bit cluttered. Also we have tilt setup for quick redeployment which watches filesystem and automates dev process. We can also run it locally with port forwarding or just `make run` (which is a bit slower and does not persist resources because the kind cluster gets re-created).

But it is just me. If people find it useful for their workflows - then it might be worth keeping it.

tmshort:
This is something that @joelanford had asked for, and I think I noticed that `OPERATOR_CONTROLLER_NAMESPACE` was not being properly set, either, but it got away from me with other work. I don't have a Tilt setup yet, so, I'm one of those where this might  be useful.

tmshort:
Added #1032 to actually fix `kind-redeploy`, but it's really just running other targets...

joelanford:
I'm +1 on removing now that Tilt is actually working properly.

joelanford (approved):

PR #1006, opened 2024-07-02
✨ Adding a token getter to get service account tokens

### LINKED ISSUE #972: Implement a Go struct for fetching and caching authentication tokens for a ServiceAccount
In order to utilize a provided `ServiceAccount` to install and manage content, operator-controller needs to be capable of fetching an authentication token for a given `ServiceAccount`. Once retrieved, this token can then be used to create clients that authenticate as the `ServiceAccount` to perform operations on the cluster.

The scope of this issue is limited to specifically implementing a Go type to fetch and cache these authentication tokens and does not include utilizing the token retrieved.

While the exact implementation may vary, here are some things to consider:
- Use the `TokenRequest` API to fetch an authentication token for a provided `ServiceAccount`
- Cache the authentication token so that when subsequent requests for the same `ServiceAccount` are made we can return a valid token without hitting the Kube API server (limits load on Kube API server)
- Token rotation. Tokens are not always long-lived and may require rotation, especially when caching tokens. If we identify that the token in the cache is no longer valid, another request for a valid token should be made.

A high-level overview of what the logic flow could look like:
```mermaid
graph LR
  A(ServiceAccount)
  B(TokenGetter)
  C(TokenRequest)
  D(Token)
  E(Token Cache)
  F{In Cache?}
  G{Expired?}
  
  A -- Provided To --> B
  B --> F
  F -- Yes --> G
  F -- No --> C
  G -- Yes --> C
  G -- No --> E
  C -- Returns --> D
  E -- Returns --> D
  D -- Stored in --> E
```

**Acceptance Criteria**:
- A new Go library/type is created that:
    - Fetches authentication tokens for a provided `ServiceAccount`
    - Caches tokens
    - Rotates cached tokens when no longer valid
- Unit tests

skattoju:
# Description

As part of #737 user provided service accounts will be used to manage cluster extensions, the token getter implemented in this PR enables this by providing an interface to fetch tokens given a service account. It also implements a cache where fetched tokens are stored saving requests to the api server for subsequent requests. Fixes #972 

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/authentication/tokengetter.go
+++ b/internal/authentication/tokengetter.go
@@ -1,0 +1,111 @@
+package authentication
+
+import (
+	"context"
+	"sync"
+	"time"
+
+	authenticationv1 "k8s.io/api/authentication/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/utils/ptr"
+)
+
+type TokenGetter struct {
+	client                     corev1.ServiceAccountsGetter
+	expirationDuration         time.Duration
+	removeAfterExpiredDuration time.Duration
+	tokens                     map[types.NamespacedName]*authenticationv1.TokenRequestStatus
+	mu                         sync.RWMutex
+}
+
+type TokenGetterOption func(*TokenGetter)
+
+const (
+	RotationThresholdPercentage       = 10
+	DefaultExpirationDuration         = 5 * time.Minute
+	DefaultRemoveAfterExpiredDuration = 90 * time.Minute
+)
+
+// Returns a token getter that can fetch tokens given a service account.
+// The token getter also caches tokens which helps reduce the number of requests to the API Server.
+// In case a cached token is expiring a fresh token is created.
+func NewTokenGetter(client corev1.ServiceAccountsGetter, options ...TokenGetterOption) *TokenGetter {
+	tokenGetter := &TokenGetter{
+		client:                     client,
+		expirationDuration:         DefaultExpirationDuration,
+		removeAfterExpiredDuration: DefaultRemoveAfterExpiredDuration,
+		tokens:                     map[types.NamespacedName]*authenticationv1.TokenRequestStatus{},
+	}
+
+	for _, opt := range options {
+		opt(tokenGetter)
+	}
+
+	return tokenGetter
+}
+
+func WithExpirationDuration(expirationDuration time.Duration) TokenGetterOption {
+	return func(tg *TokenGetter) {
+		tg.expirationDuration = expirationDuration
+	}
+}
+
+func WithRemoveAfterExpiredDuration(removeAfterExpiredDuration time.Duration) TokenGetterOption {
+	return func(tg *TokenGetter) {
+		tg.removeAfterExpiredDuration = removeAfterExpiredDuration
+	}
+}
+
+// Get returns a token from the cache if available and not expiring, otherwise creates a new token
+func (t *TokenGetter) Get(ctx context.Context, key types.NamespacedName) (string, error) {
+	t.mu.RLock()
+	token, ok := t.tokens[key]
+	t.mu.RUnlock()
+
+	expireTime := time.Time{}
+	if ok {
+		expireTime = token.ExpirationTimestamp.Time
+	}
+
+	// Create a new token if the cached token expires within DurationPercentage of expirationDuration from now
+	rotationThresholdAfterNow := metav1.Now().Add(t.expirationDuration * (RotationThresholdPercentage / 100))
+	if expireTime.Before(rotationThresholdAfterNow) {
+		var err error
+		token, err = t.getToken(ctx, key)
+		if err != nil {
+			return "", err
+		}
+		t.mu.Lock()
+		t.tokens[key] = token
+		t.mu.Unlock()
+	}
+
+	// Delete tokens that have been expired for more than ExpiredDuration
+	t.reapExpiredTokens(t.removeAfterExpiredDuration)
+
+	return token.Token, nil
+}
+
+func (t *TokenGetter) getToken(ctx context.Context, key types.NamespacedName) (*authenticationv1.TokenRequestStatus, error) {
+	req, err := t.client.ServiceAccounts(key.Namespace).CreateToken(ctx,
+		key.Name,
+		&authenticationv1.TokenRequest{
+			Spec: authenticationv1.TokenRequestSpec{ExpirationSeconds: ptr.To[int64](int64(t.expirationDuration))},
+		}, metav1.CreateOptions{})
+	if err != nil {
+		return nil, err
+	}
+	return &req.Status, nil
+}
+
+func (t *TokenGetter) reapExpiredTokens(expiredDuration time.Duration) {
+	t.mu.Lock()
+	defer t.mu.Unlock()
+	for key, token := range t.tokens {
+		if metav1.Now().Sub(token.ExpirationTimestamp.Time) > expiredDuration {
+			delete(t.tokens, key)
+		}
+	}
+}
--- a/internal/authentication/tokengetter_test.go
+++ b/internal/authentication/tokengetter_test.go
@@ -1,0 +1,88 @@
+package authentication
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	authenticationv1 "k8s.io/api/authentication/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes/fake"
+	ctest "k8s.io/client-go/testing"
+)
+
+func TestTokenGetterGet(t *testing.T) {
+	fakeClient := fake.NewSimpleClientset()
+	fakeClient.PrependReactor("create", "serviceaccounts/token",
+		func(action ctest.Action) (bool, runtime.Object, error) {
+			act, ok := action.(ctest.CreateActionImpl)
+			if !ok {
+				return false, nil, nil
+			}
+			tokenRequest := act.GetObject().(*authenticationv1.TokenRequest)
+			var err error
+			if act.Name == "test-service-account-1" {
+				tokenRequest.Status = authenticationv1.TokenRequestStatus{
+					Token:               "test-token-1",
+					ExpirationTimestamp: metav1.NewTime(metav1.Now().Add(DefaultExpirationDuration)),
+				}
+			}
+			if act.Name == "test-service-account-2" {
+				tokenRequest.Status = authenticationv1.TokenRequestStatus{
+					Token:               "test-token-2",
+					ExpirationTimestamp: metav1.NewTime(metav1.Now().Add(1 * time.Second)),
+				}
+			}
+			if act.Name == "test-service-account-3" {
+				tokenRequest.Status = authenticationv1.TokenRequestStatus{
+					Token:               "test-token-3",
+					ExpirationTimestamp: metav1.NewTime(metav1.Now().Add(-DefaultRemoveAfterExpiredDuration)),
+				}
+			}
+			if act.Name == "test-service-account-4" {
+				tokenRequest = nil
+				err = fmt.Errorf("error when fetching token")
+			}
+			return true, tokenRequest, err
+		})
+
+	tg := NewTokenGetter(fakeClient.CoreV1(),
+		WithExpirationDuration(DefaultExpirationDuration),
+		WithRemoveAfterExpiredDuration(DefaultRemoveAfterExpiredDuration))
+
+	tests := []struct {
+		testName           string
+		serviceAccountName string
+		namespace          string
+		want               string
+		errorMsg           string
+	}{
+		{"Testing getting token with fake client", "test-service-account-1",
+			"test-namespace-1", "test-token-1", "failed to get token"},
+		{"Testing getting token from cache", "test-service-account-1",
+			"test-namespace-1", "test-token-1", "failed to get token"},
+		{"Testing getting short lived token from fake client", "test-service-account-2",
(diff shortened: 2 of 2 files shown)


internal/authentication/tokengetter_test.go
```diff
@@ -0,0 +1,106 @@
...
+	"github.com/stretchr/testify/assert"
+	authenticationv1 "k8s.io/api/authentication/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes/fake"
+	ctest "k8s.io/client-go/testing"
+)
+
+func TestNewTokenGetter(t *testing.T) {
```
everettraven, line 18: Looking through the steps of the test I think the logic looks good overall. One thing I will note is that most of the tests we write for operator-controller end up being table driven tests that end up sharing the same test logic where possible. I think this test in particular could be a bit more readable and maintainable utilizing the table driven approach.

I don't think it is absolutely necessary if folks feel strongly about not requiring that change, but something I would personally like to see to stay consistent with many of the other unit tests we have throughout operator-controller

everettraven (changes requested):
Overall I think this looks really good, great work! There are a few smaller changes I'd recommend

internal/authentication/tokengetter.go
```diff
@@ -0,0 +1,99 @@
...
+	t.mu.RLock()
+	token, ok := t.tokens[key]
+	t.mu.RUnlock()
+
+	expireTime := time.Time{}
+	if ok {
+		expireTime = token.ExpirationTimestamp.Time
+	}
+
+	fiveMinutesAfterNow := metav1.Now().Add(5 * time.Minute)
```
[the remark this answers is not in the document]
skattoju: I forgot about that.. That should be set to expiration seconds. The options pattern is a good idea, thanks for pointing it out.

internal/authentication/tokengetter.go
```diff
@@ -0,0 +1,99 @@
...
+		t.mu.Lock()
+		t.tokens[key] = token
+		t.mu.Unlock()
+	}
+
+	return token.Token, nil
+}
+
+func (t *TokenGetter) getToken(ctx context.Context, key types.NamespacedName) (*authv1.TokenRequestStatus, error) {
+	req, err := t.client.ServiceAccounts(key.Namespace).CreateToken(ctx, key.Name, &authv1.TokenRequest{Spec: authv1.TokenRequestSpec{ExpirationSeconds: ptr.To[int64](3600)}}, metav1.CreateOptions{})
```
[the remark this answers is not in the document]
skattoju: Also missed this, this should also be set to expiration seconds will fix 👍

internal/authentication/tokengetter_test.go
skattoju: Ill change it to a table driven style 👍

internal/authentication/tokengetter.go
```diff
@@ -0,0 +1,117 @@
...
+	k.mu.Lock()
+	defer k.mu.Unlock()
+
+	lock, ok := k.locks[key]
+	if !ok {
+		lock = &sync.Mutex{}
+		k.locks[key] = lock
+	}
+	return lock
+}
```
joelanford, line 76: The number of locks will grow unbounded here, right? Maybe this is unnecessary complexity to start off with. WDYT about skipping the keyLock altogether. For now, maybe it's fine to get a global lock in the `Get` method. That would allow only one token to be requested at a time. If that becomes a problem, we can introduce key locks later.

everettraven: Looking back through this a second time I was wondering the same thing. I get _why_ the key lock was created but agree that this is probably more complex than necessary for a first pass

internal/authentication/tokengetter.go
```diff
@@ -0,0 +1,117 @@
...
+
+type TokenGetter struct {
+	client            corev1.ServiceAccountsGetter
+	expirationSeconds int64
+	tokens            map[types.NamespacedName]*authenticationv1.TokenRequestStatus
+	tokenLocks        keyLock[types.NamespacedName]
+	mu                sync.RWMutex
+}
+
+type TokenGetterOption func(*TokenGetter)
```
everettraven, line 23: When introducing the option pattern to change unexported fields, we should provide some already created option functions to set those values. For example we might add a function called `WithExpirationSeconds` that could be used to set the unexported expirationSeconds field on the TokenGetter

everettraven: Whoops! I commented this before I realized you had already done that - sorry!

internal/authentication/tokengetter.go
```diff
@@ -0,0 +1,117 @@
...
+	t.mu.RLock()
+	token, ok := t.tokens[key]
+	t.mu.RUnlock()
+
+	expireTime := time.Time{}
+	if ok {
+		expireTime = token.ExpirationTimestamp.Time
+	}
+
+	expirationSecondsAfterNow := metav1.Now().Add(time.Duration(t.expirationSeconds))
```
everettraven, line 92: Using the expiration seconds field here feels unintuitive to me. Maybe we should add another field to the struct called something like `rotationThreshold` that is a `time.Duration` type?

joelanford: Or maybe we make it some fraction of `expirationSeconds`?

internal/authentication/tokengetter.go
skattoju: ok ill remove the key locks for now

internal/authentication/tokengetter.go
skattoju: yeah i guess setting that to expirationSeconds may be too much.. I will add a rotationThreshold so its configurable..

skattoju: So if expirationSeconds is 100s and we set rotationThreshold to 10s we we would create a new token if its been more than 90s since the token was created like you suggested 👍

internal/authentication/tokengetter.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+func WithRotationThreshold(rotationThreshold time.Duration) TokenGetterOption {
+	return func(tg *TokenGetter) {
+		tg.rotationThreshold = rotationThreshold
+	}
+}
+
+// Get returns a token from the cache if available and not expiring, otherwise creates a new token
+func (t *TokenGetter) Get(ctx context.Context, key types.NamespacedName) (string, error) {
```
everettraven, line 56: Apologies for not noticing this earlier in the review, but Joe's earlier comment on the tokenLocks potentially growing unbounded made me realize that the tokens themselves could potentially grow unbounded with the current implementation since we only have a `Get` function.

What do you think about adding a method that could be used to "prune" expired tokens from the cache? 

I'm thinking this would be a method used as needed by a caller. I could imagine when we wire everything together we may want to ensure we prune any expired tokens that are older than X amount of time

everettraven: @joelanford What do you think?

skattoju: It appears that theres no way to invalidate tokens issued by the TokenRequest api 🤔 but for our purposes i guess we just remove keys i.e. namespaced names of the service accounts from the token getter cache

everettraven: Yeah, I'm thinking that it could simply be a process that runs every X amount of time and runs the method to ensure we aren't storing expired tokens for an unreasonable amount of time and adding bloat to the memory usage

everettraven: To clarify, I'm only recommending we implement the method that will delete the keys and tokens from the map that are expired and meet a certain criteria defined by the caller like expired for more than 2 hours

internal/authentication/tokengetter.go
```diff
@@ -0,0 +1,105 @@
...
+		client:            client,
+		expirationSeconds: int64(5 * time.Minute), // default token ttl
+		rotationThreshold: 1 * time.Minute,        // default rotation threshold
+		tokens:            map[types.NamespacedName]*authenticationv1.TokenRequestStatus{},
+	}
+
+	for _, opt := range options {
+		opt(tokenGetter)
+	}
+
```
joelanford, line 39: If we allow expiration seconds and rotation threshold to be independently configurable, we probably need to do some sanity validation to make sure they make sense.

But I'm not sure we actually need rotation threshold to be configurable if we just make it `expirationDuration * 0.1` or whatever we think is a reasonable ratio. I feel like we may be over-engineering this.

internal/authentication/tokengetter.go
```diff
@@ -0,0 +1,105 @@
...
+type TokenGetterOption func(*TokenGetter)
+
+// Returns a token getter that can fetch tokens given a service account.
+// The token getter also caches tokens which helps reduce the number of requests to the API Server.
+// In case a cached token is expiring a fresh token is created.
+func NewTokenGetter(client corev1.ServiceAccountsGetter, options ...TokenGetterOption) *TokenGetter {
+	tokenGetter := &TokenGetter{
+		client:            client,
+		expirationSeconds: int64(5 * time.Minute), // default token ttl
```
joelanford, line 31: We should change this to a `time.Duration` type. The name ("Seconds" suffix) and value (integer number of nanoseconds in 5 minutes) lead a reader to an incorrect conclusion.

internal/authentication/tokengetter.go
skattoju: Thinking about this a bit more, we would only remove entries from the cache when extensions are removed, in case there are expired tokens they would belong to service accounts of currently installed extensions. Further more the expired tokens will be replaced by fresh ones whenever an attempt is made to use the tokens so in theory tokens should not accumulate 🤔

joelanford: Do we need the complexity of a background goroutine? What if we just write a `reapExpiredTokens` function and call it every time `Get` is called?

internal/authentication/tokengetter.go
```diff
@@ -0,0 +1,105 @@
...
+	defer t.mu.RUnlock()
+	delete(t.tokens, key)
+}
+
+func (t *TokenGetter) TokenExists(ctx context.Context, key types.NamespacedName) bool {
+	t.mu.RLock()
+	defer t.mu.RUnlock()
+	_, ok := t.tokens[key]
+	return ok
+}
```
joelanford, line 105: How do we envision `Clean` and `TokenExists` being used?

skattoju: Clean would be to remove entries from the cache when extensions are uninstalled or unused items need to be cleaned up for some other reason. TokenExists is debatable, as I added to make it testable.

internal/authentication/tokengetter.go
skattoju: I guess that was int64 because the TokenRequest api takes an pointer to an int64.. its better as a time.Duration 👍

internal/authentication/tokengetter.go
skattoju: That would make it simpler. Ive added a `reapExpiredTokens` This also eliminates the need for `Clean` and `TokenExists`

internal/authentication/tokengetter.go
skattoju: These are made redundant by `reapExpiredTokens` so I have removed them

internal/authentication/tokengetter.go
```diff
@@ -0,0 +1,111 @@
...
+	return &req.Status, nil
+}
+
+func (t *TokenGetter) reapExpiredTokens(expiredDuration time.Duration) {
+	t.mu.Lock()
+	defer t.mu.Unlock()
+	for key, token := range t.tokens {
+		if metav1.Now().Sub(token.ExpirationTimestamp.Time) > expiredDuration {
+			delete(t.tokens, key)
+		}
```
joelanford, line 109: Why would we keep expired tokens around? Seems like they could be reaped as soon as they expire.

internal/authentication/tokengetter.go
```diff
@@ -0,0 +1,111 @@
...
+	token, ok := t.tokens[key]
+	t.mu.RUnlock()
+
+	expireTime := time.Time{}
+	if ok {
+		expireTime = token.ExpirationTimestamp.Time
+	}
+
+	// Create a new token if the cached token expires within DurationPercentage of expirationDuration from now
+	rotationThresholdAfterNow := metav1.Now().Add(t.expirationDuration * (RotationThresholdPercentage / 100))
```
joelanford, line 73: Will `rotationThresholdAfterNow` always evaluate to `0` when `RotationThresholdPercentage < 100`?  Seems like `(RotationThresholdPercentage / 100)` is an integer division.

RotationThresholdPercentage should probably be a `float64` so that we can do something like this:
```suggestion
	rotationThresholdAfterNow := metav1.Now().Add(time.Duration(float64(t.expirationDuration) * RotationThresholdPercentage)
```

joelanford: also, any reason for `RotationThresholdPercentage` to be exported? Seems like we could unexport it.

internal/authentication/tokengetter.go
```diff
@@ -0,0 +1,111 @@
...
+	t.reapExpiredTokens(t.removeAfterExpiredDuration)
+
+	return token.Token, nil
+}
+
+func (t *TokenGetter) getToken(ctx context.Context, key types.NamespacedName) (*authenticationv1.TokenRequestStatus, error) {
+	req, err := t.client.ServiceAccounts(key.Namespace).CreateToken(ctx,
+		key.Name,
+		&authenticationv1.TokenRequest{
+			Spec: authenticationv1.TokenRequestSpec{ExpirationSeconds: ptr.To[int64](int64(t.expirationDuration))},
```
joelanford, line 95: `t.expirationDuration` needs to be converted to an integer number of seconds to be used in the TokenRequest API.

joelanford:
```suggestion
			Spec: authenticationv1.TokenRequestSpec{ExpirationSeconds: ptr.To[int64](int64(t.expirationDuration/time.Second))},
```

internal/authentication/tokengetter.go
skattoju: Now that you mention it, it seems like we can get rid of that expiredDuration parameter 🤔

PR #1007, opened 2024-07-02
✨ Cleaner Condition Types & Reasons

bentito:
# Description

Cleaner Condition Types & Reasons. See commit messages for coverage of this PR and its sync to the RFC recommendations. https://docs.google.com/document/d/1JWJxnDXM0X1JQ67ZIDx5j1ayb1d7ajrK_uSDDaz0G98

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/api/v1alpha1/clusterextension_types.go
+++ b/api/v1alpha1/clusterextension_types.go
@@ -128,15 +128,13 @@ const (
 	ReasonErrorGettingClient = "ErrorGettingClient"
 	ReasonBundleLoadFailed   = "BundleLoadFailed"
 
-	ReasonInstallationFailed        = "InstallationFailed"
-	ReasonInstallationStatusUnknown = "InstallationStatusUnknown"
-	ReasonResolutionFailed          = "ResolutionFailed"
+	ReasonInstallationFailed = "InstallationFailed"
+	ReasonResolutionFailed   = "ResolutionFailed"
 
 	ReasonSuccess       = "Success"
 	ReasonDeprecated    = "Deprecated"
 	ReasonUpgradeFailed = "UpgradeFailed"
 
-	ReasonUnpackPending = "UnpackPending"
 	ReasonUnpackSuccess = "UnpackSuccess"
 	ReasonUnpackFailed  = "UnpackFailed"
 
@@ -163,8 +161,6 @@ func init() {
 		ReasonUpgradeFailed,
 		ReasonBundleLoadFailed,
 		ReasonErrorGettingClient,
-		ReasonInstallationStatusUnknown,
-		ReasonUnpackPending,
 		ReasonUnpackSuccess,
 		ReasonUnpackFailed,
 		ReasonErrorGettingReleaseState,
--- a/api/v1alpha1/clusterextension_types.go
+++ b/api/v1alpha1/clusterextension_types.go
@@ -173,8 +173,10 @@ type BundleMetadata struct {
 }
 
 // ClusterExtensionStatus defines the observed state of ClusterExtension.
-// InstalledBundle should only be modified when a new bundle is successfully installed. This ensures that if you've previously successfully installed a bundle before, and an upgrade fails, it is still appropriately communicated to you that there is still a bundle that is currently installed and owned by the ClusterExtension.
 type ClusterExtensionStatus struct {
+	// InstalledBundle should only be modified when a new bundle is successfully installed. This ensures that if there
+	//  is a previously successfully installed a bundle, and an upgrade fails, it is still communicated that there is
+	//  still a bundle that is currently installed and owned by the ClusterExtension.
 	// +optional
 	InstalledBundle *BundleMetadata `json:"installedBundle,omitempty"`
 	// +optional
--- a/api/v1alpha1/clusterextension_types.go
+++ b/api/v1alpha1/clusterextension_types.go
@@ -172,9 +172,8 @@ type BundleMetadata struct {
 	Version string `json:"version"`
 }
 
-// ClusterExtensionStatus defines the observed state of ClusterExtension
-// InstalledBundle should only be modified when a new bundle is successfully installed, that way if you've previously successfully installed a bundle before,
-// if an upgrade fails it is still appropriately communicated to you that there is still a bundle that is currently installed and owned by the ClusterExtension
+// ClusterExtensionStatus defines the observed state of ClusterExtension.
+// InstalledBundle should only be modified when a new bundle is successfully installed. This ensures that if you've previously successfully installed a bundle before, and an upgrade fails, it is still appropriately communicated to you that there is still a bundle that is currently installed and owned by the ClusterExtension.
 type ClusterExtensionStatus struct {
 	// +optional
 	InstalledBundle *BundleMetadata `json:"installedBundle,omitempty"`
--- a/api/v1alpha1/clusterextension_types.go
+++ b/api/v1alpha1/clusterextension_types.go
@@ -130,7 +130,6 @@ const (
 
 	ReasonInstallationFailed        = "InstallationFailed"
 	ReasonInstallationStatusUnknown = "InstallationStatusUnknown"
-	ReasonInstallationSucceeded     = "InstallationSucceeded"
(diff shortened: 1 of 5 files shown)


api/v1alpha1/clusterextension_types.go
```diff
@@ -167,7 +159,8 @@ type BundleMetadata struct {
 	Version string `json:"version"`
 }
 
-// ClusterExtensionStatus defines the observed state of ClusterExtension
+// ClusterExtensionStatus defines the observed state of ClusterExtension.
+// InstalledBundle should only be modified when a new bundle is successfully installed. This ensures that if you've previously successfully installed a bundle before, and an upgrade fails, it is still appropriately communicated to you that there is still a bundle that is currently installed and owned by the ClusterExtension.
```
joelanford, line 163: Two comments:
1. I don't think the `Status` object is the right place for details of the `InstalledBundle` field. Move this comment to that field?
2. We should be able to assess a "current installation" independent of a bunch of other reconcile tasks. Our reconcile behavior probably needs to go and actually lookup the currently installed bundle early on and affirmatively set `status.conditions[type=Installed]` and `status.installedBundle` based on what we find. If we fail trying to lookup the installed bundle, we should set the condition to Unknown, probably unset the `installedBundle` field, and then continue on with other unrelated reconcile tasks.

internal/controllers/clusterextension_controller.go
```diff
@@ -273,13 +273,12 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
 
 	switch unpackResult.State {
 	case rukpaksource.StatePending:
-		setStatusUnpackPending(ext, unpackResult.Message)
-		setInstalledStatusConditionUnknown(ext, "installation has not been attempted as unpack is pending")
+		setStatusInstallFalseUnpackFailed(ext, unpackResult.Message)
+		setInstalledStatusConditionInstalledFalse(ext, "installation has not been attempted as unpack is pending")
```
joelanford, line 277: Installed=False is not always correct here. If we have previously installed a bundle, and now the unpack of the next bundle is pending, we still have an installed bundle.

The installation status and the bundle unpack status are not related to each other.

internal/controllers/clusterextension_controller.go
```diff
@@ -273,13 +273,12 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
 
 	switch unpackResult.State {
 	case rukpaksource.StatePending:
-		setStatusUnpackPending(ext, unpackResult.Message)
-		setInstalledStatusConditionUnknown(ext, "installation has not been attempted as unpack is pending")
+		setStatusInstallFalseUnpackFailed(ext, unpackResult.Message)
```
joelanford, line 276: Unpack state is pending, but we're setting `Unpacked=False, Failed`?

api/v1alpha1/clusterextension_types.go
LalatenduMohanty: > I don't think the Status object is the right place for details of the InstalledBundle field. Move this comment to that field?

@joelanford Not sure if I understand your comments fully. Currently `InstalledBundle` is part of ClusterExtensionStatus , I guess you are suggesting it should not?

```
	// +optional
	InstalledBundle *BundleMetadata `json:"installedBundle,omitempty"`
	// +optional
	ResolvedBundle *BundleMetadata `json:"resolvedBundle,omitempty"`
	// +patchMergeKey=type
	// +patchStrategy=merge
	// +listType=map
	// +listMapKey=type
	Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type" protobuf:"bytes,1,rep,name=conditions"`
}
```

LalatenduMohanty: Also I do not see any suggested change for the `InstalledBundle` in the RFC document.

internal/controllers/clusterextension_controller.go
bentito: Well this was changed to acknowledge that we've switched to direct image registry and that as far as opr-ctrl is concerned we can't really be in an Unknown state, we're either failed or not. Perhaps we should just not process StatePending here?

api/v1alpha1/clusterextension_types.go
bentito: Fixing (1) above
on point (2), I'd llke to add https://github.com/operator-framework/operator-controller/issues/1012 and keep this PR focused on implementing the RFC.

internal/controllers/clusterextension_controller.go
joelanford: I'm not sure we should write this code with specific knowledge of the underlying `Source` implementation. In theory, the source could be switched out to a different implementation that _can_ return `StatePending`. I've been tinkering with an async direct image client, so this is not entirely hypothetical.

More broadly, I wonder if we even need an `Unpacked` condition type? I don't see it in the diagram in the RFC. The only reason we are unpacking a bundle is so that we can install it. So can we rollup unpacking status into Progressing and/or Installed conditions?

api/v1alpha1/clusterextension_types.go
bentito: @LalatenduMohanty InstalledBundle handing is getting changed here: https://github.com/operator-framework/operator-controller/blob/main/internal/controllers/common_controller.go#L111 to partially answer @joelanford and others comments about when and how we should set it, so yes, not in the RFC, hence the issue add and I'd like to deal with it more thoroughly outside this PR. I wanted to add a better comment to the field here to reflect how we intend its usage. Joe's comment is simply I think that he'd like to see the comment down by the field itself, not up on the struct, which I'll take care of shortly.

internal/controllers/clusterextension_controller.go
bentito: That's farther than the RFC went, but I think it's in the same spirit. I think we can roll it into `TypeInstalled`. I think TypeProgressing would actually be adding a new type?

internal/controllers/clusterextension_controller.go
bentito: okay, just setting the unpack as failed in 0306c9db

internal/controllers/clusterextension_controller.go
bentito: as noted above there is no `TypeProgressing` currently, if we want to eliminate `TypeUnpacked` in favor of `TypeInstalled` 
@joelanford, these two helper methods encapsulate all the places changes would be needed:
```go
func setStatusUnpackFailed(ext *ocv1alpha1.ClusterExtension, message string) {
	ext.Status.InstalledBundle = nil
	apimeta.SetStatusCondition(&ext.Status.Conditions, metav1.Condition{
		Type:               ocv1alpha1.TypeUnpacked,
		Reason:             ocv1alpha1.ReasonUnpackFailed,
		Message:            message,
	})
}

func setStatusUnpacked(ext *ocv1alpha1.ClusterExtension, message string) {
	apimeta.SetStatusCondition(&ext.Status.Conditions, metav1.Condition{
		Type:               ocv1alpha1.TypeUnpacked,
		Status:             metav1.ConditionTrue,
		Reason:             ocv1alpha1.ReasonUnpackSuccess,
		Message:            message,
	})
}
```

So on the fail case we'd want `TypeInstalled` as `ConditionUnknown` and same on the success case. The reason would be the difference. But I wonder if the success case is a bit odd: `TypeInstalled` is Uknown and the reason is Unpack success?

joelanford: I think `TypeInstalled` should literally just be based on: "is there a helm release secret with status `deployed`? if so, Installed=True. If not "Installed=False". If there was an error looking up helm release secrets (other than "not found"), then "Installed=Unknown".

Nothing else should play into it.

bentito: I created https://github.com/operator-framework/operator-controller/issues/1027 to capture this need. I'd like to keep this PR limited to the RFC covered items.

tmshort:
@joelanford are you ok with this?

tmshort (approved):
/lgtm

PR #1008, opened 2024-07-02
 🌱 fix TestClusterExtensionInstallReResolvesWhenNewCataloge2e test

### LINKED ISSUE #945: Fix: e2e test for re-reconcile on catalog change is not valid
The `TestClusterExtensionInstallReResolvesWhenNewCatalog` test is not valid because it sets up a cluster extenstion that fails to resolve. This results in an exponential backoff where retries will occur whether or not a new catalog appears.

To correct this test, we need to:
1. Create a catalog containing `foo.v1.0.0`, with a pollInterval set to `2s` (or something very short)
2. Create a cluster extension for `foo` with no channel or version specifications
3. Wait until `foo.v1.0.0` successfully installs.
4. Update the catalog to contain `foo.v2.0.0` with an upgrade edge from v1, and push the update to the same tag.
5. After step (4) the following should happen all automatically:
   - Catalogd polls and sees a new SHA for the digest, pulls it and makes it available
   - Operator-controller sees the new catalog and re-reconciles all ClusterExtensions
   - The `foo` cluster extension re-resolves, finds `foo.v2.0.0` and installs it.

yashoza19:
# Description

 - Created a new Catalog v2 with Prometheus bundle 2.0.0
 - Changes made to `build-push-e2e-catalog.sh` to allow building and pushing new catalog image
 - Changes made to TestClusterExtensionInstallReResolvesWhenNewCatalog e2e test  - 
1.  Install ClusterExtension with existing Catalog(v1) which has Prometheus v1.2.0
2. Once Prometheus v1.2.0 is installed, patch the ExtensionCatalog with the newly created catalog v2
3. Wait and confirm that ClusterExtension picks the prometheus v2.2.0 from `test-catalog:v2` and resolves to a successful install

fixes #945 

## Reviewer Checklist

- [x] API Go Documentation


test/e2e/e2e_suite_test.go
```diff
@@ -58,3 +59,19 @@ func createTestCatalog(ctx context.Context, name string, imageRef string) (*cata
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Update the ImageRef
+	catalog.Spec.Source.Image.Ref = newImageRef
+
+	// Patch the ClusterCatalog
+	err = c.Update(ctx, catalog)
+	return catalog, err
```
acornett21, line 76: This seems odd to not catch the error here, a caller of this would never know that the catalog they got back from this was the original or the updated one. IMO it would be better to catch the exception.

```suggestion
	err = c.Update(ctx, catalog)
	if err != nil {
		return nil, err
	}
	
	return catalog, err
```

yashoza19: ack, thanks for catching that!

test/tools/build-push-e2e-catalog.sh
```diff
@@ -7,59 +7,61 @@ set -o pipefail
...
+if [[ "$#" -ne 3 ]]; then
   echo "Illegal number of arguments passed"
   echo "${help}"
   exit 1
 fi
 
 namespace=$1
-tag=$2
+image=$2
+tag=$3
```
joelanford, line 26: Instead of passing an extra `tag` argument, can we parse the `image` argument? That would decrease chance of mistakes and copy/pasta errors.

test/e2e/e2e_suite_test.go
```diff
@@ -58,3 +59,23 @@ func createTestCatalog(ctx context.Context, name string, imageRef string) (*cata
...
+	catalog.Spec.Source.Image.Ref = newImageRef
+
+	// Patch the ClusterCatalog
+	err = c.Update(ctx, catalog)
+	if err != nil {
+		return nil, err
+	}
+
+	return catalog, err
+}
```
joelanford, line 81: I think this is a valid test case (where we patch the spec of a ClusterCatalog), but it does not cover the case I had in mind (and a case I think we need to cover).

What I have in mind here specifically is that the polling behavior alone is enough to trigger an upgrade of a ClusterExtension. Forcing a change by changing the image in the ClusterCatalog spec is a side-effect foils the true intention of the test.

This "polling triggers an upgrade" test is one of the most important features of OLM, so we unfortunately can't "fake" it my touching the spec.

yashoza19: That was my first approach as well. Unfortunately, I can't find a way to change the `latest` image tag on a local registry. `crane` does not allow to change tags on local registries, hence I took this approach. Let me know if you you think this is still a valid test or we can approach this issue with a different solution. @joelanford

test/tools/build-push-e2e-catalog.sh
yashoza19: good call. I will update it.

test/e2e/e2e_suite_test.go
acornett21: > This "polling triggers an upgrade" test is one of the most important features of OLM, so we unfortunately can't "fake" it my touching the spec.

How is this accomplished in the current version of OLM? The use of make/sh/kinako(spelling), to get things into the cluster, the go to `run tests`, IMO blurs what is/isn't being tested, there aren't clear separation of duties in this model.

Correct me if I'm wrong, but I think you want an image loaded into the running cluster _after_ the test has executed? Is that correct? Or do you just want this test to run longer to see if the `polling` picks up the the image already in the registry?

joelanford: > How is this accomplished in the current version of OLM?

I'm not sure

> The use of make/sh/kinako(spelling), to get things into the cluster, the go to run tests, IMO blurs what is/isn't being tested, there aren't clear separation of duties in this model.

I agree that the number of moving parts is a problem. I'm fairly confident that if we wanted something a bit leaner, we could build all of this in Go.
1. Code in go that runs during an e2e that deploys an image registry in the cluster.
2. Client-go code that runs `kubectl port-forward` to give the e2e test a way to communicate with the registry.
3. go-containerregistry code to push images, move tags, etc.

> Correct me if I'm wrong, but I think you want an image loaded into the running cluster after the test has executed? Is that correct?

No, I don't think that is sufficient because the polling behavior of catalogd queries a remote registry. `kind load` would not make it visible to the polling process. What I'm after is $something that causes the digest of a tag to change mid way through the test. That could be one of:
1. Pre-populating the registry with two catalog images as part of the setup, and then during the test re-tagging the second image over the first.
2. Pushing the first image at `repo:tag` as part of the setup, and then pushing the second image at that same `repo:tag` during the test.

test/e2e/cluster_extension_install_test.go
```diff
@@ -322,6 +257,81 @@ func TestClusterExtensionInstallSuccessorVersion(t *testing.T) {
...
 }
 
+func TestClusterExtensionInstallReResolvesWhenNewCatalog(t *testing.T) {
+	t.Log("When a cluster extension is installed from a catalog")
+	t.Log("It resolves again when a new catalog is available")
+
+	// Tag the image with the new tag
+	baseImage := "localhost:30000/e2e/test-catalog:v1"
+	var err error
+	err = crane.Tag(baseImage, "latest", crane.Insecure)
```
joelanford, line 267: Should `"latest"` be parsed from the value of `testLatestCatalogRefEnvVar` (instead of being hardcoded)?

yashoza19: `testLatestCatalogRefEnvVar` consist of the complete image path with latest tag. We just need to tag name and not the complete image name here

test/e2e/cluster_extension_install_test.go
```diff
@@ -322,6 +257,81 @@ func TestClusterExtensionInstallSuccessorVersion(t *testing.T) {
 	}, pollDuration, pollInterval)
 }
 
+func TestClusterExtensionInstallReResolvesWhenNewCatalog(t *testing.T) {
+	t.Log("When a cluster extension is installed from a catalog")
+	t.Log("It resolves again when a new catalog is available")
+
+	// Tag the image with the new tag
+	baseImage := "localhost:30000/e2e/test-catalog:v1"
```
joelanford, line 265: Is it just me or does it feel like there was some envvar sprawl before this PR, and this PR is making it more apparent with the addition of the second catalog image?

It somewhat feels like we should either:
1. Completely get rid of the envvars and just hardcode the references in the e2e test code
2. Take a more holistic look at the envvars used to carry image reference and take an opportunity to refactor/reduce
  - For example, should we have LOCAL_REGISTRY_HOST and CLUSTER_REGISTRY_HOST, and then every single repo/image/tag we'd use would get its own envvar with a consistent naming convention?

test/e2e/cluster_extension_install_test.go
```diff
@@ -322,6 +257,81 @@ func TestClusterExtensionInstallSuccessorVersion(t *testing.T) {
...
+			return
+		}
+		assert.Equal(ct, metav1.ConditionTrue, cond.Status)
+		assert.Equal(ct, ocv1alpha1.ReasonSuccess, cond.Reason)
+		assert.Contains(ct, cond.Message, "resolved to")
+		assert.Equal(ct, &ocv1alpha1.BundleMetadata{Name: "prometheus-operator.1.2.0", Version: "1.2.0"}, clusterExtension.Status.ResolvedBundle)
+	}, pollDuration, pollInterval)
+
+	// update tag on test-catalog image with v2 image
+	t.Log("By creating an ClusterExtension catalog with the updated ClusterCatalog")
```
joelanford, line 306: This log message seems incorrecT?

```suggestion
	t.Log("By updating the catalog tag to point to the v2 catalog.")
```

test/e2e/cluster_extension_install_test.go
joelanford: Right, but if we change that environment variable value to use a different tag, we'd have to change all the hardcoded `"latest"` references in the e2e code as well. If we parse the tag out of the reference, there's just the one source of truth (but also see my other comment about possibly refactoring the envvars).

Makefile
```diff
@@ -124,8 +124,16 @@ E2E_REGISTRY_NAME := docker-registry
 E2E_REGISTRY_NAMESPACE := operator-controller-e2e
 
 export REG_PKG_NAME := registry-operator
-export REGISTRY_ROOT := $(E2E_REGISTRY_NAME).$(E2E_REGISTRY_NAMESPACE).svc:5000
-export CATALOG_IMG := $(REGISTRY_ROOT)/e2e/test-catalog:e2e
+export LOCAL_REGISTRY_HOST := $(E2E_REGISTRY_NAME).$(E2E_REGISTRY_NAMESPACE).svc:5000
+export CLUSTER_REGISTRY_HOST := localhost:30000
```
joelanford, line 128: Not sure what others think, but to me it is more logical to flip these. LOCAL for `localhost`, CLUSTER for the service-based name.

Makefile
```diff
@@ -124,8 +124,16 @@ E2E_REGISTRY_NAME := docker-registry
...
+export V1_IMAGE_TAG := v1
+export V2_IMAGE_TAG := v2
+export LATEST_IMAGE_TAG := latest
+export CATALOG_IMG := $(LOCAL_REGISTRY_HOST)/e2e/test-catalog:$(V1_IMAGE_TAG)
+export UPDATED_CATALOG_IMG := $(LOCAL_REGISTRY_HOST)/e2e/test-catalog:$(V2_IMAGE_TAG)
+export LATEST_CATALOG_IMG := $(LOCAL_REGISTRY_HOST)/e2e/test-catalog:$(LATEST_IMAGE_TAG)
+export CLUSTER_V1_CATALOG_IMG := $(CLUSTER_REGISTRY_HOST)/e2e/test-catalog:$(V1_IMAGE_TAG)
+export CLUSTER_V2_CATALOG_IMG := $(CLUSTER_REGISTRY_HOST)/e2e/test-catalog:$(V2_IMAGE_TAG)
```
joelanford, line 136: Two comments here: 
1. I don't think we need envvars for every combo. Let's just concatenate.
2. Let's pick a convention like `<repoName>_<imageName_<tag>` and use that.
3. I don't think we'll need a "latest" variant here anymore. The test that needs it can deal with pushing its own tag local to the test.

So I think all of this can boil down to:
```suggestion
E2E_TESTCATALOG_V1 = e2e/testcatalog:v1
E2E_TESTCATALOG_V2 = e2e/testcatalog:v2
```

And then when we need to reference a fully-qualified image, we would just use, for example:
```
$(LOCAL_REGISTRY_HOST)/$(E2E_TEST_CATALOG_V1)
```

Makefile
acornett21: I wouldn't call `localhost` a host, this is the `ENDPOINT` into the the cluster. It's really the `external` callable address. Spitballing, because names are hard, but maybe `CLUSTER_REGISTRY_INGRESS`? 

I agree the `svc:5000` address should be called `CLUSTER`.

Makefile
yashoza19: but we still need to export the cluster image as `CLUSTER_V1_CATALOG_IMG := $(CLUSTER_REGISTRY_HOST)/$(E2E_TEST_CATALOG_V1)` in the makefile, right?

acornett21: Correct me if I'm wrong @joelanford, but I think you want `less` stuff in the Makefile, and more of the `work` in the tests that need "additional" things. IE instead of exported vars, concatenate them in the go code? So it would be something like:

```
fmt.Sprintf("%s/%s", os.Getenv("LOCAL_REGISTRY_HOST"), os.Getenv("E2E_TEST_CATALOG_V1"))
```

Or did you mean something entirely different.

yashoza19: This has been implemented now in the latest commit. Thanks

Makefile
```diff
@@ -124,8 +124,11 @@ E2E_REGISTRY_NAME := docker-registry
 E2E_REGISTRY_NAMESPACE := operator-controller-e2e
 
 export REG_PKG_NAME := registry-operator
-export REGISTRY_ROOT := $(E2E_REGISTRY_NAME).$(E2E_REGISTRY_NAMESPACE).svc:5000
-export CATALOG_IMG := $(REGISTRY_ROOT)/e2e/test-catalog:e2e
+export LOCAL_REGISTRY_HOST := $(E2E_REGISTRY_NAME).$(E2E_REGISTRY_NAMESPACE).svc:5000
+export CLUSTER_REGISTRY_HOST := localhost:30000
+export E2E_TEST_CATALOG_V1 := e2e/test-catalog:v1
+export E2E_TEST_CATALOG_V2 := e2e/test-catalog:v2
+export CATALOG_IMG := $(LOCAL_REGISTRY_HOST)/$(E2E_TEST_CATALOG_V1)
```
joelanford, line 131: Do we need `CATALOG_IMG` variable anymore?

yashoza19: Yes, it is required here: https://github.com/operator-framework/operator-controller/blob/main/test/extension-developer-e2e/setup.sh#L33

test/e2e/e2e_suite_test.go
```diff
@@ -24,6 +25,7 @@ var (
 const (
 	testCatalogRefEnvVar = "CATALOG_IMG"
 	testCatalogName      = "test-catalog"
+	LatestImageTag       = "latest"
```
joelanford, line 28: Is there a reason to export this? I'd suggest unexporting:

```suggestion
	latestImageTag       = "latest"
```

joelanford: I'll hit this in a quick follow-up.

yashoza19: I've made the changes in this PR itself

acornett21 (approved):

test/e2e/cluster_extension_install_test.go
```diff
@@ -337,6 +269,84 @@ func TestClusterExtensionInstallSuccessorVersion(t *testing.T) {
 	}, pollDuration, pollInterval)
 }
 
+func TestClusterExtensionInstallReResolvesWhenNewCatalog(t *testing.T) {
```
perdasilva, line 272: @joelanford I wonder if we should rethink the e2es to flow from each other some how and where possible. What I mean is, do we want (just about) every test to start with install a CE then wait? I'm wondering if instead we could 1) install something 2) upgrade that something 3) swap the catalog 4) bring it down, etc. I want to avoid as much as possible the 2h e2e test runtime (if possible). I know there's some pros and cons here. So, just wondering if there's another strategy that we can adopt that keep the resource usage lower while still giving us a strong signal.

Makefile
perdasilva: Do we need to address this before merging? Looks like we should...?

Makefile
```diff
@@ -124,8 +124,11 @@ E2E_REGISTRY_NAME := docker-registry
 E2E_REGISTRY_NAMESPACE := operator-controller-e2e
 
 export REG_PKG_NAME := registry-operator
-export REGISTRY_ROOT := $(E2E_REGISTRY_NAME).$(E2E_REGISTRY_NAMESPACE).svc:5000
-export CATALOG_IMG := $(REGISTRY_ROOT)/e2e/test-catalog:e2e
+export LOCAL_REGISTRY_HOST := $(E2E_REGISTRY_NAME).$(E2E_REGISTRY_NAMESPACE).svc:5000
+export CLUSTER_REGISTRY_HOST := localhost:30000
+export E2E_TEST_CATALOG_V1 := e2e/test-catalog:v1
```
perdasilva, line 129: I'm seeing a lot of os.Getenvs in the e2es. Could we (if it's not there already), group the env vars that the e2e expects and add a comment? I know we'll find out relatively quickly if we deleted something we shouldn't, but I just want to avoid the CI run for that if we can.

joelanford (approved):

Issue #971, opened 2024-06-25
Add new required field to ClusterExtension spec for users to reference a ServiceAccount


everettraven:
In order to begin using a `ServiceAccount` to install and manage content, a user needs to be able to specify a `ServiceAccount` for operator-controller to use when creating a `ClusterExtension`. Since OLMv1 will not have any permissions for installing and managing permissions by default this will need to be a required, immutable field.

An example of what a `ClusterExtension` resource could look like with this additional field:
```yaml
apiVersion: olm.operatorframework.io/v1alpha1
kind: ClusterExtension
metadata:
  name: clusterextension-sample
spec:
  installNamespace: default
  packageName: argocd-operator
  version: 0.6.0
  serviceAccount:
    name: argocd-installer
```

**Acceptance Criteria**:
- A new required, immutable field in the `ClusterExtension` spec for specifying a `ServiceAccount` that should be used to install and manage content
- Updates to unit and e2e tests as necessary

everettraven:
This is done as of #1011

PR #1011, opened 2024-07-05
✨ Add Service Account requirement to ClusterExtension

makon57:
# Description

This PR solves Issue #971 - Add new required field to ClusterExtension spec for users to reference a ServiceAccount
- Created new required 'Service Account' field in ClusterExtensionSpec
- Added Service Account to unit tests 
- Added Service Account to e2e test

## Reviewer Checklist

- [ ] API Go Documentation


api/v1alpha1/clusterextension_types.go
```diff
@@ -83,6 +83,11 @@ type ClusterExtensionSpec struct {
 	//+kubebuilder:Optional
 	// Preflight defines the configuration of preflight checks.
 	Preflight *PreflightConfig `json:"preflight,omitempty"`
+
+	//+kubebuilder:validation:MaxLength:=48
+	//+kubebuilder:validation:Pattern:=^[a-z0-9]+(-[a-z0-9]+)*$
```
everettraven, line 88: We should also make this immutable as outlined in the issue and from our earlier discussions. https://github.com/operator-framework/operator-controller/pull/1011/files#diff-8e9c0eb955489f1a05a95a75a25842a010fdc02029ce06aae7d2163b895c17cbR76 should be a good example of making a field immutable

test/e2e/cluster_extension_install_test.go
```diff
@@ -75,6 +75,7 @@ func TestClusterExtensionInstallRegistry(t *testing.T) {
 	clusterExtension.Spec = ocv1alpha1.ClusterExtensionSpec{
 		PackageName:      "prometheus",
 		InstallNamespace: "default",
+		ServiceAccount:   "default",
```
everettraven, line 78: I have a suspicion that the default ServiceAccount won't have the permissions necessary to install an extension. We may need to ensure we are creating a new ServiceAccount and the appropriate RBAC resources to install the e2e extensions

everettraven: I just realized that this is irrelevant for this particular PR because it is just adding the new field and doesn't actually implement the logic 🤦. Apologies for this distraction - I don't know what I was doing and why I thought this would be the case...

internal/controllers/clusterextension_admission_test.go
```diff
@@ -235,6 +239,52 @@ func TestClusterExtensionAdmissionInstallNamespace(t *testing.T) {
...
+func TestClusterExtensionAdmissionServiceAccount(t *testing.T) {
+	tooLongError := "spec.serviceAccount: Too long: may not be longer than 48"
+	regexMismatchError := "spec.serviceAccount in body should match"
+
+	testCases := []struct {
+		name           string
+		serviceAccount string
+		errMsg         string
+	}{
+		{"just alphanumeric", "justalphanumberic1", ""},
```
bentito, line 251:
```suggestion
		{"just alphanumeric", "justalphanumeric1", ""},
```

internal/controllers/clusterextension_admission_test.go
```diff
@@ -235,6 +239,52 @@ func TestClusterExtensionAdmissionInstallNamespace(t *testing.T) {
...
+		{"hypen-separated", "hyphenated-name", ""},
+		{"no service account name", "", regexMismatchError},
+		{"dot-separated", "dotted.name", regexMismatchError},
+		{"longest valid service account name", strings.Repeat("x", 48), ""},
+		{"too long service account name", strings.Repeat("x", 49), tooLongError},
```
bentito, line 256: this seems wrong, as the "longest valid service account name"
```bash
$ python3
Python 3.12.3 (main, Apr  9 2024, 08:09:14) [Clang 15.0.0 (clang-1500.3.9.4)] on darwin
Type "help", "copyright", "credits" or "license" for more information.

>>> len("longest valid service account name")+48
82

>>> len("too long service account name")+49
78
```
?

api/v1alpha1/clusterextension_types.go
```diff
@@ -83,6 +83,11 @@ type ClusterExtensionSpec struct {
 	//+kubebuilder:Optional
 	// Preflight defines the configuration of preflight checks.
 	Preflight *PreflightConfig `json:"preflight,omitempty"`
+
+	//+kubebuilder:validation:MaxLength:=48
```
joelanford, line 87: Service accounts can be any valid DNS 1123 subdomain. See https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names

We should adjust the MaxLength and Pattern to make sure it matches what DNS 1123 specifies.

api/v1alpha1/clusterextension_types.go
```diff
@@ -83,6 +83,11 @@ type ClusterExtensionSpec struct {
...
+	//+kubebuilder:validation:Pattern:=^[a-z0-9]+(-[a-z0-9]+)*$
+	// ServiceAccount is used to install and manage content
+	ServiceAccount string `json:"serviceAccount"`
```
joelanford, line 90: I followed up to figure out what a good convention should be for a reference to a ServiceAccount.

I propose we follow the OpenShift convention: https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md#use-specific-types-for-object-references-and-omit-ref-suffix, which would mean introducing a new `ServiceAccountReference` type that includes a `Name string` field.

api/v1alpha1/clusterextension_types.go
```diff
@@ -83,6 +83,11 @@ type ClusterExtensionSpec struct {
 	//+kubebuilder:Optional
 	// Preflight defines the configuration of preflight checks.
 	Preflight *PreflightConfig `json:"preflight,omitempty"`
+
+	//+kubebuilder:validation:MaxLength:=48
+	//+kubebuilder:validation:Pattern:=^[a-z0-9]+(-[a-z0-9]+)*$
+	// ServiceAccount is used to install and manage content
```
joelanford, line 89: The GoDoc needs to explain that the service account is expected to exist in the install namespace.

api/v1alpha1/clusterextension_types.go
```diff
@@ -83,6 +83,19 @@ type ClusterExtensionSpec struct {
...
+	// The service account is expected to exist in the InstallNamespace.
+	ServiceAccount ServiceAccountReference `json:"serviceAccount"`
+}
+
+// ServiceAccountReference references a serviceAccount.
+type ServiceAccountReference struct {
+	// name is the metadata.name of the referenced serviceAccount object.
+	//+kubebuilder:validation:MaxLength:=253
+	//+kubebuilder:validation:Pattern:=^[a-z0-9]+([.|-][a-z0-9]+)*$
+	//+kubebuilder:validation:XValidation:rule="self == oldSelf",message="serviceAccount is immutable"
```
joelanford, line 97: I think we should change this to `name is immutable`. This message will show up in the context of the `spec.serviceAccount` field like this:

```
spec.serviceAccount: Invalid value: "string": name is immutable
```

joelanford (approved):

api/v1alpha1/clusterextension_types.go
```diff
@@ -83,6 +83,19 @@ type ClusterExtensionSpec struct {
 	//+kubebuilder:Optional
 	// Preflight defines the configuration of preflight checks.
 	Preflight *PreflightConfig `json:"preflight,omitempty"`
+
+	// ServiceAccount is used to install and manage resources.
+	// The service account is expected to exist in the InstallNamespace.
+	ServiceAccount ServiceAccountReference `json:"serviceAccount"`
```
everettraven, line 89: Since we are attempting to follow the Kubernetes/OpenShift API conventions from previous comments by @joelanford, I think we should follow the naming convention for references which is `{field}Ref` as outlined by https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#naming-of-the-reference-field

```suggestion
	ServiceAccountRef ServiceAccountReference `json:"serviceAccountRef"`
```

everettraven: Ah, I do see in the openshift conventions they state that the `Ref` suffix should be omitted. @joelanford Do we have any preference in the API conventions we want to lean towards?

makon57: This is the one I followed https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md#use-specific-types-for-object-references-and-omit-ref-suffix

everettraven: As mentioned in https://kubernetes.slack.com/archives/C0181L6JYQ2/p1720534180890229 I won't hold the PR on this, but we should begin to think about the API conventions we want to favor and highlight when and why we deviate.

everettraven (approved):

Issue #736, opened 2024-04-04
[epic] Ensure no two ClusterExtensions can manage the same underlying object


joelanford:
For v1.0.0 it was decided that utilizing the underlying Helm management when concurrent reconciliations is set to a value of `1` is sufficient. While the underlying implementation details may change to prevent race conditions when scaling concurrent reconciliations > 1, the behavior that no two `ClusterExtension`s should be able to manage the same underlying objects needs to be documented.

Documentation should be contributed to the [`docs/drafts/` directory](https://github.com/operator-framework/operator-controller/tree/main/docs/drafts)

Issues:
- [x] https://github.com/operator-framework/operator-controller/issues/1247

joelanford:
I spent a little time hacking on this in the Rukpak repo and came up with:

```
---
apiVersion: admissionregistration.k8s.io/v1beta1
kind: ValidatingAdmissionPolicy
metadata:
  name: single-owner-policy
spec:
  failurePolicy: Fail
  matchConstraints:
    objectSelector:
      matchExpressions:
      - key: "core.rukpak.io/owner-name"
        operator: Exists
    resourceRules:
    - apiGroups:
      - "*"
      apiVersions:
      - "*"
      resources:
      - "*"
      operations:
      - UPDATE

  validations:
  - expression: |
      !("core.rukpak.io/owner-name" in oldObject.metadata.labels) ||
      !("core.rukpak.io/owner-name" in object.metadata.labels) ||
      oldObject.metadata.labels["core.rukpak.io/owner-name"] == object.metadata.labels["core.rukpak.io/owner-name"]
    messageExpression: '"object is owned by BundleDeployment \"" + oldObject.metadata.labels["core.rukpak.io/owner-name"]
      + "\": cannot be adopted by BundleDeployment \"" + object.metadata.labels["core.rukpak.io/owner-name"]
      + "\""'
    reason: Invalid
---
apiVersion: admissionregistration.k8s.io/v1beta1
kind: ValidatingAdmissionPolicyBinding
metadata:
  name: single-owner-policy
spec:
  policyName: single-owner-policy
  validationActions:
  - Deny
```

It will need to be adapted to change "core.rukpak.io/owner-name" to whatever label we decide to inject from the ClusterExtension API, but other than that my manual tests seemed to do the right thing.

[...]

PR #1016, opened 2024-07-08
:sparkles: mask helm conflict errors

perdasilva:
# Description
As part of #736 we'd like to mark helm errors when an installation fails due to conflict. This PR wraps the ActionClient interface we use to talk to helm and processes the errors returned give a set of programatic rules. So far, only the install conflict error translator is modeled

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -41,6 +41,7 @@ import (
 	helmclient "github.com/operator-framework/helm-operator-plugins/pkg/client"
 
 	ocv1alpha1 "github.com/operator-framework/operator-controller/api/v1alpha1"
+	"github.com/operator-framework/operator-controller/internal/action"
 	"github.com/operator-framework/operator-controller/internal/catalogmetadata/cache"
 	catalogclient "github.com/operator-framework/operator-controller/internal/catalogmetadata/client"
 	"github.com/operator-framework/operator-controller/internal/controllers"
@@ -184,9 +185,10 @@ func main() {
 		os.Exit(1)
 	}
 
-	acg, err := helmclient.NewActionClientGetter(cfgGetter,
+	acg, err := action.NewWrappedActionClientGetter(cfgGetter,
 		helmclient.WithFailureRollbacks(false),
 	)
+
 	if err != nil {
 		setupLog.Error(err, "unable to create helm client")
 		os.Exit(1)
--- a/internal/action/error/errors.go
+++ b/internal/action/error/errors.go
@@ -1,0 +1,73 @@
+package error
+
+import (
+	"fmt"
+	"regexp"
+)
+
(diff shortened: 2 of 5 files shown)


internal/error/errors.go
```diff
@@ -0,0 +1,84 @@
...
+	for _, rule := range rules {
+		if matches := rule.MatchError(err); len(matches) > 0 {
+			// first index is the full match
+			logrus.Debugf("found match with params %v\n", matches)
+			return rule.TranslateError(err, matches[1:]...)
+		}
+	}
+
+	// let's mark any unmatched errors as unknown
+	return UnknownErr(err)
```
tmshort, line 35: `UnknownErr` is not explicitly checked anywhere as a type. Do we need this. Could this simply `return err`?

internal/error/errors.go
```diff
@@ -0,0 +1,84 @@
...
+
+func (r rule) TranslateError(err error, params ...string) error {
+	return r.translator(err, params...)
+}
+
+// rules is a list of rules for error translation
+var rules = []rule{
+	{regexErrorMatcher(installConflictErrorPattern), installConflictErrorTranslator},
+	{regexErrorMatcher(installConflictErrorGeneralPattern), installConflictErrorTranslator},
+}
```
tmshort, line 64: Should these use different functions? That might make `installConflictErrorTranslater()` below simpler... no need for a default `msg`, and no need to check for `len(params)`, which might be error prone.

LalatenduMohanty:
@perdasilva can you please add the results (logging) before and after the PR here. Which will help me better understand the difference in user experience.

perdasilva:
> @perdasilva can you please add the results (logging) before and after the PR here. Which will help me better understand the difference in user experience.

Went from:

```
CustomResourceDefinition "applications.argoproj.io" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: key "meta.helm.sh/release-name" must equal "clusterextension-test": current value is "clusterextension-sample"; annotation validation error: key "meta.helm.sh/release-namespace" must equal "test": current value is "default"
```

To:

CustomResourceDefinition 'applications.argoproj.io' already exists in namespace ''

LalatenduMohanty:
> CustomResourceDefinition 'applications.argoproj.io' already exists in namespace ''

@joelanford mentioned that this logging is also not right because it does not the point right root cause to the user.

perdasilva:
> > CustomResourceDefinition 'applications.argoproj.io' already exists in namespace ''
> 
> @joelanford mentioned that this logging is also not right because it does not the point right root cause to the user.

it would be good to have an example

tmshort (approved):
/lgtm

Issue #1018, opened 2024-07-08
Add default printer columns for clusterextensions


kevinrizza:
Currently, `kubectl get clusterextensions` doesn't display any custom printer columns, just the name and namespace.

We should add some printer columns so that we can get some metadata about what is actually installed.

PR #1019, opened 2024-07-08
✨ [WIP] Add support for kubectl printer columns

kevinrizza:
Adds installed version, resolved version and age printer columns

Resolves https://github.com/operator-framework/operator-controller/issues/1018

# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/api/v1alpha1/clusterextension_types.go
+++ b/api/v1alpha1/clusterextension_types.go
@@ -183,6 +183,9 @@ type ClusterExtensionStatus struct {
 //+kubebuilder:object:root=true
 //+kubebuilder:resource:scope=Cluster
 //+kubebuilder:subresource:status
+//+kubebuilder:printcolumn:name="Installed Version",type=string,JSONPath=`.status.installedBundle.version`
+//+kubebuilder:printcolumn:name="Resolved Version",type=string,JSONPath=`.status.resolvedBundle.version`
+//+kubebuilder:printcolumn:name=Age,type=date,JSONPath=`.metadata.creationTimestamp`
 
 // ClusterExtension is the Schema for the clusterextensions API
 type ClusterExtension struct {
--- a/config/base/crd/bases/olm.operatorframework.io_clusterextensions.yaml
+++ b/config/base/crd/bases/olm.operatorframework.io_clusterextensions.yaml
@@ -14,7 +14,17 @@ spec:
     singular: clusterextension
   scope: Cluster
   versions:
-  - name: v1alpha1
+  - additionalPrinterColumns:
+    - jsonPath: .status.installedBundle.version
+      name: Installed Version
+      type: string
+    - jsonPath: .status.resolvedBundle.version
+      name: Resolved Version
+      type: string
+    - jsonPath: .metadata.creationTimestamp
+      name: Age
+      type: date
+    name: v1alpha1
     schema:
       openAPIV3Schema:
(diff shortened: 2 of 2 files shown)


kevinrizza:
These seem like reasonable printer columns to me, but we should have some agreement that these make the most sense before committing to them. Leaving this as a WIP pr for now.

joelanford (approved):
Do you know if there's a way to set a default value in the printer column if the json path evaluates to null or empty string?

Related: I'm wondering if there's a nice way to get useful information to show up if there are issues resolving or installing a bundle.

I found this https://github.com/kubernetes/enhancements/pull/4602, but obviously not usable yet.

kevinrizza:
@joelanford I looked around and didn't find anything. Imo we should keep this pr held until we make some improvements to the availability of this metadata. I'm also interested in any other thoughts on other columns we might want to have.

api/v1alpha1/clusterextension_types.go
```diff
@@ -183,6 +183,9 @@ type ClusterExtensionStatus struct {
 //+kubebuilder:object:root=true
 //+kubebuilder:resource:scope=Cluster
 //+kubebuilder:subresource:status
+//+kubebuilder:printcolumn:name="Installed Version",type=string,JSONPath=`.status.installedBundle.version`
```
perdasilva, line 186: is it possible to float up any of the conditions or errors? As a user it would be good to understand: where the CR is in its progression, version state (as above), and some kind of sense of where it stopped (and possibly why?)

joelanford: I think it is possible to write JSON path to do something like "find condition with type "Installed" and tell me it's reason", but that might not be all that helpful because:
1. Showing details of every condition type would get pretty verbose
2. Only showing details of _some_ condition types could be misleading.

I wonder if, after @bentito's work to wrangle condition updates to reflect reality better, we could do something where right before we send out the status update we pass the conditions into a function that returns $something that somewhat summarizes the state of the conditions. For example:
- Installed=False, Progressing=False => "Failed"
- Installed=False, Progressing=True => "Installing"
- Installed=True, Progressing=False, Healthy=False => "Unhealthy"
- Installed=True, Progressing=False, Healthy=True => "Ready"
- Installed=True, Progressing=True => "Upgrading"
- etc.

perdasilva: That could be cool. It would just be nice to get some signal without having to inspect the whole status.

tmshort:
Where are we on this? Keeping it? Does it need rework?

api/v1alpha1/clusterextension_types.go
tmshort: I've been looking into this, and it appears the JSON is limited to being very simply, and you really can't do something that complicated with another field having that value.

tmshort: Error I got when I tried to be a bit more "advanced":
```
The CustomResourceDefinition "clusterextensions.olm.operatorframework.io" is invalid: spec.additionalPrinterColumns[4].JSONPath: Invalid value: "{range .status.conditions[?(@.type=='Installed' && @.status=='False')]}{range .status.conditions[?(@.type=='Progressing' && @.status=='True')]}{'Installed'}{end}{end}": must be a simple json path starting with .

m1kola:
Closing if favour of #1421

PR #1020, opened 2024-07-08
✨ bump helm-operator-plugins to v0.3.0; disable failure rollbacks

### LINKED ISSUE #994: operator-controller should not uninstall/rollback a failed install/upgrade
Right now, operator-controller is using helm-operator-plugins, which automatically uninstalls failed installations and rolls back failed upgrades.

This is problematic because the installation or upgrade may have progressed to the point of no return. For example:
- a pre-install/pre-upgrade hook may have run a migration and a subsequent object failed to update.
- after any hooks run, a deployment may have started that ran a migration, but then subsequent objects fail to update.

In my opinion, a better behavior is to just stop, inform a user of the problem, and require human intervention to resolve and progress.

joelanford:
# Description

This PR bumps to a new version of helm-operator-plugins, which brings a few new fixes, features, and changes. Notably:
- ActionClientGetter: allow disabling rollbacks after failures 

This PR also:
- makes use of the new ActionClientGetter knob so that operator-controller no longer rolls back a failed install or upgrade.
- sets a max version history of 10 to avoid unlimited piling up of release secrets
- uses helm's new server-side dry-run option, which provides cluster information to helm templates (e.g. when using the `lookup` template function).

For all changes, see https://github.com/operator-framework/helm-operator-plugins/releases/tag/v0.3.0

Closes #994 

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -176,7 +176,9 @@ func main() {
 		os.Exit(1)
 	}
 
-	acg, err := helmclient.NewActionClientGetter(cfgGetter)
+	acg, err := helmclient.NewActionClientGetter(cfgGetter,
+		helmclient.WithFailureRollbacks(false),
+	)
 	if err != nil {
 		setupLog.Error(err, "unable to create helm client")
 		os.Exit(1)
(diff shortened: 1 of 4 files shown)


joelanford:
With the failure rollbacks disabled, I am seeing the expected exponential backoffs retrying (and still failing; in my case I changed a manifest to be invalid, so i expect permafailure). And then we I patch the clusterextension to allow choosing a working version, I see that upgrade go through and the installation succeed.

grokspawn (approved):

PR #1032, opened 2024-07-10
🌱 Migrate Rukpak

### LINKED ISSUE #997: Pull rukpak code into operator-controller
Pull the rukpak code into operator-controller per the RFC.

dtfranz:
Moves rukpak code imported by operator-controller into the `internal/` folder. 

Fixes #997


internal/bundledeployment/bundledeployment.go
```diff
@@ -0,0 +1,244 @@
+/*
```
dtfranz, line 1: Commenting here to call specific attention to this file. How much more of the k8s stuff here can be removed? Do we need the `json:"..."` tags? Are we OK with the fact that I copied an autogenerated func into this file in order to maintain `DeepCopy()`?

dtfranz:
General comment: there are likely some things copied over that are unused. I'll go through and make sure that's all removed.

dtfranz:
> General comment: there are likely some things copied over that are unused. I'll go through and make sure that's all removed.

I've removed everything I could find that was unused; see comparison [here](https://github.com/operator-framework/operator-controller/compare/581e4a6ce582fb8c31e45d94561f39af5b4f4bcc..ecacb47e87d2160f1d190aba98faa6c0c21727c4) to see what I left out.

joelanford:
WDYT about dumping everything in under `internal/rukpak` for now so that it's more obvious in a first pass what came from rukpak and what changed in existing operator-controller code?

I'm very much in favor of moving our rukpak code directly into this repo, but I'm also feeling a little bit like we need to (in a future PR) step back and take stock of our suddenly sprawling codebase to see if there are obsolete abstractions (e.g. unpack + storage) and other ways to reduce our package count and tidy things up a bit.

internal/controllers/clusterextension_controller.go
```diff
@@ -280,7 +280,7 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
 	case rukpaksource.StateUnpacked:
 		// TODO: Add finalizer to clean the stored bundles, after https://github.com/operator-framework/rukpak/pull/897
 		// merges.
-		if err := r.Storage.Store(ctx, ext, unpackResult.Bundle); err != nil {
+		if err := r.Storage.Store(ctx, ext.GetName(), unpackResult.Bundle); err != nil {
```
bentito, line 283: Was this just because you realized we only need to store the name and not the whole struct?

internal/handler/interfaces.go
```diff
@@ -0,0 +1,21 @@
...
+	"context"
+	"io/fs"
+
+	"helm.sh/helm/v3/pkg/chart"
+	"helm.sh/helm/v3/pkg/chartutil"
+
+	bd "github.com/operator-framework/operator-controller/internal/bundledeployment"
+)
+
+type Handler interface {
```
bentito, line 13: Should this maybe become less generic, so `RegistryHandler` perhaps?

internal/source/image_registry.go
```diff
@@ -0,0 +1,206 @@
...
+	"crypto/x509"
+	"errors"
+	"fmt"
+	"io/fs"
+	"net/http"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/containerd/containerd/archive"
```
bentito, line 16: The focus here is just pull in the code, but maybe we should add an issue to make more use of `containerd` for the future?  Seems like `containerd` has image operations that could simplify TLS and authentication handling, streamline pulling images, and optimize unpacking layers.

internal/storage/localdir.go
```diff
@@ -0,0 +1,77 @@
...
+	"path/filepath"
+
+	"github.com/nlepage/go-tarfs"
+
+	"github.com/operator-framework/operator-controller/internal/util"
+)
+
+var _ Storage = &LocalDirectory{}
+
+const DefaultBundleCacheDir = "/var/cache/bundles"
```
bentito, line 22: Never used?

dtfranz:
> WDYT about dumping everything in under `internal/rukpak` for now so that it's more obvious in a first pass what came from rukpak and what changed in existing operator-controller code?

I'm very much in agreement with this and your other thought as well. Since I filtered out a lot of the types there's not currently much reason for the abstract types like `storage` to exist in the codebase.

For now at the very least I'll move everything into the `rukpak` subdir inside `internal`.

internal/controllers/clusterextension_controller.go
dtfranz: Yes, the thing we're actually storing is the `unpackResult.Bundle`, and `ext` was passed in purely for its identifier.

internal/handler/interfaces.go
dtfranz: Yes, there's a good argument for that since we're only using one of the 3 types from rukpak.

internal/provisioner/plain/plain.go
```diff
@@ -0,0 +1,102 @@
...
+)
+
+const (
+	// ProvisionerID is the unique plain provisioner ID
+	ProvisionerID = "core-rukpak-io-plain"
+
+	manifestsDir = "manifests"
+)
+
+func HandleBundleDeployment(_ context.Context, fsys fs.FS, _ *bundledeployment.BundleDeployment) (*chart.Chart, chartutil.Values, error) {
```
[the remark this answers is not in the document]
dtfranz: Right; and that brings up another larger issue for this PR: Rukpak was kind of infamous for testing with e2e tests instead of unit tests, so we can't bring over a lot of those tests as-is. Personally I think there's probably too much new code in this PR for adding unit tests for all of it and I'd prefer to open a new issue for that if you think that's reasonable.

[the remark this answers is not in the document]
bentito: sounds good.to me. just need an issue for adding rukpakian related unit test, to do later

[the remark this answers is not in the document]
dtfranz: Created #1037

tmshort:
> For now at the very least I'll move everything into the `rukpak` subdir inside `internal`.

This is what I originally did for the original integration (before we pulled everything back)

tmshort:
We might possibly want to remove rukpak references from `.golangci.yaml`, but not critical now. With this it appears that there are no references to the external `operator-framework/rukpak` repo any more.

tmshort:
Looks as though a lot of the original code was not pulled over (good), and necessary changes made to only support imagev1.

tmshort (dismissed):
/lgtm

tmshort:
@dtfranz, you'll need to rebase due to go.mod changes....

tmshort (approved):

PR #1030, opened 2024-07-10
✨ WIP Improved error handling around installedBundle and reconcile modularization

bentito:
# Description

WIP, because:
- I'd like feedback on whether the reconciler() loop modularization seems good and whether the proposed logic flow changes match up with https://github.com/operator-framework/operator-controller/issues/1027 , which is restated below.

- Unit tests are still broken. Just looking for feedback on overall approach.

- This branch/PR cherry picks all the new commits (as of now) from branch/PR:
https://github.com/operator-framework/operator-controller/pull/1007

- vs. 1007 this PR also merged a slightly updated reconciler() from main around preflight implementation, so we should check logic still as intended

---

1. Improved Error Handling

- **State Preservation**: The installedBundle information is fetched early on and preserved throughout the reconciliation process. Even if errors occur later, the status accurately reflects the previously installed bundle.

- **Partial Progress Reporting**: Each step has its own error handling and status update logic. If an error occurs in one step, the status is updated accordingly (e.g., setResolvedStatusConditionFailed, setInstalledStatusConditionFailed), but the reconciliation process continues where possible. This ensures that partial progress is reported.

- **Progressing Status**: The setInstalledStatusConditionProgressing function is used to indicate when a new version is resolved but not yet installed. The clearInstalledStatusConditionProgressing function is called when the installation is complete or if the resolved and installed bundles match.

2. Modularization

- **Helper Functions**: The reconcile function is broken down into smaller, self-contained helper functions (handleFinalization, resolveBundle, validateAndCheckDeprecation, unpackBundle, runPreflightChecks, installOrUpgrade, and watchDynamicResources). Each function handles a specific task, making the code more organized, readable, and testable.

3. Helm Secret as Source of Truth

- **GetInstalledBundle**: The InstalledBundleGetter.GetInstalledBundle function is used to retrieve the installed bundle information directly from the Helm secret, ensuring that the status accurately reflects the actual installed bundle.

4. Comprehensive Status Updates

- **Status Updates at Each Step**: Throughout the reconcile function and its helpers, status conditions are updated to reflect the progress and outcome of each step. This provides detailed information about the reconciliation process, even if errors occur.

- **Progressing Status**: The TypeProgressing condition is used to clearly indicate when an installation or upgrade is in progress.

## Reviewer Checklist

- [ ] API Go Documentation
- [ ] Tests: Unit Tests (and E2E Tests, if appropriate)
- [ ] Comprehensive Commit Messages
- [ ] Links to related GitHub Issue(s)

PR #1033, opened 2024-07-11
🌱 use operator-registry's FBC types directly; add `Resolver` interface

joelanford:
Signed-off-by: Joe Lanford <joe.lanford@gmail.com>

# Description

This is a code refactoring to:
1. align operator-controller's resolver to use the FBC types directly, which seems like it would be able to evolve better alongside FBC evolution.
2. introduce a resolver abstraction to the reconciler, which simplifies reconciler AND resolver testability

## Reviewer Checklist

- [ ] API Go Documentation


joelanford:
I'm hoping #1050 improves the situation on the upgrade-e2e. I'll rebase after that merges.

joelanford:
Rebased. Let's see if upgrade-e2e passes now.

internal/bundleutil/bundle.go
```diff
@@ -0,0 +1,39 @@
...
+			if err != nil {
+				return nil, err
+			}
+			return &vers, nil
+		}
+	}
+	return nil, fmt.Errorf("no package property found in bundle %q", b.Name)
+}
+
+// MetadataFor returns a BundleMetadata for the given bundle. If the provided bundle is nil,
```
perdasilva, line 32: Is this still accurate? Would something like the following be more accurate?
```suggestion
// MetadataFor returns a BundleMetadata for the given bundle name and version. If the provided bundle version is nil,
```

internal/bundleutil/bundle.go
```diff
@@ -0,0 +1,39 @@
...
+// this function panics. It is up to the caller to ensure that the bundle is non-nil.
+func MetadataFor(bundleName string, bundleVersion *bsemver.Version) *ocv1alpha1.BundleMetadata {
```
perdasilva, line 34: would it make sense to wrap the bundle in an internal struct with this added method? I can see both ways working, I just don't know how "first class" this method needs to be.

internal/bundleutil/bundle.go
```diff
@@ -0,0 +1,39 @@
...
+
+	bsemver "github.com/blang/semver/v4"
+
+	"github.com/operator-framework/operator-registry/alpha/declcfg"
+	"github.com/operator-framework/operator-registry/alpha/property"
+
+	ocv1alpha1 "github.com/operator-framework/operator-controller/api/v1alpha1"
+)
+
+func GetVersion(b declcfg.Bundle) (*bsemver.Version, error) {
```
perdasilva, line 15: would it make sense for this to be part of the Bundle defn? I mean, `property` is also defined in the o-r repo. Or is this to short-circuit having to cut another release, and we'll update later?

perdasilva (dismissed):
lgtm to me - just a couple of small questions but nothing that should block this (though some tracking tickets if we want to address anything would be good =))

perdasilva:
also it may impact #1030 @bentito

internal/bundleutil/bundle.go
joelanford: `catalogmetadata.Bundle` was the wrapper that this PR removed, somewhat purposefully. What I had noticed is that we started losing parity with the FBC data model (we put channel membership and deprecation information in that wrapper). And that seemed like a recipe for more spaghetti-making when FBC evolves new types and semantics.

So I opted for functions over `declcfg.Bundle` rather than a wrapper type with methods.

internal/bundleutil/bundle.go
joelanford: I don't think we have methods (or at least it wasn't my intention to have methods) on the `declcfg.*` types. Same reasoning that there are conventions in Kubernetes API types to avoid methods, which is that methods typically pull in more dependencies that clients may want to avoid.

In the future, I 100% want to make `Version` a first class field on the `BundleV2` type that we eventually add.

joelanford:
Most recent push:
1. Rebases to `main`
2. Changes `bundleutil.MetadataFor` to fix GoDoc and avoid accepting a pointer to a blang semver version.

internal/bundleutil/bundle.go
joelanford: Addressed this in two ways:
1. Added `name and version`
2. Dropped the second sentence because I changed the signature to avoid nil-pointer panic possibilities.

dtfranz (approved):
/lgtm 
Looks like this will make the codebase a lot more manageable!

Issue #973, opened 2024-06-25
Wire up installation of content using a provided ServiceAccount


everettraven:
Following #971 and #972, we need to wire up the logic such that a `ServiceAccount` referenced in a `ClusterExtension` is used to install/upgrade/uninstall content via the Helm client.

While exact implementation may vary, here are some things to consider during implementation:
- The type implemented in #972 could likely be used to create a [`client.RestConfigMapper`](https://pkg.go.dev/github.com/operator-framework/helm-operator-plugins@v0.2.1/pkg/client#RestConfigMapper) that is used with [`client.NewActionConfigGetter`](https://pkg.go.dev/github.com/operator-framework/helm-operator-plugins@v0.2.1/pkg/client#NewActionConfigGetter) to configure the helm client created for a given ClusterExtension.
    - You can see where the `client.NewActionConfigGetter` is configured here: https://github.com/operator-framework/operator-controller/blob/2eca31ddd520a558b063880b5db0bb3ece878a3b/cmd/manager/main.go#L168-L175

**Acceptance Criteria**:
- The `client.NewActionConfigGetter` setup in https://github.com/operator-framework/operator-controller/blob/2eca31ddd520a558b063880b5db0bb3ece878a3b/cmd/manager/main.go#L168-L175 is updated to use a `client.RestConfigMapper` that creates a [`rest.Config`](https://pkg.go.dev/k8s.io/client-go/rest#Config) configured with a token from the `ServiceAccount` referenced in a `ClusterExtension`
    - The token for this should be retrieved via the implementation in #972 
 - Updates to the existing unit + e2e tests as necessary for them to continue functioning as expected. It is anticipated that some work will need to be done to configure a ServiceAccount with appropriate permissions to be used during e2e tests.
- Permissions on the operator-controller ServiceAccount should be updated to no longer require write permissions on content to be installed (and clean up any other permissions that are no longer necessary)
- Any changes to the previously implemented interfaces to facilitate the wiring of components successfully are made

joelanford:
I noticed in #1038 that we use a `*/*/*` ServiceAccount in tests. Can we change those permissions to the minimum necessary ones based on the bundle contents to make sure that future changes we make don't impose more permissions requirements?

everettraven:
Closing this one as completed as #1038 was merged.

@joelanford Your comment should be addressed via #1074

PR #1038, opened 2024-07-11
:sparkles: Wire up Service Account

theishshah:
# Description

#973 Add functionality to use Service Account

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -187,7 +187,7 @@ func main() {
 		if err != nil {
 			return nil, fmt.Errorf("Failed to extract SA token, %w", err)
 		}
-		tempConfig := rest.CopyConfig(c)
+		tempConfig := rest.AnonymousClientConfig(c)
 		tempConfig.BearerToken = token
 		return tempConfig, nil
 	}
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -40,16 +40,13 @@ import (
 
 	catalogd "github.com/operator-framework/catalogd/api/core/v1alpha1"
 	helmclient "github.com/operator-framework/helm-operator-plugins/pkg/client"
-	registryv1handler "github.com/operator-framework/rukpak/pkg/handler"
-	crdupgradesafety "github.com/operator-framework/rukpak/pkg/preflights/crdupgradesafety"
-	"github.com/operator-framework/rukpak/pkg/provisioner/registry"
-	"github.com/operator-framework/rukpak/pkg/source"
-	"github.com/operator-framework/rukpak/pkg/storage"
+
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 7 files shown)


cmd/manager/main.go
```diff
@@ -167,9 +172,34 @@ func main() {
...
+			return nil, fmt.Errorf("Failed to extract SA token, %w", err)
+		}
+		tempConfig := rest.AnonymousClientConfig(c)
+		tempConfig.BearerToken = token
+		return tempConfig, nil
+	}
 	cfgGetter, err := helmclient.NewActionConfigGetter(mgr.GetConfig(), mgr.GetRESTMapper(),
 		helmclient.StorageNamespaceMapper(installNamespaceMapper),
 		helmclient.ClientNamespaceMapper(installNamespaceMapper),
+		helmclient.RestConfigMapper(restConfigMapper),
```
joelanford, line 202: If this lands soon, all good. But just wanted to let you know that I'm working on making REST config mapping handle storage and client interactions separately. The intent is that we would use our own service account for storage (since that is an implementation detail of our controller) and we would only use the ClusterExtension's service account for the helm client's REST config.

See:
- https://github.com/operator-framework/helm-operator-plugins/pull/350
- https://github.com/operator-framework/operator-controller/pull/1057

everettraven: Thanks for the heads up! I spoke with @theishshah and I think we will focus on getting this in using the soon-to-be-deprecated `RestConfigMapper` option and as a follow up address this (if the helm-operator-plugins changes land before this gets in)

tmshort (approved):
/lgtm

internal/controllers/clusterextension_controller.go
```diff
@@ -113,11 +113,12 @@ type Preflight interface {
...
 }
 
-//+kubebuilder:rbac:groups=olm.operatorframework.io,resources=clusterextensions,verbs=get;list;watch
+//+kubebuilder:rbac:groups=olm.operatorframework.io,resources=clusterextensions,verbs=get;list;watch;update;patch
 //+kubebuilder:rbac:groups=olm.operatorframework.io,resources=clusterextensions/status,verbs=update;patch
 //+kubebuilder:rbac:groups=olm.operatorframework.io,resources=clusterextensions/finalizers,verbs=update
 //+kubebuilder:rbac:groups=core,resources=secrets,verbs=create;update;patch;delete;get;list;watch
-//+kubebuilder:rbac:groups=*,resources=*,verbs=*
+//+kubebuilder:rbac:groups=core,resources=serviceaccounts/token,verbs=create
+//+kubebuilder:rbac:groups=apiextensions.k8s.io,resources=customresourcedefinitions,verbs=get
```
everettraven, line 121: For other reviewers, the serviceaccounts/token permissions were required to get tokens for a provided SA and the customresourcedefinitions permissions are for the CRD Upgrade Safety preflight checks

everettraven (approved):

PR #1043, opened 2024-07-12
⚠ Move all packages from /pkg to /internal

kevinrizza:
Ahead of an initial 1.0.0 release, we shouldn't preemptively expose any packages publicly. Instead, let's start my making any packages internal, and we can expose them publicly later on. This commit moves everything currently in /pkg to /internal for now.


tmshort:
We have a bunch of other commits (i.e. @dtfranz's import of rukpak) that we may want to wait on before we do this.

tmshort:
@kevinrizza, this needs to be rebased as #1032 was merged

tmshort:
@kevinrizza are you going update this?

cmd/manager/main.go
```diff
@@ -57,8 +59,6 @@ import (
 	"github.com/operator-framework/operator-controller/internal/rukpak/preflights/crdupgradesafety"
 	"github.com/operator-framework/operator-controller/internal/rukpak/source"
 	"github.com/operator-framework/operator-controller/internal/version"
-	"github.com/operator-framework/operator-controller/pkg/features"
-	"github.com/operator-framework/operator-controller/pkg/scheme"
```
tmshort, line 61: The location of these shouldn't have changed. This will likely fail lint.

perdasilva: fixed!

tmshort (approved):
/lgtm
API diff is expected as components are being moved from `pkg` to `internal`

PR #1050, opened 2024-07-15
🐛 improve poor performance of helm chart conversion

joelanford:
# Description

A few ways that performance is improved:
1. Remove the unnecessary `storage` abstraction. We can convert directly from the unpacked bundle FS.
2. Use standard tooling from `cli-runtime` to parse manifests from the registry+v1 bundle (e.g. don't read file into buffer and then parse the buffer)
3. Don't write to a `plain` FS and then turn around and immediately re-parse the plain FS. Skip the plain FS and generate the chart directly.
4. Use JSON encoding rather than YAML encoding. The YAML encoding traverses JSON anyway, so this just removes an unnecessary step. Also, the Helm client has a performance benefit later by not having to convert YAML to JSON when reading the chart we generate.

Other ancillary changes:
1. Put unpacked catalogs in a separate `<cacheDir>/catalogs` directory (rather than directly in `<cacheDir>`).
2. Add debug logging and nanosecond-level granularity in log lines.
3. Remove a bunch of unused code from rukpak (handlers, provisioners, storage implementations)

# Motivation

Performance improvements of the reconcile function, and thus a more reactive system.

Locally, I noticed e2e drops from 323s to 231s.

## Reviewer Checklist

- [ ] API Go Documentation


tmshort:
And it already needs a rebase...

cmd/manager/main.go
```diff
@@ -161,7 +158,12 @@ func main() {
 	}
 
 	cl := mgr.GetClient()
-	catalogClient := catalogclient.New(cl, cache.NewFilesystemCache(cachePath, httpClient))
+	catalogsCachePath := filepath.Join(cachePath, "catalogs")
+	if err := os.MkdirAll(catalogsCachePath, 0700); err != nil {
```
tmshort, line 162: Nothing that the previous mode (for `bundles`) was `0755`, does this make a difference?

internal/controllers/clusterextension_controller.go
```diff
@@ -212,6 +212,9 @@ func checkForUnexpectedFieldChange(a, b ocv1alpha1.ClusterExtension) bool {
 */
 //nolint:unparam
 func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alpha1.ClusterExtension) (ctrl.Result, error) {
+	l := log.FromContext(ctx)
```
tmshort, line 215: Did you want to put the name of the `ClusterExtension` here?

tmshort: Possibly, also attach the V(1) to this. e.g.
```suggestion
	l := log.FromContext(ctx).V(1).AddValues("ClusterExtension", ext.GetName())
```

internal/rukpak/convert/registryv1.go
```diff
@@ -53,92 +53,73 @@ func RegistryV1ToPlain(rv1 fs.FS, installNamespace string, watchNamespaces []str
...
-			ModTime: now,
-		},
-		"manifests/manifest.yaml": &fstest.MapFile{
-			Data:    manifest.Bytes(),
-			Mode:    0644,
-			ModTime: now,
-		},
+		hash := sha256.Sum256(jsonData)
+		chrt.Templates = append(chrt.Templates, &chart.File{
+			Name: fmt.Sprintf("object-%x.json", hash[0:8]),
```
tmshort, line 117: Is this enough of the hash? Should it be longer?

internal/rukpak/convert/registryv1.go
```diff
@@ -53,92 +53,73 @@ func RegistryV1ToPlain(rv1 fs.FS, installNamespace string, watchNamespaces []str
...
-			obj := unstructured.Unstructured{}
-			err := dec.Decode(&obj)
-			if errors.Is(err, io.EOF) {
-				break
-			}
-			if err != nil {
-				return nil, fmt.Errorf("read %q: %v", e.Name(), err)
+		defer func() {
+			if err := manifestFile.Close(); err != nil {
+				l.Error(err, "error closing file", "path", path)
```
tmshort, line 73: This _appears_ to be the only place `l` is used. Should `l := log.FromContext(ctx)` be moved down here for locality?

tmshort: (or even into the WalkDir function?

tmshort (commented):
A few questions.

internal/rukpak/convert/registryv1.go
joelanford: If we did that, we'd have to repeatedly ask the context for the same logger over and over. It feels better (to me at least) to get the "logger to be used in the scope of this function" once, right at the very beginning of the function.

internal/controllers/clusterextension_controller.go
joelanford: The name is already part of the context provided in the context that arrives from the caller of `Reconcile`.

I thought about using `V(1)`, but that would mean that no logging (with that logger at least) could ever happen at the INFO level.

cmd/manager/main.go
joelanford: It shouldn't. We don't expect any other users or groups to read these files.

internal/rukpak/convert/registryv1.go
joelanford: We can make it longer, but I didn't change that aspect of this code. The name is the same as before except for `.json` instead of `.yaml`.

internal/controllers/clusterextension_controller.go
joelanford: I was going to suggest that I could do something like this:
```
l := log.FromContext(ctx)
debugLog := l.V(1)
```

But that would be misleading because levels build (i.e. `.V(1).V(1)` is equivalent to `.V(2)`), so if the caller set a `.V(1)` logger in the context, the this would suddenly _not_ be a debug logger.

internal/rukpak/convert/registryv1.go
```diff
@@ -53,92 +53,73 @@ func RegistryV1ToPlain(rv1 fs.FS, installNamespace string, watchNamespaces []str
...
 
-	var objects []*unstructured.Unstructured
 	const manifestsDir = "manifests"
-
-	entries, err := fs.ReadDir(rv1, manifestsDir)
-	if err != nil {
-		return nil, err
-	}
-	for _, e := range entries {
+	if err := fs.WalkDir(rv1, manifestsDir, func(path string, e fs.DirEntry, err error) error {
```
joelanford, line 57: There's another potential optimization here, where we spin up `runtime.NumCPU` goroutines and parse the files concurrently.  But that can happen as a follow-up.

tmshort (approved):
/lgtm

PR #1052, opened 2024-07-15
:sparkles: Use x509.CertPools instead of PEM strings

### LINKED ISSUE #999: Cleanup use of PEM with certificates
Currently, many of the interfaces used with certificates within rukpak/operator-controller depend on the use of large PEM strings. It would be better to use CertPool types/APIs.

tmshort:
Fixes #999

# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/controllers/clusterextension_controller.go
+++ b/internal/controllers/clusterextension_controller.go
@@ -19,6 +19,7 @@ package controllers
 import (
 	"bytes"
 	"context"
+	"crypto/x509"
 	"errors"
 	"fmt"
 	"io"
@@ -69,7 +70,6 @@ import (
 	catalogfilter "github.com/operator-framework/operator-controller/internal/catalogmetadata/filter"
 	catalogsort "github.com/operator-framework/operator-controller/internal/catalogmetadata/sort"
 	"github.com/operator-framework/operator-controller/internal/conditionsets"
-	"github.com/operator-framework/operator-controller/internal/httputil"
 	"github.com/operator-framework/operator-controller/internal/labels"
 	"github.com/operator-framework/operator-controller/internal/rukpak/bundledeployment"
 	"github.com/operator-framework/operator-controller/internal/rukpak/convert"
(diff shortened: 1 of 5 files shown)


internal/controllers/clusterextension_controller.go
```diff
@@ -273,9 +273,9 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
 	// Generate a BundleDeployment from the ClusterExtension to Unpack.
 	// Note: The BundleDeployment here is not a k8s API, its a simple Go struct which
 	// necessary embedded values.
-	bd := r.generateBundleDeploymentForUnpack(ctx, bundle.Image, ext)
+	bd := r.generateBundleDeploymentForUnpack(bundle.Image, ext)
 	l.V(1).Info("unpacking resolved bundle")
-	unpackResult, err := r.Unpacker.Unpack(ctx, bd)
+	unpackResult, err := r.Unpacker.Unpack(ctx, bd, r.CaCertPool)
```
joelanford, line 278: Is this the only place that `r.CaCertPool` is used in the reconciler? If so, WDYT about moving this to a field on the `Unpack` implementation's struct?

tmshort: (I had problems with mocks, so the more I can stay away from them...)

internal/httputil/httputil.go
```diff
@@ -6,48 +6,40 @@ import (
...
 		}
 		data, err := os.ReadFile(filepath.Join(caDir, e.Name()))
 		if err != nil {
-			return "", err
+			return nil, err
 		}
-		certs = append(certs, string(data))
+		// ignore ok - which just indicates if any cert was added
+		// This will ignore non-certs, which will return !ok
+		caCertPool.AppendCertsFromPEM(data)
```
joelanford, line 35: If we get `!ok`, should we error instead of silently ignoring?

Seems like we could construct a fairly nice error message like "failed to add certs from <filename>: invalid certificate data".

A cert that missed a trailing character or that accidentally included an errant whitespace character would be considered a non-cert by this code and would make troubleshooting a lot more difficult.

tmshort: The problem is that we can't tell a "badly formatted cert" from a "someone put a text file in here that we should ignore". I would say we should log an error, if anything, not error out.

joelanford (approved):

PR #1053, opened 2024-07-15
:seedling: Add unit tests for loading certs

tmshort:
# Description

## Reviewer Checklist

- [ ] API Go Documentation


internal/httputil/httputil_test.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/operator-framework/operator-controller/internal/httputil"
+)
+
+// The "good" test consists of 3 Amazon Root CAs, along with a "PRIVATE KEY" in one of the files
+// The "bad" test consists of 2 Amazon Root CAs, the second of which is gargage, and the test fails
```
perdasilva, line 12:
```suggestion
// The "bad" test consists of 2 Amazon Root CAs, the second of which is garbage, and the test fails
```

internal/httputil/httputil_test.go
```diff
@@ -0,0 +1,40 @@
...
+// The "ugly" test consist of a single file:
+// - Amazon_Root_CA_1
+// - garbage PEM
+// - Amazon_Root_CA_3
+// The error is _not_ detected because the PEM decoder skips right over the garbage
+// This demonstrates the danger of putting multiple certificates into a single file
```
perdasilva, line 18: Is this because it would violate the expectation of getting an error if _any_ of the certs are troublesome? Is this something we should document in some user-facing location? Maybe in some best-practices doc? Or, if we're not there yet, maybe we should create a tracking ticket. Seems useful to know as a user.

perdasilva (approved):
lgtm - just a small comment around documenting the wisdom imparted by the tests (or a ticket to do so)

internal/httputil/httputil_test.go
tmshort: Let me fix my native language 🤦

internal/httputil/httputil_test.go
tmshort: Right. We would expect that cert number 2 would throw an error, but the PEM decoder basically skips it. Test "bad" has the garbage certificate by itself in its own file, so the PEM decode errors out.

perdasilva (approved):
lgtm

internal/httputil/httputil_test.go
perdasilva: another idea (if we want to "go the extra mile") would be to reject multi-cert files straight off the bat, or split them into individual files in some tempfs or memfs and load the certs from there? (nothing we need for now - just spit balling)

tmshort: Or, we could to a minimal PEM decode ourself, rather than relying on `pem.Decode()`, but probably not worth the effort here.

Issue #923, opened 2024-06-12
BUG: data too long issue rendering


DrummyFloyd:
### User story

issue created , due to mention in [slack](https://kubernetes.slack.com/archives/C0181L6JYQ2/p1718174701411119?thread_ts=1718130282.869949&cid=C0181L6JYQ2)

as a recurent tester on this project, ( because i like it) 
 i test upon some wanted operator on my stack.
 
 atm i have some issue with following manifest with `0.10` OLM
 
 ```yaml
apiVersion: catalogd.operatorframework.io/v1alpha1
kind: Catalog
metadata:
  name: operatorhubio
spec:
  source:
    type: image
    image:
      ref: quay.io/operatorhubio/catalog:latest
      pollInterval: 24h

---
apiVersion: olm.operatorframework.io/v1alpha1
kind: ClusterExtension
metadata:
  name: op-mariadb
spec:
  installNamespace: operators
  packageName: mariadb-operator
  version: 0.29.0
 ```

### List if issues

error message rendered.

````
InstallationFailed:create: failed to create: Secret "sh.helm.release.v1.op-mariadb.v1" is invalid: data: Too long: must have at most 1048576 bytes
```

### New content or update?

Net-new content

### Types of documentation

_No response_

### References

_No response_

joelanford:
Thanks for submitting this issue!

There are two things that cause this bug:
1. mariadb-operator has some very large CRDs
2. Helm tries to store the entire release inside of a single secret, and etcd has a size limit for what it can store.

tmshort:
We used tar + gz to create some of these resources, and that might be necessary here... unless even that is too big.

acornett21:
This seems really strange to me that we are offloading CRD creation/management to `helm` when we know that `helm` itself can't manage the install/update of CRD's. I was under the impression (though it appears wrong), that OLMv1 was going to mange the creation of CRD's outside the context (knowledge) of `helm`. It seems that this was only in the context of `namespace scoped` installs, but even still that seems illogical, that we'd have two different control paths/flows within OLMv1.

Should we possibly look into a 4th option of managing the CRD's ourselves, or is that out of scope(effort/timeline)?

varshaprasad96:
Adding some thoughts based on initial findings:

Option 3:
The options available to us in terms of possible supported Helm backends are: Secrets, Configmaps and SQL.
[...]

Issue #1028, opened 2024-07-09
[epic]: Minimal controls for selecting catalogs to resolve from


joelanford:
Today, if multiple ClusterCatalogs exist on a cluster, they contain duplicate package names, and both catalogs contain bundles the meet the criteria of a ClusterExtension, the choice that our resolution process makes is currently undefined.

In this epic, we propose to provide users with one minimal control for catalog selection.

We would add a field to the ClusterExtension spec that gives users the ability to select over a subset of the catalogs to resolve from. This is intended to give users a similar experience as yum --enable-repo and yum --disable-repo.

In a design, we will propose an implementation that accounts for API conventions around standard patterns and anti-patterns for object selection.

- [Brief](https://docs.google.com/document/d/1-SqENbwa3nrbYqrxaD7OfjyLXWdjuYBCUUEJPPA8r7U/edit#heading=h.btd8kkom5km4)
- [RFC - Labels](https://docs.google.com/document/d/1v9iPRZHt1YXhwK9__7OITnEN430lryQzWaAaIpn---Y/edit)
- [RFC - Priority](https://docs.google.com/document/d/1jGAOvE0yf_U2lpyhy2U0KPLGOaG-8-ML0lve1mR-o1Q/edit)

Sub tasks:
- [x] #1110
- [x] #1111
- [x] #1112
- [x] #1113
- [x] #1114
- [x] #1115
- [x] #1116
- [x] #1117
- [x] #1154
- [x] https://github.com/operator-framework/catalogd/issues/370
- [x] https://github.com/operator-framework/catalogd/issues/371

joelanford:
Brief: https://docs.google.com/document/d/1-SqENbwa3nrbYqrxaD7OfjyLXWdjuYBCUUEJPPA8r7U/edit#heading=h.btd8kkom5km4

PR #1057, opened 2024-07-16
⚠️ helm: use chunking release driver in systemNamespace

joelanford:
# Description

This PR:
1. uses a new helm release storage driver from `helm-operator-plugins` that chunks release data into one or more secrets (this also incidentally means that our client and the Helm CLI client won't mistake each other's release storage for their own).
2. moves release secret storage back to our system namespace so that we don't leak details of our release bookkeeping to users.
3. Notes that the new helm-operator-plugins commit also makes it possible to use one service account for release storage and a separate service account for deploying bundle contents. This is necessary if we want to avoid leaking our internal release storage mechanism to users (who would otherwise have to configure RBAC for release storage permission)

Fixes: https://github.com/operator-framework/operator-controller/issues/923

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -30,10 +30,8 @@ import (
 	apiextensionsv1client "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1"
 	k8slabels "k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/selection"
-	"k8s.io/apimachinery/pkg/types"
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
-	"k8s.io/client-go/rest"
 	ctrl "sigs.k8s.io/controller-runtime"
 	crcache "sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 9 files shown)


internal/httputil/httputil.go
```diff
@@ -7,7 +7,7 @@ import (
 )
 
 func BuildHTTPClient(cpw *CertPoolWatcher) (*http.Client, error) {
-	httpClient := &http.Client{Timeout: 10 * time.Second}
+	httpClient := &http.Client{Timeout: 2 * time.Minute}
```
joelanford, line 10: Now that we have the operatorhubio catalog installed by default, our e2e tests are entering and staying in exponential backoff because it takes longer for the operatorhubio catalog to be pulled, unpacked, and cached than the test catalog does.

Once we have #1028, we could consider excluding the operatorhubio catalog from resolution during these e2es. Even so, we likely want this value bumped up a bit to handle the default catalog without requests timing out.

go.mod
```diff
@@ -17,7 +17,7 @@ require (
 	github.com/onsi/gomega v1.34.0
 	github.com/operator-framework/api v0.26.0
 	github.com/operator-framework/catalogd v0.20.0
-	github.com/operator-framework/helm-operator-plugins v0.3.1
+	github.com/operator-framework/helm-operator-plugins v0.3.1-0.20240722173643-2e18c5b70f9c
```
tmshort, line 20: release?

joelanford: Yeah, talking about it over on Kubernetes Slack, here: https://kubernetes.slack.com/archives/C0181L6JYQ2/p1722268675418209?thread_ts=1722268591.413059&cid=C0181L6JYQ2

cmd/manager/main.go
```diff
@@ -161,40 +159,21 @@ func main() {
...
-		}
-		return tempConfig, nil
-	}
 	cfgGetter, err := helmclient.NewActionConfigGetter(mgr.GetConfig(), mgr.GetRESTMapper(),
-		helmclient.StorageNamespaceMapper(installNamespaceMapper),
-		helmclient.ClientNamespaceMapper(installNamespaceMapper),
-		helmclient.RestConfigMapper(restConfigMapper),
+		helmclient.StorageDriverMapper(action.ChunkedStorageDriverMapper(coreClient, mgr.GetAPIReader(), systemNamespace)),
+		helmclient.ClientNamespaceMapper(func(obj client.Object) (string, error) {
+			ext := obj.(*ocv1alpha1.ClusterExtension)
```
tmshort, line 173: could this fail? do you need an `ok` on this type conversion?

tmshort: (Yes, I see it's identical to the prior code that was defined previously)

config/base/rbac/role_binding.yaml
```diff
@@ -10,3 +10,17 @@ subjects:
...
+  name: manager-rolebinding
+  namespace: system
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: manager-role
+subjects:
+  - kind: ServiceAccount
+    name: controller-manager
+    namespace: system
```
acornett21, line 26: Nit: No new line at the end of this file.

cmd/manager/main.go
joelanford: If this fails, we did something horribly wrong (i.e. we used the mapper function in a reconciler for a different type), so a panic is reasonable, IMO. There is nothing a user could do to fix it.

config/base/rbac/role_binding.yaml
joelanford: Fixed here: https://github.com/operator-framework/operator-controller/commit/bebe28ccf7b4808e046401a36b8aa6af34fdd415

tmshort (approved):
/lgtm
Looks reasonable to me
(Added a comment later about a TODO though)

internal/httputil/httputil.go
tmshort: Can you add a TODO to revert this later?

tmshort (changes requested):
OK, I want the TODO in there about the timeout.
I would prefer a release of helm-operator-plugins, failing that, perhaps a TODO.

go.mod
acornett21: You're now unblocked here, see below
https://github.com/operator-framework/helm-operator-plugins/releases/tag/v0.4.0

tmshort:
And... a rebase is needed...

joelanford:
Ok, rebased and reverted the catalog client timeout change, so it is back to 10s. Looks like e2e still passes in GH CI.

I'm also seeing `main` fail the same as locally for me as this branch, so let's chalk this up as a "me" problem for now.

tmshort (approved):
/lgtm

tmshort (approved):
/lgtm

internal/httputil/httputil.go
joelanford: I reverted this. Seems to be a problem locally for me only.

Issue #1026, opened 2024-07-09
Operator Controller Should Provide a Standard Install Process


acornett21:
Today the install is done from `yamls` from the the release pages, which is less the ideal for end users. We should mimic the install process of OLMv0 (as far as steps go, the mechanism could be different). Sample process below:

1. Install `dependencies`
2. Install `operator-controller`
3. Install the same catalog from OperatorHub.io that OLMv0 uses.

joelanford:
I actually don't think the `curl <releaseScript> | bash` is all that bad of an installation process. The fact that that's the only place we have it documented isn't ideal though. I'd almost split this into 3 different things to do, in order:
1. Add a `ClusterCatalog` for `quay.io/operatorhubio/catalog:latest` to our existing install script.
2. Improve our docs (and also docs process/developer awareness) to have a more user-centric (rather than developer-centric) place to find installation instructions.
3. (maybe) Build tooling into a common tool (maybe kubectl-operator?) that can automate installation/upgrade of OLMv1.

trgeiger:
Item number 1 from Joe's message is addressed by MR 1079. Not sure if we should break items 2 and 3 out into separate issues and close this one or if we should track those here.

DrummyFloyd:
hi , from what i understand here, you're talking about how OLMV1 will be installed ? 
because OLMV1, fill the gitops gap from the V0 , you shoudl definitely , add an Helm release of the Operator
could be use with some tools like FLuxCD or Argocd

djmcgreal:
Is running curl and bash from terraform the best option? If so, it’s far from ideal.

PR #1060, opened 2024-07-17
✨ Apply OperatorHub.io ClusterCatalog during installation

trgeiger:
This matches the default configuration in OLMv0 by including OperatorHub

# Description
This change simply adds the OperatorHub ClusterCatalog to the installation script. On a fresh installation, users will now have the same default operator listings as in OLMv0.

Right now I just have the script applying the sample YAML from this repository. If we wanted to match the way we install actual releases of Cert Manager and Catalogd, then I could make a PR in Catalogd to add the OperatorHub catalog yaml to the release artifacts and cut a release of Catalogd for those to be included.

Related issue: #1026

I'll follow up this PR with documentation changes as suggested in 1026

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/scripts/install.tpl.sh
+++ b/scripts/install.tpl.sh
@@ -38,5 +38,8 @@ kubectl_wait "cert-manager" "deployment/cert-manager-webhook" "60s"
 kubectl apply -f "https://github.com/operator-framework/catalogd/releases/download/${catalogd_version}/catalogd.yaml"
 kubectl_wait "olmv1-system" "deployment/catalogd-controller-manager" "60s"
 
+kubectl apply -f "https://raw.githubusercontent.com/operator-framework/operator-controller/main/config/samples/catalogd_operatorcatalog.yaml"
+kubectl wait --for=condition=Unpacked "clustercatalog/operatorhubio" --timeout="60s"
+
 kubectl apply -f "${operator_controller_manifest}"
 kubectl_wait "olmv1-system" "deployment/operator-controller-controller-manager" "60s"


scripts/install.tpl.sh
```diff
@@ -38,5 +38,7 @@ kubectl_wait "cert-manager" "deployment/cert-manager-webhook" "60s"
 kubectl apply -f "https://github.com/operator-framework/catalogd/releases/download/${catalogd_version}/catalogd.yaml"
 kubectl_wait "olmv1-system" "deployment/catalogd-controller-manager" "60s"
 
+kubectl apply -f "https://raw.githubusercontent.com/operator-framework/operator-controller/main/config/samples/catalogd_operatorcatalog.yaml"
+
 kubectl apply -f "${operator_controller_manifest}"
-kubectl_wait "olmv1-system" "deployment/operator-controller-controller-manager" "60s"
+kubectl_wait "olmv1-system" "deployment/operator-controller-controller-manager" "60s"
```
itroyano, line 44: nit: missing newline

scripts/install.tpl.sh
```diff
@@ -38,5 +38,8 @@ kubectl_wait "cert-manager" "deployment/cert-manager-webhook" "60s"
 kubectl apply -f "https://github.com/operator-framework/catalogd/releases/download/${catalogd_version}/catalogd.yaml"
 kubectl_wait "olmv1-system" "deployment/catalogd-controller-manager" "60s"
 
+kubectl apply -f "https://raw.githubusercontent.com/operator-framework/operator-controller/main/config/samples/catalogd_operatorcatalog.yaml"
```
everettraven, line 41: I'm not sure that we should utilize the samples here as the default catalogs, despite this being one that utilizes the OperatorHub catalog. 

What do you think about adding officially maintained defaults to catalogd? This could take the form of plain YAML or another kustomize overlay, but I think the benefits of this is that the default `ClusterCatalog` resources are co-located with the code that would be modified when changes are made to the API.

trgeiger: That sounds like a good idea to me, I suggested something like that in the description of the PR. I can close this PR and make those changes there. Are you thinking something where the catalog is installed by default when Catalogd is installed, and then operator-controller would just inherit that when it installs Catalogd?

PR #1062, opened 2024-07-17
:seedling: Add support for CA/certificate rotation

### LINKED ISSUE #915: Handle CA rotation for Catalogd web server trust
With the addition of TLS communication between operator-controller and catalogd, [we now mount](https://github.com/operator-framework/operator-controller/pull/888) the CA certificate into operator-controller so it can trust connections to the catalogd web server.

It would be ideal if we could handle rotations of the mounted CA certificate without restarting the manager. As of now, the contents of the mounted CA certificate *do* change on disk when the certificate rotates, but the manager does not pick up any changes since Go [does not have a mechanism](https://github.com/golang/go/issues/35887#issuecomment-1798581909) for reloading rootCA.

There is the possibility to implement custom connection verification logic in [tls.Config.VerifyConnection()](https://pkg.go.dev/crypto/tls#example-Config-VerifyConnection), but for this to work you would also need to set InsecureSkipVerify to true which isn't ideal. The ideal solution would probably be to contribute something akin to GetCertificate but for the rootCA upstream into Go and then use that feature in operator-controller.

It might also be worth modifying the CertificateWatcher in controller-runtime so it can function without providing a key--currently it requires both the cert and key since it's built for managing rotation on servers. A GetRootCA-type functionality for re-loading the CA cert in tls.Config would utilize the modified CertificateWatcher in a similar manner to how the certificate rotation is [handled in Catalogd](https://github.com/operator-framework/catalogd/blob/main/internal/serverutil/serverutil.go#L33).

tmshort:
Fixes #915
Mounted secrets are automatically updated into pods, but...
* It doesn't work with `subPath` mountings
* When `subPath` is not used, then a bunch of directories are mounted
* And one of those directories is a symlink, so `IsDir()` returns false
* And a watch is needed to notice the change

So, update the certificate volume patch, which requires a change in how we look for certificates in the CA cert directory.

Add a watch, so when the certs do change, we update the cert pool.
Also look at validity dates of certificates, and error on expired certs.

The default cert-manager certificates have 90 days validities.

# Description

## Reviewer Checklist

- [ ] API Go Documentation


tmshort:
This also adds a unit-test for an empty cert dir, and for an expired certificate. And also adds an e2e to make sure secret updates and rotations actually work.

cmd/manager/main.go
```diff
@@ -176,11 +176,15 @@ func main() {
 		os.Exit(1)
 	}
 
-	certPool, err := httputil.NewCertPool(caCertDir)
+	certPool, err := httputil.NewCertPool(caCertDir, setupLog, func() {
+		setupLog.WithName("cert-pool").Info("restarting to reload CA certificates")
+		os.Exit(1)
```
everettraven, line 181: I'm a bit concerned with this approach from the perspective of a user that might see the pod having restarts and suspecting it is a bug, filing a bug report, and then us as maintainers having to perform deeper investigation to find that it was a cert rotation and stating that it is expected behavior for the pod to "crash" to reload CA certificates.

While it may be easier to simply crash the pod and it be restarted, I think it will throw a false negative signal to users. I think a caching based approach, while more complex, would ensure we aren't issuing false negative signals.

tmshort: It's clear from the (prior) logs that it's a restart from cert rotation.
CA certificates should rarely, if ever, rotate. We're talking a minimum of 3 months (for default cert-manager certificates) to years for commercial CAs.
According to Steve K, this is a common approach (I specifically asked him about it in another project).

Not sure what you mean by a cache-based approach? These are files mounted in the pod itself, not Secrets or ConfigMaps that are being monitored.

cmd/manager/main.go
```diff
@@ -259,6 +267,21 @@ func main() {
...
+	setupLog.Info("adding certificate watch", "directory", caCertDir)
+	if err := httputil.AddCertWatch(caCertDir, func() error {
+		ctrl.Log.WithName("cert-pool").Info("reloading certificate pool")
+		certPool, err := httputil.NewCertPool(caCertDir, ctrl.Log.WithName("cert-pool"))
+		if err != nil {
+			return err
+		}
+		ceRecon.CaMutex.Lock()
+		defer ceRecon.CaMutex.Unlock()
+		ceRecon.CaCertPool = certPool
```
everettraven, line 279: Instead of updating the cert pool on the ClusterExtensionReconciler, what if we instead used a type that cached this information and that object was passed around?

So instead of having to do something like:
```go
	ceRecon := &controllers.ClusterExtensionReconciler{}

	unpacker := &source.ImageRegistry{
		BaseCachePath: filepath.Join(cachePath, "unpack"),
		// TODO: This needs to be derived per extension via ext.Spec.InstallNamespace
		AuthNamespace: systemNamespace,
		GetCaCertPool: func() (*x509.CertPool, error) {
			ceRecon.CaMutex.RLock()
			defer ceRecon.CaMutex.RUnlock()
			return ceRecon.CaCertPool.Clone(), nil
		},
	}
```

it could instead be:
```go
        certPooler := httputil.NewCertPooler(...)
	unpacker := &source.ImageRegistry{
		BaseCachePath: filepath.Join(cachePath, "unpack"),
		// TODO: This needs to be derived per extension via ext.Spec.InstallNamespace
		AuthNamespace: systemNamespace,
		// Used the existing approach as a way to show the example, but this could
		// probably also just be updated to be a cert pooler instead of a function
		// although keeping it as a function is also fine by me
		GetCaCertPool: func() (*x509.CertPool, error) {
			return certPooler.Get()
		},
	}
```

tmshort: Hmmm.... Would still need to store the pool somewhere, and that place right now is on the reconciler, but doesn't have to be. I can look at this,. but not sure how much "tidier" it would be.

everettraven: The pool would be stored on the `CertPoolWatcher`(?) type

hack/test/cert-e2e.sh
```diff
@@ -0,0 +1,31 @@
...
+}
+
+WAIT_TIME=0
+MAX_TIME=300
+until (( NEXT_WAIT_TIME == MAX_TIME )) || check_logs; do
+    echo -n "."
+    sleep "$(( WAIT_TIME += 10 ))"
+done
+(( WAIT_TIME < MAX_TIME ))
+echo "#"
```
m1kola, line 31: This is quite hard to follow, so I might be missing something. But it seems to me that we will never exit in the unhappy path (string never appears in logs) because we are waiting for `NEXT_WAIT_TIME` to be equal to `MAX_TIME` and `NEXT_WAIT_TIME` never defined.

m1kola: If you decide to switch to Go tests for this at some point - you might find this useful:

https://github.com/operator-framework/operator-controller/blob/95b9f0dc5943497028deb47969310dec45792a2d/test/upgrade-e2e/post_upgrade_test.go#L116-L145

internal/catalogmetadata/cache/cache.go
```diff
@@ -95,7 +95,11 @@ func (fsc *filesystemCache) FetchCatalogContents(ctx context.Context, catalog *c
 		return nil, fmt.Errorf("error forming request: %v", err)
 	}
 
-	resp, err := fsc.client.Do(req)
+	client, err := fsc.getClient()
+	if err != nil {
+		return nil, fmt.Errorf("error getting HTTP client: %w", err)
+	}
```
m1kola, line 101: As far as I understand this PR - we are creating a new client each time we need to make a request to catalogd to make sure that we read a new cert.

But if it is the case - why do we need a cert watcher?

m1kola (commented):
I'm still reviewing.

For these interested, here are some relevant issues in Go and K8s on root CA reloading:

* [Go lang proposal for root CA reloading](https://github.com/golang/go/issues/64796)
* [K8s client-go issue for root CA reloading](https://github.com/kubernetes/kubernetes/issues/119483)

TLDR: it looks like there is no equivalent of `GetClientCertificate`/`GetCertificate`, but for root CAs in TLS config in standard library, but it is possible to implement a workaround with `VerifyPeerCertificate`. This would require setting `InsecureSkipVerify` to `true` and re-implementing the standard verification in `VerifyPeerCertificate` as far as I understand - see [this example](https://pkg.go.dev/crypto/tls@master#example-Config-VerifyConnection) from Go docs.

hack/test/cert-e2e.sh
tmshort: As of now, it doesn't fail... but you are correct. I didn't update every variable.

tmshort:
> TLDR: it looks like there is no equivalent of `GetClientCertificate`/`GetCertificate`, but for root CAs in TLS config in standard library, but it is possible to implement a workaround with `VerifyPeerCertificate`. This would require setting `InsecureSkipVerify` to `true` and re-implementing the standard verification in `VerifyPeerCertificate` as far as I understand - see [this example](https://pkg.go.dev/crypto/tls@master#example-Config-VerifyConnection) from Go docs.


And we really don't want to do that (set `InsecureSkipVerify`) by default. And reimplementing the verification code is probably not worth it, when we can just offer up the updated certificate pool on demand.

internal/catalogmetadata/cache/cache.go
tmshort: Because we don't want to read the certs off disk every single time we want to make a connection. The cert pool is used in multiple places.

hack/test/cert-e2e.sh
```diff
@@ -0,0 +1,31 @@
...
+kubectl delete secret -n olmv1-system olmv1-cert
+
+# delete the existing operator-controller, to force it to get a new secret
+# (deleting the secret itself isn't enough)
+kubectl delete pod -n olmv1-system -l control-plane=controller-manager
+kubectl wait --for=condition=Available --namespace=olmv1-system "deployment/operator-controller-controller-manager" --timeout="60s"
+
+# and then search through the previous logs
+function check_logs() {
+    kubectl logs -c manager -n olmv1-system -l control-plane=controller-manager | grep "updating certificate pool" >& /dev/null
```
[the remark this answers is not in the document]
tmshort: You're correct in that the cert-e2e only test the reloading of certificates. It would be _much_ more complex and time-consuming to test an end-to-end connection with catalogd. The renewal period would have to be adjusted to make this test possibly run 10+ minutes, which is undesirable. Now that there's a separate CertPoolWatcher, I can likely do it as a unit test; the code before was not amenable to a unit-test.

hack/test/cert-e2e.sh
```diff
@@ -0,0 +1,31 @@
...
+kubectl patch certificate.cert-manager.io -n olmv1-system olmv1-cert --type='merge' -p '{"spec":{"duration":"1h", "renewBefore":"59m"}}'
+
+# delete the old secrets, so new secrets are generated
+kubectl delete secret -n cert-manager olmv1-ca
+kubectl delete secret -n olmv1-system olmv1-cert
+
+# delete the existing operator-controller, to force it to get a new secret
+# (deleting the secret itself isn't enough)
+kubectl delete pod -n olmv1-system -l control-plane=controller-manager
+kubectl wait --for=condition=Available --namespace=olmv1-system "deployment/operator-controller-controller-manager" --timeout="60s"
```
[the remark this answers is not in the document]
tmshort: You might be right, I can verify. But if I can use a unit test instead, this wouldn't be necessary.

internal/httputil/certpoolwatcher.go
```diff
@@ -0,0 +1,80 @@
...
+					cpw.mx.Lock()
+					defer cpw.mx.Unlock()
+					cpw.pool = pool
+				}()
+			case err = <-watcher.Errors:
+				log.Error(err, "error watching certificate dir")
+				os.Exit(1)
+			}
+		}
+	}()
```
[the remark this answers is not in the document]
tmshort: I can break it up (specifically the update function), which would make it a bit easier to understand.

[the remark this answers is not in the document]
tmshort: The complexity is there due to the number if `func()` that are present due to the use of `defer` and `return`...

internal/catalogmetadata/cache/cache.go
m1kola: Makes sense. Resolving this question.

m1kola:
> And we really don't want to do that (set `InsecureSkipVerify`) by default. And reimplementing the verification code is probably not worth it, when we can just offer up the updated certificate pool on demand.

Setting `InsecureSkipVerify` to `true` and `VerifyPeerCertificate` means that we take over the responsibility of verifying the certs from standard library. That's not great, but it is not a lot of code (see standard library [code](https://github.com/golang/go/blob/70491a81113e7003e314451f3e3cf134c4d41dd7/src/crypto/tls/handshake_client.go#L1117-L1128), and [client example](https://pkg.go.dev/crypto/tls@master#example-Config-VerifyConnection) from the docs). It doesn't sounds too terrible to me, so I'm open to consider this.

As far as I understand in this implementation we are re-creating a client each time and will have to establish a connection each we use it which is not optimal.

I don't know how significantly it will affect our performance. One one hand - creating new connections every time is not great, but on the other hand - not sure if we are going to benefit from re-using the connection in these use cases.

hack/test/cert-e2e.sh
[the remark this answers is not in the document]
tmshort: I removed the e2e, and now use a unit test

hack/test/cert-e2e.sh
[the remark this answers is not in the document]
tmshort: I removed the e2e and now use a unit test

internal/httputil/certpoolwatcher.go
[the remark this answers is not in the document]
tmshort: I broke it up, and @joelanford offered some suggests for performance, so there's still a `drain()` function, but it's isolated.

hack/test/cert-e2e.sh
tmshort: I removed the e2e in favor of a unit test

Makefile
```diff
@@ -188,6 +188,15 @@ test-upgrade-e2e: export TEST_CLUSTER_CATALOG_NAME := test-catalog
 test-upgrade-e2e: export TEST_CLUSTER_EXTENSION_NAME := test-package
 test-upgrade-e2e: kind-cluster run-latest-release image-registry build-push-e2e-catalog registry-load-bundles pre-upgrade-setup docker-build kind-load kind-deploy post-upgrade-checks kind-clean #HELP Run upgrade e2e tests on a local kind cluster
 
+.PHONY: run-cert-e2e
+run-cert-e2e:
+	./hack/test/cert-e2e.sh
+
+.PHONY: test-cert-e2e
+test-cert-e2e: KIND_CLUSTER_NAME := operator-controller-cert-e2e
+test-cert-e2e: KUSTOMIZE_BUILD_DIR := config/overlays/cert-manager
```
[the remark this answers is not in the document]
tmshort: I removed the e2e in favor of a unit test

internal/httputil/certpoolwatcher.go
```diff
@@ -0,0 +1,108 @@
...
+		dir:        caDir,
+		pool:       pool,
+		log:        log,
+		watcher:    watcher,
+		done:       make(chan bool),
+	}
+	go func() {
+		for {
+			select {
+			case <-watcher.Events:
```
bentito, line 62: Could there be spurious events in this dir that are going to cause us to do work when we don't really need to? Just wondering how clean cert manager, and anything else with access, is in terms of usage of this dir? Maybe we should filter it down further or just watch .crt files, etc.

tmshort: That's what this function is for... it's used to "drain" the event cue to avoid spurious events.
It's really just kubernetes that has write access, it updates the certs based on when a Secret is updated. The directory is read-only within the pod.

bentito: sounds like it'll be okay then

internal/httputil/certpoolwatcher.go
```diff
@@ -0,0 +1,108 @@
...
+				cpw.drainEvents()
+				cpw.update()
+			case err := <-watcher.Errors:
+				log.Error(err, "error watching certificate dir")
+				os.Exit(1)
+			case <-cpw.done:
+				err := watcher.Close()
+				if err != nil {
+					log.Error(err, "error closing watcher")
```
everettraven, line 71: What happens with the watcher in the case that the close fails? I see the goroutine for handling events exits, but does the watcher continue to stick around filling the event queue and consuming memory?

internal/httputil/certpoolwatcher.go
```diff
@@ -0,0 +1,108 @@
...
+	}()
+	return cpw, nil
+}
+
+func (cpw *CertPoolWatcher) update() {
+	cpw.log.Info("updating certificate pool")
+	pool, err := NewCertPool(cpw.dir, cpw.log)
+	if err != nil {
+		cpw.log.Error(err, "error updating certificate pool")
+		os.Exit(1)
```
everettraven, line 85: Is it expected that encountering an error when performing an update to the certificate pool that restarting the pod will result in the error being resolved?

internal/httputil/certpoolwatcher.go
```diff
@@ -0,0 +1,108 @@
...
+		pool:       pool,
+		log:        log,
+		watcher:    watcher,
+		done:       make(chan bool),
+	}
+	go func() {
+		for {
+			select {
+			case <-watcher.Events:
+				cpw.drainEvents()
```
everettraven, line 63: Do you know if any events include _only_ the directory being watched as the event.Name value? If so, I wonder if instead of performing this drainEvents action we could do some event filtering similar to https://github.com/fsnotify/fsnotify/blob/c1467c02fba575afdb5f4201072ab8403bbf00f4/cmd/fsnotify/file.go#L66-L78

I won't block the PR merging on this, but something that could make it so we don't have any "sleep" actions

tmshort: Some filtering might be useful. The only time this path should be updated is when a Secret is updated. The directory is read-only within the pod.

internal/httputil/certpoolwatcher.go
tmshort: If we can't create the cert pool, then something is wrong with the pod. The mount of the certs directory is broken, or the set of certificates is bad, or the configuration of the pod is bad. Joe has given a previous comment he'd rather have the pod crash (and be noticed) than have it continue to run in an error condition.

internal/httputil/certpoolwatcher.go
tmshort: Since the `cpw` is still likely referenced _somewhere_ the memory for the notifier is still around; but the notify OS resources have been cleaned up the _best that they can be_.

internal/httputil/certpoolwatcher.go
everettraven: >We need to recognize new files, deleted files, updated files, etc.
If we only react to "directory has been updated" type events wouldn't we catch these events as well?

everettraven (approved):

internal/httputil/certpoolwatcher.go
tmshort: But it wouldn't catch updates to files within, as that's a change to the file, not the directory.

everettraven: I thought the reasoning for the drain events operation was because when we receive updates we get mass events on everything when something changed. Maybe I misunderstood, which led me to thinking that if any change happened in the directory (including an individual file), it would trigger an event for the directory as well.

tmshort: The Watch is on the directory, and that includes the contents. It also depends on how things are mounted. A change to a file within a directory does not necessarily indicate a change to the directory.
The drain is there because the update of a single secret may trigger a number of events (I was seeing 4+, because of how the mounted files were presented), and only one reload of the certs is necessary. 
Based on my testing, if there's an update on a file, it only reports the update on that file (i.e. create/update); it doesn't trigger a second update on the directory as well.

tmshort: I added debug output to the cert watcher test for every event in the unit test:
```
=== RUN   TestCertPoolWatcher
    certpoolwatcher_test.go:72: Create cert file at "/tmp/cert-pool4285276756/test1.pem"
    certpoolwatcher_test.go:87: Create cert file at "/tmp/cert-pool4285276756/test2.pem"
Event: CREATE        "/tmp/cert-pool4285276756/test2.pem"
Event: WRITE         "/tmp/cert-pool4285276756/test2.pem"
--- PASS: TestCertPoolWatcher (1.11s)
```
So, there's an event for the create of the new PEM, and one for the write, but nothing on the directory itself. There are two events, and that would cause two reloads without the drain mechanism in place.

PR #1068, opened 2024-07-18
:sparkles: refactor ClusterExtensionReconciler.reconcile()

everettraven:
# Description
Makes the `reconcile` method on the `ClusterExtensionReconciler` easier to unit test by encapsulating distinct operations as interfaces that can be mocked:
- (Done before this PR) Resolution is now done via a `Resolver` interface.
    - (Done as part of this PR) Bundle validation is now done by the `Resolver`. The reasoning behind this is that it reduces the complexity of the `reconcile` method by making the assumption that the configured `Resolver` will _always_ return a valid bundle. Validation unit tests that existed in the `internal/controller` package have been moved and updated in the `internal/resolve` package
- Unpacking was unchanged
- Installation of content is now abstracted into a more generic `Applier` interface. This makes it easier to, in the future, swap out the underlying installation method by implementing a new `Applier` interface and swapping it out for the existing one. The existing Helm-based installation logic was moved to an implementation of this interface.
- Watching of managed resources was moved to a new `Watcher` interface that was implemented as part of the provided service account work. This was going to be a change that took place in the near future. From a functionality perspective, nothing should have changed.
- Unit tests added for the `ClusterExtensionReconciler.reconcile()` method

## Reviewer Checklist

- [ ] API Go Documentation


joelanford:
Nice! I like this idea. I'll take a closer look at this tomorrow, but one thing I would suggest off the bat is talking to @thetechnick about the `Applier` interface. I was chatting with him last week about experimenting with plugging `package-operator`'s apply logic into operator-controller. I figure having a second implementation candidate would help validate our ideas behind the shape of that interface.

thetechnick:
I do have a very rough first PoC here:
https://github.com/thetechnick/operator-controller/pull/1
It directly uses Package Operator APIs to offload bundle reconciliation, so operator-controller just has to create a `ClusterPackage`, read status from it and lean back, while getting features similar to the OLMv0 Subscription config for free.

everettraven:
@joelanford Sounds good. Are you thinking to utilize the `ClusterPackage` similarly to what @thetechnick has done (via the PoC) or more along the lines of importing the apply logic and using it as a library?

I'd be a bit hesitant to move back to an approach where we are creating an underlying resource like the `ClusterPackage` since we intentionally decided to deviate from that approach relatively recently (~3ish months ago).

internal/applier/helm.go
```diff
@@ -0,0 +1,166 @@
...
+	// check is unsuccessful
+	Upgrade(context.Context, *release.Release) error
+}
+
+type Helm struct {
+	ActionClientGetter helmclient.ActionClientGetter
+	Preflights         []Preflight
+}
+
+func (h *Helm) Apply(ctx context.Context, contentFS fs.FS, ext *ocv1alpha1.ClusterExtension, labels map[string]string) ([]client.Object, string, error) {
```
everettraven, line 54: Ideally we make this function more easily unit testable and add explicit unit tests for this function and the individual components but I would put that out of scope of the intention of this PR.

internal/catalogmetadata/filter/successors.go
```diff
@@ -20,17 +20,17 @@ func SuccessorsOf(installedBundle *ocv1alpha1.BundleMetadata, channels ...declcf
 
 	installedBundleVersion, err := mmsemver.NewVersion(installedBundle.Version)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("parsing installed bundle version %q: %w | installedBundle %v", installedBundle.Version, err, installedBundle)
```
tmshort, line 23:
```suggestion
		return nil, fmt.Errorf("parsing installed bundle %q version %q: %w | installedBundle %v", installedBundle.Name, installedBundle.Version, err)
```

everettraven: Ah shoot, some of that is still debugging changes I didn't rollback - thanks for catching that!

everettraven: should be updated in https://github.com/operator-framework/operator-controller/pull/1068/commits/8c127588ceff671306f53a290b9f4acec2247f75

everettraven:
I will squash commits after reviews are fully addressed

tmshort:
Looks reasonable to me, but needs a rebase.

tmshort (approved):
/lgtm

Issue #920, opened 2024-06-11
[epic] v1.0.0 Performance and Scale


joelanford:
# Epic Goal

- Measure and implement any necessary improvements to ensure OLM v1.0.0 meets or exceeds OCP guidelines around performance and scalability.

# Why is this important?

- OLM v1.0.0 will be a payload component that always runs in OCP clusters. In order to reduce SD and customer costs, we need to minimize this overhead.
- OLM v1.0.0 is intended to be used on a wide variety of clusters, ranging from single node clusters with just a few namespaces to clusters 2-3 orders of magnitude larger. We need make sure that it runs just as well on a small cluster as it does a large cluster.
- In order to reduce user frustration, we need to provide a responsive user experience. Reconciliation needs to be fast and non-blocking to ensure users receive the experience they have come to expect from OCP. To the extent possible, long-running tasks (e.g. catalog fetching/caching and image pulling) should be performed asynchronously.

# Scenarios

1. Collect pprof profiles for CPU and memory when running standard user flows around installing, upgrading, and removing operators from public catalogs (e.g. operatorhub)
2. Find the most resource intensive code paths. Provide documentation and recommendations related to making improvements in those areas.
3. Coordinate with OLM maintainers to make improvements in areas deemed to provide the most significant performance and scale gain.
4. Implement automated performance and scale regression tests in the existing upstream CI test suite.

Examples of known areas for improvement include:
- When reconciling a ClusterExtension to resolve a bundle from the criteria provided by a user, the reconciler should return a desired bundle within 100ms and allocate no more memory than the size of the catalog metadata for the named spec.packageName.
- When the ClusterExtension reconciler does not have the contents of a resolved image bundle available, it does not block waiting for the image to be pulled and processed. Rather, it starts an asychronous job, reports the pending image pull via the ClusterExtension status, and returns from reconcile.

joelanford:
I think I've found one unexpected slowdown: the bundle handler that converts a registry+v1 bundle to plain and then to helm. It takes 5s on my machine in the "Force upgfade" epic.

joelanford:
@OchiengEd just wanted to check to make sure you didn't find any critical (as in "must fix for 1.0.0") issues in your performance and scale research?

If not, can we move the remaining scope of this epic to v1.x?

everettraven:
Item to include in performance and scale (although maybe more of a release blocker based on other discussions): 
[...]

PR #1071, opened 2024-07-18
:sparkles: feat: implement initial benchmark tests

OchiengEd:
# Description

This PR is for implementing benchmarking tests which will be used in CI to evaluate change in performance / compute resource utilization as new features are implemented.

## Reviewer Checklist

- [ ] API Go Documentation

Relates to #920


internal/contentmanager/contentmanager_test.go
```diff
@@ -108,6 +108,48 @@ func TestWatch(t *testing.T) {
...
+		ctrl, err := controller.New("test-controller", mgr, controller.Options{
+			Reconciler: reconcile.Func(func(context.Context, reconcile.Request) (reconcile.Result, error) {
+				return reconcile.Result{}, nil
+			}),
+		})
+		if err != nil {
+			b.Logf("creating controller; %v", err)
+		}
+
+		instance := New(rcm, config, mgr.GetRESTMapper())
```
joelanford, line 145: This is quite a bit of setup that will impact the benchmark results, and I would expect that quite a bit CPU time and memory allocation happens in this setup, potentially much more than in the `instance.Watch` call.

internal/contentmanager/contentmanager_test.go
```diff
@@ -108,6 +108,48 @@ func TestWatch(t *testing.T) {
...
+		err = instance.Watch(context.Background(), ctrl, ce, objs)
```
joelanford, line 146: Without a manager or controller actually running, I'm not sure `instance.Watch` does all that much.

internal/authentication/tokengetter_test.go
```diff
@@ -85,3 +85,65 @@ func TestTokenGetterGet(t *testing.T) {
 		}
 	}
 }
+
+func BenchmarkTokenGetterGet(b *testing.B) {
+	fakeClient := fake.NewSimpleClientset()
+	fakeClient.PrependReactor("create", "serviceaccounts/token",
+		func(action ctest.Action) (bool, runtime.Object, error) {
+			act, ok := action.(ctest.CreateActionImpl)
```
joelanford, line 93: It might make sense to stop the benchmark timer for the duration of this function. Otherwise, this code will be included in the benchmark results.

joelanford (changes requested):
Broadly, I would suggest that we focus benchmarks on areas of code that we think may be particularly prone to CPU time and allocation.

Do we have any indication that the contentmanager, tokengetter, or schemebuilder are particularly problematic? If not, I'm concerned that these benchmarks could become a source of noise and toil.

In particular, I think the four best places to add benchmarks are:
- The catalog cache and/or client.
- The `Resolver` interface implementation.
- The image unpacker.
- The registry+v1 to helm chart conversion.

In my observation, these seem to be the largest contributors to the duration of a `Reconcile` call, so even performance regressions in these areas likely have a bigger impact that other areas of the codebase.

internal/contentmanager/contentmanager_test.go
OchiengEd: Removing this test and focusing on areas more likely to be impactful on performance

internal/contentmanager/contentmanager_test.go
OchiengEd: Removing initial tests and focusing on areas more likely to be impactful on performance

internal/authentication/tokengetter_test.go
OchiengEd: Removing this test and implementing benchmarks for more consequential functions

.github/workflows/benchmark.yaml
```diff
@@ -0,0 +1,46 @@
...
+    - name: Compare benchmark results
+      if: steps.download.outcome == 'success'
+      run: benchstat benchmark.txt stats/benchmark.txt
+
+    - name: Upload artifacts
+      uses: actions/upload-artifact@v4
+      with:
+        name: benchmark-results
+        path: benchmark.txt
+        overwrite: true
```
OchiengEd, line 46: Still needs a step to compare the change in performance as a result of the PR .

In addition, what do we think would be a good threshold for a computing performance degradation

.github/workflows/benchmark.yaml
```diff
@@ -0,0 +1,85 @@
...
+        uses: jannekem/run-python-script-action@v1
+        with:
+          script: |
+            import pandas as pd
+
+            df = pd.read_csv('stats.csv',
+              names=['name','reference','secs/op','benchmark','_secs/op','change','P'])
+            df.change = df.change.replace("~", "0")
+            df.change = df.change.str.rstrip("%")
+            print(df.loc[df.change > 20, ['name','change','P']])
```
OchiengEd, line 85: Should we error if there is a significant drop of performance and should we print all the performance changes?

itroyano: I think we should,  and maybe there is a way to make it show on the PR itself (similarly to how code coverage does)?

internal/rukpak/source/image_registry.go
```diff
@@ -114,17 +115,16 @@ func (i *ImageRegistry) Unpack(ctx context.Context, bundle *BundleSource) (*Resu
 		hexVal := strings.TrimPrefix(digest.DigestStr(), "sha256:")
 		unpackPath := filepath.Join(i.BaseCachePath, bundle.Name, hexVal)
 		if stat, err := os.Stat(unpackPath); err == nil && stat.IsDir() {
-			l.V(1).Info("found image in filesystem cache", "digest", hexVal)
```
tmshort, line 117: Why are we removing logs?

OchiengEd: The logs would end up in the benchmarks which would eventually result in an error when benchstat is run against the benchmarks.

internal/rukpak/source/image_registry_test.go
```diff
@@ -0,0 +1,41 @@
...
+			},
+		),
+	)
+
+	ctx := log.IntoContext(context.Background(), logger)
+
+	bundleSource := &source.BundleSource{
+		Type: source.SourceTypeImage,
+		Image: &source.ImageSource{
+			Ref: "quay.io/eochieng/litmus-edge-operator-bundle@sha256:104b294fa1f4c2e45aa0eac32437a51de32dce0eff923eced44a0dddcb7f363f",
```
OchiengEd, line 33: This bundle image will also need to be changed to something that we expect to stay around longer.

internal/rukpak/source/image_registry.go
OchiengEd: @tmshort it might be possible to suppress logs if we detect benchmarks are being run. Would that be a preferred approach?

OchiengEd:
While reviewing, consider:

1. What should be threshold for an increase in computational resource utilization?
2. And should we error when above threshold is exceeded?
3. Should we add a comment with the benchmark results?
4. Informational logs in the function being benchmarked will break the benchmark format. As a workaround, should we remove the comments altogether or suppress them when benchmarks are running?

internal/rukpak/source/image_registry.go
OchiengEd: The controller-runtime zap logger will to write to `io.Discard`  which consequently, will require no changes to the Unpack() function.

OchiengEd:
During a meeting on the benchmarking of operator-controller functions, we discussed:
- [ ]  Committing a golden benchmark to the repository and comparing benchmarks when new PRs are created to the golden / reference benchmark
- [ ] Set the threshold as low as possible, preferrably zero, and alerting when there is any measurable change
- [ ] Switch the analysis of the benchmark data to use Golang tools / modules from Pandas which is written in Python
- [ ] Fail / Alert when a benchmark is removed from the golden benchmark or when a benchmark which does not exist in the Golden benchmark is added to a PR

OchiengEd:
Will be revisited at a later time.

PR #1074, opened 2024-07-19
:sparkles: wire up ServiceAccount based caching layer

### LINKED ISSUE #975: Wire up dynamic caching layer to manage installed content

### LINKED ISSUE #983: [Bug] Reconciliation of cluster extension content not working on main branch
Since we setup watches for all objects installed by a cluster extension. It is expected that the controller will reconcile cluster extension content in case any of it is modified. However this does not work as expected. 

Steps to reproduce:
1. Switch to main branch
2. `make run` 
3.  Deploy sample catalog
4. Deploy sample extension
5. Delete deployment of extension controller

Expected Result:
The controller should react and recreate the deployment

Actual Result:
The controller does not react

everettraven:
# Description
- Updates the RestConfigMapper to use a new `TokenTripper` that wraps a `http.RoundTripper` to always set the Authorization header to the referenced ServiceAccount
- Updates the E2E ServiceAccounts introduced in #1038 to use a scoped set of permissions instead of giving it cluster-admin equivalent permissions (used `bind` + `escalate` for RBAC to keep the SA permissions as succinct as possible. Open to changing this if desired)
- Updates the `ClusterExtensionReconciler` to have an exported field that can be used to configure the `contentmanager.Watcher` that should be used to establish watches on the ClusterExtension managed objects
- Updates the `contentmanager.Instance.Watch()` method to create watches with the same configurations as existed in the `ClusterExtensionReconciler.reconcile()` method previously
- Adds an E2E test to verify that when a change is made to managed objects (in the test case we delete a managed object), that the managed object triggers reconciliation and changes are reverted (in the test case we ensure the managed object is recreated)

resolves #975 
resolves #983 

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"net/http"
 	"os"
 	"path/filepath"
 	"time"
@@ -49,6 +50,7 @@ import (
 	"github.com/operator-framework/operator-controller/internal/authentication"
 	"github.com/operator-framework/operator-controller/internal/catalogmetadata/cache"
 	catalogclient "github.com/operator-framework/operator-controller/internal/catalogmetadata/client"
+	"github.com/operator-framework/operator-controller/internal/contentmanager"
 	"github.com/operator-framework/operator-controller/internal/controllers"
 	"github.com/operator-framework/operator-controller/internal/httputil"
 	"github.com/operator-framework/operator-controller/internal/labels"
@@ -179,12 +181,14 @@ func main() {
 			Name:      cExt.Spec.ServiceAccount.Name,
 			Namespace: cExt.Spec.InstallNamespace,
 		}
-		token, err := tokenGetter.Get(ctx, namespacedName)
-		if err != nil {
-			return nil, fmt.Errorf("failed to extract SA token, %w", err)
-		}
 		tempConfig := rest.AnonymousClientConfig(c)
-		tempConfig.BearerToken = token
+		tempConfig.WrapTransport = func(rt http.RoundTripper) http.RoundTripper {
+			return &authentication.TokenInjectingRoundTripper{
+				Tripper:     rt,
+				TokenGetter: tokenGetter,
+				Key:         namespacedName,
+			}
+		}
 		return tempConfig, nil
 	}
 	cfgGetter, err := helmclient.NewActionConfigGetter(mgr.GetConfig(), mgr.GetRESTMapper(),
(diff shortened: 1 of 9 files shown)


internal/authentication/tripper.go
```diff
@@ -0,0 +1,29 @@
...
+	Tripper     http.RoundTripper
+	TokenGetter *TokenGetter
+	Key         types.NamespacedName
+}
+
+func (tt *TokenTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	reqClone := utilnet.CloneRequest(req)
+	token, err := tt.TokenGetter.Get(reqClone.Context(), tt.Key)
+	if err != nil {
+		return nil, err
```
everettraven, line 23: One thing I think worth noting is the comment on the http.RoundTripper interface that calls out scenarios where RoundTrip errors:

>	// RoundTrip must always close the body, including on errors,
	// but depending on the implementation may do so in a separate
	// goroutine even after RoundTrip returns. This means that
	// callers wanting to reuse the body for subsequent requests
	// must arrange to wait for the Close call before doing so.
	
Should we still return the error here, or should we log the error and let the chain continue with an empty token value (that will likely result in installation failure)?

everettraven: Should be resolved by: https://github.com/operator-framework/operator-controller/pull/1074/files#diff-5afa0f97777f966f560e34685db2c58fe887216c09e41def816eb4d9232f45e3R25-R29

skattoju (approved):
/lgtm

cmd/manager/main.go
```diff
@@ -200,7 +204,6 @@ func main() {
 	acg, err := action.NewWrappedActionClientGetter(cfgGetter,
 		helmclient.WithFailureRollbacks(false),
 	)
-
```
tmshort, line 203: Nit: change in whitespace

internal/controllers/clusterextension_controller.go
```diff
@@ -533,13 +508,11 @@ func (r *ClusterExtensionReconciler) SetupWithManager(mgr ctrl.Manager) error {
 				},
 			})).
 		Build(r)
-
```
tmshort, line 536: Nit: whitespace-only change

tmshort (approved):
/lgtm
Just small wicked minor nits.

PR #1079, opened 2024-07-24
✨ Install default-catalogs.yaml from Catalogd release

trgeiger:
# Description
Addresses #1026 by adding OperatorHub to the default installation script.

Installs default-catalogs.yaml from Catalogd release.

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/scripts/install.tpl.sh
+++ b/scripts/install.tpl.sh
@@ -11,6 +11,7 @@ fi
 
 catalogd_version=$CATALOGD_VERSION
 cert_mgr_version=$CERT_MGR_VERSION
+install_default_catalogs=$INSTALL_DEFAULT_CATALOGS
 
 if [[ -z "$catalogd_version" || -z "$cert_mgr_version" ]]; then
     err="Error: Missing component version(s) for: "
@@ -37,6 +38,11 @@ kubectl_wait "cert-manager" "deployment/cert-manager-webhook" "60s"
 
 kubectl apply -f "https://github.com/operator-framework/catalogd/releases/download/${catalogd_version}/catalogd.yaml"
 kubectl_wait "olmv1-system" "deployment/catalogd-controller-manager" "60s"
+
+if [[ "${install_default_catalogs,,}" != "false" ]]; then
+    kubectl apply -f "https://github.com/operator-framework/catalogd/releases/download/${catalogd_version}/default-catalogs.yaml"
+    kubectl wait --for=condition=Unpacked "clustercatalog/operatorhubio" --timeout="60s"
+fi
 
 kubectl apply -f "${operator_controller_manifest}"
 kubectl_wait "olmv1-system" "deployment/operator-controller-controller-manager" "60s"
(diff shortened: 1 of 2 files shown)


tmshort:
@trgeiger - requires rebase, and one of the e2e's is failing

trgeiger:
@tmshort e2e passes now, it was mad about the presence of operatorhub.io so I just edited the setup script to remove it for the test. Codecov is mad but everything else is passing

test/extension-developer-e2e/setup.sh
```diff
@@ -74,6 +74,15 @@ reg_bundle_img="${LOCAL_REGISTRY_HOST}/bundles/registry-v1/registry-bundle:v0.0.
...
 
+
+########################################
+# Remove the default OperatorHub.io 
+# ClusterCatalog
+########################################
+
+# We only want to worry about the test catalogs
+kubectl delete clustercatalog operatorhubio
+
```
tmshort, line 85: I would prefer a way for this not to have to be here. i.e. have a way to parameterize `install.tpl.sh` about whether to actually install the default catalog, rather than always installing it, then removing it...
Would it make more sense to do:

```
kubectl delete -f "https://github.com/operator-framework/catalogd/releases/download/${catalogd_version}/default-catalogs.yaml"
```
?

trgeiger: Oh yeah I could do something like pass in another variable to the install.tpl.sh that determines if the install script installs default-catalogs.yaml

trgeiger: Do you think it would make more sense to have a variable like NO_DEFAULT_CATALOGS where if that's empty or not set then we install default-catalogs.yaml, or a positive variable like INSTALL_DEFAULT_CATALOGS=true that is set to true by default and then is set to false for the ext-dev-e2e test?

tmshort: Thanks. I'd prefer the later "INSTALL_DEFAULT_CATALOG=true" or "False"... I dislike double-negatives. :)

scripts/install.tpl.sh
```diff
@@ -38,5 +39,10 @@ kubectl_wait "cert-manager" "deployment/cert-manager-webhook" "60s"
 kubectl apply -f "https://github.com/operator-framework/catalogd/releases/download/${catalogd_version}/catalogd.yaml"
 kubectl_wait "olmv1-system" "deployment/catalogd-controller-manager" "60s"
 
+if [[ "${install_default_catalogs,,}" == "true" ]]
```
tmshort, line 42:
```suggestion
if [[ "${install_default_catalogs,,}" == "true" ]]; then
```

trgeiger: thanks, I threw this up right before lunch and didn't run it locally yet to catch that typo :)

scripts/install.tpl.sh
```diff
@@ -38,5 +39,10 @@ kubectl_wait "cert-manager" "deployment/cert-manager-webhook" "60s"
 kubectl apply -f "https://github.com/operator-framework/catalogd/releases/download/${catalogd_version}/catalogd.yaml"
 kubectl_wait "olmv1-system" "deployment/catalogd-controller-manager" "60s"
 
+if [[ "${install_default_catalogs,,}" != "false" ]]; then
```
trgeiger, line 42: I thought this made more sense than checking for true since we want the default to be installing the catalogs.

tmshort (approved):
/lgtm

Issue #1022, opened 2024-07-09
bug: Misleading error message when no successors available


joelanford:
I have a scenario setup with 6.0.1 installed, where I updated the spec.version to 6.0.0 with upgrade policy enforcement enabled.

I got the following error message:
```
    message: 'error upgrading from currently installed version "6.0.1": no package
      "cockroachdb" matching version "6.0.0" found'
```

This makes it sound like:
  - there just isn't a 6.0.0 version in the catalog
  - if there was a 6.0.0 version (there is), the upgrade would proceed (it doesn't, because 6.0.0 is not a successor)

A good initial improvement would be something like:
```
    message: 'error upgrading from currently installed version "6.0.1": desired package "cockroachdb" with version range "6.0.0" does not match any successor of "6.0.1"
```

It would be sugar on top if we listed the best successor in both the z stream and y stream. For instance:
```
    message: 'error upgrading from currently installed version "6.0.1": desired package "cockroachdb" with version range "6.0.0" does not match any successor of "6.0.1". Highest version successors of "6.0.1" are "6.3.11" (y-stream) and "6.0.10" (z-stream)
```

LalatenduMohanty:
@joelanford mentioned that he has already fixed it, so assigning it to him to cross check.

joelanford:
I cross checked. I don't think I did fix this actually. My mistake!

PR #1083, opened 2024-07-26
🐛 WIP: Updating error message when no successors available 

dhoover103:
# Description

This is an update to error messages received when trying to upgrade an operator, per #1022 

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,7 @@ require (
 	carvel.dev/kapp v0.63.3
 	github.com/Masterminds/semver/v3 v3.3.0
 	github.com/blang/semver/v4 v4.0.0
-	github.com/containerd/containerd v1.7.21
+	github.com/containerd/containerd v1.7.22
 	github.com/fsnotify/fsnotify v1.7.0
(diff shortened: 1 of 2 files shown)


internal/resolve/catalog.go
```diff
@@ -101,6 +102,7 @@ func (r *CatalogResolver) Resolve(ctx context.Context, ext *ocv1alpha1.ClusterEx
 				return nil
 			}
 			if compare.ByVersion(*resolvedBundle, thisBundle) < 0 {
+				oldBundle = &thisBundle
```
joelanford, line 105: I'm not sure this approach covers all of the possibilities. I know the scenario talks about trying to "go backwards", but I think there's a similar scenario where the `spec.version` is ahead of the installed bundle, but contains no successors.

In general, I think we need to find a way to distinguish between the following:
- No bundles in this range
- The bundles in this range are not successors of the currently installed bundle

Distinguishing these may require more refactoring. Perhaps applying the various predicates separately would help us tell which predicates cause the resulting set of bundles to reduce to an empty list.

PR #1084, opened 2024-07-26
🌱 Bundle + CRD Upgrade Safety Tests

dtfranz:
Adds unit test coverage to the `bundleutil` and `crdupgradesafety` packages. Also prunes some unused code.

## Reviewer Checklist

- [ ] API Go Documentation


tmshort:
Any missed error conditions?

dtfranz:
> Any missed error conditions?

I haven't found a good way to cover the errors from `ToUnstructured` and `FromUnstructured` yet.

internal/bundleutil/bundle_test.go
```diff
@@ -0,0 +1,73 @@
...
+				Value: json.RawMessage(`{"version": "1.0.0"}`),
+			},
+		},
+		{
+			name: "invalid version",
+			pkgProperty: &property.Property{
+				Type:  property.TypePackage,
+				Value: json.RawMessage(`{"version": "abcd"}`),
+			},
+			wantErr: errors.New("No Major.Minor.Patch elements found"),
```
everettraven, line 35: should we test for explicit error strings / matches? I wonder if just checking that an error is returned or not is sufficient for our needs. It also makes these unit tests more resilient to changes of the error strings returned.

This is presuming we don't really care that much about the exact error string that is returned and more so just that this case results in an error.

internal/rukpak/preflights/crdupgradesafety/crdupgradesafety_test.go
```diff
@@ -0,0 +1,203 @@
...
+				Name:     "test-release",
+				Manifest: "abcd",
+			},
+			wantErrMsgs: []string{"json: cannot unmarshal string into Go value of type unstructured.detector"},
+		},
+		{
+			name: "release with no CRD objects",
+			release: &release.Release{
+				Name:     "test-release",
+				Manifest: getManifestString(t, "../../../../testdata/manifests/no-crds.json"),
```
everettraven, line 106: Nit: I'm personally not a huge fan of unit tests having dependencies on a file existing. I find it harder to grok what the unit test is trying to achieve when I have to jump back and forth between the unit test code and a file with $content

Instead of having JSON files (which are harder to read than YAML, IMO), what do you think about having the Go types for the manifests defined in the test itself and then at test runtime building the release with that data?

everettraven:
>I haven't found a good way to cover the errors from ToUnstructured and FromUnstructured yet.

@dtfranz You might be able to create an interface that the CRD Upgrade Safety preflight type takes in as an exported, configurable field and then mock it for testing

internal/bundleutil/bundle_test.go
dtfranz: I agree with you on this test; the explicit message checking is probably unnecessary. For the `crdupgradesafety` tests, are you OK with the semi-explicit checking there? To me it seemed useful to do so in that context.

internal/rukpak/preflights/crdupgradesafety/crdupgradesafety_test.go
dtfranz: I think we may differ in opinion here :sweat_smile: I personally don't like having to scroll through hundreds of lines of code for static structs and strings in test files; putting them in separate json files makes the test code cleaner IMHO. 

This doesn't help with your nit at all but I can at least make a `const` for the folder location so we don't see the `.../../../../` all over the place...

dtfranz:
> > I haven't found a good way to cover the errors from ToUnstructured and FromUnstructured yet.
> 
> @dtfranz You might be able to create an interface that the CRD Upgrade Safety preflight type takes in as an exported, configurable field and then mock it for testing

Was thinking about this but do you think it's worthwhile? It doesn't really buy us that much and would require us to hack up the preflight type just to cover two lines. 

The other thing was I could never get those errors to trigger with the json I provided. If the json is invalid, we hit an earlier error, and if the CRD schema inside is invalid it doesn't care at all (it ignores unknown fields by default).

internal/rukpak/preflights/crdupgradesafety/crdupgradesafety_test.go
everettraven: @dtfranz No need to do anything for the nit unless you'd like. Just something I felt worth calling out in case others felt more strongly one way or the other.

everettraven:
>Was thinking about this but do you think it's worthwhile?

@dtfranz probably not. Covering those 2 lines isn't going to block me from approving

internal/bundleutil/bundle_test.go
everettraven: @dtfranz re: crdupgradesafety tests - I'm OK with the semi explicit there if that is the only way to validate and ensure that the appropriate validation(s) are failing. In an ideal world, we would be able to have a separate test for each validation but if that isn't possible what is in this PR seems fine.

internal/rukpak/util/fs.go
tmshort, line 1: So these files are no longer necessary?

internal/rukpak/preflights/crdupgradesafety/crdupgradesafety_test.go
tmshort: I'd prefer YAML files as well, even though the JSONs are kinda-sorta formatted like YAML...

tmshort (approved):
/lgtm
@everettraven you good with this, too?

everettraven (approved):

PR #1100, opened 2024-08-07
🌱 Internal Rukpak Package Tests

### LINKED ISSUE #1037: Add unit test coverage for migrated rukpak packages
Seems like a unit test on this with Bundle in/Chart out would be good. Just in case Helm bumps, etc, shift something on us

_Originally posted by @bentito in https://github.com/operator-framework/operator-controller/pull/1032#discussion_r1673683851_

dtfranz:
Adds more tests to the packages migrated from rukpak.

Closes #1037 

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/rukpak/source/image_registry_test.go
+++ b/internal/rukpak/source/image_registry_test.go
@@ -1,0 +1,360 @@
+package source_test
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"log"
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/google/go-containerregistry/pkg/crane"
+	"github.com/google/go-containerregistry/pkg/name"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/operator-framework/operator-controller/internal/rukpak/source"
+)
+
+const (
+	testFilePathBase string = ".image-registry-test"
+	bogusDigestHex   string = "0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef"
+	testImageTag     string = "test-tag"
+	testImageName    string = "test-image"
+	badImageName     string = "bad-image"
+	testFileName     string = "test-file"
+	testFileContents string = "test-content"
+)
+
+func newReference(host, repo, ref string) (name.Reference, error) {
+	tag, err := name.NewTag(fmt.Sprintf("%s/%s:%s", host, repo, ref), name.WeakValidation)
+	if err == nil {
+		return tag, nil
+	}
+	return name.NewDigest(fmt.Sprintf("%s/%s@%s", host, repo, ref), name.WeakValidation)
+}
+
+func imageToRawManifest(img v1.Image) ([]byte, error) {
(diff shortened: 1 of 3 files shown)


internal/rukpak/util/hash_test.go
```diff
@@ -0,0 +1,76 @@
...
+				num: 22,
+				obj: struct{ foo string }{foo: "bar"},
+				arr: []int{0, 1},
+				b:   true,
+				n:   nil,
+			},
+			expectedHash: "16jfjhihxbzhfhs1k5mimq740kvioi98pfbea9q6qtf9",
+		},
+		{
+			// An empty object generates the same hash as the populated object above... that feels wrong?
```
dtfranz, line 40: Commenting here to call attention to this.

internal/rukpak/util/util_test.go
```diff
@@ -0,0 +1,148 @@
...
+spec:
+  clusterIP: None
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: service-b
+  namespace: ns-b
+spec:
+  clusterIP: 0.0.0.0`,
```
LalatenduMohanty, line 93: wrong formatting?

dtfranz: I'm defining a file inline here literally with the tilde character instead of creating a separate file for it. It'll look odd compared to the formatting of the rest of the code here because it doesn't have leading whitespace unless required for yaml format - assuming that's what you're referring to, or did you see an issue in the yaml content?

LalatenduMohanty: Got it. All good. Thanks

internal/rukpak/util/hash_test.go
dtfranz: Figured this out thanks to @joelanford , the object passed into `DeepHashObject` needs to have exported fields or else the JSON encoder will treat the object the same as an entirely empty object. I've updated the tests to show this behavior.

internal/rukpak/util/hash_test.go
```diff
@@ -0,0 +1,117 @@
...
+				Str string
+				Num int
+				Obj interface{}
+				Arr []int
+				B   bool
+				N   interface{}
+			}{
+				Str: "foobar",
+				Num: 22,
+				Obj: struct{ foo string }{foo: "bar"},
```
joelanford, line 32: Nit: might want to also export `Foo` here. As is, `Obj` will be encoded as `{}` since foo is still unexported.

dtfranz: Oops, missed that one. Thanks!

internal/rukpak/source/image_registry_test.go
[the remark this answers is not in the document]
everettraven: Just to clarify, this is not something that I think should block this PR from going in. Just something that stood out to me during review and having the contextual knowledge that we implemented unit tests for similar functionality in catalogd a while back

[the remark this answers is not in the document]
dtfranz: Thanks for showing me that! I missed that in there; I'll follow that example and see how much simplification can be done in here.

dtfranz:
> Overall looks good to me. Left a general comment for comparison of the image registry tests implemented here and in catalogd, but it doesn't need to be addressed for this PR to merge IMO.

It currently looks to me like I'll need to keep the low-level http response stuff in there for some of the error returns; I'll revisit this at another time and merge for now. Thank you!

PR #1104, opened 2024-08-08
:bug: CRD Upgrade Safety pre-flight fixes

everettraven:
# Description
- Updates CRD Upgrade Safety pre-flight check to pass on old CRD not found
- Updates CRD Upgrade Safety pre-flight check to run on both install and upgrade
- Updates unit tests for CRD Upgrade Safety pre-flight check
- Updates E2E tests to add a test that ensures fixing the error from a failed initial installation results in a successful install (or "upgrade" in helm's semantics)

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/rukpak/preflights/crdupgradesafety/crdupgradesafety.go
+++ b/internal/rukpak/preflights/crdupgradesafety/crdupgradesafety.go
@@ -10,6 +10,7 @@ import (
 	"helm.sh/helm/v3/pkg/release"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	apiextensionsv1client "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 
@@ -64,11 +65,15 @@ func NewPreflight(crdCli apiextensionsv1client.CustomResourceDefinitionInterface
 	return p
 }
 
-func (p *Preflight) Install(_ context.Context, _ *release.Release) error {
-	return nil
+func (p *Preflight) Install(ctx context.Context, rel *release.Release) error {
+	return p.runPreflight(ctx, rel)
 }
 
 func (p *Preflight) Upgrade(ctx context.Context, rel *release.Release) error {
+	return p.runPreflight(ctx, rel)
+}
+
+func (p *Preflight) runPreflight(ctx context.Context, rel *release.Release) error {
 	if rel == nil {
 		return nil
 	}
@@ -96,6 +101,11 @@ func (p *Preflight) Upgrade(ctx context.Context, rel *release.Release) error {
 
 		oldCrd, err := p.crdClient.Get(ctx, newCrd.Name, metav1.GetOptions{})
 		if err != nil {
+			// if there is no existing CRD, there is nothing to break
+			// so it is immediately successful.
+			if apierrors.IsNotFound(err) {
+				return nil
+			}
 			return fmt.Errorf("getting existing resource for CRD %q: %w", newCrd.Name, err)
 		}
 
--- a/internal/rukpak/preflights/crdupgradesafety/crdupgradesafety_test.go
+++ b/internal/rukpak/preflights/crdupgradesafety/crdupgradesafety_test.go
@@ -12,8 +12,10 @@ import (
 	"helm.sh/helm/v3/pkg/release"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	apiextensionsv1client "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 
 	"github.com/operator-framework/operator-controller/internal/rukpak/preflights/crdupgradesafety"
 	"github.com/operator-framework/operator-controller/internal/rukpak/util"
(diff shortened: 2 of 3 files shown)


internal/rukpak/preflights/crdupgradesafety/crdupgradesafety.go
```diff
@@ -96,6 +101,11 @@ func (p *Preflight) Upgrade(ctx context.Context, rel *release.Release) error {
 
 		oldCrd, err := p.crdClient.Get(ctx, newCrd.Name, metav1.GetOptions{})
 		if err != nil {
+			// if there is no existing CRD, there is nothing to break
+			// so it is immediately successful.
+			if apierrors.IsNotFound(err) {
+				return nil
```
joelanford, line 107: Should this be `continue`?

internal/rukpak/preflights/crdupgradesafety/crdupgradesafety_test.go
```diff
@@ -71,8 +73,140 @@ func getManifestString(t *testing.T, crdFile string) string {
 // TestInstall exists only for completeness as Install() is currently a no-op. It can be used as
 // a template for real tests in the future if the func is implemented.
 func TestInstall(t *testing.T) {
-	preflight := newMockPreflight(nil, nil, nil)
-	require.Nil(t, preflight.Install(context.Background(), nil))
+	tests := []struct {
+		name          string
+		oldCrdPath    string
```
joelanford, line 78: Based on my comment above about `return nil` vs `continue`, it seems like we may want to test scenarios where there are multiple CRDs at play.

test/e2e/cluster_extension_install_test.go
```diff
@@ -593,6 +598,90 @@ func TestClusterExtensionInstallReResolvesWhenManagedContentChanged(t *testing.T
...
+		}
+		assert.Equal(ct, metav1.ConditionFalse, cond.Status)
+		assert.Equal(ct, ocv1alpha1.ReasonInstallationFailed, cond.Reason)
+		assert.Contains(ct, cond.Message, "forbidden")
+	}, pollDuration, pollInterval)
+
+	t.Log("By fixing the ServiceAccount permissions")
+	require.NoError(t, createClusterRoleAndBindingForSA(context.Background(), name, sa))
+
+	t.Log("By eventually installing the package successfully")
```
joelanford, line 671: I assume we can't actually watch for the fixed service account, so this is predicated on exponential backoff retries?  If so, we should:

1. Be very careful to avoid touching the ClusterExtension after creating the necessary permissions (maybe we add a comment here about that)
2. Add a comment in the reconciler at the `return` site of the permission error to make sure that we don't swallow it and return a `nil` error.

test/e2e/cluster_extension_install_test.go
```diff
@@ -593,6 +598,90 @@ func TestClusterExtensionInstallReResolvesWhenManagedContentChanged(t *testing.T
 	}, pollDuration, pollInterval)
 }
 
+func TestClusterExtensionRecoversFromInitialInstallFailedWhenFailureFixed(t *testing.T) {
```
joelanford, line 601: Is this scenario testable in our reconcile unit tests yet? I'd love to avoid adding more e2es if possible.

internal/rukpak/preflights/crdupgradesafety/crdupgradesafety.go
everettraven: Ah, you are right - good catch. For some reason I missed that this was in a for loop 🤦

test/e2e/cluster_extension_install_test.go
everettraven: I'm not sure this test in particular is well suited to unit tests. In our unit tests we aren't necessarily testing the whole system. I imagine that we would still want an E2E test like this that flexes the entire system in this scenario.

I do think that separately we should go back through all the existing e2e tests and see if there are any that are focused in on a particular thing that could now be unit tested

test/e2e/cluster_extension_install_test.go
everettraven: Correct. I am happy to add comments in places we feel necessary, but the comment about the permissions error in the reconciler is a bit iffy for me. The permissions error is being returned by the Helm Applier implementation, which gets the error from the Helm client. Maybe adding a comment saying "we should never eat errors returned from X"?

internal/rukpak/preflights/crdupgradesafety/crdupgradesafety.go
everettraven: Switched to continue in https://github.com/operator-framework/operator-controller/pull/1104/commits/233d5f641e1e8056d052d888a195ed7ce5104d85

internal/rukpak/preflights/crdupgradesafety/crdupgradesafety_test.go
everettraven: In your opinion, is this a blocker for this PR? I do agree we should do this, but wonder if it would be better suited to have as a follow-up. I don't mind doing it, but I'm not sure holding this bug fix on a, not insignificant, unit test refactor is worth it.

joelanford: Not a blocker!

test/e2e/cluster_extension_install_test.go
joelanford: > Maybe adding a comment saying "we should never eat errors returned from X"?

Yeah, that's what I had in mind generally. I think we should explain _why_ we are not going to eat those errors. That way our future selves are reminded about this specific thing if we ever decide we _do_ want to eat that error for some reason. In that kind of situation, it would help us remember that we would need to check the error to decide whether to eat it or not.

test/e2e/cluster_extension_install_test.go
joelanford: The reason I ask is that it seems like there is a set of preconditions that we could setup and an expected output. If we have scenarios that can be tested in a single execution of the `Reconcile` function, I think we should be trying to put those in unit tests.

test/e2e/cluster_extension_install_test.go
everettraven: Comments added in appropriate places in https://github.com/operator-framework/operator-controller/pull/1104/commits/a1d5ead220d75fe44fb5f1e17f7a6c62bbf16351

test/e2e/cluster_extension_install_test.go
joelanford: @everettraven and I chatted, and landed on making a separate epic that tracks improvements to make our code more testable at various layers.

We can merge with this e2e, and make testing/code structure improvements a separate task.

internal/applier/helm.go
```diff
@@ -90,6 +90,11 @@ func (h *Helm) Apply(ctx context.Context, contentFS fs.FS, ext *ocv1alpha1.Clust
 		}
 	}
 
+	// NOTE: We need to be cautious of eating errors here.
+	// We should always return any error that occurs during an
+	// attempt to install/upgrade/reconcile. Only when there is
+	// a verifiable reason to eat the error (i.e it is recoverable)
+	// should an exception be made.
```
joelanford, line 97: I think we should call out here the specific error that cannot be eaten. Something like:

```
The following kinds of errors should be returned up the stack to ensure exponential backoff can occur:
- Permission errors (it is not possible to watch changes to permissions. The only way to eventually recover from permission errors is to keep retrying).
```

joelanford: Also, I think it is _probably_ safe to delete this instance of the comment, and leave just the one in `clusterextension_controller.go`.

My reasoning there is that I think it is unlikely that we would swallow errors here if we didn't have a way to handle them well. In the reconcile code, I could easily see us swallowing errors after updating our status object with their contents.

everettraven: Updated the comments in https://github.com/operator-framework/operator-controller/pull/1104/commits/97ce5d773526bb19691a8bf4ae1330b72e039350

joelanford (approved):

PR #1105, opened 2024-08-09
🌱 enable OwnerReferencesPermissionEnforcement in project kind config

joelanford:
# Description

Kubernetes has many admission policy plugins that are disabled by default when using vanilla Kubernetes. Vendors of Kubernetes tend to enable the `OwnerReferencesPermissionEnforcement` plugin. One prominent example of a vendor enabling this plugin is OpenShift.
In order to ensure OLM works seamlessly on as many Kubernetes distributions as possible, we should enable this plugin in our kind config.

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/config/samples/olm_v1alpha1_clusterextension.yaml
+++ b/config/samples/olm_v1alpha1_clusterextension.yaml
@@ -28,7 +28,12 @@ kind: ClusterRole
 metadata:
   name: argocd-installer-clusterrole
 rules:
-# Manage ArgoCD CRDs 
+# Allow ClusterExtension to set blockOwnerDeletion ownerReferences
+- apiGroups: [olm.operatorframework.io]
+  resources: [clusterextensions/finalizers]
+  verbs: [update]
+  resourceNames: [argocd]
+# Manage ArgoCD CRDs
 - apiGroups: [apiextensions.k8s.io]
   resources: [customresourcedefinitions]
   verbs: [create]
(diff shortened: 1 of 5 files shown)


config/samples/olm_v1alpha1_clusterextension.yaml
```diff
@@ -28,7 +28,11 @@ kind: ClusterRole
 metadata:
   name: argocd-installer-clusterrole
 rules:
-# Manage ArgoCD CRDs 
+# Allow ClusterExtension to set blockOwnerDeletion ownerReferences
+- apiGroups: [olm.operatorframework.io]
+  resources: [clusterextensions/finalizers]
+  verbs: [update]
```
joelanford, line 34: Technically, this is the only necessary change to RBAC.

The other RBAC changes are cleanups that I missed in #1102

everettraven: Is this a permission that should need to be set by a user of a ClusterExtension or is this a permission we should ensure is available on all ClusterExtensions by operator-controller's ServiceAccount?

everettraven: It just seems a bit of an odd UX to me to have a user need to provide permissions for the ClusterExtension controller to do things with a ClusterExtension

joelanford: The plugin says: "The user creating an object that contains an ownerReference with `blockOwnerDeletion: true`, must have permission to control the owner's deletion configuration (i.e. to set finalizers on the owner)"

The user in this case has to be the ClusterExtension service account because it is the one creating these objects that have these owner refs back to the ClusterExtension.

everettraven: Ah I see. That makes sense. I'm not familiar with the plugin and found a brief reference to it. Could you update the PR description to include this information? Is there a way we could document this behavior as well so that users know that when this plugin is enabled they _have_ to add this to the ServiceAccount being referenced in the ClusterExtension?

I'm fine with the documentation being a follow-up, just would like to make sure we have a tracker for it somewhere to ensure we don't forget.

everettraven (dismissed):

everettraven (approved):

config/samples/olm_v1alpha1_clusterextension.yaml
joelanford: Added a draft doc for this in this PR!

PR #1157, opened 2024-08-20
✨ update ClusterExtension.Spec.PakageName to be immutable

### LINKED ISSUE #1152: Update the `ClusterExtension.Spec.PackageName` field to be immutable
The `ClusterExtension.Spec.PackageName` field should be made immutable to prevent unknown behavior from occurring when changing the name of a package to be installed.

yashoza19:
# Description
Fixes #1152 

## Reviewer Checklist

- [x] API Go Documentation

### DIFF
--- a/config/base/crd/bases/olm.operatorframework.io_clusterextensions.yaml
+++ b/config/base/crd/bases/olm.operatorframework.io_clusterextensions.yaml
@@ -105,6 +105,9 @@ spec:
                 maxLength: 48
                 pattern: ^[a-z0-9]+(-[a-z0-9]+)*$
                 type: string
+                x-kubernetes-validations:
+                - message: packageName is immutable
+                  rule: self == oldSelf
               preflight:
                 description: Preflight defines the configuration of preflight checks.
                 properties:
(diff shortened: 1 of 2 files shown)


config/base/crd/bases/olm.operatorframework.io_clusterextensions.yaml
```diff
@@ -105,6 +105,9 @@ spec:
                 maxLength: 48
                 pattern: ^[a-z0-9]+(-[a-z0-9]+)*$
                 type: string
+                x-kubernetes-validations:
+                  - message: packageName is immutable
+                    rule: self == oldSelf
```
everettraven, line 110: Looks like there might be some formatting issues here causing the sanity/verify check to fail.

yashoza19: seems to have passed the sanity check, but still fails on tilt.

everettraven: I think the tilt check is currently expected to fail based on some recent changes that were made in catalogd. I don't think it is a required check.

everettraven (approved):

PR #1159, opened 2024-08-21
✨ convert field CRDUpgradeSafetyPolicy to be an enum

### LINKED ISSUE #1145: Convert boolean fields to enums
As outlined in the [Do not use Boolean Fields subsection of the OpenShift API Conventions](https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md#do-not-use-boolean-fields), boolean fields are to be replaced with an enumeration of values that describe the action instead. 

At the time of writing this issue, the only field that is boolean type is the `ClusterExtension.Spec.Preflight.CRDUpgradeSafety.Disabled` field: https://github.com/operator-framework/operator-controller/blob/cb9ea0032c95d59d8a4951593f6486ae6c03c0bb/api/v1alpha1/clusterextension_types.go#L112

This field should be updated to use an enumeration instead.

For inspiration, an implementation could look something like:
```go
type CRDUpgradeSafetyPolicy string 

const (
  CRDUpgradeSafetyPolicyEnabled CRDUpgradeSafetyPolicy = "Enabled"
  CRDUpgradeSafetyPolicyDisabled CRDUpgradeSafetyPolicy = "Disabled"
)

  //+kubebuilder:Required
  // +kubebuilder:validation:Enum:="Enabled","Disabled"
  // policy represents the state of the CRD upgrade safety preflight check. Allowed values are "Enabled", and Disabled".
  Policy CRDUpgradeSafetyPolicy `json:"policy,omitempty"`
}
```

yashoza19:
# Description

- Fixes #1145 

## Reviewer Checklist

- [x] API Go Documentation

### DIFF
--- a/api/v1alpha1/clusterextension_types.go
+++ b/api/v1alpha1/clusterextension_types.go
@@ -28,6 +28,7 @@ var (
 )
 
 type UpgradeConstraintPolicy string
+type CRDUpgradeSafetyPolicy string
 
 const (
 	// The extension will only upgrade if the new version satisfies
@@ -107,14 +108,16 @@ type ServiceAccountReference struct {
 type PreflightConfig struct {
 	//+kubebuilder:Required
 	// CRDUpgradeSafety holds necessary configuration for the CRD Upgrade Safety preflight checks.
-	CRDUpgradeSafety *CRDUpgradeSafetyPreflightConfig `json:"crdUpgradeSafety,omitempty"`
+	CRDUpgradeSafety *CRDUpgradeSafetyPreflightConfig `json:"crdUpgradeSafety"`
 }
 
 // CRDUpgradeSafetyPreflightConfig is the configuration for CRD upgrade safety preflight check.
 type CRDUpgradeSafetyPreflightConfig struct {
 	//+kubebuilder:Required
-	// Disabled represents the state of the CRD upgrade safety preflight check being disabled/enabled.
-	Disabled bool `json:"disabled,omitempty"`
+	//+kubebuilder:validation:Enum:="Enabled";"Disabled"
+	//+kubebuilder:default:=Enabled
+	// policy represents the state of the CRD upgrade safety preflight check. Allowed values are "Enabled", and Disabled".
+	Policy CRDUpgradeSafetyPolicy `json:"policy"`
 }
 
 const (
@@ -144,6 +147,9 @@ const (
 	ReasonUnpackFailed  = "UnpackFailed"
 
 	ReasonErrorGettingReleaseState = "ErrorGettingReleaseState"
+
+	CRDUpgradeSafetyPolicyEnabled  CRDUpgradeSafetyPolicy = "Enabled"
+	CRDUpgradeSafetyPolicyDisabled CRDUpgradeSafetyPolicy = "Disabled"
 )
 
 func init() {
--- a/config/base/crd/bases/olm.operatorframework.io_clusterextensions.yaml
+++ b/config/base/crd/bases/olm.operatorframework.io_clusterextensions.yaml
@@ -112,11 +112,20 @@ spec:
                     description: CRDUpgradeSafety holds necessary configuration for
                       the CRD Upgrade Safety preflight checks.
                     properties:
-                      disabled:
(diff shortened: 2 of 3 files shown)


yashoza19:
@everettraven Do you know why this is failing the apidiff check? Is there something I'm missing?

everettraven:
>@everettraven Do you know why this is failing the apidiff check? Is there something I'm missing?

@yashoza19 this is a breaking change to the API surface. I would expect that check to fail in this case. It is not a required check, but one that is in place to grab our attention when API changes are being made. Once we cut our stable v1.0.0 release, we will likely convert it to a required check

api/v1alpha1/clusterextension_types.go
```diff
@@ -113,8 +114,9 @@ type PreflightConfig struct {
 // CRDUpgradeSafetyPreflightConfig is the configuration for CRD upgrade safety preflight check.
 type CRDUpgradeSafetyPreflightConfig struct {
 	//+kubebuilder:Required
-	// Disabled represents the state of the CRD upgrade safety preflight check being disabled/enabled.
-	Disabled bool `json:"disabled,omitempty"`
+	// +kubebuilder:validation:Enum:="Enabled,Disabled"
```
everettraven, line 117: This would make the only allowed value the exact string `Enabled,Disabled`. I think we are looking for:
```suggestion
	// +kubebuilder:validation:Enum:="Enabled","Disabled"
```

api/v1alpha1/clusterextension_types.go
```diff
@@ -113,8 +114,9 @@ type PreflightConfig struct {
...
+	// policy represents the state of the CRD upgrade safety preflight check. Allowed values are "Enabled", and Disabled".
+	Policy CRDUpgradeSafetyPolicy `json:"policy,omitempty"`
```
everettraven, line 119: This is a required field so we should be good to remove the `omitempty` here.

api/v1alpha1/clusterextension_types.go
yashoza19: Good catch. I've updated this in the new commit. Thanks

config/base/crd/bases/olm.operatorframework.io_clusterextensions.yaml
```diff
@@ -112,10 +112,13 @@ spec:
...
-                      disabled:
-                        description: Disabled represents the state of the CRD upgrade
-                          safety preflight check being disabled/enabled.
-                        type: boolean
+                      policy:
+                        description: policy represents the state of the CRD upgrade
+                          safety preflight check. Allowed values are "Enabled", and
+                          Disabled".
+                        enum:
+                        - Enabled,Disabled
```
everettraven, line 120: Need to regenerate the CRD to get the distinct allowed values here

api/v1alpha1/clusterextension_types.go
```diff
@@ -107,14 +108,15 @@ type ServiceAccountReference struct {
...
 }
 
 // CRDUpgradeSafetyPreflightConfig is the configuration for CRD upgrade safety preflight check.
 type CRDUpgradeSafetyPreflightConfig struct {
 	//+kubebuilder:Required
-	// Disabled represents the state of the CRD upgrade safety preflight check being disabled/enabled.
-	Disabled bool `json:"disabled,omitempty"`
+	// +kubebuilder:validation:Enum:="Enabled","Disabled"
+	// policy represents the state of the CRD upgrade safety preflight check. Allowed values are "Enabled", and Disabled".
+	Policy CRDUpgradeSafetyPolicy `json:"policy"`
```
everettraven, line 119: Since we are using an Enum now, we should probably set the default value as `Enabled`.

We also need to update the code for the preflight checks here:
https://github.com/operator-framework/operator-controller/blob/1fb6bd28f917b881fcdd8fdc6ef512b157eebd94/internal/applier/helm.go#L72-L78

yashoza19: I've already updated the code for the preflight check skipping logic in this commit. I'll set the default value as Enabled now.

everettraven (approved):
Changes look good to me! Thanks @yashoza19 !

PR #1164, opened 2024-08-22
:sparkles: Add in catalog label selection support

tmshort:
Fix #1112

Adds some tests to validate the label selector.
Modified some tests to add labels and selectors. They are not necessary now, but will be once 1110 is resolved; as that issue adds the ambiguity that this PR helps resolve.

# Description

## Reviewer Checklist

- [ ] API Go Documentation
- [ ] Tests: Unit Tests (and E2E Tests, if appropriate)
- [ ] Comprehensive Commit Messages
- [ ] Links to related GitHub Issue(s)

PR #1165, opened 2024-08-22
⚠: Ambiguous Resolution

### LINKED ISSUE #1110: Error on ambiguous resolution + test

dtfranz:
Changes the behavior of catalog resolution by returning an error when multiple packages are available from multiple catalogs. 

Users will need to use forthcoming label selectors to narrow down catalogs when this occurs.

Closes #1110

# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/resolve/catalog.go
+++ b/internal/resolve/catalog.go
@@ -51,8 +51,9 @@ func (r *CatalogResolver) Resolve(ctx context.Context, ext *ocv1alpha1.ClusterEx
 	}
 
 	var (
-		resolvedBundle      *declcfg.Bundle
-		resolvedDeprecation *declcfg.Deprecation
+		resolvedBundles  []*declcfg.Bundle
+		matchedCatalogs  []string
+		priorDeprecation *declcfg.Deprecation
 	)
 
 	listOptions := []client.ListOption{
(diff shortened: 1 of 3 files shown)


internal/resolve/catalog.go
```diff
@@ -95,31 +96,23 @@ func (r *CatalogResolver) Resolve(ctx context.Context, ext *ocv1alpha1.ClusterEx
...
-	if resolvedBundle == nil {
+	if len(resolvedBundles) != 1 {
 		errPrefix := ""
 		if installedBundle != nil {
 			errPrefix = fmt.Sprintf("error upgrading from currently installed version %q: ", installedBundle.Version)
 		}
 		switch {
+		case len(resolvedBundles) > 1:
+			slices.Sort(matchedCatalogs) // sort for consistent error message
+			return nil, nil, nil, fmt.Errorf("%smatching packages found in multiple catalogs: %v; please specify a catalog", errPrefix, matchedCatalogs)
```
dtfranz, line 115: This error message needs work I think on second glance. Open to suggestions!

internal/resolve/catalog.go
```diff
@@ -95,31 +96,23 @@ func (r *CatalogResolver) Resolve(ctx context.Context, ext *ocv1alpha1.ClusterEx
...
-				return nil
-			}
-			if compare.ByVersion(*resolvedBundle, thisBundle) < 0 {
-				return nil
-			}
-		}
-		resolvedBundle = &thisBundle
+		resolvedBundles = append(resolvedBundles, &packageFBC.Bundles[0])
+		matchedCatalogs = append(matchedCatalogs, cat.GetName())
 		resolvedDeprecation = thisDeprecation
```
tmshort, line 101: The updated code does not use deprecation to disambiguate, and I think it needs to. If we have the case where two bundles are found, but one of them is deprecated, we should not error out and should return the non-deprecated bundle.

Specifically, if `resolvedBundles` contains a deprecated bundle, and the new bundle is non-deprecated, we should ignore the deprecated bundles in `resolvedBundles`, and continue to add only non-deprecated bundles. Something like:

if list empty: add bundle to list, and update `resolvedDeprecation`
if list not empty:
  if `thisBundle` is not deprecated and list is deprecated: clear out list and add `thisBundle` to list and update `resolvedDeprecation`
  if `thisBundle` is deprecated and list is not deprecated: ignore new deprecated `thisBundle`
  otherwise (both new bundle and list share deprecation status): add `thisBundle` to list
  
What this does is make a homogeneous list of either all deprecated or all non-deprecated bundles. If we _then_ have multiple bundles, then it's an error.

Also note that `resolvedDeprecation` is always the last deprecation status, regardless of any other bundle on the list.

internal/resolve/catalog.go
tmshort: `"matching bundles found in multiple catalogs"` followed by a list of the catalogs or the list of catalogs+versions

tmshort: No need for "please specify a catalog"

internal/resolve/catalog.go
dtfranz: I may have simply remembered this wrong; I was under the impression that we'd give an ambiguous resolution error in this case as well. If that wasn't the agreement I'll put this back as a deciding factor.

tmshort:
You need to update the test-e2e, which will be a pain in the ass until the selector is available (see #1164) so we might want to merge that one in first...

tmshort:
You might want to rebase now that #1164 is merged.

internal/resolve/catalog.go
dtfranz: Updating this but I'm keeping `packages` vocabulary vs `bundles` since it matches more closely with the other errors and I think is more directly descriptive of the issue + what the user provided in `packageName`.

internal/resolve/catalog.go
tmshort: Deprecated was definitely kept; it was just version being the tie breaker. You have a good approach, because you have the list of found operations.

dtfranz:
PR has been rebased and updated with @tmshort 's e2e test.

tmshort (approved):
/lgtm

Issue #1146, opened 2024-08-19
Update fields in the spec to no longer be a pointer


everettraven:
As outlined in the [Pointers subsection of the OpenShift API Conventions](https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md#pointers), in configuration APIs, pointers should only be used when there is an explicit need to distinguish between a zero value and a nil value. An exception here is pointers in the Status subresource since a nil value has a different meaning in the Status subresource than the Spec.

We propose that the following fields, not having the explicit need to distinguish between the zero and nil values, should be updated to no longer be pointers:

The following fields do not have the explicit need to distinguish between the zero and nil values, meaning they should be updated to no longer be pointers based on the OpenShift API Conventions:
- `ClusterExtension.Spec.Preflight`: https://github.com/operator-framework/operator-controller/blob/baac660980549ef68329cd72b45906ddefa7f3c2/api/v1alpha1/clusterextension_types.go#L84-L85
- `ClusterExtension.Spec.Preflight.CRDUpgradeSafety`: https://github.com/operator-framework/operator-controller/blob/baac660980549ef68329cd72b45906ddefa7f3c2/api/v1alpha1/clusterextension_types.go#L103-L105

PR #1171, opened 2024-08-26
 ✨ Update fields in the spec to no longer be a pointer

rashmigottipati:
# Description

Fixes https://github.com/operator-framework/operator-controller/issues/1146

## Reviewer Checklist

- [ ] API Go Documentation


api/v1alpha1/clusterextension_types.go
```diff
@@ -319,7 +319,7 @@ type PreflightConfig struct {
 	// consequences of upgrading a CRD, such as data loss.
 	//
 	// This field is required if the spec.preflight field is specified.
-	CRDUpgradeSafety *CRDUpgradeSafetyPreflightConfig `json:"crdUpgradeSafety"`
+	CRDUpgradeSafety CRDUpgradeSafetyPreflightConfig `json:"crdUpgradeSafety,omitempty"`
```
everettraven, line 322: This field is a required field and therefore should not have the `omitempty` in the JSON tag:
```suggestion
	CRDUpgradeSafety CRDUpgradeSafetyPreflightConfig `json:"crdUpgradeSafety"`
```

api/v1alpha1/clusterextension_types.go
```diff
@@ -340,7 +340,7 @@ type CRDUpgradeSafetyPreflightConfig struct {
 	//
 	//+kubebuilder:validation:Enum:="Enabled";"Disabled"
 	//+kubebuilder:default:=Enabled
-	Policy CRDUpgradeSafetyPolicy `json:"policy"`
+	Policy CRDUpgradeSafetyPolicy `json:"policy,omitempty"`
```
everettraven, line 343: This is a required field and therefore should not have the `omitempty` JSON tag:
```suggestion
	Policy CRDUpgradeSafetyPolicy `json:"policy,"`
```

api/v1alpha1/clusterextension_types.go
rashmigottipati: I originally left it without the `omitempty` tag, however, unit tests are failing with the below error i.e. the field having a null value when it is empty:
`        	Error Trace:	/Users/rashmigottipati/go/src/github.com/operator-framework/operator-controller/internal/controllers/clusterextension_controller_test.go:63
        	Error:      	Received unexpected error:
        	            	ClusterExtension.olm.operatorframework.io "cluster-extension-test-s58x6kfr" is invalid: [spec.preflight.crdUpgradeSafety.policy: Unsupported value: "": supported values: "Enabled", "Disabled", <nil>: Invalid value: "null": some validation rules were not checked because the object was invalid; correct the existing errors to complete validation]`

 Since these fields are required only if the `spec.preflight` field is specified, there's a possibility of it being null when the `spec.preflight` field is not being set and so I think it would be ok to add the `omitempty` tag to this field and also the `Policy` field  but open to suggestions on how to appropriately handle this in a better way.

everettraven: Is there a way we could ensure the default value is set when the top level `spec.preflight` field is not explicitly set?

everettraven: if not, we may need to have some more complex CEL validations in place. We need to be careful there as well though to ensure the cost of the CEL expression is not too expensive.

everettraven: I found https://github.com/kubernetes-sigs/controller-tools/issues/622#issuecomment-1690128862 which may be what we need to do in this case.

joelanford: Keep in mind. This is required for now, only because there are no other sibling preflight check configurations. As soon as there are sibling configurations (e.g. for a permissions preflight check), then we'll a validation like "at least one of the preflight check fields is required".

everettraven: Thats a good point @joelanford . I do think it makes sense to go ahead and update it such that we have the "at least one of" approach

PR #1173, opened 2024-08-26
:warning: allow filtering with a list of channels

### LINKED ISSUE #1150: Update the `ClusterExtension.Spec.Channel` field to be a list of channels
Based on discussions for v1.0.0 features, we are heading into a reality where a version range can be specified and a set of catalogs can be specified. We should allow for a set of channels to be specified as well. This would mean:
- `ClusterExtension.Spec.Channel` would become `ClusterExtension.Spec.Channels`
- The field would remain optional
- The field would now be a list of channel names
- If the channel does not exist, no bundles will be returned for that channel (i.e nothing happens). If none of the channels exist, no bundle would be resolved and a resolution error would occur

everettraven:
# Description
- Updates the `ClusterExtension.Spec.Channel` field to now be `ClusterExtension.Spec.Channels` and accept a list of channel names for filtering of bundles during resolution
- Updates resolution logic, unit, and e2e tests as necessary

- Resolves #1150 

## Reviewer Checklist

- [ ] API Go Documentation


internal/controllers/clusterextension_controller.go
```diff
@@ -314,10 +314,16 @@ func SetDeprecationStatus(ext *ocv1alpha1.ClusterExtension, bundleName string, d
 			case declcfg.SchemaPackage:
 				deprecations[ocv1alpha1.TypePackageDeprecated] = entry
 			case declcfg.SchemaChannel:
-				if ext.Spec.Channel != entry.Reference.Name {
-					continue
+				// TODO: How should we handle doing channel deprecation with
+				// multiple channels being specified?
+				// For now, if it is possible it is from a deprecated channel
+				// set it.
```
everettraven, line 320: I wasn't sure how we should approach this so I went with this for now. I do wonder if we should even support "channel deprecation" when looking at multiple channels for filtering.

tmshort: I guess it matters if you care about "any" vs. "all". If you have to check for "all", then you might need a new map (or similar) to track the deprecation status for each channel.

internal/resolve/catalog.go
```diff
@@ -170,6 +165,39 @@ func (r *CatalogResolver) Resolve(ctx context.Context, ext *ocv1alpha1.ClusterEx
...
+type resolutionErrorInfo struct {
+	PackageName     string
+	Version         string
+	Channels        []string
+	Catalogs        []string
+	InstalledBundle *ocv1alpha1.BundleMetadata
+	ResolvedBundles []*declcfg.Bundle
+}
+
+func buildResolutionError(rei resolutionErrorInfo) error {
```
tmshort, line 177: I tried this once (an error generation function) and there was opposition to it. At that time, it was called in multiple places, this seems to be simply pulling out the code into a function.

tmshort: (And of course, this requires changing all the error messages)

internal/resolve/catalog_test.go
```diff
@@ -462,19 +462,24 @@ func TestCatalogWalker(t *testing.T) {
...
 			PackageName:             pkg,
-			Channel:                 channel,
 			Version:                 version,
 			UpgradeConstraintPolicy: upgradeConstraintPolicy,
 		},
 	}
+
+	if channel != "" {
+		ce.Spec.Channels = []string{channel}
+	}
```
tmshort, line 480: This feels like it should be done via a closure above.

internal/resolve/catalog.go
everettraven: I pulled it out into a separate function because the linter was complaining about the nesting depth being too high.

Regarding changing all the error messages, _I_ found it a bit more complex to read the previous error logic in a switch statement and thought it was a bit more readable using this new `strings.Builder` approach because it is more linear in nature (i.e as you read each line of the code you see the scenarios that result in new information being added to the error string we return).

everettraven: If folks prefer the previous approach, I can revert this change.

internal/resolve/catalog_test.go
everettraven: Do you mean instead of doing this, do something like:

```go
channels := []string{}
if channel != "" {
	channels = append(channels, channel)
}

return ...
```
?

everettraven: I'm fine with making a change here if you feel strongly about how it is implemented

everettraven: Updated in https://github.com/operator-framework/operator-controller/pull/1173/commits/e8ea4ff20524e3ef9bbdb3a64aeb87221c68a3d7 . Let me know if this is insufficient

tmshort: I was thinking something like this, so the assignment is all in one place. Not a big deal (don't even know if the following is legal Go!)
```
		Spec: ocv1alpha1.ClusterExtensionSpec{
			InstallNamespace:        "default",
			ServiceAccount:          ocv1alpha1.ServiceAccountReference{Name: "default"},
			PackageName:             pkg,
			Channel: func() []string{} {
				if channels != "" {
					return []string{channels}
				}
				return []string{}
			}(),
			Version:                 version,
			UpgradeConstraintPolicy: upgradeConstraintPolicy,
		},

Comment (everettraven) [reply internal/resolve/catalog_test.go]: I do think that is legal Go, but i'm not sure it is idiomatic. Going to keep it as is for now.

File: internal/resolve/catalog.go (hunk 2 above)
Comment (tmshort) [reply internal/resolve/catalog.go:177]: I don't care, I just wanted to make the note.

**Review by tmshort — APPROVED** [review]
/lgtm

File: internal/controllers/clusterextension_controller.go [hunk 4]
```diff
@@ -306,22 +307,32 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
...
 			case declcfg.SchemaChannel:
-				if ext.Spec.Source.Catalog.Channel != entry.Reference.Name {
-					continue
+				if channelSet.Has(entry.Reference.Name) {
+					if deps, ok := deprecations[ocv1alpha1.TypeChannelDeprecated]; ok {
+						deps = append(deps, entry)
+						deprecations[ocv1alpha1.TypeChannelDeprecated] = deps
+						continue
+					}
+					deprecations[ocv1alpha1.TypeChannelDeprecated] = []declcfg.DeprecationEntry{entry}
```
joelanford, line 329:
```suggestion
					deprecations[ocv1alpha1.TypeChannelDeprecated] = append(deprecations[ocv1alpha1.TypeChannelDeprecated], entry)
```

internal/controllers/clusterextension_controller_test.go
```diff
@@ -1137,7 +1133,7 @@ func TestSetDeprecationStatus(t *testing.T) {
 					Source: ocv1alpha1.SourceConfig{
 						SourceType: "Catalog",
 						Catalog: &ocv1alpha1.CatalogSource{
-							Channel: "badchannel",
+							Channels: []string{"badchannel"},
```
joelanford, line 1136: Add a test case with multiple deprecated channels?

internal/resolve/catalog.go
```diff
@@ -28,11 +31,17 @@ type CatalogResolver struct {
 	Validations      []ValidationFunc
 }
 
+type foundBundle struct {
+	bundle   *declcfg.Bundle
+	catalog  string
+	priority int32
+}
```
joelanford, line 38: Any reason this type needs to be used outside the `Resolve` function? If not, it seems better IMO to leave it scoped to that function.

internal/resolve/catalog.go
```diff
@@ -190,6 +178,42 @@ func (r *CatalogResolver) Resolve(ctx context.Context, ext *ocv1alpha1.ClusterEx
...
+type resolutionErrorInfo struct {
+	PackageName     string
+	Version         string
+	Channels        []string
+	InstalledBundle *ocv1alpha1.BundleMetadata
+	ResolvedBundles []foundBundle
+}
+
+func buildResolutionError(rei resolutionErrorInfo) error {
+	var errStringBuilder strings.Builder
```
joelanford, line 190: I might be in the minority, but I'd suggest a much shorter name for this variable since:
1. It is used in like 50% of the lines in this function.
2. The whole purpose of the function is to build up a string, so a short variable name will not really lose meaning
3. IMO, it would make the code look cleaner.

```suggestion
	var sb strings.Builder
```

internal/resolve/catalog.go
```diff
@@ -190,6 +178,42 @@ func (r *CatalogResolver) Resolve(ctx context.Context, ext *ocv1alpha1.ClusterEx
 	return resolvedBundle, resolvedBundleVersion, priorDeprecation, nil
 }
 
+type resolutionErrorInfo struct {
```
joelanford, line 181: Should we just make this type implement `error` and return it directly. (And then change `func buildResolutionError` to be `func (rei resolutionErrorInfo) Error() string`?

internal/controllers/clusterextension_controller.go
everettraven: Does this handle the case where `deprecations[ocv1alpha1.TypeChannelDeprecated]` has not yet been populated (i.e a new `[]declfg.DeprecationEntry` is initialized if this is the first time we are adding this key/value pair)?

internal/resolve/catalog.go
```diff
@@ -190,6 +178,42 @@ func (r *CatalogResolver) Resolve(ctx context.Context, ext *ocv1alpha1.ClusterEx
...
+
+	if len(rei.ResolvedBundles) > 1 {
+		matchedCatalogs := []string{}
+		for _, r := range rei.ResolvedBundles {
+			matchedCatalogs = append(matchedCatalogs, r.catalog)
+		}
+		slices.Sort(matchedCatalogs) // sort for consistent error message
+		errStringBuilder.WriteString(fmt.Sprintf("package %q found in multiple catalogs: %v", rei.PackageName, matchedCatalogs))
+	} else {
+		errStringBuilder.WriteString(fmt.Sprintf("no package %q found ", rei.PackageName))
```
joelanford, line 203: Looking at the tests, I wonder if we could improve terminology a bit in these cases. I feel like the current error messages have some ambiguity in the bundle vs package terms.

Change this:
```
package "foo" found in multiple catalogs: [fizz buzz] matching version ">=1.0.0"
```
To:
```
found bundles in package "foo" matching version ">=1.0.0" in channels [a b] in multiple catalogs of the same priority: [fizz buzz]
```


Change this:
```
no package "foo" found matching version ">=1.0.0" in channels [a b]
```
To:
```
no bundles found in package "foo" matching version ">=1.0.0" in channels [a b]
```

internal/resolve/catalog_test.go
```diff
@@ -935,3 +940,26 @@ func TestMultiplePriority(t *testing.T) {
...
+		"b": func() (*declcfg.DeclarativeConfig, *catalogd.ClusterCatalogSpec, error) {
+			return &declcfg.DeclarativeConfig{}, nil, nil
+		},
+		"c": func() (*declcfg.DeclarativeConfig, *catalogd.ClusterCatalogSpec, error) {
+			return genPackage(pkgName), nil, nil
+		},
+	}
+	r := CatalogResolver{WalkCatalogsFunc: w.WalkCatalogs}
+	ce := buildFooClusterExtension(pkgName, "beta", "", ocv1alpha1.UpgradeConstraintPolicyEnforce)
+	ce.Spec.Source.Catalog.Channels = append(ce.Spec.Source.Catalog.Channels, "alpha")
```
joelanford, line 959: This seems like it could be annoying to maintain. Change `buildFooClusterExtension` to accept a slice of channels?

internal/resolve/catalog.go
everettraven: I use it as a field type in the `resolutionErrorInfo` struct

internal/resolve/catalog_test.go
```diff
@@ -935,3 +940,26 @@ func TestMultiplePriority(t *testing.T) {
 	assert.Nil(t, gotVersion)
 	assert.Nil(t, gotDeprecation)
 }
+
+func TestMultipleChannels(t *testing.T) {
```
joelanford, line 944: We should make some change to ensure there's a different outcome between:
1. leave channels empty
2. set both `alpha` and `beta`

Maybe we should add a `gamma` channel in `genPackage` that has a `3.0.0` version in it. That way there would be different outcomes in the above cases:
1. select 3.0.0
2. select 2.0.0

internal/controllers/clusterextension_controller.go
everettraven: Used go playground to verify it does - TIL!

internal/resolve/catalog.go
everettraven: That sounds reasonable to me

internal/controllers/clusterextension_controller_test.go
```diff
@@ -1205,6 +1201,85 @@ func TestSetDeprecationStatus(t *testing.T) {
...
+							Type:               ocv1alpha1.TypePackageDeprecated,
+							Reason:             ocv1alpha1.ReasonDeprecated,
+							Status:             metav1.ConditionFalse,
+							ObservedGeneration: 1,
+						},
+						{
+							Type:               ocv1alpha1.TypeChannelDeprecated,
+							Reason:             ocv1alpha1.ReasonDeprecated,
+							Status:             metav1.ConditionTrue,
+							ObservedGeneration: 1,
```
joelanford, line 1252: Hmm, I didn't realize this test ignored the message. Follow-up?

internal/resolve/catalog.go
```diff
@@ -190,6 +177,46 @@ func (r *CatalogResolver) Resolve(ctx context.Context, ext *ocv1alpha1.ClusterEx
...
+	} else {
+		sb.WriteString(fmt.Sprintf("no bundles found for package %q ", rei.PackageName))
+	}
+
+	if rei.Version != "" {
+		sb.WriteString(fmt.Sprintf("matching version %q ", rei.Version))
+	}
+
+	if len(rei.Channels) > 0 {
+		sb.WriteString(fmt.Sprintf("in channels %v", rei.Channels))
```
joelanford, line 205:
```suggestion
		sb.WriteString(fmt.Sprintf("in channels %v ", rei.Channels))
```

internal/resolve/catalog.go
```diff
@@ -190,6 +177,46 @@ func (r *CatalogResolver) Resolve(ctx context.Context, ext *ocv1alpha1.ClusterEx
...
+	}
+
+	matchedCatalogs := []string{}
+	for _, r := range rei.ResolvedBundles {
+		matchedCatalogs = append(matchedCatalogs, r.catalog)
+	}
+	slices.Sort(matchedCatalogs) // sort for consistent error message
+	if len(matchedCatalogs) > 1 {
+		sb.WriteString(fmt.Sprintf("in multiple catalogs with the same priority %v", matchedCatalogs))
```
joelanford, line 214: Not strictly necessary, but following the pattern and making it easier for additional context to be added later:
```suggestion
		sb.WriteString(fmt.Sprintf("in multiple catalogs with the same priority %v ", matchedCatalogs))
```

internal/controllers/clusterextension_controller_test.go
everettraven: Is there a specific reason we should be testing the message?

I understand wanting to verify our message crafting logic is correct, but I lean towards the camp where the message is any arbitrary string and is not part of our API guarantees - testing for it explicitly makes our unit tests brittle

internal/resolve/catalog.go
everettraven: Should be updated in the latest push

internal/resolve/catalog.go
everettraven: Should be updated in the latest push

joelanford (approved):

PR #1174, opened 2024-08-26
🐛: Improve Catalog Selection Testing + Fix small issue

dtfranz:
We initially assumed that the resolvedBundle index we check doesn't matter when it comes to checking for prior package deprecations in the resolver but the index needs to match the prior result or the bundle names may not match and give a false negative.

Also improved the unit test cases (which caught the above issue).

# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/resolve/catalog.go
+++ b/internal/resolve/catalog.go
@@ -114,7 +114,7 @@ func (r *CatalogResolver) Resolve(ctx context.Context, ext *ocv1alpha1.ClusterEx
 		if len(resolvedBundles) != 0 {
 			// We've already found one or more package candidates
 			currentIsDeprecated := isDeprecated(thisBundle, thisDeprecation)
-			priorIsDeprecated := isDeprecated(*resolvedBundles[0], priorDeprecation) // Slice index doesn't matter; the whole slice is either deprecated or not
+			priorIsDeprecated := isDeprecated(*resolvedBundles[len(resolvedBundles)-1], priorDeprecation)
 			if currentIsDeprecated && !priorIsDeprecated {
 				// Skip this deprecated package and retain the non-deprecated package(s)
 				return nil
--- a/internal/resolve/catalog_test.go
+++ b/internal/resolve/catalog_test.go
@@ -236,49 +236,116 @@ func TestAcceptDeprecated(t *testing.T) {
 
 func TestPackageVariationsBetweenCatalogs(t *testing.T) {
 	pkgName := randPkg()
-	genImgRef := func(catalog, name string) string {
-		return fmt.Sprintf("%s/%s", catalog, name)
-	}
 	w := staticCatalogWalker{
 		"a": func() (*declcfg.DeclarativeConfig, error) { return &declcfg.DeclarativeConfig{}, nil },
 		"b": func() (*declcfg.DeclarativeConfig, error) {
-			fbc := genPackage(pkgName)
-			fbc.Bundles = append(fbc.Bundles, genBundle(pkgName, "1.0.3"))
-			for i := range fbc.Bundles {
-				fbc.Bundles[i].Image = genImgRef("catalog-b", fbc.Bundles[i].Name)
+			fbc := &declcfg.DeclarativeConfig{
+				Packages: []declcfg.Package{{Name: pkgName}},
+				Bundles:  []declcfg.Bundle{genBundle(pkgName, "1.0.0")},
+				Deprecations: []declcfg.Deprecation{{
+					Package: pkgName,
+					Entries: []declcfg.DeprecationEntry{
+						{
+							Reference: declcfg.PackageScopedReference{Schema: declcfg.SchemaBundle, Name: bundleName(pkgName, "1.0.0")},
+							Message:   fmt.Sprintf("bundle %s is deprecated", bundleName(pkgName, "1.0.0")),
+						},
+					},
+				}},
 			}
 			return fbc, nil
 		},
 		"c": func() (*declcfg.DeclarativeConfig, error) {
-			fbc := genPackage(pkgName)
-			fbc.Bundles = append(fbc.Bundles, genBundle(pkgName, "0.1.1"))
-			fbc.Deprecations = nil
-			for i := range fbc.Bundles {
-				fbc.Bundles[i].Image = genImgRef("catalog-c", fbc.Bundles[i].Name)
+			fbc := &declcfg.DeclarativeConfig{
+				Packages: []declcfg.Package{{Name: pkgName}},
+				Bundles:  []declcfg.Bundle{genBundle(pkgName, "1.0.1")},
+				Deprecations: []declcfg.Deprecation{{
+					Package: pkgName,
+					Entries: []declcfg.DeprecationEntry{
+						{
+							Reference: declcfg.PackageScopedReference{Schema: declcfg.SchemaBundle, Name: bundleName(pkgName, "1.0.1")},
+							Message:   fmt.Sprintf("bundle %s is deprecated", bundleName(pkgName, "1.0.1")),
+						},
+					},
+				}},
+			}
+			return fbc, nil
+		},
+		"d": func() (*declcfg.DeclarativeConfig, error) {
+			fbc := &declcfg.DeclarativeConfig{
+				Packages: []declcfg.Package{{Name: pkgName}},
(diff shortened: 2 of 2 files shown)


internal/resolve/catalog.go
```diff
@@ -114,7 +114,7 @@ func (r *CatalogResolver) Resolve(ctx context.Context, ext *ocv1alpha1.ClusterEx
 		if len(resolvedBundles) != 0 {
 			// We've already found one or more package candidates
 			currentIsDeprecated := isDeprecated(thisBundle, thisDeprecation)
-			priorIsDeprecated := isDeprecated(*resolvedBundles[0], priorDeprecation) // Slice index doesn't matter; the whole slice is either deprecated or not
+			priorIsDeprecated := isDeprecated(*resolvedBundles[len(resolvedBundles)-1], priorDeprecation)
```
everettraven, line 117: Am I correct in understanding that the expectation is that if the previously installed package is deprecated we expect it to be at the end of the slice of resolved bundles?

How would this work if there are more than one deprecated resolved bundles and the previously installed bundle is of higher semver than other deprecated bundles?

internal/resolve/catalog_test.go
```diff
@@ -236,46 +236,97 @@ func TestAcceptDeprecated(t *testing.T) {
...
+				Packages: []declcfg.Package{{Name: pkgName}},
+				Bundles:  []declcfg.Bundle{genBundle(pkgName, "1.0.4")},
 			}
 			return fbc, nil
 		},
 	}
 	r := CatalogResolver{WalkCatalogsFunc: w.WalkCatalogs}
 
-	t.Run("when catalog b has a newer version that matches the range", func(t *testing.T) {
+	t.Run("when package is deprecated in all but one catalog", func(t *testing.T) {
```
everettraven, line 306: Based on the test string, this seems like it is when the entire package is deprecated in all but one catalog but seems to be testing when a specific bundle is deprecated in all but one catalog.

What happens if there are multiple bundles, some with deprecations and some without?

As an example:
- Given package `foo` and version range `1.x`
- Given catalog A contains package `foo` with bundles `1.0.0`, `1.1.0` marked as deprecated but `1.2.0`, `1.3.0` not deprecated
- Given catalog B contains package `foo` with bundles `1.0.0`, `1.1.0` and `1.2.0` marked as deprecated but `1.3.0` and `1.4.0` not deprecated

What happens?

internal/resolve/catalog.go
dtfranz: The package at the end of the slice isn't the currently installed package, it's just the package we got from the previous catalog we looked through. 

But regardless, in your scenario, if the user isn't filtering out packages of a lower version (`UpgradeConstraintPolicyIgnore=true`) and our resolver finds multiple packages that are also deprecated then we would have an ambiguous resolution i.e. we will not decide for the user which package in which catalog we will downgrade to.

internal/resolve/catalog_test.go
dtfranz: You're right on the wording, I'm conflating `package` and `bundle` here. I'll fix that :) 

In your scenario we would give an ambiguous resolution because we have multiple resolved candidates from multiple catalogs:
* `1.2.0` from catalog `A`
* `1.3.0` from catalog `A` and `B` 
* `1.4.0` from catalog `B`

The deprecated bundles won't be considered at all since we encountered greater than zero non-deprecated bundles.
In more fine-grained detail:
* We check catalog A
  * resolved list is empty, add 1.0.0
  * 1.1.0 is also deprecated, add 1.1.0 to the resolved list
  * 1.2.0 is not deprecated, clear the list and add 1.2.0
* We check catalog B
  * 1.0.0 is deprecated, 1.2.0 in the list is not, skip
  * same for 1.1.0
  * same for 1.2.0 - probably not a valid scenario but one I think we need to handle
  * 1.3.0 is not deprecated, add it to the list
  * 1.4.0 is not deprecated, add it to the list

So we finish resolution with candidates `1.2.0(A), 1.3.0(A), 1.3.0(B), 1.4.0(B)`, and the error generated would be `matching packages found in multiple catalogs: [A B]`

You made me realize though, and maybe this was your intent, we need to adjust the error when there are multiple resolved packages in one catalog. But then again, my understanding from the RFC was that we were removing the highest-semver decision entirely, but looking at this again it seems like we should keep highest-semver when we only have one catalog to choose from, right?

tmshort (dismissed):
/lgtm

Unless @everettraven's issues have not been resolved.

internal/resolve/catalog_test.go
tmshort: Within a catalog, it should be highest semver, there should only be one return from a given catalog. Since we support automatic updates, this has to be the case. The ambiguity comes from when there are two operators in different catalogs, and we're not sure which operator the user meant, regardless of semver.

So, in @everettraven's example, we should have:
1. Picked `foo-1.3.0` from catalog `A`: 1.0.0 and 1.1.0 are deprecated, and thus ignored when 1.2.0 is found. 1.3.0 takes precedence over 1.2.0 due to semver.
2. Picked `foo-1.4.0` from catalog `B`. 1.0.0, 1.1.0, and 1.2.0 are deprecated, and thus ignored when 1.3.0 is found. 1.4.0 takes precedence over 1.3.0 due to semver.

Now, there are two bundles with the same name, but different versions. 

So, I don;'t think we can completely get rid of semver, because we might run into the the situation you have (i.e. one catalog with multiple candidates), and we should pick the highest semver (that fulfills the `version` requirement).

So, perhaps the `resolvedBundles` array needs to be a map indexed by catalog? Or something similar? Still have to deal with deprecated though. Before these changes, only one bundle was ever returned, either from 1 catalog or _n_ catalogs. I think we need to choose the best bundle from _each_ catalog, and then make sure we don't have ambiguity after that process.

tmshort: TL;DR: pick the best bundle from each catalog; if there are multiple results, then look at deprecation status. If there are multiple non-deprecated or all bundles are deprecated and there are multiple deprecated, then there's ambiguity, and we error out.

So, if there is a deprecated bundle from catalog A, and a non-deprecated bundle from catalog B, the bundle from catalog B wins.

dtfranz: Apologies to you both - I confused myself here and forgot some key detail.

When we're looking at one catalog, if we get multiple bundle candidates, we sort them by semver and deprecation. 

That means that each time we run this `func`, we're already choosing the best candidate per catalog, and there's no change needed.

The key thing here is that if more than one catalog is giving us back a "best candidate", and all the candidates share deprecated or non-deprecated status, that's when we give the error. The code already takes care of this well, AFAICT.

Of course if either of you still think I'm missing something here let me know! In the meantime though I'll add another test case to hopefully make this clearer.

dtfranz: https://github.com/operator-framework/operator-controller/pull/1174/files#diff-0eaea52edc2ea467f970f25ee8c9a64e6038cfdfd040284c65bd0195409f7b0bR341

internal/resolve/catalog_test.go
```diff
@@ -236,50 +236,117 @@ func TestAcceptDeprecated(t *testing.T) {
...
 		require.Error(t, err)
-		assert.ErrorContains(t, err, "found in multiple catalogs: [b c]")
+		// We will not make a decision on which catalog to use
+		assert.ErrorContains(t, err, "found in multiple catalogs: [d f]")
 		assert.Nil(t, gotBundle)
 		assert.Nil(t, gotVersion)
 		assert.Nil(t, gotDeprecation)
 	})
+
+	t.Run("highest semver bundle is chosen when candidates are all from the same catalog", func(t *testing.T) {
```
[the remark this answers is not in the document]
joelanford: Nvmd. I see `PreferNonDeprecated` and `AcceptDeprecated` tests.

internal/resolve/catalog_test.go
everettraven: All of this sounds and looks good to me. I mostly wanted to verify my understanding of the expected behaviors, so thank you both for the detailed answers!

everettraven (approved):

internal/resolve/catalog_test.go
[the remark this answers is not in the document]
tmshort: Created https://github.com/operator-framework/operator-controller/issues/1176 (just to make sure)

PR #1175, opened 2024-08-26
:warning: update name validation patterns to match Kubernetes as close as possible

### LINKED ISSUE #1149: Update validations for fields referencing the name of something to follow Kubernetes
Where possible we should restrict validation of fields representing names of things to follow the stricter RFC 1035 naming conventions. Where we need less restrictive naming validation we should fall back to the RFC 1123 conventions. We should not deviate from the validations outlined in https://kubernetes.io/docs/concepts/overview/working-with-objects/names/ 

The following fields have been identified as needing updates to adhere to the above mentioned criteria:
- `ClusterExtension.Spec.PackageName` should be updated to follow the RFC 1123 DNS Label pattern and validations. We fall back to RFC 1123 for this field due to packages existing in the current ecosystem that start with numeric characters (3scale-operator is one such example).
- `ClusterExtension.Spec.Channel` would be updated to follow the RFC 1123 DNS Label pattern and validations. We fall back to RFC 1123 for this field due to channels existing in the current ecosystem that start with numeric characters (1.1.x is one such example that exists in the OperatorHub community catalog today).
- `ClusterExtension.Spec.InstallNamespace` would be updated to follow the RFC 1123 DNS Label pattern and validations. This is the restriction on `Namespace` names by Kubernetes ([ref](https://github.com/kubernetes/kubernetes/blob/e54c8ef2024e638d721242224f6f925b15ee43f5/staging/src/k8s.io/apimachinery/pkg/api/validation/generic.go#L63)).
- `ClusterExtension.Spec.ServiceAccount.Name` would be updated to follow the RFC 1123 DNS Subdomain pattern and validations. This is the same validation that Kubernetes uses for `ServiceAccount` names ([ref](https://github.com/kubernetes/kubernetes/blob/e54c8ef2024e638d721242224f6f925b15ee43f5/staging/src/k8s.io/apimachinery/pkg/api/validation/generic.go#L68)).

The RFC 1123 DNS Label validations can be found here: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names

The RFC 1123 Subdomain validations can be found here: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names

everettraven:
# Description
- Updates existing field comments and validations to adhere more closely to Kubernetes naming conventions

- Resolves #1149 

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/api/v1alpha1/clusterextension_types.go
+++ b/api/v1alpha1/clusterextension_types.go
@@ -67,7 +67,7 @@ type ClusterExtensionSpec struct {
 	// applied to other Namespaces. This Namespace is expected to exist.
 	//
 	// installNamespace is required, immutable, and follows the DNS label standard
-	// as defined in RFC 1123. This means that valid values:
+	// as defined in [RFC 1123]. This means that valid values:
 	//   - Contain no more than 63 characters
 	//   - Contain only lowercase alphanumeric characters or '-'
 	//   - Start with an alphanumeric character
@@ -84,6 +84,8 @@ type ClusterExtensionSpec struct {
 	//   - some-namespace-
 	//   - thisisareallylongnamespacenamethatisgreaterthanthemaximumlength
 	//   - some.namespace
+	//
+	// [RFC 1123]: https://tools.ietf.org/html/rfc1123
 	//
 	//+kubebuilder:validation:Pattern:=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
 	//+kubebuilder:validation:MaxLength:=63
@@ -140,11 +142,10 @@ type CatalogSource struct {
 	// packageName is a reference to the name of the package to be installed
 	// and is used to filter the content from catalogs.
 	//
-	// This field is required, immutable and follows the DNS label standard as defined in RFC
-	// 1123, with a deviation in the maximum length being no more than 48
-	// characters. This means that valid values:
-	//   - Contain no more than 48 characters
-	//   - Contain only lowercase alphanumeric characters or '-'
+	// This field is required, immutable and follows the DNS subdomain name
+	// standard as defined in [RFC 1123]. This means that valid entries:
+	//   - Contain no more than 253 characters
+	//   - Contain only lowercase alphanumeric characters, '-', or '.'
 	//   - Start with an alphanumeric character
 	//   - End with an alphanumeric character
 	//
@@ -160,8 +161,10 @@ type CatalogSource struct {
 	//   - thisisareallylongpackagenamethatisgreaterthanthemaximumlength
 	//   - some.package
 	//
-	//+kubebuilder:validation:MaxLength:=48
-	//+kubebuilder:validation:Pattern:=^[a-z0-9]+(-[a-z0-9]+)*$
+	// [RFC 1123]: https://tools.ietf.org/html/rfc1123
+	//
+	//+kubebuilder:validation:MaxLength:=253
+	//+kubebuilder:validation:Pattern:=^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$
 	//+kubebuilder:validation:XValidation:rule="self == oldSelf",message="packageName is immutable"
 	PackageName string `json:"packageName"`
 
@@ -259,10 +262,9 @@ type CatalogSource struct {
 	//
 	// When unspecified, upgrade edges across all channels will be used to identify valid automatic upgrade paths.
 	//
-	// This field follows the DNS subdomain name standard as defined in RFC
-	// 1123, with a deviation in the maximum length being no more than 48
-	// characters. This means that valid values:
-	//   - Contain no more than 48 characters
+	// This field follows the DNS subdomain name standard as defined in [RFC
+	// 1123]. This means that valid entries:
+	//   - Contain no more than 253 characters
 	//   - Contain only lowercase alphanumeric characters, '-', or '.'
 	//   - Start with an alphanumeric character
 	//   - End with an alphanumeric character
@@ -281,9 +283,13 @@ type CatalogSource struct {
 	//   - -some-channel
 	//   - some-channel-
 	//   - thisisareallylongchannelnamethatisgreaterthanthemaximumlength
+	//   - original_40
+	//   - --default-channel
 	//
-	//+kubebuilder:validation:MaxLength:=48
-	//+kubebuilder:validation:Pattern:=^[a-z0-9]+([\.-][a-z0-9]+)*$
+	// [RFC 1123]: https://tools.ietf.org/html/rfc1123
+	//
+	//+kubebuilder:validation:MaxLength:=253
+	//+kubebuilder:validation:Pattern:=^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$
 	//+optional
 	Channel string `json:"channel,omitempty"`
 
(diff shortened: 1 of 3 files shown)


api/v1alpha1/clusterextension_types.go
```diff
@@ -49,10 +49,9 @@ type ClusterExtensionSpec struct {
 	// packageName is a reference to the name of the package to be installed
 	// and is used to filter the content from catalogs.
 	//
-	// This field is required, immutable and follows the DNS label standard as defined in RFC
-	// 1123, with a deviation in the maximum length being no more than 48
-	// characters. This means that valid values:
-	//   - Contain no more than 48 characters
+	// This field is required, immutable and follows the DNS label standard as
+	// defined in [RFC 1123]. This means that valid values:
+	//   - Contain no more than 63 characters
```
joelanford, line 54: We had originally decided on a lower maximum so that we left a few characters for controllers and other clients to use packages names in their own objects with prefixes or suffixes totaling 15 characters.

Did we talk specifically about eliminating this cushion?

api/v1alpha1/clusterextension_types.go
```diff
@@ -168,11 +169,11 @@ type ClusterExtensionSpec struct {
...
 	//
-	// This field follows the DNS subdomain name standard as defined in RFC
-	// 1123, with a deviation in the maximum length being no more than 48
-	// characters. This means that valid values:
-	//   - Contain no more than 48 characters
-	//   - Contain only lowercase alphanumeric characters, '-', or '.'
+	// This field follows the DNS label name standard as defined in [RFC 1123]
+	// with the addition of '.' and '_' characters being allowed.
+	// This means that valid values:
+	//   - Contain no more than 63 characters
```
joelanford, line 175: Similar question here re: removing the 15 character cushion.

api/v1alpha1/clusterextension_types.go
joelanford, line 1: Do the before or after changes from this PR align to the validations that happen in `opm validate`? We need to make sure that package and channel names are fully specified and validated the exact same way in both places.

everettraven: As far as I can tell, the validation for the package name does match. There does not appear to be any validation on channel names in `opm validate`. 

IMO we should merge the validation changes here and then add the validation of the channel name to `opm validate`.

api/v1alpha1/clusterextension_types.go
everettraven: Could you provide an example of a use case where the 48 character limit would be beneficial and actually be used? An additional note, the [package name validation](https://github.com/operator-framework/operator-registry/blob/2e0d6a237a1ae06998584043e44c44029f5df352/alpha/declcfg/declcfg_to_model.go#L26-L28) in `opm validate` uses the `validation.IsDNS1123Label` function implemented in https://github.com/kubernetes/kubernetes/blob/7436ca32bc766ff202109a7541d2e7bb41ee7d13/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L185-L202

which validates a maximum length of 63 characters. In your scenario, a package that has gone through `opm validate` may not be able to be installed due to the 48 character length limit we are imposing in the `ClusterExtension` API.

joelanford: The 48 character limit I think comes from the early days of rukpak. And the idea was originally because rukpak itself used that package name in order to create other objects on the cluster. So we decided to give ourselves and other rukpak provisioner developers 15 characters of slush in case they wanted to use package names to construct other object names.

That's a good point about `opm validate`. If we wanted to keep things at 48, we'd have to double check all known catalogs to see if that would break anything.

api/v1alpha1/clusterextension_types.go
joelanford: On second thought, I'm definitely good with 63 here. I can't imagine there being any cases where one would want channel names in other objects.

api/v1alpha1/clusterextension_types.go
joelanford: I think we need to check known catalogs against these proposed validations before we merge so that we have more information to make a decision. We may ultimately decide to implement these validations despite breaking existing content, but we should know before we act.

everettraven: Gist that verified this validation works for all channel names in all known catalogs (except for one in the redhat certified operators index that looks like an unintentional channel name): https://gist.github.com/everettraven/4691c8cc6181f28b129f46cceeb602b5

api/v1alpha1/clusterextension_types.go
```diff
@@ -168,11 +169,11 @@ type ClusterExtensionSpec struct {
 	//
 	// When unspecified, upgrade edges across all channels will be used to identify valid automatic upgrade paths.
 	//
-	// This field follows the DNS subdomain name standard as defined in RFC
-	// 1123, with a deviation in the maximum length being no more than 48
-	// characters. This means that valid values:
-	//   - Contain no more than 48 characters
-	//   - Contain only lowercase alphanumeric characters, '-', or '.'
+	// This field follows the DNS label name standard as defined in [RFC 1123]
+	// with the addition of '.' and '_' characters being allowed.
```
joelanford, line 173: I assume `.` and `_` are disallowed as the first or last characters? Seems like that's an important detail to include.

internal/controllers/clusterextension_admission_test.go
joelanford, line 1: Add more tests for `_` and `.` being allowed (but not as first or last characters)?

api/v1alpha1/clusterextension_types.go
joelanford, line 173: Is this valid: `a._-b`?

If it is, I don't think it should be.

everettraven: I do believe that would be valid

internal/controllers/clusterextension_admission_test.go
everettraven: Since we updated to use the RFC 1123 DNS Subdomain pattern I added tests for the `.` character being allowed (but not as first character) and `_` not being allowed at all

api/v1alpha1/clusterextension_types.go
everettraven: With the new use of the RFC 1123 DNS Subdomain pattern, something like `a.-b` is not valid. I will add another admission test for this.

internal/controllers/clusterextension_admission_test.go
```diff
@@ -243,6 +248,9 @@ func TestClusterExtensionAdmissionChannel(t *testing.T) {
 		{"ends with hyphen", "end-with-hyphen-", regexMismatchError},
 		{"starts with period", ".start-with-period", regexMismatchError},
 		{"ends with period", "end-with-period.", regexMismatchError},
+		{"contains underscore", "some_thing", regexMismatchError},
+		{"starts with dot", ".some.thing", regexMismatchError},
```
everettraven, line 252:
```suggestion
```
already existing test

api/v1alpha1/clusterextension_types.go
everettraven: should be added now

api/v1alpha1/clusterextension_types.go
```diff
@@ -67,7 +67,7 @@ type ClusterExtensionSpec struct {
 	// applied to other Namespaces. This Namespace is expected to exist.
 	//
 	// installNamespace is required, immutable, and follows the DNS label standard
```
joelanford, line 69:
```suggestion
	// installNamespace is required, immutable, and follows the DNS subdomain standard
```

joelanford (approved):

PR #1191, opened 2024-08-29
✨ Add ServedVersionValidator preflight check

### LINKED ISSUE #1091: CRD Upgrade Safety preflight checks should check for incompatibility between served versions
After speaking with folks more familiar with the nuances of CRD management, we need to update the CRD Upgrade Safety preflight check to add a check that for _new_ instances of the CRD only:
- Checks compatibility across all served versions, failing if there is incompatibility
- Fails if incompatibility is found _AND_ no conversion strategy is specified
- Passes if conversion webhook is specified as the conversion strategy

For historical context the previous description of this issue is saved below. It can be used to get clarity on the original concern, but all that is needed for implementation is what is mentioned above.
## History

The existing CRD Upgrade Safety preflight checks are limited to checking for breaking changes across existing versions of a CRD (i.e old v1alpha1 vs new v1alpha1), but would not handle the following case:
- CRD Foo v1alpha1 has optional field bar, required field baz and is a stored version
- CRD Foo v1alpha2 is introduced and only has required field bar
- No conversion strategy is specified

Any instances of the v1alpha1 API would be unable to successfully convert to the v1alpha2 API, thus invalidating any instances of v1alpha1 in etcd.

We need to add a validation to the CRD Upgrade Safety preflight check that:
- Checks for breaking changes between existing stored versions (from the old CRD status), and served versions (from the new CRD)
- Returns an error if any breaking changes are found and a conversion strategy is NOT specified
    - If a conversion strategy is specified we could log and/or add an event to warn (similar to OLMv0) or do nothing

trgeiger:
New preflight check that checks for either the presence of the "Webhook" conversion strategy to pass, or checks for any incompatible changes between served versions in the new CRD which, if found, causes failure..

Closes #1091

# Description
Adds a new preflight check, ServedVersionValidator, which checks for breaking changes between served versions in the new CRD when upgrading.

Happy to change up any naming or error messages.

Existing test data had to be edited so it wouldn't trigger the new check--the empty v1 in all the test data has just been filled in to match the test data v2.

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/rukpak/preflights/crdupgradesafety/checks.go
+++ b/internal/rukpak/preflights/crdupgradesafety/checks.go
@@ -1,0 +1,72 @@
+package crdupgradesafety
+
+import (
+	"errors"
+	"fmt"
+	"slices"
+
+	kappcus "carvel.dev/kapp/pkg/kapp/crdupgradesafety"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	versionhelper "k8s.io/apimachinery/pkg/version"
+)
+
+type ServedVersionValidator struct {
+	Validations []kappcus.ChangeValidation
+}
+
+func (c *ServedVersionValidator) Validate(old, new apiextensionsv1.CustomResourceDefinition) error {
+	// If conversion webhook is specified, pass check
+	if new.Spec.Conversion != nil && new.Spec.Conversion.Strategy == apiextensionsv1.WebhookConverter {
+		return nil
+	}
+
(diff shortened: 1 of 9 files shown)


everettraven (changes requested):
Overall the changes look really good! Great work @trgeiger !

I have a couple minor comments, but ones I think are worth addressing

trgeiger:
Thanks for the review and suggestions. All ready to go again.

everettraven (approved):
Looks good to me!

PR #1193, opened 2024-08-29
:sparkles: Use a catalog's priority to as a bundle selection tie-breaker

tmshort:
Fixes: #1117 

Lots of changes to the catalog test infrastructure, as I needed to include information about the catalog itself.

# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/resolve/catalog.go
+++ b/internal/resolve/catalog.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"slices"
+	"sort"
 
 	mmsemver "github.com/Masterminds/semver/v3"
 	bsemver "github.com/blang/semver/v4"
@@ -50,9 +51,14 @@ func (r *CatalogResolver) Resolve(ctx context.Context, ext *ocv1alpha1.ClusterEx
 		}
 	}
 
+	type foundBundle struct {
+		bundle   *declcfg.Bundle
+		catalog  string
+		priority int32
+	}
+
+	resolvedBundles := []foundBundle{}
 	var (
-		resolvedBundles  []*declcfg.Bundle
-		matchedCatalogs  []string
 		priorDeprecation *declcfg.Deprecation
 	)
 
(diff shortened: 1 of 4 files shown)


internal/resolve/catalog_test.go
```diff
@@ -727,3 +854,44 @@ func TestClusterExtensionNoMatchLabel(t *testing.T) {
...
+			return genPackage(pkgName), &catalogd.ClusterCatalogSpec{Priority: 1}, nil
+		},
+		"b": func() (*declcfg.DeclarativeConfig, *catalogd.ClusterCatalogSpec, error) {
+			return genPackage(pkgName), &catalogd.ClusterCatalogSpec{Priority: 0}, nil
+		},
+	}
+	r := CatalogResolver{WalkCatalogsFunc: w.WalkCatalogs}
+
+	ce := buildFooClusterExtension(pkgName, "", ">=1.0.0 <=1.0.1", ocv1alpha1.UpgradeConstraintPolicyEnforce)
+	_, _, _, err := r.Resolve(context.Background(), ce, nil)
```
dtfranz, line 871: I think we should ensure we actually got the right bundle here (from catalog `"a"`)

internal/resolve/catalog.go
```diff
@@ -114,33 +120,47 @@ func (r *CatalogResolver) Resolve(ctx context.Context, ext *ocv1alpha1.ClusterEx
...
-		matchedCatalogs = append(matchedCatalogs, cat.GetName())
+		resolvedBundles = append(resolvedBundles, foundBundle{&thisBundle, cat.GetName(), cat.Spec.Priority})
 		priorDeprecation = thisDeprecation
 		return nil
 	}, listOptions...); err != nil {
 		return nil, nil, nil, fmt.Errorf("error walking catalogs: %w", err)
 	}
 
+	// Resolve for priority
+	if len(resolvedBundles) > 1 {
```
dtfranz, line 142: Couldn't we resolve for priority while we're in the process of walking through the catalogs? Similar to how we ignore a bundle when it's deprecated and the last one was not, couldn't we ignore a bundle when the priority is lower than the last resolved bundle (and it shares deprecation status)? Not that it gains us much because we don't have that many catalogs to sort through later, but just a thought.

tmshort: We still want to be able to list the catalogs, including those that don't have a maximal priority. If you see the test I added, `TestMultiplePriority`, all 3 catalogs are listed, not just the two highest priority. If this is not what we want, then, yes, it might make sense to adjust the list as we create it. Keeping it separate like this, makes it easier to follow, IMHO.

internal/resolve/catalog_test.go
tmshort: Not sure we can actually check that out, since we don't return the catalog it comes from...

internal/resolve/catalog.go
tmshort: Tradeoff between performance and simplicity/readability.

internal/resolve/catalog_test.go
dtfranz: What we've done in other tests is populated the different catalogs with different bundle versions, so i.e. catalog "a" has 1.0.0 and "b" has 1.0.1, so if you want the bundle from catalog a you can just check that the bundle version matches. Just make sure of course that the semver matches both (which it does now as written).

tmshort: Fair, I cheated and just used `genPackage()`

dtfranz (approved):
/lgtm

PR #1194, opened 2024-08-30
✨unpacker: switch from google/go-containerregistry to containers/image

joelanford:
# Description

See https://docs.google.com/document/d/1s-nfpJqecWGuriFIVXr2REbZLUB8N8QxLp1MQKLPhlE/edit#heading=h.x3tfh25grvnv

## Reviewer Checklist

- [ ] API Go Documentation


internal/controllers/clusterextension_controller.go
```diff
@@ -237,6 +237,7 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
 	setResolvedStatusConditionSuccess(ext, fmt.Sprintf("resolved to %q", resolvedBundle.Image))
 
 	bundleSource := &rukpaksource.BundleSource{
+		Name: ext.GetName(),
```
joelanford, line 240: Seems like we must have a bug if we're not setting this. I think it means that:
1. All cluster extensions unpacking is happening in the same directory
2. Because of (1), only the most recently reconciled bundle will have its FS extracted at once.
3. Because of (2), when a ClusterExtension reconciles it has to re-pull and re-unpack its image if another ClusterExtension has reconciled in the meantime.
4. (Not a problem today) If we turned on concurrent reconciles, then things would likely break even harder because the unpacker would be racing to delete the what it thinks are unused unpack paths while other reconciles are writing to them for other ClusterExtensions.

internal/rukpak/source/containers_image.go
```diff
@@ -0,0 +1,274 @@
...
+	if unpackStat, err := os.Stat(unpackPath); err == nil {
+		if !unpackStat.IsDir() {
+			return nil, fmt.Errorf("unexpected file at unpack path %q: expected a directory", unpackPath)
+		}
+		l.Info("image already unpacked", "ref", imgRef.String(), "digest", canonicalRef.Digest().String())
+		return successResult(bundle.Name, unpackPath, canonicalRef), nil
+	}
+	//if err := os.MkdirAll(i.bundlePath(bundle.Name), 0755); err != nil {
+	//	return nil, fmt.Errorf("error creating unpack directory: %w", err)
+	//}
```
everettraven, line 73: Remove this?

internal/rukpak/source/containers_image_test.go
```diff
@@ -0,0 +1,397 @@
...
+		// Attempt to pull and unpack the image
+		_, err := unpacker.Unpack(context.Background(), bundleSource)
+		if err != nil {
+			t.Error("func should have panicked")
+		}
+	}
+	require.Panics(t, shouldPanic)
+}
+
+func TestUnpackInvalidNilImage(t *testing.T) {
```
everettraven, line 218: Should we ensure that the returned error here is the unrecoverable type?

internal/rukpak/source/containers_image_test.go
```diff
@@ -0,0 +1,397 @@
...
+	}
+
+	// Attempt to unpack
+	result, err := unpacker.Unpack(context.Background(), bundleSource)
+	require.ErrorContains(t, err, "nil image source")
+	require.NoDirExists(t, filepath.Join(unpacker.BaseCachePath, bundleSource.Name))
+	assert.Nil(t, result)
+}
+
+func TestUnpackInvalidImageRef(t *testing.T) {
```
everettraven, line 236: Should we ensure the error type returned is unrecoverable?

joelanford: [f21e3e5003b70b5b14eb73fc3c5170dd4f147425](https://github.com/operator-framework/operator-controller/compare/df384cd6cb07fbd53f8d037b94d24a96a60c50b9..f21e3e5003b70b5b14eb73fc3c5170dd4f147425)

internal/rukpak/source/containers_image.go
joelanford: https://github.com/operator-framework/operator-controller/compare/df384cd6cb07fbd53f8d037b94d24a96a60c50b9..f21e3e5003b70b5b14eb73fc3c5170dd4f147425

everettraven (approved):

cmd/manager/main.go
```diff
@@ -194,14 +195,18 @@ func main() {
 		setupLog.Error(err, "unable to create CA certificate pool")
 		os.Exit(1)
 	}
-	unpacker := &source.ImageRegistry{
-		BaseCachePath:   filepath.Join(cachePath, "unpack"),
-		CertPoolWatcher: certPoolWatcher,
```
tmshort, line 199: Loss of `CertPoolWatcher: certPoolWatcher` here...

internal/controllers/clusterextension_controller.go
tmshort: This this in a separate PR?

cmd/manager/main.go
tmshort: Not seeing `ContainersImageRegistry` use certificate, so is this necessary?

tmshort: Yes, in BuildHTTPClient

joelanford: Yeah, the containers/image source implementation lets us set the CA files on a context struct, which I think are reloaded on every pull.

And yeah, the certPoolWatcher _should_ still be used to connect to catalogd.

internal/rukpak/source/containers_image.go
```diff
@@ -0,0 +1,271 @@
...
+		return nil, fmt.Errorf("error getting digest of manifest: %w", err)
+	}
+	return reference.WithDigest(reference.TrimNamed(imgRef), imgDigest)
+}
+
+func (i *ContainersImageRegistry) unpackImage(ctx context.Context, unpackPath string, imageReference types.ImageReference) error {
+	img, err := imageReference.NewImage(ctx, i.SourceContext)
+	if err != nil {
+		return fmt.Errorf("error reading image: %w", err)
+	}
```
thetechnick, line 207:
```suggestion
	img, err := imageReference.NewImage(ctx, i.SourceContext)
	if err != nil {
		return fmt.Errorf("error reading image: %w", err)
	}
	defer func() {
		if err := img.Close(); err != nil {
			panic(err)
		}
	}()
```
Missing call to img.Close().

cmd/manager/main.go
```diff
@@ -194,14 +195,22 @@ func main() {
...
+	unpacker := &source.ContainersImageRegistry{
+		BaseCachePath: filepath.Join(cachePath, "unpack"),
+		SourceContext: &types.SystemContext{
+			DockerCertPath: caCertDir,
+			OCICertPath:    caCertDir,
+		},
+	}
+	defaultAuthConfigPath := filepath.Join("/etc", "containers", "auth.json")
+	if _, err := os.Stat(defaultAuthConfigPath); err == nil {
+		unpacker.SourceContext.AuthFilePath = defaultAuthConfigPath
```
joelanford, line 208: This doesn't seem to be a default for `containers/image`, interestingly. Researching this a bit more.

internal/rukpak/source/containers_image.go
joelanford: Panic, you think? We could also pull a logger out of the context and log the error.

thetechnick: Nobody is going to see that log.
I'd either panic or raise the error further up the stack.

cmd/manager/main.go
everettraven: Primary read/write file for auth seems to be `${XDG_RUNTIME_DIR}/containers/auth.json` or `$HOME/.config/containers/auth.json`: https://github.com/containers/image/blob/main/docs/containers-auth.json.5.md#description

everettraven: Looking at this, we should probably revisit the RFC and highlight the default location for all the configs we may care about

joelanford: So it turns out that file location is just not a thing (yet at least): https://github.com/containers/image/pull/1746

joelanford: We could follow the pattern that podman and skopeo do. If `REGISTRY_AUTH_FILE` is set in the environment, we set its value on this field. I'll switch the PR to do that for now.

internal/rukpak/source/containers_image.go
joelanford: Panic it is. I expect this will never happen, but if it does, agree that it is likely something for a maintainer to look into.

cmd/manager/main.go
joelanford: I took all of the auth stuff back out of the PR for now. We should discuss and handle separately.

This PR is now focused on `containers/image` and use of `/etc/containers` system configuration (as an unsupported/subject-to-change internal API)

internal/controllers/clusterextension_controller.go
joelanford: https://github.com/operator-framework/operator-controller/pull/1243

internal/rukpak/source/unpacker.go
```diff
@@ -2,10 +2,28 @@ package source
...
+	// Ref contains the reference to a container image containing Bundle contents.
+	Ref string
+}
+
+// Unrecoverable represents an error that can not be recovered
+// from without user intervention. When this error is returned
+// the request should not be requeued.
+type Unrecoverable struct {
+	error
+}
```
tmshort, line 21: Is there an explicit check for this type of error?

config/components/registries-conf/registries_conf_configmap.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: e2e-registries-conf
+  namespace: olmv1-system
```
tmshort, line 5:
```suggestion
  namespace: system
```
The namespace is overwritten by the value in `kustomization.yaml`?

internal/rukpak/source/containers_image_test.go
```diff
@@ -0,0 +1,400 @@
...
+		},
+	}
+
+	oldBundlePath := filepath.Join(unpacker.BaseCachePath, bundleSource.Name, "old")
+	err := os.MkdirAll(oldBundlePath, 0755)
+	require.NoError(t, err)
+
+	// Attempt to pull and unpack the image
+	result, err := unpacker.Unpack(context.Background(), bundleSource)
+	// Check Result
```
tmshort, line 54: I think all the `// Check Result` comments might be a little redundant...

tmshort (commented):
A few comments. Mostly surrounding the need for a new Issuer?

Makefile
```diff
@@ -155,10 +155,19 @@ UNIT_TEST_DIRS := $(shell go list ./... | grep -v /test/)
...
 test-unit: $(SETUP_ENVTEST) #HELP Run the unit tests
 	rm -rf $(COVERAGE_UNIT_DIR) && mkdir -p $(COVERAGE_UNIT_DIR)
-	eval $$($(SETUP_ENVTEST) use -p env $(ENVTEST_VERSION) $(SETUP_ENVTEST_BIN_DIR_OVERRIDE)) && CGO_ENABLED=1 go test -count=1 -race -short $(UNIT_TEST_DIRS) -cover -coverprofile ${ROOT_DIR}/coverage/unit.out -test.gocoverdir=$(ROOT_DIR)/coverage/unit
+	eval $$($(SETUP_ENVTEST) use -p env $(ENVTEST_VERSION) $(SETUP_ENVTEST_BIN_DIR_OVERRIDE)) && \
+            CGO_ENABLED=1 go test \
+                -tags '$(GO_BUILD_TAGS)' \
+                -cover -coverprofile ${ROOT_DIR}/coverage/unit.out \
+                -count=1 -race -short \
+                $(UNIT_TEST_DIRS) \
+                -test.gocoverdir=$(ROOT_DIR)/coverage/unit
```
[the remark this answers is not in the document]
joelanford: 🤷‍♂️ 

I was mainly focused on "like flags" and "can I see it without side-to-side scrolling?"

Notes
- I wanted UNIT_TEST_DIRS as close to the end as possible because I sometimes find myself copying/pasting the command line from a full run so that I can run unit tests in a specific package (and yes, I know we could make `UNIT_TEST_DIRS` overrideable, but in other projects that do that I always forget what the var name is, so I end up copy/pasting anyway)
- I was forced to put `test.govoerdir` _after_ the UNIT_TEST_DIRS because those flags are passed to the test binary.

internal/rukpak/source/unpacker.go
joelanford: Hmm. It appears that we do not check for this error type in our reconciler. I brought this over from the now-deleted `image_registry.go` where it was used.

I do check for it in unit tests.

WDYT about addressing this in a follow-up?

internal/rukpak/source/containers_image_test.go
joelanford: Copy/pasta from the original test suite. I can remove.

internal/rukpak/source/unpacker.go
tmshort: Depends on how many other changes will occur if you update... but at least it's isolated.

Makefile
```diff
@@ -173,6 +182,7 @@ build-push-e2e-catalog: ## Build the testdata catalog used for e2e tests and pus
 test-e2e: KIND_CLUSTER_NAME := operator-controller-e2e
 test-e2e: KUSTOMIZE_BUILD_DIR := config/overlays/e2e
 test-e2e: GO_BUILD_FLAGS := -cover
+test-e2e: E2E_REGISTRY_CERT_REF := Issuer/selfsigned-issuer
```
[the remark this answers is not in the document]
joelanford: The intention here is to setup our standard e2e test with:
- an image registry the serves with an untrusted certificate
- a mounted `/etc/containers/registries.conf` file that instructs operator-controller to allow insecure connections to the image registry.

This setup proves that our image puller respect what shows up in `/etc/containers/registries.conf` (if the `registries.conf` volume mount is removed, the tests fail)

The extension developer e2e and the upgrade e2e are left using the standard olmv1-ca to:
1. Verify that op-con works without a mounted /etc/containers/registry.conf
2. Avoid issues with pre-upgrade setup which depends on using the unmodified manifest of the previous release (meaning we can't mount `/etc/containers` to allow op-con to trust an insecure registry)

hack/test/image-registry.sh
```diff
@@ -7,27 +7,40 @@ set -o pipefail
...
 metadata:
   name: ${namespace}
+---
+ apiVersion: cert-manager.io/v1
+ kind: Issuer
+ metadata:
+   name: selfsigned-issuer
+   namespace: ${namespace}
+ spec:
+   selfSigned: {}
```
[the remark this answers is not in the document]
joelanford: See https://github.com/operator-framework/operator-controller/pull/1194#discussion_r1752626728

Makefile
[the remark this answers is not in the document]
joelanford: I honestly don't love the current setup. It seems overly complex and brittle, and I do acknowledge that this PR makes it ever so slightly more complex and brittle.

I'd be in favor of looking at our e2e test code and fixtures holistically in order to do a bit of cleanup and refactoring, but IMO, in a separate PR.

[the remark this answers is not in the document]
tmshort: As long as the _intent_ was to have a separate trust chain for the registries, then ok. The current CAs are used for communication catalogd <-> operator-controller, and api-server -> catalogd (for webhooks)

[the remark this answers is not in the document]
joelanford: Yes, the intent is to explicitly have a separate (untrusted) trust chain. The other places the CA were used are:
- catalogd -> e2e image registry
- operator-controller -> e2e image registry

Prior to this PR, the same exact image registry setup was used for every e2e (e2e, upgrade-e2e, and developer-extension-e2e).

With this PR, the vanilla e2e is explicitly changed to use the untrusted self-signed cert (while the other e2e variants remain unchanged)

tmshort (approved):
/lgtm

everettraven (approved):

LalatenduMohanty (approved):
/lgtm

Issue #1148, opened 2024-08-19
Logically group fields in the spec


everettraven:
This change is unrelated to either the OpenShift or Kubernetes API Conventions and is focused on ensuring that fields are logically grouped and that these logical groupings are valid as our feature set and API is expanded post-v1.0.0.

Building on #1088 , the RFC proposed a structure similar to:
```go 
type ClusterExtensionSpec struct {
  Source SourceConfig `json:"resolution"`
  Install InstallationConfig `json:"install"`
}

type InstallationConfig struct {
  Namespace string `json:"namespace"`
  ServiceAccount ServiceAccountReference `json:"serviceAccount"`
  Preflight *PreflightConfig `json:"preflight,omitempty"`
}
```
An example of a ClusterExtension, in YAML, with these changes (and building on top of the discriminated union changes):
```yaml
apiVersion: olm.operatorframework.io/v1alpha1
kind: ClusterExtension
metadata:
  name: argocd
spec:
  resolution:
    type: Catalog
    catalog:
      packageName: argocd-operator
      version: 0.6.0
  install:
    namespace: argocd
    serviceAccount:
      name: argocd-installer
```

Additionally, the `ClusterExtension.Status` should be updated to have logically grouped status information as well:
```go
type ClusterExtensionStatus struct {
  Install *InstallStatus `json:"install,omitempty"`
  Resolution *ResolutionStatus `json:"resolution,omitempty"`

  Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type" protobuf:"bytes,1,rep,name=conditions"`
}

type InstallStatus struct {
  Bundle *BundleMetadata `json:"bundle,omitempty"`
}

type ResolutionStatus struct {
  Bundle *BundleMetadata `json:"bundle,omitempty"`
}
```

Having these distinct, logical sections in both the `ClusterExtension.Spec` and the `ClusterExtension.Status` allows for logical expansion of fields. For example, if we wanted to add the `ClusterCatalog` that the bundle we resolved to was from in the status, it would be a straightforward update to add a new field to the `ResolutionStatus` struct above.

PR #1200, opened 2024-08-30
✨ logically group fields in the spec

yashoza19:
# Description
Fixes: #1148 

## Reviewer Checklist

- [x] API Go Documentation


api/v1alpha1/clusterextension_types.go
```diff
@@ -46,27 +46,66 @@ const (
 
 // ClusterExtensionSpec defines the desired state of ClusterExtension
 type ClusterExtensionSpec struct {
-	// source is a required field which selects the installation source of content
+	// resolution is a required field which selects the installation source of content
```
everettraven, line 49: We had a discussion about this in the RFC and it looks like I had forgotten to update appropriately in the issue that this PR is linked to, but we want this to remain named `source` instead of `resolution`. This applies for all the types  associated with this as well.

My apologies for dropping the ball on updating the issue!

api/v1alpha1/clusterextension_types.go
```diff
@@ -46,27 +46,66 @@ const (
...
+
+	// catalog is used to configure how information is sourced from a catalog. This field must be defined when sourceType is set to "Catalog",
+	// and must be the only field defined for this sourceType.
+	//
+	// +optional.
+	Catalog *CatalogSource `json:"catalog,omitempty"`
+}
+
+// InstallationConfig is a discriminated union which selects the installation config.
+// +union
```
everettraven, line 101: This is not a discriminated union

everettraven (changes requested):

internal/controllers/clusterextension_controller.go
```diff
@@ -184,8 +184,8 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
 		//  This is not ideal, and we should consider a more nuanced approach that resolves
 		//  as much status as possible before returning, or at least keeps previous state if
 		//  it is properly labeled with its observed generation.
-		ext.Status.ResolvedBundle = nil
-		ext.Status.InstalledBundle = nil
+		ext.Status.Resolution.Bundle = nil
+		ext.Status.Install.Bundle = nil
```
everettraven, line 188: Anywhere this change has been made runs a risk of accessing a `nil` pointer for the `ext.Status.Resolution` and `ext.Status.Install` fields. It looks like this is why some of the tests are failing (although maybe not caused by this line explicitly).

All areas where we update to this should perform some kind of nil check, and if the parent field is nil, set it. Something like:
```suggestion
		if ext.Status.Resolution == nil {
		        ext.Status.Resolution = &ocv1alpha1.ResolutionStatus{}
		}
		ext.Status.Resolution.Bundle = nil
	        if ext.Status.Install == nil {
		        ext.Status.Install = &ocv1alpha1.InstallStatus{}
		}
		ext.Status.Install.Bundle = nil
```

Since you'd have to likely do this `nil` check on every place we set this, it might be helpful to extract it into a helper function for each status that looks something like:
```go
func setResolutionStatus(ext *ocv1alpha1.ClusterExtension, resStatus *ocv1alpha1.ResolutionStatus) { ... }

func setInstallStatus(ext *ocv1alpha1.ClusterExtension, installStatus *ocv1alpha1.InstallStatus) { ... }
```

internal/controllers/common_controller.go
```diff
@@ -87,3 +87,17 @@ func setStatusUnpacked(ext *ocv1alpha1.ClusterExtension, message string) {
 		ObservedGeneration: ext.GetGeneration(),
 	})
 }
+
+func setResolutionStatus(ext *ocv1alpha1.ClusterExtension, resStatus *ocv1alpha1.ResolutionStatus) {
+	if ext.Status.Resolution == nil {
+		ext.Status.Resolution = &ocv1alpha1.ResolutionStatus{}
+	}
+	ext.Status.Resolution.Bundle = resStatus.Bundle
```
everettraven, line 95:
```suggestion
	ext.Status.Resolution = resStatus
```

internal/controllers/common_controller.go
```diff
@@ -87,3 +87,17 @@ func setStatusUnpacked(ext *ocv1alpha1.ClusterExtension, message string) {
...
+}
+
+func setInstallStatus(ext *ocv1alpha1.ClusterExtension, installStatus *ocv1alpha1.InstallStatus) {
+	if ext.Status.Install == nil {
+		ext.Status.Install = &ocv1alpha1.InstallStatus{}
+	}
+	ext.Status.Install.Bundle = installStatus.Bundle
```
everettraven, line 102:
```suggestion
	ext.Status.Install = installStatus
```

everettraven (changes requested):

test/e2e/cluster_extension_install_test.go
```diff
@@ -319,7 +326,7 @@ func TestClusterExtensionInstallRegistryMultipleBundles(t *testing.T) {
 		assert.Equal(ct, metav1.ConditionFalse, cond.Status)
 		assert.Equal(ct, ocv1alpha1.ReasonResolutionFailed, cond.Reason)
 		assert.Contains(ct, cond.Message, "matching packages found in multiple catalogs")
-		assert.Nil(ct, clusterExtension.Status.ResolvedBundle)
+		assert.Nil(ct, clusterExtension.Status.Resolution.Bundle)
```
everettraven, line 329: This is probably where the e2e panic is
```suggestion
		assert.Nil(ct, clusterExtension.Status.Resolution)
```

everettraven (commented):
Overall it looks really good! Fix the e2e panic and I'll be happy to merge!

api/v1alpha1/clusterextension_types.go
```diff
@@ -61,12 +61,53 @@ type ClusterExtensionSpec struct {
...
+	// install is a required field used to configure the installation options
+	// for the ClusterExtension such as the installation namespace,
+	// the service account and the pre-flight check configuration.
+	//
+	// Below is a minimal example of an installation definition (in yaml):
+	// install:
+	//    namespace: example-namespace
+	//    serviceAccount:
+	//      name: example-sa
+	Install InstallationConfig `json:"install"`
```
joelanford, line 73: Let's try to be consistent with word forms between field names and type names:
```suggestion
	Install InstallConfig `json:"install"`
```

joelanford: Also, should we be prefixing these type names with `ClusterExtension`?

This package will contain structs for any future root API types and their sub-structs.

A hypothetical `Extension` API might have its own `InstallConfig` struct, but it will be different because it would not have a `namespace` field.

test/e2e/cluster_extension_install_test.go
```diff
@@ -644,7 +718,13 @@ func TestClusterExtensionInstallReResolvesWhenNewCatalog(t *testing.T) {
 		assert.Equal(ct, metav1.ConditionTrue, cond.Status)
 		assert.Equal(ct, ocv1alpha1.ReasonSuccess, cond.Reason)
 		assert.Contains(ct, cond.Message, "resolved to")
-		assert.Equal(ct, &ocv1alpha1.BundleMetadata{Name: "prometheus-operator.2.0.0", Version: "2.0.0"}, clusterExtension.Status.ResolvedBundle)
+		assert.Equal(ct,
+			&ocv1alpha1.ResolutionStatus{Bundle: &ocv1alpha1.BundleMetadata{
+				Name:    "prometheus-operator.1.2.0",
+				Version: "1.2.0",
```
joelanford, line 724: This is not a change I expect to see in a PR that is rearranging types. What's the background here?

yashoza19: In order to compare the `BundleMetadata` to the new `Resolution.Bundle` spec, I had to make these changes.

api/v1alpha1/clusterextension_types.go
yashoza19: Do you mean it should be `ClusterExtensionInstallConfig`?

everettraven: I think `ClusterExtensionInstallConfig` is reasonable

everettraven: But I think @joelanford is referring to all the new types here. So something like:
- `ClusterExtensionInstallConfig`
- `ClusterExtensionInstallStatus`
- `ClusterExtensionResolutionStatus`
- ...

test/e2e/cluster_extension_install_test.go
everettraven: To clarify, since these are in an `Eventually` block, the e2e tests were panicking due to accessing the nil `clusterExtension.Status.Resolution` object to get the `Bundle` field. Doing this compares the entire `clusterExtension.Status.Resolution` object and avoids a potential panic. Alternatively an explicit `nil` check could be done before attempting to access the `Bundle` field.

joelanford: I should have been more specific in this comment (and it's no big deal since @tmshort fixed it in #1222), but what I was trying to ask is why this changed from 2.0.0 to 1.2.0.

I'm guessing that the `Eventually` aspect of this test is what tripped us up here. It is likely the case that the value is 1.2.0 when this check executes the first time, which explains why this would pass. However the purpose of this test is to see if a new catalog showing up under the same tag will trigger a re-reconcile and resolve a new version.

By setting this to 1.2.0, we were not actually testing what this test intended. Again, no big deal and the actual code was still behaving as expected, which is how Todd caught it.

However, I wonder: can we design our tests to be more robust against this kind of thing? Use of `Eventually` is likely the primary source of this problem because it implies both of:
- the current state of the field is not what we expect (otherwise, we would just directly check it)
- the value of the field we are asserting against will change over time.

And _that_ means that the test can pass using any assertion that passes on any initial or transitory state of that field.

Issue #528, opened 2023-11-08
Bug: `Catalog` cache doesn't update when the `Catalog` changes


everettraven:
I was taking a look at #496 and while testing the `Catalog` polling logic by pushing catalog updates, I noticed the `Catalog` would update but the resolution on the `Operator` would still fail. I found that the cached `Catalog` contents are likely not being updated due to https://github.com/operator-framework/operator-controller/blob/b5c6bc958d43c900f8cd6e6417e0c46b36effbe0/internal/catalogmetadata/cache/cache.go#L86-L91

In catalogd v0.8.0 the meaning of the `Status.ResolvedSource.Image.Ref` field changed and a new `Status.ResolvedSource.Image.ResolvedRef` field was added (see https://github.com/operator-framework/catalogd/pull/185/files#diff-e36279cce9b48e5f464ffa40cb649dc902d9a850d128b39796d351610ad1ffffR114-R123)

The fix for this bug should be a pretty straightforward change to use the `Status.ResolvedSource.Image.ResolvedRef` in place of `Status.ResolvedSource.Image.Ref` as the check above is always true if a `Catalog` has entered the cache and the image is unchanged so it misses when the resolved reference's SHA256 changes after polling has performed an update.

PR #1202, opened 2024-09-02
🐛 Fix cache check

m1kola:
# Description

There is a bug in the optimisation where we try to avoid writing, if another thread already updated the cache: I believe we should be checking `catalog.Status.ResolvedSource.Image.ResolvedRef` (includes sha) as in other places instead of `catalog.Status.ResolvedSource.Image.Ref`.

Related:
* #529
* #528

## Reviewer Checklist

- [ ] API Go Documentation


m1kola:
Accidentally deleted the branch. Restored and reopened.

everettraven (approved):

PR #1207, opened 2024-09-03
✨ Remove cache when catalog is deleted

### LINKED ISSUE #948: Bug: cached catalog contents are not cleaned up when catalog is deleted.

m1kola:
# Description

We add a finaliser to `ClusterCatalog` which enables us to remove catalog cache from filesystem on catalog deletion.

Fixes #948

## Reviewer Checklist

- [ ] API Go Documentation


cmd/manager/main.go
```diff
@@ -200,13 +200,11 @@ func main() {
 	}
 
 	clusterExtensionFinalizers := crfinalizer.NewFinalizers()
-	domain := ocv1alpha1.GroupVersion.Group
-	cleanupUnpackCacheKey := fmt.Sprintf("%s/cleanup-unpack-cache", domain)
```
m1kola, line 204: Note to self: should probably move this refactoring into a separate PR.

joelanford:
I like the direction of this. Maybe not something to do in this PR, but once we have a separate catalog controller, I think it would make sense to have _that_ controller be the thing that also populates `operator-controller`'s cache.

Right now, the operator-controller catalog cache is populated on-demand when a ClusterExtension is reconciled. The outcome is that some ClusterExtension reconciles take quite a bit longer because they have to pause to pull and extract catalog content from catalogd.

If we moved pull/extract logic to a separate `ClusterCatalog` controller, then operator-controller could do this work async. At that point the only ClusterExtensions that would still have to wait are the ones that are reconciled just after a ClusterCatalog has new data available.

m1kola:
Trying to understand why tests are failing. https://github.com/operator-framework/catalogd/issues/389 and the fact that pull interval is set to 1s in e2e makes it hard to debug.

internal/catalogmetadata/cache/cache.go
```diff
@@ -166,3 +166,25 @@ func (fsc *filesystemCache) FetchCatalogContents(ctx context.Context, catalog *c
...
+func (fsc *filesystemCache) Remove(catalogName string) error {
+	cacheDir := fsc.cacheDir(catalogName)
+
+	fsc.mutex.Lock()
+	defer fsc.mutex.Unlock()
+
+	if _, exists := fsc.cacheDataByCatalogName[catalogName]; !exists {
+		return nil
+	}
+	delete(fsc.cacheDataByCatalogName, catalogName)
```
m1kola, line 180: I missed this line. This should resolve E2Es failures.

cmd/manager/main.go
```diff
@@ -277,6 +278,23 @@ func main() {
 		setupLog.Error(err, "unable to create controller", "controller", "ClusterExtension")
 		os.Exit(1)
 	}
+
+	clusterCatalogFinalizers := crfinalizer.NewFinalizers()
+	if err := clusterCatalogFinalizers.Register(controllers.ClusterCatalogCacheDeletionFinalizer, finalizerFunc(func(ctx context.Context, obj client.Object) (crfinalizer.Result, error) {
+		return crfinalizer.Result{}, cacheFetcher.Remove(obj.GetName())
+	})); err != nil {
+		setupLog.Error(err, "unable to register finalizer", "finalizerKey", controllers.ClusterCatalogCacheDeletionFinalizer)
```
rashmi43, line 286: this includes the catalog name?

joelanford: Good call out. I don't think it does. We should use `log.FromContext` to get a logger from the caller's context, which in this case will be the Reconciler context that (I think) includes the catalog name.

cmd/manager/main.go
```diff
@@ -277,6 +278,23 @@ func main() {
...
+		os.Exit(1)
+	}
+
+	if err = (&controllers.ClusterCatalogReconciler{
+		Client:     cl,
+		Finalizers: clusterCatalogFinalizers,
```
joelanford, line 292: I worry a little bit about using finalizers here. We'll end up with two different controllers that always race when catalogs are created and destroyed, which I'm guessing will add unnecessary reconcile churn.

WDYT about a best-effort garbage collector that runs the cleanup logic along the lines of:
```
if err := r.Get(req); errors.IsNotFound(err) {
    if err := cache.Cleanup(req); err != nil {
          metrics.IncrementCatalogCacheCleanupFailure()
          eventrecorder.Event("failed to clear catalog cache for %q", req)
    }
}
```

Another option is a garbage collector somewhat like catalogd has here: https://github.com/operator-framework/catalogd/blob/main/internal/garbagecollection/garbage_collector.go

cmd/manager/main.go
rashmi43: okay

cmd/manager/main.go
everettraven: Can we still go with a "controller" approach that just doesn't modify the `ClusterCatalog` resource?

If the pod crashes, is our cache still populated? If so, we could do an initial list + cleanup on startup and then rely on informers to react to events that trigger our cleanup logic without modification of `ClusterCatalog` resources. Once the informer is up I am pretty confident it won't miss any events. Any events we miss are likely because the operator-controller manager container crashed and is going to restart and do the initial list+cleanup.

Does this sound like it could be a good middle ground?

m1kola: @everettraven I think this is the same as the best-effort option Joe suggested.

Anyway - I updated the controller to not change `ClusterCatalog`. Please take another look.


> If the pod crashes, is our cache still populated? If so, we could do an initial list + cleanup on startup and then rely on informers to react to events that trigger our cleanup logic without modification of `ClusterCatalog` resources. Once the informer is up I am pretty confident it won't miss any events. Any events we miss are likely because the operator-controller manager container crashed and is going to restart and do the initial list+cleanup.

On pod crash we will end up with filesystem containing cache since we are currently using `emptyDir` which persists between crashes. But [we will be rewriting](https://github.com/operator-framework/operator-controller/blob/df0e8481f09e6beac92d05c9ff62e31675d383b6/internal/catalogmetadata/cache/cache.go#L156-L158) the cache for each catalog since [this map](https://github.com/operator-framework/operator-controller/blob/df0e8481f09e6beac92d05c9ff62e31675d383b6/internal/catalogmetadata/cache/cache.go#L54) is not populated on restart. So it looks like using `emptyDir` is pointless here.

But my point was not about crashes: I was talking about issues when the manager is still running, but can't connect to API server. In this case it will still think that the cache is still present and valid because it hasn't seen delete event.

Perhaps getting rid of `emptyDir` will help?

internal/controllers/clustercatalog_controller.go
```diff
@@ -0,0 +1,77 @@
...
+	existingCatalog := &catalogd.ClusterCatalog{}
+	err := r.Client.Get(ctx, req.NamespacedName, existingCatalog)
+	if client.IgnoreNotFound(err) != nil {
+		return ctrl.Result{}, err
+	}
+	if apierrors.IsNotFound(err) {
+		return ctrl.Result{}, r.Cache.Remove(req.Name)
+	}
+
+	return ctrl.Result{}, nil
```
joelanford, line 54: Nit: it took me an extra second to parse the `IgnoreNotFound` + `IsNotFound` checks. We could simplify to the following.
```suggestion
	if apierrors.IsNotFound(err) {
		return ctrl.Result{}, r.Cache.Remove(req.Name)
	}
	return ctrl.Result{}, err
```

If you aren't a fan of `return ctrl.Result{}, err` at the end serving the dual purpose of returning nil or non-nil errors, we could also refactor like this:
```suggestion
	if apierrors.IsNotFound(err) {
		return ctrl.Result{}, r.Cache.Remove(req.Name)
	}
	if err != nil {
		return ctrl.Result{}, err
	}
	return ctrl.Result{}, nil
```

internal/controllers/clustercatalog_controller.go
```diff
@@ -0,0 +1,77 @@
...
+//+kubebuilder:rbac:groups=olm.operatorframework.io,resources=clustercatalogs,verbs=get;list;watch
+
+func (r *ClusterCatalogReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	existingCatalog := &catalogd.ClusterCatalog{}
+	err := r.Client.Get(ctx, req.NamespacedName, existingCatalog)
+	if client.IgnoreNotFound(err) != nil {
+		return ctrl.Result{}, err
+	}
+	if apierrors.IsNotFound(err) {
+		return ctrl.Result{}, r.Cache.Remove(req.Name)
```
joelanford, line 51: Should we create one or more new metrics related to this cache? For example:

- gauge: number of known ClusterCatalogs
- gauge: number of local caches

Or maybe just store the diff as a single gauge (that we would expect to always be 0)?

internal/controllers/clustercatalog_controller.go
everettraven: Do we need the `Get()` call? In our `SetupWithManager()` method below we filter events to only care about the delete events, so reconcile should only ever be called when a delete event is triggered for a `ClusterCatalog` resource.

internal/controllers/clustercatalog_controller.go
everettraven: I don't think this is unreasonable, but I would probably consider this outside the scope of this specific PR. I _do_ think we should eventually have a discussion about instrumenting OLMv1 with meaningful metrics.

joelanford: Fair!

internal/controllers/clustercatalog_controller.go
joelanford: Yeah, true. We could make it that simple for now.

I'm guessing we will follow-up at some point to do what I mentioned here: https://github.com/operator-framework/operator-controller/pull/1207#issuecomment-2326975817

But the code is so simple as is that it would be easy to refactor later without accidentally introducing bugs or causing a regression of this code.

m1kola: I went with the second suggestion for error handling for the same reason: will be adding more into this controller soon.

internal/controllers/clustercatalog_controller.go
perdasilva: I've created #1287 to keep track of this. Are there any labels we'd want to apply to the issue?

everettraven (approved):

Issue #1151, opened 2024-08-19
Update the allowed values of the `ClusterExtension.Spec.UpgradeConstraintPolicy` to be more descriptive


everettraven:
For the `ClusterExtension.Spec.UpgradeConstraintPolicy` field, the “Enforce” and “Ignore” allowed values are not very descriptive of their behavior. In an effort to be more descriptive it is proposed that the following changes are made to the allowed value names:
- `Enforce` → `CatalogProvided`
- `Ignore` → `SelfCertified`

The GoDoc comment should be updated to include explicit information as to what each allowed value means when specified.

PR #1215, opened 2024-09-04
📖 [Docs] Add support for skip ranges and replacements

### LINKED ISSUE #744: Docs: Implement full support for replaces, skips, and skipRange (and make default)
# User story

As an admin, when I have an installed ClusterExtension, I expect that the upgrade edges defined in the catalog (i.e. replaces, skips, and skipRange) will be honored.

## List the user-facing issues in the epic that require documentation.

- Issue: #738 (single issue epic)

## Does this epic require new content or updates to existing content?

- [ ] Net-new content
- [x] Update existing content

My assumption is that there are already docs that talk about how OLMv1 uses semver semantics when assessing upgrade successors in a channel. If that's true, those docs need to be updated to talk about replaces, skips, and skipRange.

Even with this new support for replaces, skips, and skipRange, OLMv1 logic is subtly different:
- In OLMv0 if there are multiple possible successors, the one closest to the channel head is chosen
- In OLMv1 if there are multiple possible successors, the one with the highest semver version is chosen.

## What type of documentation is required for this epic?

- [ ] Release note
- [ ] Concept
- [ ] Procedure
- [ ] Reference
- [x] Not sure

## Is there prior art or reference materials that would help write the documentation?

https://olm.operatorframework.io/docs/concepts/olm-architecture/operator-catalog/creating-an-update-graph/

Documentation should be contributed to the [`docs/drafts/` directory](https://github.com/operator-framework/operator-controller/tree/main/docs/drafts)

michaelryanpeter:
Closes #744

@perdasilva: I am not sure if you are using this file as part of your work. I just wanted to make sure we aren't stepping on each other's toes :)

# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/docs/drafts/upgrade-support.md
+++ b/docs/drafts/upgrade-support.md
@@ -1,33 +1,58 @@
 # Upgrade support
 
-This document explains how OLM 1.0 handles upgrades.
+This document explains how OLM v1 handles upgrades.
 
-OLM 1.0 introduces a simplified UX for package authors and package admins to implicitly define upgrade edges via [Semantic Versioning](https://semver.org/).
+OLM v1 introduces a simplified UX for package authors and package admins to implicitly define upgrade edges via [Semantic Versioning](https://semver.org/).
 
 It also introduces an API to enable independently verified upgrades and downgrades.
 
 ## Upgrade constraint semantics
 
-As of operator-controller release 0.10.0, OLM 1.0 supports the following upgrade constraint semantics:
+When determining upgrade edges, also known as upgrade paths or upgrade constraints, for an installed cluster extension, Operator Lifecycle Manager (OLM) v1 supports [legacy OLM semantics](https://olm.operatorframework.io/docs/concepts/olm-architecture/operator-catalog/creating-an-update-graph/) by default. This support follows the behavior from legacy OLM, including `replaces`, `skips`, and `skipRange` directives, with a few noted differences.
 
-* [Semantic Versioning](https://semver.org/) (Semver)
-* [Legacy OLM 0 semantics](https://olm.operatorframework.io/docs/concepts/olm-architecture/operator-catalog/creating-an-update-graph/#methods-for-specifying-updates): the `replaces`/`skips`/`skipRange` directives
+By supporting legacy OLM semantics, OLM v1 now honors the upgrade graph from catalogs accurately.
 
-The Kubernetes manifests in this repo enable legacy support by default. Cluster admins can control which semantics to use by passing one of the following arguments to the `manager` binary:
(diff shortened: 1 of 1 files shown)


docs/drafts/upgrade-support.md
```diff
@@ -1,33 +1,58 @@
...
+kind: ClusterExtension
+metadata:
+  name: <extension_name>
+spec:
+  installNamespace: <namespace>
+  packageName: <package_name>
+  serviceAccount:
+    name: <service_account>
+  upgradeConstraintPolicy: Ignore
+  version: "<version_or_version_range>"
```
joelanford, line 46: Non-blocking (I think we'll need to sweep the entire repo once all the API changes are finished): All these fields are getting moved around: 
- This example is not currently valid.
- Even if you made it valid according to what is in `main` right now, there are other open PRs that will invalidate it.

docs/drafts/upgrade-support.md
```diff
@@ -1,33 +1,58 @@
...
 ```
 
-In a future release, it is planned to remove the `ForceSemverUpgradeConstraints` feature gate and allow package authors to specify upgrade constraint semantics at the catalog level.
+where setting the `upgradeConstraintPolicy` to:
+
+`Ignore`
+:   Does not limit the next version to the set of successors, and instead allows for any downgrade, sidegrade, or upgrade.
+
+`Enforce`
```
joelanford, line 54: Another upcoming change: https://github.com/operator-framework/operator-controller/issues/1151

docs/drafts/upgrade-support.md
michaelryanpeter: With Material MK Docs, we can actually embed YAML files into a markdown page.

Maybe we can create an issue right before or after 1.0.0 GA to switch our inline examples to embedded YAML files? That way we can update our examples all in one go.

michaelryanpeter: It looks like #1200 has the relevant doc changes queued up in that PR.

docs/drafts/upgrade-support.md
michaelryanpeter: @joelanford Is this something we would want to change to the new value now? Or do we want to update #1151 with a comment or a docs issue?

joelanford: Use what you have, which is what is on `main`. The PR for that change should hopefully find/replace these.

docs/drafts/upgrade-support.md
joelanford: #1200 merged, which makes this section incorrect if it merges. I'm happy to merge this and fix in a follow up (hence my approval on the PR), but I'll defer to you if you want to go ahead and fix now.

joelanford (approved):

PR #1216, opened 2024-09-05
:seedling: Add catalog exploration and extension rbac/installation scripts

perdasilva:
# Description

The lack of tooling can make finding and installing extensions hard. This PR adds some `hack/` scripts to help smooth over this rough edge until the proper tools are ready

## Reviewer Checklist

- [ ] API Go Documentation


everettraven:
Is it worth adding some of these scripts and having to maintain them until we have proper tooling in place? Is there things that we've learned from writing these scripts that we can write up documentation on how to manually do all these things until the proper tooling is in place?

perdasilva:
> Is it worth adding some of these scripts and having to maintain them until we have proper tooling in place? Is there things that we've learned from writing these scripts that we can write up documentation on how to manually do all these things until the proper tooling is in place?

If you tell me we'll have the tools next week, then agree. Otherwise, in my opinion, yes. I don't think I'd use a package manager that requires me to read/understand documentation and go through a whole bunch of manual including manually finding/pulling/unpacking the bundle, and inspecting its content. I think it makes sense to show the minimum amount of effort towards tooling to help facilitate user's lives. This script is in hack, the documentation calls it out as a not-guaranteed-to-work helper script to smooth things out until we have proper tools. How much maintenance do you foresee happening here? Will it be that much of a time sync? Is it _really_ not worth it?

hack/tools/catalogs/download-catalog
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/env bash
+
+SCRIPT_ROOT=$(dirname "$(realpath "$0")")
+source "${SCRIPT_ROOT}/lib/utils.sh"
+
+# Constants
+SERVICE_NAMESPACE="olmv1-system"
+SERVICE_NAME="catalogd-catalogserver"
+LOCAL_PORT=8001
```
everettraven, line 9: Can a user override this value easily? I can't remember off the top of my head if `=` sets a variable if the environment variable doesn't already exist in bash

hack/tools/catalogs/download-catalog
```diff
@@ -0,0 +1,72 @@
...
+if [ -z "$1" ]; then
+    usage
+    exit 1
+fi
+
+CATALOG_NAME="$1"
+
+# Check if the clustercatalog resource exists
+echo "Checking if ClusterCatalog $CATALOG_NAME exists..."
+CLUSTER_CATALOG=$(kubectl get clustercatalog "$CATALOG_NAME" -o json 2>/dev/null)
```
everettraven, line 35: Should we have a check/documentation that lets a user know that `kubectl` is required?

hack/tools/catalogs/download-catalog
everettraven, line 1: Put the contents of this file through shellcheck and got the following back:
```
Line 4:
source "${SCRIPT_ROOT}/lib/utils.sh"
       ^-- SC1091 (info): Not following: ./lib/utils.sh was not specified as input (see shellcheck -x).
 
Line 62:
LOCAL_CONTENT_URL=$(echo "$CONTENT_URL" | sed "s|https://[^/]*|https://localhost:$LOCAL_PORT|")
                    ^-- SC2001 (style): See if you can use ${variable//search/replace} instead.
```

hack/tools/catalogs/download-catalog
```diff
@@ -0,0 +1,72 @@
...
+if [ -z "$CONTENT_URL" ]; then
+    echo "Content URL not found for ClusterCatalog $CATALOG_NAME."
+    exit 1
+fi
+
+# Start port forwarding
+echo "Starting kubectl port-forward to $SERVICE_NAME on port $LOCAL_PORT..."
+kubectl port-forward -n "$SERVICE_NAMESPACE" svc/"$SERVICE_NAME" $LOCAL_PORT:$SERVICE_PORT &>/dev/null &
+PORT_FORWARD_PID=$!
+sleep 2  # Wait for the port-forwarding to start
```
everettraven, line 59: Is there a way to do this without using `sleep`? Could we try to `ping` the local port and exit on the first response that signals the local port is "live"

hack/tools/catalogs/generate-manifests
```diff
@@ -0,0 +1,178 @@
...
+    echo ""
+    echo "Usage:"
+    echo ""
+    echo "Generate installation manifests"
+    echo "$0 install <package-name> <package-version> [-n namespace] [-e cluster-extension-name] [-cr cluster-role-name] [-r role-name] [-s service-account-name] [--use-defaults]"
+    echo ""
+    echo "Generate RBAC manifests"
+    echo "$0 rbac <package-name> <package-version> [-n namespace] [-e cluster-extension-name] [-cr cluster-role-name] [-r role-name] [-s service-account-name] [--use-defaults]"
+    echo ""
+    echo "Generate installation or RBAC manifests for a registry+v1 package given an FBA catalog in stdin"
```
everettraven, line 27:
```suggestion
    echo "Generate installation or RBAC manifests for a registry+v1 package given an FBC catalog in stdin"
```

hack/tools/catalogs/generate-manifests
```diff
@@ -0,0 +1,178 @@
...
+fi
+
+# Command and package details
+COMMAND=$1
+export PACKAGE_NAME=$2
+export PACKAGE_VERSION=$3
+
+# Initialize environment variables with template defaults
+export NAMESPACE="\${NAMESPACE}"
+export EXTENSION_NAME="\${EXTENSION-NAME}"
```
everettraven, line 82: Based on the usage it looks like this should be:
```suggestion
export EXTENSION_NAME="\${EXTENSION_NAME}"
```

hack/tools/catalogs/is-bundle-supported
```diff
@@ -0,0 +1,39 @@
...
+SCRIPT_ROOT=$(dirname "$(realpath "$0")")
+
+source "${SCRIPT_ROOT}/lib/unpack.sh"
+source "${SCRIPT_ROOT}/lib/collect-rbac.sh"
+source "${SCRIPT_ROOT}/lib/utils.sh"
+
+usage() {
+    print-banner
+    echo ""
+    echo "Usage: $0 <manifest_dirn>"
```
everettraven, line 13:
```suggestion
    echo "Usage: $0 <manifest_dir>"
```

hack/tools/catalogs/lib/bundle.sh
```diff
@@ -0,0 +1,191 @@
+# Library of functions for interacting with bundles and their manifests
+
+# SCRIPT_ROOT is the root of the script
+source "$(dirname "${BASH_SOURCE[0]}")/hash.sh"
+
+# Given package and version grabs the bundle image from stdin FBC stream
+function get-bundle-image(){
```
everettraven, line 7: Should we document anywhere that this is used, or maybe as part of this utility function first do a check for the presence, that `jq` is required to be installed?

hack/tools/catalogs/lib/bundle.sh
```diff
@@ -0,0 +1,191 @@
...
+        exit 1
+    fi
+
+    echo "${image}"
+}
+
+function is-all-namespace-mode-enabled() {
+    local csv="${1}"
+    local valid
+    valid=$(yq eval '(.spec.installModes[] | select(.type == "AllNamespaces" and .supported == true)) and true' "${csv}")
```
everettraven, line 25: Document that `yq` is required where this is used?

everettraven: Out of curiosity, why does this need the additional `and true` at the end?

hack/tools/catalogs/list-compatible-bundles
```diff
@@ -0,0 +1,93 @@
...
+function regex_it() {
+    jq --arg regex "$REGEX" '
+        if $regex != "" then
+            map(select(.packageName | test($regex)))
+        else
+            .
+        end'
+}
+
+cat - | looking-for-a-bundle | in-finance | with-a-trust-fund | six-five | blue-eyes
```
everettraven, line 90: lol

tmshort: I thought I read something in _Clean Code_ against this type of stuff...? @everettraven ?

hack/tools/catalogs/download-catalog
perdasilva: I was originally thinking of them as constants. If the values are already set, they will be overriden in the sub-shell the script gets executed in. It shouldn't affect the users environment. However, it could be useful to be able to override them. So, I've made the necessary changes to use the environment value (or default) and also made the variable names a bit less generic.

hack/tools/catalogs/download-catalog
perdasilva: I have added a README to the directory and also assertions at the beginning of each script.

hack/tools/catalogs/download-catalog
perdasilva: I think the first issue seems to be related to shellcheck not being able to find the file or it wasn't included in the context when invoking it....?

I've updated the second to:

```
LOCAL_CONTENT_URL=${CONTENT_URL//https:\/\/$CATALOGD_SERVICE_NAME.$CATALOGD_CATALOGD_SERVICE_NAMESPACE.svc/https:\/\/localhost:$LOCAL_PORT}
```

wdyt?

perdasilva: I'm not sure what the downsides of sed are. Here we have to assume the url.

hack/tools/catalogs/download-catalog
perdasilva: Made it so!

hack/tools/catalogs/lib/bundle.sh
perdasilva: I put it there to force yq to evaluate it to "true" (or false) rather than returning the matching snippet (or nothing). E.g. without it, it would look something like this:

```
match=$(yq eval '(.spec.installModes[] | select(.type == "AllNamespaces" and .supported == true))' "${csv}")
    if [ -n "${match}" ]; then
        echo "true"
    else
        echo "false"
    fi
```

I'll add a comment explaining it

hack/tools/catalogs/list-compatible-bundles
perdasilva: Clean code schlean code - real developers leave easter eggs XD

perdasilva: I've updated it to something boring

hack/tools/catalogs/lib/bundle.sh
```diff
@@ -0,0 +1,194 @@
+# Library of functions for interacting with bundles and their manifests
+
+# SCRIPT_ROOT is the root of the script
+source "$(dirname "${BASH_SOURCE[0]}")/hash.sh"
+
+# Given package and version grabs the bundle image from stdin FBC stream
+function get-bundle-image(){
```
gavinmbell, line 7: nit: You don't need to put both the "function" keyword and "()".  The latter is in good form, the former is outdated.

perdasilva: Fixed up ^^

perdasilva:
@everettraven I think this is ready for merge ^^

everettraven (approved):

Issue #1115, opened 2024-08-12
E2E tests for labeling webhooks and selector


tmshort:
There is a composed system to make the final correct response happen:  one, or possibly two, webhooks -- a mutating webhook and possibly a validating webhook to enforce policy; operator-controller, catalogd. Because of this composed system, we should have add at least one e2e that confirms the happy path outcome works.

tmshort:
This should be done in both catalogd and operator-controller.

Looks extremely straightforward to augment the existing e2e here: https://github.com/operator-framework/catalogd/blob/db28a52cd57400cbdac88f42e00bc2d4eabcff84/test/e2e/unpack_test.go#L64

Right after create succeeds, we can check that catalog var for the right label/value

PR #1222, opened 2024-09-05
:seedling: Update catalogd e2e and update e2e

tmshort:
Fixes: #1115 

# Description

## Reviewer Checklist

- [ ] API Go Documentation


tmshort:
The upgrade-e2e failure is likely, due to the addition of a webhook

test/e2e/cluster_extension_install_test.go
```diff
@@ -1021,3 +1021,30 @@ func getArtifactsOutput(t *testing.T) {
 		}
 	}
 }
+
+func TestClusterCatalogLabels(t *testing.T) {
```
joelanford, line 1025: Do we need this separate test in `operator-controller`? It seems like
1. it is functionality that is fully encompassed (and now tested) in catalogd.
2. we are already testing this indirectly via operator-controller's support of catalog selectors in its ClusterExtension API.

tmshort: If catalogd's behavior changes, then this would detect it, but so would other tests. I can live without it

joelanford: > but so would other tests

Right, our other tests depend on that functionality working now.

everettraven: +1 on not testing catalogd's behavior explicitly in operator-controller

tmshort:
It would seem that using `GenerateName` (as used here: https://github.com/operator-framework/operator-controller/blob/c4470cc02d4871975d25747fee6fdd9467a3af6d/test/extension-developer-e2e/extension_developer_test.go#L44) won't work with ClusterCatalogs...

perdasilva (approved):

everettraven:
> It would seem that using `GenerateName`, as used here:
> 
> https://github.com/operator-framework/operator-controller/blob/c4470cc02d4871975d25747fee6fdd9467a3af6d/test/extension-developer-e2e/extension_developer_test.go#L44
> 
> won't work with ClusterCatalogs...

Why is that?

tmshort:
> Why is that?

Because the `name` field doesn't exist, the webhook `matchConditions` errors out. This has been fixed in my latest merged PR to catalogd: https://github.com/operator-framework/catalogd/pull/382

PR #1223, opened 2024-09-06
⚠️ types conversions from review

grokspawn:
Fixes https://github.com/operator-framework/operator-controller/issues/1146
Replaces #1171 

# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/api/v1alpha1/clusterextension_types.go
+++ b/api/v1alpha1/clusterextension_types.go
@@ -378,7 +378,7 @@ type ServiceAccountReference struct {
 }
 
 // PreflightConfig holds the configuration for the preflight checks.  If used, at least one preflight check must be non-nil.
-// +kubebuilder:validation:XValidation:rule="has(self.crdUpgradeSafety)",message="crdUpgradeSafety is required when preflight is specified"
+// +kubebuilder:validation:XValidation:rule="has(self.crdUpgradeSafety)",message="at least one of [crdUpgradeSafety] are required when preflight is specified"
 type PreflightConfig struct {
 	// crdUpgradeSafety is used to configure the CRD Upgrade Safety pre-flight
 	// checks that run prior to upgrades of installed content.
--- a/api/v1alpha1/clusterextension_types.go
+++ b/api/v1alpha1/clusterextension_types.go
@@ -377,7 +377,8 @@ type ServiceAccountReference struct {
 	Name string `json:"name"`
 }
 
-// PreflightConfig holds the configuration for the preflight checks.
+// PreflightConfig holds the configuration for the preflight checks.  If used, at least one preflight check must be non-nil.
+// +kubebuilder:validation:XValidation:rule="has(self.crdUpgradeSafety)",message="crdUpgradeSafety is required when preflight is specified"
 type PreflightConfig struct {
 	// crdUpgradeSafety is used to configure the CRD Upgrade Safety pre-flight
 	// checks that run prior to upgrades of installed content.
--- a/config/base/crd/bases/olm.operatorframework.io_clusterextensions.yaml
+++ b/config/base/crd/bases/olm.operatorframework.io_clusterextensions.yaml
@@ -128,6 +128,9 @@ spec:
                     required:
                     - crdUpgradeSafety
                     type: object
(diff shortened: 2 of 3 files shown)


api/v1alpha1/clusterextension_types.go
```diff
@@ -377,7 +377,8 @@ type ServiceAccountReference struct {
 	Name string `json:"name"`
 }
 
-// PreflightConfig holds the configuration for the preflight checks.
+// PreflightConfig holds the configuration for the preflight checks.  If used, at least one preflight check must be non-nil.
+// +kubebuilder:validation:XValidation:rule="has(self.crdUpgradeSafety)",message="crdUpgradeSafety is required when preflight is specified"
```
grokspawn, line 381: This CEL is a result of the conversation [here](https://github.com/operator-framework/operator-controller/pull/1171#discussion_r1731360575). 
In order for this to function, the instance of the `PreflightConfig` in `ClusterExtensionInstallConfig` had to remain an [optional/pointer](https://github.com/operator-framework/operator-controller/blob/c4470cc02d4871975d25747fee6fdd9467a3af6d/api/v1alpha1/clusterextension_types.go#L153).

everettraven: I recently ran into an issue when playing around with CEL in a CRD that came back saying that I had exceeded the CEL cost budget. I haven't been able to find if this is a cost budget for the entire CRD or an individual CEL expression but realizing that there even is one made me think that we are better off using the `anyOf` field in the CRD to enforce the "at least one of" clause.

Maybe @joelanford has some thoughts on this?

everettraven: Unfortunately there is no controller-gen support for automatically creating the `anyOf`, `oneOf` directives but maybe that is something we could try to contribute in the future?

joelanford: Oh yeah, if this is an area that we need "at least one" semantics and that's what `anyOf` does, let's do that instead.

joelanford: Example in rukpak that may help (it uses `oneOf`, not `anyOf`, but I assume it is otherwise an example of what we want to do.

 https://github.com/operator-framework/rukpak/blob/5ffcfff615566f3a1a482a5f6649cd3e07f2427f/manifests/base/apis/crds/patches/bundledeployment_validation.yaml#L9-L20

grokspawn: Aside from preferential semantics, what's the difference between `has(self.crdUpgradeSafety) || has(self.NewType || ...` and `anyOf`?  It feels like the only real difference is manual vs controller-gen maintenance.

everettraven: I think the biggest difference is there is no "cost budget" associated with the `anyOf`. That being said, I just played around with chaining `has(...)` OR statements in https://playcel.undistro.io/ and it doesn't look like there is a cost budget increase for the chaining and it stays statically at a cost of `1`.

My original concern was that as the subset of fields grew so would our cost of evaluation, but that seems to not be the case.

With that in mind, I'm OK with CEL going in if you feel that we should use CEL here @grokspawn

grokspawn: The [runtime-cost-budget](https://kubernetes.io/docs/reference/using-api/cel/#runtime-cost-budget) suggest that the intent is to avoid runaway evaluation costs.  
The [resource use by validation functions](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#resource-use-by-validation-functions) says specifically
> You are unlikely to encounter issues with the resource budget for validation if you only specify rules that always take the same amount of time regardless of how large their input is.

I don't believe that the anyOf approach `has(self.property) [|| has(self.property2)... || has(self.propertyN)]` suffers from a non-linear cost (internally, this should refer to a limited sequence of nil pointer checks with early escape), so I don't think that this rises to the level of something that we'd want to maintain via kustomize patch instead (with accompanying `maxItems`, `maxProperties`, `maxLength` presumably).

everettraven (approved):

PR #1230, opened 2024-09-06
📖 [Docs] API Reference Doc Generation

### LINKED ISSUE #1126: Docs: Operator-Controller API reference
Write OperatorController API reference documentation. Sync with @michaelryanpeter to see if there are any tools that can go from godoc to the API reference. Another option might to source the information from the CRD which is in openAPI format. There might already be tools to process that. (See what k8s does here: https://kubernetes.io/docs/contribute/generate-ref-docs/kubernetes-api/)

- [ ] ClusterExtension API reference documentation

Current documentation is posted up at https://operator-framework.github.io/operator-controller/
New docs should be placed in [docs/drafts](https://github.com/operator-framework/operator-controller/tree/main/docs/drafts)

> [!NOTE]
> This documentation should be written _after_ #740

dtfranz:
Adds API reference doc generation via github.com/elastic/crd-ref-docs, and adds make target + dependency to the Makefile so we can ensure the document is always up-to-date.

Closes #1126 

# Description

## Reviewer Checklist

- [ ] API Go Documentation


Makefile
```diff
@@ -299,6 +299,16 @@ quickstart: $(KUSTOMIZE) manifests #EXHELP Generate the installation release man
 
 ##@ Docs
 
+.PHONY: crd-ref-docs
+API_REFERENCE_FILENAME := operator-controller-api-reference.md
```
dtfranz, line 303: Any strong thoughts on the filename? I prepended with `operator-controller` on the off-chance we put other APIs in here.

Makefile
```diff
@@ -299,6 +299,16 @@ quickstart: $(KUSTOMIZE) manifests #EXHELP Generate the installation release man
...
+crd-ref-docs: $(CRD_REF_DOCS)
+	rm -f $(ROOT_DIR)/docs/drafts/$(API_REFERENCE_FILENAME)
+	$(CRD_REF_DOCS) --source-path=$(ROOT_DIR)/api \
```
dtfranz, line 306: There are a wealth of options and configuration we can use we we generate these; if there are any that seem interesting or necessary [here](https://github.com/elastic/crd-ref-docs?tab=readme-ov-file#usage) please let me know! Otherwise for now I'm just dumping the entire API reference into one `.md` file. We could for instance build a config file to ignore the `ClusterExtensionList` type if we wanted to.

dtfranz:
`sanity` workflow is currently failing because it's generating a different file than local. I'm investigating...

Makefile
LalatenduMohanty: `operator-controller-api-reference.md` sounds good to me.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/operator-framework/operator-controller
 
-go 1.22.5
+go 1.22.7
```
dtfranz, line 3: Version bump required by the `crd-ref-docs` latest version

docs/drafts/operator-controller-api-reference.md
michaelryanpeter, line 1: Because this content is autogenerated and verified, I think that this could go right into the `docs/refs/api` directory.

We would need to update the topic map here: https://github.com/operator-framework/operator-controller/blob/main/mkdocs.yml

Maybe adding a subfolder/toggle in the navigation for API reference docs around[ line 19/20](https://github.com/operator-framework/operator-controller/blob/1df52c71e52bc6b46c9ecbb1462254c9a89b61ca/mkdocs.yml#L19)

```
  - References:
    - API references:
      - Operator controller API reference: 'refs/api/operator-controller-api-reference.md'
    - Catalog queries: 'refs/catalog-queries.md'
    - CRD Upgrade Safety: 'refs/crd-upgrade-safety.md'

Comment (dtfranz) [reply docs/drafts/operator-controller-api-reference.md:1]: Done!

**Comment by michaelryanpeter** [comment]
Other than the nits I made in Slack, my only other comments are refinements to try in the future. 
LGTM.

File: docs/refs/api/operator-controller-api-reference.md [hunk 4]
```diff
@@ -0,0 +1,281 @@
+# API Reference
```
michaelryanpeter, line 1: Could we change this heading easily?

```suggestion
# Operator Controller API Reference
```

Makefile
```diff
@@ -299,6 +299,16 @@ quickstart: $(KUSTOMIZE) manifests #EXHELP Generate the installation release man
...
 ##@ Docs
 
+.PHONY: crd-ref-docs
+API_REFERENCE_FILENAME := operator-controller-api-reference.md
+crd-ref-docs: $(CRD_REF_DOCS)
+	rm -f $(ROOT_DIR)/docs/refs/api/$(API_REFERENCE_FILENAME)
+	$(CRD_REF_DOCS) --source-path=$(ROOT_DIR)/api \
+	--config=$(ROOT_DIR)/config/base/crd/bases/olm.operatorframework.io_clusterextensions.yaml \
+	--renderer=markdown \
+	--output-path=$(ROOT_DIR)/docs/refs/api/
```
everettraven, line 309: I did some digging in the source code and it looks like they support this being a path to a file as well - could you verify if setting `--output-path=$(ROOT_DIR)/docs/refs/api/$(API_REFERENCE_FILENAME)` would allow us to drop the `mv` operation after this?

everettraven (dismissed):
Just a minor nit - I don't think it should hold this PR. Thanks @dtfranz !

Makefile
dtfranz: Ah - you're right. I did try that before but I must have formatted it wrong or had some other issue; I assumed it didn't work. Thanks for digging in to find that! I've updated the PR.

everettraven (approved):

PR #1232, opened 2024-09-06
📖 Add provided ServiceAccount documentation to drafts

### LINKED ISSUE #977: Write Provided ServiceAccount Documentation
# User story

Describe the user story or stories for this epic:

- As a user, I expect to be able to define a ServiceAccount that OLM v1 uses to manage content
- As a user, I expect the ClusterExtension installation to fail if the provided ServiceAccount has insufficient permissions
- As a user, I expect the ClusterExtension to be rejected if a ServiceAccount is not provided
- As a user, I expect to be able to identify the permissions required to install a bundle so that I can create a ServiceAccount with all the permissions necessary to install an extension
     
## List the user-facing issues in the epic that require documentation.
- #971 

## Does this epic require new content or updates to existing content?

- [x] Net-new content
- [x] Update existing content

> [!NOTE]
> Preemptively setting that this will need both. I'm not 100% sure, but I think it is likely that both existing content will need to be updated and that new content will need to be written

## What type of documentation is required for this epic?

- [x] Release note
- [x] Concept
- [ ] Procedure
- [x] Reference

## Is there prior art or reference materials that would help write the documentation?
- Once the feature is complete and in place, there will likely be some unit and e2e tests that could be used as reference

Documentation should be contributed to the [`docs/drafts/` directory](https://github.com/operator-framework/operator-controller/tree/main/docs/drafts)

trgeiger:
# Description

Closes #977 

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/docs/drafts/provided-serviceaccount.md
+++ b/docs/drafts/provided-serviceaccount.md
@@ -1,0 +1,31 @@
+# Provided ServiceAccount for ClusterExtension Installation and Management
+
+Adhering to OLM v1's "Secure by Default" tenet, OLM v1 does not have the permissions
+necessary to install content. This follows the least privilege principle and reduces
+the chance of a [confused deputy attack](https://en.wikipedia.org/wiki/Confused_deputy_problem).
+Instead, users must explicitly specify a ServiceAccount that will be used to perform the
+installation and management of a specific ClusterExtension. The ServiceAccount is specified
+in the ClusterExtension manifest as follows:
+
+```yaml
+apiVersion: olm.operatorframework.io/v1alpha1
+kind: ClusterExtension
+metadata:
+  name: argocd
+spec:
+  source:
+    sourceType: Catalog
+    catalog:
+      packageName: argocd-operator
+      version: 0.6.0
+  install:
+    namespace: argocd
+    serviceAccount:
+      name: argocd-installer
+```
+
+The ServiceAccount must be configured with the RBAC permissions required by the ClusterExtension.
+If the permissions do not meet the minimum requirements, installation will fail. If no ServiceAccount
+is provided in the ClusterExtension manifest, then the manifest will be rejected.
+
+//TODO: Add link to documentation on determining least privileges required for the ServiceAccount


trgeiger:
There are some locations where it would make sense to add hyperlinks to this doc, but I'm not sure how we want to handle that while it's in the drafts directory so I didn't add any.

docs/drafts/provided-serviceaccount.md
```diff
@@ -0,0 +1,34 @@
+# Provided ServiceAccount for ClusterExtension Installation and Management
+
+While OLMv0 provides many mechanisms for permission configuration via OperatorGroups,
+or by manipulating CatalogSources, cluster administrators must be aware of these options
+and actually implement them. If no ServiceAccount is explicitly specified for installing
+and upgrading operators, then cluster-admin is used by default. This can pose security risks
+by providing more permissions than are actually required for the management of any specific bundle.
```
everettraven, line 7: Not sure we really need to mention OLMv0 here - what do others think?

docs/drafts/provided-serviceaccount.md
```diff
@@ -0,0 +1,34 @@
...
+
+OLMv1 will not grant cluster-admin permissions. It instead requires ServiceAccounts provided
+by users to install, upgrade and delete content.
```
everettraven, line 10: What do you think of something like:
```suggestion
Adhering to OLM v1's "Secure by Default" tenet, OLM v1 does not have the permissions necessary to install content. This follows the least privilege principle and reduces the chance of a confused deputy attack. Instead, a ServiceAccount must be provided by users to install and manage content.
```
?

docs/drafts/provided-serviceaccount.md
joelanford: I definitely see an argument for certain OLMv0 callouts on specific notable changes. And this is certainly one of those.

But maybe we should just have a separate doc that highlights all of those things in one place?

trgeiger: That sounds like a good idea. So do you think I should remove that mention here and open a ticket to create a different page outlining the differences between OLMv0/v1?

everettraven: I'm +1 for a separate "changes from OLMv0" type documentation item

rashmi43: I agree with this, we can have a separate document for changes from OLMv0 and not mention OLMv0 in this doc.

rashmi43: Maybe we can rephrase as below:
```
OLMv1 differs in terms of permission management from its predecessor, it requires you to specify the permissions required to install cluster extensions upfront. It does not provider cluster admin privileges by default. Cluster administrators can specify the exact permissions that are actually required for the management of any specific bundle.

Comment (rashmi43) [reply docs/drafts/provided-serviceaccount.md:7]: I added the following to my doc:

OLMv1 does not provide cluster admin privileges by default for installing cluster extensions. It depends on the cluster extension developer to specify the permissions required for the management of any specific bundle. A ServiceAccount needs to be explicitly specified for installing and upgrading operators else will face errors when deploying your cluster extension.

**Comment by trgeiger** [comment]
I have reworked most of the doc, integrating everyone's feedback. Ready for another look.

File: docs/drafts/provided-serviceaccount.md [hunk 3]
```diff
@@ -0,0 +1,31 @@
...
+    catalog:
+      packageName: argocd-operator
+      version: 0.6.0
+  install:
+    namespace: argocd
+    serviceAccount:
+      name: argocd-installer
+```
+
+The ServiceAccount must be configured with the necessary permissions required by the ClusterExtension.
```
OchiengEd, line 27: Having `necessary` and `required` in the same sentence is redundant. I'd recommend perhaps replacing `necessary` with RBAC.

docs/drafts/provided-serviceaccount.md
```diff
@@ -0,0 +1,31 @@
+# Provided ServiceAccount for ClusterExtension Installation and Management
+
+Adhering to OLM v1's "Secure by Default" tenet, OLM v1 does not have the permissions
+necessary to install content. This follows the least privilege principle and reduces
+the chance of a [confused deputy attack](https://en.wikipedia.org/wiki/Confused_deputy_problem).
+Instead, users must explicitly specify a ServiceAccount that will be perform the
```
everettraven, line 6:
```suggestion
Instead, users must explicitly specify a ServiceAccount that will be used to perform the
```

everettraven (commented):
One minor nit, but other than that it LGTM!

tmshort (approved):
/lgtm

PR #1240, opened 2024-09-09
📖 catalogd API Reference Generation

### LINKED ISSUE #1125: Docs: Catalogd API reference
Write Catalogd API reference documentation. Sync with @michaelryanpeter to see if there are any tools that can go from godoc to the API reference. Another option might to source the information from the CRD which is in openAPI format. There might already be tools to process that. (See what k8s does here: https://kubernetes.io/docs/contribute/generate-ref-docs/kubernetes-api/)

- [ ] ClusterCatalog API reference documentation

Current documentation is posted up at https://operator-framework.github.io/operator-controller/
New docs should be placed in [docs/drafts](https://github.com/operator-framework/operator-controller/tree/main/docs/drafts)

> [!NOTE]
> This documentation should be written _after_ https://github.com/operator-framework/catalogd/issues/239

dtfranz:
Generates the catalogd API reference along with the operator-controller's

Closes #1125 

# Description

## Reviewer Checklist

- [ ] API Go Documentation


Makefile
```diff
@@ -300,13 +300,18 @@ quickstart: $(KUSTOMIZE) manifests #EXHELP Generate the installation release man
...
+	$(Q)go mod vendor
+	rm -f $(ROOT_DIR)/docs/refs/api/$(OPERATOR_CONTROLLER_API_REFERENCE_FILENAME)
 	$(CRD_REF_DOCS) --source-path=$(ROOT_DIR)/api \
-	--config=$(ROOT_DIR)/config/base/crd/bases/olm.operatorframework.io_clusterextensions.yaml \
-	--renderer=markdown \
-	--output-path=$(ROOT_DIR)/docs/refs/api/$(API_REFERENCE_FILENAME)
+	--config=$(ROOT_DIR)/docs/refs/api/crd-ref-docs-gen-config.yaml \
+	--renderer=markdown --output-path=$(ROOT_DIR)/docs/refs/api/$(OPERATOR_CONTROLLER_API_REFERENCE_FILENAME);
+	rm -f $(ROOT_DIR)/docs/refs/api/$(CATALOGD_API_REFERENCE_FILENAME)
+	$(CRD_REF_DOCS) --source-path=$(ROOT_DIR)/vendor/github.com/operator-framework/catalogd/api \
```
dtfranz, line 312: Are we OK with grabbing this from vendor? I feel like there has to be a better option but nothing else obvious occurred to me.

Makefile
```diff
@@ -300,13 +300,18 @@ quickstart: $(KUSTOMIZE) manifests #EXHELP Generate the installation release man
...
+OPERATOR_CONTROLLER_API_REFERENCE_FILENAME := operator-controller-api-reference.md
+CATALOGD_API_REFERENCE_FILENAME := catalogd-api-reference.md
+crd-ref-docs: tidy $(CRD_REF_DOCS)
+	$(Q)go mod vendor
+	rm -f $(ROOT_DIR)/docs/refs/api/$(OPERATOR_CONTROLLER_API_REFERENCE_FILENAME)
 	$(CRD_REF_DOCS) --source-path=$(ROOT_DIR)/api \
-	--config=$(ROOT_DIR)/config/base/crd/bases/olm.operatorframework.io_clusterextensions.yaml \
-	--renderer=markdown \
-	--output-path=$(ROOT_DIR)/docs/refs/api/$(API_REFERENCE_FILENAME)
+	--config=$(ROOT_DIR)/docs/refs/api/crd-ref-docs-gen-config.yaml \
```
dtfranz, line 309: I'd misunderstood the usage of this flag when I first implemented this, but turns out if you don't specify `--source-path` and _do_ specify `--config` as your CRD file, it generates the same doc. Confusing? Or maybe it finds the `api` folder automatically if not specified?

Regardless, this is the actual intended usage. `--config` specifies a configuration file for the generator. At the moment we just specify the reference k8s version, which also consequently fixes the broken k8s API reference links in our doc.

docs/refs/api/operator-controller-api-reference.md
```diff
@@ -81,7 +81,7 @@ _Appears in:_
...
-| `selector` _[LabelSelector](https://kubernetes.io/docs/reference/generated/kubernetes-api/v/#labelselector-v1-meta)_ | selector is an optional field that can be used<br />to filter the set of ClusterCatalogs used in the bundle<br />selection process.<br /><br />When unspecified, all ClusterCatalogs will be used in<br />the bundle selection process. |  |  |
+| `selector` _[LabelSelector](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.31/#labelselector-v1-meta)_ | selector is an optional field that can be used<br />to filter the set of ClusterCatalogs used in the bundle<br />selection process.<br /><br />When unspecified, all ClusterCatalogs will be used in<br />the bundle selection process. |  |  |
```
dtfranz, line 84: As noted in earlier comment, these links were broken due to incorrect config file usage.

Makefile
joelanford: We don't vendor our dependencies in this repo, so I think we need to do something different.

However, there is a plan afoot to make a monorepo out of operator-controller and catalogd. If there isn't an interim solution for generating docs from out-of-tree APIs, we'll be able to do this after we monorepo-ize.

Makefile
dtfranz: From what I can tell, no, the `api/` folder is the only available source of truth for this tool. If the `--source-path` isn't specified, it'll crawl through your folder structure and find the `api/` folder itself.

Do you have an example of what kind of change we would make from a patch like that? Something like field descriptions, validations, or other things not required to be present in the `_types.go` file is what I was thinking of. We may be able to cover this kind of thing using their custom markers and templates ([reference](https://github.com/elastic/crd-ref-docs?tab=readme-ov-file#advanced-features)).

joelanford: Here's an example from catalogd: https://github.com/operator-framework/catalogd/blob/main/config/base/crd/patches/catalog_validation.yaml

Makefile
dtfranz: Not having thought of anything else, I've updated this to do above. Let me know if you're alright with this method. It's definitely a little clumsy though, since a local environment isn't guaranteed to have `catalogd` checked out in the right spot.

everettraven: @dtfranz What do you think about using `git` to clone the version of catalogd we depend on in `go.mod` into a temporary directory and pointing to that instead?

I saw that you were able to clone catalogd in CI to the appropriate path, but I'd like to see it work without any additional configuration on the dev side.

That being said, this probably doesn't really matter that much since we are actively planning to move to a monorepo (already seems to be good consensus on the monorepo idea, now we are hashing out the _how_).

everettraven (approved):

Makefile
dtfranz: Thanks @everettraven ! That's a good call. I'll merge this for now and do that as an immediate follow-up. As a bonus, it would simplify the github action a little bit as well.

PR #1250, opened 2024-09-11
⚠ minor API improvements

### LINKED ISSUE #1251: :bug: More API cleanups
I found a few more little things in the API that need to be updated. See https://github.com/operator-framework/operator-controller/pull/1250

joelanford:
# Description

More minor improvements to the API in preparation for 1.0.0.

Fixes #1251

## Reviewer Checklist

- [ ] API Go Documentation


api/v1alpha1/clusterextension_types.go
```diff
@@ -23,8 +23,7 @@ import (
 )
 
 var (
-	ClusterExtensionGVK  = SchemeBuilder.GroupVersion.WithKind("ClusterExtension")
```
joelanford, line 26: `ClusterExtensionGVK` is unused, and is not really necessary for inclusion as an exported global variable in the API.

Callers can construct themselves like this:
```go
gvk := ocv1alpha1.GroupVersion.WithKind(v1alpha1.ClusterExtensionKind)
```

api/v1alpha1/clusterextension_types.go
```diff
@@ -485,15 +484,17 @@ type ClusterExtensionStatus struct {
...
+	// The global set of condition types are:
+	//   - "Installed", represents whether or not the a bundle has been installed for this ClusterExtension
 	//   - "Resolved", represents whether or not a bundle was found that satisfies the selection criteria outlined in the spec
-	//   - "Deprecated", represents an aggregation of the PackageDeprecated, ChannelDeprecated, and BundleDeprecated condition types.
-	//   - "PackageDeprecated", represents whether or not the package specified in the spec.packageName field has been deprecated
-	//   - "ChannelDeprecated", represents whether or not the channel specified in spec.channel has been deprecated
-	//   - "BundleDeprecated", represents whether or not the bundle installed is deprecated
 	//   - "Unpacked", represents whether or not the bundle contents have been successfully unpacked
 	//
+	// When the ClusterExtension is sourced from a catalog, the following conditions are also possible:
```
joelanford, line 492: I grouped the `Deprecated*` conditions into their own section for two reasons:
1. I think it is helpful to see the other "main" conditions without the "noise" of these being intermingled.
2. Deprecation metadata comes from catalogs, and while we currently only support sourcing from catalogs, I thought it might be good to call out the fact that deprecations will only ever show up if the ClusterExtension is sourced from a catalog.

api/v1alpha1/clusterextension_types.go
```diff
@@ -485,15 +484,17 @@ type ClusterExtensionStatus struct {
...
+	//   - "Deprecated", represents an aggregation of the PackageDeprecated, ChannelDeprecated, and BundleDeprecated condition types
+	//   - "PackageDeprecated", represents whether or not the package specified in the spec.source.catalog.packageName field has been deprecated
+	//   - "ChannelDeprecated", represents whether or not any channel specified in spec.source.catalog.channels has been deprecated
+	//   - "BundleDeprecated", represents whether or not the installed bundle is deprecated
```
joelanford, line 496: I'm pretty sure this line is not true, but want to double check. A quick glance at the code makes me think we should make the following edit as well:
```suggestion
	//   - "BundleDeprecated", represents whether or not the resolved bundle is deprecated
```

However, I think the original sentiment of having this condition be based on the installed bundle is what we actually want. That might require a bit more conversation and consideration though, as it isn't clear to me how exactly we would assess the deprecation status of an installed bundle during subsequent reconciles after it is installed.

api/v1alpha1/clusterextension_types.go
```diff
@@ -511,22 +512,17 @@ type ClusterExtensionInstallStatus struct {
...
 	// A "bundle" is a versioned set of content that represents the resources that
 	// need to be applied to a cluster to install a package.
-	//
-	// This field is only updated once a bundle has been successfully installed and
-	// once set will only be updated when a new version of the bundle has
-	// successfully replaced the currently installed version.
-	//
-	//+optional
-	Bundle *BundleMetadata `json:"bundle,omitempty"`
+	Bundle BundleMetadata `json:"bundle"`
```
joelanford, line 515: I removed the pointers here because the parent struct is a pointer. If the parent struct is non-nil, this field is _always_ expected to be set.

api/v1alpha1/clusterextension_types.go
```diff
@@ -511,22 +512,17 @@ type ClusterExtensionInstallStatus struct {
...
 type ClusterExtensionResolutionStatus struct {
 	// bundle is a representation of the bundle that was identified during
 	// resolution to meet all installation/upgrade constraints and is slated to be
 	// installed or upgraded to.
 	//
-	//+optional
-	Bundle *BundleMetadata `json:"bundle,omitempty"`
+	// A "bundle" is a versioned set of content that represents the resources that
+	// need to be applied to a cluster to install a package.
+	Bundle BundleMetadata `json:"bundle"`
```
joelanford, line 525: See: https://github.com/operator-framework/operator-controller/pull/1250/files#r1754304430

api/v1alpha1/clusterextension_types.go
perdasilva: I guess we'd have two possibilities: 
 1. You want to install a bundle from the catalog, and it's deprecated -  which is straightforward
 2. You install a bundle that isn't deprecated, and in the next catalog version it is (I guess this is only meaningful for version ranges that don't let the bundle upgrade to the next (non-deprecated) version.
 
 In the 2nd case, shouldn't it pick up the condition during reconcile? Maybe it's OK that we only express deprecated conditions for resolved bundles (and not worry about the the installed bundles) since either the CE will get upgraded, or if it can't/won't, we'd anyway resolve to the installed bundle?

config/base/crd/bases/olm.operatorframework.io_clusterextensions.yaml
```diff
@@ -474,15 +474,17 @@ spec:
...
                     - "Unpacked", represents whether or not the bundle contents have been successfully unpacked
 
+                  When the ClusterExtension is sourced from a catalog, the following conditions are also possible:
+                    - "Deprecated", represents an aggregation of the PackageDeprecated, ChannelDeprecated, and BundleDeprecated condition types
+                    - "PackageDeprecated", represents whether or not the package specified in the spec.source.catalog.packageName field has been deprecated
+                    - "ChannelDeprecated", represents whether or not any channel specified in spec.source.catalog.channels has been deprecated
+                    - "BundleDeprecated", represents whether or not the installed bundle is deprecated
+
                   The current set of reasons are:
                     - "Success", this reason is set on the "Unpacked", "Resolved" and "Installed" conditions when unpacking a bundle's content, resolution and installation/upgrading is successful
```
perdasilva, line 489: Do we need to call out the reasons for the deprecated conditions?

perdasilva (approved):
Nice one! Thank you ^^ 

I left a question about whether we ought to call out the Deprecated* condition reasons in the docs. If not, take the approval, if yes, happy to approve again =D

api/v1alpha1/clusterextension_types.go
joelanford: My initial thought is that we only touch the `BundleDeprecated` condition when:
1. We successfully install a bundle. In this case the resolved bundle and the installed bundle are the same, so the resolved deprecation is applicable to the installed bundle.
2. We detect that a bundle is somehow no longer installed, in which case we unset (or set to `Unknown`) the `BundleDeprecated` condition.

config/base/crd/bases/olm.operatorframework.io_clusterextensions.yaml
joelanford: I don't think so. The "Reason" for these conditions are basically 1-to-1 with the Status. "The reason this is deprecated is because the author provided a deprecation"

PR #1269, opened 2024-09-13
:sparkles: reduce use of carvel validators

### LINKED ISSUE #1246: [CRD Upgrade Safety] Add a change validation for handling field type changes
It is considered a breaking change to update the type of a field when upgrading a CRD. We currently do not have an explicit check for this type of change so we get an output similar to:
```yaml
message: 'validating upgrade for CRD "nginxolm75123s.cache.example.com" failed:
      CustomResourceDefinition nginxolm75123s.cache.example.com failed upgrade safety
      validation. "ChangeValidator" validation failed: version "v1alpha1", field "^.spec.field"
      has unknown change, refusing to determine that change is safe'
```

We should add a new change validation that explicitly handles this case to improve the UX in identifying the exact reason why a CRD upgrade was rejected and help users make better informed decisions as to if this is a change they are comfortable with going through on their cluster.

This is not a v1.0.0 blocker as it is a QoL improvement and the existing pre-flight check fails closed on an unrecognized change so we aren't causing any harm to a user's cluster with the current state of the check.

### LINKED ISSUE #1248: [CRD Upgrade Safety] Message output use the human readable value instead of exact bytes
When upgrading a CRD, 
A new default value is added to a field that did not previously have a default value
The default value of a field is changed
An existing default value of a field is removed

This looks like it is behaving properly but the exact message output seems pretty unhelpful since it seems to be giving the exact bytes instead of a human readable value. 
we get an output similar to:
<pre>
message: 'validating upgrade for CRD "nginxolm75124s.cache.example.com" failed:
      CustomResourceDefinition nginxolm75124s.cache.example.com failed upgrade safety
      validation. "ChangeValidator" validation failed: version "v1alpha1", field "^.spec.field":
      new value added as default when previously no default value existed: &JSON{Raw:*[34
      100 101 102 97 117 108 116 45 115 116 114 105 110 103 45 106 105 116 108 105
      34],}'
</pre>

<pre>
message: 'validating upgrade for CRD "nginxolm75124s.cache.example.com" failed:
      CustomResourceDefinition nginxolm75124s.cache.example.com failed upgrade safety
      validation. "ChangeValidator" validation failed: version "v1alpha1", field "^.spec.defaultenum":
      default value has been changed from [34 118 97 108 117 101 49 34] to [34 118
      97 108 117 101 51 34]'
</pre>

<pre>
message: 'validating upgrade for CRD "nginxolm75124s.cache.example.com" failed:
      CustomResourceDefinition nginxolm75124s.cache.example.com failed upgrade safety
      validation. "ChangeValidator" validation failed: version "v1alpha1", field "^.spec.defaultint":
      default value has been removed when previously a default value existed: [57]'
</pre>

everettraven:
# Description
- Reduces the use of the field change validators in carvel/kapp in favor of our own implementations. This makes the CRD Upgrade Safety pre-flight check more maintainable by allowing us to make bugfixes and changes in our implementation without having to contribute to carvel/kapp and wait for a new release.

- fixes #1248 
- fixes #1246

## Reviewer Checklist

- [ ] API Go Documentation


internal/rukpak/preflights/crdupgradesafety/checks.go
```diff
@@ -70,3 +73,246 @@ func (c *ServedVersionValidator) Validate(old, new apiextensionsv1.CustomResourc
...
+	}
+
+	var err error
+
+	switch {
+	case diff.Old.Maximum == nil && diff.New.Maximum != nil:
+		err = fmt.Errorf("maximum constraint %v added when there were no restrictions previously", *diff.New.Maximum)
+	case diff.Old.Maximum != nil && diff.New.Maximum != nil && *diff.New.Maximum < *diff.Old.Maximum:
+		err = fmt.Errorf("maximum constraint decreased from %v to %v", *diff.Old.Maximum, *diff.New.Maximum)
+	}
```
everettraven, line 146: This switch statement is for the most part repeated across all the `Max*` functions. I'm planning to make an update to have a function that is re-used across all of them, but didn't have time left and wanted to get this PR up before I logged off for the day. I figure it's not the end of the world if this goes in as-is since I could file a follow-up. If it hasn't merged by the time I get to circle back around to it I'll push the update :)

internal/rukpak/preflights/crdupgradesafety/checks.go
```diff
@@ -70,3 +73,246 @@ func (c *ServedVersionValidator) Validate(old, new apiextensionsv1.CustomResourc
...
+
+	var err error
+
+	switch {
+	case diff.Old.Minimum == nil && diff.New.Minimum != nil:
+		err = fmt.Errorf("minimum constraint %v added when there were no restrictions previously", *diff.New.Minimum)
+	case diff.Old.Minimum != nil && diff.New.Minimum != nil && *diff.New.Minimum > *diff.Old.Minimum:
+		err = fmt.Errorf("minimum constraint increased from %v to %v", *diff.Old.Minimum, *diff.New.Minimum)
+	}
```
everettraven, line 222: This switch statement is for the most part repeated across all the `Min*` functions. I'm planning to make an update to have a function that is re-used across all of them, but didn't have time left and wanted to get this PR up before I logged off for the day. I figure it's not the end of the world if this goes in as-is since I could file a follow-up. If it hasn't merged by the time I get to circle back around to it I'll push the update :)

internal/rukpak/preflights/crdupgradesafety/crdupgradesafety.go
```diff
@@ -32,17 +32,18 @@ type Preflight struct {
 
 func NewPreflight(crdCli apiextensionsv1client.CustomResourceDefinitionInterface, opts ...Option) *Preflight {
 	changeValidations := []kappcus.ChangeValidation{
```
joelanford, line 34: How much further would we need to go to fully drop the kapp dependency?

everettraven: Not 100% sure, but level of effort depends on whether we want to re-implement in our own way such that we don't need to adhere to Apache 2.0 license guidelines or if we would want to copy-paste the code we need including maintaining and adhering to the Apache 2.0 license guidelines

everettraven: That being said, I would say totally dropping the kapp dependency should be seen outside the scope of this PR. While this PR reduces our use of the carvel provided validators, the main reason for this PR is to address some minor bugs in the output when the checks fail. My approach to doing this just happens to reduce the carvel validator use because I felt it would be easier for us to maintain long-term as opposed to contributing the changes all the way upstream.

internal/rukpak/preflights/crdupgradesafety/checks.go
everettraven: Done in https://github.com/operator-framework/operator-controller/pull/1269/commits/787b14815e628f68fae683e2125f6b4289dd380e

internal/rukpak/preflights/crdupgradesafety/checks.go
everettraven: Done in https://github.com/operator-framework/operator-controller/pull/1269/commits/787b14815e628f68fae683e2125f6b4289dd380e

internal/rukpak/preflights/crdupgradesafety/crdupgradesafety.go
joelanford: Are the check implementations copy/pasted from kapp or new implementations? I ask because:
- If they are copy/pasted:
    - I don't need to review them as carefully,
    - BUT, we need some sort of attribution?
- If they are not copy/pasted,
    - We should review thoroughly
    - We don't need attribution

everettraven: They are not copy/paste-ed but *inspired* by the carvel implementation (which I helped write so hard to not be inspired by), but I think the implementation is changed significantly enough that it doesn't require any attribution

tmshort:
Where are we on this one?

PR #1270, opened 2024-09-14
✨ follow-ups for containers/image from catalogd and previous PR

joelanford:
# Description

## Reviewer Checklist

- [ ] API Go Documentation


internal/rukpak/source/containers_image.go
```diff
@@ -89,7 +85,11 @@ func (i *ContainersImageRegistry) Unpack(ctx context.Context, bundle *BundleSour
 	if err != nil {
 		return nil, fmt.Errorf("error creating temporary directory: %w", err)
 	}
-	defer os.RemoveAll(layoutDir)
+	defer func() {
+		if err := os.RemoveAll(layoutDir); err != nil {
```
tmshort, line 89: Should this be `deleteRecursive()`?

internal/rukpak/source/containers_image.go
```diff
@@ -174,31 +169,60 @@ func (i *ContainersImageRegistry) unpackPath(bundleName string, digest digest.Di
...
+	}
+
+	canonicalRef, isCanonical, err := resolveCanonicalRef(ctx, imgRef, sourceContext)
+	if err != nil {
+		return nil, nil, false, fmt.Errorf("error resolving canonical reference: %w", err)
+	}
+	return imgRef, canonicalRef, isCanonical, nil
+}
+
+func resolveCanonicalRef(ctx context.Context, imgRef reference.Named, imageCtx *types.SystemContext) (reference.Canonical, bool, error) {
```
tmshort, line 185: It's not clear what the `bool` return is for. There's no explanation (I assume it's `isCanonical`?)

internal/rukpak/source/containers_image.go
```diff
@@ -217,7 +241,7 @@ func (i *ContainersImageRegistry) unpackImage(ctx context.Context, unpackPath st
 		return fmt.Errorf("error creating image source: %w", err)
 	}
 
-	if err := os.MkdirAll(unpackPath, 0755); err != nil {
+	if err := os.MkdirAll(unpackPath, 0700); err != nil {
```
tmshort, line 244: Why the restrictive mode?

tmshort: Also noting `setReadOnlyRecursive`... these only set the owner's bits, not group/other.

internal/rukpak/source/containers_image.go
```diff
@@ -45,14 +46,9 @@ func (i *ContainersImageRegistry) Unpack(ctx context.Context, bundle *BundleSour
 	// Resolve a canonical reference for the image.
 	//
 	//////////////////////////////////////////////////////
-	imgRef, err := reference.ParseNamed(bundle.Image.Ref)
+	imgRef, canonicalRef, _, err := resolveReferences(ctx, bundle.Image.Ref, i.SourceContext)
```
LalatenduMohanty, line 49: @joelanford I am curious about this change. What triggered this change?

LalatenduMohanty: We are moving from a containers/image method to local implementation, hence the question.

LalatenduMohanty: Nevermind I saw that it is just a wrapper on top of ` reference.ParseNamed(ref)`

internal/rukpak/source/containers_image.go
LalatenduMohanty: It is the same mode as we are setting in ` applyLayerFilter()`

internal/rukpak/source/containers_image.go
joelanford: It was based on a suggestion from catalogd's implementation PR for containers/image. I'm trying to keep them as close to the same as I can.

internal/rukpak/source/containers_image.go
joelanford: This one doesn't have to be because `layoutDir` only ever exists while this function is running, and we never set it to be read-only like we do for the unpack dir.

internal/rukpak/source/containers_image.go
joelanford: Yeah, its `isCanonical`. One way to tell is by looking at the call site in line 178.

But, this one is a bit weird because operator-controller doesn't actually use this variable. Keep moving up the stack and you'll see we discard its value on line 49.

As best I can, I'm trying to keep the function signatures the same between catalogd and operator-controller to facilitate merging later.

internal/rukpak/source/containers_image.go
joelanford: Just trying to be consistent every where. These files are always written and read just by us, so these permissions should be fine. The cache dir is in an emptyDir mount that starts out empty every time a pod starts, so we should never need to be reading files written by another UID.

internal/rukpak/source/containers_image.go
```diff
@@ -269,9 +300,65 @@ func (i *ContainersImageRegistry) deleteOtherImages(bundleName string, digestToK
...
+				// do not follow symlinks
+				// 1. if they resolve to other locations in the root, we'll find them anyway
+				// 2. if they resolve to other locations outside the root, we don't want to change their permissions
+				return nil
+			case os.ModeDir:
+				return os.Chmod(path, 0500)
+			case 0: // regular file
+				return os.Chmod(path, 0400)
+			default:
+				return fmt.Errorf("refusing to change ownership of file %q with type %v", path, typ.String())
```
everettraven, line 333: Nit: Should we use the type format directive here?
```suggestion
				return fmt.Errorf("refusing to change ownership of file %q with type %T", path, typ)
```

joelanford: No, I don't think so. This is a file type, not a Go type.

tmshort (approved):
/lgtm

I'm ok with the explanations.

PR #1276, opened 2024-09-16
✨ use controller-runtime Terminal error instead of our custom Unrecoverable error

### LINKED ISSUE #1271: Instead of custom `Unrecoverable` error, use controller-runtime's `TerminalError`
See https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.19.0/pkg/reconcile#TerminalError

joelanford:
Fixes #1271

# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/rukpak/source/containers_image.go
+++ b/internal/rukpak/source/containers_image.go
@@ -21,6 +21,7 @@ import (
 	"github.com/containers/image/v5/types"
 	"github.com/opencontainers/go-digest"
 	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
 type ContainersImageRegistry struct {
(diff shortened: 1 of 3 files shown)


LalatenduMohanty (approved):
/lgtm

LalatenduMohanty:
This makes sense, was not aware of func [TerminalError](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.19.0/pkg/reconcile/reconcile.go#L159) [¶](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.19.0/pkg/reconcile#TerminalError)

everettraven (approved):

everettraven:
I don't think it needs to block this PR since it could be done in a follow-up, but it seems like we never made use of the unrecoverable error in the `ClusterExtensionReconciler` and we should probably update that.

joelanford:
@everettraven the beauty of this error type is that controller-runtime type checks for it after we return from `Reconcile` and does the same logic that we did (but with a few extras like incrementing a different metric)

https://github.com/kubernetes-sigs/controller-runtime/blob/2eb879f25c4829825e6f2511cda035de41ec7030/pkg/internal/controller/controller.go#L306-L310

PR #1283, opened 2024-09-19
🐛 fix: do not silently drop finalizer updates when status is also updated

joelanford:
This is a fix for a bug that @anik120 stumbled upon in catalogd while working on https://github.com/operator-framework/catalogd/pull/379.

Wanted to fix it here as well while we're thinking about it.

# Description

## Reviewer Checklist

- [ ] API Go Documentation


joelanford:
`codecov/patch` test seems to be failing because I moved the check for unexpected changes.

anik120 (dismissed):
> `codecov/patch` test seems to be failing because I moved the check for unexpected changes.

Fwiw there were no tests added to confirm this was a problem before(which is fine coz we didn't know it was a problem), but there's no tests now to confirm on paper it is still not a problem

tmshort:
@anik120 not sure we want to test a panic, also not sure how to get a unit test to modify fields while we're in `reconcile()` This really is a "should never happen" type of situation during normal runs. It _might_ panic if someone were making an incorrect code change.

tmshort (approved):
/lgtm

everettraven (approved):

PR #1284, opened 2024-09-19
✨ Populate/update cache on ClusterCatalog reconcile

m1kola:
# Description

Currently we fetch catalog data and populate cache on demand during `ClusterExtension` reconciliation. This works but the first reconciliation after `ClusterCatalog` creation or update is slow due to the need to fetch data.

With this change we proactively populate cache on `ClusterCatalog` creation and check if cache needs to be updated on `ClusterCatalog` update.

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/controllers/clustercatalog_controller.go
+++ b/internal/controllers/clustercatalog_controller.go
@@ -18,25 +18,30 @@ package controllers
 
 import (
 	"context"
+	"fmt"
+	"io/fs"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	ctrl "sigs.k8s.io/controller-runtime"
-	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/event"
-	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
 	catalogd "github.com/operator-framework/catalogd/api/core/v1alpha1"
 )
 
-type CatalogCacheRemover interface {
+type CatalogCache interface {
+	Get(catalogName, resolvedRef string) (fs.FS, error)
 	Remove(catalogName string) error
+}
+
+type CatalogCachePopulator interface {
+	PopulateCache(ctx context.Context, catalog *catalogd.ClusterCatalog) (fs.FS, error)
 }
 
 // ClusterCatalogReconciler reconciles a ClusterCatalog object
 type ClusterCatalogReconciler struct {
 	client.Client
-	Cache CatalogCacheRemover
+	CatalogCache          CatalogCache
+	CatalogCachePopulator CatalogCachePopulator
 }
 
 //+kubebuilder:rbac:groups=olm.operatorframework.io,resources=clustercatalogs,verbs=get;list;watch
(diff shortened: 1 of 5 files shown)


internal/controllers/clustercatalog_controller.go
```diff
@@ -0,0 +1,64 @@
...
+	err := r.Client.Get(ctx, req.NamespacedName, existingCatalog)
+	if client.IgnoreNotFound(err) != nil {
+		return ctrl.Result{}, err
+	}
+	if apierrors.IsNotFound(err) {
+		return ctrl.Result{}, r.Cache.Remove(req.Name)
+	}
+
+	_, err = r.Cache.FetchCatalogContents(ctx, existingCatalog)
+	return ctrl.Result{}, err
```
joelanford, line 54: One benefit of the previous approach is that catalog pull errors will show up in ClusterExtension status.

If we move catalog cache populate to a separate controller, do users lose all visibility into the specific details of the reason the catalog caches aren't populated?

I wonder if we should store these errors somewhere that is available to the ClusterExtension reconciler so that we can still propagate them into ClusterExtension status?

m1kola: I'm considering keeping on-demand cache population. If cache fetch fails here - we will keep re-queueing until it gets populated. If it doesn't and it is something permanent - cache population will fail during `ClusterExtension` itself and it will be propagated to `ClusterExtension` status as it currently works.

It should work, but I don't like it very much because it is not straightforward to understand.

> I wonder if we should store these errors somewhere that is available to the ClusterExtension reconciler so that we can still propagate them into ClusterExtension status?

Any ideas? We can probably store somewhere in `filesystemCache` ([e.g. in `cacheDataByCatalogName`](https://github.com/operator-framework/operator-controller/blob/f3e2ce0727136bbfe572e750fef1ce1e38795552/internal/catalogmetadata/cache/cache.go#L54))?

joelanford: High level idea would be for the interface that `ClusterExtension` uses to remain the same (e.g. it can return an error).

Yeah `cacheDataByCatalogName` seems like the obvious place. We could have `cacheData` include an `Err` field?

m1kola: @joelanford could you please take a look at https://github.com/operator-framework/operator-controller/pull/1318?

I refactored the client and cache: the client is now responsible for http requests and cache is responsible for caching only (previously it was making request). And I also added error propagation.

Once it is merged I'll be able to rebase this PR on top of the changes.

internal/controllers/clustercatalog_controller.go
```diff
@@ -45,31 +50,44 @@ func (r *ClusterCatalogReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+		existingCatalog.Status.ResolvedSource.Image.Ref == "" {
+		// Reference is not known yet - skip cache population with no error.
+		// Once the reference is resolved another reconcile cycle
+		// will be triggered and we will progress further.
+		return ctrl.Result{}, nil
+	}
+
+	catalogFsys, err := r.CatalogCache.Get(existingCatalog.Name, existingCatalog.Status.ResolvedSource.Image.Ref)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("error retrieving cache for catalog %q: %v", existingCatalog.Name, err)
```
joelanford, line 73: WDYT about propagating this error into the cache so that cache clients have more context about why the content is not available?

internal/controllers/clustercatalog_controller.go
```diff
@@ -45,31 +50,44 @@ func (r *ClusterCatalogReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+	}
+	if catalogFsys != nil {
+		// Cache already exists so we do not need to populate it
+		return ctrl.Result{}, nil
+	}
+
+	if _, err = r.CatalogCachePopulator.PopulateCache(ctx, existingCatalog); err != nil {
+		return ctrl.Result{}, fmt.Errorf("error fetching catalog contents for catalog %q: %v", existingCatalog.Name, err)
```
joelanford, line 81:
```suggestion
		return ctrl.Result{}, fmt.Errorf("error populating cache for catalog %q: %v", existingCatalog.Name, err)
```

internal/controllers/clustercatalog_controller.go
m1kola: @joelanford it is currently implemented how we discussed here: https://github.com/operator-framework/operator-controller/pull/1284#discussion_r1767560927

Cache client caches content fetch errors and `r.CatalogCache.Get(...)` will return fetch error to the clients. So here we will have something like `error retrieving cache for catalog "my-catalog": network request timeout or other error from actual network call`

joelanford: Ah I see now. I didn't realize most of the plumbing had already been done! Beautiful!

m1kola:
Updated the error wording. @joelanford please take another look.

joelanford (approved):

PR #1302, opened 2024-09-23
✨ add progressing condition

### LINKED ISSUE #1292: [v1 API Review] Add `Progressing` status condition
To better signal to a user when a ClusterExtension is transitioning to a new state we need to add a `Progressing` status condition. This condition's state does not affect the `Installed` condition state unless the ClusterExtension has not progressed successfully through the initial installation. 

Some examples of when state transitions may occur:
- Resolving bundle to be installed for the ClusterExtension
- Unpacking bundle contents
- Installing/Upgrading resources for the ClusterExtension

Proposed `Status` and `Reason` combinations are:
- **Status**: _False_, **Reason**: _Success_ - reconciliation has completed successfully, no further progression necessary
- **Status**: _False_, **Reason**: _Blocked_ - error that requires manual user intervention was encountered during reconciliation. We cannot or will not progress further. As an example, the bundle image reference we resolved from a catalog is invalid (unparseable).
- **Status**: _True_, **Reason**: _Retrying_ - error encountered during reconciliation that is recoverable without manual intervention. ClusterExtension is re-queued for reconciliation. Some examples of when this condition case will be populated:
    - Upgrade operation fails leaving the ClusterExtension in an intermediate state
    - Permission errors related to informer establishment
    
everettraven:
# Description
- Adds a new `Progressing` status condition
- Removes unpack states that we never expect to be returned from our implementation, makes reconciliation panic if we get an unexpected state since it is a programmer error
- Consolidates some unit tests
- fixes #1292 

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/controllers/clusterextension_controller.go
+++ b/internal/controllers/clusterextension_controller.go
@@ -201,6 +201,7 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
 		setInstallStatus(ext, nil)
 		setResolutionStatus(ext, nil)
 		setResolvedStatusConditionFailed(ext, err.Error())
+		setStatusProgressing(ext, err)
 		ensureAllConditionsWithReason(ext, ocv1alpha1.ReasonFailed, err.Error())
 		return ctrl.Result{}, err
 	}
@@ -216,6 +217,7 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
 		setInstallStatus(ext, nil)
 		// TODO: use Installed=Unknown
 		setInstalledStatusConditionFailed(ext, err.Error())
+		setStatusProgressing(ext, err)
 		return ctrl.Result{}, err
 	}
 
@@ -227,6 +229,7 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
 		setInstallStatus(ext, nil)
 		setResolutionStatus(ext, nil)
 		setResolvedStatusConditionFailed(ext, err.Error())
+		setStatusProgressing(ext, err)
 		ensureAllConditionsWithReason(ext, ocv1alpha1.ReasonFailed, err.Error())
 		return ctrl.Result{}, err
 	}
@@ -247,8 +250,9 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
 	//         all catalogs?
 	SetDeprecationStatus(ext, resolvedBundle.Name, resolvedDeprecation)
 
+	resolvedBundleMetadata := bundleutil.MetadataFor(resolvedBundle.Name, *resolvedBundleVersion)
 	resStatus := &ocv1alpha1.ClusterExtensionResolutionStatus{
-		Bundle: bundleutil.MetadataFor(resolvedBundle.Name, *resolvedBundleVersion),
+		Bundle: resolvedBundleMetadata,
 	}
 	setResolutionStatus(ext, resStatus)
 	setResolvedStatusConditionSuccess(ext, fmt.Sprintf("resolved to %q", resolvedBundle.Image))
@@ -264,20 +268,18 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
 	unpackResult, err := r.Unpacker.Unpack(ctx, bundleSource)
 	if err != nil {
 		setStatusUnpackFailed(ext, err.Error())
+		// Wrap the error passed to this with the resolution information until we have successfully
+		// installed since we intend for the progressing condition to replace the resolved condition
+		// and will be removing the .status.resolution field from the ClusterExtension status API
+		setStatusProgressing(ext, wrapErrorWithResolutionInfo(resolvedBundleMetadata, err))
 		return ctrl.Result{}, err
 	}
 
 	switch unpackResult.State {
-	case rukpaksource.StatePending:
-		setStatusUnpackFailed(ext, unpackResult.Message)
-		ensureAllConditionsWithReason(ext, ocv1alpha1.ReasonFailed, "unpack pending")
-		return ctrl.Result{}, nil
 	case rukpaksource.StateUnpacked:
 		setStatusUnpacked(ext, unpackResult.Message)
 	default:
-		setStatusUnpackFailed(ext, "unexpected unpack status")
-		// We previously exit with a failed status if error is not nil.
-		return ctrl.Result{}, fmt.Errorf("unexpected unpack status: %v", unpackResult.Message)
+		panic(fmt.Sprintf("unexpected unpack state %q", unpackResult.State))
 	}
 
 	objLbls := map[string]string{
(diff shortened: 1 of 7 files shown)


internal/controllers/clusterextension_controller_test.go
```diff
@@ -89,89 +89,17 @@ func TestClusterExtensionResolutionFails(t *testing.T) {
 	require.Equal(t, ocv1alpha1.ReasonFailed, cond.Reason)
 	require.Equal(t, fmt.Sprintf("no package %q found", pkgName), cond.Message)
 
-	verifyInvariants(ctx, t, reconciler.Client, clusterExtension)
-	require.NoError(t, cl.DeleteAllOf(ctx, &ocv1alpha1.ClusterExtension{}))
-}
-
-func TestClusterExtensionResolutionSucceeds(t *testing.T) {
```
everettraven, line 96: Removed this test function because I don't think we should be using a state we never expect to be returned as part of our standard code paths in a mock to end reconciliation without an error but still have successful resolution. Removing the unused, outside of testing, unpack states resulted in making the mock unpacker return an error. We already had a test (immediately after this one I think) that did this same logic so I renamed it and removed this test.

everettraven: Note: the github diff view makes this look a tad wonky

internal/controllers/common_controller_test.go
```diff
@@ -0,0 +1,66 @@
...
+				Reason:  ocv1alpha1.ReasonBlocked,
+				Message: "boom",
+			},
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			setStatusProgressing(tc.clusterExtension, tc.err)
+			progressingCond := meta.FindStatusCondition(tc.clusterExtension.Status.Conditions, ocv1alpha1.TypeProgressing)
+			require.NotNil(t, progressingCond, "progressing condition should be set but was not")
+			diff := cmp.Diff(*progressingCond, tc.expected, cmpopts.IgnoreFields(metav1.Condition{}, "Message", "LastTransitionTime", "ObservedGeneration"))
```
everettraven, line 62: Need to remove "Message" from ignored fields

internal/controllers/common_controller_test.go
```diff
@@ -0,0 +1,66 @@
...
+			require.Empty(t, diff, "difference between actual and expected Progressing conditions", diff)
```
everettraven, line 63: This probably will already show the diff so I'll remove including it in the additional message output.

internal/rukpak/source/unpacker.go
```diff
@@ -59,15 +59,6 @@ type Result struct {
...
 const (
-	// StatePending conveys that a request for unpacking a bundle has been
-	// acknowledged, but not yet started.
-	StatePending State = "Pending"
-
-	// StateUnpacking conveys that the source is currently unpacking a bundle.
-	// This state should be used when the bundle contents are being downloaded
-	// and processed.
-	StateUnpacking State = "Unpacking"
-
```
everettraven, line 70: This is somewhat outside of the scope of the intent of this PR, but while I was identifying where to set the `Progressing` condition I noted that we don't ever expect these unpack states from being returned. I figured it is a small enough change that it made sense to include it as part of this PR.

internal/controllers/clusterextension_controller.go
```diff
@@ -264,20 +267,15 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
...
 		return ctrl.Result{}, err
 	}
 
 	switch unpackResult.State {
-	case rukpaksource.StatePending:
-		setStatusUnpackFailed(ext, unpackResult.Message)
-		ensureAllConditionsWithReason(ext, ocv1alpha1.ReasonFailed, "unpack pending")
-		return ctrl.Result{}, nil
 	case rukpaksource.StateUnpacked:
 		setStatusUnpacked(ext, unpackResult.Message)
```
skattoju, line 276: Should we setStatusProgressing here ? or just remove this completely (probably in the removed unpacked status pr) ?

everettraven: I would anticipate that we remove it completely in the unpacked status PR. There isn't really any value to setting it here since if there are no errors throughout reconciliation we set it to False, Succeeded

internal/controllers/clusterextension_controller_test.go
```diff
@@ -244,6 +172,11 @@ func TestClusterExtensionUnpackFails(t *testing.T) {
 	require.Equal(t, metav1.ConditionFalse, unpackedCond.Status)
 	require.Equal(t, ocv1alpha1.ReasonFailed, unpackedCond.Reason)
 
+	progressingCond := apimeta.FindStatusCondition(clusterExtension.Status.Conditions, ocv1alpha1.TypeProgressing)
+	require.NotNil(t, progressingCond)
+	require.Equal(t, metav1.ConditionTrue, progressingCond.Status)
+	require.Equal(t, ocv1alpha1.ReasonRetrying, progressingCond.Reason)
+
```
joelanford, line 179: Since we expected the `Progressing` condition message to be the place that contains details of the resolution, and there will eventually be no other place with that information, I think we should have some test assertions that look for <name>, <version>, and maybe <resolvedImageRef> in the message.

internal/controllers/common_controller.go
```diff
@@ -98,3 +101,26 @@ func setResolutionStatus(ext *ocv1alpha1.ClusterExtension, resStatus *ocv1alpha1
...
+		Status:             metav1.ConditionFalse,
+		Reason:             ocv1alpha1.ReasonSuccess,
+		Message:            "desired state reached",
+		ObservedGeneration: ext.GetGeneration(),
+	}
+
+	if err != nil {
+		progressingCond.Status = metav1.ConditionTrue
+		progressingCond.Reason = ocv1alpha1.ReasonRetrying
+		progressingCond.Message = err.Error()
```
joelanford, line 117: Should we go ahead and add the name, version, and resolved image reference of the bundle here?

That way, the PR that removes the `status.resolved` field can literally just remove the field and not have to remember to come here and update the message.

internal/controllers/clusterextension_controller_test.go
everettraven: I think this is reasonable

internal/controllers/common_controller.go
everettraven: I would expect the string we receive from the `err.Error()` call to return a string that contains this information + the error encountered. I can update the errors we send to this `setStatusProgressing` error to have that information.

internal/controllers/clusterextension_controller_test.go
everettraven: should be updated in the latest push

internal/controllers/clusterextension_controller_test.go
```diff
@@ -583,6 +422,12 @@ func TestClusterExtensionManagerFailed(t *testing.T) {
 	require.Equal(t, metav1.ConditionTrue, installedCond.Status)
 	require.Equal(t, ocv1alpha1.ReasonSuccess, installedCond.Reason)
 
+	t.Log("By checking the expected progressing conditions")
+	progressingCond := apimeta.FindStatusCondition(clusterExtension.Status.Conditions, ocv1alpha1.TypeProgressing)
+	require.NotNil(t, progressingCond)
+	require.Equal(t, metav1.ConditionTrue, progressingCond.Status)
+	require.Equal(t, ocv1alpha1.ReasonRetrying, progressingCond.Reason)
```
joelanford, line 429: Check for the resolution message name and version here too?

joelanford (approved):

PR #1303, opened 2024-09-24
✨ Use creds if present for pulling bundle images

anik120:
# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -26,6 +26,7 @@ import (
 	"time"
 
 	"github.com/containers/image/v5/types"
+	"github.com/go-logr/logr"
 	"github.com/spf13/pflag"
 	"go.uber.org/zap/zapcore"
 	apiextensionsv1client "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1"
@@ -66,6 +67,8 @@ var (
 	setupLog               = ctrl.Log.WithName("setup")
 	defaultSystemNamespace = "olmv1-system"
 )
+
+const authFilePath = "/etc/operator-controller/auth.json"
 
 // podNamespace checks whether the controller is running in a Pod vs.
 // being run locally by inspecting the namespace file that gets mounted
@@ -201,6 +204,7 @@ func main() {
 		SourceContext: &types.SystemContext{
 			DockerCertPath: caCertDir,
 			OCICertPath:    caCertDir,
+			AuthFilePath:   authFilePathIfPresent(setupLog),
 		},
 	}
 
@@ -310,4 +314,16 @@ type finalizerFunc func(ctx context.Context, obj client.Object) (crfinalizer.Res
 
 func (f finalizerFunc) Finalize(ctx context.Context, obj client.Object) (crfinalizer.Result, error) {
 	return f(ctx, obj)
+}
+
+func authFilePathIfPresent(logger logr.Logger) string {
+	_, err := os.Stat(authFilePath)
+	if os.IsNotExist(err) {
+		return ""
+	}
+	if err != nil {
+		logger.Error(err, "could not stat auth file path", "path", authFilePath)
+		os.Exit(1)
+	}
+	return authFilePath
 }


cmd/manager/main.go
```diff
@@ -204,6 +206,17 @@ func main() {
...
+
+	if err != nil && !os.IsNotExist(err) {
+		setupLog.Error(err, "unable to check for auth file")
+		os.Exit(1)
+	}
+
+	if !os.IsNotExist(err) {
+		unpacker.SourceContext.AuthFilePath = authFilePath
+	}
+
```
joelanford, line 219: I'd suggest putting this in a helper function. Something like:

```
func authFilePathIfPresent(log logr.Logger) string {
	_, err = os.Stat(authFilePath)
        if os.IsNotExist(err) {
                return ""
        }
        if err != nil {
                log.Error(err, "could not stat auth file path", "path", authFilePath)
                return ""
        }
        return path
}
```

Then you can directly call that function when originally initializing the `SourceContext` in lines 203-206.

cmd/manager/main.go
```diff
@@ -67,6 +67,8 @@ var (
 	defaultSystemNamespace = "olmv1-system"
 )
 
+const authFilePath = "/etc/containers/config.json"
```
joelanford, line 70:
```suggestion
const authFilePath = "/etc/operator-controller/auth.json"
```

cmd/manager/main.go
```diff
@@ -311,3 +315,15 @@ type finalizerFunc func(ctx context.Context, obj client.Object) (crfinalizer.Res
...
 }
+
+func authFilePathIfPresent(logger logr.Logger) string {
+	_, err := os.Stat(authFilePath)
+	if os.IsNotExist(err) {
+		return ""
+	}
+	if err != nil {
+		logger.Error(err, "could not stat auth file path", "path", authFilePath)
+		return ""
```
anik120, line 326: Are we okay with not panicking here if an unknown stat error occurs here?

joelanford: Yeah, you're probably right. Should we have this function return `(string, error)`?  That makes calling it slightly more verbose to call, but then we could do the error check and have the logging/os.Exit happen in main. And at that point, no need to pass a logger in.

anik120: I like the idea of keeping the signature the same and then `os.Exit(1)` instead of `return ""`. Updated the PR.

cmd/manager/main.go
```diff
@@ -311,3 +315,15 @@ type finalizerFunc func(ctx context.Context, obj client.Object) (crfinalizer.Res
 func (f finalizerFunc) Finalize(ctx context.Context, obj client.Object) (crfinalizer.Result, error) {
 	return f(ctx, obj)
 }
+
+func authFilePathIfPresent(logger logr.Logger) string {
```
everettraven, line 319: Nit: Could we name this something like `mustAuthFilePathIfPresent` or `authFilePathIfPresentOrDie`? I think that it would help someone reading through the code where this function is used to understand that if there is an error (unhandled error type in this case) that this function will exit the program.

anik120: I see where you're coming from. 
....

not sure if I would understand what `mustAuthFilePathIfPresent` would mean if I didn't know the context....

`authFilePathIfPresentOrDie` sounds like it must find an auth file or panic if it's not present, which is not the case. 

So I think this one's not that easy 😅

anik120: Actually, I could just leave comment on top that says unhandled error will cause is to exit with error code 1. Gets the job done.

anik120: While writing the comment I realized I have to basically summarize in English words every single thing that the function is doing, just to get to the part about the error code 1, which seems superfluous 😅 

```
// authFilePathIfPresent returns the value of the constant
// authFilePath ("/etc/catalogd/auth.json") if the file is
// present, emptry string ("") otherwise, using os.Stat(filePath).
// If os.Stat returns an error that's not IsNotExist, the function
// causes the program to exit with error code 1
```

everettraven: If there is no reasonable way we can come up with to convey this as part of the function name that is fine. I won't hold the PR on this nit, just thought it worthwhile to call out

joelanford (approved):

PR #1309, opened 2024-09-25
✨ Remove unpacked status

### LINKED ISSUE #1294: [v1 API Review] Remove the `Unpacked` status condition
The `Unpacked` status condition will be replaced by the `Progressing` condition implemented as part of #1292. Once it is implemented, we need to remove the `Unpacked` condition.

skattoju:
# Description
This PR removes the "Unpacked" status which has now been rolled into a "Progressing" status. 
Fixes #1294 

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/controllers/clusterextension_controller.go
+++ b/internal/controllers/clusterextension_controller.go
@@ -267,7 +267,6 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
 	l.V(1).Info("unpacking resolved bundle")
 	unpackResult, err := r.Unpacker.Unpack(ctx, bundleSource)
 	if err != nil {
-		setStatusUnpackFailed(ext, err.Error())
 		// Wrap the error passed to this with the resolution information until we have successfully
 		// installed since we intend for the progressing condition to replace the resolved condition
 		// and will be removing the .status.resolution field from the ClusterExtension status API
@@ -275,10 +274,7 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
 		return ctrl.Result{}, err
 	}
 
-	switch unpackResult.State {
-	case rukpaksource.StateUnpacked:
-		setStatusUnpacked(ext, unpackResult.Message)
-	default:
+	if unpackResult.State != rukpaksource.StateUnpacked {
 		panic(fmt.Sprintf("unexpected unpack state %q", unpackResult.State))
 	}
 
--- a/internal/controllers/clusterextension_controller_test.go
+++ b/internal/controllers/clusterextension_controller_test.go
@@ -167,12 +167,6 @@ func TestClusterExtensionResolutionSuccessfulUnpackFails(t *testing.T) {
 	require.Equal(t, ocv1alpha1.ReasonSuccess, resolvedCond.Reason)
 	require.Equal(t, "resolved to \"quay.io/operatorhubio/prometheus@fake1.0.0\"", resolvedCond.Message)
 
-	t.Log("By checking the expected unpacked conditions")
-	unpackedCond := apimeta.FindStatusCondition(clusterExtension.Status.Conditions, ocv1alpha1.TypeUnpacked)
-	require.NotNil(t, unpackedCond)
-	require.Equal(t, metav1.ConditionFalse, unpackedCond.Status)
-	require.Equal(t, ocv1alpha1.ReasonFailed, unpackedCond.Reason)
-
 	progressingCond := apimeta.FindStatusCondition(clusterExtension.Status.Conditions, ocv1alpha1.TypeProgressing)
 	require.NotNil(t, progressingCond)
 	require.Equal(t, metav1.ConditionTrue, progressingCond.Status)
@@ -233,9 +227,12 @@ func TestClusterExtensionUnpackUnexpectedState(t *testing.T) {
 			Image:   "quay.io/operatorhubio/prometheus@fake1.0.0",
 		}, &v, nil, nil
 	})
+
 	require.Panics(t, func() {
 		_, _ = reconciler.Reconcile(ctx, ctrl.Request{NamespacedName: extKey})
 	}, "reconciliation should panic on unknown unpack state")
+
+	require.NoError(t, cl.DeleteAllOf(ctx, &ocv1alpha1.ClusterExtension{}))
 }
 
 func TestClusterExtensionResolutionAndUnpackSuccessfulApplierFails(t *testing.T) {
@@ -311,12 +308,6 @@ func TestClusterExtensionResolutionAndUnpackSuccessfulApplierFails(t *testing.T)
 	require.Equal(t, metav1.ConditionTrue, resolvedCond.Status)
 	require.Equal(t, ocv1alpha1.ReasonSuccess, resolvedCond.Reason)
 	require.Equal(t, "resolved to \"quay.io/operatorhubio/prometheus@fake1.0.0\"", resolvedCond.Message)
-
-	t.Log("By checking the expected unpacked conditions")
-	unpackedCond := apimeta.FindStatusCondition(clusterExtension.Status.Conditions, ocv1alpha1.TypeUnpacked)
-	require.NotNil(t, unpackedCond)
-	require.Equal(t, metav1.ConditionTrue, unpackedCond.Status)
-	require.Equal(t, ocv1alpha1.ReasonSuccess, unpackedCond.Reason)
 
 	t.Log("By checking the expected installed conditions")
 	installedCond := apimeta.FindStatusCondition(clusterExtension.Status.Conditions, ocv1alpha1.TypeInstalled)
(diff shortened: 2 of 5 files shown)


internal/controllers/clusterextension_controller.go
```diff
@@ -267,21 +267,13 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
...
 		setStatusProgressing(ext, wrapErrorWithResolutionInfo(resolvedBundleMetadata, err))
 		return ctrl.Result{}, err
 	}
 
-	switch unpackResult.State {
-	case rukpaksource.StateUnpacked:
-		setStatusUnpacked(ext, unpackResult.Message)
-	default:
-		panic(fmt.Sprintf("unexpected unpack state %q", unpackResult.State))
-	}
```
everettraven, line 283: I'm not sure we should entirely remove the panic here because the API of `r.Unpacker.Unpack` doesn't have any guarantees a `nil` error returned means that the unpack was successful and will have the bundle data populated in the result returned. I would consider it out of scope for this PR to refactor the `Unpacker` API, but I would still want us to catch any programmer errors possible from the current state of the API contract.

Instead of removing the switch statement, I think changing it to something like this allows us to catch programmer errors:
```go
if unpackResult.State != rukpaksource.StateUnpacked {
	panic(fmt.Sprintf("unexpected unpack state %q", unpackResult.State))
}
```

internal/controllers/clusterextension_controller_test.go
```diff
@@ -312,11 +379,82 @@ func TestClusterExtensionResolutionAndUnpackSuccessfulApplierFails(t *testing.T)
...
 
-	t.Log("By checking the expected unpacked conditions")
-	unpackedCond := apimeta.FindStatusCondition(clusterExtension.Status.Conditions, ocv1alpha1.TypeUnpacked)
-	require.NotNil(t, unpackedCond)
-	require.Equal(t, metav1.ConditionTrue, unpackedCond.Status)
-	require.Equal(t, ocv1alpha1.ReasonSuccess, unpackedCond.Reason)
+	require.NoError(t, cl.DeleteAllOf(ctx, &ocv1alpha1.ClusterExtension{}))
+}
+
+func TestClusterExtensionInstallationFailedApplierFails(t *testing.T) {
```
everettraven, line 385: This function was intentionally removed in the progressing conditions PR

internal/controllers/clusterextension_controller_test.go
```diff
@@ -303,6 +369,7 @@ func TestClusterExtensionResolutionAndUnpackSuccessfulApplierFails(t *testing.T)
 	t.Log("By checking the status fields")
 	expectedBundleMetadata := ocv1alpha1.BundleMetadata{Name: "prometheus.v1.0.0", Version: "1.0.0"}
 	require.Equal(t, expectedBundleMetadata, clusterExtension.Status.Resolution.Bundle)
+	require.Equal(t, ocv1alpha1.BundleMetadata{Name: "prometheus.v1.0.0", Version: "1.0.0"}, clusterExtension.Status.Resolution.Bundle)
```
everettraven, line 372: This line seems like a duplicate of the one above it, just instead it is manually building the bundle metadata instead of using the variable on line 370

internal/controllers/clusterextension_controller_test.go
```diff
@@ -138,6 +138,75 @@ func TestClusterExtensionResolutionSuccessfulUnpackFails(t *testing.T) {
...
+	resolvedCond := apimeta.FindStatusCondition(clusterExtension.Status.Conditions, ocv1alpha1.TypeResolved)
+	require.NotNil(t, resolvedCond)
+	require.Equal(t, metav1.ConditionTrue, resolvedCond.Status)
+	require.Equal(t, ocv1alpha1.ReasonSuccess, resolvedCond.Reason)
+	require.Equal(t, "resolved to \"quay.io/operatorhubio/prometheus@fake1.0.0\"", resolvedCond.Message)
+
+	require.NoError(t, cl.DeleteAllOf(ctx, &ocv1alpha1.ClusterExtension{}))
+}
+
+func TestClusterExtensionUnpackFails(t *testing.T) {
```
everettraven, line 171: This test function was intentionally removed as part of the progressing status condition PR

skattoju: will remove it

internal/controllers/clusterextension_controller_test.go
skattoju: will remove it

internal/controllers/clusterextension_controller_test.go
skattoju: will remove it

internal/controllers/clusterextension_controller.go
skattoju: yeah should have figured the default logic stays will update

internal/controllers/clusterextension_controller_test.go
```diff
@@ -148,16 +148,14 @@ func TestClusterExtensionResolutionSuccessfulUnpackFails(t *testing.T) {
...
 	require.Equal(t, ctrl.Result{}, res)
-	require.Error(t, err)
 
 	t.Log("By fetching updated cluster extension after reconcile")
 	require.NoError(t, cl.Get(ctx, extKey, clusterExtension))
 
 	t.Log("By checking the status fields")
-	expectedBundleMetadata := ocv1alpha1.BundleMetadata{Name: "prometheus.v1.0.0", Version: "1.0.0"}
-	require.Equal(t, expectedBundleMetadata, clusterExtension.Status.Resolution.Bundle)
+	require.Equal(t, ocv1alpha1.BundleMetadata{Name: "prometheus.v1.0.0", Version: "1.0.0"}, clusterExtension.Status.Resolution.Bundle)
```
everettraven, line 158: I would do the inverse here (keep what is being removed, get rid of what is being added)

internal/controllers/clusterextension_controller_test.go
```diff
@@ -148,16 +148,14 @@ func TestClusterExtensionResolutionSuccessfulUnpackFails(t *testing.T) {
 			Image:   "quay.io/operatorhubio/prometheus@fake1.0.0",
 		}, &v, nil, nil
 	})
-	res, err := reconciler.Reconcile(ctx, ctrl.Request{NamespacedName: extKey})
+	res, _ := reconciler.Reconcile(ctx, ctrl.Request{NamespacedName: extKey})
 	require.Equal(t, ctrl.Result{}, res)
-	require.Error(t, err)
```
everettraven, line 153: Why are we removing the error check here? I would expect if the unpack process fails and returns an error then reconcile would return an error.

internal/controllers/clusterextension_controller_test.go
```diff
@@ -312,25 +301,6 @@ func TestClusterExtensionResolutionAndUnpackSuccessfulApplierFails(t *testing.T)
...
-	require.NotNil(t, installedCond)
-	require.Equal(t, metav1.ConditionFalse, installedCond.Status)
-	require.Equal(t, ocv1alpha1.ReasonFailed, installedCond.Reason)
-
-	t.Log("By checking the expected progressing conditions")
-	progressingCond := apimeta.FindStatusCondition(clusterExtension.Status.Conditions, ocv1alpha1.TypeProgressing)
-	require.NotNil(t, progressingCond)
-	require.Equal(t, metav1.ConditionTrue, progressingCond.Status)
-	require.Equal(t, ocv1alpha1.ReasonRetrying, progressingCond.Reason)
-	require.Contains(t, progressingCond.Message, fmt.Sprintf("for resolved bundle %q with version %q", expectedBundleMetadata.Name, expectedBundleMetadata.Version))
```
everettraven, line 332: Installed and Progressing condition checks are ones I would expect to stay

internal/controllers/clusterextension_controller_test.go
skattoju: Thought it was a lint error when in fact it was a rebase error. But fixed now.

internal/controllers/clusterextension_controller_test.go
skattoju: fixed 👍

internal/controllers/clusterextension_controller_test.go
```diff
@@ -167,18 +165,6 @@ func TestClusterExtensionResolutionSuccessfulUnpackFails(t *testing.T) {
...
-	unpackedCond := apimeta.FindStatusCondition(clusterExtension.Status.Conditions, ocv1alpha1.TypeUnpacked)
-	require.NotNil(t, unpackedCond)
-	require.Equal(t, metav1.ConditionFalse, unpackedCond.Status)
-	require.Equal(t, ocv1alpha1.ReasonFailed, unpackedCond.Reason)
-
-	progressingCond := apimeta.FindStatusCondition(clusterExtension.Status.Conditions, ocv1alpha1.TypeProgressing)
-	require.NotNil(t, progressingCond)
-	require.Equal(t, metav1.ConditionTrue, progressingCond.Status)
-	require.Equal(t, ocv1alpha1.ReasonRetrying, progressingCond.Reason)
-	require.Contains(t, progressingCond.Message, fmt.Sprintf("for resolved bundle %q with version %q", expectedBundleMetadata.Name, expectedBundleMetadata.Version))
```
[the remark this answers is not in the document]
skattoju: yup yup bad rebase 😅

internal/controllers/clusterextension_controller_test.go
skattoju: updated 👍

everettraven (approved):
Changes look good to me! Thanks @skattoju !

CI also looks green except for go-apidiff which has the expected failure

skattoju:
There is an e2e failure in the merge queue but it passes locally 🤔 and its not in a function that had an unpacked condition removed.. specifically `TestClusterExtensionBlockInstallNonSuccessorVersion` It might be intermittent ...

PR #1312, opened 2024-09-26
✨ Remove the Resolved status condition

yashoza19:
# Description

Fixes #1293 

## Reviewer Checklist

- [x] API Go Documentation
- [x] Tests: Unit Tests (and E2E Tests, if appropriate)
- [x] Comprehensive Commit Messages
- [x] Links to related GitHub Issue(s)

PR #1314, opened 2024-09-26
✨ Update Installed status condition handling

### LINKED ISSUE #1291: [v1 API Review] Update `Installed` Status condition usage
The `Installed` status condition represents whether or not a bundle is installed. It should be set to `True` once the initial installation is successful. If it was set to `True`, it should transition to `False`, or be removed, **ONLY IF** all workloads associated with the ClusterExtension have been removed from the cluster. When unset, this condition should be interpreted the same as "False". Successful upgrades should update this condition and other status fields with the appropriate values to reflect the state of the currently installed bundle.

**Acceptance Criteria**:
- The logic for setting the `Installed` status condition is updated to:
    - Set the `Installed` status condition to `True` and `status.install.bundle` field with installed bundle information after initial installation is successful
    - Update the `Installed` status condition and `status.install.bundle` field to reflect the currently installed bundle information after a successful upgrade
    - Set the `Installed` status condition to `False` if there are any failures encountered during the initial installation attempt
    - Only transition from `True` to `False` if all workloads associated with the ClusterExtension have been removed from the cluster (having not looked into it further, this may mean we never perform this transition)
    
trgeiger:
# Description
Updates Installed status condition per status condition RFC. Closes #1291.

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/controllers/clusterextension_controller.go
+++ b/internal/controllers/clusterextension_controller.go
@@ -192,15 +192,9 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
 	l.Info("handling finalizers")
 	finalizeResult, err := r.Finalizers.Finalize(ctx, ext)
 	if err != nil {
-		// TODO: For now, this error handling follows the pattern of other error handling.
-		//  Namely: zero just about everything out, throw our hands up, and return an error.
-		//  This is not ideal, and we should consider a more nuanced approach that resolves
-		//  as much status as possible before returning, or at least keeps previous state if
-		//  it is properly labeled with its observed generation.
-		setInstallStatus(ext, nil)
 		setResolutionStatus(ext, nil)
 		setStatusProgressing(ext, err)
-		ensureAllConditionsWithReason(ext, ocv1alpha1.ReasonFailed, err.Error())
+
 		return ctrl.Result{}, err
 	}
 	if finalizeResult.Updated || finalizeResult.StatusUpdated {
@@ -297,8 +291,11 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
 	//     The only way to eventually recover from permission errors is to keep retrying).
 	managedObjs, _, err := r.Applier.Apply(ctx, unpackResult.Bundle, ext, objLbls, storeLbls)
 	if err != nil {
-		setInstalledStatusConditionFailed(ext, err.Error())
 		setStatusProgressing(ext, wrapErrorWithResolutionInfo(resolvedBundleMetadata, err))
+		// If bundle is not already installed, set Installed status condition to False
+		if installedBundle == nil {
+			setInstalledStatusConditionFailed(ext, err.Error())
+		}
 		return ctrl.Result{}, err
 	}
 
--- a/internal/controllers/clusterextension_controller_test.go
+++ b/internal/controllers/clusterextension_controller_test.go
@@ -18,6 +18,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/rand"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	crfinalizer "sigs.k8s.io/controller-runtime/pkg/finalizer"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	"github.com/operator-framework/operator-registry/alpha/declcfg"
@@ -25,6 +26,7 @@ import (
 	ocv1alpha1 "github.com/operator-framework/operator-controller/api/v1alpha1"
 	"github.com/operator-framework/operator-controller/internal/conditionsets"
 	"github.com/operator-framework/operator-controller/internal/controllers"
+	"github.com/operator-framework/operator-controller/internal/finalizers"
 	"github.com/operator-framework/operator-controller/internal/resolve"
 	"github.com/operator-framework/operator-controller/internal/rukpak/source"
 )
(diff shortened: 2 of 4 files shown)


internal/controllers/clusterextension_controller.go
```diff
@@ -193,16 +193,17 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
...
 	finalizeResult, err := r.Finalizers.Finalize(ctx, ext)
 	if err != nil {
-		// TODO: For now, this error handling follows the pattern of other error handling.
-		//  Namely: zero just about everything out, throw our hands up, and return an error.
-		//  This is not ideal, and we should consider a more nuanced approach that resolves
-		//  as much status as possible before returning, or at least keeps previous state if
-		//  it is properly labeled with its observed generation.
-		setInstallStatus(ext, nil)
 		setResolutionStatus(ext, nil)
 		setResolvedStatusConditionFailed(ext, err.Error())
```
everettraven, line 197: Missing a rebase? https://github.com/operator-framework/operator-controller/pull/1312 merged that, from what I can tell, removed the `Resolved` status condition all together.

internal/controllers/clusterextension_controller.go
```diff
@@ -193,16 +193,17 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
...
 		setStatusProgressing(ext, err)
-		ensureAllConditionsWithReason(ext, ocv1alpha1.ReasonFailed, err.Error())
+
+		// Here we check if the bundle is already installed.
+		// We only set Installed status to False if the bundle is actually no longer on cluster.
+		installedBundle, _ := r.InstalledBundleGetter.GetInstalledBundle(ctx, ext)
+		if installedBundle == nil {
+			setInstallStatus(ext, nil)
+			setInstalledStatusConditionFailed(ext, err.Error())
```
everettraven, line 205: I know we talked about this about during our sync yesterday, but wanted to pose the question here as well in case other maintainers have opinions on it:

Do we need to set the Installed condition to False here and overwrite any errors that may exist in it with the finalizer error?

My feeling is that we do not need to do this as a user may want to still be able to see the original failed installation message. We already set the Progressing condition with the finalizer error so I think that should be sufficient here.

Additionally, if the Installed condition did not previously exist, I don't think there is any reason to set it here. Non-existent Installed condition should be interpreted as "Unknown".

Maybe a more fundamental question (that popped into my head as I was typing ^) - if there is no reason for us to change any of the status related to installation of the extension can we just remove any logic to do with the installed status from the finalizer error handling?

internal/controllers/clusterextension_controller_test.go
```diff
@@ -325,6 +327,112 @@ func TestClusterExtensionResolutionAndUnpackSuccessfulApplierFails(t *testing.T)
...
+	expectedBundleMetadata := ocv1alpha1.BundleMetadata{Name: "prometheus.v1.0.0", Version: "1.0.0"}
+	require.Equal(t, expectedBundleMetadata, clusterExtension.Status.Resolution.Bundle)
+	require.Equal(t, expectedBundleMetadata, clusterExtension.Status.Install.Bundle)
+
+	t.Log("By checking the expected resolution conditions")
+	resolvedCond := apimeta.FindStatusCondition(clusterExtension.Status.Conditions, ocv1alpha1.TypeResolved)
+	require.NotNil(t, resolvedCond)
+	require.Equal(t, metav1.ConditionTrue, resolvedCond.Status)
+	require.Equal(t, ocv1alpha1.ReasonSuccess, resolvedCond.Reason)
+	require.Equal(t, "resolved to \"quay.io/operatorhubio/prometheus@fake1.0.0\"", resolvedCond.Message)
```
everettraven, line 418: Resolved condition removed in https://github.com/operator-framework/operator-controller/pull/1312

internal/controllers/clusterextension_controller_test.go
```diff
@@ -601,6 +709,101 @@ func TestClusterExtensionInstallationSucceeds(t *testing.T) {
...
+		bundle: &ocv1alpha1.BundleMetadata{Name: "prometheus.v1.0.0", Version: "1.0.0"},
+	}
+	err = reconciler.Finalizers.Register("fake.testfinalizer.io", finalizerFunc(func(ctx context.Context, obj client.Object) (crfinalizer.Result, error) {
+		return crfinalizer.Result{}, errors.New("still have finalizers")
+	}))
+	require.NoError(t, err)
+
+	res, err := reconciler.Reconcile(ctx, ctrl.Request{NamespacedName: extKey})
+	require.Equal(t, ctrl.Result{}, res)
+	require.NoError(t, err)
```
everettraven, line 779: Maybe this is unnecessary since it might test the finalizer library logic, but should we add a check after this first reconciliation that causes the test to fail early if the finalizer doesn't exist on the ClusterExtension like we expect?

internal/controllers/clusterextension_controller_test.go
```diff
@@ -601,6 +709,101 @@ func TestClusterExtensionInstallationSucceeds(t *testing.T) {
...
+
+	require.NoError(t, cl.DeleteAllOf(ctx, &ocv1alpha1.ClusterExtension{}))
+	res, err = reconciler.Reconcile(ctx, ctrl.Request{NamespacedName: extKey})
+	require.Error(t, err, res)
+
+	t.Log("By fetching updated cluster extension after reconcile")
+	require.NoError(t, cl.Get(ctx, extKey, clusterExtension))
+	cond = apimeta.FindStatusCondition(clusterExtension.Status.Conditions, ocv1alpha1.TypeInstalled)
+	require.NotNil(t, cond)
+	require.Equal(t, metav1.ConditionTrue, cond.Status)
```
everettraven, line 798: We should probably also check that the `.status.install` field contains the information we expect

internal/controllers/clusterextension_controller_test.go
```diff
@@ -601,6 +709,101 @@ func TestClusterExtensionInstallationSucceeds(t *testing.T) {
...
+}
+
+type finalizerFunc func(ctx context.Context, obj client.Object) (crfinalizer.Result, error)
+
+func (f finalizerFunc) Finalize(ctx context.Context, obj client.Object) (crfinalizer.Result, error) {
+	return f(ctx, obj)
+}
```
everettraven, line 805: To reduce duplication of this code, since this is taken from main.go, we should probably put this logic in a separate "finalizers" pkg that we import where we need it.

internal/controllers/clusterextension_controller.go
```diff
@@ -193,15 +193,16 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
...
+		}
```
everettraven, line 205: Circling back around to this after spending some time doing some more thinking on this. I can't think of a scenario where we would want to actually modify any of the information in the status related to installation on finalizer issues, so I think we should just remove all logic related to install status here:
```suggestion

```

My thinking for context:
- If we have successfully installed the ClusterExtension before, we do nothing
- If we have not successfully installed the ClusterExtension before:
    - We should preserve installation failure messages if they are set
    - If we failed before ever setting the installed condition and this ClusterExtension is being deleted, there really isn't any point to populate the empty install status field+condition.

everettraven: Looks like this change didn't actually make it into the PR? I still see the logic that I think we could remove.

trgeiger: weird, maybe i accidentally undid it before I pushed. thanks for spotting it.

internal/controllers/clusterextension_controller_test.go
```diff
@@ -591,6 +692,103 @@ func TestClusterExtensionInstallationSucceeds(t *testing.T) {
...
+	require.NoError(t, err)
+
+	t.Log("By fetching updated cluster extension after first reconcile")
+	require.NoError(t, cl.Get(ctx, extKey, clusterExtension))
+	cond := apimeta.FindStatusCondition(clusterExtension.Status.Conditions, ocv1alpha1.TypeInstalled)
+	expectedBundleMetadata := ocv1alpha1.BundleMetadata{Name: "prometheus.v1.0.0", Version: "1.0.0"}
+	require.Equal(t, expectedBundleMetadata, clusterExtension.Status.Install.Bundle)
+	require.NotNil(t, cond)
+	require.Equal(t, metav1.ConditionTrue, cond.Status)
+	require.Equal(t, fakeFinalizer, clusterExtension.Finalizers[0])
```
everettraven, line 777: I think this might need to move up above line 766 as this is currently being called after the second reconcile

internal/controllers/clusterextension_controller_test.go
```diff
@@ -591,6 +692,103 @@ func TestClusterExtensionInstallationSucceeds(t *testing.T) {
...
+	res, err = reconciler.Reconcile(ctx, ctrl.Request{NamespacedName: extKey})
+	require.Error(t, err, res)
+
+	t.Log("By fetching updated cluster extension after second reconcile")
+	require.NoError(t, cl.Get(ctx, extKey, clusterExtension))
+	cond = apimeta.FindStatusCondition(clusterExtension.Status.Conditions, ocv1alpha1.TypeInstalled)
+	require.Equal(t, expectedBundleMetadata, clusterExtension.Status.Install.Bundle)
+	require.NotNil(t, cond)
+	require.Equal(t, metav1.ConditionTrue, cond.Status)
+	require.Equal(t, fakeFinalizer, clusterExtension.Finalizers[0])
```
everettraven, line 789: Should we add a check that the progressing status condition is being set appropriately as well (i.e with the finalizer error)?

everettraven (approved):

PR #1317, opened 2024-09-27
✨main.go: switch to klog-based logger

joelanford:
Use standard klog now that klog works with controller-runtime's expectation of a logr.Logger interface. 

We get a standard --v flag and we are more in line with k8s standards. 

We also get client-go and other low-level logging from k8s libraries at the higher verbosity levels which could be immensely helpful for debugging.

See similar PR in catalogd: https://github.com/operator-framework/catalogd/pull/419

# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/config/base/manager/manager.yaml
+++ b/config/base/manager/manager.yaml
@@ -90,12 +90,12 @@ spec:
           capabilities:
             drop:
               - "ALL"
-        image: gcr.io/kubebuilder/kube-rbac-proxy:v0.13.1
+        image: gcr.io/kubebuilder/kube-rbac-proxy:v0.15.0
         args:
-        - "--secure-listen-address=0.0.0.0:8443"
-        - "--upstream=http://127.0.0.1:8080/"
-        - "--logtostderr=true"
-        - "--v=0"
+        - --secure-listen-address=0.0.0.0:8443
+        - --http2-disable
+        - --upstream=http://127.0.0.1:8080/
+        - --logtostderr=true
         ports:
         - containerPort: 8443
           protocol: TCP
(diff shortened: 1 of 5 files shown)


config/base/manager/manager.yaml
```diff
@@ -90,12 +90,12 @@ spec:
...
+        image: gcr.io/kubebuilder/kube-rbac-proxy:v0.15.0
         args:
-        - "--secure-listen-address=0.0.0.0:8443"
-        - "--upstream=http://127.0.0.1:8080/"
-        - "--logtostderr=true"
-        - "--v=0"
+        - --secure-listen-address=0.0.0.0:8443
+        - --http2-disable
+        - --upstream=http://127.0.0.1:8080/
+        - --logtostderr=true
```
joelanford, line 98: Making arguments for krp look the same as catalogd (which also requires bumping krp version).

I'm doing this to ensure consistency in the contents and ordering of arguments between catalogd and operator controller so that vendors have an easier time manipulating flags for their specific use cases.

internal/controllers/clusterextension_controller.go
```diff
@@ -101,8 +101,8 @@ func (r *ClusterExtensionReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	l := log.FromContext(ctx).WithName("operator-controller")
 	ctx = log.IntoContext(ctx, l)
 
-	l.V(1).Info("reconcile starting")
-	defer l.V(1).Info("reconcile ending")
+	l.Info("reconcile starting")
+	defer l.Info("reconcile ending")
```
joelanford, line 105: With the original zap logger `l.V(1).Info` was being output by default.

With klogr, we need to put these log messages at a lower verbosity level to keep them coming by default.

joelanford:
As a test, I manually set `--v=8` on the controller and verified that low-level client-go logs were being emitted. 🎉

everettraven (approved):

PR #1318, opened 2024-09-29
🌱 Refactor `catalogmetadata` client and cache

m1kola:
# Description

It is now possible to cache errors occured on the cache population. This will be useful when we start proactively populating cache from `ClusterCatalogReconciler` instead of doing this on-demand.

This refactoring also moves the responsibility of performing network requests out of the cache backend into the client.

This is done in preparation for #1284

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/catalogmetadata/cache/cache.go
+++ b/internal/catalogmetadata/cache/cache.go
@@ -1,35 +1,24 @@
 package cache
 
 import (
-	"context"
 	"fmt"
+	"io"
 	"io/fs"
-	"net/http"
 	"os"
 	"path/filepath"
 	"sync"
 
-	catalogd "github.com/operator-framework/catalogd/api/core/v1alpha1"
 	"github.com/operator-framework/operator-registry/alpha/declcfg"
 
 	"github.com/operator-framework/operator-controller/internal/catalogmetadata/client"
 )
 
-var _ client.Fetcher = &filesystemCache{}
+var _ client.Cache = &filesystemCache{}
 
-// NewFilesystemCache returns a client.Fetcher implementation that uses a
-// local filesystem to cache Catalog contents. When fetching the Catalog contents
-// it will:
-// - Check if the Catalog is cached
-//   - IF !cached it will fetch from the catalogd HTTP server and cache the response
-//   - IF cached it will verify the cache is up to date. If it is up to date it will return
-//     the cached contents, if not it will fetch the new contents from the catalogd HTTP
-//     server and update the cached contents.
-func NewFilesystemCache(cachePath string, clientFunc func() (*http.Client, error)) *filesystemCache {
+func NewFilesystemCache(cachePath string) *filesystemCache {
 	return &filesystemCache{
 		cachePath:              cachePath,
 		mutex:                  sync.RWMutex{},
-		getClient:              clientFunc,
 		cacheDataByCatalogName: map[string]cacheData{},
 	}
 }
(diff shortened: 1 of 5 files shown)


internal/catalogmetadata/cache/cache_test.go
```diff
@@ -58,291 +52,113 @@ const (
...
+	assert.NoError(t, equalFilesystems(actualFSPut, actualFSGet))
+
+	t.Log("Put v1 error into cache")
+	actualFSPut, err = c.Put(catalogName, resolvedRef1, nil, errors.New("fake put error"))
+	// Errors do not override previously successfully populated cache
+	assert.NoError(t, err)
+	assert.NoError(t, equalFilesystems(defaultFS(), actualFSPut))
+	assert.NoError(t, equalFilesystems(actualFSPut, actualFSGet))
+
+	t.Log("Put v2 error into cache")
```
everettraven, line 103: Should we add a test after populating something for a newly resolved ref that doing a GET for the old resolved ref returns a nil cache and nil error (IIUC that is the intended behavior)?

everettraven (dismissed):
Had a comment about potentially adding another test case, but overall this refactor looks really good!

internal/catalogmetadata/cache/cache_test.go
m1kola: Actually you are right - "Get v1 from cache" will return `nil` because we map by catalog name and override on `Put`. This is a bug: I want `Remove` to work even if we do a `Put` of another version before `Remove`.

I think #1207 (and `main`) has the same issue...

internal/catalogmetadata/cache/cache.go
```diff
@@ -117,19 +65,35 @@ func (fsc *filesystemCache) FetchCatalogContents(ctx context.Context, catalog *c
...
-		if data.ResolvedRef == catalog.Status.ResolvedSource.Image.ResolvedRef {
-			return os.DirFS(cacheDir), nil
-		}
+	// the cached contents.
+	if cache, err := fsc.get(catalogName, resolvedRef); err == nil && cache != nil {
+		// We only return here if the was no error during
+		// the previous (likely concurrent) cache population attempt.
+		// If there was an error - we want to try and populate the cache again.
+		return cache, nil
+	}
```
everettraven, line 74: Apologies for missing this the first time around, but do we actually still need this?

I think this was put in place with more complex thread safety logic than we require now since we are doing thread-safe "Put" and "Get" operations. Refreshing my memory on the previous logic, it was done for the specific scenario of:
- We do a read-only lock to verify if the content has been cached before
- We release the read-only lock once we have verified that either we have or have not cached this content before
- Some other thread could now potentially get a write lock and write the contents for this catalog to the cache
- When we go to write content for the catalog to the cache, we get a write lock and re-verify whether or not it was written by some other process. We let it win because we don't know whether our revision of the content is newer or not.

I don't think the same scenario is possible _here_ since we are doing a lock for the entire duration of the `Put` method and as such we can probably discard this initial get and ensure `Put` has the single responsibility of _always_ writing the new cache.

everettraven: That being said, it may still be valid to say "we have successfully cached this revision before so we will ignore a failed put for the same revision". I would lean more towards putting the responsibility of this verification to the caller though to simplify the responsibility of these cache functions.

m1kola: @everettraven it is still possible that two callers still do the following concurrently:

1. Get cache - it is missing - we lock for reading
2. Do an HTTP request - we do not lock this
3. Write into FS - we lock this

So if this happens we have and we remove the get from `Put` we have two redundant things: 1) network request 2) FS IO.

So if we remove this check we are risking two sequential writes to FS which will make reads blocked as well. With this check we ensure that only one FS write happens even if there was multiple concurrent attempts to fetch cache.

everettraven (approved):
Had one comment about the responsibility of the `Put` method, but I don't think that it should hold this PR because:
- It is an internal package, we can modify it as much as we want
- It isn't any worse than what we had before

internal/catalogmetadata/cache/cache.go
everettraven: Right. I get that it would still be possible for that scenario to occur, but I'm questioning whether it should be the responsibility of the `Put` method to handle that or the caller. I would think the caller should take that into consideration as opposed to the `Put` method.

m1kola: @everettraven if we make this responsibility of the caller then the caller will have to be responsible for locking of the mutex as well. And we probably don't want to make locking the caller's responsibility.

Or am I still not getting the idea?

Issue #1015, opened 2024-07-05
[epic] Support installing bundles from a private registry


everettraven:
Currently, there is no support for creating a `ClusterExtension` that can install a bundle that is in a private registry and requires authentication. This epic is scoped for enabling that support. There are a couple different approaches that should be supported:
- A global pull secret
    - Note: this should also be done in Catalogd, which currently has support for direct references to a `Secret`
- A direct reference to a `Secret`

Brief: https://docs.google.com/document/d/14abNwdBEe6bGaLczGIlgYR9ghz6gbp_qene_7znIa8o/edit

RFC: https://docs.google.com/document/d/1BXD6kj5zXHcGiqvJOikU2xs8kV26TPnzEKp6n7TKD4M/edit

bentito:
Can the ServiceAccount be patched to attach the cluster pull secret?

everettraven:
>Can the ServiceAccount be patched to attach the cluster pull secret?

@bentito Could you elaborate on what you mean by this?

bentito:
The user/admin would need to do this:

```bash
kubectl patch serviceaccount default -p '{"imagePullSecrets": [{"name": "my-pull-secret"}]}'
```
and then the code would need to reference it, for example on Pod creation:

```go
  // Define a new Pod object
  pod := &corev1.Pod{
    ObjectMeta: metav1.ObjectMeta{
      Name:      "demo-pod",
      Namespace: req.Namespace,
    },
    Spec: corev1.PodSpec{
      Containers: []corev1.Container{
        {
          Name:  "demo-container",
          Image: "my-registry.example.com/my-image:latest",
        },
      },
      ServiceAccountName: "default", // Ensure this ServiceAccount has the pull secret
    },
  }
```

Benefits of blessing the SA with the pull secret like this are 1) it's centralized for any usage (Pod, Deployment, etc); 2) Just one place to go to for creds rotation.

everettraven:
@bentito thanks for more info on that. I forgot that in catalogd's unpacking implementation for image sources we have the ServiceAccount configured secrets disabled for pull secrets:
https://github.com/operator-framework/catalogd/blob/e6da1538304ec2af0648c918bff127632c4d0d0e/internal/source/image_registry_client.go#L58-L61

I think we should enable them for both catalogd and operator-controller (o-c should follow a similar pattern to catalogd)

LalatenduMohanty:
[...]

PR #1322, opened 2024-09-30
✨ Add PullSecret controller to save pull secret data locally

anik120:
RFC: https://docs.google.com/document/d/1BXD6kj5zXHcGiqvJOikU2xs8kV26TPnzEKp6n7TKD4M/edit#heading=h.x3tfh25grvnv

# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -151,12 +151,14 @@ func main() {
 	}
 	if globalPullSecretKey != nil {
 		cacheOptions.ByObject[&corev1.Secret{}] = crcache.ByObject{
-			Field: fields.SelectorFromSet(map[string]string{
-				"metadata.name": globalPullSecretKey.Name,
-			}),
-		}
-		cacheOptions.DefaultNamespaces[globalPullSecretKey.Namespace] = crcache.Config{
-			LabelSelector: k8slabels.Everything(),
+			Namespaces: map[string]crcache.Config{
+				globalPullSecretKey.Namespace: {
+					LabelSelector: k8slabels.Everything(),
+					FieldSelector: fields.SelectorFromSet(map[string]string{
+						"metadata.name": globalPullSecretKey.Name,
+					}),
+				},
+			},
 		}
 	}
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -23,13 +23,18 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/containers/image/v5/types"
 	"github.com/go-logr/logr"
 	"github.com/spf13/pflag"
+	corev1 "k8s.io/api/core/v1"
 	apiextensionsv1client "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1"
+	"k8s.io/apimachinery/pkg/fields"
 	k8slabels "k8s.io/apimachinery/pkg/labels"
+	k8stypes "k8s.io/apimachinery/pkg/types"
+	apimachineryrand "k8s.io/apimachinery/pkg/util/rand"
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 	"k8s.io/klog/v2"
@@ -67,7 +72,7 @@ var (
 	defaultSystemNamespace = "olmv1-system"
 )
 
-const authFilePath = "/etc/operator-controller/auth.json"
+const authFilePrefix = "operator-controller-global-pull-secrets"
 
 // podNamespace checks whether the controller is running in a Pod vs.
 // being run locally by inspecting the namespace file that gets mounted
@@ -90,6 +95,7 @@ func main() {
 		operatorControllerVersion bool
 		systemNamespace           string
 		caCertDir                 string
+		globalPullSecret          string
 	)
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
@@ -100,6 +106,7 @@ func main() {
 	flag.StringVar(&cachePath, "cache-path", "/var/cache", "The local directory path used for filesystem based caching")
 	flag.BoolVar(&operatorControllerVersion, "version", false, "Prints operator-controller version information")
 	flag.StringVar(&systemNamespace, "system-namespace", "", "Configures the namespace that gets used to deploy system resources.")
+	flag.StringVar(&globalPullSecret, "global-pull-secret", "", "The <namespace>/<name> of the global pull secret that is going to be used to pull bundle images.")
 
 	klog.InitFlags(flag.CommandLine)
 
@@ -116,27 +123,49 @@ func main() {
 
 	setupLog.Info("starting up the controller", "version info", version.String())
 
+	authFilePath := filepath.Join(os.TempDir(), fmt.Sprintf("%s-%s.json", authFilePrefix, apimachineryrand.String(8)))
+	var globalPullSecretKey *k8stypes.NamespacedName
+	if globalPullSecret != "" {
+		secretParts := strings.Split(globalPullSecret, "/")
+		if len(secretParts) != 2 {
+			setupLog.Error(fmt.Errorf("incorrect number of components"), "value of global-pull-secret should be of the format <namespace>/<name>")
+			os.Exit(1)
+		}
+		globalPullSecretKey = &k8stypes.NamespacedName{Name: secretParts[1], Namespace: secretParts[0]}
+	}
+
 	if systemNamespace == "" {
 		systemNamespace = podNamespace()
 	}
 
 	setupLog.Info("set up manager")
+	cacheOptions := crcache.Options{
+		ByObject: map[client.Object]crcache.ByObject{
+			&ocv1alpha1.ClusterExtension{}: {Label: k8slabels.Everything()},
+			&catalogd.ClusterCatalog{}:     {Label: k8slabels.Everything()},
+		},
+		DefaultNamespaces: map[string]crcache.Config{
+			systemNamespace: {LabelSelector: k8slabels.Everything()},
+		},
+		DefaultLabelSelector: k8slabels.Nothing(),
+	}
+	if globalPullSecretKey != nil {
+		cacheOptions.ByObject[&corev1.Secret{}] = crcache.ByObject{
+			Field: fields.SelectorFromSet(map[string]string{
+				"metadata.name": globalPullSecretKey.Name,
+			}),
+		}
+		cacheOptions.DefaultNamespaces[globalPullSecretKey.Namespace] = crcache.Config{
+			LabelSelector: k8slabels.Everything(),
+		}
+	}
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
 		Scheme:                 scheme.Scheme,
 		Metrics:                server.Options{BindAddress: metricsAddr},
 		HealthProbeBindAddress: probeAddr,
 		LeaderElection:         enableLeaderElection,
 		LeaderElectionID:       "9c4404e7.operatorframework.io",
-		Cache: crcache.Options{
-			ByObject: map[client.Object]crcache.ByObject{
-				&ocv1alpha1.ClusterExtension{}: {Label: k8slabels.Everything()},
-				&catalogd.ClusterCatalog{}:     {Label: k8slabels.Everything()},
-			},
-			DefaultNamespaces: map[string]crcache.Config{
-				systemNamespace: {LabelSelector: k8slabels.Everything()},
-			},
-			DefaultLabelSelector: k8slabels.Nothing(),
-		},
+		Cache:                  cacheOptions,
 		// LeaderElectionReleaseOnCancel defines if the leader should step down voluntarily
 		// when the Manager ends. This requires the binary to immediately end when the
 		// Manager is stopped, otherwise, this setting is unsafe. Setting this significantly
@@ -191,12 +220,21 @@ func main() {
 
 	unpacker := &source.ContainersImageRegistry{
 		BaseCachePath: filepath.Join(cachePath, "unpack"),
-		SourceContext: &types.SystemContext{
-			DockerCertPath: caCertDir,
-			OCICertPath:    caCertDir,
-			AuthFilePath:   authFilePathIfPresent(setupLog),
-		},
-	}
+		SourceContextFunc: func(logger logr.Logger) (*types.SystemContext, error) {
+			srcContext := &types.SystemContext{
+				DockerCertPath: caCertDir,
+				OCICertPath:    caCertDir,
+			}
+			if _, err := os.Stat(authFilePath); err == nil && globalPullSecretKey != nil {
+				logger.Info("using available authentication information for pulling image")
+				srcContext.AuthFilePath = authFilePath
+			} else if os.IsNotExist(err) {
+				logger.Info("no authentication information found for pulling image, proceeding without auth")
+			} else {
+				return nil, fmt.Errorf("could not stat auth file, error: %w", err)
+			}
+			return srcContext, nil
+		}}
 
 	clusterExtensionFinalizers := crfinalizer.NewFinalizers()
 	if err := clusterExtensionFinalizers.Register(controllers.ClusterExtensionCleanupUnpackCacheFinalizer, finalizers.FinalizerFunc(func(ctx context.Context, obj client.Object) (crfinalizer.Result, error) {
@@ -281,6 +319,19 @@ func main() {
 		os.Exit(1)
 	}
 
+	if globalPullSecretKey != nil {
+		setupLog.Info("creating SecretSyncer controller for watching secret", "Secret", globalPullSecret)
+		err := (&controllers.PullSecretReconciler{
+			Client:       mgr.GetClient(),
+			AuthFilePath: authFilePath,
+			SecretKey:    *globalPullSecretKey,
+		}).SetupWithManager(mgr)
+		if err != nil {
+			setupLog.Error(err, "unable to create controller", "controller", "SecretSyncer")
+			os.Exit(1)
+		}
+	}
+
 	//+kubebuilder:scaffold:builder
 
 	if err := mgr.AddHealthzCheck("healthz", healthz.Ping); err != nil {
@@ -298,18 +349,8 @@ func main() {
 		setupLog.Error(err, "problem running manager")
 		os.Exit(1)
 	}
-}
-
-func authFilePathIfPresent(logger logr.Logger) string {
-	_, err := os.Stat(authFilePath)
-	if os.IsNotExist(err) {
-		logger.Info("auth file not found, skipping configuration of global auth file", "path", authFilePath)
-		return ""
-	}
-	if err != nil {
-		logger.Error(err, "unable to access auth file path", "path", authFilePath)
+	if err := os.Remove(authFilePath); err != nil {
+		setupLog.Error(err, "failed to cleanup temporary auth file")
 		os.Exit(1)
 	}
-	logger.Info("auth file found, configuring globally for image registry interactions", "path", authFilePath)
-	return authFilePath
 }
--- a/internal/rukpak/source/containers_image.go
+++ b/internal/rukpak/source/containers_image.go
@@ -26,8 +26,8 @@ import (
 )
 
 type ContainersImageRegistry struct {
-	BaseCachePath string
-	SourceContext *types.SystemContext
+	BaseCachePath     string
+	SourceContextFunc func(logger logr.Logger) (*types.SystemContext, error)
 }
 
 func (i *ContainersImageRegistry) Unpack(ctx context.Context, bundle *BundleSource) (*Result, error) {
@@ -41,12 +41,16 @@ func (i *ContainersImageRegistry) Unpack(ctx context.Context, bundle *BundleSour
 		return nil, reconcile.TerminalError(fmt.Errorf("error parsing bundle, bundle %s has a nil image source", bundle.Name))
 	}
 
+	srcCtx, err := i.SourceContextFunc(l)
+	if err != nil {
+		return nil, err
+	}
 	//////////////////////////////////////////////////////
 	//
 	// Resolve a canonical reference for the image.
 	//
 	//////////////////////////////////////////////////////
-	imgRef, canonicalRef, _, err := resolveReferences(ctx, bundle.Image.Ref, i.SourceContext)
+	imgRef, canonicalRef, _, err := resolveReferences(ctx, bundle.Image.Ref, srcCtx)
 	if err != nil {
 		return nil, err
 	}
@@ -102,7 +106,7 @@ func (i *ContainersImageRegistry) Unpack(ctx context.Context, bundle *BundleSour
 	// a policy context for the image pull.
 	//
 	//////////////////////////////////////////////////////
-	policyContext, err := loadPolicyContext(i.SourceContext, l)
+	policyContext, err := loadPolicyContext(srcCtx, l)
 	if err != nil {
 		return nil, fmt.Errorf("error loading policy context: %w", err)
 	}
@@ -118,7 +122,7 @@ func (i *ContainersImageRegistry) Unpack(ctx context.Context, bundle *BundleSour
 	//
 	//////////////////////////////////////////////////////
 	if _, err := copy.Image(ctx, policyContext, layoutRef, dockerRef, &copy.Options{
-		SourceCtx: i.SourceContext,
+		SourceCtx: srcCtx,
 	}); err != nil {
 		return nil, fmt.Errorf("error copying image: %w", err)
 	}
@@ -129,7 +133,7 @@ func (i *ContainersImageRegistry) Unpack(ctx context.Context, bundle *BundleSour
 	// Mount the image we just pulled
 	//
 	//////////////////////////////////////////////////////
-	if err := i.unpackImage(ctx, unpackPath, layoutRef); err != nil {
+	if err := i.unpackImage(ctx, unpackPath, layoutRef, srcCtx); err != nil {
 		if cleanupErr := deleteRecursive(unpackPath); cleanupErr != nil {
 			err = errors.Join(err, cleanupErr)
 		}
@@ -225,8 +229,8 @@ func loadPolicyContext(sourceContext *types.SystemContext, l logr.Logger) (*sign
 	return signature.NewPolicyContext(policy)
 }
 
-func (i *ContainersImageRegistry) unpackImage(ctx context.Context, unpackPath string, imageReference types.ImageReference) error {
-	img, err := imageReference.NewImage(ctx, i.SourceContext)
+func (i *ContainersImageRegistry) unpackImage(ctx context.Context, unpackPath string, imageReference types.ImageReference, sourceContext *types.SystemContext) error {
+	img, err := imageReference.NewImage(ctx, sourceContext)
 	if err != nil {
 		return fmt.Errorf("error reading image: %w", err)
 	}
@@ -236,7 +240,7 @@ func (i *ContainersImageRegistry) unpackImage(ctx context.Context, unpackPath st
 		}
 	}()
 
-	layoutSrc, err := imageReference.NewImageSource(ctx, i.SourceContext)
+	layoutSrc, err := imageReference.NewImageSource(ctx, sourceContext)
 	if err != nil {
 		return fmt.Errorf("error creating image source: %w", err)
 	}
--- a/internal/rukpak/source/containers_image_test.go
+++ b/internal/rukpak/source/containers_image_test.go
@@ -14,6 +14,7 @@ import (
 	"github.com/containers/image/v5/docker/reference"
 	"github.com/containers/image/v5/pkg/sysregistriesv2"
 	"github.com/containers/image/v5/types"
+	"github.com/go-logr/logr"
 	"github.com/google/go-containerregistry/pkg/crane"
 	"github.com/google/go-containerregistry/pkg/registry"
 	"github.com/opencontainers/go-digest"
@@ -34,8 +35,8 @@ func TestUnpackValidInsecure(t *testing.T) {
 	defer cleanup()
 
 	unpacker := &source.ContainersImageRegistry{
-		BaseCachePath: t.TempDir(),
-		SourceContext: buildPullContext(t, imageTagRef),
+		BaseCachePath:     t.TempDir(),
+		SourceContextFunc: buildPullContextfunc(t, imageTagRef),
 	}
 	bundleSource := &source.BundleSource{
 		Name: "test-bundle",
@@ -69,8 +70,8 @@ func TestUnpackValidUsesCache(t *testing.T) {
 	defer cleanup()
 
 	unpacker := &source.ContainersImageRegistry{
-		BaseCachePath: t.TempDir(),
-		SourceContext: buildPullContext(t, imageDigestRef),
+		BaseCachePath:     t.TempDir(),
+		SourceContextFunc: buildPullContextfunc(t, imageDigestRef),
 	}
 
 	bundleSource := &source.BundleSource{
@@ -102,8 +103,8 @@ func TestUnpackCacheCheckError(t *testing.T) {
 	defer cleanup()
 
 	unpacker := &source.ContainersImageRegistry{
-		BaseCachePath: t.TempDir(),
-		SourceContext: buildPullContext(t, imageTagRef),
+		BaseCachePath:     t.TempDir(),
+		SourceContextFunc: buildPullContextfunc(t, imageTagRef),
 	}
 	bundleSource := &source.BundleSource{
 		Name: "test-bundle",
@@ -131,8 +132,8 @@ func TestUnpackNameOnlyImageReference(t *testing.T) {
 	defer cleanup()
 
 	unpacker := &source.ContainersImageRegistry{
-		BaseCachePath: t.TempDir(),
-		SourceContext: buildPullContext(t, imageTagRef),
+		BaseCachePath:     t.TempDir(),
+		SourceContextFunc: buildPullContextfunc(t, imageTagRef),
 	}
 	bundleSource := &source.BundleSource{
 		Name: "test-bundle",
@@ -153,8 +154,8 @@ func TestUnpackUnservedTaggedImageReference(t *testing.T) {
 	defer cleanup()
 
 	unpacker := &source.ContainersImageRegistry{
-		BaseCachePath: t.TempDir(),
-		SourceContext: buildPullContext(t, imageTagRef),
+		BaseCachePath:     t.TempDir(),
+		SourceContextFunc: buildPullContextfunc(t, imageTagRef),
 	}
 	bundleSource := &source.BundleSource{
 		Name: "test-bundle",
@@ -175,8 +176,8 @@ func TestUnpackUnservedCanonicalImageReference(t *testing.T) {
 	defer cleanup()
 
 	unpacker := &source.ContainersImageRegistry{
-		BaseCachePath: t.TempDir(),
-		SourceContext: buildPullContext(t, imageTagRef),
+		BaseCachePath:     t.TempDir(),
+		SourceContextFunc: buildPullContextfunc(t, imageTagRef),
 	}
 
 	origRef := imageDigestRef.String()
@@ -232,7 +233,11 @@ func TestUnpackInvalidNilImage(t *testing.T) {
 }
 
 func TestUnpackInvalidImageRef(t *testing.T) {
-	unpacker := &source.ContainersImageRegistry{}
+	unpacker := &source.ContainersImageRegistry{
+		SourceContextFunc: func(logr.Logger) (*types.SystemContext, error) {
+			return &types.SystemContext{}, nil
+		},
+	}
 	// Create BundleSource with malformed image reference
 	bundleSource := &source.BundleSource{
 		Name: "test-bundle",
@@ -255,8 +260,8 @@ func TestUnpackUnexpectedFile(t *testing.T) {
 	defer cleanup()
 
 	unpacker := &source.ContainersImageRegistry{
-		BaseCachePath: t.TempDir(),
-		SourceContext: buildPullContext(t, imageTagRef),
+		BaseCachePath:     t.TempDir(),
+		SourceContextFunc: buildPullContextfunc(t, imageTagRef),
 	}
 	bundleSource := &source.BundleSource{
 		Name: "test-bundle",
@@ -280,8 +285,8 @@ func TestUnpackCopySucceedsMountFails(t *testing.T) {
 	defer cleanup()
 
 	unpacker := &source.ContainersImageRegistry{
-		BaseCachePath: t.TempDir(),
-		SourceContext: buildPullContext(t, imageTagRef),
+		BaseCachePath:     t.TempDir(),
+		SourceContextFunc: buildPullContextfunc(t, imageTagRef),
 	}
 	bundleSource := &source.BundleSource{
 		Name: "test-bundle",
@@ -305,8 +310,8 @@ func TestCleanup(t *testing.T) {
 	defer cleanup()
 
 	unpacker := &source.ContainersImageRegistry{
-		BaseCachePath: t.TempDir(),
-		SourceContext: buildPullContext(t, imageTagRef),
+		BaseCachePath:     t.TempDir(),
+		SourceContextFunc: buildPullContextfunc(t, imageTagRef),
 	}
 	bundleSource := &source.BundleSource{
 		Name: "test-bundle",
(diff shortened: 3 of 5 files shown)


config/base/rbac/role.yaml
```diff
@@ -4,6 +4,14 @@ kind: ClusterRole
...
   name: manager-role
 rules:
+- apiGroups:
+  - ""
+  resources:
+  - secrets
+  verbs:
+  - get
+  - list
+  - watch
```
joelanford, line 14: I think we should remove this.

I would expect the base install to:
- not define this flag
- therefore not need these rules

If someone deploying OLMv1 wants to enable this feature, they can provide the appropriate (and minimal) RBAC.

internal/controllers/secret_syncer_controller.go
```diff
@@ -0,0 +1,117 @@
...
+		if !ok {
+			return false
+		}
+		return secret.Name == secretName && secret.Namespace == secretNamespace
+	})
+}
+
+// writeSecretToFile writes the secret data to the specified file
+func (r *SecretSyncerReconciler) writeSecretToFile(secret *corev1.Secret, logger logr.Logger) (ctrl.Result, error) {
+	jsonData, err := json.Marshal(secret.Data)
```
joelanford, line 89: I don't think this will work. We need to write a specific key to the file, not the entire set of keys from `Data`.

The standard pull secret key for the data we want is: `.dockerconfigjson`.

internal/controllers/secret_syncer_controller.go
```diff
@@ -0,0 +1,116 @@
...
+		logger.Info("Received unexpected request for secret", req.Name, "in namespace", req.Namespace)
+		return ctrl.Result{}, nil
+	}
+
+	secret := &corev1.Secret{}
+	err := r.Get(ctx, req.NamespacedName, secret)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			logger.Info("Secret", req.Name, "in namespace", req.Namespace, "deleted.")
+			return r.deleteSecretFile(logger)
```
rashmi43, line 57: can we have an empty jsonstring, do we have to delete the auth json file?

internal/controllers/secret_syncer_controller.go
rashmi43: @joelanford does this mean we need a diff of what's in the pull-secret and auth file?

anik120: Looks like the pull secret we're concerned with only has that one key in it anyway, at least for now: 


```
apiVersion: v1
data:
  .dockerconfigjson: ewogICJhdX....==
kind: Secret
metadata:
   .
   .
  name: pull-secret
  namespace: ....
.
.
.
type: kubernetes.io/dockerconfigjson
```

And what gets saved is a json object: 


```
$ cat /tmp/operator-controller/auth.json 
{".dockerconfigjson":"ewogICJhdX....=="}/ $
```

Are you saying that we need to hardcode extract and save only `.dockerconfigjson`? Are all Secrets that users could use for this purpose only ever going to have that key and no other key?

rashmi43: yes I believe so: https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/#registry-secret-existing-credentials

cmd/manager/main.go
```diff
@@ -289,6 +298,18 @@ func main() {
 		os.Exit(1)
 	}
 
+	if globalPullSecretName != "" && globalPullSecretNamespace != "" {
+		err := (&controllers.SecretSyncerReconciler{
+			Client:          mgr.GetClient(),
+			SecretName:      globalPullSecretName,
+			SecretNamespace: globalPullSecretNamespace,
+		}).SetupWithManager(mgr)
```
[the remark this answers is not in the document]
anik120: Yes I missed that initially

internal/controllers/secret_syncer_controller.go
anik120: That's a good point. Looks like we need to create an empty file initially too if the flag is present, since [authFilePathIfPresent](https://github.com/operator-framework/operator-controller/blob/main/cmd/manager/main.go#L312) will not pass this file path to `SourceContext` if the file is not present. 

And in that case, on deletion we need to leave the file empty, otherwise `SourceContext` will try to use a non-existent file.

cmd/manager/main.go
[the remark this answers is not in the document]
anik120: Added it.

config/base/rbac/role.yaml
anik120: That's fair.

anik120:
Where the PR stands so far: 

* On start up: 

```
.
.
I1001 16:50:17.917567       1 controller.go:217] "Starting workers" controller="secret" controllerGroup="" controllerKind="Secret" worker count=1
.
.
```

* On creation of a secret, the namespace/name for which is passed via the flag: 

```
I1001 16:51:08.001691       1 secret_syncer_controller.go:103] "Saved Secret data locally" controller="secret" controllerGroup="" controllerKind="Secret" Secret="olmv1-system/pull-secret" namespace="olmv1-system" name="pull-secret" reconcileID="fe0d20fe-17dd-4af7-9a76-d433de66ed3a"
```

- The data in the file system: 

```
$ cat /tmp/operator-controller/auth.json
{".dockerconfigjson":"{\n  \"auths\": {\n    \"registry.build09.ci.openshift.org\": {\n      \"auth\": \"dXNlcjp..
```
More details about this file [here](https://github.com/operator-framework/operator-controller/pull/1322#discussion_r1783214244)


* On deletion of the Secret: 

```
I1001 17:03:31.719907       1 secret_syncer_controller.go:54] "Secret" controller="secret" controllerGroup="" controllerKind="Secret" Secret="olmv1-system/pull-secret" namespace="olmv1-system" name="pull-secret" reconcileID="ea3490a2-fa78-46d4-a396-a10cad77dddc" pull-secret="in namespace" olmv1-system="deleted."
I1001 17:03:31.720283       1 secret_syncer_controller.go:109] "Emptying local auth file." controller="secret" controllerGroup="" controllerKind="Secret" Secret="olmv1-system/pull-secret" namespace="olmv1-system" name="pull-secret" reconcileID="ea3490a2-fa78-46d4-a396-a10cad77dddc"
I1001 17:03:31.720629       1 secret_syncer_controller.go:116] "Auth file emptied successfully" controller="secret" controllerGroup="" controllerKind="Secret" Secret="olmv1-system/pull-secret" namespace="olmv1-system" name="pull-secret" reconcileID="ea3490a2-fa78-46d4-a396-a10cad77dddc" file="/tmp/operator-controller/auth.json"
```

- Content in the file 
```
$ cat /tmp/operator-controller/auth.json
$
```

internal/controllers/secret_syncer_controller.go
rashmi43: this is correct:
```{
  ".dockerconfigjson": {
    "auths": {
      .
      .
      .
      "cloud.openshift.com": {
        "auth": "b3BlbnNoaWZ0LXJlbGVhc2UtZGV2K29jbV9hY2N...",
        "email": "aos-team-dp-testplatform@redhat.com"
      },
      "quay.io": {
        "auth": "b3BlbnNoaWZ0LXJlbGVhc2UtZGV2K29jbV9hY2N...",
        "email": "aos-team-dp-testplatform@redhat.com"
      },
      "quay.io/openshift/ci": {
        "auth": "b3BlbnNoaWZ0K3FjaV9wdWxsX3JvYm90OkVMVD..."
      }
    }
  }
}

File: internal/controllers/secret_syncer_controller.go (hunk 3 above)
Comment (rashmi43) [reply internal/controllers/secret_syncer_controller.go]: when empty is should look this:
```
{
  ".dockerconfigjson": {
    "auths": {}
}

rashmi43:
```
oc get secret regcred -o yaml
apiVersion: v1
data:
  .dockerconfigjson: ewogICAgImF1dGhzIjogewogICAgfQp9Cg==
```

and
```
echo "ewogICAgImF1dGhzIjogewogICAgfQp9Cg=="|base64 -d
{
    "auths": {
    }
}
```

internal/controllers/secret_syncer_controller.go
joelanford: Here is the specification of the auth file: https://github.com/containers/image/blob/main/docs/containers-auth.json.5.md

We need to drop the top-level `".dockerconfigjson"` key. The file needs to contain the _value_ of that key.

cmd/manager/main.go
```diff
@@ -289,6 +305,21 @@ func main() {
...
 	}
 
+	if globalPullSecret != "" {
+		secretParts := strings.Split(globalPullSecret, "/")
+		setupLog.Info("Creating SecretSyncer controller for watching secret", secretParts[1], "in namespace", secretParts[0])
+		err := (&controllers.SecretSyncerReconciler{
+			Client:          mgr.GetClient(),
+			AuthFilePath:    authFilePath,
+			SecretName:      secretParts[1],
+			SecretNamespace: secretParts[0],
```
joelanford, line 315: We need to configure the cache to start a secret informer in the specified secret's namespace. We should not have a cluster-wide secret infomer. That would require the controller to have cluster-wide list-watch permissions on secrets, and that's a big no-no.

internal/controllers/secret_syncer_controller.go
```diff
@@ -0,0 +1,124 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+)
+
+// SecretSyncerReconciler reconciles a specific secret object
+type SecretSyncerReconciler struct {
+	client.Client
+	SecretName      string
+	SecretNamespace string
```
joelanford, line 38: Use `SecretKey types.NamespacedName` here instead?

skattoju:
fixes #1015

internal/controllers/secret_syncer_controller.go
anik120: Aha thanks Joe.

internal/controllers/secret_syncer_controller.go
```diff
@@ -0,0 +1,150 @@
...
+
+	return err
+}
+
+func newSecretPredicate(secretName, secretNamespace string) predicate.Predicate {
+	return predicate.NewPredicateFuncs(func(obj client.Object) bool {
+		secret, ok := obj.(*corev1.Secret)
+		if !ok {
+			return false
+		}
```
everettraven, line 90: Nit: Do we need to have this additional validation from the type conversion? I imagine we should be able to "trust" that controller-runtime will respect our controller configuration that we build to only send events for `Secret` resources.

internal/controllers/secret_syncer_controller.go
```diff
@@ -0,0 +1,150 @@
...
+	// if err := r.Client.Get(context.TODO(), client.ObjectKey{
+	// 	Name:      r.SecretKey.Name,
+	// 	Namespace: r.SecretKey.Namespace,
+	// }, &corev1.Secret{}); err != nil {
+	// 	if apierrors.IsNotFound(err) {
+	// 		r.Logger.Info("Secret not found during initialization.", "Secret", fmt.Sprintf("%v/%v", r.SecretKey.Name, r.SecretKey.Namespace))
+	// 	} else {
+	// 		return err
+	// 	}
+	// }
```
everettraven, line 75: This code needed any longer?

internal/controllers/secret_syncer_controller.go
```diff
@@ -0,0 +1,150 @@
...
+	client.Client
+	Logger       logr.Logger
+	SecretKey    types.NamespacedName
+	AuthFilePath string
+}
+
+func (r *SecretSyncerReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+
+	if req.Name != r.SecretKey.Name || req.Namespace != r.SecretKey.Namespace {
+		r.Logger.Info("Received unexpected request for Secret", "Secret", fmt.Sprintf("%v/%v", req.Namespace, req.Name))
```
everettraven, line 46: Should we be careful about leaking the namespace and name of secrets on the cluster in our logs?

internal/controllers/secret_syncer_controller.go
```diff
@@ -0,0 +1,150 @@
...
+			}
+			// expected format for auth.json
+			// https://github.com/containers/image/blob/main/docs/containers-auth.json.5.md
+			secretData["auths"] = dockerConfig["auths"]
+		}
+	}
+	if err := r.writeDataToFile(secretData); err != nil {
+		return ctrl.Result{}, err
+	}
+	r.Logger.Info("Saved Secret data locally", "file", r.AuthFilePath)
```
everettraven, line 114: Same question regarding potentially leaking information in logs about credentials. Is this something that a malicious actor could potentially leverage to impact a cluster using OLM?

cmd/manager/main.go
```diff
@@ -117,6 +123,14 @@ func main() {
 
 	setupLog.Info("starting up the controller", "version info", version.String())
 
+	var globalPullSecretKey *k8stypes.NamespacedName
+	if globalPullSecret != "" {
+		secretParts := strings.Split(globalPullSecret, "/")
```
joelanford, line 128: We need:
```go
if len(secretParts) != 2 {
    setupLog.Error(...)
    os.Exit(1)
}
```

cmd/manager/main.go
```diff
@@ -117,6 +123,14 @@ func main() {
...
+		globalPullSecretKey = &k8stypes.NamespacedName{Name: secretParts[1], Namespace: secretParts[0]}
+	} else {
+		globalPullSecretKey = nil
+	}
```
joelanford, line 132: This is unnecessary and can be deleted. `nil` is the default value of a pointer that gets assigned in line 126 when you declare the variable.

internal/controllers/secret_syncer_controller.go
```diff
@@ -0,0 +1,150 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+)
+
+// SecretSyncerReconciler reconciles a specific secret object
+type SecretSyncerReconciler struct {
+	client.Client
+	Logger       logr.Logger
```
joelanford, line 38: `Logger` should not be a field on the reconciler. We should inherit from the `context.Context`. (e.g. https://github.com/operator-framework/operator-controller/blob/801388ac7bed43355d38efaa4e3bd42378ad83a4/internal/controllers/clusterextension_controller.go#L101)

Though I'm actually not sure the `WithName` is necessary there.

internal/controllers/secret_syncer_controller.go
```diff
@@ -0,0 +1,150 @@
...
+	// 	if apierrors.IsNotFound(err) {
+	// 		r.Logger.Info("Secret not found during initialization.", "Secret", fmt.Sprintf("%v/%v", r.SecretKey.Name, r.SecretKey.Namespace))
+	// 	} else {
+	// 		return err
+	// 	}
+	// }
+
+	_, err := ctrl.NewControllerManagedBy(mgr).
+		For(&corev1.Secret{}).
+		WithEventFilter(newSecretPredicate(r.SecretKey.Name, r.SecretKey.Namespace)).
```
joelanford, line 79: Just pass the key?
```suggestion
		WithEventFilter(newSecretPredicate(r.SecretKey)).
```

internal/controllers/secret_syncer_controller.go
joelanford: Correct. We also don't need to assert to get the object's name or namespace.

I would rename this to `namespacedNamePredicate` and then use `obj.GetName()` and `obj.GetNamespace()` directly. No need for the type assertion.

internal/controllers/secret_syncer_controller.go
```diff
@@ -0,0 +1,150 @@
...
+		// image registry secrets are always stored with the key .dockerconfigjson
+		// ref: https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/#registry-secret-existing-credentials
+		if key == ".dockerconfigjson" {
+			var dockerConfig map[string]interface{}
+			if err := json.Unmarshal(value, &dockerConfig); err != nil {
+				return ctrl.Result{}, fmt.Errorf("error unmarshalling .dockerconfigjson: %w", err)
+			}
+			// expected format for auth.json
+			// https://github.com/containers/image/blob/main/docs/containers-auth.json.5.md
+			secretData["auths"] = dockerConfig["auths"]
```
joelanford, line 108: We don't need to have a json Unmarshal/Marshal roundtrip here do we?

We can use `os.WriteFile` to directly write the `value` to the file.

internal/controllers/secret_syncer_controller.go
rashmi43: that's right `.dockerconfigjson` is only for k8s image secrets , its values is the auths json which we will store.

cmd/manager/main.go
anik120: right, I was using `var globalPullSecretKey string` before I decided to switch to a pointer, so remnant of that. Removing it.

internal/controllers/secret_syncer_controller.go
anik120: Right, changed this to sending just the key. But another thought I had was, do we even need this event filter? We're already setting up the cache in a way that we only cache the specific secret, do we additionally need to configure even filters here too?

internal/controllers/secret_syncer_controller.go
anik120: Yup, remanent of the experiment I was doing with the format, simplifying this to just write the value.

internal/controllers/secret_syncer_controller.go
anik120: Removing this, since we log the value of the flag when we're starting the controller: 

```
I1003 11:02:18.606018       1 main.go:320] "creating SecretSyncer controller for watching secret" logger="setup" Secret="default/pull-secret"
```

cmd/manager/main.go
anik120: [So the only combination that seems to work](https://github.com/operator-framework/operator-controller/pull/1322/files#diff-435eecb1d2af40ee96747f88ab71eb2758da989c92f76dcb1f714fc1b300c633R158-R167) (controller actually receives events for and reconciles the object) is 

```
if globalPullSecretKey != nil {
		cacheOptions.ByObject[&corev1.Secret{}] = crcache.ByObject{
			Field: fields.SelectorFromSet(map[string]string{
				"metadata.name": globalPullSecretKey.Name,
			}),
		}
		cacheOptions.DefaultNamespaces[globalPullSecretKey.Namespace] = crcache.Config{
			LabelSelector: k8slabels.Everything(),
		}
}
```

ie 
* in `ByObject`: use field selector 
* in `Config`: use labels.Everything()

If I switch the field used in `Config` to `FieldSelector` and pass in 

```
fields.SelectorFromSet(map[string]string{
	"metadata.name": globalPullSecretKey.Name,
})
```

the controller doesn't reconcile. 

I'm guessing the combination I mentioned will do the job (cache Secrets with `metadata.Name`=<name-we-want>, with labels.Everything() from the namespace). I'm guessing because:what's a good way of checking what's being cached?

internal/controllers/secret_syncer_controller.go
anik120: I removed the `key/value` pair, but since we're deriving the log from the context, the logger already logs the info about the secret, like Joe mentioned below

internal/controllers/secret_syncer_controller.go
```diff
@@ -0,0 +1,150 @@
...
+	if req.Name != r.SecretKey.Name || req.Namespace != r.SecretKey.Namespace {
+		r.Logger.Info("Received unexpected request for Secret", "Secret", fmt.Sprintf("%v/%v", req.Namespace, req.Name))
+		return ctrl.Result{}, nil
+	}
+
+	secret := &corev1.Secret{}
+	err := r.Get(ctx, req.NamespacedName, secret)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			r.Logger.Info("Secret deleted", "Secret", fmt.Sprintf("%v/%v", req.Namespace, req.Name))
```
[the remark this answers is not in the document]
anik120: Simplifying this by removing the key/value pair, logger still logs the secret. Resolving since conversation is happening above too.

cmd/manager/main.go
```diff
@@ -101,6 +105,7 @@ func main() {
 	flag.StringVar(&cachePath, "cache-path", "/var/cache", "The local directory path used for filesystem based caching")
 	flag.BoolVar(&operatorControllerVersion, "version", false, "Prints operator-controller version information")
 	flag.StringVar(&systemNamespace, "system-namespace", "", "Configures the namespace that gets used to deploy system resources.")
+	flag.StringVar(&globalPullSecret, "global-pull-secret", "", "The namespace/name of the global pull secret that is going to be used to pull bundle images.")
```
everettraven, line 108: Just as a thing to note, there is a [`flag.Var()` method](https://pkg.go.dev/flag#FlagSet.Var) that would allow us to implement a custom flag value with custom logic for handling how it is set, including validations. 

I did something similar to this in kapp back when we were contributing preflight checks: https://github.com/carvel-dev/kapp/blob/ca6887d26d782636fe62a2b3f3bbbf5c91a965f3/pkg/kapp/preflight/registry.go#L64-L97

Might be overkill for this particular case, but this may be a good way to handle any custom flag validations we want in the future

internal/controllers/secret_syncer_controller.go
anik120: @rashmi43 fyi pivoted this a little bit after having a conversation with Joe. Now if the secret is deleted, we delete the auth file, and the [second commit](https://github.com/operator-framework/operator-controller/pull/1322/commits/bec4e50966531e3e9fe826dab93a5fc7ce668df9) changes the containerRegistry.Unpacker to check if the file exists before passing it to `container/images` library.

internal/controllers/secret_syncer_controller.go
anik120: resolving this conversation because I have the write stuff down for this now, thank you both 🙏🏽

cmd/manager/main.go
```diff
@@ -130,22 +145,33 @@ func main() {
...
+	}
+	if globalPullSecretKey != nil {
+		cacheOptions.ByObject[&corev1.Secret{}] = crcache.ByObject{
+			Field: fields.SelectorFromSet(map[string]string{
+				"metadata.name": globalPullSecretKey.Name,
+			}),
+		}
+		cacheOptions.DefaultNamespaces[globalPullSecretKey.Namespace] = crcache.Config{
+			LabelSelector: k8slabels.Everything(),
+		}
```
joelanford, line 166: This doesn't seem quite right to me, but we may need to experiment to know for sure. I feel like the following should be sufficient (but maybe it isn't?):
```suggestion
		cacheOptions.ByObject[&corev1.Secret{}] = crcache.ByObject{
			Namespaces: map[string]crcache.Config{
				globalPullSecretKey.Namespace: crcache.Config{
					Field: fields.SelectorFromSet(map[string]string{
						"metadata.name": globalPullSecretKey.Name,
					},
				},	
				systemNamespace: crcache.Config{
					LabelSelector: k8slabels.Everything(),
				},
			}
		}
```

internal/controllers/secretsyncer_controller.go
```diff
@@ -0,0 +1,125 @@
...
+type SecretSyncerReconciler struct {
+	client.Client
+	SecretKey    types.NamespacedName
+	AuthFilePath string
+}
+
+func (r *SecretSyncerReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	logger := log.FromContext(ctx)
+	if req.Name != r.SecretKey.Name || req.Namespace != r.SecretKey.Namespace {
+		logger.Info("received unexpected request for Secret")
```
joelanford, line 45: Seems like this would be more appropriately logged as an error?

internal/controllers/secretsyncer_controller.go
```diff
@@ -0,0 +1,125 @@
...
+
+func newSecretPredicate(key types.NamespacedName) predicate.Predicate {
+	return predicate.NewPredicateFuncs(func(obj client.Object) bool {
+		return obj.GetName() == key.Name && obj.GetNamespace() == key.Namespace
+	})
+}
+
+// writeSecretToFile writes the secret data to the specified file
+func (r *SecretSyncerReconciler) writeSecretToFile(logger logr.Logger, secret *corev1.Secret) (ctrl.Result, error) {
+	for key, value := range secret.Data {
```
joelanford, line 81: This is an unnecessary loop. We should do basically this instead:

```go
dockerConfigJSON, ok := secret.Data[".dockerconfigjson"]
if !ok {
    // delete the file
    // log error because expected key not found
    // return ctrl.Result{}, nil
}
if err := r.writeDataToFile(value); err != nil {
	return ctrl.Result{}, err
}
logger.Info("saved Secret data locally", "file", r.AuthFilePath)
return ctrl.Result{}, nil
```

internal/controllers/secretsyncer_controller.go
```diff
@@ -0,0 +1,125 @@
...
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to delete secret file: %w", err)
+		}
+		logger.Info("auth file deleted successfully", "file", r.AuthFilePath)
+	} else if os.IsNotExist(err) {
+		logger.Info("auth file does not exist, nothing to delete", "file", r.AuthFilePath)
+	} else {
+		return ctrl.Result{}, fmt.Errorf("failed to check secret file: %w", err)
+	}
+	return ctrl.Result{}, nil
```
joelanford, line 110: Seems like quite a bit of complexity. Could we simply call `os.Remove(r.AuthFilePath)`, do the error check, and have a log line for both possibilities (failed to remove, successfully removed)?

I don't think the extra `os.Stat` adds much.

internal/controllers/secretsyncer_controller.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+func (r *SecretSyncerReconciler) writeDataToFile(data []byte) error {
+	// create the directory first if it does not exists
+	dir := filepath.Dir(r.AuthFilePath)
+	if err := os.MkdirAll(dir, 0755); err != nil {
+		return fmt.Errorf("unable to create directory for storing auth: %w", err)
+
+	}
+	err := os.WriteFile(r.AuthFilePath, data, 0644)
```
joelanford, line 120:
```suggestion
	err := os.WriteFile(r.AuthFilePath, data, 0600)
```

internal/controllers/secretsyncer_controller_test.go
```diff
@@ -0,0 +1,99 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
+func TestSecretSyncerReconciler(t *testing.T) {
+	secretData := []byte(`{"auths":{"exampleRegistry": "exampledata"}}`)
+	authFile := "/tmp/test-auth.json"
```
joelanford, line 20: Don't write to the user's disk this way.

We should use `t.TempDir()` called inside `t.Run()`'s inline function and a relative path for the file from there so that we avoid:
1. Reusing the same path between different test cases
2. Accidentally leaving test files behind on a user's system.

internal/controllers/secretsyncer_controller_test.go
```diff
@@ -0,0 +1,99 @@
...
+			cl := clientBuilder.Build()
+
+			secretKey := types.NamespacedName{Namespace: tt.secret.Namespace, Name: tt.secret.Name}
+			r := &controllers.SecretSyncerReconciler{
+				Client:       cl,
+				SecretKey:    secretKey,
+				AuthFilePath: authFile,
+			}
+			if tt.fileShouldExistBefore {
+				err := os.WriteFile(authFile, secretData, 0644)
```
joelanford, line 76:
```suggestion
				err := os.WriteFile(authFile, secretData, 0600)
```

internal/rukpak/source/containers_image.go
```diff
@@ -41,12 +42,21 @@ func (i *ContainersImageRegistry) Unpack(ctx context.Context, bundle *BundleSour
 		return nil, reconcile.TerminalError(fmt.Errorf("error parsing bundle, bundle %s has a nil image source", bundle.Name))
 	}
 
+	srcCtx := *i.SourceContext
```
joelanford, line 45: This is a shallow copy. In this particular case, adding a `AuthFilePath` to the shallow copy won't affect anything, but it would be easy for our future selves to overlook this and accidentally modify a pointer within the shallow copy (which would affect the original)

I wonder two things:
1. If we hardcode `AuthFilePath` in `main.go` but the file doesn't exist, can `containers/image` tolerate that? I think the answer is no, but have we tried?
2. If the answer to (1) is no, maybe we need `SourceContextFunc` on the `ContainersImageRegistry` struct, which can generate a brand new `*types.SystemContext` for one-time-use when we call it (which we would do in `Unpack`. That way we can provide a function that can do the below check and optionally setup `AuthFilePath`.

cmd/manager/main.go
```diff
@@ -69,7 +72,7 @@ var (
 	defaultSystemNamespace = "olmv1-system"
 )
 
-const authFilePath = "/etc/operator-controller/auth.json"
+const authFilePath = "/tmp/operator-controller/auth.json"
```
joelanford, line 75: This change is interesting. I think I understand the motivation behind it, but I feel like this is something we should chat through.

Would it make sense to use a path directly in the current working directory, and then tell the controller and the unpacker the absolute path for it? Maybe something like `$PWD/operator-controller-global-pull-secret.json` (but maybe $TMPDIR is fine too; we should use `$TMPDIR`, not `/tmp` if we stick with an absolute location).

IMO, we should:
- do it in a place where we don't have to create a directory to store it.
- delete it when the process is shutting down.

internal/controllers/secret_syncer_controller.go
joelanford: I think we do. There may be other secrets in the systemNamespace that show up here.

cmd/manager/main.go
anik120: Nah that didn't work, what I have is the only combination that's worked so far.

anik120: Also in case you missed it: [comment](https://github.com/operator-framework/operator-controller/pull/1322#discussion_r1786044859)

cmd/manager/main.go
anik120: Fyi https://docs.google.com/document/d/1BXD6kj5zXHcGiqvJOikU2xs8kV26TPnzEKp6n7TKD4M/edit?disco=AAABTsk4FzQ 

* We run into permission issues if we try to create the file in /etc. 
* $PWD is root (/), so we run into permission issues trying go create /operator-controller-global-pull-secret.json too. 
* I don't see $TMPDIR set when I deploy locally. We could do something like 
```
tmpDir := os.Getenv("TMPDIR")
if tmpDir == "" {
    tmpDir = "/tmp"
} 
```

but is that really necessary?

internal/rukpak/source/containers_image.go
anik120: [Done](https://github.com/operator-framework/operator-controller/pull/1322/commits/33f8089a6afa7c7dffe4998e432881036f8e80c0#diff-435eecb1d2af40ee96747f88ab71eb2758da989c92f76dcb1f714fc1b300c633R230-R244)

cmd/manager/main.go
```diff
@@ -130,22 +145,33 @@ func main() {
...
+		},
+		DefaultLabelSelector: dependentSelector,
+	}
+	if globalPullSecretKey != nil {
+		cacheOptions.ByObject[&corev1.Secret{}] = crcache.ByObject{
+			Field: fields.SelectorFromSet(map[string]string{
+				"metadata.name": globalPullSecretKey.Name,
+			}),
+		}
+		cacheOptions.DefaultNamespaces[globalPullSecretKey.Namespace] = crcache.Config{
```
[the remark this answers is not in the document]
anik120: So the answer for these can be anchored around the RBAC. 

- The Role and the RoleBinding for our SA must exist before we can set up a watch from our controller. 
- If the namespace does not exist, the Role/RoleBinding cannot exist. So the controller pod will keep crashing 
```
E1004 12:06:32.363508       1 reflector.go:158] "Unhandled Error" err="pkg/cache/internal/informers.go:106: Failed to watch *v1.Secret: failed to list *v1.Secret: secrets \"pull-secret\" is forbidden: User \"system:serviceaccount:olmv1-system:operator-controller-controller-manager\" cannot list resource \"secrets\" in API group \"\" in the namespace \"test-namespace\"" logger="UnhandledError"
```

The last two questions were interesting finds for me. Looks like if the namespace exists when the controller starts but then is deleted, I don't see any error immediately. Nor for the ~5 mins that I kept the `logs -f` up. 
When the namespace gets created, I still don't see anything in the logs. 

TLDR of course is that you must have the proper Role/Rolebinding (and as a result the namespace) before you start our controller with the flag to watch for the specific secret in that namespace. Otherwise the controller will fail to start.

internal/rukpak/source/containers_image.go
joelanford: Nice! Looks great!

cmd/manager/main.go
joelanford: Use `os.TempDir()`. It's basically that.

cmd/manager/main.go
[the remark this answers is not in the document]
joelanford: > If the namespace does not exist, the Role/RoleBinding cannot exist.

Yeah, in my scenarios let's assume that there is a ClusterRole and ClusterRoleBinding, so that we can assume we have permission everywhere all the time. (That obviously wouldn't be what we'd suggest, but its useful to isolate the behavior of the caching logic and informers)

I don't think we need to actually _solve_ any weirdness that comes up with my scenarios. I just think we need to understand it and write it down in docs and in the help text of the flag we add.

For example, I could see us saying:
```
--global-pull-secret    A secret reference with the format <namespace>/<name>. The namespace is expected to exist for the life of the process. Creation or deletion of the namespace after the controller starts is likely to ... <whatever we see happens when we test this>"
```

internal/controllers/secretsyncer_controller.go
```diff
@@ -0,0 +1,124 @@
...
+	if !ok {
+		logger.Error(fmt.Errorf("expected secret.Data key not found"), "expected secret Data to contain key .dockerconfigjson")
+		return ctrl.Result{}, nil
+	}
+	// expected format for auth.json
+	// https://github.com/containers/image/blob/main/docs/containers-auth.json.5.md
+	if err := r.writeDataToFile(dockerConfigJSON); err != nil {
+		return ctrl.Result{}, err
+	}
+	logger.Info("saved Secret data locally", "file", r.AuthFilePath)
```
m1kola, line 93: Do we want to log file path here and in other places? I don't think it is necessary to log a file from a temp dir (especially in info logs). It is like a location of a cache in this case.

internal/rukpak/source/containers_image_test.go
```diff
@@ -103,7 +108,9 @@ func TestUnpackCacheCheckError(t *testing.T) {
 
 	unpacker := &source.ContainersImageRegistry{
 		BaseCachePath: t.TempDir(),
-		SourceContext: buildPullContext(t, imageTagRef),
+		SourceContextFunc: func(_ logr.Logger) (*types.SystemContext, error) {
+			return buildPullContext(t, imageTagRef)
+		},
```
m1kola, line 113: Nit: `_` is not necessary.

```suggestion
		SourceContextFunc: func(logr.Logger) (*types.SystemContext, error) {
			return buildPullContext(t, imageTagRef)
		},
```

Since we are modifying `buildPullContext` anyway we can even make `buildPullContext` to be `buildSourceContextFunc` which returns `func(logr.Logger) (*types.SystemContext, error)` and here we can have something like:

```suggestion
		SourceContextFunc: buildSourceContextFunc(t, imageTagRef),
```

cmd/manager/main.go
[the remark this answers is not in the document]
m1kola: Agree with the approach of saying that "this and that is a prerequisite" and going simple. We. can come up with something more sophisticated later, if we see the need for it.

m1kola (commented):
Just a few comments/questions and suggestions (nits, really). Other than that - no additional concerns apart from what has already been raised.

cmd/manager/main.go
anik120: Done in a [new commit](https://github.com/operator-framework/operator-controller/pull/1322/commits/527e324664d0761aa4cd6b793b2e8f906d652604)

internal/rukpak/source/containers_image_test.go
anik120: Yea you're right, I got lazy. Incorporated this in the [new commit](https://github.com/operator-framework/operator-controller/pull/1322/commits/d7648e61a5825255733702e41826926c26c9b385)

internal/controllers/secretsyncer_controller.go
anik120: Yea I've had others raise the same issue too, I've just removed it. Was helpful while I was debugging everything but you're right doesn't make sense to put this in the info logs.

internal/controllers/secretsyncer_controller.go
```diff
@@ -0,0 +1,124 @@
...
+	SecretKey    types.NamespacedName
+	AuthFilePath string
+}
+
+func (r *SecretSyncerReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	logger := log.FromContext(ctx)
+	if req.Name != r.SecretKey.Name || req.Namespace != r.SecretKey.Namespace {
+		logger.Error(fmt.Errorf("received unexpected request for Secret %v/%v", req.Namespace, req.Name), "reconciliation error")
+		return ctrl.Result{}, nil
+	}
```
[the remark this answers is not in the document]
anik120: Most likely not, but doesn't hurt to keep the check?

cmd/manager/main.go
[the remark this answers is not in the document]
anik120: That is a LONG message for the flag though. Probably neat-er to document this in the doc we write for this.

internal/controllers/secretsyncer_controller.go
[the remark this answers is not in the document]
m1kola: I'm fine with it. Resolving.

cmd/manager/main.go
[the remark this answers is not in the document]
m1kola: I think it is fine. Kubectl has longer messages, for example. And this one is something which gets configured once and probably rarely modified.

But adding it into docs in addition to the flag description sounds good to me to.

cmd/manager/main.go
joelanford: We got this sorted out. I pushed a new commit with improved cache options.

internal/controllers/secretsyncer_controller.go
joelanford, line 1: File and struct name nit:
- `pull_secret_controller.go`
- `PullSecretReconciler`

WDYT?

cmd/manager/main.go
joelanford: We got it sorted out in the latest commit!

cmd/manager/main.go
joelanford: Looks good!

cmd/manager/main.go
```diff
@@ -200,12 +230,21 @@ func main() {
...
+		SourceContextFunc: func(logger logr.Logger) (*types.SystemContext, error) {
+			srcContext := &types.SystemContext{
+				DockerCertPath: caCertDir,
+				OCICertPath:    caCertDir,
+			}
+			if _, err := os.Stat(authFilePath); err == nil {
+				logger.Info("using available authentication information for pulling image")
+				srcContext.AuthFilePath = authFilePath
+			} else if os.IsNotExist(err) {
+				logger.Info("no authentication information found for pulling image, proceeding without auth")
```
joelanford, line 242: Just had an interesting thought. This behavior means that an enterprising person could mount a file at `authFilePath`, NOT set `--global-pull-secret` and have this work.

If we want to avoid unintended usage like this, we may want to:
1. Make this code conditional on the `--global-pull-secret` flag being set, AND
2. Doing something like this to get a random file name every time we start:
   ```
   authFilePath := filepath.Join(os.TempDir(), fmt.Sprintf("%s-%s.json", authFilePrefix, apimachineryrand.String(8))
   ```

cmd/manager/main.go
```diff
@@ -308,17 +360,3 @@ func main() {
 		os.Exit(1)
 	}
```
joelanford, line 361:
```suggestion
	}
	if err := os.Remove(authFilePath); err != nil {
		setupLog.Error(err, "failed to cleanup temporary auth file")
		os.Exit(1)	
	}
```

cmd/manager/main.go
m1kola: Slightly related: I was actually thinking about us using an optional secret mount instead of a flag: https://kubernetes.io/docs/concepts/configuration/secret/#restriction-secret-must-exist

But I think this will limit users to having to create a global pull secret in the same namespace as operator-controller.

internal/controllers/secretsyncer_controller.go
anik120: Sure, looks cleaner than "secretsyncer_controller.go` 👍🏽

cmd/manager/main.go
anik120: >But I think this will limit users to having to create a global pull secret in the same namespace as operator-controller.

Yea that's not a viable option for us. 

Also incorporated Joe's suggestions.

cmd/manager/main.go
anik120: Done.

cmd/manager/main.go
```diff
@@ -67,7 +72,7 @@ var (
 	defaultSystemNamespace = "olmv1-system"
 )
 
-const authFilePath = "/etc/operator-controller/auth.json"
+const authFilePrefix = "operator-controller-global-pull-secrets.json"
```
everettraven, line 75: `authFilePrefix` might be a misleading name since this looks more like a file name which I would expect to be considered a "suffix" on the path.

anik120: Actually the `.json` should not have been there, the fmt.Sprintf directive down below already includes the `.json`. 

Should be fixed now, good catch! 

here's what it looks like now: https://go.dev/play/p/x35HITorSPj

cmd/manager/main.go
```diff
@@ -116,27 +123,51 @@ func main() {
...
+		cacheOptions.ByObject[&corev1.Secret{}] = crcache.ByObject{
+			Namespaces: map[string]crcache.Config{
+				globalPullSecretKey.Namespace: {
+					LabelSelector: k8slabels.Everything(),
+					FieldSelector: fields.SelectorFromSet(map[string]string{
+						"metadata.name": globalPullSecretKey.Name,
+					}),
+				},
+			},
+		}
```
everettraven, line 162: Does this also override the configuration for secrets in the `systemNamespace`?  IIUC we want to make sure we have access to all the release secrets that get created by the secret driver we use for partitioning helm release secrets for larger bundles.

everettraven (approved):

cmd/manager/main.go
everettraven: Approved pending an answer to this

cmd/manager/main.go
```diff
@@ -116,27 +123,51 @@ func main() {
...
+		globalPullSecretKey = &k8stypes.NamespacedName{Name: secretParts[1], Namespace: secretParts[0]}
+	}
+
 	if systemNamespace == "" {
 		systemNamespace = podNamespace()
 	}
 
 	setupLog.Info("set up manager")
+	cacheOptions := crcache.Options{
```
LalatenduMohanty, line 142: Nitpick: If you can add some comments around what the below code doing it will add a lot to the readability of the code.

cmd/manager/main.go
LalatenduMohanty: Can we mark this thread resolved?

cmd/manager/main.go
joelanford: Dirty secret: iirc, the release secret driver is still a live client. So checking whether that is working won't actually tell you anything.

anik120: Also I don't actually see any configuration for secretes in the `systemNamespace`. There's configuration for `&ocv1alpha1.ClusterExtension{}` and `&catalogd.ClusterCatalog{}`, but nothing for Secrets.

cmd/manager/main.go
anik120: I'm not convinced any further documentation is necessary. The `crcache.Options` struct is pretty well documented. Once the reader understands what the controller-runtime cache does and what `Options` can be used to configure, the code reads "configuring caching options for the objects we own + additional object Secret in a specific namespace".

everettraven (dismissed):

LalatenduMohanty (approved):

PR #1323, opened 2024-10-01
🐛 dynamicsource: fix race condition with wait and valid scheme

joelanford:
Also decreasing 1 second poll interval to 100ms. The `HasSynced` check that repeats every interval simply locks mutexes and checks variables available in memory, so it is extremely fast and has low over head. This decreased the test's average time from >1000ms to ~250ms 

# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/contentmanager/source/dynamicsource.go
+++ b/internal/contentmanager/source/dynamicsource.go
@@ -139,7 +139,7 @@ func (dis *dynamicInformerSource) Start(ctx context.Context, q workqueue.TypedRa
 			close(dis.syncedChan)
 		})
 
-		_ = wait.PollUntilContextCancel(dis.informerCtx, time.Second, true, func(_ context.Context) (bool, error) {
+		_ = wait.PollUntilContextCancel(dis.informerCtx, time.Millisecond*100, true, func(_ context.Context) (bool, error) {
 			if sharedIndexInf.HasSynced() {
 				syncOnce()
 				return true, nil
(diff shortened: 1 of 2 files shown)


LalatenduMohanty (approved):
/lgtm

internal/contentmanager/source/dynamicsource.go
```diff
@@ -139,7 +139,7 @@ func (dis *dynamicInformerSource) Start(ctx context.Context, q workqueue.TypedRa
 			close(dis.syncedChan)
 		})
 
-		_ = wait.PollUntilContextCancel(dis.informerCtx, time.Second, true, func(_ context.Context) (bool, error) {
+		_ = wait.PollUntilContextCancel(dis.informerCtx, time.Millisecond*100, true, func(_ context.Context) (bool, error) {
```
bentito, line 142: If this works, it seems like it's possibly just a temp fix. Seems like there might still be an undiagnosed concurrency issue.

joelanford: This wasn't the concurrency fix. 

The concurrency fix was to `WaitForSync` in the test, which forced the panic to happen.
And then the panic fix was to use `scheme.Scheme` instead of `runtime.NewScheme()` in the test.

_This_ change just helps `WaitForSync` have less latency.

bentito: ah, okay, I get it now.

PR #1330, opened 2024-10-03
🌱 Flakes: replace `require` with `assert` in retry funcs

m1kola:
# Description

When `require` assertions are not satisfied they fails a test immediately. This is not desired when used
with retry function such as `EventuallyWithT` since they are supposed to retry on a failure.

This can cause flakes when, for example, we are asserting on a resource condition. On the first hit the resource might not have a condition and `require` will fail `EventuallyWithT` despite the fact that the code is correct and the condition gets set eventually.

Example failures from today & yesterday:
* https://github.com/operator-framework/operator-controller/actions/runs/11123906109/job/30908338221
* https://github.com/operator-framework/operator-controller/actions/runs/11145592682/job/30975544168
* https://github.com/operator-framework/operator-controller/actions/runs/11147969636/job/30983531236
* https://github.com/operator-framework/operator-controller/actions/runs/11148330698/job/30984708630

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/test/e2e/cluster_extension_install_test.go
+++ b/test/e2e/cluster_extension_install_test.go
@@ -274,7 +274,9 @@ func TestClusterExtensionInstallRegistry(t *testing.T) {
 			require.EventuallyWithT(t, func(ct *assert.CollectT) {
 				assert.NoError(ct, c.Get(context.Background(), types.NamespacedName{Name: clusterExtension.Name}, clusterExtension))
 				cond := apimeta.FindStatusCondition(clusterExtension.Status.Conditions, ocv1alpha1.TypeProgressing)
-				require.NotNil(ct, cond)
+				if !assert.NotNil(ct, cond) {
+					return
+				}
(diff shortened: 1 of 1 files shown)


grokspawn:
Can we generalize the statement that we should not use `require` in `Eventually` expressions? 

I know this is just test code, but the extra lines are messing with my OCD. 😉 
Would it flow better to do something like (example from [these lines](https://github.com/operator-framework/operator-controller/pull/1330/files#diff-597f97f8a5e0b9fc20fbd288c0f75a734309ab56c52872f0e6b881522244cb73R277-R281)):

```go
				assert.NoError(ct, c.Get(context.Background(), types.NamespacedName{Name: clusterExtension.Name}, clusterExtension))
				cond := apimeta.FindStatusCondition(clusterExtension.Status.Conditions, ocv1alpha1.TypeProgressing)
				if assert.NotNil(ct, cond) {
				}
```

grokspawn (approved):

PR #1333, opened 2024-10-03
:sparkles: :book: Minor updates to API+GoDoc, Documentation, and removal of a no longer needed label selector

everettraven:
# Description
- Removes the `dependentSelector` label selector requirement placed as the default label selector for our controller-runtime cache. This was necessary when we used native controller-runtime logic for managing dependent resources, but is no longer necessary now that we have some custom implementations of controller-runtime abstractions that use ServiceAccounts to establish informers and manage dependent resources.
- Updates the GoDoc of the `ClusterExtension.Status.Conditions` field to reflect the new set of potential status conditions and reasons
- Removes no longer used condition reasons from the Go types
- Updates stale references to ClusterExtension and ClusterCatalog APIs in our documentation

## Reviewer Checklist

- [ ] API Go Documentation


cmd/manager/main.go
```diff
@@ -144,7 +135,7 @@ func main() {
 			DefaultNamespaces: map[string]crcache.Config{
 				systemNamespace: {LabelSelector: k8slabels.Everything()},
 			},
-			DefaultLabelSelector: dependentSelector,
+			DefaultLabelSelector: k8slabels.Everything(),
```
joelanford, line 138: I don't think we want everything. I think we want nothing else. We don't want to accidentally explode our cache with stuff we don't explicitly say we need.

If we just drop this field entirely, is that be semantically equivalent to "nothing else"? I can't remember. It might take a little experimentation.

joelanford (commented):
Just one fairly important question on the cache options.

cmd/manager/main.go
joelanford: Bryce and I chatted about this more and agreed that an explicit `k8slabels.Nothing()` is the DefaultLabelSelector that we want. That will ensure that nothing makes it into our cache unless we explicitly add an override LabelSelector for it elsewhere in the cache options.

joelanford (approved):
Looks good to me (other than the rebase that needs to happen).

joelanford (approved):

PR #1346, opened 2024-10-04
✨ Fix a flake in upgrade e2e

m1kola:
# Description

There is a potential for nil pointer dereference in the test. This change adds condition to check for nil.

Example failures:
* https://github.com/operator-framework/operator-controller/actions/runs/11168271140/job/31046293874?pr=1284

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/test/upgrade-e2e/post_upgrade_test.go
+++ b/test/upgrade-e2e/post_upgrade_test.go
@@ -86,7 +86,7 @@ func TestClusterExtensionAfterOLMUpgrade(t *testing.T) {
 		assert.Equal(ct, metav1.ConditionTrue, cond.Status)
 		assert.Equal(ct, ocv1alpha1.ReasonSuccess, cond.Reason)
 		assert.Contains(ct, cond.Message, "Installed bundle")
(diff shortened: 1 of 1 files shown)


test/upgrade-e2e/post_upgrade_test.go
```diff
@@ -86,7 +86,7 @@ func TestClusterExtensionAfterOLMUpgrade(t *testing.T) {
 		assert.Equal(ct, metav1.ConditionTrue, cond.Status)
 		assert.Equal(ct, ocv1alpha1.ReasonSuccess, cond.Reason)
 		assert.Contains(ct, cond.Message, "Installed bundle")
-		if assert.NotEmpty(ct, clusterExtension.Status.Install.Bundle) {
```
m1kola, line 89: `clusterExtension.Status.Install.Bundle` is not a pointer and it doesn't make much sense to check for it being empty: we can just check the value (`Version`) we care about after checking that `clusterExtension.Status.Install` is not `nil`.

joelanford (approved):

PR #1349, opened 2024-10-07
:sparkles: Add checks for helm deployment status

### LINKED ISSUE #1296: Ensure rollouts are completed before starting a new upgrade.
Interruptions during bundle reconciliation could lead to operator-controller not completing the current running rollout and instead pick up the next available release from the catalog.

Not rolling out a release completely, may lead to inconsistent results and unexpected behavior.

;tldr: Binary restarts and transient errors should not trigger re-resolve bundle images when a rollout is still in flight.

tmshort:
Fix #1296 

Don't allow an update to occur when the deployment is still installing.

# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/controllers/clusterextension_controller.go
+++ b/internal/controllers/clusterextension_controller.go
@@ -25,6 +25,7 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
+	"helm.sh/helm/v3/pkg/release"
 	"helm.sh/helm/v3/pkg/storage/driver"
 	"k8s.io/apimachinery/pkg/api/equality"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
@@ -246,6 +247,7 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
 			Ref: resolvedBundle.Image,
 		},
 	}
+
 	l.Info("unpacking resolved bundle")
 	unpackResult, err := r.Unpacker.Unpack(ctx, bundleSource)
 	if err != nil {
@@ -470,16 +472,26 @@ func (d *DefaultInstalledBundleGetter) GetInstalledBundle(ctx context.Context, e
 		return nil, err
 	}
 
-	release, err := cl.Get(ext.GetName())
+	rel, err := cl.Get(ext.GetName())
 	if err != nil && !errors.Is(err, driver.ErrReleaseNotFound) {
 		return nil, err
 	}
-	if release == nil {
+	if rel == nil {
 		return nil, nil
 	}
 
+	switch rel.Info.Status {
+	case release.StatusUnknown:
+		return nil, fmt.Errorf("installation status is unknown")
+	case release.StatusDeployed, release.StatusUninstalled, release.StatusSuperseded, release.StatusFailed:
+	case release.StatusUninstalling, release.StatusPendingInstall, release.StatusPendingRollback, release.StatusPendingUpgrade:
+		return nil, fmt.Errorf("installation is still pending: %s", rel.Info.Status)
+	default:
+		return nil, fmt.Errorf("unknown installation status: %s", rel.Info.Status)
+	}
+
 	return &ocv1alpha1.BundleMetadata{
-		Name:    release.Labels[labels.BundleNameKey],
-		Version: release.Labels[labels.BundleVersionKey],
+		Name:    rel.Labels[labels.BundleNameKey],
+		Version: rel.Labels[labels.BundleVersionKey],
 	}, nil
 }


internal/controllers/clusterextension_controller.go
```diff
@@ -470,16 +472,31 @@ func (d *DefaultInstalledBundleGetter) GetInstalledBundle(ctx context.Context, e
 		return nil, err
 	}
 
-	release, err := cl.Get(ext.GetName())
+	rel, err := cl.Get(ext.GetName())
```
everettraven, line 475: Instead of continuing to use the `Get()` functionality here (I know this is all that helm-operator-plugins provides us), maybe we should:
- Update the [helm-operator-plugins `ActionInterface`](https://github.com/operator-framework/helm-operator-plugins/blob/1017ef84e14e4a35a69f946894d6387ae8a0180a/pkg/client/actionclient.go#L53) to have a `List()` method that creates a new [`List` type](https://pkg.go.dev/helm.sh/helm/v3@v3.16.1/pkg/action#List) and runs it with the provided options. The `List` type looks like it has the ability to be configured to filter releases by their various states and even sort them.
- Pull in the change from helm-operator-plugins and use the `List()` method here to:
    - Filter by "Deployed" releases
    - Sort by date (might need to reverse sort as well, I'm not sure if sorting by date would put newest first)

WDYT?

everettraven: Ah, the [`Sorter` type](https://pkg.go.dev/helm.sh/helm/v3@v3.16.1/pkg/action#Sorter) is a uint with a possible value of sorting by date descending

everettraven: You can filter by release name as well. Another thing probably worth verifying here, do we store history of the helm releases in our custom storage driver? If we don't, then should we?

everettraven: Also, should we be using only the ClusterExtension name or should we have something more deterministic based on ClusterExtension name + version?

tmshort: I think you might be getting a bit out-of-scope with history.

internal/controllers/clusterextension_controller.go
```diff
@@ -470,16 +472,31 @@ func (d *DefaultInstalledBundleGetter) GetInstalledBundle(ctx context.Context, e
...
 		return nil, nil
 	}
 
+	switch {
+	case rel.Info.Status == release.StatusUnknown:
+		return nil, fmt.Errorf("installation status is unknown")
+	case rel.Info.Status == release.StatusDeployed:
+	case rel.Info.Status == release.StatusUninstalled:
+	case rel.Info.Status == release.StatusSuperseded:
+	case rel.Info.Status == release.StatusFailed:
```
everettraven, line 489: If the release is failed, do we want to allow it to be overridden?

everettraven: Based on my understanding of https://github.com/operator-framework/operator-controller/issues/1296 I would assume we would not want to do this.

internal/controllers/clusterextension_controller.go
```diff
@@ -470,16 +472,31 @@ func (d *DefaultInstalledBundleGetter) GetInstalledBundle(ctx context.Context, e
...
+	case rel.Info.Status == release.StatusUninstalling:
+		return nil, fmt.Errorf("installation is still pending: %s", rel.Info.Status)
+	case rel.Info.Status.IsPending():
+		return nil, fmt.Errorf("installation is still pending: %s", rel.Info.Status)
+	default:
+		return nil, fmt.Errorf("unknown installation status: %s", rel.Info.Status)
+	}
```
everettraven, line 496: Nit: combining cases might make this a little more terse. Also it might make sense to not even enter the switch if the status is still pending.
```suggestion
        if rel.Info.Status.IsPending() {
                return nil, fmt.Errorf("installation is still pending: %s", rel.Info.Status)
        }

	switch rel.Info.Status {
	case release.StatusUnknown:
		return nil, fmt.Errorf("installation status is unknown")
	case release.StatusDeployed, release.StatusUninstalled, release.StatusSuperseded, release.StatusFailed:
	case rel.Info.Status == release.StatusUninstalling:
		return nil, fmt.Errorf("installation is still pending: %s", rel.Info.Status)
	default:
		return nil, fmt.Errorf("unknown installation status: %s", rel.Info.Status)
	}
```

tmshort: Not a fan of the longer lines. Splitting out the `IsPending()` makes it seem as though it's separate when it's not; it's just checking the set of `StatusPendingXxxxx` values.
Also, this happens to match the order they are defined by Helm.

internal/controllers/clusterextension_controller.go
tmshort: It's in a terminal state; and if an upgrade is the only possible way to fix it, then we need to allow it to happen. Otherwise, the installation will remain in a failed state and be unable to be replaced without manual intervention (i.e. removal).

tmshort:
ping @thetechnick

LalatenduMohanty (approved):
/lgtm

PR #1359, opened 2024-10-09
🌱 Use catalog availability

bentito:
# Description

## Reviewer Checklist

- [ ] API Go Documentation


internal/resolve/catalog.go
```diff
@@ -233,18 +234,50 @@ type CatalogWalkFunc func(context.Context, *catalogd.ClusterCatalog, *declcfg.De
...
+			}
+
 			if walkErr := f(ctx, cat, fbc, fbcErr); walkErr != nil {
 				return walkErr
 			}
 		}
+
+		// If no catalogs were processed at all, return a 'no catalogs' error
+		if !processedCatalogs {
+			return fmt.Errorf("no enabled catalogs found for package: %s", packageName)
```
joelanford, line 273: This error message is misleading. It makes it sound like there might be some enabled catalogs, just none that are enabled that contain `packageName`. But the `packageName` isn't really relevant, right? If this value is false, it literally means that all catalogs are disabled.

```suggestion
			return errors.New("no enabled catalogs found")
```

internal/resolve/catalog.go
```diff
@@ -233,18 +234,50 @@ type CatalogWalkFunc func(context.Context, *catalogd.ClusterCatalog, *declcfg.De
...
+		}
+
+		// Return an error if no valid bundle was found in any processed catalog
+		if !bundleFound {
+			return fmt.Errorf("no bundles found for package: %s", packageName)
+		}
```
joelanford, line 279: This code is simply the catalog walking code, and is decoupled from the concept of finding/resolving bundles. It is _used_ by the resolver.

But maybe the terminology is confusing me? We call `getPackage` which returns the entire FBC for `packageName` from the catalog, so if we keep this, it should really be called `foundPackage`.

But I'm not sure we need this variable or error at all. The walk function (`f`), receives the packageName, fbc, and fbcErr and can decide how to handle any potential result of `getPackage`.

internal/resolve/catalog_test.go
```diff
@@ -963,3 +963,150 @@ func TestMultipleChannels(t *testing.T) {
...
+	}
+
+	ce := buildFooClusterExtension(pkgName, []string{}, ">=1.0.0", ocv1alpha1.UpgradeConstraintPolicyCatalogProvided)
+	gotBundle, gotVersion, _, err := r.Resolve(context.Background(), ce, nil)
+	require.NoError(t, err)
+	require.NotNil(t, gotBundle)
+	require.Equal(t, bsemver.MustParse("3.0.0"), *gotVersion)
+}
+
+func TestPriorityRespectedWithDisabledCatalogs(t *testing.T) {
```
joelanford, line 1030: This test seems unnecessary to me since priority and availability are orthogonal concepts (right?)

We ignore the disabled catalogs in the `CatalogWalker`, so the resolver (where we consider priority) never even knows of the existence of the disabled catalog.

internal/resolve/catalog_test.go
```diff
@@ -963,3 +963,150 @@ func TestMultipleChannels(t *testing.T) {
...
+		}, nil
+	}
+
+	getPackage := func(ctx context.Context, cat *catalogd.ClusterCatalog, packageName string) (*declcfg.DeclarativeConfig, error) {
+		return genPackage(pkgName), nil
+	}
+
+	r := CatalogResolver{
+		WalkCatalogsFunc: CatalogWalker(listCatalogs, getPackage),
+	}
```
joelanford, line 1055: It seems to me like if we managed to break the fact that disabled catalogs are ignored, this test would still pass?
1. The catalog with higher priority is enabled, so we'd resolve from it anyway.
2. If the disabled catalog was higher priority (and we have broken logic for ignoring disabled catalogs), `getPackage` would be called for both catalogs, which means that the resolver would still find the same bundle.
3. If we also broke the logic for considering enabled catalogs, this test looks like it would _still_ pass because `getPackage` would be called for the disabled catalog and it would still return the same result that the assertions expect.

internal/resolve/catalog_test.go
```diff
@@ -963,3 +963,150 @@ func TestMultipleChannels(t *testing.T) {
...
+				Spec: catalogd.ClusterCatalogSpec{
+					Availability: "Disabled",
+				},
+			},
+		}, nil
+	}
+
+	getPackage := func(ctx context.Context, cat *catalogd.ClusterCatalog, packageName string) (*declcfg.DeclarativeConfig, error) {
+		return genPackage(pkgName), nil
+	}
```
joelanford, line 986: The expectation is that this function would never even be called. Perhaps we should panic?

```suggestion
	getPackage := func(ctx context.Context, cat *catalogd.ClusterCatalog, packageName string) (*declcfg.DeclarativeConfig, error) {
		panic("getPackage should never be called when all catalogs are disabled")
	}
```

internal/resolve/catalog_test.go
```diff
@@ -963,3 +963,150 @@ func TestMultipleChannels(t *testing.T) {
...
+			{
+				Spec: catalogd.ClusterCatalogSpec{
+					Availability: "Disabled",
+				},
+			},
+		}, nil
+	}
+
+	getPackage := func(ctx context.Context, cat *catalogd.ClusterCatalog, packageName string) (*declcfg.DeclarativeConfig, error) {
+		return genPackage(pkgName), nil
```
joelanford, line 1016: In order to properly test that we ignore disabled catalogs, don't we need to the enabled and disabled catalogs to have disjoint sets of bundle versions, with the disable catalog having higher versions than the enabled catalog? That way if we ever somehow stop ignoring disabled catalogs, resolution would incorrectly resolve to a bundle we find in the disabled catalog and this test would fail.

internal/resolve/catalog_test.go
```diff
@@ -963,3 +963,150 @@ func TestMultipleChannels(t *testing.T) {
...
+	require.Equal(t, bsemver.MustParse("3.0.0"), *gotVersion)
+}
+
+func TestCatalogWithoutAvailabilityIsEnabled(t *testing.T) {
+	pkgName := randPkg()
+	listCatalogs := func(ctx context.Context, options ...client.ListOption) ([]catalogd.ClusterCatalog, error) {
+		return []catalogd.ClusterCatalog{
+			{
+				Spec: catalogd.ClusterCatalogSpec{
+					Priority: 1, // No Availability field set
```
joelanford, line 1070: This should be impossible according to our CRD validation. It seems like `""` should be treated as an error.

internal/resolve/catalog.go
bentito: Yeah I agree. I'm removing the tracking variables and associated errors, the CatalogWalker function no longer makes decisions about whether any bundles were found or whether any catalogs were processed. It simply walks through the catalogs, now aware of catalog availability.

internal/resolve/catalog_test.go
bentito: modifying the test so it now effectively verifies that disabled catalogs are ignored even when they have higher priority and versions. If the logic to skip disabled catalogs breaks, the test will fail either by panic or by selecting the wrong bundle.

internal/resolve/catalog_test.go
bentito: Implemented the panic in getPackage when called for a disabled catalog to ensure that disabled catalogs are not processed.

internal/resolve/catalog_test.go
bentito: Yes, we've modified the `getPackage` function to return a higher version for the disabled catalog and a lower version for the enabled catalog. This ensures that if the disabled catalog is incorrectly processed, the test will fail by selecting the higher version from the disabled catalog.

internal/resolve/catalog_test.go
bentito: We are now explicitly setting the Availability field for both catalogs to comply with CRD validation.

internal/resolve/catalog.go
```diff
@@ -231,20 +232,33 @@ func isDeprecated(bundle declcfg.Bundle, deprecation *declcfg.Deprecation) bool
...
 		catalogs, err := listCatalogs(ctx, catalogListOpts...)
 		if err != nil {
 			return fmt.Errorf("error listing catalogs: %w", err)
 		}
 
 		for i := range catalogs {
 			cat := &catalogs[i]
+
+			// skip catalogs with Availability set to "Disabled"
+			if cat.Spec.Availability == "Disabled" {
```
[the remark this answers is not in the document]
bentito: Yeah, that makes send, updated the code to remove disabled catalogs using slices.DeleteFunc before entering the loop.

bentito:
Going to leave this as draft with test failed until https://github.com/operator-framework/catalogd/pull/421 (its companion) merges.

LalatenduMohanty (dismissed):

internal/resolve/catalog.go
```diff
@@ -3,6 +3,7 @@ package resolve
 import (
 	"context"
 	"fmt"
+	"sigs.k8s.io/controller-runtime/pkg/log"
```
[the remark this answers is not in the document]
bentito: forgot to gci it _doing_

PR #1379, opened 2024-10-15
:sparkles: Use Helm List operator to determine Deployed status

tmshort:
Latest fixes for #1296 

# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/controllers/clusterextension_controller.go
+++ b/internal/controllers/clusterextension_controller.go
@@ -84,7 +84,7 @@ type Applier interface {
 }
 
 type InstalledBundleGetter interface {
-	GetInstalledBundle(ctx context.Context, ext *ocv1alpha1.ClusterExtension) (*ocv1alpha1.BundleMetadata, error)
+	GetInstalledBundle(ctx context.Context, ext *ocv1alpha1.ClusterExtension) (*InstalledBundle, error)
 }
 
 //+kubebuilder:rbac:groups=olm.operatorframework.io,resources=clusterextensions,verbs=get;list;watch;update;patch
@@ -206,19 +206,22 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
 	installedBundle, err := r.InstalledBundleGetter.GetInstalledBundle(ctx, ext)
 	if err != nil {
 		setInstallStatus(ext, nil)
-		// TODO: use Installed=Unknown
-		setInstalledStatusConditionFailed(ext, err.Error())
-		setStatusProgressing(ext, err)
+		setInstalledStatusConditionUnknown(ext, err.Error())
+		setStatusProgressing(ext, errors.New("retrying to get installed bundle"))
 		return ctrl.Result{}, err
 	}
 
 	// run resolution
 	l.Info("resolving bundle")
-	resolvedBundle, resolvedBundleVersion, resolvedDeprecation, err := r.Resolver.Resolve(ctx, ext, installedBundle)
+	var bm *ocv1alpha1.BundleMetadata
+	if installedBundle != nil {
+		bm = &installedBundle.BundleMetadata
+	}
+	resolvedBundle, resolvedBundleVersion, resolvedDeprecation, err := r.Resolver.Resolve(ctx, ext, bm)
 	if err != nil {
 		// Note: We don't distinguish between resolution-specific errors and generic errors
-		setInstallStatus(ext, nil)
 		setStatusProgressing(ext, err)
+		setInstalledStatusFromBundle(ext, installedBundle)
 		ensureAllConditionsWithReason(ext, ocv1alpha1.ReasonFailed, err.Error())
 		return ctrl.Result{}, err
 	}
@@ -255,6 +258,7 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
 		// installed since we intend for the progressing condition to replace the resolved condition
 		// and will be removing the .status.resolution field from the ClusterExtension status API
 		setStatusProgressing(ext, wrapErrorWithResolutionInfo(resolvedBundleMetadata, err))
+		setInstalledStatusFromBundle(ext, installedBundle)
 		return ctrl.Result{}, err
 	}
 
@@ -268,9 +272,10 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
 	}
 
 	storeLbls := map[string]string{
-		labels.BundleNameKey:    resolvedBundle.Name,
-		labels.PackageNameKey:   resolvedBundle.Package,
-		labels.BundleVersionKey: resolvedBundleVersion.String(),
+		labels.BundleNameKey:      resolvedBundle.Name,
+		labels.PackageNameKey:     resolvedBundle.Package,
+		labels.BundleVersionKey:   resolvedBundleVersion.String(),
+		labels.BundleReferenceKey: resolvedBundle.Image,
 	}
 
 	l.Info("applying bundle contents")
@@ -286,18 +291,17 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
 	managedObjs, _, err := r.Applier.Apply(ctx, unpackResult.Bundle, ext, objLbls, storeLbls)
 	if err != nil {
 		setStatusProgressing(ext, wrapErrorWithResolutionInfo(resolvedBundleMetadata, err))
-		// If bundle is not already installed, set Installed status condition to False
-		if installedBundle == nil {
-			setInstalledStatusConditionFailed(ext, err.Error())
-		}
+		// Now that we're actually trying to install, use the error
+		setInstalledStatusFromBundle(ext, installedBundle)
 		return ctrl.Result{}, err
 	}
 
-	installStatus := &ocv1alpha1.ClusterExtensionInstallStatus{
-		Bundle: resolvedBundleMetadata,
+	newInstalledBundle := &InstalledBundle{
+		BundleMetadata: resolvedBundleMetadata,
+		Image:          resolvedBundle.Image,
 	}
-	setInstallStatus(ext, installStatus)
-	setInstalledStatusConditionSuccess(ext, fmt.Sprintf("Installed bundle %s successfully", resolvedBundle.Image))
+	// Successful install
+	setInstalledStatusFromBundle(ext, newInstalledBundle)
 
 	l.Info("watching managed objects")
 	cache, err := r.Manager.Get(ctx, ext)
@@ -466,32 +470,45 @@ type DefaultInstalledBundleGetter struct {
 	helmclient.ActionClientGetter
 }
 
-func (d *DefaultInstalledBundleGetter) GetInstalledBundle(ctx context.Context, ext *ocv1alpha1.ClusterExtension) (*ocv1alpha1.BundleMetadata, error) {
+type InstalledBundle struct {
+	ocv1alpha1.BundleMetadata
+	Image string
+}
+
+func (d *DefaultInstalledBundleGetter) GetInstalledBundle(ctx context.Context, ext *ocv1alpha1.ClusterExtension) (*InstalledBundle, error) {
 	cl, err := d.ActionClientFor(ctx, ext)
 	if err != nil {
 		return nil, err
 	}
 
-	rel, err := cl.Get(ext.GetName())
+	relhis, err := cl.History(ext.GetName())
 	if err != nil && !errors.Is(err, driver.ErrReleaseNotFound) {
 		return nil, err
 	}
-	if rel == nil {
+	if len(relhis) == 0 {
 		return nil, nil
 	}
 
-	switch rel.Info.Status {
-	case release.StatusUnknown:
-		return nil, fmt.Errorf("installation status is unknown")
-	case release.StatusDeployed, release.StatusUninstalled, release.StatusSuperseded, release.StatusFailed:
-	case release.StatusUninstalling, release.StatusPendingInstall, release.StatusPendingRollback, release.StatusPendingUpgrade:
-		return nil, fmt.Errorf("installation is still pending: %s", rel.Info.Status)
-	default:
-		return nil, fmt.Errorf("unknown installation status: %s", rel.Info.Status)
+	// relhis[0].Info.Status is the status of the most recent install attempt.
+	// But we need to look for the most-recent _Deployed_ release
+	for _, rel := range relhis {
+		if rel.Info != nil && rel.Info.Status == release.StatusDeployed {
+			// If there are blank values, we should consider this as not installed
+			if n, ok := rel.Labels[labels.BundleNameKey]; !ok || n == "" {
+				return nil, nil
+			}
+			if v, ok := rel.Labels[labels.BundleVersionKey]; !ok || v == "" {
+				return nil, nil
+			}
+			// Not checking BundleReferenceKey, as it's new; upgrade test would fail
+			return &InstalledBundle{
+				BundleMetadata: ocv1alpha1.BundleMetadata{
+					Name:    rel.Labels[labels.BundleNameKey],
+					Version: rel.Labels[labels.BundleVersionKey],
+				},
+				Image: rel.Labels[labels.BundleReferenceKey],
+			}, nil
+		}
 	}
-
-	return &ocv1alpha1.BundleMetadata{
-		Name:    rel.Labels[labels.BundleNameKey],
-		Version: rel.Labels[labels.BundleVersionKey],
-	}, nil
+	return nil, nil
 }
--- a/internal/controllers/clusterextension_controller_test.go
+++ b/internal/controllers/clusterextension_controller_test.go
@@ -12,6 +12,9 @@ import (
 	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"helm.sh/helm/v3/pkg/chart"
+	"helm.sh/helm/v3/pkg/release"
+	"helm.sh/helm/v3/pkg/storage/driver"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -21,12 +24,14 @@ import (
 	crfinalizer "sigs.k8s.io/controller-runtime/pkg/finalizer"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
+	helmclient "github.com/operator-framework/helm-operator-plugins/pkg/client"
 	"github.com/operator-framework/operator-registry/alpha/declcfg"
 
 	ocv1alpha1 "github.com/operator-framework/operator-controller/api/v1alpha1"
 	"github.com/operator-framework/operator-controller/internal/conditionsets"
 	"github.com/operator-framework/operator-controller/internal/controllers"
 	"github.com/operator-framework/operator-controller/internal/finalizers"
+	"github.com/operator-framework/operator-controller/internal/labels"
 	"github.com/operator-framework/operator-controller/internal/resolve"
 	"github.com/operator-framework/operator-controller/internal/rukpak/source"
 )
@@ -392,7 +397,10 @@ func TestClusterExtensionApplierFailsWithBundleInstalled(t *testing.T) {
 		cache: &MockManagedContentCache{},
 	}
 	reconciler.InstalledBundleGetter = &MockInstalledBundleGetter{
-		bundle: &ocv1alpha1.BundleMetadata{Name: "prometheus.v1.0.0", Version: "1.0.0"},
+		bundle: &controllers.InstalledBundle{
+			BundleMetadata: ocv1alpha1.BundleMetadata{Name: "prometheus.v1.0.0", Version: "1.0.0"},
+			Image:          "quay.io/operatorhubio/prometheus@fake1.0.0",
+		},
 	}
 	reconciler.Applier = &MockApplier{
 		objs: []client.Object{},
@@ -745,7 +753,10 @@ func TestClusterExtensionDeleteFinalizerFails(t *testing.T) {
 		cache: &MockManagedContentCache{},
 	}
 	reconciler.InstalledBundleGetter = &MockInstalledBundleGetter{
-		bundle: &ocv1alpha1.BundleMetadata{Name: "prometheus.v1.0.0", Version: "1.0.0"},
+		bundle: &controllers.InstalledBundle{
+			BundleMetadata: ocv1alpha1.BundleMetadata{Name: "prometheus.v1.0.0", Version: "1.0.0"},
+			Image:          "quay.io/operatorhubio/prometheus@fake1.0.0",
+		},
 	}
 	err = reconciler.Finalizers.Register(fakeFinalizer, finalizers.FinalizerFunc(func(ctx context.Context, obj client.Object) (crfinalizer.Result, error) {
 		return crfinalizer.Result{}, errors.New(finalizersMessage)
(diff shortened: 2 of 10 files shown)


internal/controllers/clusterextension_controller.go
```diff
@@ -84,7 +84,7 @@ type Applier interface {
 }
 
 type InstalledBundleGetter interface {
-	GetInstalledBundle(ctx context.Context, ext *ocv1alpha1.ClusterExtension) (*ocv1alpha1.BundleMetadata, error)
+	GetInstalledBundle(ctx context.Context, ext *ocv1alpha1.ClusterExtension) (*ocv1alpha1.BundleMetadata, release.Status, error)
```
joelanford, line 87: I'd say only add this if we actually need it. Looks like we aren't using it yet.

And I don't think we need to. IIRC, yesterday we said this function would return the most recent revision that has status==Deployed, otherwise return `nil` (or an error).

internal/controllers/clusterextension_controller.go
```diff
@@ -203,21 +203,34 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
...
+		installStatus := &ocv1alpha1.ClusterExtensionInstallStatus{
+			Bundle: *installedBundle,
+		}
+		setInstallStatus(ext, installStatus)
+		if !apimeta.IsStatusConditionTrue(ext.Status.Conditions, ocv1alpha1.TypeInstalled) {
+			setInstalledStatusConditionSuccess(ext, "Installed extension successfully")
+		}
+	} else {
+		setInstallStatus(ext, nil)
+		apimeta.RemoveStatusCondition(&ext.Status.Conditions, ocv1alpha1.TypeInstalled)
```
joelanford, line 226: Unfortunately, I don't think we can directly update the status this way. Here's the problematic scenario.
1. We start `reconcile` and we have `status.Conditions[type=Installed]` is `True`.
2. We don't find a deployed bundle, so we remove the `Installed` status condition (really, we should be setting this to `False` instead of removing it)
3. We continue on, resolve, unpack, and install successfully, resulting in us setting `status.Conditions[type=Installed]` back to `True`.

The problem is that this will cause `lastTransitionTime` to update without an actual transition of the Status. (It stays `True` in etcd the whole time).

We need to figure out a way to ensure that we only touch each condition type once in a given run of `reconcile`.

internal/controllers/clusterextension_controller.go
```diff
@@ -466,32 +479,36 @@ type DefaultInstalledBundleGetter struct {
...
+func (d *DefaultInstalledBundleGetter) GetInstalledBundle(ctx context.Context, ext *ocv1alpha1.ClusterExtension) (*ocv1alpha1.BundleMetadata, release.Status, error) {
+	status := release.StatusUnknown
 	cl, err := d.ActionClientFor(ctx, ext)
 	if err != nil {
-		return nil, err
+		return nil, status, err
 	}
 
-	rel, err := cl.Get(ext.GetName())
+	relhis, err := cl.History(ext.GetName())
```
joelanford, line 489: Refactor suggestion:
```go
relhis, err := cl.History(ext.GetName())
if errors.Is(err, driver.ErrReleaseNotFound)
	return nil, nil
}
if err != nil {
	return nil, err
}
for _, rel := range relhis {
	if rel.Info.Status == release.StatusDeployed {
 		return &ocv1alpha1.BundleMetadata{
 			Name:    rel.Labels[labels.BundleNameKey],
 			Version: rel.Labels[labels.BundleVersionKey],
	 	}, nil
 	}
 }
return nil, nil
```

internal/controllers/clusterextension_controller.go
joelanford: Maybe we need a helper function that can encapsulate the update logic and then we still end up calling that helper function in all of the various error handling `if err  != nil` blocks? We'd pass `installedBundle` into it to handle setting the condition and other status fields appropriately.

internal/controllers/clusterextension_controller.go
```diff
@@ -206,8 +206,7 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
 	installedBundle, err := r.InstalledBundleGetter.GetInstalledBundle(ctx, ext)
 	if err != nil {
 		setInstallStatus(ext, nil)
-		// TODO: use Installed=Unknown
-		setInstalledStatusConditionFailed(ext, err.Error())
+		setInstalledStatusConditionUnknown(ext, err.Error())
 		setStatusProgressing(ext, err)
```
joelanford, line 210: Nit-ish: should the progressing message end up being something like: "trying again to get installed bundle" rather than a repeat of the message in the Installed status condition?

internal/controllers/clusterextension_controller.go
```diff
@@ -217,7 +216,7 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
 	resolvedBundle, resolvedBundleVersion, resolvedDeprecation, err := r.Resolver.Resolve(ctx, ext, installedBundle)
 	if err != nil {
 		// Note: We don't distinguish between resolution-specific errors and generic errors
-		setInstallStatus(ext, nil)
+		setInstalledStatusFromBundle(ext, installedBundle, err)
```
joelanford, line 219: Since we successfully got the installed bundle (or know there isn't an installed bundle), it seems like there shouldn't be any error messages in the `Installed` condition. It would be one of:

```yaml
- type: Installed
  status: True
  reason: Success
  # This is interesting because I don't think we store the reference in the 
  # installed bundle metadata, maybe we should so that the message is
  # able to be consistently recreated.  Just need to be careful not to assume
  # the reference is an image because it could (in the future) come from
  # somewhere else.
  message: "Installed bundle <bundleImageRef> successfully"
```

Or:
```yaml
- type: Installed
  status: False
  reason: Failed
  message: "no installed bundle found"
```

So should be pass a `nil` error here or `errors.New("no installed bundle found")`?

internal/controllers/clusterextension_controller.go
tmshort: That's exactly what I did (before reading this comment)

internal/controllers/clusterextension_controller.go
tmshort: I was thinking that... but wondering if it makes it harder to read...

internal/controllers/clusterextension_controller.go
```diff
@@ -286,10 +286,7 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
 	managedObjs, _, err := r.Applier.Apply(ctx, unpackResult.Bundle, ext, objLbls, storeLbls)
 	if err != nil {
 		setStatusProgressing(ext, wrapErrorWithResolutionInfo(resolvedBundleMetadata, err))
-		// If bundle is not already installed, set Installed status condition to False
-		if installedBundle == nil {
-			setInstalledStatusConditionFailed(ext, err.Error())
-		}
+		setInstalledStatusFromBundle(ext, installedBundle, err)
```
joelanford, line 289: It's too bad we can't make lines 293-297 reuse the `setInstalledStatusFromBundle`. What would it take to do that? Seems like we'd need to construct an `installedBundle` out of the `resolvedBundle`

internal/controllers/clusterextension_controller.go
joelanford: Similar comment for other places that we call this function after a successful call to `GetInstalledBundle`

internal/controllers/clusterextension_controller.go
joelanford: IMO, its easier with my suggestion, but I realize this is subjective. 

- The first two `if` conditionals seem more straightforward to grok.
- The `return` inside the loop is _maybe_ slightly more difficult to read, but it eliminates the extra variable, `break` and assignment that would be required to return after the loop completes.

internal/controllers/clusterextension_controller.go
tmshort: I thought about that too...

tmshort: However, the `setInstalledStatusFromBundle()` function will try to avoid changing the timestamp of the condition if it's already true; so it's not constantly overwriting the timestamp (or the message) when the prior bundle was (detected as) installed.
Lines 293-297 force an update of the condition.

internal/controllers/clusterextension_controller.go
```diff
@@ -472,26 +469,24 @@ func (d *DefaultInstalledBundleGetter) GetInstalledBundle(ctx context.Context, e
...
 
-	switch rel.Info.Status {
-	case release.StatusUnknown:
-		return nil, fmt.Errorf("installation status is unknown")
-	case release.StatusDeployed, release.StatusUninstalled, release.StatusSuperseded, release.StatusFailed:
-	case release.StatusUninstalling, release.StatusPendingInstall, release.StatusPendingRollback, release.StatusPendingUpgrade:
-		return nil, fmt.Errorf("installation is still pending: %s", rel.Info.Status)
-	default:
-		return nil, fmt.Errorf("unknown installation status: %s", rel.Info.Status)
+	// TODO: relhis[0].Info.Status is the status of the highest semver install attempt.
```
joelanford, line 480: Not necessarily highest semver. Just the most recent install attempt. A failed downgrade would result in the first release in the slice having a lower semver than subsequent releases in the list.

internal/controllers/clusterextension_controller.go
tmshort: No longer an issue

internal/controllers/clusterextension_controller.go
joelanford: Sorry, I think the error message should go on the Installed condition and the "retrying to get install bundle" should go on the Progressing condition.

The Installed condition is unknown _because_ of that error.

internal/controllers/clusterextension_controller_test.go
```diff
@@ -1400,3 +1411,122 @@ func TestSetDeprecationStatus(t *testing.T) {
...
+
+func (mag *MockActionGetter) Uninstall(name string, opts ...helmclient.UninstallOption) (*release.UninstallReleaseResponse, error) {
+	return nil, nil
+}
+
+func (mag *MockActionGetter) Reconcile(rel *release.Release) error {
+	return nil
+}
+
+func TestGetInstalledBundleHistory(t *testing.T) {
```
joelanford, line 1449: Nit: This seems ripe for a table-driven test sort of thing where we declare a test case struct with the mock history/error, and expected return values. Then a loop over the test cases where we setup the mock, call the getter, and then make the assertion.

tmshort:
I was able to run `tilt ci` locally, and it seemed to work. I am seeing this warning in the actual CI run (but not locally), and not sure if it related to the failure?
```
operator-con… │ [manager] Version of Delve is too old for Go version go1.23.2 (maximum supported version 1.22, suppress this error with --check-go-version=false)
```

internal/controllers/clusterextension_controller.go
```diff
@@ -206,19 +206,24 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
 	installedBundle, err := r.InstalledBundleGetter.GetInstalledBundle(ctx, ext)
 	if err != nil {
 		setInstallStatus(ext, nil)
-		// TODO: use Installed=Unknown
-		setInstalledStatusConditionFailed(ext, err.Error())
-		setStatusProgressing(ext, err)
+		// The error is put into Progressing
+		setInstalledStatusConditionUnknown(ext, err.Error())
+		setStatusProgressing(ext, errors.New("retrying to get installed bundle"))
```
everettraven, line 211: Seems backwards based on the added comment?

internal/controllers/clusterextension_controller.go
```diff
@@ -206,19 +206,24 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
...
 	}
 
 	// run resolution
 	l.Info("resolving bundle")
-	resolvedBundle, resolvedBundleVersion, resolvedDeprecation, err := r.Resolver.Resolve(ctx, ext, installedBundle)
+	var bm *ocv1alpha1.BundleMetadata
+	if installedBundle != nil {
+		bm = &installedBundle.BundleMetadata
+	}
+	resolvedBundle, resolvedBundleVersion, resolvedDeprecation, err := r.Resolver.Resolve(ctx, ext, bm)
```
everettraven, line 221: Any particular reason for this change? is `nil` an unacceptable value in the resolver?

internal/controllers/clusterextension_controller.go
```diff
@@ -286,18 +294,16 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
 	managedObjs, _, err := r.Applier.Apply(ctx, unpackResult.Bundle, ext, objLbls, storeLbls)
 	if err != nil {
 		setStatusProgressing(ext, wrapErrorWithResolutionInfo(resolvedBundleMetadata, err))
-		// If bundle is not already installed, set Installed status condition to False
-		if installedBundle == nil {
-			setInstalledStatusConditionFailed(ext, err.Error())
-		}
+		// Now that we're actually trying to install, use the error
+		setInstalledStatusFromBundle(ext, installedBundle, err)
```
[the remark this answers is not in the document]
everettraven: >In this case, the Apply() failed. If we already had a bundle installed, we should continue to report that? But how do we report the error?

My understanding tracks with this continuing to be the `installedBundle` since we want to continue reporting that we successfully installed the previous bundle but are failing to upgrade. My understanding is that the error is reported through the progressing condition a bit higher up in this conditional.

internal/controllers/clusterextension_controller.go
```diff
@@ -283,21 +291,20 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
 	// to ensure exponential backoff can occur:
 	//   - Permission errors (it is not possible to watch changes to permissions.
 	//     The only way to eventually recover from permission errors is to keep retrying).
+	installedBundle = &InstalledBundle{
+		BundleMetadata: resolvedBundleMetadata,
+		Image:          resolvedBundle.Image,
+	}
```
everettraven, line 297: This should probably be moved below the `r.Applier.Apply()` call since we haven't successfully installed until after the `Apply` step has completed. We also don't want to overwrite the information of a previously successful installation.

internal/controllers/clusterextension_controller.go
```diff
@@ -283,21 +291,20 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
...
 		return ctrl.Result{}, err
 	}
 
-	installStatus := &ocv1alpha1.ClusterExtensionInstallStatus{
-		Bundle: resolvedBundleMetadata,
-	}
-	setInstallStatus(ext, installStatus)
-	setInstalledStatusConditionSuccess(ext, fmt.Sprintf("Installed bundle %s successfully", resolvedBundle.Image))
+	// Successful install
+	setInstalledStatusFromBundle(ext, installedBundle, nil)
```
everettraven, line 307: In this case, `installedBundle` is actually the `resolvedBundle` right?

internal/controllers/clusterextension_controller.go
```diff
@@ -466,32 +473,46 @@ type DefaultInstalledBundleGetter struct {
 	helmclient.ActionClientGetter
 }
 
-func (d *DefaultInstalledBundleGetter) GetInstalledBundle(ctx context.Context, ext *ocv1alpha1.ClusterExtension) (*ocv1alpha1.BundleMetadata, error) {
+type InstalledBundle struct {
+	ocv1alpha1.BundleMetadata
+	Image string
+}
```
everettraven, line 479: Any particular reason why this type is added and used only within the `ClusterExtension` controller instead of extending the `ocv1alpha1.BundleMetadata` type to also include the image?

internal/controllers/clusterextension_controller.go
```diff
@@ -466,32 +473,46 @@ type DefaultInstalledBundleGetter struct {
...
+	// the status of an upgrade attempt.
+	for _, rel := range relhis {
+		if rel.Info != nil && rel.Info.Status == release.StatusDeployed {
+			// If there are blank values, we should consider this as not installed
+			if n, ok := rel.Labels[labels.BundleNameKey]; !ok || n == "" {
+				return nil, nil
+			}
+			if v, ok := rel.Labels[labels.BundleVersionKey]; !ok || v == "" {
+				return nil, nil
+			}
```
everettraven, line 506: Should we return early in these cases or continue looking for the next successfully deployed release that contains these labels?

Is there any case where we wouldn't have set the labels on a release or is the absence of these labels indicative of a much bigger issue?

internal/controllers/clusterextension_controller_test.go
```diff
@@ -1400,3 +1411,137 @@ func TestSetDeprecationStatus(t *testing.T) {
...
+
+	for _, tst := range mag {
+		t.Log(tst.description)
+		getter.ActionClientGetter = &tst
+		md, err := getter.GetInstalledBundle(context.Background(), &ext)
+		if tst.expectedError {
+			require.Error(t, err)
+		} else {
+			require.NoError(t, err)
+		}
```
everettraven, line 1539: Nit: could probably simplify this to something like:
```suggestion
		require.Equal(t, err != nil, tst.expectedError, fmt.Sprintf("error state doesn't match expected. should error? %v, err: %w", tst.expectedError, err))
```

internal/controllers/clusterextension_controller_test.go
```diff
@@ -1400,3 +1411,137 @@ func TestSetDeprecationStatus(t *testing.T) {
...
+		if tst.expectedBundle == nil {
+			require.Nil(t, md)
+		} else {
+			require.NotNil(t, md)
+			require.Equal(t, tst.expectedBundle, md)
+		}
```
everettraven, line 1545: Nit: I think this could just be a simple equality check:
```suggestion
		require.Equal(t, tst.expectedBundle, md)
```

internal/controllers/common_controller.go
```diff
@@ -26,6 +27,31 @@ import (
...
+func setInstalledStatusFromBundle(ext *ocv1alpha1.ClusterExtension, installedBundle *InstalledBundle, err error) {
+	// Something is installed
+	if installedBundle != nil {
+		installStatus := &ocv1alpha1.ClusterExtensionInstallStatus{
+			Bundle: installedBundle.BundleMetadata,
+		}
+		setInstallStatus(ext, installStatus)
+		// If both conditions occur, this is a failed Apply when a bundle was already installed
+		if err != nil {
+			setInstalledStatusConditionFailed(ext, err.Error())
```
everettraven, line 40: Wouldn't we want to preserve the historical installation information on a failed "upgrade" attempt, which is what I interpret a "failed Apply when a bundle was already installed" to mean.

internal/controllers/clusterextension_controller.go
tmshort: @joelanford wanted the retry status under Progressing

internal/controllers/clusterextension_controller.go
tmshort: `nil` is acceptable, and was possible if no installed bundle was found. We need to replicate that.

internal/controllers/clusterextension_controller.go
tmshort: @joelanford asked for this to be up here, so that we can indicate a failed install, since we are about to `Apply()`

internal/controllers/clusterextension_controller.go
tmshort: Yes, I used the same name, but I could use a different named variable.

internal/controllers/clusterextension_controller.go
tmshort: @joelanford didn't want to update the external API (which is what would happen if `ocv1alpha1.BundleMetadata` is updated, as it's part of the `ClusterExtension.status`. We didn't want to put the Image location in there, only the metadata.

internal/controllers/clusterextension_controller.go
```diff
@@ -466,32 +473,46 @@ type DefaultInstalledBundleGetter struct {
...
+			// Not checking BundleReferenceKey, as it's new; upgrade test would fail
```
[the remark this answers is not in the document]
tmshort: It fails the upgrade tests, as these fields are not necessarily included in the prior version of OLMv1 installs  (without the latest helm-operator-plugin).

But yes, we could remove these checks. I would actually prefer to remove them once we push a release, to make sure they really aren't needed.

internal/controllers/clusterextension_controller.go
tmshort: These checks should be able to be removed once we do a release, as they're only necessary for the upgrade.

internal/controllers/clusterextension_controller_test.go
tmshort: Perhaps, not sure that's actually simpler... it's one line of code, but it takes a bit more time to grok.

internal/controllers/common_controller.go
tmshort: I was concerned about this, but the problem is that we lose the fact that we failed an upgrade. I'm not sure which is better. But this is what @joelanford suggested.

internal/controllers/clusterextension_controller.go
tmshort: We didn't previously check for for empty, but we still might want to consider doing it.

internal/controllers/clusterextension_controller.go
tmshort: Renamed to `toBeInstalledBundle`. Can't use the old name of `resolvedBundle` because it's a different type. 

Changing `Resolve()` to use the `InstalledBundle` could be possible, but that would make this change much larger.

internal/controllers/common_controller.go
tmshort: Another way to look at it: if we tried an install, we store the bundle information. Should we not have the same information on an upgrade?
I almost think we need to have an upgrade field...

tmshort: There's additional discussion here: https://github.com/operator-framework/operator-controller/pull/1379#discussion_r1805660826

internal/controllers/clusterextension_controller.go
tmshort: See: https://github.com/operator-framework/operator-controller/pull/1379#discussion_r1805462712

internal/controllers/clusterextension_controller.go
tmshort: To clarify, the type has changed, and can no longer be input into `Resolve()` directly. Since `installedBundle` could have been `nil`, we want to be be able to replicate that. If `Resolve()` is modified to accept the `InstalledBundle` type, this conversion would be unnecessary.

internal/controllers/clusterextension_controller_test.go
tmshort: Ok, found a better way!

internal/controllers/clusterextension_controller.go
tmshort: The change would be:
`8 files changed, 70 insertions(+), 57 deletions(-)`

internal/controllers/clusterextension_controller.go
everettraven: That's fine, but the comment seems to be stating the opposite. If this is the desired ordering, we should probably update the comment such that it reflects what the code actually says.

tmshort: Forgot to remove that comment...

internal/controllers/clusterextension_controller.go
everettraven: Makes sense. I had a brief chat with @joelanford yesterday and we touched on this and the desire for the public api to have an explicit design for how we include references without constraining ourselves to specific assumptions.

internal/controllers/common_controller.go
tmshort: Also note, that this change (no longer preserving `Installed` after the `Apply()`) is a separate fixup commit, so it can be looked at in isolation.

everettraven: If a ClusterExtension has been installed successfully in the past, we should keep the installed condition as is and include the upgrade failure information in the progressing condition. The resolution information can be included in the error message so that a user can better connect the dots that an upgrade failed and what bundles were involved.

internal/controllers/clusterextension_controller.go
joelanford: After v1, it seems there are three possibilities for missing labels on the release:
1. We make a new release that adds a new label (say in `1.2`), but we just upgraded from `1.1`, so any releases that exist prior to `1.2` would not have all of `1.2`'s expected labels.
2. Something deleted our labels.
3. We introduced a bug somehow where we didn't set the labels.

joelanford, line 506: I personally think we should return an error if the expected labels are missing. Or the resolver should return an error if the `installedBundle` is non-nil, but then expected struct fields are unset.

I don't think it makes sense to assume "not installed" because clearly _something_ is installed.

tmshort: This occurs on an upgrade; seemingly. But we can't tell if it's a temporary property of the lack of a helm-operater-plugin update, or something else.

internal/controllers/common_controller.go
```diff
@@ -26,6 +27,26 @@ import (
...
+		setInstalledStatusConditionSuccess(ext, fmt.Sprintf("Installed bundle %s successfully", installedBundle.Image))
+		return
+	}
+	// Nothing is installed, if there's no error, it means no installed bundle was found
+	setInstallStatus(ext, nil)
+	if err != nil {
+		setInstalledStatusConditionFailed(ext, err.Error())
+	} else {
+		setInstalledStatusConditionFailed(ext, "No installed bundle found")
+	}
```
everettraven, line 47: Nit: 
```suggestion
	}
	
	setInstalledStatusConditionFailed(ext, "No installed bundle found")
```

internal/controllers/clusterextension_controller.go
```diff
@@ -466,32 +470,46 @@ type DefaultInstalledBundleGetter struct {
...
-	case release.StatusUnknown:
-		return nil, fmt.Errorf("installation status is unknown")
-	case release.StatusDeployed, release.StatusUninstalled, release.StatusSuperseded, release.StatusFailed:
-	case release.StatusUninstalling, release.StatusPendingInstall, release.StatusPendingRollback, release.StatusPendingUpgrade:
-		return nil, fmt.Errorf("installation is still pending: %s", rel.Info.Status)
-	default:
-		return nil, fmt.Errorf("unknown installation status: %s", rel.Info.Status)
+	// TODO: relhis[0].Info.Status is the status of the most recent install attempt.
+	// This might be useful informaton if it's not release.StatusDeployed, in telling us
+	// the status of an upgrade attempt.
```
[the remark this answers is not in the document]
tmshort: Not sure if we actually plan to do anything there. I think it makes more sense to remove the comment, since I think it would be an issue that would not be worked on.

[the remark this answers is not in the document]
tmshort: Removed the TODO, and changed it to indicate that we're looking for the first Deployed.

internal/controllers/clusterextension_controller.go
tmshort: I plan to clean this up after a release and a clean upgrade-e2e.

internal/controllers/clusterextension_controller.go
```diff
@@ -286,18 +291,17 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1alp
 	managedObjs, _, err := r.Applier.Apply(ctx, unpackResult.Bundle, ext, objLbls, storeLbls)
 	if err != nil {
 		setStatusProgressing(ext, wrapErrorWithResolutionInfo(resolvedBundleMetadata, err))
-		// If bundle is not already installed, set Installed status condition to False
-		if installedBundle == nil {
-			setInstalledStatusConditionFailed(ext, err.Error())
-		}
+		// Now that we're actually trying to install, use the error
+		setInstalledStatusFromBundle(ext, installedBundle, err)
```
joelanford, line 295: Yesterday we said:
- Failed install
   - status.conditions[Installed] = false
- Failed upgrade
   - status.conditions[Installed] = true, message indicates successfully installed bundle ref

So should this be:

```suggestion
		// Now that we're actually trying to install, use the error
		setInstalledStatusFromBundle(ext, installedBundle, nil)
```

So that the `Installed`

tmshort: No, we want the `err` in there, the code is correct.

A failed (initial) install means that `installedBundle` is nil, and `err` is not nil, so the following happens:
```
setInstallStatus(ext, nil)
setInstalledStatusConditionFailed(ext, err.Error())
```

A failed upgrade means that `installedBundle` is not nil, and `err` is not nil, so the following happens:
```
setInstallStatus(ext, installStatus)
setInstalledStatusConditionSuccess(ext, fmt.Sprintf("Installed bundle %s successfully", installedBundle.Image))
```

tmshort: If `err` is nil in here, then the result will be a message of "No install bundle found", rather than the installation error from `err`.

joelanford (dismissed):

joelanford (approved):

PR #1404, opened 2024-10-22
⚠️ updates from api audit

grokspawn:
This is API changes in preparation for our v1.0 release, mostly from the openshift API team's [audit](https://github.com/openshift/api/pull/2067) but also from feedback from the maintainers and community. 

It is expected that this PR contains breaking changes as we prepare (tightening types, enhancing docs, etc.) for the release.

# Description

## Reviewer Checklist

- [ ] API Go Documentation


api/v1alpha1/clusterextension_types.go
```diff
@@ -496,6 +507,7 @@ type ClusterExtensionStatus struct {
 	// +patchStrategy=merge
 	// +listType=map
 	// +listMapKey=type
+	// +kubebuilder:validation:Required
```
everettraven, line 510: This should be optional?

config/base/crd/bases/olm.operatorframework.io_clusterextensions.yaml
```diff
@@ -574,6 +574,8 @@ spec:
                 required:
                 - bundle
                 type: object
+            required:
+            - conditions
```
everettraven, line 578: I would have expected no diff in our generated CRD by only adding the required tags to the already required fields

grokspawn: Ditto! And yet ... 
It's even more apparently-empty in the generated docs!

everettraven: I left another comment that I think the conditions should be optional. We include `omitempty` in the JSON tag. If it should be required then this change is fine.

api/v1alpha1/clusterextension_types.go
grokspawn: Yes.  A copy/paste error.

api/v1alpha1/clusterextension_types.go
```diff
@@ -258,7 +264,7 @@ type CatalogSource struct {
 	// For more information on semver, please see https://semver.org/
 	//
 	//+kubebuilder:validation:MaxLength:=64
-	//+kubebuilder:validation:Pattern=`^(\s*(=||!=|>|<|>=|=>|<=|=<|~|~>|\^)\s*(v?(0|[1-9]\d*|[x|X|\*])(\.(0|[1-9]\d*|x|X|\*]))?(\.(0|[1-9]\d*|x|X|\*))?(-([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?(\+([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?)\s*)((?:\s+|,\s*|\s*\|\|\s*)(=||!=|>|<|>=|=>|<=|=<|~|~>|\^)\s*(v?(0|[1-9]\d*|x|X|\*])(\.(0|[1-9]\d*|x|X|\*))?(\.(0|[1-9]\d*|x|X|\*]))?(-([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?(\+([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?)\s*)*$`
+	//+kubebuilder:validation.XValidation:rule="self.matches(r'^(\s*(=||!=|>|<|>=|=>|<=|=<|~|~>|\^)\s*(v?(0|[1-9]\d*|[x|X|\*])(\.(0|[1-9]\d*|x|X|\*]))?(\.(0|[1-9]\d*|x|X|\*))?(-([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?(\+([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?)\s*)((?:\s+|,\s*|\s*\|\|\s*)(=||!=|>|<|>=|=>|<=|=<|~|~>|\^)\s*(v?(0|[1-9]\d*|x|X|\*])(\.(0|[1-9]\d*|x|X|\*))?(\.(0|[1-9]\d*|x|X|\*]))?(-([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?(\+([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?)\s*)*$')"", message="invalid version expression in the catalog source"
```
perdasilva, line 267: Is the error message correct here?

api/v1alpha1/clusterextension_types.go
```diff
@@ -80,14 +82,15 @@ const SourceTypeCatalog = "Catalog"
 type SourceConfig struct {
```
everettraven, line 82: I believe we need to update the CEL validation here to match the changes in https://github.com/operator-framework/catalogd/pull/443/files#diff-d446855f2187e6d61c4719c3202fcb21e4aea9b09e09f9e1d01419f341d8fc9fR233

It looks like Joel had requested this as well: https://github.com/openshift/api/pull/2067/files#r1812412946

api/v1alpha1/clusterextension_types.go
```diff
@@ -80,14 +82,15 @@ const SourceTypeCatalog = "Catalog"
...
 	// When this field is set to "Catalog", information for determining the appropriate
 	// bundle of content to install will be fetched from ClusterCatalog resources existing
 	// on the cluster. When using the Catalog sourceType, the catalog field must also be set.
 	//
 	// +unionDiscriminator
 	// +kubebuilder:validation:Enum:="Catalog"
+	// +kubebuilder:validation:Required
 	SourceType string `json:"sourceType"`
 
 	// catalog is used to configure how information is sourced from a catalog. This field must be defined when sourceType is set to "Catalog",
```
everettraven, line 96: Should probably update this to reflect any updates we make to the CEL validation following my comment on this structs validations

api/v1alpha1/clusterextension_types.go
```diff
@@ -130,6 +133,7 @@ type ClusterExtensionInstallConfig struct {
 	//+kubebuilder:validation:Pattern:=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
 	//+kubebuilder:validation:MaxLength:=63
 	//+kubebuilder:validation:XValidation:rule="self == oldSelf",message="namespace is immutable"
+	//+kubebuilder:validation:Required
 	Namespace string `json:"namespace"`
```
everettraven, line 137: There were a few comments from Joel related to the GoDoc for this field that I don't see reflected here. Did we decide against accepting those suggestions or making adjustments to clarify some of the questions asked?

Ref:
- https://github.com/openshift/api/pull/2067/files#r1812416875
- https://github.com/openshift/api/pull/2067/files#r1812425452
- https://github.com/openshift/api/pull/2067/files#r1812430084
- https://github.com/openshift/api/pull/2067/files#r1812416228

api/v1alpha1/clusterextension_types.go
```diff
@@ -130,6 +133,7 @@ type ClusterExtensionInstallConfig struct {
 	//+kubebuilder:validation:Pattern:=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
```
everettraven, line 133: Should we use CEL here instead of pattern validation?

api/v1alpha1/clusterextension_types.go
```diff
@@ -69,6 +70,7 @@ type ClusterExtensionSpec struct {
 	//    namespace: example-namespace
 	//    serviceAccount:
 	//      name: example-sa
+	// +kubebuilder:validation:Required
```
joelanford, line 73: Add an extra new line above the Required comment marker?

api/v1alpha1/clusterextension_types.go
joelanford: We don't need to say "in the catalog source" because the error emitted by the validator includes the parent path leading up to this field.

```suggestion
	//+kubebuilder:validation.XValidation:rule="self.matches(r'^(\s*(=||!=|>|<|>=|=>|<=|=<|~|~>|\^)\s*(v?(0|[1-9]\d*|[x|X|\*])(\.(0|[1-9]\d*|x|X|\*]))?(\.(0|[1-9]\d*|x|X|\*))?(-([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?(\+([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?)\s*)((?:\s+|,\s*|\s*\|\|\s*)(=||!=|>|<|>=|=>|<=|=<|~|~>|\^)\s*(v?(0|[1-9]\d*|x|X|\*])(\.(0|[1-9]\d*|x|X|\*))?(\.(0|[1-9]\d*|x|X|\*]))?(-([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?(\+([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?)\s*)*$')"", message="invalid version expression"
```

api/v1alpha1/clusterextension_types.go
```diff
@@ -386,28 +393,30 @@ type PreflightConfig struct {
...
 	CRDUpgradeSafety *CRDUpgradeSafetyPreflightConfig `json:"crdUpgradeSafety"`
 }
 
 // CRDUpgradeSafetyPreflightConfig is the configuration for CRD upgrade safety preflight check.
 type CRDUpgradeSafetyPreflightConfig struct {
-	// policy is used to configure the state of the CRD Upgrade Safety pre-flight check.
+	// enforcement is used to configure the state of the CRD Upgrade Safety pre-flight check.
 	//
 	// This field is required when the spec.install.preflight.crdUpgradeSafety field is
 	// specified.
```
joelanford, line 405: I think I saw a comment from Joel about not needing the conditional statement here. We can just say "enforcement is required" unconditionally (it is implicit that it being required is conditional on the parent struct being specified).

api/v1alpha1/clusterextension_types.go
```diff
@@ -258,14 +264,14 @@ type CatalogSource struct {
 	// For more information on semver, please see https://semver.org/
 	//
 	//+kubebuilder:validation:MaxLength:=64
-	//+kubebuilder:validation:Pattern=`^(\s*(=||!=|>|<|>=|=>|<=|=<|~|~>|\^)\s*(v?(0|[1-9]\d*|[x|X|\*])(\.(0|[1-9]\d*|x|X|\*]))?(\.(0|[1-9]\d*|x|X|\*))?(-([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?(\+([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?)\s*)((?:\s+|,\s*|\s*\|\|\s*)(=||!=|>|<|>=|=>|<=|=<|~|~>|\^)\s*(v?(0|[1-9]\d*|x|X|\*])(\.(0|[1-9]\d*|x|X|\*))?(\.(0|[1-9]\d*|x|X|\*]))?(-([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?(\+([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?)\s*)*$`
+	//+kubebuilder:validation.XValidation:rule="self.matches(r'^(\s*(=||!=|>|<|>=|=>|<=|=<|~|~>|\^)\s*(v?(0|[1-9]\d*|[x|X|\*])(\.(0|[1-9]\d*|x|X|\*]))?(\.(0|[1-9]\d*|x|X|\*))?(-([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?(\+([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?)\s*)((?:\s+|,\s*|\s*\|\|\s*)(=||!=|>|<|>=|=>|<=|=<|~|~>|\^)\s*(v?(0|[1-9]\d*|x|X|\*])(\.(0|[1-9]\d*|x|X|\*))?(\.(0|[1-9]\d*|x|X|\*]))?(-([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?(\+([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?)\s*)*$')"", message="invalid version expression in the catalog source"
```
everettraven, line 267:
```suggestion
	//+kubebuilder:validation:XValidation:rule="self.matches(r'^(\s*(=||!=|>|<|>=|=>|<=|=<|~|~>|\^)\s*(v?(0|[1-9]\d*|[x|X|\*])(\.(0|[1-9]\d*|x|X|\*]))?(\.(0|[1-9]\d*|x|X|\*))?(-([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?(\+([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?)\s*)((?:\s+|,\s*|\s*\|\|\s*)(=||!=|>|<|>=|=>|<=|=<|~|~>|\^)\s*(v?(0|[1-9]\d*|x|X|\*])(\.(0|[1-9]\d*|x|X|\*))?(\.(0|[1-9]\d*|x|X|\*]))?(-([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?(\+([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?)\s*)*$')"", message="invalid version expression in the catalog source"
```

api/v1alpha1/clusterextension_types.go
everettraven: The error message here definitely seems a bit vague to me. Could we update the message to also include some helpful information for crafting a valid version expression?

api/v1alpha1/clusterextension_types.go
```diff
@@ -140,6 +144,7 @@ type ClusterExtensionInstallConfig struct {
 	// the ServiceAccount provided via this field should be configured with the
 	// appropriate permissions to perform the necessary operations on all the
 	// resources that are included in the bundle of content being applied.
+	//+kubebuilder:validation:Required
 	ServiceAccount ServiceAccountReference `json:"serviceAccount"`
 
 	// preflight is an optional field that can be used to configure the preflight checks run before installation or upgrade of the content for the package specified in the packageName field.
```
everettraven, line 150: Do we need to be more descriptive about what the default preflight checks are?

Context: https://github.com/openshift/api/pull/2067/files#r1812435988

internal/applier/helm.go
```diff
@@ -56,6 +57,25 @@ type Helm struct {
 	Preflights         []Preflight
 }
 
+// shouldSkipPreflight is a helper to determine if the preflight check is CRDUpgradeSafety AND
+// if it is set to enforcement None.
+func shouldSkipPreflight(preflight Preflight, ctx context.Context, ext *ocv1alpha1.ClusterExtension, state string) bool {
+	l := log.FromContext(ctx)
+	if ext.Spec.Install.Preflight != nil && ext.Spec.Install.Preflight.CRDUpgradeSafety != nil {
+		if _, ok := preflight.(*crdupgradesafety.Preflight); ok {
```
joelanford, line 65: Maybe move this line out to the calling function, probably a more appropriate place for it, and might solve the linter concern around nested blocks.

api/v1alpha1/clusterextension_types.go
```diff
@@ -181,6 +186,7 @@ type CatalogSource struct {
 	//+kubebuilder:validation:MaxLength:=253
 	//+kubebuilder:validation:Pattern:=^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$
```
everettraven, line 187: Should this also be CEL validation?

api/v1alpha1/clusterextension_types.go
```diff
@@ -258,14 +264,14 @@ type CatalogSource struct {
...
 	//
 	//+kubebuilder:validation:MaxLength:=64
-	//+kubebuilder:validation:Pattern=`^(\s*(=||!=|>|<|>=|=>|<=|=<|~|~>|\^)\s*(v?(0|[1-9]\d*|[x|X|\*])(\.(0|[1-9]\d*|x|X|\*]))?(\.(0|[1-9]\d*|x|X|\*))?(-([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?(\+([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?)\s*)((?:\s+|,\s*|\s*\|\|\s*)(=||!=|>|<|>=|=>|<=|=<|~|~>|\^)\s*(v?(0|[1-9]\d*|x|X|\*])(\.(0|[1-9]\d*|x|X|\*))?(\.(0|[1-9]\d*|x|X|\*]))?(-([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?(\+([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?)\s*)*$`
+	//+kubebuilder:validation.XValidation:rule="self.matches(r'^(\s*(=||!=|>|<|>=|=>|<=|=<|~|~>|\^)\s*(v?(0|[1-9]\d*|[x|X|\*])(\.(0|[1-9]\d*|x|X|\*]))?(\.(0|[1-9]\d*|x|X|\*))?(-([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?(\+([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?)\s*)((?:\s+|,\s*|\s*\|\|\s*)(=||!=|>|<|>=|=>|<=|=<|~|~>|\^)\s*(v?(0|[1-9]\d*|x|X|\*])(\.(0|[1-9]\d*|x|X|\*))?(\.(0|[1-9]\d*|x|X|\*]))?(-([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?(\+([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?)\s*)*$')"", message="invalid version expression in the catalog source"
 	//+optional
 	Version string `json:"version,omitempty"`
 
 	// channels is an optional reference to a set of channels belonging to
```
everettraven, line 271: There is an open-question on if we should set a max number of items in the list of channels: https://github.com/openshift/api/pull/2067/files#r1812455558

I think it is reasonable to have a max number of entries in the slice and Joe's suggested 256 seems fine to me as well.

api/v1alpha1/clusterextension_types.go
grokspawn: What do you have in mind?  I was specifically modeling after https://github.com/openshift/api/blob/4f6053f954b069f7fea5788d4307f4213db53678/config/v1alpha1/types_image_policy.go#L56-L60

api/v1alpha1/clusterextension_types.go
```diff
@@ -181,6 +186,7 @@ type CatalogSource struct {
 	//+kubebuilder:validation:MaxLength:=253
 	//+kubebuilder:validation:Pattern:=^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$
 	//+kubebuilder:validation:XValidation:rule="self == oldSelf",message="packageName is immutable"
+	//+kubebuilder:validation:Required
 	PackageName string `json:"packageName"`
```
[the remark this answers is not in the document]
grokspawn: Is [has one](https://github.com/operator-framework/operator-controller/pull/1404/files#diff-8e9c0eb955489f1a05a95a75a25842a010fdc02029ce06aae7d2163b895c17cbR163), and I thought that the crux of confusion lay in the pattern getting barfed back to users on errors.   
I'll look back over the comments to see if I misunderstood something.

api/v1alpha1/clusterextension_types.go
grokspawn: [[1]](https://github.com/openshift/api/pull/2067#discussion_r1812368363) comment that mentioning optional/required isn't necessary on status

Happy to make the change, but I didn't see a comment like this (maybe yet!).

api/v1alpha1/clusterextension_types.go
LalatenduMohanty: Looks like most of the comments are addressed . Also https://github.com/openshift/api/pull/2067/files#r1812416228 is an optional change. Either way it is fine.

api/v1alpha1/clusterextension_types.go
LalatenduMohanty: Do we want to add a validation for max number of entries?

LalatenduMohanty: Never mind , we have added `+kubebuilder:validation:MaxItems:=256`

internal/applier/helm.go
grokspawn: I resolved this by externalizing the conditions somewhat, but still putting all the 'CRDUpgradeSafety' checks in one place.  
My assertion is still that this kind of thing should be exposed via an interface.

api/v1alpha1/clusterextension_types.go
grokspawn: Yeah, @everettraven was way ahead of my WIP PR here. 😄 
I think I've caught up now, and took the 'WIP' off for a full review.

api/v1alpha1/clusterextension_types.go
```diff
@@ -69,100 +68,93 @@ type ClusterExtensionSpec struct {
...
 
 // ClusterExtensionInstallConfig is a union which selects the clusterExtension installation config.
 // ClusterExtensionInstallConfig requires the namespace and serviceAccount which should be used for the installation of packages.
+//
 // +union
 type ClusterExtensionInstallConfig struct {
-	// namespace is a reference to the Namespace in which the bundle of
-	// content for the package referenced in the packageName field will be applied.
+	// namespace designates the kubernetes Namespace where bundle content
+	// for the 'packageName' package field will be applied and the necessary
```
everettraven, line 110: I know Joel was a bit confused by the previous verbiage, but now reading this I'm getting confused. Maybe something like:
```suggestion
	// for the package, referenced in the packageName field, will be applied and the necessary
```

api/v1alpha1/clusterextension_types.go
```diff
@@ -69,100 +68,93 @@ type ClusterExtensionSpec struct {
...
+	// start and end with an alphanumeric character, and be no longer than 63 characters
 	//
 	// [RFC 1123]: https://tools.ietf.org/html/rfc1123
 	//
-	//+kubebuilder:validation:Pattern:=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
-	//+kubebuilder:validation:MaxLength:=63
-	//+kubebuilder:validation:XValidation:rule="self == oldSelf",message="namespace is immutable"
+	// +kubebuilder:validation:MaxLength:=63
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="namespace is immutable"
+	// +kubebuilder:validation:XValidation:rule="self.matches(\"^[a-z0-9]([-a-z0-9]*[a-z0-9])?$\")",message="namespace must be a valid DNS1123 label"
```
everettraven, line 123: I know we describe what a RFC 1123 DNS label is in the description of this field, but a user that inputs an incorrect value won't know how to fix "namespace must be a valid DNS1123 label" without some additional searching. Should we add some additional information in the message such that a user can more easily understand what the validation criteria is from the admission error message they receive?

Maybe something like:
```suggestion
	// +kubebuilder:validation:XValidation:rule="self.matches(\"^[a-z0-9]([-a-z0-9]*[a-z0-9])?$\")",message="namespace must be a valid DNS1123 label. Valid DNS1123 labels contain only lowercase alphanumeric characters or hyphens, starting and ending with an alphanumeric character"
```

api/v1alpha1/clusterextension_types.go
```diff
@@ -178,9 +170,11 @@ type CatalogSource struct {
 	//
 	// [RFC 1123]: https://tools.ietf.org/html/rfc1123
 	//
-	//+kubebuilder:validation:MaxLength:=253
-	//+kubebuilder:validation:Pattern:=^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$
-	//+kubebuilder:validation:XValidation:rule="self == oldSelf",message="packageName is immutable"
+	// +kubebuilder:validation.Required
+	// +kubebuilder:validation:MaxLength:=253
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="packageName is immutable"
+	// +kubebuilder:validation:XValidation:rule="self.matches(\"^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$\")",message="packageName must be a valid DNS1123 subdomain"
```
everettraven, line 176: Similar comment re: more helpful error message that includes information on what a valid DNS1123 subdomain means

api/v1alpha1/clusterextension_types.go
```diff
@@ -303,9 +295,10 @@ type CatalogSource struct {
 	//
 	// [RFC 1123]: https://tools.ietf.org/html/rfc1123
 	//
-	//+kubebuilder:validation:items:MaxLength:=253
-	//+kubebuilder:validation:items:Pattern:=^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$
-	//+optional
+	// +kubebuilder:validation:items:MaxLength:=253
+	// +kubebuilder:validation:MaxItems:=256
```
everettraven, line 299: I don't see mention of the max items in the GoDoc for this field. Should we include it to ensure it is communicated to a user doing something like `kubectl explain ...`?

api/v1alpha1/clusterextension_types.go
```diff
@@ -370,48 +361,45 @@ type ServiceAccountReference struct {
...
-	//+kubebuilder:validation:XValidation:rule="self == oldSelf",message="name is immutable"
+	// +kubebuilder:validation:MaxLength:=253
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="name is immutable"
+	// +kubebuilder:validation:XValidation:rule="self.matches(\"^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$\")",message="name must be a valid DNS1123 subdomain"
+	// +kubebuilder:validation:Required
 	Name string `json:"name"`
 }
 
 // PreflightConfig holds the configuration for the preflight checks.  If used, at least one preflight check must be non-nil.
+//
```
LalatenduMohanty, line 372: Empty line?

api/v1alpha1/clusterextension_types.go
```diff
@@ -303,9 +295,10 @@ type CatalogSource struct {
 	//
 	// [RFC 1123]: https://tools.ietf.org/html/rfc1123
 	//
-	//+kubebuilder:validation:items:MaxLength:=253
-	//+kubebuilder:validation:items:Pattern:=^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$
-	//+optional
+	// +kubebuilder:validation:items:MaxLength:=253
+	// +kubebuilder:validation:MaxItems:=256
+	// +kubebuilder:validation:items:XValidation:rule="self.matches(\"^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$\")",message="channels entries must be valid DNS1123 subdomains"
```
everettraven, line 300: Similar comment re: more helpful error message clarifying what makes a valid DNS1123 subdomain

api/v1alpha1/clusterextension_types.go
```diff
@@ -370,48 +361,45 @@ type ServiceAccountReference struct {
...
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="name is immutable"
+	// +kubebuilder:validation:XValidation:rule="self.matches(\"^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$\")",message="name must be a valid DNS1123 subdomain"
+	// +kubebuilder:validation:Required
 	Name string `json:"name"`
 }
 
 // PreflightConfig holds the configuration for the preflight checks.  If used, at least one preflight check must be non-nil.
+//
 // +kubebuilder:validation:XValidation:rule="has(self.crdUpgradeSafety)",message="at least one of [crdUpgradeSafety] are required when preflight is specified"
 type PreflightConfig struct {
```
LalatenduMohanty, line 374: We do not have optional tag here because we have `// +kubebuilder:validation:XValidation:rule="has(self.crdUpgradeSafety)"`?

api/v1alpha1/clusterextension_types.go
```diff
@@ -370,48 +361,45 @@ type ServiceAccountReference struct {
 	//
 	// [RFC 1123]: https://tools.ietf.org/html/rfc1123
 	//
-	//+kubebuilder:validation:MaxLength:=253
-	//+kubebuilder:validation:Pattern:=^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$
-	//+kubebuilder:validation:XValidation:rule="self == oldSelf",message="name is immutable"
+	// +kubebuilder:validation:MaxLength:=253
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="name is immutable"
+	// +kubebuilder:validation:XValidation:rule="self.matches(\"^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$\")",message="name must be a valid DNS1123 subdomain"
```
everettraven, line 366: Ditto re: helpful error message for valid DNS1123 subdomain criteria

api/v1alpha1/clusterextension_types.go
```diff
@@ -370,48 +361,45 @@ type ServiceAccountReference struct {
...
 	//
-	// When set to "Enabled", the CRD Upgrade Safety pre-flight check will be run when
+	// When set to "Strict", the CRD Upgrade Safety pre-flight check will be run when
 	// performing an upgrade operation.
 	//
-	//+kubebuilder:validation:Enum:="Enabled";"Disabled"
-	//+kubebuilder:default:=Enabled
-	Policy CRDUpgradeSafetyPolicy `json:"policy"`
+	// +kubebuilder:validation:Enum:="None";"Strict"
+	// +kubebuilder:default:=Strict
```
everettraven, line 397: I think @joelanford had originally called this out, but does this default actually do anything? We mark this as required but set a default value which I think tends to only ever be done on optional fields.

api/v1alpha1/clusterextension_types.go
```diff
@@ -428,110 +416,109 @@ const (
...
+	// +kubebuilder:validation:XValidation:rule="self.matches(\"^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$\")",message="packageName must be a valid DNS1123 subdomain"
 	Name string `json:"name"`
-	// version is a required field and is a reference
-	// to the version that this bundle represents
+
+	// version is a required field and is a reference to the version that this bundle represents
+	// version follows the semantic versioning standard as defined in https://semver.org/, but permits a leading 'v' character.
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:XValidation:rule="self.matches(\"^v?([0-9]+)(\\\\.[0-9]+)?(\\\\.[0-9]+)?(-([-0-9A-Za-z]+(\\\\.[-0-9A-Za-z]+)*))?(\\\\+([-0-9A-Za-z]+(-\\\\.[-0-9A-Za-z]+)*))?\")",message="version name must foo"
```
everettraven, line 440: Version name must foo ... fighters?

api/v1alpha1/clusterextension_types.go
```diff
@@ -428,110 +416,109 @@ const (
...
 type BundleMetadata struct {
-	// name is a required field and is a reference
-	// to the name of a bundle
+	// name is required and follows the DNS subdomain standard
+	// as defined in [RFC 1123]. It must contain only lowercase alphanumeric characters,
+	// hyphens (-) or periods (.), start and end with an alphanumeric character,
+	// and be no longer than 253 characters.
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:XValidation:rule="self.matches(\"^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$\")",message="packageName must be a valid DNS1123 subdomain"
```
everettraven, line 433: Ditto re: more helpful message

api/v1alpha1/clusterextension_types.go
grokspawn: Correct.  There is a discussion in the other PR about how this `CRDUpgradeSafety` is a pointer, but also a required type, where we assert that we'll change this in the short term to make this an "at least one non-nil field" enforcement... and right now we only have the CRDUpgradeSafety type right now.

api/v1alpha1/clusterextension_types.go
grokspawn: How important is this, given that the godoc has the details?  I'm curious because my first impulse when I get a message like this would be to web search 'RFC 1123 label', which would pretty quickly get to the answer.

api/v1alpha1/clusterextension_types.go
grokspawn: I tried to be consistent and always include a line between godoc and kubebuilder statements.

api/v1alpha1/clusterextension_types.go
grokspawn: This was left half-done I think.  WDYT about something like
> namespace designates the kubernetes Namespace which will be used to install the package, and where the necessary service account can be found.

?

api/v1alpha1/clusterextension_types.go
grokspawn: We could view it as belt-and-suspenders, but practically I don't think it will have any effect. Removing.

api/v1alpha1/clusterextension_types.go
everettraven: I wouldn't block the PR on this, but my thinking is that while we are making the change we might as well make sure we are providing as helpful of an error message to the end user as possible. While doing a web search may be many of our first impulse, I would like to think many of us would prefer error messages that clearly signal the general requirements of the field such that it saves us needing to perform that additional web search step.

This is purely a nice-to-have from the end user perspective and I would rank it high on the importance scale, but not high enough to block this since updating the message to be "nicer" would be a non-breaking change.

internal/applier/helm.go
```diff
@@ -56,6 +57,26 @@ type Helm struct {
 	Preflights         []Preflight
 }
 
+// shouldSkipPreflight is a helper to determine if the preflight check is CRDUpgradeSafety AND
+// if it is set to enforcement None.
+func shouldSkipPreflight(preflight Preflight, ctx context.Context, ext *ocv1alpha1.ClusterExtension, state string) bool {
```
everettraven, line 62: IIRC the context is usually the first parameter in "idiomatic" Go:
```suggestion
func shouldSkipPreflight(ctx context.Context, preflight Preflight, ext *ocv1alpha1.ClusterExtension, state string) bool {
```

LalatenduMohanty (approved):
/lgtm

everettraven (approved):

grokspawn:
Solves #1426

PR #1425, opened 2024-11-02
🌱 Test Registry

dtfranz:
# Test Registry

Adds a bare-bones test registry and a folder tree for our test images.

This PR aims to improve our test infrastructure by removing scripts and external images for registry operations in favor of bare-bones tools built in golang using well-known libraries (`go-containerregistry` in particular).

## Structure

The PR adds two tools:`registry` and  `push`. It also aggregates all our image-related makefile targets into one: `image-registry`.

### `registry`

Just about the simplest possible image registry that can be built using the `go-containerregistry` lib. See README [here](https://github.com/operator-framework/operator-controller/blob/2355c522ce854b4054f319347dd27ac915a12744/testdata/registry/README.md).

### `push`

Utilizes `crane` within `go-containerregistry` to build and push our catalog and bundle images. It uses a directory walker which is intended to make adding future test bundles and catalogs easily. It also (IMO - totally subjective) makes our test image collection more WYSIWYG by clearly showing all the images and tags we use in the folder rather than hiding them in scripts or makefiles.

See README [here](https://github.com/operator-framework/operator-controller/blob/2355c522ce854b4054f319347dd27ac915a12744/testdata/push/README.md) for more info on the expected folder structure.

### `make image-registry`

Builds the `registry` and `push` binaries, bundles them into one image, then deploys them both. `registry` is deployed essentially as before. `push` executes as a job, building all the images bundled inside of it then pushing them to the specified registry address. 

## TODOs / Follow-Ups

~~PR size is currently huge due to the amount of prometheus manifests being copied for each bundle tag... I'm open to suggestions on how we could address this. Maybe we could remove the `prometheus-operator` in favor of a test operator with fewer manifests?~~ Addressed this, see below.

Remaining items to address:
* [x] ~~Either:~~ Done - going with option **C** for now.
 * A: ~~Find a better way to organize tags such that we need not have many copies of the same manifests.~~
 * B: ~~Keep the current folder structure for tags but simplify the manifests we're using in test to reduce file count/complexity.~~
 * C: Modify the catalog yaml so that every version of `prometheus-operator` uses the same bundle image (`v1.0.0`). We were previously doing this anyway essentially because each tag we created for this bundle had the exact same content.
* [x] ~~Metadata reading for bundle images is using one path instead of reading dynamically (which we can get away with right now because all the bundles are identical).~~ Done.
* [x] ~~Configurable cert/key path for `registry` (not super critical right now).~~ Done.
* [x] ~~Use mounted secrets for `push` instead of `crane.Insecure` (also not critical).~~ Done.

# Description

## Reviewer Checklist

- [ ] API Go Documentation


testdata/test-catalog.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: olm.operatorframework.io/v1alpha1
+kind: ClusterCatalog
+metadata:
+  finalizers:
+  - olm.operatorframework.io/delete-server-cache
```
tmshort, line 5: Not sure this is needed...
Shouldn't the finalizers be added by the controller?

testdata/test-catalog.yaml
```diff
@@ -0,0 +1,16 @@
...
+  labels:
+    olm.operatorframework.io/metadata.name: test-catalog
```
tmshort, line 7: This label is automatically added.

Makefile
```diff
@@ -165,12 +165,13 @@ test-unit: $(SETUP_ENVTEST) #HELP Run the unit tests
...
-image-registry: ## Setup in-cluster image registry
-	./hack/test/image-registry.sh $(E2E_REGISTRY_NAMESPACE) $(E2E_REGISTRY_NAME)
-
-build-push-e2e-catalog: ## Build the testdata catalog used for e2e tests and push it to the image registry
-	./hack/test/build-push-e2e-catalog.sh $(E2E_REGISTRY_NAMESPACE) $(LOCAL_REGISTRY_HOST)/$(E2E_TEST_CATALOG_V1)
-	./hack/test/build-push-e2e-catalog.sh $(E2E_REGISTRY_NAMESPACE) $(LOCAL_REGISTRY_HOST)/$(E2E_TEST_CATALOG_V2)
+E2E_REGISTRY_IMAGE=localhost/e2e-test-registry:devel
+image-registry: ## Build the testdata catalog used for e2e tests and push it to the image registry
+	go build -o ./testdata/registry/bin/registry ./testdata/registry/registry.go
+	go build -o ./testdata/push/bin/push         ./testdata/push/push.go
```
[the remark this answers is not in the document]
tmshort: I would simplify the Dockerfile and the gitignore if they were in the same directory.

tmshort:
verdiff is complaining about new `go.mod` files; we should be ok with that.

tmshort:
I have a wicked simple bundle that can be installed: https://github.com/tmshort/test-image
It consists of just a configmap

dtfranz:
> I have a wicked simple bundle that can be installed: https://github.com/tmshort/test-image It consists of just a configmap

If you're suggesting this to replace the `prometheus-operator`, I'm happy with that as long as we're OK with not having things like: crd, deployment with runtime, etc. in our test bundle. But I think that having at least a running pod as a result of the ClusterExtension is important for the e2e to check for installation success.

I did also just test that we can definitely get away with just having the test-catalog's bundles all point to the same `prometheus-operator` version.

testdata/test-catalog.yaml
dtfranz: I think this file might actually be vestigial, I'm doing another test run but it doesn't appear to be used anywhere. I'll remove it if that's the case (but also thanks, that's good to know!).

tmshort:
It's a bundle, it can be installed... Not sure how critical it is to actually have something with a CRD or deployment. That being said, it's just an offer. Not making it a requirement.

testdata/push/push.go
```diff
@@ -0,0 +1,179 @@
...
+	if err != nil {
+		log.Fatalf("failed to build bundles: %s", err.Error())
+	}
+	catalogs, err := buildCatalogs(catalogsFullPath)
+	if err != nil {
+		log.Fatalf("failed to build catalogs: %s", err.Error())
+	}
+
+	// Push the images
+	// TODO without insecure option
```
tmshort, line 47: Ugh? Make an issue?

dtfranz: I've fixed this - `push` job is using the mounted certs now.

testdata/build-test-registry.sh
```diff
@@ -7,23 +7,23 @@ set -o pipefail
 help="
 image-registry.sh is a script to stand up an image registry within a cluster.
```
bentito, line 8: Has the script name changed and this line needs to be updated?

Makefile
```diff
@@ -165,12 +165,13 @@ test-unit: $(SETUP_ENVTEST) #HELP Run the unit tests
...
 
-image-registry: ## Setup in-cluster image registry
-	./hack/test/image-registry.sh $(E2E_REGISTRY_NAMESPACE) $(E2E_REGISTRY_NAME)
-
-build-push-e2e-catalog: ## Build the testdata catalog used for e2e tests and push it to the image registry
-	./hack/test/build-push-e2e-catalog.sh $(E2E_REGISTRY_NAMESPACE) $(LOCAL_REGISTRY_HOST)/$(E2E_TEST_CATALOG_V1)
-	./hack/test/build-push-e2e-catalog.sh $(E2E_REGISTRY_NAMESPACE) $(LOCAL_REGISTRY_HOST)/$(E2E_TEST_CATALOG_V2)
+E2E_REGISTRY_IMAGE=localhost/e2e-test-registry:devel
+image-registry: ## Build the testdata catalog used for e2e tests and push it to the image registry
+	go build $(GO_BUILD_FLAGS) -tags '$(GO_BUILD_TAGS)' -ldflags '$(GO_BUILD_LDFLAGS)' -gcflags '$(GO_BUILD_GCFLAGS)' -asmflags '$(GO_BUILD_ASMFLAGS)' -o ./testdata/registry/bin/registry ./testdata/registry/registry.go
```
bentito, line 170:
```suggestion
	GOOS=linux GOARCH=amd64 go build $(GO_BUILD_FLAGS) -tags '$(GO_BUILD_TAGS)' -ldflags '$(GO_BUILD_LDFLAGS)' -gcflags '$(GO_BUILD_GCFLAGS)' -asmflags '$(GO_BUILD_ASMFLAGS)' -o ./testdata/registry/bin/registry ./testdata/registry/registry.go
```

Makefile
```diff
@@ -165,12 +165,13 @@ test-unit: $(SETUP_ENVTEST) #HELP Run the unit tests
...
+	go build $(GO_BUILD_FLAGS) -tags '$(GO_BUILD_TAGS)' -ldflags '$(GO_BUILD_LDFLAGS)' -gcflags '$(GO_BUILD_GCFLAGS)' -asmflags '$(GO_BUILD_ASMFLAGS)' -o ./testdata/push/bin/push         ./testdata/push/push.go
```
bentito, line 171:
```suggestion
	GOOS=linux GOARCH=amd64 go build $(GO_BUILD_FLAGS) -tags '$(GO_BUILD_TAGS)' -ldflags '$(GO_BUILD_LDFLAGS)' -gcflags '$(GO_BUILD_GCFLAGS)' -asmflags '$(GO_BUILD_ASMFLAGS)' -o ./testdata/push/bin/push         ./testdata/push/push.go
```

Makefile
```diff
@@ -181,7 +182,7 @@ build-push-e2e-catalog: ## Build the testdata catalog used for e2e tests and pus
 test-e2e: KIND_CLUSTER_NAME := operator-controller-e2e
 test-e2e: KUSTOMIZE_BUILD_DIR := config/overlays/e2e
 test-e2e: GO_BUILD_FLAGS := -cover
```
bentito, line 184:
```suggestion
test-e2e: GOOS=linux
test-e2e: GOARCH=amd64
test-e2e: GO_BUILD_FLAGS := -cover
```

bentito: all the GOOS, GOARCH suggestions are needed in the case where the local platform is not Linux/AMD64 then the tests and image reg binaries will still work inside Kind cluster (which needs Linux/AMD64)

tmshort: We should be careful here; @joelanford had some changes in the Makefile to do the proper thing, we should emulate that, where possible.

Makefile
tmshort: Since the `test-e2e` target has a dependency on the `image-registry` target, the `GOOS` and `GOARCH` variables should be set from the `test-e2e` target, and don't need to be set explicitly here. This way, you can build `image-registry` locally, and get the correct (local) images when no running the e2e.

Makefile
tmshort: Since the `test-e2e` target has a dependency on the `image-registry` target, the `GOOS` and `GOARCH` variables should be set from the `test-e2e` target, and don't need to be set explicitly here. This way, you can build `image-registry` locally, and get the correct (local) images when no running the e2e.

Makefile
tmshort: This _should_ be the right place to set these variables; _not_ in the `image-registry` recipe.

dtfranz: Updated

testdata/build-test-registry.sh
dtfranz: Fixed, thanks!

Makefile
```diff
@@ -180,8 +181,10 @@ build-push-e2e-catalog: ## Build the testdata catalog used for e2e tests and pus
 .PHONY: test-e2e
 test-e2e: KIND_CLUSTER_NAME := operator-controller-e2e
 test-e2e: KUSTOMIZE_BUILD_DIR := config/overlays/e2e
+test-e2e: GOOS=linux
+test-e2e: GOARCH=amd64
```
tmshort, line 185: Use `export` here.

dtfranz: Would that not force these values across the entire makefile? Is that what we want?

dtfranz: Oh nevermind, I think I see what you mean. Like this?
```suggestion
test-e2e: export GOOS=linux
test-e2e: export GOARCH=amd64
```

dtfranz: Updated.

Makefile
camilamacedo86: Can we use `GOOS=$(go env GOOS) GOARCH=$(go env GOARCH)` instead? This would make the target compatible with both macOS and Linux environments. Setting GOOS=linux GOARCH=amd64 explicitly restricts it to those environments, meaning we can’t test or run it locally on other systems like macOS.

camilamacedo86: Hi @tmshort and @bentito,

Would it make more sense to use GOOS=$(go env GOOS) GOARCH=$(go env GOARCH) instead?

This approach would make the target compatible with both macOS and Linux environments, allowing us to run it locally as well. If we explicitly set GOARCH=amd64, for example, it won’t run on macOS systems with a different architecture, where the expected result would be darwin/arm64. Is targeting linux/amd64 specifically what we're aiming for here?

bentito: yep. current changes take this into account.

camilamacedo86: Hi @tmshort,

> Not sure that's the right approach. Doing the export should work when doing the image-target, and setting the variable equal to the go env ought to be a noop.

I’m not sure if I fully follow your comment here. The export has fixed values as well:

```
image-registry: export GOOS=linux
image-registry: export GOARCH=amd64
```

Either way, this suggests **fixed** values for go build.

So, how would `GOOS=linux GOARCH=amd64 go build` build a binary for darwin/arm64 (my local env, for example)?

tmshort: Right, but doing this:
```
GOOS=$(go env GOOS)
```
Is basically setting the `GOOS` variable to the environment value that the `go` command already has available to it. It's equivalent to doing:
```
GOOS=${GOOS}
```
Which is a no-op.

tmshort: @camilamacedo86  I didn't answer your question. 

> So, how would `GOOS=linux GOARCH=amd64 go build` build a binary for darwin/arm64 (my local env, for example)?

It wouldn't. In _this_ particular case, it's building executables for the docker container, which have to be linux/amd64. These two executables are never expected to run locally.

My point is that doing `GOOS=$(go env GODS` is pointless.

dtfranz: This build is specifically for the `image-registry` image to be run in `kind`; comparing to how we build the `operator-controller`'s `manager` binary, we use the same `GOOS=linux GOARCH=amd64` ENV when building (`make run` uses `docker-build` which then uses `build-linux`). I'm just following that same convention.

If you feel it's important, what we could potentially do is split up the `image-registry` target so that `GOOS=linux GOARCH=amd64` is only set when actually building the image. But I think this is unnecessary or at least non-critical enough to address in a follow-up.

camilamacedo86: > These two executables are never expected to run locally.

If that is the case, (we never run it locally and would not make sense run it locally) then I am OK with fixed values 👍

tmshort (approved):
/lgtm

tmshort (approved):
/lgtm
After the squash, it's still the same.

PR #1430, opened 2024-11-06
📖 docs: add local build and deploy instructions to CONTRIBUTING.md

camilamacedo86:
- Added steps for building, loading Docker images, and deploying changes locally with Kind

PR #1431, opened 2024-11-06
🌱 (chore): configure Codecov to include only relevant directories for coverage

camilamacedo86:
- Updated .codecov.yml to include only specific directories (api/, cmd/, internal/) in coverage reports
- Excluded documentation, YAML configurations, and test files to improve coverage accuracy and prevent irrelevant CI coverage failures

Example: 

The PR https://github.com/operator-framework/operator-controller/pull/1430 just add docs and should not fail with but it is:

Screenshot 2024-11-06 at 20 25 57

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,3 +1,11 @@
 codecov:
   notify:
     after_n_builds: 2
+
+# Configure the paths to include in coverage reports.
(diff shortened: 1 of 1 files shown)


tmshort (approved):
/lgtm

PR #1432, opened 2024-11-06
🐛 fix: put annotations in deployment's pod template

joelanford:
# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/rukpak/convert/registryv1.go
+++ b/internal/rukpak/convert/registryv1.go
@@ -195,6 +195,7 @@ func Convert(in RegistryV1, installNamespace string, targetNamespaces []string) 
 	for _, depSpec := range in.CSV.Spec.InstallStrategy.StrategySpec.DeploymentSpecs {
 		annotations := util.MergeMaps(in.CSV.Annotations, depSpec.Spec.Template.Annotations)
 		annotations["olm.targetNamespaces"] = strings.Join(targetNamespaces, ",")
+		depSpec.Spec.Template.Annotations = annotations
 		deployments = append(deployments, appsv1.Deployment{
 			TypeMeta: metav1.TypeMeta{
 				Kind:       "Deployment",
@@ -202,10 +203,9 @@ func Convert(in RegistryV1, installNamespace string, targetNamespaces []string) 
 			},
 
 			ObjectMeta: metav1.ObjectMeta{
-				Namespace:   installNamespace,
-				Name:        depSpec.Name,
-				Labels:      depSpec.Label,
-				Annotations: annotations,
+				Namespace: installNamespace,
+				Name:      depSpec.Name,
+				Labels:    depSpec.Label,
 			},
 			Spec: depSpec.Spec,
 		})
--- a/internal/rukpak/convert/registryv1_test.go
+++ b/internal/rukpak/convert/registryv1_test.go
@@ -2,11 +2,13 @@ package convert
 
 import (
 	"fmt"
+	"strings"
 	"testing"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	schedulingv1 "k8s.io/api/scheduling/v1"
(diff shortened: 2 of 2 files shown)


camilamacedo86 (approved):
It seems like a small fix, and everything makes sense to me after reviewing the code. 
LGTM

internal/rukpak/convert/registryv1_test.go
```diff
@@ -224,41 +246,51 @@ var _ = Describe("RegistryV1 Suite", func() {
 
 				By("verifying if plain bundle has required objects")
 				Expect(plainBundle).ShouldNot(BeNil())
-				Expect(plainBundle.Objects).To(HaveLen(6))
+				Expect(plainBundle.Objects).To(HaveLen(5))
+
+				By("verifying olm.targetNamespaces annotation in the deployment's pod template")
+				dep := findObjectByName("testDeployment", plainBundle.Objects)
+				Expect(dep).NotTo(BeNil())
+				Expect(dep.(*appsv1.Deployment).Spec.Template.Annotations).To(HaveKeyWithValue("olm.targetNamespaces", strings.Join(watchNamespaces, ",")))
```
camilamacedo86, line 254: We are checking here the change 👍

internal/rukpak/convert/registryv1_test.go
```diff
@@ -167,22 +169,38 @@ var _ = Describe("RegistryV1 Suite", func() {
...
-				csv = v1alpha1.ClusterServiceVersion{
+				baseCSV = v1alpha1.ClusterServiceVersion{
 					ObjectMeta: metav1.ObjectMeta{
 						Name: "testCSV",
 						Annotations: map[string]string{
 							"olm.properties": fmt.Sprintf("[{\"type\": %s, \"value\": \"%s\"}]", property.TypeConstraint, "value"),
 						},
 					},
 					Spec: v1alpha1.ClusterServiceVersionSpec{
-						InstallModes: []v1alpha1.InstallMode{{Type: v1alpha1.InstallModeTypeMultiNamespace, Supported: true}},
```
camilamacedo86, line 183: I understand that it was removed because, after the conversion, we expect InstallStrategy, correct?

joelanford: This is the piece of the CSV that I want to configure in each of the test scenarios, so I removed it here and then each test scenario does a deep copy before setting InstallModes for that scenario.

internal/rukpak/convert/registryv1.go
```diff
@@ -195,17 +195,17 @@ func Convert(in RegistryV1, installNamespace string, targetNamespaces []string)
...
 			TypeMeta: metav1.TypeMeta{
 				Kind:       "Deployment",
 				APIVersion: appsv1.SchemeGroupVersion.String(),
 			},
 
 			ObjectMeta: metav1.ObjectMeta{
-				Namespace:   installNamespace,
-				Name:        depSpec.Name,
-				Labels:      depSpec.Label,
-				Annotations: annotations,
```
camilamacedo86, line 208: @joelanford, I have a quick question: should we keep the annotations in both places, given that we’re merging other annotations with annotations := util.MergeMaps(in.CSV.Annotations, depSpec.Spec.Template.Annotations)?

Could we not have more annotations ?

joelanford: I'm just trying to match OLMv0's behavior (which I may _still_ not have exactly right, btw. It's complicated). We are also dealing with the etcd size limit, so I don't want to duplicate stuff more than necessary.  There is a `Subscription.spec.config.annotations` in OLMv0 that I think propagates to the deployment's annotations. So once we have the parameterized value stuff design/implemented/supported, users will be able to provide custom annotations.

camilamacedo86: > dealing with the etcd size limit
Bit by bit, memories are resurfacing... I remember this

Make sense 👍

internal/rukpak/convert/registryv1_test.go
```diff
@@ -167,22 +169,38 @@ var _ = Describe("RegistryV1 Suite", func() {
...
 			var (
 				svc             corev1.Service
-				csv             v1alpha1.ClusterServiceVersion
+				baseCSV         v1alpha1.ClusterServiceVersion
 				watchNamespaces []string
 			)
 
 			BeforeEach(func() {
-				csv = v1alpha1.ClusterServiceVersion{
+				baseCSV = v1alpha1.ClusterServiceVersion{
```
perdasilva, line 177: would it be worth adding a comment here to say that this will be deep copied and modified by each test case?
```suggestion
                                // base CSV definition that each test case will deep copy and modify 
				baseCSV = v1alpha1.ClusterServiceVersion{
```

perdasilva (approved):
lgtm - just left a small suggestion. If you don't think it's necessary, I think it's ok to merge.

tmshort (approved):
/lgtm

perdasilva (approved):
lgtm =D

PR #1434, opened 2024-11-07
⚠️ Bump catalogd to v0.36.0

m1kola:
# Description

There were some breaking changes in catalogd (https://github.com/operator-framework/catalogd/pull/443) so this requires a manual bump.

## Reviewer Checklist

- [ ] API Go Documentation


test/e2e/e2e_suite_test.go
```diff
@@ -53,8 +53,8 @@ func createTestCatalog(ctx context.Context, name string, imageRef string) (*cata
 			Source: catalogd.CatalogSource{
 				Type: catalogd.SourceTypeImage,
 				Image: &catalogd.ImageSource{
-					Ref:          imageRef,
-					PollInterval: &metav1.Duration{Duration: time.Second},
+					Ref:                 imageRef,
+					PollIntervalMinutes: ptr.To(1),
```
m1kola, line 57: Will probably have to bump timeouts in e2es because we went from 1 second to 1 minute.

camilamacedo86: So should not be here `&metav1.Duration{Duration: time.Minute},`

m1kola: @camilamacedo86 no. It is no longer a `*metav1.Duration` field. It was changed in the linked PR [here](https://github.com/operator-framework/catalogd/pull/443/files#diff-d446855f2187e6d61c4719c3202fcb21e4aea9b09e09f9e1d01419f341d8fc9fR352).

The value now accepts only minutes with the minimum of 1.

m1kola: But CI seems to pass fine with the existing timeouts.

perdasilva (approved):
lgtm

camilamacedo86 (commented):
/lgtm

LalatenduMohanty (approved):

test/e2e/e2e_suite_test.go
everettraven: @m1kola Is there a way we can validate that this doesn't introduce flakiness into CI by timeouts not being sufficient? What are the existing timeouts?

m1kola: @everettraven I was mostly worried about pulling like `require.Eventually`. We have some defaults here:

https://github.com/operator-framework/operator-controller/blob/26904a29b8622d4665d10b160be0a038fc04f540/test/e2e/cluster_extension_install_test.go#L38-L39

But some tests can use own values in theory. Our pull interval for `require.Eventually` is usually quite low, so it should be good.

I think the more historic data from CI runs we get - the easier it will be asses if we need to increase something.

PR #1449, opened 2024-11-13
:seedling: Add a make target and check in CI to verify CRD compatibility

everettraven:
# Description

- Adds https://github.com/everettraven/crd-diff as a bingo dependency
- Adds a make target called `verify-crd-compatibility` to run `crd-diff` and check for CRD incompatibilities
- Adds a GitHub Actions Workflow to run `make verify-crd-compatibility` on Pull Requests

## Reviewer Checklist

- [ ] API Go Documentation


.bingo/crd-diff.mod
```diff
@@ -0,0 +1,5 @@
+module _ // Auto generated by https://github.com/bwplotka/bingo. DO NOT EDIT
+
+go 1.22.5
```
camilamacedo86, line 3: It seems to be failing in the api-diff to build

camilamacedo86: See: https://github.com/operator-framework/operator-controller/actions/runs/11808262605/job/32896538347?pr=1449

everettraven: Yeah, this looks like a failure that got masked by our breaking API changes. I'm planning to investigate in a bit

.bingo/Variables.mk
```diff
@@ -1,4 +1,4 @@
-# Auto generated binary variables helper managed by https://github.com/bwplotka/bingo v0.9. DO NOT EDIT.
+# Auto generated binary variables helper managed by https://github.com/bwplotka/bingo v0.8. DO NOT EDIT.
```
bentito, line 1: should we just remove the version from this comment so we don't have to keep it aligned to be correct?

everettraven: Ah, I didn't realize I wasn't up to date on my bingo version. I don't think we should touch the auto-generated files though. If we are concerned about this particular change I can update my bingo version to v0.9

everettraven:
Closing and re-opening to pull in upgrade-e2e CI fixes in #1451

.bingo/Variables.mk
```diff
@@ -29,6 +29,12 @@ $(CONTROLLER_GEN): $(BINGO_DIR)/controller-gen.mod
 	@echo "(re)installing $(GOBIN)/controller-gen-v0.16.1"
 	@cd $(BINGO_DIR) && GOWORK=off $(GO) build -mod=mod -modfile=controller-gen.mod -o=$(GOBIN)/controller-gen-v0.16.1 "sigs.k8s.io/controller-tools/cmd/controller-gen"
 
+CRD_DIFF := $(GOBIN)/crd-diff-v0.0.0-20241112183958-25304aa59cdb
+$(CRD_DIFF): $(BINGO_DIR)/crd-diff.mod
+	@# Install binary/ries using Go 1.14+ build command. This is using bwplotka/bingo-controlled, separate go module with pinned dependencies.
+	@echo "(re)installing $(GOBIN)/crd-diff-v0.0.0-20241112183958-25304aa59cdb"
+	@cd $(BINGO_DIR) && GOWORK=off $(GO) build -mod=mod -modfile=crd-diff.mod -o=$(GOBIN)/crd-diff-v0.0.0-20241112183958-25304aa59cdb "github.com/everettraven/crd-diff"
```
bentito, line 36: Should this be in an `operator-framework` org repo? Or maybe just all the code under `hack/` and get it built from there?

everettraven: We use tools like this from non operator-framework repos already.

I'm not opposed to donating it to the operator-framework, but we could update this whenever that happens. Using as is gets us these checks in sooner rather than later.

everettraven:
the go-verdiff check looks like it is expected to fail in this scenario because this PR adds a go.mod file that didn't previously exist.

crd-diff-config.yaml
```diff
@@ -0,0 +1,109 @@
...
+    scope:
+      enabled: true
+    existingFieldRemoval:
+      enabled: true
+    storedVersionRemoval:
+      enabled: true
+  version:
+    sameVersion:
+      enabled: true
+      unhandledFailureMode: "Closed"
```
bentito, line 12: I think the crd-diff repo needs a docs dir. In that it needs at a minimum a doc outlining how to use the config options. I gleaned from the code for instance that we could doc this field like:

```unhandledFailureMode: Specifies handling for unexpected validation failures. "Closed" indicates strict enforcement, where failures are considered critical.```

I'd be happy to contribute a doc like this over on the tool side of this if it's desired?

Comment (everettraven) [reply crd-diff-config.yaml:12]: You're more than welcome to create a PR! I haven't gotten around to writing proper docs yet

File: .bingo/crd-diff.mod (hunk 1 above)
Comment (tmshort) [reply .bingo/crd-diff.mod:3]: https://github.com/operator-framework/operator-controller/pull/1468

Comment (tmshort) [reply .bingo/crd-diff.mod:3]: This is in now, if you rebase, it should pick up the fix for go-verdiff

File: crd-diff-config.yaml (hunk 4 above)
Comment (everettraven) [reply crd-diff-config.yaml:12]: Added much more detailed documentation for the tool and configuration options. Docs site at https://everettraven.github.io/crd-diff/

File: .bingo/Variables.mk (hunk 3 above)
Comment (everettraven) [reply .bingo/Variables.mk]: Chatted a bit with folks at KubeCon NA 2024 and there seems to be some interest in bringing something like this up to the sig-apimachinery folks to consolidate some of the various CRD validation efforts that have been going on into a single concerted effort. I'm planning to get involved in that area in the hopes we can get some official, Kubernetes sanctioned, tooling in place. 

That said, I still think it is worth it to implement the check as-is since it provides some immediate value.

File: .bingo/crd-diff.mod (hunk 1 above)
Comment (everettraven) [reply .bingo/crd-diff.mod:3]: Looks to have passed now. Thanks!

File: crd-diff-config.yaml [hunk 5]
```diff
@@ -0,0 +1,109 @@
...
+      enum:
+        enabled: true
+        removalEnforcement: "Strict"
+        additionEnforcement: "Strict"
```
everettraven, line 16: This enforcement means addition of new allowed enum values is considered incompatible. I know we have some areas of our API where we anticipate adding some enum values so we may want to move this to the less strict `IfPreviouslyConstrained` that allows addition of new allowed enum values only if the property was previously constrained by enum values. (i.e adding net new enum restrictions to a property is still considered incompatible)

LalatenduMohanty (approved):
/lgtm

PR #1429, opened 2024-11-06
🐛 fix: resolve JSONPath error in caBundle readiness check by adding kubectl_wait_for_query function

camilamacedo86:
BY running `make kind-deploy` against a kind cluster  or installing the released script from https://operator-framework.github.io/operator-controller/getting-started/olmv1_getting_started/ the following error is faced:

```sh
...
deployment.apps/cert-manager-webhook condition met
deployment.apps/cert-manager-cainjector condition met
deployment.apps/cert-manager condition met
error: jsonpath wait format must be --for=jsonpath='{.status.readyReplicas}'=3
```

This PR fixes an issue with kubectl wait when used to check the caBundle field in `mutatingwebhookconfigurations` and `validatingwebhookconfigurations`. 

### Solution

This PR introduces the `kubectl_wait_for_query` function, which replaces `kubectl wait` for this specific use case. The function repeatedly checks the `caBundle` field by using `kubectl get` in a loop, ensuring that the `caBundle` is populated without relying on status-based conditions. This approach provides a more flexible solution compatible with webhook configurations, bypassing the limitations of `kubectl wait`.

### After the fix:

```sh
$ make kind-deploy
/Users/camiladeomacedo/go/bin/controller-gen-v0.16.1 rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/base/crd/bases output:rbac:artifacts:config=config/base/rbac
/Users/camiladeomacedo/go/bin/kustomize-v4.5.7 build config/overlays/cert-manager > operator-controller.yaml
envsubst '$CATALOGD_VERSION,$CERT_MGR_VERSION,$INSTALL_DEFAULT_CATALOGS,$MANIFEST' < scripts/install.tpl.sh | bash -s
namespace/cert-manager created
customresourcedefinition.apiextensions.k8s.io/certificaterequests.cert-manager.io created
customresourcedefinition.apiextensions.k8s.io/certificates.cert-manager.io created
customresourcedefinition.apiextensions.k8s.io/challenges.acme.cert-manager.io created
customresourcedefinition.apiextensions.k8s.io/clusterissuers.cert-manager.io created
customresourcedefinition.apiextensions.k8s.io/issuers.cert-manager.io created
customresourcedefinition.apiextensions.k8s.io/orders.acme.cert-manager.io created
serviceaccount/cert-manager-cainjector created
serviceaccount/cert-manager created
serviceaccount/cert-manager-webhook created
[...]

PR #1450, opened 2024-11-13
🌱 (ci) add GitHub Action to validate installation

camilamacedo86:
- Adds a GitHub Action to validate the installation process in CI, detecting issues in `scripts/install.tpl.sh`.
- This workflow follows steps from the Getting Started guide, reducing installation errors.

**Motivation**:  
The error seen in [PR #1429](https://github.com/operator-framework/operator-controller/pull/1429) would have been detected with this action in place.

### DIFF
--- a/.github/workflows/install.yaml
+++ b/.github/workflows/install.yaml
@@ -1,0 +1,51 @@
+name: Install
+
+on:
+  workflow_dispatch:
+  pull_request:
+  merge_group:
+  push:
+    branches:
+      - main
+
+jobs:
+  kind-deploy:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v5
+        with:
+          go-version-file: go.mod
+
+      - name: Install the latest version of kind
+        run: |
+          curl -Lo ./kind https://kind.sigs.k8s.io/dl/latest/kind-linux-amd64
+          chmod +x ./kind
+          sudo mv ./kind /usr/local/bin/kind
+
+      - name: Create Kind cluster for operator-controller
+        run: |
+          kind create cluster --name operator-controller
+
+      - name: Build the project
+        run: |
+          make docker-build
+
+      - name: Load image into Kind cluster and deploy
+        run: |
+          make kind-load kind-deploy
+
+      - name: Logs and Details
+        if: always()
+        run: |
+          # Capture high-level information
+          echo "Gathering details for all resources in namespace olmv1-system..."
+          kubectl get all -n olmv1-system
+          
+          # Describe each pod in the namespace for more details
+          for pod in $(kubectl get pods -n olmv1-system -o jsonpath='{.items[*].metadata.name}'); do
+            echo "Describing pod $pod..."
+            kubectl describe pod $pod -n olmv1-system
+            echo "Logs for pod $pod:"
+            kubectl logs $pod -n olmv1-system --all-containers=true
+          done


joelanford:
I don't think this will work on MacOS.

On MacOS, when we build a container, we need to specifically cross-compile to GOOS=linux. That's why the `go-build-linux` targets exist.

joelanford:
I think we _can_probably remove the hardcoded GOARCH setting though.

camilamacedo86:
Hi @joelanford 

Thank you for your time and look for this one. 

> I don't think this will work on MacOS.

I use macOS, which is my motivation for these changes. 😊 It works well for me, as you can see below.

Additionally, to ensure compatibility across macOS and Linux, I updated this PR to run all installation steps on both operating systems in the GitHub Actions workflow. This setup makes reviewing easier and gives us confidence that we won’t unintentionally break compatibility in the future.


```
$ make docker-build
/Users/camiladeomacedo/go/bin/controller-gen-v0.16.1 rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/base/crd/bases output:rbac:artifacts:config=config/base/rbac
/Users/camiladeomacedo/go/bin/controller-gen-v0.16.1 object:headerFile="hack/boilerplate.go.txt" paths="./..."
go fmt ./...
go vet -tags 'containers_image_openpgp' ./...
go build  -tags 'containers_image_openpgp' -ldflags '-s -w -X 'github.com/operator-framework/operator-controller/internal/version.version=6ef6948' ' -gcflags 'all=-trimpath=/Users/camiladeomacedo/go/src/github.com/operator-framework/operator-controller' -asmflags 'all=-trimpath=/Users/camiladeomacedo/go/src/github.com/operator-framework/operator-controller' -o bin/manager ./cmd/manager
docker build -t quay.io/operator-framework/operator-controller:devel -f Dockerfile ./bin/
[+] Building 5.0s (7/7) FINISHED                                                
 => [internal] load build definition from Dockerfile                       0.0s
 => => transferring dockerfile: 37B                                        0.0s
 => [internal] load .dockerignore                                          0.0s
 => => transferring context: 2B                                            0.0s
 => [internal] load metadata for gcr.io/distroless/static:nonroot          0.5s
 => [internal] load build context                                          3.8s
 => => transferring context: 75.21MB                                       3.8s
 => CACHED [1/3] FROM gcr.io/distroless/static:nonroot@sha256:d71f4b239be  0.0s
 => [2/3] COPY manager manager                                             0.4s
 => exporting to image                                                     0.2s
 => => exporting layers                                                    0.2s
 => => writing image sha256:36948c6c6d8b467c570a6724f3da9e68ccb831b767f69  0.0s
 => => naming to quay.io/operator-framework/operator-controller:devel      0.0s
```

> On MacOS, when we build a container, we need to specifically cross-compile to GOOS=linux. That's why the go-build-linux targets exist.

Thanks for the context! _(as far I understand)_ Since Docker builds in a Linux-based container, it already compiles as if it’s on a Linux system, so I thought GOOS=linux might be unnecessary here—it just seemed to work without it. 😊

joelanford:
@camilamacedo86 That container won't be runnable in a cluster because it is `darwin/arm64`, which can't run in a linux-based kubernetes environment.

.github/workflows/install.yaml
```diff
@@ -0,0 +1,51 @@
+name: Install
```
everettraven, line 1: I don't think that we need a dedicated action to run the install script. All e2e runs in CI will use the install script so I would expect install script issues to be present in e2e runs.

camilamacedo86:
Hi @joelanford 

I tried to add to the action the build for macOS but unfortunately, it is not working very well: https://github.com/actions/runner/issues/1456 (I tried a couple of options but the docker does not work :-( )

I put above that for mac/docker we do not need to export GOOS=linux, see: https://github.com/operator-framework/operator-controller/pull/1450#issuecomment-2473919621 . But if we add just the GOOS it should works too.

So since you requested I added this part back and checked that it still working

```
 $ make docker-build
/Users/camiladeomacedo/go/bin/controller-gen-v0.16.1 rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/base/crd/bases output:rbac:artifacts:config=config/base/rbac
/Users/camiladeomacedo/go/bin/controller-gen-v0.16.1 object:headerFile="hack/boilerplate.go.txt" paths="./..."
go fmt ./...
go vet -tags 'containers_image_openpgp' ./...
docker build -t quay.io/operator-framework/operator-controller:devel -f Dockerfile ./bin
[+] Building 2.9s (7/7) FINISHED                                                
 => [internal] load build definition from Dockerfile                       0.2s
 => => transferring dockerfile: 294B                                       0.0s
 => [internal] load .dockerignore                                          0.1s
 => => transferring context: 2B                                            0.0s
 => [internal] load metadata for gcr.io/distroless/static:nonroot          0.5s
 => [internal] load build context                                          2.0s
 => => transferring context: 75.21MB                                       2.0s
 => [1/3] FROM gcr.io/distroless/static:nonroot@sha256:d71f4b239be2d41201  0.0s
 => CACHED [2/3] COPY manager manager                                      0.0s
 => exporting to image                                                     0.0s
 => => exporting layers                                                    0.0s
 => => writing image sha256:36948c6c6d8b467c570a6724f3da9e68ccb831b767f69  0.0s
 => => naming to quay.io/operator-framework/operator-controller:devel      0.0s
```

 
I think we are good to fly

.github/workflows/install.yaml
camilamacedo86: Unfortunately, that doesn't seem accurate. We’ve been addressing issues with the script for exactly this reason. For example, the error in [PR #1429](https://github.com/operator-framework/operator-controller/pull/1429) would have prevented the CI from passing. **Could you please check and let me know where this test might be duplicated elsewhere?**

Additionally, we are validating the build/docker process locally as well. This validation currently cannot be tested in any other environment because it’s broken and fails. Error Examples: (before changes):  make docker-build failed on both macOS and Linux due to a hardcoded ./bin/linux path that does not exist. Here’s an example of the error: [Error example](https://github.com/operator-framework/operator-controller/actions/runs/11814261205/job/32912965441) Even if we create the dir it will still failing [see error](https://github.com/operator-framework/operator-controller/actions/runs/11814618324/job/32914032322)

trgeiger: The Makefile as it is on 'main' right now works fine for me on Linux. The test you linked to above is failing because it appears it was before you changed the docker-build line to target the ./bin directory rather than ./bin/linux.

I don't have access to a mac but I don't really see how the changes in this PR are really different from how it worked before--it seems to just be changing where we build the binaries from ./bin/linux to ./bin? And it's also effectively removing the functionality of the existing 'go-build-local' target and replacing it with the current 'go-build-linux,' just renaming them and removing the former since you're still setting GOOS=linux.

It is interesting that the existing go-build-local on main isn't directly used anywhere--I'm assuming it's there for cases where somebody wanted to compile locally and would just locally edit the docker-build target to use 'build' for one-off testing. Maybe we should just make another docker-build target that uses -local, i.e. docker-build-local, if that's a workflow we see people needing/wanting

joelanford: I agree with @everettraven. We are (in theory) automatically generating a manifest and rendering an install script as part of the e2e.

If there are gaps in what happens with the installation during the e2e, let's fix it in the e2e instead of introducing a separate test.

everettraven: @camilamacedo86 Part of running our e2e tests includes doing, more or less, what you are doing here.

If you look at the `test-e2e` make target at https://github.com/operator-framework/operator-controller/blob/6d79092c8921f888b125e084b494576e87d5706e/Makefile#L179-L188 and dig even further into the `run` target at https://github.com/operator-framework/operator-controller/blob/6d79092c8921f888b125e084b494576e87d5706e/Makefile#L279-L280 you can see that it runs the make targets `docker-build`, `kind-cluster`, `kind-load`, and `kind-deploy`. The `kind-deploy` target at https://github.com/operator-framework/operator-controller/blob/6d79092c8921f888b125e084b494576e87d5706e/Makefile#L222-L226 uses the same installation script that we generate for releases.

Essentially, everything you are doing in this CI action is already done as part of all of our e2e runs in CI. 

I'm not quite sure why CI didn't catch the issues you linked, but I also never ended up running into those problems. Is it possible that there may be some difference in tooling versions that resulted in you experiencing those issues?

camilamacedo86: HI @everettraven 

>  docker-build kind-cluster kind-load kind-deploy 

I see that is calling so that is true. 
Thank you for point out. 

>  Is it possible that there may be some difference in tooling versions that resulted in you experiencing those issues?

Not that is not really possible 
if you revert https://github.com/operator-framework/operator-controller/pull/1429 OR https://github.com/operator-framework/operator-controller/pull/1411 you will see that the error occurs always.

**So probably we move forward besides the e2e failures.**

camilamacedo86:
Closing this one. 
@everettraven point out that we are doing those tests; https://github.com/operator-framework/operator-controller/pull/1450#discussion_r1840852324
if we revert https://github.com/operator-framework/operator-controller/pull/1429 OR https://github.com/operator-framework/operator-controller/pull/1411 you will see that the error occurs always.

So, we probably move forward besides the e2e failures.
that is fine

PR #1453, opened 2024-11-13
🌱 Remove CGO_ENABLED=1 from test-unit target to fix tests on Mac OS

camilamacedo86:
**Motivation:**

By using CGO_ENABLED=1 it does not work well on MacOS and I am facing warnings such as:

```
 ld: warning: ‘/private/var/folders/xf/_py4d4yn7qqdlr5pym3pxvnc0000gn/T/go-link-38050737/000012.o’ has malformed LC_DYSYMTAB, expected 98 undefined symbols to start at index 1626, found 95 undefined symbols starting at index 1626
```

By disabling it all works fine.
PS.: we do not enable CGO for our builds

# Description

## Reviewer Checklist

- [ ] API Go Documentation


joelanford:
Are you sure the race detector is actually doing what it is supposed to be doing if we set CGO_ENABLED=0?

From https://go.dev/doc/articles/race_detector

> The race detector requires cgo to be enabled, and on non-Darwin systems requires an installed C compiler. The race detector supports linux/amd64, linux/ppc64le, linux/arm64, linux/s390x, freebsd/amd64, netbsd/amd64, darwin/amd64, darwin/arm64, and windows/amd64.

The warnings are harmless (but I agree also annoying and clutter the output) https://github.com/golang/go/issues/61229

Issue #341, opened 2023-08-18
Switch to `go tool` for tool dependency management


m1kola:
There is a proposal for go to receive support for tool dependency management which might land as soon as Go 1.22: https://github.com/golang/go/issues/48429

Once it is available and our codebase is updated to the corresponding Go version we should switch to native support.

At the moment of writing we use `github.com/bwplotka/bingo` for tool pinning.

yuriolisa:
Can I sign up for this one?

m1kola:
@yuriolisa thanks for your interest in this. Unfortunately from https://github.com/golang/go/issues/48429 it looks like we will have to wait a longer. As far as I understand this feature didn't make it into Go 1.22 and will likely not be release in 1.23.

m1kola:
It seems like this will be released in 1.24 with chis about 2 months away: https://github.com/golang/go/issues/48429#issuecomment-2504513067

m1kola:
It seems like all the relevant changes got merged in Go and the issue is now closed. We should be able to use it once we upgrade to 1.24.

m1kola:
Removed good first issue label because it is more complicated than it seems :)

I've been experimenting with 1.24 tool management [in this PR](https://github.com/operator-framework/operator-controller/pull/1512).

It seems like it is possible to achieve what we currently have with bingo:

1. We can have a separate modfile for each tool so their dependencies won't conflict (as of today - our tools do have conflicts between themselves).
2. Go 1.24 will handle tool installation and caching between projects (in the same way it caches modules between projects).
3. We will be able to to continue using `make` the same way as we currently do

What will change:
1. Instead of `bingo` CLI we will have to use Go CLI. And this is where bingo is easier:
    ````
    # Can just run this in the project root
    bingo get sigs.k8s.io/controller-tools/cmd/controller-gen@new-version 
    ````
    versus go 1.24:
    ````
    # Have to specify path to to the right module
    go get -modfile tools/controller-gen.mod -tool sigs.k8s.io/controller-tools/cmd/controller-gen@new-version 
    ````
    We can probably do some makefile magic for this, but still.
    Also - if we accidentally run `go mod tidy` again one of these tools modfiles - `go mod tidy` will add a bunch of dependencies from the main module, I think.
3. Go 1.24 currently doesn't support build tags for tools. I think it is currently recommended to use `go run` for tools with build tags, but I haven't tried this out yet.

[...]

PR #1454, opened 2024-11-13
🌱 Configure ENVTEST Binaries for IDE Debugging

camilamacedo86:
This change introduces downloading the specific binaries required to run ENVTEST-based tests into the project/bin directory. This setup makes configuring tests for debugging directly in an IDE easier.

### Motivation
1. **Ease of IDE Debugging**:
   - By setting the `BinaryAssetsDirectory` dynamically, contributors can debug tests in their preferred IDEs (e.g., GoLand, VSCode) without requiring additional environment configuration such as setting `KUBEBUILDER_ASSETS`.

Screenshot 2024-11-13 at 17 38 40

2. **Project-Specific Binaries:**

Installing binaries inside the project avoids conflicts with other projects or system-wide installations, particularly useful for contributors with restricted environments (e.g., corporate laptops).

The binaries installed by envtest-setup are no longer placed in the HOME/global directory. See the default behavior here:
https://github.com/kubernetes-sigs/controller-runtime/blob/e3347b5405cdd0da5bff527af3d406117938ba6b/tools/setup-envtest/README.md?plain=1#L57-L70

This change ensures binaries are placed locally, avoiding the conflicts and clutter caused by global installations.

I’m referring to the binaries required for ENVTEST **(not those managed by BINGO)**.

Screenshot 2024-11-19 at 01 22 59

Before this PR those are placed for example in a Mac OS at: `/Users/camilam/Library/Application Support/io.kubebuilder.envtest/k8s/1.31.0-darwin-arm64`

```
$ pwd
/Users/camilam/Library/Application Support/io.kubebuilder.envtest/k8s/1.31.0-darwin-arm64

ls -la
total 337856
dr-xr-xr-x  5 camilam  staff       160 19 Nov 09:20 .
drwxr-xr-x  3 camilam  staff        96 19 Nov 09:20 ..
-r-xr-xr-x  1 camilam  staff  26058562 19 Nov 09:20 etcd
-r-xr-xr-x  1 camilam  staff  90359954 19 Nov 09:20 kube-apiserver
-r-xr-xr-x  1 camilam  staff  56560754 19 Nov 09:20 kubectl

```

3. **Simplified Contributor Onboarding**:
   - New contributors often face a learning curve to:
     - Understand ENVTEST and its dependency on binaries.
     - Configure `KUBEBUILDER_ASSETS` manually.
     - Set up their IDE for debugging tests.
   - This setup minimizes friction and ensures tests can be run consistently across different environments.

While setting `KUBEBUILDER_ASSETS` is sufficient for experienced contributors, this change makes it easier for newcomers and supports debugging directly in IDEs without additional configuration.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,10 +29,10 @@ export WAIT_TIMEOUT := 60s
 # Install default ClusterCatalogs
 export INSTALL_DEFAULT_CATALOGS := true
 
-# By default setup-envtest will write to $XDG_DATA_HOME, or $HOME/.local/share if that is not defined.
+# By default setup-envtest binary will write to $XDG_DATA_HOME, or $HOME/.local/share if that is not defined.
 # If $HOME is not set, we need to specify a binary directory to prevent an error in setup-envtest.
 # Useful for some CI/CD environments that set neither $XDG_DATA_HOME nor $HOME.
-SETUP_ENVTEST_BIN_DIR_OVERRIDE=
+SETUP_ENVTEST_BIN_DIR_OVERRIDE += --bin-dir $(ROOT_DIR)/bin/envtest-binaries
 ifeq ($(shell [[ $$HOME == "" || $$HOME == "/" ]] && [[ $$XDG_DATA_HOME == "" ]] && echo true ), true)
 	SETUP_ENVTEST_BIN_DIR_OVERRIDE += --bin-dir /tmp/envtest-binaries
 endif
@@ -158,19 +158,25 @@ test-ext-dev-e2e: $(OPERATOR_SDK) $(KUSTOMIZE) $(KIND) #HELP Run extension creat
 	test/extension-developer-e2e/setup.sh $(OPERATOR_SDK) $(CONTAINER_RUNTIME) $(KUSTOMIZE) $(KIND) $(KIND_CLUSTER_NAME) $(E2E_REGISTRY_NAMESPACE)
 	go test -count=1 -v ./test/extension-developer-e2e/...
 
-.PHONY: test-unit
 ENVTEST_VERSION := $(shell go list -m k8s.io/client-go | cut -d" " -f2 | sed 's/^v0\.\([[:digit:]]\{1,\}\)\.[[:digit:]]\{1,\}$$/1.\1.x/')
 UNIT_TEST_DIRS := $(shell go list ./... | grep -v /test/)
 COVERAGE_UNIT_DIR := $(ROOT_DIR)/coverage/unit
-test-unit: $(SETUP_ENVTEST) #HELP Run the unit tests
+
+.PHONY: envtest-k8s-bins #HELP Uses setup-envtest to download and install the binaries required to run ENVTEST-test based locally at the project/bin directory.
+envtest-k8s-bins: $(SETUP_ENVTEST)
+	mkdir -p $(ROOT_DIR)/bin
+	$(SETUP_ENVTEST) use -p env $(ENVTEST_VERSION) $(SETUP_ENVTEST_BIN_DIR_OVERRIDE)
+
+.PHONY: test-unit
+test-unit: $(SETUP_ENVTEST) envtest-k8s-bins #HELP Run the unit tests
 	rm -rf $(COVERAGE_UNIT_DIR) && mkdir -p $(COVERAGE_UNIT_DIR)
-	eval $$($(SETUP_ENVTEST) use -p env $(ENVTEST_VERSION) $(SETUP_ENVTEST_BIN_DIR_OVERRIDE)) && \
+	KUBEBUILDER_ASSETS="$(shell $(SETUP_ENVTEST) use -p path $(ENVTEST_VERSION) $(SETUP_ENVTEST_BIN_DIR_OVERRIDE))" \
             CGO_ENABLED=1 go test \
                 -tags '$(GO_BUILD_TAGS)' \
                 -cover -coverprofile ${ROOT_DIR}/coverage/unit.out \
                 -count=1 -race -short \
                 $(UNIT_TEST_DIRS) \
-                -test.gocoverdir=$(ROOT_DIR)/coverage/unit
+                -test.gocoverdir=$(COVERAGE_UNIT_DIR)
 
 .PHONY: image-registry
 E2E_REGISTRY_IMAGE=localhost/e2e-test-registry:devel
--- a/internal/controllers/suite_test.go
+++ b/internal/controllers/suite_test.go
@@ -26,7 +26,7 @@ import (
 
 	"github.com/stretchr/testify/require"
 	"k8s.io/apimachinery/pkg/api/meta"
-	"k8s.io/apimachinery/pkg/runtime"
+	apimachineryruntime "k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -64,7 +64,7 @@ func (m *MockUnpacker) Cleanup(_ context.Context, _ *source.BundleSource) error 
 func newClient(t *testing.T) client.Client {
 	// TODO: this is a live client, which behaves differently than a cache client.
 	//  We may want to use a caching client instead to get closer to real behavior.
-	sch := runtime.NewScheme()
+	sch := apimachineryruntime.NewScheme()
 	require.NoError(t, ocv1.AddToScheme(sch))
 	cl, err := client.New(config, client.Options{Scheme: sch})
 	require.NoError(t, err)
@@ -162,6 +162,21 @@ func TestMain(m *testing.M) {
 		ErrorIfCRDPathMissing: true,
 	}
 
+	// ENVTEST-based tests require specific binaries. By default, these binaries are located
+	// in paths defined by controller-runtime. However, the `BinaryAssetsDirectory` needs
+	// to be explicitly set when running tests directly (e.g., debugging tests in an IDE)
+	// without using the Makefile targets.
+	//
+	// This is equivalent to configuring your IDE to export the `KUBEBUILDER_ASSETS` environment
+	// variable before each test execution. The following function simplifies this process
+	// by handling the configuration for you.
+	//
+	// To ensure the binaries are in the expected path without manual configuration, run:
+	// `make envtest-k8s-bins`
+	if getFirstFoundEnvTestBinaryDir() != "" {
+		testEnv.BinaryAssetsDirectory = getFirstFoundEnvTestBinaryDir()
+	}
+
 	var err error
 	config, err = testEnv.Start()
 	utilruntime.Must(err)
@@ -178,4 +193,16 @@ func TestMain(m *testing.M) {
 	code := m.Run()
 	utilruntime.Must(testEnv.Stop())
 	os.Exit(code)
+}
+
+// getFirstFoundEnvTestBinaryDir finds and returns the first directory under the given path.
+func getFirstFoundEnvTestBinaryDir() string {
+	basePath := filepath.Join("..", "..", "bin", "envtest-binaries", "k8s")
+	entries, _ := os.ReadDir(basePath)
+	for _, entry := range entries {
+		if entry.IsDir() {
+			return filepath.Join(basePath, entry.Name())
+		}
+	}
+	return ""
 }
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -59,6 +59,32 @@ you can follow the steps below to test your changes:
     make kind-load kind-deploy
     ```
 
+## How to debug controller tests using ENVTEST
+
+[ENVTEST](https://book.kubebuilder.io/reference/envtest) requires k8s binaries to be downloaded to run the tests.
+To download the necessary binaries, follow the steps below:
+
+```sh
+make envtest-k8s-bins
+```
+
+Note that the binaries are downloaded to the `bin/envtest-binaries` directory.
+
+```sh
+$ tree
+.
+├── envtest-binaries
+│   └── k8s
+│       └── 1.31.0-darwin-arm64
+│           ├── etcd
+│           ├── kube-apiserver
+│           └── kubectl
+```
+
+Now, you can debug them with your IDE:
+
+![Screenshot IDE example](https://github.com/user-attachments/assets/3096d524-0686-48ca-911c-5b843093ad1f)
+
 ### Communication Channels
 
 - Email: [operator-framework-olm-dev](mailto:operator-framework-olm-dev@googlegroups.com)


Makefile
```diff
@@ -29,14 +29,6 @@ export WAIT_TIMEOUT := 60s
 # Install default ClusterCatalogs
 export INSTALL_DEFAULT_CATALOGS := true
 
-# By default setup-envtest will write to $XDG_DATA_HOME, or $HOME/.local/share if that is not defined.
-# If $HOME is not set, we need to specify a binary directory to prevent an error in setup-envtest.
-# Useful for some CI/CD environments that set neither $XDG_DATA_HOME nor $HOME.
-SETUP_ENVTEST_BIN_DIR_OVERRIDE=
-ifeq ($(shell [[ $$HOME == "" || $$HOME == "/" ]] && [[ $$XDG_DATA_HOME == "" ]] && echo true ), true)
-	SETUP_ENVTEST_BIN_DIR_OVERRIDE += --bin-dir /tmp/envtest-binaries
```
camilamacedo86, line 37: We are updating the setup to use the project's bin directory (used already for the other staff), ensuring we don’t interfere with contributors' local development environments."

joelanford: iirc, the reason this is here is because Red Hat reported issues when they ran CI because of quirks of how their CI system sets up the environment. We would be good stewards if we validated changes in this area in RH's CI system before we make them here.

joelanford: Note the conditional: in order for this override to be used we need both:
- $HOME=="" or $HOME=="/"
- XDG_DATA_HOME==""

That will not happen in a developer's local environment.

camilamacedo86: My whole idea is for us 
in local/dev is to have the bins inside of the project
So that I can config the ENVTEST properly and debug the tests with IDE
I will give a look

I do not want either the targets installing staff in my go src or any other place outside of the project.

Makefile
```diff
@@ -29,13 +29,6 @@ export WAIT_TIMEOUT := 60s
...
-endif
```
camilamacedo86, line 38: @joelanford I ONLY move it to be close to the others related configurations so that make easier for us check the values and etc

Makefile
```diff
@@ -341,4 +365,4 @@ deploy-docs: venv
 	. $(VENV)/activate; \
 	mkdocs gh-deploy --force
 
-include Makefile.venv
+include Makefile.venv
```
camilamacedo86, line 368: CAMILA : TODO: give space

internal/controllers/suite_test.go
```diff
@@ -160,6 +162,14 @@ func TestMain(m *testing.M) {
...
 		},
 		ErrorIfCRDPathMissing: true,
+
+		// The BinaryAssetsDirectory is only required if you want to run the tests directly
+		// without call the makefile target test. If not informed it will look for the
+		// default path defined in controller-runtime which is /usr/local/kubebuilder/.
+		// Note that you must have the required binaries setup under the bin directory to perform
+		// the tests directly. When we run make u it will be setup and used automatically.
+		BinaryAssetsDirectory: filepath.Join("..", "..", "bin", "envtest-binaries", "k8s",
+			fmt.Sprintf("1.31.0-%s-%s", runtime.GOOS, runtime.GOARCH)),
```
camilamacedo86, line 172: I will change it 
Please wait :-)

Makefile
```diff
@@ -151,19 +144,49 @@ test-ext-dev-e2e: $(OPERATOR_SDK) $(KUSTOMIZE) $(KIND) #HELP Run extension creat
...
-.PHONY: test-unit
+# By default, setup-envtest will write to $XDG_DATA_HOME or $HOME/.local/share if neither is set.
+# If $HOME is unset, specify a binary directory to prevent an error in setup-envtest.
+# This is useful for CI/CD environments without $XDG_DATA_HOME or $HOME.
+LOCALBIN := $(PWD)/bin
+SETUP_ENVTEST := $(LOCALBIN)/setup-envtest
+SETUP_ENVTEST_BIN_DIR_OVERRIDE := --bin-dir $(LOCALBIN)/envtest-binaries
+
+ifeq ($(shell [[ -z "$$HOME" || "$$HOME" == "/" ]] && [[ -z "$$XDG_DATA_HOME" ]] && echo true), true)
+	SETUP_ENVTEST := /tmp/setup-envtest
```
camilamacedo86, line 155: I am not sure if we are really using this part
but the binary used to setup-envtest is either downloaded.

Makefile
camilamacedo86: @joelanford Is not only used because the default path will be /usr/local/kubebuilder/, which does not work in the GitHub/ci. If we move to the project's bin dir, would the issues be reported differently?

Makefile
```diff
@@ -151,19 +151,26 @@ test-ext-dev-e2e: $(OPERATOR_SDK) $(KUSTOMIZE) $(KIND) #HELP Run extension creat
...
-.PHONY: test-unit
 ENVTEST_VERSION := $(shell go list -m k8s.io/client-go | cut -d" " -f2 | sed 's/^v0\.\([[:digit:]]\{1,\}\)\.[[:digit:]]\{1,\}$$/1.\1.x/')
 UNIT_TEST_DIRS := $(shell go list ./... | grep -v /test/)
 COVERAGE_UNIT_DIR := $(ROOT_DIR)/coverage/unit
+
+.PHONY: setup-envtest #HELP Install setup-envtest binary in the bin directory.
+setup-envtest:
+	mkdir -p $(ROOT_DIR)/bin
+	GOBIN=$(ROOT_DIR)/bin go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
+	$(SETUP_ENVTEST) use -p env $(ENVTEST_VERSION) $(SETUP_ENVTEST_BIN_DIR_OVERRIDE)
```
m1kola, line 162: Where do we call this target? As far as I see `test-unit` still runs `$(SETUP_ENVTEST)` (from bingo).

camilamacedo86:
Hi @m1kola,

Thank you for catching that—really good observation! :raised_hands:

Your comment highlights an additional motivation for this change: addressing scenarios like those outlined in [kubernetes-sigs/controller-runtime#2744](https://github.com/kubernetes-sigs/controller-runtime/issues/2744) and [kubernetes-sigs/controller-runtime#2646](https://github.com/kubernetes-sigs/controller-runtime/issues/2646).

The required changes have been addressed, and we no longer use bingo here as intended. Please feel free to review this one.

Makefile
```diff
@@ -32,7 +32,7 @@ export INSTALL_DEFAULT_CATALOGS := true
 # By default setup-envtest will write to $XDG_DATA_HOME, or $HOME/.local/share if that is not defined.
 # If $HOME is not set, we need to specify a binary directory to prevent an error in setup-envtest.
 # Useful for some CI/CD environments that set neither $XDG_DATA_HOME nor $HOME.
-SETUP_ENVTEST_BIN_DIR_OVERRIDE=
+SETUP_ENVTEST_BIN_DIR_OVERRIDE := --bin-dir $(ROOT_DIR)/bin/envtest-binaries
 ifeq ($(shell [[ $$HOME == "" || $$HOME == "/" ]] && [[ $$XDG_DATA_HOME == "" ]] && echo true ), true)
 	SETUP_ENVTEST_BIN_DIR_OVERRIDE += --bin-dir /tmp/envtest-binaries
```
perdasilva, line 37: could the += cause problems here and we end up with SETUP_ENVTEST_BIN_DIR_OVERRIDE = --bin-dir /$(ROOT_DIR)/... --bin-dir /tmp/envtest-binaries 

Is this what we are after?

internal/controllers/suite_test.go
```diff
@@ -162,6 +162,15 @@ func TestMain(m *testing.M) {
 		ErrorIfCRDPathMissing: true,
 	}
 
+	// The BinaryAssetsDirectory is only necessary when we run tests directly
+	// without using the target. (e.g., when debugging tests in an IDE)
+	// If left unset, it will default to the path defined in controller-runtime (/usr/local/kubebuilder/).
+	// Ensure that required binaries are set up in the bin directory bin/k8s/envtest-binaries for direct testing.
+	// Run `make setup-envtest` to set up the binaries.
+	if getFirstFoundEnvTestBinaryDir() != "" {
```
perdasilva, line 170: would it be cleaner if we just used an environment variable like ENV_TEST_BINARY_DIR or something like that?
Then, it doesn't really matter where the binaries are placed.

Makefile
camilamacedo86: I do not think so, but we can replace `:=` with `+=`
It will not change the result 👍

Makefile
```diff
@@ -151,19 +151,27 @@ test-ext-dev-e2e: $(OPERATOR_SDK) $(KUSTOMIZE) $(KIND) #HELP Run extension creat
...
 ENVTEST_VERSION := $(shell go list -m k8s.io/client-go | cut -d" " -f2 | sed 's/^v0\.\([[:digit:]]\{1,\}\)\.[[:digit:]]\{1,\}$$/1.\1.x/')
+CONTROLLER_RUNTIME_VERSION := release-$(shell go list -m sigs.k8s.io/controller-runtime | awk '{print $$2}' | sed -E 's/^v([0-9]+)\.([0-9]+)\..*/\1.\2/')
 UNIT_TEST_DIRS := $(shell go list ./... | grep -v /test/)
 COVERAGE_UNIT_DIR := $(ROOT_DIR)/coverage/unit
-test-unit: $(SETUP_ENVTEST) #HELP Run the unit tests
+
+.PHONY: setup-envtest #HELP Install setup-envtest binary in the bin directory.
+setup-envtest:
+	mkdir -p $(ROOT_DIR)/bin
+	GOBIN=$(ROOT_DIR)/bin go install sigs.k8s.io/controller-runtime/tools/setup-envtest@$(CONTROLLER_RUNTIME_VERSION)
```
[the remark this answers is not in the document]
camilamacedo86: I wouldn't say I liked this bingo thing because I work with other projects, and it replaces the modules and setup of things that are set up in my local environment and have no relation to the project.

The binary in the dir bin ensures that the project can use it and does not affect my local environment. Also, it allows me to download other envtest binaries if I need to.

internal/controllers/suite_test.go
camilamacedo86: By default, the binaries required for the project should be placed in the `bin` directory of the project.  

If we don’t enforce a specific location, then anyone wanting to debug tests **must**:  
- Understand that the tests require certain binaries.  
- Identify the correct environment variables needed.  
- Manually configure these variables every single time they want to debug something.  

By standardizing this, everything works seamlessly without requiring contributors to deeply understand or reconfigure the setup.  
I don’t need to set up test execution repeatedly—things “just work” as they should, hassle-free.

Makefile
camilamacedo86: Done

Makefile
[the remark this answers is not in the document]
camilamacedo86: In other words, setup-envtest is a binary and not a GO module to be set in go/bin.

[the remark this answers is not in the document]
bentito: IMO this is better off not cluttering the Makefile and left to bingo to handle. Bingo does not change local binaries except for the project.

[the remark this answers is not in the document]
camilamacedo86: Bingo will setup things under $GOPATH/go/bin
That is valid for ANY Golang project in my local env
So, yes, it affects all my other projects.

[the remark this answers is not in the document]
joelanford: Bingo includes the version number on the binaries, so projects can share binaries because every project is still able to specify a specific version to use.

The maintainers aligned on bingo over a year ago, and I think it is more important to be consistent on tool usage. If we want to move away from bingo, we should do it consistently, not as a one-off.

joelanford:
I like the integration in TestMain to make envtest work for IDE-based test execution. Could we put that piece in its own PR?

Makefile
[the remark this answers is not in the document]
camilamacedo86: Hi @joelanford,

This setup is really problematic as it doesn’t respect my local environment.

**A good example of the issue is:**

I encountered a significant problem when trying to run kind load image directly (outside of the project target). The command failed because of a version mismatch between the kind binary installed by Bingo and the version I expected to use.

To make matters worse, if I want to run any kind command, I now have to remember that Bingo overrides my setup. This means I need to use GOPATH/bin/kind_<version> instead of simply running the command from my own installed version.

This behavior is frustrating and doesn’t align with respecting local setups, which is critical for a smooth development experience.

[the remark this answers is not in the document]
camilamacedo86: Anyway, as it is as it is
I will change it to use the BINGO
and came back to the BINGO staff.

m1kola (changes requested):
Putting a hold for now. Let's discuss this first.

I see that this PR makes envtest be unmanaged by bingo. I think this needs a discussion because we intentionally decided to manage dependencies such as envtest and kind with bingo for reproducibility reasons.

There was another PR which was trying to do something similar in catalogd with kind [here](https://github.com/operator-framework/catalogd/pull/462#discussion_r1846231104).

This suggests that there are some issues with this approach.

Given that would be good to gather some more information about the issues we are facing so we can reevaluate our decision on bingo and decide the way forward.

@camilamacedo86 could you please create an issue or an RFC document describing the issues you are facing across our repos with the dependencies? I think it will be very helpful to have this information in one place so we can discuss before we act here.

internal/controllers/suite_test.go
m1kola: We also need to take into account that bingo will eventually become obsolete and we will likely switch to `go tool` for managing tool dependencies once it lands in Go. See #341 and relevant Go issues mentioned there.

m1kola:
> The goal of the PR never was to remove BINGO

Still - it is related as it changes how we manage our tool dependency. If we merge this - we will have two ways: we will still have bingo managing a bunch of tools and there will be `go install` managing `evntest` in `Makefile` installing it into a custom location. I would like us to be consistent with this the approach, if possible.

If there are issues with what we have today - let's address them. But first we need to understand these issues. That's why I invite you to write down your feedback in one place so we ca discuss it in a more structured way.

internal/controllers/suite_test.go
camilamacedo86: This PR never was about BINGO
Sorry, I got lost in the journey and reviews and reverted it to solve what it should address and was addressing since the beginner.

camilamacedo86:
Hi @joelanford, @m1kola, @perdasilva, @bentito,

Apologies for the confusion earlier!

This PR is NOT about Bingo. **The main goal is to make debugging ENVTEST-based tests in the IDE simpler and more straightforward.**

> _I suspect that some team members might think it’s not possible to debug with Ginkgo because the tests in question are actually ENVTEST-based. This change addresses that scenario and clarifies the process._

I’ve addressed all comments, closed all as done or outdated to avoid confusion, reverted the BINGO changes, and kept the suggestions.

@M1kola, could you please revert your block?

Thanks in advance! 😊

internal/controllers/suite_test.go
```diff
@@ -162,6 +162,21 @@ func TestMain(m *testing.M) {
...
+	// to be explicitly set when running tests directly (e.g., debugging tests in an IDE)
+	// without using the Makefile targets.
+	//
+	// This is equivalent to configuring your IDE to export the `KUBEBUILDER_ASSETS` environment
+	// variable before each test execution. The following function simplifies this process
+	// by handling the configuration for you.
+	//
+	// To ensure the binaries are in the expected path without manual configuration, run:
+	// `make setup-envtest`
+	if getFirstFoundEnvTestBinaryDir() != "" {
```
camilamacedo86, line 176: Hi @perdasilva,  

You added this comment, and I deviated—sorry about that!  

> Would it be cleaner if we just used an environment variable like `ENV_TEST_BINARY_DIR` or something like that?  
> Then, it doesn't really matter where the binaries are placed.  

We already have an environment variable (`KUBEBUILDER_ASSETS`) that can be used, but it’s not very convenient and involves a steep learning curve:  
- Know that ENVTEST requires binaries.
- Then, by default, ENVTEST is expected to find the bins in `local/...`.  
- That we can override this using the `KUBEBUILDER_ASSETS` environment variable.  
- Honestly, I’m not sure where Bingo was placing the specific bins used in the tests before.  
- Then, there’s the added challenge of knowing how to configure the tests properly in an IDE.  

This PR simplifies the process so everything “just works” without extra setup or deep knowledge. 
Making tests and debugging smoother and hassle-free.  

I’ve also updated the code comments to clarify things further.  
Hope this clears everything up! 😊

Makefile
```diff
@@ -151,19 +151,25 @@ test-ext-dev-e2e: $(OPERATOR_SDK) $(KUSTOMIZE) $(KIND) #HELP Run extension creat
...
+.PHONY: setup-envtest #HELP Install setup-envtest binary in the bin directory.
+setup-envtest: $(SETUP_ENVTEST)
+	mkdir -p $(ROOT_DIR)/bin
+	$(SETUP_ENVTEST) use -p env $(ENVTEST_VERSION) $(SETUP_ENVTEST_BIN_DIR_OVERRIDE)
+
+.PHONY: test-unit
 test-unit: $(SETUP_ENVTEST) #HELP Run the unit tests
 	rm -rf $(COVERAGE_UNIT_DIR) && mkdir -p $(COVERAGE_UNIT_DIR)
-	eval $$($(SETUP_ENVTEST) use -p env $(ENVTEST_VERSION) $(SETUP_ENVTEST_BIN_DIR_OVERRIDE)) && \
+	KUBEBUILDER_ASSETS="$(shell $(SETUP_ENVTEST) use -p path $(ENVTEST_VERSION) $(SETUP_ENVTEST_BIN_DIR_OVERRIDE))" \
```
camilamacedo86, line 166: Here’s a more concise version:

Hi @m1kola,

Just to clarify, the binaries installed by envtest-setup are no longer placed in the HOME/global directory. See the default behavior here:
https://github.com/kubernetes-sigs/controller-runtime/blob/e3347b5405cdd0da5bff527af3d406117938ba6b/tools/setup-envtest/README.md?plain=1#L57-L70

This change ensures binaries are placed locally, avoiding the conflicts and clutter caused by global installations.

To clarify, I’m referring to the binaries required for ENVTEST (not those managed by BINGO).

Screenshot 2024-11-19 at 01 22 59



Hope this clears things up! 😊

camilamacedo86: @perdasilva ^ That might clarify better.

Makefile
m1kola: Unresolving because I have a related question.

WIth `--bin-dir` set to a local directory - do we need a conditional for `/tmp/envtest-binaries`?

Makefile
```diff
@@ -151,19 +151,25 @@ test-ext-dev-e2e: $(OPERATOR_SDK) $(KUSTOMIZE) $(KIND) #HELP Run extension creat
...
+
+.PHONY: setup-envtest #HELP Install setup-envtest binary in the bin directory.
+setup-envtest: $(SETUP_ENVTEST)
+	mkdir -p $(ROOT_DIR)/bin
+	$(SETUP_ENVTEST) use -p env $(ENVTEST_VERSION) $(SETUP_ENVTEST_BIN_DIR_OVERRIDE)
+
+.PHONY: test-unit
 test-unit: $(SETUP_ENVTEST) #HELP Run the unit tests
 	rm -rf $(COVERAGE_UNIT_DIR) && mkdir -p $(COVERAGE_UNIT_DIR)
-	eval $$($(SETUP_ENVTEST) use -p env $(ENVTEST_VERSION) $(SETUP_ENVTEST_BIN_DIR_OVERRIDE)) && \
```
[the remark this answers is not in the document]
camilamacedo86: The goal of the change is to simplify.
Make it easier to figure out that the tests use `KUBEBUILDER_ASSETS` to set the path for the bins.

Makefile
camilamacedo86: @m1kola 

Is that make sense ?
Can we close this one?

Makefile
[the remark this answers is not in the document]
m1kola: I'm on the fence here because of repetition: we would have to now ensure that we pass the same args to both `$(SETUP_ENVTEST)` calls. But explicit is better than implicit I guess.

So I'll leave it up to you.

Makefile
```diff
@@ -151,19 +151,25 @@ test-ext-dev-e2e: $(OPERATOR_SDK) $(KUSTOMIZE) $(KIND) #HELP Run extension creat
...
 COVERAGE_UNIT_DIR := $(ROOT_DIR)/coverage/unit
-test-unit: $(SETUP_ENVTEST) #HELP Run the unit tests
+
+.PHONY: setup-envtest #HELP Install the binaries required to run ENVTEST-test based in the project/bin directory.
+setup-envtest: $(SETUP_ENVTEST)
+	mkdir -p $(ROOT_DIR)/bin
+	$(SETUP_ENVTEST) use -p env $(ENVTEST_VERSION) $(SETUP_ENVTEST_BIN_DIR_OVERRIDE)
+
+.PHONY: test-unit
+test-unit: $(SETUP_ENVTEST) setup-envtest #HELP Run the unit tests
```
m1kola, line 164: A bit strange to have `$(SETUP_ENVTEST)` and `setup-envtest` targets.

Is it beneficial to have a separate `setup-envtest` target here?

Maybe we can just do this?



```suggestion
test-unit: $(SETUP_ENVTEST) #HELP Run the unit tests
	mkdir -p $(ROOT_DIR)/bin
	$(SETUP_ENVTEST) use -p env $(ENVTEST_VERSION) $(SETUP_ENVTEST_BIN_DIR_OVERRIDE)
```

m1kola (commented):
Looks good to me overall.

Just one question about `$(SETUP_ENVTEST)` and `setup-envtest` targets.

Makefile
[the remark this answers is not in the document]
camilamacedo86: We need use eval for the go-test works otherwise it will not
So, I am closing it as sorted out

Makefile
camilamacedo86: - $(SETUP_ENVTEST) -> Downloads the binary for setting up envtest.
- setup-envtest -> Downloads the Kubernetes binaries and actually sets up envtest.

If we make the suggested change where we are not calling setup-envtest and we do not have an option to do only that, it would allow us to have the required binaries to run the IDE without running the WHOLE unit tests.

m1kola (dismissed):

internal/controllers/suite_test.go
perdasilva: Might also be nice to update the developer docs: https://github.com/operator-framework/operator-controller/blob/main/docs/contribute/developer.md

Makefile
perdasilva: maybe we could call the target: `setup-local-env` or something like that. Then, it could be a generic target for developers to bootstrap their local development environment. Right now, that may only be the envtest binaries, but there could me more in the future. Maybe we could also add an @echo "adding envtest binaries to bin/ - see [link to developer docs] for more details"?

internal/controllers/suite_test.go
camilamacedo86: Both are done 👍 
Thank you 🥇

camilamacedo86:
Thank you @perdasilva and @m1kola 

I am moving forward with this one
I think all suggestions and etc are addressed

However, we can always to push enhancements in a follow-up as well.

PR #1457, opened 2024-11-13
🌱 Change from using prometheus operator to a custom test operator for e2e

dtfranz:
If e2e is run on a cluster that already has prometheus-operator installed then all of our tests will fail. This PR replaces the prometheus-operator manifests with a fake bundle and also removes our reliance on the operatorhub catalog for the multiple catalogs test.

Borrows a lot from @tmshort 's [simple-bundle](https://github.com/tmshort/test-image/tree/main/simple-bundle) repo for the bundle content.


perdasilva (approved):
lgtm - Thank you ^^

Issue #1459, opened 2024-11-14
Update docs to use example which better demonstrates upgrades


m1kola:
Follow up to https://github.com/operator-framework/operator-controller/pull/1443:

- [ ] Need a better example for ["Upgrade an Extension" doc](https://github.com/operator-framework/operator-controller/blob/aaa0e0090ec0a04bbc990bbda0a4146852cf70bf/docs/tutorials/upgrade-extension.md): currently we use  argocd 0.5.0 -> 0.6.0 which fails CRD checks.
- [ ] We also need a better example for "Getting Started": we are [currently suggesting to upgrade](https://github.com/operator-framework/operator-controller/blob/main/docs/getting-started/olmv1_getting_started.md?plain=1#L79-L83) to 0.11.0 and it fails.

trgeiger:
So there's a couple things of note here:

[As you can see here](https://github.com/operator-framework/operator-controller/blob/main/docs/concepts/upgrade-support.md#upgrades-within-the-major-version-zero) we don't support automatic upgrades between minor versions within major version zero. So even if the CRD Upgrade Safety check passes, we can't automatically upgrade between, for example, 0.5.0 and 0.6.0 anyway. There's not many patch versions to choose from in ArgoCD, but I did find that 0.2.0 will upgrade to 0.2.1. Any versions newer than 0.6.0 have webhooks and so aren't supported.

Would changing the example to use 0.2.0 and 0.2.1 be good enough? If we do that, we should add a reference to this major version 0 limitation on the tutorial page. Or do we want to find a new ClusterExtension example entirely that has workable non-zero major versions? If we go that route I'll need to derive all the RBAC and basically create an entirely new sample manifest.

trgeiger:
Please see the draft linked PR for an example of changing to using 0.2.0 and 0.2.1.

As for the 2nd item in this issue, the use of 0.11.0 as the upgrade example, I'm not sure we can really use any of the newer versions of argocd without running afoul of either the CRD Upgrade Safety preflight check, the presence of webhooks, or the fact that we don't automatically upgrade between minor versions with a major version of 0. Is it okay to just stick to 0.2.0 and 0.2.1 of argocd in our example documentation or should we find another operator that would fulfill all these requirements as an example?

trgeiger:
[...]

PR #1469, opened 2024-11-14
📖 Update extension upgrade tutorial doc

trgeiger:
With the addition of the CRD Upgrade Safety preflight check, our existing example of upgrading argocd from 0.5.0 to 0.6.0 no longer serves as a good example. This commit changes the example to use an update from 0.2.0 to 0.2.1 which no longer fails the CRD Upgrade check, and also doesn't violate our restriction on automatic upgrades between minor versions with a major version of 0.

Addresses #1459 

# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/docs/getting-started/olmv1_getting_started.md
+++ b/docs/getting-started/olmv1_getting_started.md
@@ -73,12 +73,11 @@ kubectl apply -f https://raw.githubusercontent.com/operator-framework/operator-c
 To upgrade the installed extension, update the version field in the ClusterExtension resource. Note that
 there must be CRD compatibility between the versions being upgraded, and the target version must be
 compatible with OLM v1. More information on CRD upgrade safety can be found [here](../concepts/crd-upgrade-safety.md),
-compatible with OLM v1. More information on CRD upgrade safety can be found [here](../concepts/crd-upgrade-safety.md),
-and on the extension upgrade process [here](../tutorials/upgrade-extension.md).
+and more information on the extension upgrade process can be found [here](../tutorials/upgrade-extension.md).
 
 ```bash
 # Update to v0.11.0
-kubectl patch clusterextension argocd --type='merge' -p '{"spec": {"source": {"catalog": {"version": "0.11.0"}}}}'
+kubectl patch clusterextension argocd --type='merge' -p '{"spec": {"source": {"catalog": {"version": "0.2.1"}}}}'
 ```
 
 For information on the downgrade process, see [here](../tutorials/downgrade-extension.md).
--- a/docs/getting-started/olmv1_getting_started.md
+++ b/docs/getting-started/olmv1_getting_started.md
@@ -77,7 +77,7 @@ and more information on the extension upgrade process can be found [here](../tut
 
 ```bash
 # Update to v0.11.0
-kubectl patch clusterextension argocd --type='merge' -p '{"spec": {"source": {"catalog": {"version": "0.2.1"}}}}'
+kubectl patch clusterextension argocd --type='merge' -p '{"spec": {"source": {"catalog": {"version": "0.11.0"}}}}'
 ```
 
 For information on the downgrade process, see [here](../tutorials/downgrade-extension.md).
--- a/docs/tutorials/upgrade-extension.md
+++ b/docs/tutorials/upgrade-extension.md
@@ -11,16 +11,17 @@ For information on downgrading an extension, see [Downgrade an Extension](downgr
 
 ## Prerequisites
 
-* You have an extension installed
+* You have a ClusterExtension installed
 * The target version is compatible with OLM v1 (see [OLM v1 limitations](../project/olmv1_limitations.md))
-* CRD compatibility between the versions being upgraded or downgraded (see [CRD upgrade safety](../concepts/crd-upgrade-safety.md))
-* The installer service account's RBAC permissions are adequate for the target version (see [Minimal RBAC for Installer Service Account](../howto/derive-service-account.md))
+* Any changes to the CustomResourceDefinition in the new version meet compatibility requirements (see [CRD upgrade safety](../concepts/crd-upgrade-safety.md))
+* The installer ServiceAccount's RBAC permissions are adequate for the target version (see [Minimal RBAC for Installer Service Account](../howto/derive-service-account.md))
+* You are not attempting to upgrade between minor versions with a major version of zero (see [Upgrades within the major version zero](../concepts/upgrade-support.md#upgrades-within-the-major-version-zero))
 
 For more detailed information see [Upgrade Support](../concepts/upgrade-support.md).
 
 ## Procedure
 
-Suppose we have successfully created and installed v0.5.0 of the ArgoCD operator with the following `ClusterExtension`:
+Suppose we have successfully created and installed v0.2.0 of the ArgoCD operator with the following `ClusterExtension`:
 
 ``` yaml title="Example CR"
 apiVersion: olm.operatorframework.io/v1
@@ -35,7 +36,7 @@ spec:
     sourceType: Catalog
     catalog:
       packageName: argocd-operator
-      version: 0.5.0
+      version: 0.2.0
 ```
 
 * Update the version field in the ClusterExtension resource:
@@ -54,7 +55,7 @@ spec:
         sourceType: Catalog
         catalog:
           packageName: argocd-operator
-          version: 0.6.0 # Update to version 0.6.0
+          version: 0.2.1 # Update to version 0.2.1
     EOF
     ```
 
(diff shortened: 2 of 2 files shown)


trgeiger:
Should we change the ClusterExtension sample argocd manifest to match our docs at version 0.2.0 or 0.2.1? Or is it good enough to make these docs examples just work and leave the manifest alone at 0.6.0?

docs/getting-started/olmv1_getting_started.md
```diff
@@ -73,12 +73,11 @@ kubectl apply -f https://raw.githubusercontent.com/operator-framework/operator-c
...
 there must be CRD compatibility between the versions being upgraded, and the target version must be
 compatible with OLM v1. More information on CRD upgrade safety can be found [here](../concepts/crd-upgrade-safety.md),
-compatible with OLM v1. More information on CRD upgrade safety can be found [here](../concepts/crd-upgrade-safety.md),
-and on the extension upgrade process [here](../tutorials/upgrade-extension.md).
+and more information on the extension upgrade process can be found [here](../tutorials/upgrade-extension.md).
 
 ```bash
 # Update to v0.11.0
-kubectl patch clusterextension argocd --type='merge' -p '{"spec": {"source": {"catalog": {"version": "0.11.0"}}}}'
+kubectl patch clusterextension argocd --type='merge' -p '{"spec": {"source": {"catalog": {"version": "0.2.1"}}}}'
```
m1kola, line 80: Will this work?

I believe `https://raw.githubusercontent.com/operator-framework/operator-controller/main/config/samples/olm_v1_clusterextension.yaml` is currently installing `0.6.0`.

So if as a previous step we install `0.6.0` then here we effectively are trying to downgrade to `0.2.1`.

docs/tutorials/upgrade-extension.md
```diff
@@ -100,15 +88,15 @@ spec:
         metadata:
           annotations:
             kubectl.kubernetes.io/last-applied-configuration: |
-              {"apiVersion":"olm.operatorframework.io/v1","kind":"ClusterExtension","metadata":{"annotations":{},"name":"argocd"},"spec":{"namespace":"argocd","serviceAccount":{"name":"argocd-installer"},"source":{"catalog":{"packageName":"argocd-operator","version":"0.5.0"},"sourceType":"Catalog"}}}
-          creationTimestamp: "2024-11-11T14:13:12Z"
+              {"apiVersion":"olm.operatorframework.io/v1","kind":"ClusterExtension","metadata":{"annotations":{},"name":"argocd"},"spec":{"namespace":"argocd","serviceAccount":{"name":"argocd-installer"},"source":{"catalog":{"packageName":"argocd-operator","version":"0.2.1"},"sourceType":"Catalog"}}}
```
m1kola, line 91: I have a feeling that it should be `0.2.0` if we are upgrading from `0.2.0` to `0.2.1`.

I think it would be really good to have a reproducible example here. Something people can run end to end and see how it works.

m1kola: > Should we change the ClusterExtension sample argocd manifest to match our docs at version 0.2.0 or 0.2.1? Or is it good enough to make these docs examples just work and leave the manifest alone at 0.6.0?

I think the above comment is related to what you are asking here.

On one hand - having a asample which installs 0.2.0 version, while there is 0.12.0 already, is not great. But on the other hand - if it helps us to demonstrate things in docs, the it might be worth doing it.

Alternatively, we can include all the manifests (RBAC, service account, etc) required for 0.2.0 & 0.2.1 just in docs and make docs less reliant on the sample manifest. I think doing this might be the best approach here.

docs/getting-started/olmv1_getting_started.md
trgeiger: Yeah, that's why I have my comment/question above about if we should change the sample manifest. The existing example of using 0.11.0 doesn't work either because it's attempting to upgrade a minor version within a zero major version which we don't support.

This specific line in 'olmv1_getting_started.md' technically isn't part of any tutorial or anything, but rather a one-off example of what an upgrade command might look like. So I think we could get away with whatever we want here, although having it be functional and not show an example of non-functional minor version upgrade attempts would probably be best. I could rewrite it to not have any actual version in the command, maybe, but rather some placeholder?

docs/tutorials/upgrade-extension.md
trgeiger: That's an interesting idea. Do you think it would make the most sense to have those manifests somewhere in the beginning of this "getting started" doc?

And as for the output version here on line 91, it's showing 0.2.1 because this is the verification output after you perform the upgrade.

everettraven: >I expected the spec to have a new version, but kubectl.kubernetes.io/last-applied-configuration to have the previous spec (before we applied a new version). Is it not how it should work?

That annotation matches my expectations. We patched the ClusterExtension to 0.2.1 in using kubectl so it updated this annotation as part of that patch since we are applying a "new" configuration.

everettraven: I'm +1 for a collapsible admonition if we are going to deviate from the sample.

trgeiger: I can make a note admonition explaining this nuance with the last-applied-configuration value

docs/getting-started/olmv1_getting_started.md
trgeiger: @m1kola @everettraven so given we are adding a 0.2.0 manifest to the upgrade tutorial, should we just leave this line in olmv1_getting_started the same at 0.11.0 or maybe some other change?

trgeiger:
Admonition added at the beginning of the tutorial with the v0.2.0 manifests.

Admonition added at the very end explaining the possible discrepancy in last-applied-configuration depending on if you applied a new yaml file or used patch/edit to upgrade the ClusterExtension.

docs/tutorials/upgrade-extension.md
```diff
@@ -11,34 +11,310 @@ For information on downgrading an extension, see [Downgrade an Extension](downgr
...
+        sourceType: Catalog
+        catalog:
+          packageName: argocd-operator
+          version: 0.2.0
+    ```
+
+If we view the current state of our ClusterExtension we should see that we have installed version 0.2.0:
+
+```terminal
+kubectl get clusterextension argocd -o yaml | tail -n4
```
m1kola, line 306: Maybe worth using `jsonpath` or `jsonpath-as-json` instead of `tail` here? Or just show full yaml?

```
# kubectl get clusterextension argocd -o jsonpath="{.status.install}" | yq -p json
bundle:
  name: argocd-operator.v0.2.0
  version: 0.2.0
```

```
# kubectl get clusterextension argocd -o jsonpath-as-json="{.status.install}"
[
    {
        "bundle": {
            "name": "argocd-operator.v0.2.0",
            "version": "0.2.0"
        }
    }
]
```

docs/tutorials/upgrade-extension.md
```diff
@@ -117,49 +380,55 @@ spec:
...
             type: Progressing
           install:
             bundle:
-              name: argocd-operator.v0.6.0
-              version: 0.6.0
+              name: argocd-operator.v0.2.1
+              version: 0.2.1
         ```
+
+!!! note "Note on `last-applied-configuration`"
```
m1kola, line 430:
```suggestion
!!! note "Note on `kubectl.kubernetes.io/last-applied-configuration` annotation"
```

docs/tutorials/upgrade-extension.md
m1kola: You are right. I was patching. All good then. Resolving this thread.

docs/getting-started/olmv1_getting_started.md
m1kola: I don't like that we have duplication of content. Maybe we should consolidate the tutorial and getting started somehow?

m1kola (commented):
Just a few comments. The "Upgrade an Extension" looks good to me. We need to sort out the "Getting Started" page, but we can do it in a separate PR, if you prefer.

docs/tutorials/upgrade-extension.md
m1kola: Just to clarify: I'm suggesting this because I'm not sure if field ordering is guaranteed somehow in `kubectl get -o yaml` output.

m1kola:
@trgeiger makes senes. Should we revert / partially revert `docs/getting-started/olmv1_getting_started.md` here in that case? Or do you think it is still valuable to have these changes?

trgeiger:
I will revert for this PR, yeah. Best to leave it alone and address it directly elsewhere.

I did just notice something else, though--with the example 0.2.0 manifest that I added, when I apply that to a cluster my `kubectl get clusterextension argocd -o yaml` now includes this bit:
```
kind: List
metadata:
  resourceVersion: ""
  selfLink: ""
```

What is that from?

docs/tutorials/upgrade-extension.md
trgeiger: I think it would clean up this whole doc quite a bit to just use this jsonpath-as-path command for both instances where we check the clusterextension status. This would also allow us to remove that last note I added about the last-applied-configuration annotation since that won't show up in the tutorial anymore. What do you think?

trgeiger: I guess the note can stay, in case anyone does notice that discrepancy. Doesn't hurt anything.

m1kola: I had a look at the updated doc - looks good to me. Resolving this thread.

trgeiger:
I organized each step of the tutorial into "example" admonitions with clearer titles for the code blocks and admonition blocks. I think this should improve the overall organization and formatting of the doc. It should be ready for final review now.

PR #1475, opened 2024-11-17
✨ Replace kube-rbac-proxy with controller-runtime metrics authentication/authorization

camilamacedo86:
This commit removes the use of the kube-rbac-proxy image and replaces it with metrics authentication/authorization provided by controller-runtime. The kube-rbac-proxy image is deprecated and will no longer be maintained, which introduces risks to production environments. For more details, see: https://github.com/kubernetes-sigs/kubebuilder/discussions/3907

PR for catalogd: https://github.com/operator-framework/catalogd/pull/460
Motivation: https://github.com/operator-framework/operator-controller/issues/1509

## Tests done locally

**To ensure backwards compatibility**

Create the clusterrolebinding
```
$ kubectl create clusterrolebinding operator-controller-metrics-binding \
   --clusterrole=operator-controller-metrics-reader \
   --serviceaccount=olmv1-system:operator-controller-controller-manager
```

Create the TOKEN
```
TOKEN=$(kubectl create token operator-controller-controller-manager -n olmv1-system)
echo $TOKEN
```

Use curl to validate
```
kubectl label namespace olmv1-system pod-security.kubernetes.io/enforce-
kubectl run curl-metrics --rm -it --restart=Never \
  --image=curlimages/curl:7.87.0 -n olmv1-system -- /bin/sh
```

Call the metrics:

```
 curl -v -k -H "Authorization: Bearer $TOKEN" \
> https://operator-controller-controller-manager-metrics-service.olmv1-system.sv
c.cluster.local:8443/metrics
*   Trying 10.96.16.165:8443...
* Connected to operator-controller-controller-manager-metrics-service.olmv1-system.svc.cluster.local (10.96.16.165) port 8443 (#0)
* ALPN: offers h2
* ALPN: offers http/1.1
* [CONN-0-0][CF-SSL] TLSv1.3 (OUT), TLS handshake, Client hello (1):
* [CONN-0-0][CF-SSL] TLSv1.3 (IN), TLS handshake, Server hello (2):
* [CONN-0-0][CF-SSL] TLSv1.3 (IN), TLS handshake, Encrypted Extensions (8):
* [CONN-0-0][CF-SSL] TLSv1.3 (IN), TLS handshake, Certificate (11):
* [CONN-0-0][CF-SSL] TLSv1.3 (IN), TLS handshake, CERT verify (15):
* [CONN-0-0][CF-SSL] TLSv1.3 (IN), TLS handshake, Finished (20):
* [CONN-0-0][CF-SSL] TLSv1.3 (OUT), TLS change cipher, Change cipher spec (1):
* [CONN-0-0][CF-SSL] TLSv1.3 (OUT), TLS handshake, Finished (20):
* SSL connection using TLSv1.3 / TLS_AES_128_GCM_SHA256
* ALPN: server accepted h2
* Server certificate:
*  subject: [NONE]
*  start date: Dec 12 20:40:15 2024 GMT
*  expire date: Mar 12 20:40:15 2025 GMT
*  issuer: CN=olmv1-ca
*  SSL certificate verify result: unable to get local issuer certificate (20), continuing anyway.
* Using HTTP2, server supports multiplexing
* Copying HTTP/2 data in stream buffer to connection buffer after upgrade: len=0
* h2h3 [:method: GET]
* h2h3 [:path: /metrics]
* h2h3 [:scheme: https]
* h2h3 [:authority: operator-controller-controller-manager-metrics-service.olmv1-system.svc.cluster.local:8443]
* h2h3 [user-agent: curl/7.87.0-DEV]
* h2h3 [accept: */*]
* h2h3 [authorization: Bearer TOKEN
* Using Stream ID: 1 (easy handle 0xffff82fa3aa0)
> GET /metrics HTTP/2
> Host: operator-controller-controller-manager-metrics-service.olmv1-system.svc.cluster.local:8443
> user-agent: curl/7.87.0-DEV
> accept: */*
> authorization: Bearer TOKEN 
* [CONN-0-0][CF-SSL] TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
* Connection state changed (MAX_CONCURRENT_STREAMS == 250)!
< HTTP/2 200 
< content-type: text/plain; version=0.0.4; charset=utf-8; escaping=values
< date: Thu, 12 Dec 2024 20:43:10 GMT
< 
# HELP certwatcher_read_certificate_errors_total Total number of certificate read errors
# TYPE certwatcher_read_certificate_errors_total counter
certwatcher_read_certificate_errors_total 0
# HELP certwatcher_read_certificate_total Total number of certificate reads
# TYPE certwatcher_read_certificate_total counter
certwatcher_read_certificate_total 1
# HELP controller_runtime_active_workers Number of currently used workers per controller
# TYPE controller_runtime_active_workers gauge
....
workqueue_work_duration_seconds_count{controller="clusterextension",name="clusterextension"} 0
* Connection #0 to host operator-controller-controller-manager-metrics-service.olmv1-system.svc.cluster.local left intact
```

**Now, to validate the call with the certificates**

Create the POD with the secret
```
kubectl apply -f - <<EOF
apiVersion: v1
kind: Pod
metadata:
  name: curl-metrics
  namespace: olmv1-system
spec:
  serviceAccountName: operator-controller-controller-manager
  containers:
  - name: curl
    image: curlimages/curl:7.87.0
    command:
    - sh
    - -c
    - sleep 3600
    volumeMounts:
    - mountPath: /tmp/cert
      name: olm-cert
      readOnly: true
  volumes:
  - name: olm-cert
    secret:
      secretName: olmv1-cert
  restartPolicy: Never
EOF
```

```
$ kubectl exec -it curl-metrics -n olmv1-system -- sh

$ curl -v --cacert /tmp/cert/ca.crt --cert /tmp/cert/tls.crt --key /tmp/cert/t
ls.key \
> -H "Authorization: Bearer $TOKEN" \
> https://operator-controller-controller-manager-metrics-service.olmv1-system.sv
c.cluster.local:8443/metrics
*   Trying 10.96.16.165:8443...
* Connected to operator-controller-controller-manager-metrics-service.olmv1-system.svc.cluster.local (10.96.16.165) port 8443 (#0)
* ALPN: offers h2
* ALPN: offers http/1.1
*  CAfile: /tmp/cert/ca.crt
*  CApath: none
* [CONN-0-0][CF-SSL] TLSv1.3 (OUT), TLS handshake, Client hello (1):
* [CONN-0-0][CF-SSL] TLSv1.3 (IN), TLS handshake, Server hello (2):
* [CONN-0-0][CF-SSL] TLSv1.3 (IN), TLS handshake, Encrypted Extensions (8):
* [CONN-0-0][CF-SSL] TLSv1.3 (IN), TLS handshake, Certificate (11):
* [CONN-0-0][CF-SSL] TLSv1.3 (IN), TLS handshake, CERT verify (15):
* [CONN-0-0][CF-SSL] TLSv1.3 (IN), TLS handshake, Finished (20):
* [CONN-0-0][CF-SSL] TLSv1.3 (OUT), TLS change cipher, Change cipher spec (1):
* [CONN-0-0][CF-SSL] TLSv1.3 (OUT), TLS handshake, Finished (20):
* SSL connection using TLSv1.3 / TLS_AES_128_GCM_SHA256
* ALPN: server accepted h2
* Server certificate:
*  subject: [NONE]
*  start date: Dec 12 20:40:15 2024 GMT
*  expire date: Mar 12 20:40:15 2025 GMT
*  subjectAltName: host "operator-controller-controller-manager-metrics-service.olmv1-system.svc.cluster.local" matched cert's "operator-controller-controller-manager-metrics-service.olmv1-system.svc.cluster.local"
*  issuer: CN=olmv1-ca
*  SSL certificate verify ok.
* Using HTTP2, server supports multiplexing
* Copying HTTP/2 data in stream buffer to connection buffer after upgrade: len=0
* h2h3 [:method: GET]
* h2h3 [:path: /metrics]
* h2h3 [:scheme: https]
* h2h3 [:authority: operator-controller-controller-manager-metrics-service.olmv1-system.svc.cluster.local:8443]
* h2h3 [user-agent: curl/7.87.0-DEV]
* h2h3 [accept: */*]
* h2h3 [authorization: Bearer TOKEN
* Using Stream ID: 1 (easy handle 0xffffb0cf9aa0)
> GET /metrics HTTP/2
> Host: operator-controller-controller-manager-metrics-service.olmv1-system.svc.cluster.local:8443
> user-agent: curl/7.87.0-DEV
> accept: */*
> authorization: Bearer TOKEN
> 
* [CONN-0-0][CF-SSL] TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
* Connection state changed (MAX_CONCURRENT_STREAMS == 250)!
< HTTP/2 200 
< content-type: text/plain; version=0.0.4; charset=utf-8; escaping=values
< date: Thu, 12 Dec 2024 20:46:29 GMT
< 

```


camilamacedo86:
c/c @joelanford

cmd/manager/main.go
```diff
@@ -97,7 +102,11 @@ func main() {
 		caCertDir                 string
 		globalPullSecret          string
 	)
-	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
+	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8443", "The address the metric endpoint binds to.")
```
camilamacedo86, line 105: We should use 8443 not 8080

cmd/manager/main.go
```diff
@@ -97,7 +102,11 @@ func main() {
...
+	flag.BoolVar(&secureMetrics, "metrics-secure", true,
+		"If set, the metrics endpoint is served securely via HTTPS. Use --metrics-secure=false to use HTTP instead.")
+	flag.BoolVar(&enableHTTP2, "enable-http2", false,
+		"If set, HTTP/2 will be enabled for the metrics and webhook servers")
```
camilamacedo86, line 109: Here we just adding the same flags as we have in kubebuilder

cmd/manager/main.go
```diff
@@ -97,7 +102,11 @@ func main() {
 		caCertDir                 string
 		globalPullSecret          string
 	)
-	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
+	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8443", "The address the metric endpoint binds to.")
+	flag.BoolVar(&secureMetrics, "metrics-secure", true,
+		"If set, the metrics endpoint is served securely via HTTPS. Use --metrics-secure=false to use HTTP instead.")
+	flag.BoolVar(&enableHTTP2, "enable-http2", false,
```
perdasilva, line 108: should we omit this for now, until http2 becomes a thing we need to think about? Otherwise, we may increase our support surface unnecessarily. Though, it's easy to add if we ever need to.

cmd/manager/main.go
```diff
@@ -161,9 +164,33 @@ func main() {
 			},
 		}
 	}
+
+	// http/2 should be disabled due to its vulnerabilities. More specifically,
+	// disabling http/2 will prevent from being vulnerable to the HTTP/2 Stream
```
bentito, line 169:
```suggestion
	// disabling http/2 will prevent it from being vulnerable to the HTTP/2 Stream
```
Though more generally I think we could just replace the whole comment with: `http/2 should be disabled due to its vulnerabilities.` If they want to know more they can search the web.

config/base/rbac/auth_proxy_role.yaml
joelanford, line 1: Don't we still need this?

config/base/rbac/auth_proxy_service.yaml
joelanford, line 1: Don't we still need this?

config/base/rbac/auth_proxy_role.yaml
joelanford: Oh nevermind, I see this is re-created with a different name.

After we release 1.0.0, we will need to be extremely careful about renaming objects. Renaming things will generally mean that the old objects will be left around on a cluster unless we specifically have a mechanism to clean them up.

config/base/manager/manager.yaml
```diff
@@ -52,7 +52,7 @@ spec:
         - /manager
         args:
         - "--health-probe-bind-address=:8081"
-        - "--metrics-bind-address=127.0.0.1:8080"
+        - "--metrics-bind-address=127.0.0.1:8443"
```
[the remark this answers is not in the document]
camilamacedo86: Good catcher

config/base/rbac/auth_proxy_service.yaml
camilamacedo86: No, we do not need any more

The 

```
- auth_proxy_service.yaml
- auth_proxy_role.yaml
- auth_proxy_role_binding.yaml
- auth_proxy_client_clusterrole.yaml
```

Is replace for

```
- metrics_auth_role.yaml
- metrics_auth_role_binding.yaml
- metrics_reader_role.yaml
```

joelanford:
/hold

This doesn't seem like a change that should be made after we've released 1.0.0-rc1. Let's wait until after we cut 1.0.0.

config/base/manager/manager.yaml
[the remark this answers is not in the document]
camilamacedo86: Done

camilamacedo86:
Hi @joelanford 

The way that it is today, it will stop working in the first quarter of the next year.
So, are you sure that we should release 1.0.0 before fix it?

joelanford:
Let's prioritize a design for this. I do not want to add any risk to our 1.0.0 in the final week after we've already released an RC.

config/base/rbac/auth_proxy_role.yaml
camilamacedo86: I reverted and I kept the same files/names 
no changes on that 👍

cmd/manager/main.go
```diff
@@ -97,7 +100,7 @@ func main() {
 		caCertDir                 string
 		globalPullSecret          string
 	)
-	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
+	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8443", "The address the metric endpoint binds to.")
```
tmshort, line 103: I would suggest mirroring `catalogd`'s options for certificates, rather than using fixed names below. This will provide flexibility (i.e. downstream).

cmd/manager/main.go
```diff
@@ -161,9 +164,40 @@ func main() {
...
+		// These configurations ensure that only authorized users and service accounts
+		// can access the metrics endpoint.
+		FilterProvider: filters.WithAuthenticationAndAuthorization,
+
+		// Ensure that metrics is protected with certs managed by cert-manager
+		// If not informed, the metrics service provided by controller-runtime will generate
+		// and use its own self-assigned certs which is not recommended for production envs.
+		CertDir:  "/var/certs/",
+		CertName: "olm-ca.crt",
+		KeyName:  "ca.crt",
```
tmshort, line 195: These ought to be configured similarly to catalogd.

cmd/manager/main.go
```diff
@@ -161,9 +164,40 @@ func main() {
...
+	// disabling http/2 will prevent it from being vulnerable to the HTTP/2 Stream
+	// Cancellation and Rapid Reset CVEs. For more information see:
+	// - https://github.com/advisories/GHSA-qppj-fm5r-hxr3
+	// - https://github.com/advisories/GHSA-4374-p667-p6c8
+	disableHTTP2 := func(c *tls.Config) {
+		setupLog.Info("disabling http/2")
+		c.NextProtos = []string{"http/1.1"}
+	}
+
+	tlsOpts = append(tlsOpts, disableHTTP2)
```
tmshort, line 178: It appears that TLSv1.2 is the default minimum for TLS version, so we probably don't need to add an explicit minimum.

cmd/manager/main.go
camilamacedo86: Yes, we need to do the same here: https://github.com/operator-framework/catalogd/pull/460

cmd/manager/main.go
camilamacedo86: That is an amazing good point for us to validate!

camilamacedo86: I think you are right. But, we need to set the above code  "http/1.1 to prevent the CVE

cmd/manager/main.go
```diff
@@ -161,9 +175,48 @@ func main() {
...
+
+	// Create a new certificate watcher to watch for changes in the certificate files.
+	// If the certificate files change, the certificate watcher will reload the certificate
+	// and key files and update the TLS configuration.
+	cw, err := certwatcher.New(certFile, keyFile)
```
camilamacedo86, line 194: @tmshort to speed it up I might need your help.
See that all works fine until we try to use certwatcher, see that e2e tests before the last commit and this one fail in the e2e tests: https://github.com/operator-framework/operator-controller/actions/runs/12032812266/job/33545503759?pr=1475

cmd/manager/main.go
trgeiger: Those CVEs are for golang.org/x/net below version 0.17.0 and grpc below version 1.58.3. We are on 0.31.0 net and 1.67.1 grpc so we don't need to do anything here.

camilamacedo86: OH thank you for check that so that is something that we can remove great !!!!

cmd/manager/main.go
```diff
@@ -161,9 +178,37 @@ func main() {
...
+		// is not recommended for production use.
+		if len(certFile) > 0 && len(keyFile) > 0 {
+			// If the certificate files change, the watcher will reload them.
+			var err error
+			certWatcher, err = certwatcher.New(certFile, keyFile)
+			if err != nil {
+				log.Fatalf("Failed to initialize certificate watcher: %v", err)
+			}
+			metricsServerOptions.TLSOpts = append(metricsServerOptions.TLSOpts, func(config *tls.Config) {
+				config.GetCertificate = certWatcher.GetCertificate
```
camilamacedo86, line 204: @tmshort, everything will work fine until we start to use certWatcher here.
I don't understand why the tests are failing.
I will need to investigate and try to understand those tests, but if you already know that, I would appreciate your help.

cmd/manager/main.go
```diff
@@ -89,17 +94,24 @@ func podNamespace() string {
...
 		operatorControllerVersion bool
 		systemNamespace           string
 		caCertDir                 string
 		globalPullSecret          string
 	)
-	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
+	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8443", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.StringVar(&caCertDir, "ca-certs-dir", "", "The directory of TLS certificate to use for verifying HTTPS connections to the Catalogd and docker-registry web servers.")
+	flag.BoolVar(&secureMetrics, "metrics-secure", true,
```
trgeiger, line 111: If this is a default value to true, would it make more sense to make this something like --disable-metrics-https and make the description state that it's secure by default?

cmd/manager/main.go
```diff
@@ -89,17 +94,24 @@ func podNamespace() string {
...
+		"If set, the metrics endpoint is served securely via HTTPS. Use --metrics-secure=false to use HTTP instead.")
+	flag.StringVar(&certFile, "tls-cert", "", "The certificate file used for serving catalog contents over HTTPS. Requires tls-key.")
+	flag.StringVar(&keyFile, "tls-key", "", "The key file used for serving catalog contents over HTTPS. Requires tls-cert.")
```
trgeiger, line 114: The descriptions of these two flags are copied from catalogd, they do not apply to their usage here in operator-controller

config/components/tls/patches/manager_deployment_cert.yaml
```diff
@@ -1,9 +1,15 @@
 - op: add
   path: /spec/template/spec/volumes/-
-  value: {"name":"olmv1-certificate", "secret":{"secretName":"olmv1-cert", "optional": false, "items": [{"key": "ca.crt", "path": "olm-ca.crt"}]}}
+  value: {"name":"olmv1-certificate", "secret":{"secretName":"olmv1-cert", "optional": false, "items": [{"key": "ca.crt", "path": "olm-ca.crt"}, {"key": "tls.crt", "path": "tls.crt"}, {"key": "tls.key", "path": "tls.key"}]}}
```
tmshort, line 3:
```suggestion
  value: {"name":"olmv1-certificate", "secret":{"secretName":"olmv1-cert", "optional": false, "items": [{"key": "ca.crt", "path": "olm-ca.crt"}, {"key": "tls.crt", "path": "tls.cert"}, {"key": "tls.key", "path": "tls.key"}]}}
```

tmshort:
I am seeing the following in the logs:
```
  error resolving canonical reference: error creating image source: missing client certificate tls.cert for key tls.key
```
and
```
E1203 01:00:09.253529       1 controller.go:316] "Reconciler error" err="error resolving canonical reference: error creating image source: missing client certificate tls.cert for key tls.key" controller="clusterextension" controllerGroup="olm.operatorframework.io" controllerKind="ClusterExtension" ClusterExtension="clusterextension-ntd2r7bq" namespace="" name="clusterextension-ntd2r7bq" reconcileID="db373c5a-1cc0-4aa5-91dd-c249e06e5cd1"
```
It seems that the golang library code is looking for `tls.cert` and not `tls.crt` (even though I thought it was smart enough to handle either...)

**HOWEVER…**

It’s odd that it’s saying `missing client certificate`, which sounds as though we may be trying to connect using TLS client authentication, because the client may be finding `tls.key` and wanting a certificate to go with it, but we aren’t doing TLS client authentication.
This tells me that we may need to separate the CA certs into `/var/ca-certs` and the TLS server servers into `/var/certs` as is done for catalogd, instead of having them in the same directory. This likely due to the client finding the `tls.key` file among the CA certs and thinking it should do certificate authentication. Separating the CA certs and the server key+cert may be the way to go.

cmd/manager/main.go
tmshort: See my comments below. This is due to how the manifests are set up and we may want to change how things are arranged. It's not an issue with the code proper.

camilamacedo86:
HI @tmshort,

It seems a very reasonable thought https://github.com/operator-framework/operator-controller/pull/1475#issuecomment-2513374826

cmd/manager/main.go
camilamacedo86: If we use `--disable-metrics-https`, then we would need to explicitly set `true` in the code source to say that metrics is secure by default, rather than as the flag's default value. Would `--disable-metrics-https` default to `false`? And if it’s not explicitly disabled, do we assume metrics are secure? 

_(if !siableMetricsHttps then == MetricsSecure .. little confuse right?)_

Here’s how I see it:

- **Is the metrics server secure (`true`|`false`)?**
- **Do we want the metrics server to start secure or not by default?**

I believe we should secure metrics by default. Therefore, the value of the flag should be `true`. This means, if someone wants to start the manager with metrics in an insecure mode, they would need to explicitly set `--metrics-secure=false`.

For this reason, I’d prefer to keep the current flag name as it aligns with the conventions used in tools like Kubebuilder/SDK. While we’ve deviated in some areas (IHMO), staying as close as possible to those conventions may just make the project more intuitive for contributors.

I hope that makes sense. 
But if everybody wants to change the name, it is fine for me too.

trgeiger: The issue is that the description of the flag is not accurate (and this applies to kubebuilder, too, which I just looked at per your comment and it uses the same flag and description). It says "if set, then the endpoint is secured" but if it defaults to true, then that's not entirely true--it's also secured when this flag isn't even called. 

As for how my suggestion would work, I was just imagining a boolean field that defaults to false, then if set to true we skip all the https logic. That's typically how boolean flags to disable a default work, i.e. the -k/--insecure flag in curl for disabling https verification. Having a boolean flag that you pass an explicit boolean value to like `--flag=true or --flag=false` is an odd usage pattern that I haven't really seen. You can see we use this pattern in the other 2 flag.BoolVar values in operator-controller, 'leader-elect' and 'version'.

To illustrate what I mean, in the current PR code:
```
# Secure
manager 
manager --metrics-secure

# Intentionally not secure
manager --metrics-secure=false
```
The two secure examples do the same thing, the metrics are secured, so putting the flag in the 2nd case is unnecessary and potentially confusing if a user is expecting the first case to be insecure by default--typically if you have a flag named something like --metrics-secure, the existence of that flag implies you need that flag in order to enable that capability.

Compare that to something like:
```
# Secure by default
manager

# Intentionally not secure
manager --disable-metrics-tls
```

Here the flag name is directly stating what's changing and since it's optional, a user doesn't have to worry about it at all if they want to leave things secure.

Defaulting boolean flags to false is a pretty widespread standard practice because it's better UX. Some more explanation: https://serendipidata.com/posts/naming-guidelines-for-boolean-variables#guideline-2-always-choose-names-that-enable-setting-false-by-default

camilamacedo86: If we follow your suggestion

`manager --disable-metrics-tls`

That means, it MUST be secure metrics be default right

So, that means we have in the code

If --disable-metrics-tls NOT true then ensure metrics secure.

That is confusing.

camilamacedo86: > it's also secured when this flag isn't even called.

Becuase --metrics-secure by default, it is true
We want it secure by default.
If no flag is informed is secure

camilamacedo86: Because `--metrics-secure` defaults to `true`, we ensure security by default.

**If no flag is provided, the service remains secure.**

We do **not** want a situation where someone runs the project and the Metrics Service is **unsecured** simply because they didn't set the flag. By default, it must be protected. Users need to consciously decide and understand the consequences of running it in an unsecure mode.

This is why I disagree with the suggestion to default it to `false`.

camilamacedo86: Hi @tylerslaton 

Thank you for the detailed feedback! I appreciate the clarification and examples you provided. Here's my perspective:

**Flag Description Accuracy**
You’re absolutely right that the previous flag description (“if set, then the endpoint is secured”) was misleading, as the service is secure by default if the flag isn’t explicitly set. To address this, I’ve updated the description to explicitly reflect the default behavior, reducing potential confusion:

```go
flag.BoolVar(&secureMetrics, "metrics-secure", true, "Controls whether the metrics endpoint is served securely via HTTPS. Defaults to true (secure). Use --metrics-secure=false to serve metrics over HTTP (insecure).") 
```

**Defaulting to Secure by Default**
**The default of true prioritizes security without requiring explicit configuration.** This aligns with best practices for secure-by-default design and minimizes the risk of accidental misconfiguration.

> Security by default "means that the [default](https://itlaw.fandom.com/wiki/Default) [configuration settings](https://itlaw.fandom.com/wiki/Configuration_setting) are the most [secure](https://itlaw.fandom.com/wiki/Secure) settings possible, which are not necessarily the most [user friendly](https://itlaw.fandom.com/wiki/User_friendly) settings."
> https://itlaw.fandom.com/wiki/Security_by_default

REF: https://www.ncsc.gov.uk/information/secure-default

WDYT? Could we agree with the approach?
Can we mark it as solved?

camilamacedo86: Thank you for the suggestion! I agree that using --disable-metrics-tls is a clearer and more intuitive approach. It aligns well with secure-by-default principles while making it explicit when someone opts for an insecure configuration

Done so I am closing this one :-)

cmd/manager/main.go
```diff
@@ -337,6 +392,14 @@ func main() {
 
 	//+kubebuilder:scaffold:builder
 
+	if !disableMetricsTLS && certWatcher != nil {
```
trgeiger, line 395: Could this be moved up to the section starting at line 188 where you're checking !disableMetricsTLS already? Near the end of it after you're creating the certWatcher, around 210 or something?

camilamacedo86: I cannot be done before the `mgr, err := ctrl.NewManager` and must be done with the others controllers. 
We can move to up but not close to 188 line. It can be after the `certPoolWatcher ` for example,

tmshort (approved):
/lgtm

camilamacedo86:
Rebased with the tests to ensure backwords compatibility: https://github.com/operator-framework/operator-controller/pull/1522

camilamacedo86:
Hi @joelanford 

Can we unhold this one now?

camilamacedo86:
Hi @joelanford 

Can we unhold this one now?

cmd/manager/main.go
```diff
@@ -161,9 +175,49 @@ func main() {
 			},
 		}
 	}
+
+	// Force serving be disabled by default if no certs are provided
```
trgeiger, line 179: This comment is actually describing line 184, right?

cmd/manager/main.go
```diff
@@ -97,9 +104,11 @@ func main() {
 		caCertDir                 string
 		globalPullSecret          string
 	)
-	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
+	flag.StringVar(&metricsAddr, "metrics-bind-address", "0", "The address for the metrics endpoint. Use :8443 for HTTPS or set to 0 to disable. Requires both tls-cert and tls-key.")
```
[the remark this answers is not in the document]
trgeiger: I see in the Options struct that setting to 0 disables the entire metric server, so I answered that question for myself. In which case, that should be more directly stated in the help message here, i.e. "or set to 0 to disable the metrics server."

Was a decision made that we now want to have the metrics server disabled by default?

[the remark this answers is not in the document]
camilamacedo86: We can improve the description if you see that is required, but renaming the flag is not an option because it is already in use, and changing it would result in a breaking change for us. So, I think we do not discuss it. 

The flag is intended to specify the port used for the metrics server. If set to 0, it will disable the metrics server.
See the RFC: https://docs.google.com/document/d/1hYOtHWcfoX_zWvy2YeolQBGuDgmP72vXMlrj9q9kPBQ/edit?tab=t.0#heading=h.x3tfh25grvnv

Screenshot 2024-12-18 at 17 59 29

You can find there the link for the code in controller-runtime.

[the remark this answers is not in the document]
trgeiger: Also, you don't have to use 8443 to serve HTTPS, you can use any port you want. Is this something we are wanting to enforce on users?

If we do want to force using 8443 or 0 to disable, then this flag should be rewritten as a boolean flag.

If we don't care which port users use for the metrics server, then this needs to have rewritten help text and probably a different default value.

Do we want to allow users to run the metrics server on plain HTTP?

As the code is right now, as a user I might read these help messages for the available flags and think I just need to provide a TLS cert and key to have my secure metrics server, but since metrics-bind-address defaults to 0, it's going to be disabled. 

I think depending on the answers to my questions above, we need to better enforce that the three flag options of keyfile, certfile, and metricsBindAddr are either all used together, or none are used--if a user uses one or two but not the third, then we should probably fail and tell the user why. Right now some of those edge cases of using only some of the options are not covered.

cmd/manager/main.go
camilamacedo86: I will improve the comment to bring more clarity. That might be a good point. 🥇

cmd/manager/main.go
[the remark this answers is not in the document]
camilamacedo86: > Also, you don't have to use 8443 to serve HTTPS, you can use any port you want. Is this something we are wanting to enforce on users?

We want to set 8443. See: 

https://github.com/operator-framework/operator-controller/pull/1475/files#diff-e40019a0cad999763c42de1c2a364c2fa74258f5d5cabc4c36306b50ddfac0cbR57-R60

> If we do want to force using 8443 or 0 to disable, then this flag should be rewritten as a boolean flag.

No, because tomorrow, we might want or need to change it.
Changing the type from string to boolean can cause breaking changes. 
Today, it uses 8443 by default for the metrics, and we want to avoid addressing breaking changes.

cmd/manager/main.go
```diff
@@ -161,9 +175,49 @@ func main() {
...
+			// Rapid Reset CVEs. For more information see:
+			// - https://github.com/advisories/GHSA-qppj-fm5r-hxr3
+			// - https://github.com/advisories/GHSA-4374-p667-p6c8
+			// Besides, those CVEs are solved already; the solution is still insufficient, and we need to mitigate
+			// the risks. More info https://github.com/golang/go/issues/63417
+			setupLog.Info("disabling http/2")
+			config.NextProtos = []string{"http/1.1"}
+		})
+	} else {
+		setupLog.Info("WARNING: Metrics Server will not be serving")
```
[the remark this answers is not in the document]
camilamacedo86: Changed

cmd/manager/main.go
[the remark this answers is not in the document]
trgeiger: It's not using 8443 by default, you're defaulting it to 0. That link to the manager.yaml file shows that we set 8443 by default in that manifest, but the manifest defaults do not necessitate a hard-coded value in the go code--users can change the manifest all they want, the code needs to be able to handle those kinds of changes. The current code on main allows the user to set any port they want, that's the purpose of the flag.

If we want to preserve as much of the current user experience as possible, then I see two different options, depending on if we want to allow users to run metrics without HTTPS or not:

- If we don't want to allow HTTP, then you need to make the metrics-bind-address flag default to 8443, change the message to something like "The address for the metrics endpoint. Set to 0 to disable the metrics server." and make tls-cert and tls-key required flags.
- If we do want to allow HTTP, then make the metrics-bind-address flag default to 8443, change the message to something like "The address for the metrics endpoint. Will serve on HTTPS if tls-cert and tls-key are provided. Set to 0 to fully disable the metrics server." and add logic for handling setting up the metrics server options without TLS and with the provided port

But all of those cases should be able to handle any valid port number passed to the argument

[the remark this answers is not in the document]
trgeiger: Another thought I had was that in the code on 'main,' the ability to disable the metrics server is not exposed or explained to users--being able to set the flag to 0 to disable the server is a detail you only see if you dig into the controller-runtime package, so it's not obvious to me that we wanted users to be able to disable the server?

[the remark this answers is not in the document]
camilamacedo86: ## Why can't `metrics-bind-address` default to any value other than `0`?

We want to ensure that the metrics server is **disabled by default** (`0` if no port is specified). This follows a **security-by-default** principle, prioritizing safety and minimizing unnecessary exposure.

## Why not make `tls-cert` and `tls-key` required flags?

The solution should be able to run **without exporting metrics**. If metrics are not required, we simply do not want the metrics server running in such cases, avoiding unnecessary configuration or resources.

## Why change the description?

We recommend using **port 8443** for exporting metrics, as it is our **default port**.

Using any other port might lead to potential issues:
- Conflicts with ports already in use (e.g., webhook ports).  
- Inconsistencies with our documentation and examples, which often assume metrics are exported on **port 8443**.  

While we don’t block the use of other ports (to remain flexible if a change is needed), there’s no strong reason to suggest using an arbitrary port. Consistency and clarity make 8443 the preferred option.

But if you have a better suggestion for the description that still adding, recommending 8443(our default) I am totally fine to change it out.

cmd/manager/main.go
```diff
@@ -97,9 +104,11 @@ func main() {
 		caCertDir                 string
 		globalPullSecret          string
 	)
-	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
+	flag.StringVar(&metricsAddr, "metrics-bind-address", "0", "The address for the metrics endpoint. Use :8443 for HTTPS or set to 0 to disable the metrics server. Requires both tls-cert and tls-key.")
```
trgeiger, line 107: Okay, so I see that in the accepted RFC you wrote that you specified that this would be set to 0 by default, which means that metrics are disabled by default.

Since that's the accepted change in the RFC, I would suggest at least tweaking the help text here so it doesn't sound like the flag only accepts either 8443 or 0--I know that we would *prefer* users use 8443 and that our manifest will use 8443 by default, but it's not technically true that this flag only accepts 8443 and 0, which is how it reads right now. 

Since it's set to 0 by default, you don't need to mention setting it to 0 manually. Simply omitting the flag disables the server, which means setting the flag at all enables it. So, the text should say something like "Enable the metrics server at the specified endpoint address. Requires both tls-cert and tls-key. Suggested value of \":8443\""

I do have reservations about this design of defaulting to 0 though. Since we do have a preferred port of :8443, it makes more sense to me to make the default be 8443, with help text like "The address for the metrics endpoint. Defaults to :8443. Requires both tls-cert and tls-key. Set to 0 to completely disable the metrics server." but even this seems like it's cramming too much functionality into a single flag since you're doing what the flag is named, setting the metrics bind address, as well as controlling if the server is enabled or disabled, AND it's dependent on the presence of 2 other flags. Complex flags like this can quickly become confusing to a user. Ideally you'd control if it's disabled/enabled with one flag, what the address is with another flag (the existing one), and check for the proper application of those flags together like we do with tls-key and tls-cert.

__
A separate concern I have is by changing the default from 8080 to 0, any previous usage of operator-controller that didn't specify a value for --metrics-bind-address but rather relied on the default value of 8080 will go from having a functioning server endpoint to having it disabled after this merge (not that there's tons of usage out there right now, but just in the spirit of not introducing breaking changes). That's a silent breaking change which won't error out to alert the user that core functionality is altered. Whereas, if you default to 8443, you keep the program running the same way the user intended, with the server enabled, but they'll get the proper error output alerting them to the fact that they now need to provide tls-cert and tls-key.

camilamacedo86: >but it's not technically true that this flag only accepts 8443 and 0

Why the flag only accepts :8443 
Why could we not pass other value
We want / recommend :8443 but still possible use another right

camilamacedo86: I updated the description.
Also, see that I fixed the map of the port
It should be done in the service
Let me know if that address your main concerns and if we can move forward.

LalatenduMohanty: I agree with most of what @trgeiger mentioned. if the user can omit the flag then the default of `metrics-bind-address` flag should be 8443. Also have same concern about breaking change.

camilamacedo86: Takeaways

- If cert/key are not provided, metrics-bind-addr is not allowed
- If cert/key are provided, and metrics-bind-addr is not, we default to :8443
- If cert/key are provided and metrics-bind-addr is also provided, we honor what is provided.

camilamacedo86: So closing this one as sorted out

limck5856 (approved):

LalatenduMohanty (approved):
/lgtm

cmd/manager/main.go
```diff
@@ -97,9 +103,11 @@ func main() {
 		caCertDir                 string
 		globalPullSecret          string
 	)
-	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
+	flag.StringVar(&metricsAddr, "metrics-bind-address", "", "The address for the metrics endpoint. Requires tls-cert and tls-key. (Default: ':8443')")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.StringVar(&caCertDir, "ca-certs-dir", "", "The directory of TLS certificate to use for verifying HTTPS connections to the Catalogd and docker-registry web servers.")
+	flag.StringVar(&certFile, "tls-cert", "", "The certificate file used for serving metrics contents over HTTPS. Requires tls-key.")
```
trgeiger, line 109: "The certificate file used for the metrics server. Required to enable the metrics server. Requires tls-key."

cmd/manager/main.go
```diff
@@ -97,9 +103,11 @@ func main() {
...
+	flag.StringVar(&keyFile, "tls-key", "", "The key file used for serving metrics contents over HTTPS. Requires tls-cert.")
```
trgeiger, line 110: "The key file used for the metrics server. Required to enable the metrics server. Requires tls-cert."

trgeiger (approved):
lgtm

LalatenduMohanty (approved):
/lgtm

PR #1476, opened 2024-11-17
✨ (Blocked by #1475) - Ensure that metrics is protected with TLS for Prometheus integration.

camilamacedo86:
⚠️ This PR is blocked by #1475 ; the goal is to merge the second commit. After we merge (#1475) we can rebase and move forward within 

This change replaces controllers-runtime's self-signed certificates for metrics with cert-manager-managed certificates, ensuring security. We should not use `insecureSkipVerify: true`. Also, it is not a good practice for production env relay on the certs self-signed generated by controller-runtime and/or the old kube-rbac-proxy according to sig-security. 

Key updates:
- Configured `metricsServerOptions` to use cert-manager-managed certificates (`/var/metrics/certs`).
- Added `ClusterIssuer` (`olmv1-metrics-ca`) and `Certificate` resources for automated certificate management.
- Updated `ServiceMonitor` to enable secure TLS scraping by Prometheus using certificates issued by `olmv1-metrics-ca`.
- The deployment was pushed to mount metrics certificates as secrets.

Benefits:
- Enhanced security with automated certificate lifecycle management.
- Production-ready TLS setup for Prometheus metrics scraping.


tmshort:
We don't need a separate service (although we do have a separate one now)/certificate for the metrics, they can all share the same certificates by adding additional hostnames to the Certificate. Even if they are different services, they can still share the same certificate.
1) Make the configuration here much simpler
2) The certificate should use the existing Certificate/Issuer`olmv1-ca` rather than creating another one. Even better would be to add the service names into the existing certificate so that the certificate can be shared.
4

tmshort:
This should use the existing `olmv1-ca` to sign the certificate. No need to add another.

This should add support to rotate the certificates via `controller-runtime/pkg/certwatcher`

camilamacedo86:
Hi @tmshort 

I will create a Brief and RFC to explain why it is required
Also, I will follow your suggestion that will simplify things
By last keep all in the same PR to add as example in the RFC

And yes, you are right. We need to do the same with catalogD

camilamacedo86:
Hi @tmshort 

Thank you for your input. That was very, very helpful.
Since we could simplify for now, I will include all in #1475 so it will be easier to use it in the RFC.
But I will follow your input. !

Closing in favor of #1475

Issue #1297, opened 2024-09-23
Improve error message when ClusterExtension ServiceAccount is missing


thetechnick:
When creating a ClusterExtension with `.spec.install.serviceAccount` pointing to a non-existent ServiceAccount, Operatort Controller is showing a potentially misleading message in the condition status. (See in example below)

`Kubernetes cluster unreachable: Get "https://10.96.0.1:443/version":` should not be part of the message.

```
  - lastTransitionTime: "2024-09-16T15:20:44Z"
    message: 'Kubernetes cluster unreachable: Get "https://10.96.0.1:443/version":
      serviceaccounts "nginx" not found'
    observedGeneration: 2
    reason: InstallationFailed
    status: "False"
    type: Installed
```

rashmi43:
latest error message:
```
status:
  conditions:
  - lastTransitionTime: "2024-11-19T13:10:46Z"
    message: 'Kubernetes cluster unreachable: Get "[https://10.96.0.1:443/version](https://10.96.0.1/version)":  Unable
      to authenticate with Kubernetes cluster using ServiceAccount "nginx-install":
      ServiceAccount "nginx-install" not found.'
    observedGeneration: 1

**Comment by rashmi43** [comment]
able to fix this issue:
```
  conditions:
  - lastTransitionTime: "2024-11-19T14:45:08Z"
    message: ' Unable to authenticate with Kubernetes cluster using ServiceAccount
      "nginx-install": ServiceAccount "nginx-install" not found.'
    observedGeneration: 1
    reason: Failed
    status: Unknown
    type: Installed
  - lastTransitionTime: "2024-11-19T14:45:08Z"
    message: retrying to get installed bundle
    observedGeneration: 1
    reason: Retrying
    status: "True"
    type: Progressing

PR #1477, opened 2024-11-18
🐛  wrap service account not found error

rashmi43:
# Description

## Reviewer Checklist

- [ ] API Go Documentation


rashmi43:
fixes https://github.com/operator-framework/operator-controller/issues/1297

thetechnick:
While PR creates a predictable error string, the message is still pretty much the same and does not yet fix the issue:
```
  - lastTransitionTime: "2024-11-18T14:41:05Z"
    message: 'Kubernetes cluster unreachable: Get "https://10.96.0.1:443/version":
      service account not found'
    observedGeneration: 1
    reason: Failed
    status: Unknown
    type: Installed
```

@rashmi43 Would you please create your own error type[1] and catch this error before it is written to the Condition message to provide your own custom error message?
[1]: https://go.dev/tour/methods/19

everettraven:
Alternatively, with the custom error type we could attempt to parse the error string/type wherever it is returned and attempt to strip out what we think is the unnecessary prefix.

camilamacedo86 (changes requested):
HI @rashmi43 

Could you please rebase the commits?
Otherwise, when it gets merged, and we run `git log`, it will result in a very lousy commit message.

thetechnick:
@everettraven By creating a custom error into the TokenGetter we can keep the error message short and precise.
I don't see a reason to expose the whole error wrap-chain to the user, when we know exactly that the Service Account is missing in this case and the additional information is not useful or actually missleading.

To pick up your example, I would be very happy if the condition message just reads:
`Unable to authenticate with Kubernetes cluster using ServiceAccount "foo": ServiceAccount "foo" not found.`

rashmi43:
can we get someone to review @camilamacedo86 perhaps @thetechnick can review my changes. we can hold the merge till v1 is shipped

internal/authentication/tokengetter.go
```diff
@@ -86,7 +95,8 @@ func (t *TokenGetter) getToken(ctx context.Context, key types.NamespacedName) (*
 			Spec: authenticationv1.TokenRequestSpec{ExpirationSeconds: ptr.To(int64(t.expirationDuration / time.Second))},
 		}, metav1.CreateOptions{})
 	if err != nil {
-		return nil, err
+		saErr := &SANotFoundError{key.Name}
```
everettraven, line 98: Is it guaranteed that the only kind of error that is returned here is that the ServiceAccount we are requesting a token for is not found?

internal/authentication/tokengetter.go
```diff
@@ -19,6 +20,14 @@ type TokenGetter struct {
 	mu                 sync.RWMutex
 }
 
+type SANotFoundError struct {
```
everettraven, line 23:
```suggestion
type ServiceAccountNotFoundError struct {
```

internal/controllers/clusterextension_controller.go
```diff
@@ -206,6 +207,11 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1.Cl
 	installedBundle, err := r.InstalledBundleGetter.GetInstalledBundle(ctx, ext)
 	if err != nil {
 		setInstallStatus(ext, nil)
+		var saerr *tokengetter.SANotFoundError
+		if errors.As(err, &saerr) {
+			l.Info("is a SAError")
+			err = saerr
+		}
```
everettraven, line 214: I think there are a lot of other places where this error could be returned, likely anywhere we are doing interactions with the cluster since we are using the provided ServiceAccount for those interactions. Off the top of my head, that would be:
- Getting the installed bundle (here)
- Applying contents to the cluster 
- Setting up caches and informers for managed content

Is there a way we could more effectively capture this error type when setting the status conditions we are wanting a more specific message for?

internal/authentication/tokengetter.go
rashmi43: @thetechnick is this a fair assumption, like discussed above since we are doing a GET for a serviceaccount that would be the case.

internal/controllers/clusterextension_controller.go
everettraven: I understand how this is working in this particular error check, but I think there are other error checks in which it is possible for this error type to be returned as part of the wrapped error chain. Some off the top of my head is:
- When we attempt to fetch the currently installed bundle (which you already cover with this check here)
- When we attempt to apply content to the cluster (not covered by your changes)
- When we attempt to create informers to manage the content owned by a ClusterExtension (not covered by your changes)

The second piece of my comment was an attempt to get at the fact that instead of repeating this across all the error scenarios, and since we really only care about the message that is set in some of the status conditions when receiving this error, we could probably put this error check in the logic that is used to set the conditions we want to "customize" the errors for.

internal/authentication/tokengetter.go
```diff
@@ -86,7 +95,8 @@ func (t *TokenGetter) getToken(ctx context.Context, key types.NamespacedName) (*
 			Spec: authenticationv1.TokenRequestSpec{ExpirationSeconds: ptr.To(int64(t.expirationDuration / time.Second))},
 		}, metav1.CreateOptions{})
 	if err != nil {
-		return nil, err
+		saErr := &ServiceAccountNotFoundError{key.Name}
+		return nil, saErr
```
everettraven, line 99: I think we need to:
- Wrap this in a NotFound error check
- If the returned error is not a NotFound error, return the original error

This makes it so we are only returning our custom ServiceAccountNotFoundError whenever we know it is a NotFound error returned from our interaction with the Kubernetes API server. Not doing this means we assume that an error in this case always means the ServiceAccount is not found, which I think is a bad assumption to be making and will result in obfuscating any other errors that may be returned during this interaction.

internal/authentication/tokengetter.go
everettraven: I left another comment here with my thoughts on this: https://github.com/operator-framework/operator-controller/pull/1477#discussion_r1850476014

internal/controllers/clusterextension_controller.go
rashmi43: will work on 
1) unit testcases
2) encompassing more scenarios:
```
When we attempt to apply content to the cluster (not covered by your changes)
When we attempt to create informers to manage the content owned by a ClusterExtension (not covered by your changes)

File: internal/authentication/tokengetter.go (hunk 4 above)
Comment (rashmi43) [reply internal/authentication/tokengetter.go:99]: I rebuild and it seems to be working as expected:
```
status:
  conditions:
  - lastTransitionTime: "2024-11-26T16:08:21Z"
    message: 'Unable to authenticate with Kubernetes cluster using ServiceAccount
      "nginx-install": ServiceAccount "nginx-install" not found.'
    observedGeneration: 1
    reason: Failed
    status: Unknown
    type: Installed
  - lastTransitionTime: "2024-11-26T16:08:21Z"
    message: retrying to get installed bundle
    observedGeneration: 1
    reason: Retrying
    status: "True"
    type: Progressing

internal/authentication/tokengetter.go
```diff
@@ -95,8 +96,12 @@ func (t *TokenGetter) getToken(ctx context.Context, key types.NamespacedName) (*
...
 		}, metav1.CreateOptions{})
 	if err != nil {
-		saErr := &ServiceAccountNotFoundError{key.Name}
-		return nil, saErr
+		if errors.IsNotFound(err) {
+			saErr := &ServiceAccountNotFoundError{key.Name}
+			return nil, saErr
+		} else {
+			return nil, err
+		}
```
everettraven, line 104: Nit:
```suggestion
		if errors.IsNotFound(err) {
			saErr := &ServiceAccountNotFoundError{key.Name}
			return nil, saErr
		} 
		return nil, err
```

internal/authentication/tokengetter_test.go
```diff
@@ -71,7 +71,7 @@ func TestTokenGetterGet(t *testing.T) {
 		{"Testing token that expired 10 seconds ago", "test-service-account-3",
 			"test-namespace-3", "test-token-3", "failed to get token"},
 		{"Testing error when getting token from fake client", "test-service-account-4",
-			"test-namespace-4", "error when fetching token", "error when fetching token"},
+			"test-namespace-4", "error when fetching token.", "error when fetching token"},
```
everettraven, line 74: Why is this change made?

rashmi43: reverted it

internal/controllers/clusterextension_controller.go
```diff
@@ -206,6 +207,11 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1.Cl
 	installedBundle, err := r.InstalledBundleGetter.GetInstalledBundle(ctx, ext)
 	if err != nil {
 		setInstallStatus(ext, nil)
+		var saerr *tokengetter.ServiceAccountNotFoundError
+		if errors.As(err, &saerr) {
+			l.Info("is a SAError")
```
camilamacedo86, line 212: Could we improve the wording and explanation here? What exactly is the error, and under what circumstances does it occur?

As a cluster admin or extension admin, I want to clearly understand why and when an expected error happens so that I can handle the scenario more effectively.

camilamacedo86:
Hi @rashmi43,

Could you please squash the commits? Since this is a small change, it should ideally have just one commit.

I’m concerned that if we leave it as is, all these commits will show up in the logs (e.g., run git log to see what I mean). A clean and meaningful git history is important because, when debugging or identifying breaking changes, it’s common to rely on the history to pinpoint the cause. If the history isn’t well-organized, it can make life much harder. 😊

internal/controllers/clusterextension_controller.go
```diff
@@ -47,6 +47,7 @@ import (
 	"github.com/operator-framework/api/pkg/operators/v1alpha1"
 	catalogd "github.com/operator-framework/catalogd/api/v1"
 	helmclient "github.com/operator-framework/helm-operator-plugins/pkg/client"
+	tokengetter "github.com/operator-framework/operator-controller/internal/authentication"
```
camilamacedo86, line 50:
```suggestion
	"github.com/operator-framework/operator-controller/internal/authentication"
```

I think we don't need the alias here right?
Can we keep the original import?

internal/authentication/tokengetter.go
```diff
@@ -19,6 +21,14 @@ type TokenGetter struct {
 	mu                 sync.RWMutex
 }
 
+type ServiceAccountNotFoundError struct {
+	ServiceAccountName string
+}
+
+func (e *ServiceAccountNotFoundError) Error() string {
+	return fmt.Sprintf("Unable to authenticate with Kubernetes cluster using ServiceAccount \"%s\": ServiceAccount \"%s\" not found.", e.ServiceAccountName, e.ServiceAccountName)
+}
```
camilamacedo86, line 30:
```suggestion
// ServiceAccountNotFoundError represents an error when a specified ServiceAccount is not found.
type ServiceAccountNotFoundError struct {
	ServiceAccountName string // The name of the missing ServiceAccount.
}

// Error implements the error interface for ServiceAccountNotFoundError.
func (e *ServiceAccountNotFoundError) Error() string {
	return fmt.Sprintf("ServiceAccount \"%s\" not found: Unable to authenticate with the Kubernetes cluster.", e.ServiceAccountName)
}
```

internal/authentication/tokengetter.go
```diff
@@ -86,6 +96,10 @@ func (t *TokenGetter) getToken(ctx context.Context, key types.NamespacedName) (*
 			Spec: authenticationv1.TokenRequestSpec{ExpirationSeconds: ptr.To(int64(t.expirationDuration / time.Second))},
 		}, metav1.CreateOptions{})
 	if err != nil {
+		if errors.IsNotFound(err) {
+			saErr := &ServiceAccountNotFoundError{key.Name}
+			return nil, saErr
+		}
```
camilamacedo86, line 102:
```suggestion
		if errors.IsNotFound(err) {
			return nil, &ServiceAccountNotFoundError{ServiceAccountName: key.Name}
		}
```

Define the attribute name explicitly, which the value represents. This improves the readability of the code
We can also return directly

internal/controllers/clusterextension_controller.go
```diff
@@ -206,6 +207,11 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1.Cl
 	installedBundle, err := r.InstalledBundleGetter.GetInstalledBundle(ctx, ext)
 	if err != nil {
 		setInstallStatus(ext, nil)
+		var saerr *tokengetter.ServiceAccountNotFoundError
+		if errors.As(err, &saerr) {
+			l.Info("is a SAError")
+			err = saerr
+		}
```
camilamacedo86, line 214: Why do we need to log something here?
Also, we are checking the error and replacing it 
Why do we need it?

internal/controllers/clusterextension_controller_test.go
```diff
@@ -677,6 +677,97 @@ func TestClusterExtensionInstallationSucceeds(t *testing.T) {
...
+	require.Equal(t, metav1.ConditionTrue, installedCond.Status)
+	require.Equal(t, ocv1.ReasonSucceeded, installedCond.Reason)
+
+	t.Log("By checking the expected progressing conditions")
+	progressingCond := apimeta.FindStatusCondition(clusterExtension.Status.Conditions, ocv1.TypeProgressing)
+	t.Log("Progressing condition message", progressingCond.Message)
+	require.NotNil(t, progressingCond)
+	t.Log("Progressing condition status", progressingCond.Status)
+	t.Log("Progressing condition reason", progressingCond.Reason)
+	//require.Equal(t, metav1.ConditionTrue, progressingCond.Status)
```
camilamacedo86, line 762: Why is this line commented on? 
Do we need to do this assert? 
We need uncomment and fix or remove, IHMO we cannot leave commented lines

internal/controllers/clusterextension_controller_test.go
```diff
@@ -677,6 +677,97 @@ func TestClusterExtensionInstallationSucceeds(t *testing.T) {
 	require.NoError(t, cl.DeleteAllOf(ctx, &ocv1.ClusterExtension{}))
 }
 
+func TestClusterExtensionInstallationFailsWithNoServiceAccount(t *testing.T) {
```
camilamacedo86, line 680: Hi @rashmi43,

Without diving into the code itself, I think we should improve how t.Log is being used. The messages should describe actions like "Validating the status to ensure it is X" or "Checking the progressing condition." This makes the logs more meaningful and helps anyone reading them understand what’s happening in the test.

Also, we can make things cleaner by using defer for resource cleanup. I would like to suggest something like:

```go
func TestClusterExtensionInstallationFailsWithNoServiceAccount(t *testing.T) {

	reconciler.Unpacker = &MockUnpacker{
		result: &source.Result{
			State:  source.StateUnpacked,
			Bundle: fstest.MapFS{},
		},
	}

	ctx := context.Background()
	extKey := types.NamespacedName{Name: fmt.Sprintf("cluster-extension-test-%s", rand.String(8))}
	defer func() {
		t.Log("Cleaning up all cluster extension resources")
		cl.DeleteAllOf(ctx, &ocv1.ClusterExtension{})
	}()

	t.Log("Initializing cluster extension with a channel, version, and non-existent service account")
	pkgName := "prometheus"
	pkgVer := "1.0.0"
	pkgChan := "beta"
	namespace := fmt.Sprintf("test-ns-%s", rand.String(8))
	serviceAccount := fmt.Sprintf("test-does-not-exist-%s", rand.String(8))

	clusterExtension := &ocv1.ClusterExtension{
		ObjectMeta: metav1.ObjectMeta{Name: extKey.Name},
		Spec: ocv1.ClusterExtensionSpec{
			Source: ocv1.SourceConfig{
				SourceType: "Catalog",
				Catalog: &ocv1.CatalogSource{
					PackageName: pkgName,
					Version:     pkgVer,
					Channels:    []string{pkgChan},
				},
			},
			Namespace: namespace,
			ServiceAccount: ocv1.ServiceAccountReference{
				Name: serviceAccount,
			},
		},
	}
	require.NoError(t, cl.Create(ctx, clusterExtension), "Failed to create the cluster extension resource")

	t.Log("Setting up mock resolver and applier for reconciliation")
	reconciler.Resolver = resolve.Func(func(_ context.Context, _ *ocv1.ClusterExtension, _ *ocv1.BundleMetadata) (*declcfg.Bundle, *bsemver.Version, *declcfg.Deprecation, error) {
		v := bsemver.MustParse("1.0.0")
		return &declcfg.Bundle{
			Name:    "prometheus.v1.0.0",
			Package: "prometheus",
			Image:   "quay.io/operatorhubio/prometheus@fake1.0.0",
		}, &v, nil, nil
	})
	reconciler.Applier = &MockApplier{
		objs: []client.Object{},
	}
	reconciler.Manager = &MockManagedContentCacheManager{
		cache: &MockManagedContentCache{},
	}

	t.Log("Running reconcile to process the cluster extension")
	require.Equal(t, ctrl.Result{}, res, "Unexpected reconcile result")
	require.NoError(t, err, "Reconcile failed with an error")

	t.Log("Fetching updated cluster extension after reconciliation")
	require.NoError(t, cl.Get(ctx, extKey, clusterExtension), "Failed to retrieve the updated cluster extension resource")

	t.Log("Verifying the status fields of the cluster extension")
	require.Equal(t, ocv1.BundleMetadata{Name: "prometheus.v1.0.0", Version: "1.0.0"}, clusterExtension.Status.Install.Bundle, "Bundle metadata does not match expected values")

	t.Log("Verifying the installed condition of the cluster extension")
	installedCond := apimeta.FindStatusCondition(clusterExtension.Status.Conditions, ocv1.TypeInstalled)
	require.NotNil(t, installedCond, "Installed condition not found")
	require.Equal(t, metav1.ConditionTrue, installedCond.Status, "Installed condition status should be True")
	require.Equal(t, ocv1.ReasonSucceeded, installedCond.Reason, "Installed condition reason should indicate success")

	t.Log("Verifying the progressing condition of the cluster extension")
	progressingCond := apimeta.FindStatusCondition(clusterExtension.Status.Conditions, ocv1.TypeProgressing)
	require.NotNil(t, progressingCond, "Progressing condition not found")
	require.Equal(t, ocv1.ReasonFailed, progressingCond.Reason, "Progressing condition reason should indicate failure")
}
```

internal/authentication/tokengetter.go
camilamacedo86: Hey! Just a quick note:

Anything we write with a capital letter at the front, like a function or struct name, is exported—meaning it can be accessed by other packages. That also means we should add proper documentation for those with // comments on top. This makes the code easier to understand for anyone using it  😊

internal/controllers/clusterextension_controller.go
rashmi43: we are checking the type of the error and then returning our error message

internal/controllers/clusterextension_controller.go
rashmi43: fixed

internal/authentication/tokengetter.go
rashmi43: sure

internal/controllers/clusterextension_controller_test.go
rashmi43: going to stash these changes

internal/authentication/tokengetter.go
```diff
@@ -19,6 +21,15 @@ type TokenGetter struct {
 	mu                 sync.RWMutex
 }
 
+type ServiceAccountNotFoundError struct {
+	ServiceAccountName string // The name of the missing ServiceAccount.
+}
+
+// Error implements the error interface for ServiceAccountNotFoundError.
+func (e *ServiceAccountNotFoundError) Error() string {
+	return fmt.Sprintf("ServiceAccount \"%s\" not found: Unable to authenticate with the Kubernetes cluster.", e.ServiceAccountName)
```
joelanford, line 30: Some style suggestions to make this error message follow standard conventions:
```suggestion
	return fmt.Sprintf("service account %q not found: unable to authenticate with the Kubernetes cluster", e.ServiceAccountName)
```

internal/authentication/tokengetter.go
```diff
@@ -19,6 +21,15 @@ type TokenGetter struct {
 	mu                 sync.RWMutex
 }
 
+type ServiceAccountNotFoundError struct {
+	ServiceAccountName string // The name of the missing ServiceAccount.
+}
```
joelanford, line 26: Maybe not strictly necessary, but it may prove useful to:
1. include the underlying error value in the struct
2. implement the Unwrap function for this struct so that the underlying error can be recovered if necessary.

I recognize that we don't want to show up in the error string, so I'm _not_ suggesting that we use `fmt.Errorf("blah blah: %w", err)`

But we can do this:
```go
type ServiceAccountNotFoundError struct {
	ServiceAccountName string // The name of the missing ServiceAccount.
	Err error // The underlying error
 }

func (e *ServiceAccountNotFoundError) Unwrap() error {
	return e.Err
}
```

internal/authentication/tokengetter.go
joelanford: Also, should we include the expected namespace in the error message as well?

internal/controllers/clusterextension_controller.go
```diff
@@ -206,6 +207,11 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1.Cl
 	installedBundle, err := r.InstalledBundleGetter.GetInstalledBundle(ctx, ext)
 	if err != nil {
 		setInstallStatus(ext, nil)
+		var saerr *authentication.ServiceAccountNotFoundError
+		if errors.As(err, &saerr) {
+			l.Info("ServiceAccount defined in the ClusterExtension is not found")
+			err = saerr
+		}
```
joelanford, line 214: By returning this error in line 217, aren't we getting logging by controller-runtime already? If so, this seems extraneous.

rashmi43: removing the log

internal/controllers/clusterextension_controller_test.go
rashmi43: I need some help with this test case

internal/authentication/tokengetter.go
rashmi43: added the same

internal/controllers/clusterextension_controller.go
```diff
@@ -206,6 +207,10 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1.Cl
 	installedBundle, err := r.InstalledBundleGetter.GetInstalledBundle(ctx, ext)
 	if err != nil {
 		setInstallStatus(ext, nil)
+		var saerr *authentication.ServiceAccountNotFoundError
+		if errors.As(err, &saerr) {
+			err = saerr
```
camilamacedo86, line 212:
```suggestion
			setInstalledStatusConditionUnknown(ext, saerr.Error())
                         setStatusProgressing(ext, errors.New("installation cannot proceed due to missing ServiceAccount"))
                         return ctrl.Result{}, err
```

It seems misslanding.
We should provide the specific info for the status condition ( the new error ) 
and return directly.

internal/controllers/clusterextension_controller_test.go
```diff
@@ -677,6 +677,89 @@ func TestClusterExtensionInstallationSucceeds(t *testing.T) {
...
+	require.Equal(t, ctrl.Result{}, res)
+	require.NoError(t, err)
+
+	t.Log("By fetching updated cluster extension after reconcile")
+	require.NoError(t, cl.Get(ctx, extKey, clusterExtension))
+
+	t.Log("By checking the status fields")
+	require.Equal(t, ocv1.BundleMetadata{Name: "prometheus.v1.0.0", Version: "1.0.0"}, clusterExtension.Status.Install.Bundle)
+	res, err = reconciler.Reconcile(ctx, ctrl.Request{NamespacedName: extKey})
+	require.Error(t, err, res)
```
camilamacedo86, line 746:
```suggestion
	require.Error(t, err)
        require.Contains(t, err.Error(), "service account", "Expected error about missing ServiceAccount but got: %v", err)
```

I think we need to something like here.

rashmi43: added the changes

internal/controllers/clusterextension_controller_test.go
```diff
@@ -677,6 +677,79 @@ func TestClusterExtensionInstallationSucceeds(t *testing.T) {
...
+	}
+	reconciler.Manager = &MockManagedContentCacheManager{
+		cache: &MockManagedContentCache{},
+	}
+
+	_, err = reconciler.Reconcile(ctx, ctrl.Request{NamespacedName: extKey})
+	require.Error(t, err)
+
+	t.Log("By fetching updated cluster extension after reconcile")
+	require.NoError(t, cl.Get(ctx, extKey, clusterExtension))
```
camilamacedo86, line 747: Hi @rashmi43 

Your changes are in the reconcile.
We can confirm that the reconciliation code (your changes) is not triggered between the previous commit and this one. Note that we are mocking its result here. 

So, to validate your code changes now, the best way is to use e2e tests, which will do everything and trigger your code. See: https://github.com/operator-framework/operator-controller/blob/main/test/e2e/cluster_extension_install_test.go


You would need to add something like:

```go
func TestClusterExtensionInstallFailsWithoutServiceAccount(t *testing.T) {
	t.Log("When a cluster extension is installed from a catalog")
	t.Log("And the ServiceAccount does not exist")

	// Step 1: Setup test environment without creating an SA
	clusterExtensionName := fmt.Sprintf("clusterextension-%s", rand.String(8))

	ns, err := createNamespace(context.Background(), clusterExtensionName)
	require.NoError(t, err)

	extensionCatalog, err := createTestCatalog(context.Background(), testCatalogName, os.Getenv(testCatalogRefEnvVar))
	require.NoError(t, err)

	// Important: We do NOT create a ServiceAccount here
	defer testCleanup(t, extensionCatalog, nil, nil, ns)
	defer utils.CollectTestArtifacts(t, artifactName, c, cfg)

	// Step 2: Define the ClusterExtension with a non-existent SA
	clusterExtension := &ocv1.ClusterExtension{
			Name: clusterExtensionName,
		},
		Spec: ocv1.ClusterExtensionSpec{
			Source: ocv1.SourceConfig{
				SourceType: "Catalog",
				Catalog: &ocv1.CatalogSource{
					PackageName: "test",
					Selector: &metav1.LabelSelector{
						MatchLabels: map[string]string{"olm.operatorframework.io/metadata.name": extensionCatalog.Name},
					},
				},
			},
			Namespace: ns.Name,
			ServiceAccount: ocv1.ServiceAccountReference{
				Name: "non-existent-sa", // This SA does NOT exist
			},
		},
	}

	t.Log("By creating the ClusterExtension resource without a ServiceAccount")
	require.NoError(t, c.Create(context.Background(), clusterExtension))

	// Step 3: Validate failure due to missing ServiceAccount
	t.Log("By eventually reporting an installation failure due to missing ServiceAccount")
		assert.NoError(ct, c.Get(context.Background(), types.NamespacedName{Name: clusterExtension.Name}, clusterExtension))

		// Validate that Progressing reports failure
		cond := apimeta.FindStatusCondition(clusterExtension.Status.Conditions, ocv1.TypeProgressing)
		if assert.NotNil(ct, cond) {
			assert.Equal(ct, ocv1.ReasonRetrying, cond.Reason)
			assert.Contains(ct, cond.Message, "installation cannot proceed due to missing ServiceAccount")
		}

		// Validate that Installed condition is false
		installCond := apimeta.FindStatusCondition(clusterExtension.Status.Conditions, ocv1.TypeInstalled)
		if assert.NotNil(ct, installCond) {
			assert.Equal(ct, metav1.ConditionFalse, installCond.Status)
			assert.Equal(ct, ocv1.ReasonFailed, installCond.Reason)
			assert.Contains(ct, installCond.Message, "No bundle installed")
		}
}
```

rashmi43: added the e2e test case

test/e2e/cluster_extension_install_test.go
```diff
@@ -251,21 +251,30 @@ func testCleanup(t *testing.T, cat *catalogd.ClusterCatalog, clusterExtension *o
 		return errors.IsNotFound(err)
 	}, pollDuration, pollInterval)
```
camilamacedo86, line 252: Hi @rashmi43 

I added a comment in the slack:

Apologies for any confusion—I should have verified it locally more carefully instead of relying solely on what was presented in the PR and what you were initially trying to accomplish.

We are only adding a new message here, so to incorporate the tests, we just need to improve the existing ones to also check for this message.
There’s no need to write the tests the way they were done before or use end-to-end (e2e) tests.

A unit test for GET(token) will be sufficient since it directly calls your changes at the line  token, err = t.getToken(ctx, key):https://github.com/operator-framework/operator-controller/blob/main/internal/authentication/tokengetter.go#L67
See that I get your changes, and added supplement the test to check your change here (all is working):
[GitHub PR Comparison for Help](https://github.com/operator-framework/operator-controller/compare/main...camilamacedo86:operator-controller:pr-help-rashmi43-sa-err-msg?expand=1)
Let me know if you need any clarification! (

rashmi43: removed the e2e and added the unit test

test/e2e/cluster_extension_install_test.go
```diff
@@ -251,21 +251,30 @@ func testCleanup(t *testing.T, cat *catalogd.ClusterCatalog, clusterExtension *o
...
+	if clusterExtension != nil {
+		t.Logf("By deleting ClusterExtension %q", clusterExtension.Name)
+		require.NoError(t, c.Delete(context.Background(), clusterExtension))
+		require.Eventually(t, func() bool {
+			err := c.Get(context.Background(), types.NamespacedName{Name: clusterExtension.Name}, &ocv1.ClusterExtension{})
+			return errors.IsNotFound(err)
+		}, pollDuration, pollInterval)
+	} else {
+		t.Log(" ClusterExtension is nil", clusterExtension)
+	}
```
rashmi43, line 265: I ll remove these changes as well?

rashmi43:
Closing this PR in favour of https://github.com/operator-framework/operator-controller/pull/1698

PR #1499, opened 2024-11-26
:seedling: improve the verify-crd-compatibility make target

everettraven:
for use during local development by making the default updated source be the local file, picking up changes not yet committed. The previous git sourcing required changes to be committed to be picked up and is not a very good developer experience as you should validate the changes prior to commiting them. Since the git source is necessary for CI, the GitHub Action is updated to specify the git source for the updated file.

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -132,10 +132,10 @@ bingo-upgrade: $(BINGO) #EXHELP Upgrade tools
 
 .PHONY: verify-crd-compatibility
 CRD_DIFF_ORIGINAL_REF := main
-CRD_DIFF_UPDATED_REF  := HEAD
+CRD_DIFF_UPDATED_SOURCE := file://config/base/crd/bases/olm.operatorframework.io_clusterextensions.yaml
 CRD_DIFF_CONFIG := crd-diff-config.yaml
-verify-crd-compatibility: $(CRD_DIFF)
-	$(CRD_DIFF) --config="${CRD_DIFF_CONFIG}" "git://${CRD_DIFF_ORIGINAL_REF}?path=config/base/crd/bases/olm.operatorframework.io_clusterextensions.yaml" "git://${CRD_DIFF_UPDATED_REF}?path=config/base/crd/bases/olm.operatorframework.io_clusterextensions.yaml"
+verify-crd-compatibility: $(CRD_DIFF) manifests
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -132,10 +132,10 @@ bingo-upgrade: $(BINGO) #EXHELP Upgrade tools
 
 .PHONY: verify-crd-compatibility
 CRD_DIFF_ORIGINAL_REF := main
-CRD_DIFF_UPDATED_REF  := HEAD
+CRD_DIFF_UPDATED_SOURCE := file://config/base/crd/bases/olm.operatorframework.io_clusterextensions.yaml
```
tmshort, line 135: I';m assuming this isn't treated as an absolute path?

Makefile
```diff
@@ -132,10 +132,10 @@ bingo-upgrade: $(BINGO) #EXHELP Upgrade tools
...
 CRD_DIFF_CONFIG := crd-diff-config.yaml
-verify-crd-compatibility: $(CRD_DIFF)
-	$(CRD_DIFF) --config="${CRD_DIFF_CONFIG}" "git://${CRD_DIFF_ORIGINAL_REF}?path=config/base/crd/bases/olm.operatorframework.io_clusterextensions.yaml" "git://${CRD_DIFF_UPDATED_REF}?path=config/base/crd/bases/olm.operatorframework.io_clusterextensions.yaml"
+verify-crd-compatibility: $(CRD_DIFF) manifests
```
tmshort, line 137: Adding `manifests` to this target will cause it to be run in the crd-diff, and if there's a diff, it won't be caught by the GitHub workflow, because it's looking directly at SHAs.
That being said, this should fail `make verify`

tmshort (approved):
Two minor comments/questions.

tmshort:
⛔ This should be held, unless we consider it a bug.

Makefile
everettraven: Correct. In CI, the expectation is that it is looking at the committed changes. I would push for us to rely on the distinct checks for "uncommitted, generated, changes" in CI.

Makefile
everettraven: Nope. It will read it as a relative path.

perdasilva:
unholding as v1 has been released :tada:

PR #1502, opened 2024-11-26
📖 add clarification about if is possible to migrate from V0 to V1

camilamacedo86:
In order to try to be more transparent about our goals and current state.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -37,3 +37,12 @@ For more details on the evolution of OLM and the roadmap for v1, explore the fol
 
 - [Multi-Tenancy Challenges, Lessons Learned, and Design Shifts](project/olmv1_design_decisions.md)
 - [OLM v1 Roadmap](project/olmv1_roadmap.md)
+
+## Can I Migrate from OLMv0 to OLMv1?
+
+There is currently no concrete migration strategy due to the [conceptual differences between OLMv0 and OLMv1](project/olmv1_design_decisions.md).
+OLMv1, as of writing, supports a subset of the existing content supported by OLMv0.
+For more information regarding the current limitations of OLMv1, see [limitations](project/olmv1_limitations.md).
+
+If your current usage of OLMv0 is compatible with the limitations and expectations of OLMv1, you may be able to manually
+transition to using OLMv1 following the standard workflows we have documented.


camilamacedo86:
@joelanford @LalatenduMohanty 
I was thinking of something like that so that we can make it clear.
WDYT?

docs/index.md
```diff
@@ -37,3 +37,14 @@ For more details on the evolution of OLM and the roadmap for v1, explore the fol
...
+## Can I Migrate from OLMv0 to OLMv1?
+
+Not yet. Currently, OLMv1 is a **Minimum Viable Product (MVP)** with a subset of OLMv0’s functionality.
+While it makes a best-effort attempt to remain compatible with and manage Bundles in the Registry+v1 format (using CSV, as supported in OLMv0),
+it comes with certain [limitations](project/olmv1_limitations.md). Additionally, users may encounter edge cases that 
+might have not been fully mapped due to the [conceptual differences between OLMv0 and OLMv1](project/olmv1_design_decisions.md).
+
+OLMv1 is actively evolving towards adopting more common package formats widely used by the community, 
+such as **HelmCharts**. Until OLMv1 achieves key milestones on its [roadmap—such](project/olmv1_roadmap.md) as support for **Webhooks**, 
+**HelmCharts**, and other critical features—we strongly recommend that users **do not migrate from OLMv0 to OLMv1** at this time.
```
everettraven, line 50: Strongly recommended that users don't migrate from OLMv0 to OLMv1 feels like a stretch to me. What if someone if using OLMv0 in a way that is totally compatible with OLMv1? Should we provide them details on how to go from using OLMv0 to OLMv1 in that case?

The migration story right now feels much more nuanced than I think that this messaging conveys.

camilamacedo86: > Should we provide them details on how to go from using OLMv0 to OLMv1 in that case?

We do not have the migration steps yet. As we discussed yesterday in the community meeting, we want to address more aspects before starting to work on the migration guidance to ensure it is done properly. (I.,e We want to support HelmChart and etc)

The idea here is to make this clear for both us and the community. **Better wording suggestions are very welcome.** However, at this moment, I understand that we convey we strongly recommend against using OLMv1 in production or starting any migration.

everettraven: I haven't spent any time thinking of better wording yet, but will plan to make some suggestions later today.

>However, at this moment, I understand that we convey we strongly recommend against using OLMv1 in production or starting any migration

In my opinion, this is a bad precedent to set. It sounds like we have prematurely released v1.0.0 if we say that it isn't recommended to use in production. I think we _do_ recommend using it in production for the use cases and workflows that are currently enabled by the features we have in place. We can _separately_ state that we do not currently have any migration strategies in place for going from OLMv0 --> OLMv1.

docs/index.md
```diff
@@ -37,3 +37,14 @@ For more details on the evolution of OLM and the roadmap for v1, explore the fol
 
 - [Multi-Tenancy Challenges, Lessons Learned, and Design Shifts](project/olmv1_design_decisions.md)
 - [OLM v1 Roadmap](project/olmv1_roadmap.md)
+
+## Can I Migrate from OLMv0 to OLMv1?
```
everettraven, line 41: Should this be an entirely separate entry in the documentation that we can update in the future when we _do_ have a migration strategy?

docs/index.md
camilamacedo86: Ok. Think about how we can convey the way in a way that is nice and transparent 
Open to suggestions !!!

docs/index.md
camilamacedo86: > you may be able to manually transition to using OLMv1 following the standard workflows we have documented.

We did not document that, right?

docs/index.md
```diff
@@ -37,3 +37,14 @@ For more details on the evolution of OLM and the roadmap for v1, explore the fol
...
+
+Not yet. Currently, OLMv1 is a **Minimum Viable Product (MVP)** with a subset of OLMv0’s functionality.
+While it makes a best-effort attempt to remain compatible with and manage Bundles in the Registry+v1 format (using CSV, as supported in OLMv0),
+it comes with certain [limitations](project/olmv1_limitations.md). Additionally, users may encounter edge cases that 
+might have not been fully mapped due to the [conceptual differences between OLMv0 and OLMv1](project/olmv1_design_decisions.md).
+
+OLMv1 is actively evolving towards adopting more common package formats widely used by the community, 
+such as **HelmCharts**. Until OLMv1 achieves key milestones on its [roadmap—such](project/olmv1_roadmap.md) as support for **Webhooks**,
```
LalatenduMohanty, line 49: I would suggest that we remove the reference to project/olmv1_roadmap.md  as we will mark it deprecated soon.

docs/index.md
```diff
@@ -37,3 +37,14 @@ For more details on the evolution of OLM and the roadmap for v1, explore the fol
 
 - [Multi-Tenancy Challenges, Lessons Learned, and Design Shifts](project/olmv1_design_decisions.md)
 - [OLM v1 Roadmap](project/olmv1_roadmap.md)
+
+## Can I Migrate from OLMv0 to OLMv1?
+
+Not yet. Currently, OLMv1 is a **Minimum Viable Product (MVP)** with a subset of OLMv0’s functionality.
```
LalatenduMohanty, line 43: Here are some suggested text. 
```
yes, but with [limitations](project/olmv1_limitations.md). OLM v1 is actively working to address some of the limitations. However because OLMv1's design principles are different than OLMv0 it is hard for us to say if they will be fully compatible with each-other in future.  We will update this space with more information when they are available.

LalatenduMohanty: I do not think we need to provide more information here. IMO less is more here.

LalatenduMohanty (changes requested):

LalatenduMohanty:
@joelanford PTAL and let us know if you are fine with the language we have used.

camilamacedo86:
Hi @LalatenduMohanty 

It is with your suggestion now, so it seems good to fly 🚀

PR #1510, opened 2024-12-03
🌱 add Github action to capture logs and pod descriptions when e2e fails

camilamacedo86:
# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -21,6 +21,20 @@ jobs:
     - name: Run the extension developer e2e test
       run: make extension-developer-e2e
 
+    - name: Collect pod logs and descriptions on failure
+      if: failure()
+      run: |
+        echo "Fetching pod logs and descriptions..."
+        kubectl get pods -n olmv1-system -o wide || true
+        for pod in $(kubectl get pods -n olmv1-system -o name); do
+          echo "Collecting logs for $pod..."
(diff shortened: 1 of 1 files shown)


everettraven:
Just a note that our e2e action should already have logic to upload artifacts upon failure. For example, https://github.com/operator-framework/operator-controller/actions/runs/11746095340

The artifacts are gathered by https://github.com/operator-framework/operator-controller/blob/e51c0c205ec09e51691579cacbedc3ac53e4bd54/test/e2e/cluster_extension_install_test.go#L865-L991 on test failure

camilamacedo86:
Hi @everettraven 

> Just a note that our e2e action should already have logic to upload artifacts upon failure. For example, https://github.com/operator-framework/operator-controller/actions/runs/11746095340

I see that, but IHMO it is not so straightforward way helpful as looking the logs in the action. I do not see any harm to add it but if you folks see any reason for not please let me know.

joelanford:
IMO, we should not have two different ways of doing this. If we can improve on what's there, great! If we need to replace what we already have with an alternate method, let's make sure we pull in the folks that were involved with the original implementation to make sure there we account for everything that was originally considered. For example, was there some specific reason that we did _NOT_ use a log-based approach?

camilamacedo86:
Fair enough.
Closing this one as not accepted.

PR #1523, opened 2024-12-13
Adds a utility to make an "ownership tree" like an ls-tree for a k8s namespace

bentito:
Also allows to store a state "fingerprint" based on getting the YAML for each resource in the tree and prompting an LLM for what is wrong with the ClusterExtension install.

commands are:

`hack/tools/ownership_tree.py <namespace>`
prints an lstree style list of the resources found in the namespace

`hack/tools/ownership_tree.py <namespace> --no-events`
same as w/ no flags, but Event resources are not shown

`hack/tools/ownership_tree.py <namespace> --with-event-info`
same as w/ no flags, but Event resources messages are also shown

`hack/tools/ownership_tree.py <namespace> --gather-cluster-extension-state`
same as w/ no flags, but also a "fingerprint" is created from getting all the YAML for the each resource in the tree, file is JSON and will have the name: `<namespace>-state.json`

`hack/tools/ownership_tree.py <namespace> --prompt`
same as with `--gather-cluster-extension-state` flag set, but the "fingerprint" is created and sent as part of a prompt to an LLM for diagnosis of ClusterExtension problems.

### DIFF
--- a/hack/tools/ownership_tree.py
+++ b/hack/tools/ownership_tree.py
@@ -1,0 +1,153 @@
+#!/usr/bin/env python3
+import json
+import subprocess
+import sys
+import argparse
+from collections import defaultdict
+
+parser = argparse.ArgumentParser(description="Print a tree of ownership for all resources in a namespace, grouped by kind.")
+parser.add_argument("namespace", help="The namespace to inspect")
+parser.add_argument("--no-events", action="store_true", help="Do not show Events kind grouping")
+args = parser.parse_args()
+
+NAMESPACE = args.namespace
+SHOW_EVENTS = not args.no_events
+
+# Get all namespaced resource types
+api_resources_cmd = ["kubectl", "api-resources", "--verbs=list", "--namespaced", "-o", "name"]
+resource_types = subprocess.check_output(api_resources_cmd, text=True).strip().split('\n')
+
+uid_to_resource = {}
+all_uids = set()
+
+def get_resources_for_type(r_type):
+    try:
+        items_json = subprocess.check_output(
+            ["kubectl", "get", r_type, "-n", NAMESPACE, "-o", "json"],
+            text=True
+        )
+    except subprocess.CalledProcessError:
+        return []
+    data = json.loads(items_json)
+    if "items" not in data:
+        return []
+    return data["items"]
+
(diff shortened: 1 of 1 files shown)


camilamacedo86 (changes requested):
I do not mean to be contrary, but this raises some interesting questions:

1. Are we considering adding tooling options to the operator project itself, or are we planning to use something like an OLM tool to help users? If so, is this the right place to address that?
2. Why would we introduce tools in Python when the entire project is written in Go? If we decide to add tooling, shouldn’t it also be implemented in Go to stay consistent with the rest of the project? What would justify deviating from this approach?
3. Everything we add, we must support. Do we really want to take on the responsibility of supporting this tooling in the operator itself? 
4. Do we have an RFC or brief for this, or are you still in the exploration phase?

bentito:
Hi Camila,
Thanks for the comment, see inline below for specifics, but TL;DR; of my response: It's a Draft PR and just exploration.

> I do not mean to be contrary, but this raises some interesting questions:
> 
>     1. Are we considering adding tooling options to the operator project itself, or are we planning to use something like an OLM tool to help users? If so, is this the right place to address that?

Likely we would want something located here in hack/ or maybe off separately in a tooling repo if we end up creating one.

> 
>     2. Why would we introduce tools in Python when the entire project is written in Go? If we decide to add tooling, shouldn’t it also be implemented in Go to stay consistent with the rest of the project? What would justify deviating from this approach?

As noted in the live (and recorded) demo: This was done in Python as a PoC and one-off. Definitely there is no reason to do this in Python if we wanted to make a tool like this.

> 
>     3. Everything we add, we must support. Do we really want to take on the responsibility of supporting this tooling in the operator itself?

This is a draft PR. I think in general, no need to comment about where stuff is going to land or anything like for a draft PR.

> 
>     4. Do we have an RFC or brief for this, or are you still in the exploration phase?

This was a Hack and Hustle Day exploration. It's up in Draft PR form so that people can see the code, and get more insight beyone the demo.

joelanford:
How much overlap is there between this and https://github.com/ahmetb/kubectl-tree?

perdasilva:
closing as stale - please re-open if its still relevant

PR #1539, opened 2025-01-01
:bug: desired state in applier

azych:
<del>A few minor fixes:</del>
1. _internal/applier/helm.go:getReleaseState()_ contained two `if` blocks that handled  `driver.ErrReleaseNotFound` with the first one returning, which meant the second block would never run.
    
    Additionally, the first handler always returned a `nil`  `desiredState` causing the only currently used `Preflight` validator (`crdupgradesafety`) and potentially others in the future to exit immediately because of not having anything to validate.

    This fix leaves only the second handler block in, which actually triggers a dry-run installation and is able to return a proper `desiredState` that any `Preflight` validation can work with. It also has the added benefit of potentially spotting any issues and exiting earlier in case the dry-run failed.

<del>2. _hack/tools/catalogs/download-catalog_ script - changes the name of the catalogd service from `catalogd-server` to the currently used `catalogd-service`.</del>

<del>3. _hack/tools/catalogs/list-compatible-bundles_ script - adds the missing `extract-olm-package-property` function to the call chain and uses the correct name of regex filtering function (`filter-by-regex-if-necessary`), making sure the script works.
    The main issues here were two:
    1. `regex_it` did not exist and was renamed during review of the original PR to `filter-by-regex-if-necessary`
    2. `extract-olm-package-property` call was missing from the chain meaning the filter function couldn't recognize the structure it was expecting</del>

Fixes to `hack/tools/catalogs/` scripts have been separated from this PR into PRs of their own: https://github.com/operator-framework/operator-controller/pull/1546 and https://github.com/operator-framework/operator-controller/pull/1545 as per https://github.com/operator-framework/operator-controller/pull/1539#discussion_r1900596361 

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/applier/helm.go
+++ b/internal/applier/helm.go
@@ -153,13 +153,6 @@ func (h *Helm) Apply(ctx context.Context, contentFS fs.FS, ext *ocv1.ClusterExte
 
 func (h *Helm) getReleaseState(cl helmclient.ActionInterface, ext *ocv1.ClusterExtension, chrt *chart.Chart, values chartutil.Values, post postrender.PostRenderer) (*release.Release, *release.Release, string, error) {
 	currentRelease, err := cl.Get(ext.GetName())
-	if err != nil && !errors.Is(err, driver.ErrReleaseNotFound) {
-		return nil, nil, StateError, err
-	}
-	if errors.Is(err, driver.ErrReleaseNotFound) {
-		return nil, nil, StateNeedsInstall, nil
-	}
-
 	if errors.Is(err, driver.ErrReleaseNotFound) {
 		desiredRelease, err := cl.Install(ext.GetName(), ext.Spec.Namespace, chrt, values, func(i *action.Install) error {
 			i.DryRun = true
@@ -171,6 +164,10 @@ func (h *Helm) getReleaseState(cl helmclient.ActionInterface, ext *ocv1.ClusterE
 		}
 		return nil, desiredRelease, StateNeedsInstall, nil
 	}
+	if err != nil {
+		return nil, nil, StateError, err
+	}
+
 	desiredRelease, err := cl.Upgrade(ext.GetName(), ext.Spec.Namespace, chrt, values, func(upgrade *action.Upgrade) error {
 		upgrade.MaxHistory = maxHelmReleaseHistory
 		upgrade.DryRun = true
--- a/internal/applier/helm_test.go
+++ b/internal/applier/helm_test.go
@@ -1,0 +1,321 @@
+package applier_test
+
+import (
+	"context"
+	"errors"
+	"os"
+	"testing"
+	"testing/fstest"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"helm.sh/helm/v3/pkg/action"
+	"helm.sh/helm/v3/pkg/chart"
+	"helm.sh/helm/v3/pkg/release"
+	"helm.sh/helm/v3/pkg/storage/driver"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	helmclient "github.com/operator-framework/helm-operator-plugins/pkg/client"
+
+	v1 "github.com/operator-framework/operator-controller/api/v1"
+	"github.com/operator-framework/operator-controller/internal/applier"
+)
+
+type mockPreflight struct {
+	installErr error
+	upgradeErr error
+}
+
+func (mp *mockPreflight) Install(context.Context, *release.Release) error {
+	return mp.installErr
+}
+
+func (mp *mockPreflight) Upgrade(context.Context, *release.Release) error {
+	return mp.upgradeErr
+}
+
+type mockActionGetter struct {
+	actionClientForErr error
+	getClientErr       error
+	installErr         error
+	dryRunInstallErr   error
+	upgradeErr         error
+	dryRunUpgradeErr   error
+	reconcileErr       error
+	desiredRel         *release.Release
+	currentRel         *release.Release
+}
+
+func (mag *mockActionGetter) ActionClientFor(ctx context.Context, obj client.Object) (helmclient.ActionInterface, error) {
+	return mag, mag.actionClientForErr
+}
+
+func (mag *mockActionGetter) Get(name string, opts ...helmclient.GetOption) (*release.Release, error) {
+	return mag.currentRel, mag.getClientErr
+}
+
+func (mag *mockActionGetter) History(name string, opts ...helmclient.HistoryOption) ([]*release.Release, error) {
+	return nil, mag.getClientErr
+}
+
+func (mag *mockActionGetter) Install(name, namespace string, chrt *chart.Chart, vals map[string]interface{}, opts ...helmclient.InstallOption) (*release.Release, error) {
+	i := action.Install{}
+	for _, opt := range opts {
+		if err := opt(&i); err != nil {
+			return nil, err
+		}
+	}
+	if i.DryRun {
+		return mag.desiredRel, mag.dryRunInstallErr
+	}
+	return mag.desiredRel, mag.installErr
+}
+
+func (mag *mockActionGetter) Upgrade(name, namespace string, chrt *chart.Chart, vals map[string]interface{}, opts ...helmclient.UpgradeOption) (*release.Release, error) {
+	i := action.Upgrade{}
+	for _, opt := range opts {
+		if err := opt(&i); err != nil {
+			return nil, err
+		}
+	}
+	if i.DryRun {
+		return mag.desiredRel, mag.dryRunUpgradeErr
+	}
+	return mag.desiredRel, mag.upgradeErr
+}
+
+func (mag *mockActionGetter) Uninstall(name string, opts ...helmclient.UninstallOption) (*release.UninstallReleaseResponse, error) {
+	return nil, nil
+}
+
+func (mag *mockActionGetter) Reconcile(rel *release.Release) error {
+	return mag.reconcileErr
+}
+
+var (
+	// required for unmockable call to convert.RegistryV1ToHelmChart
+	validFS = fstest.MapFS{
+		"metadata/annotations.yaml": &fstest.MapFile{Data: []byte("{}")},
+		"manifests": &fstest.MapFile{Data: []byte(`apiVersion: operators.operatorframework.io/v1alpha1
+kind: ClusterServiceVersion
+metadata:
+  name: test.v1.0.0
+  annotations:
+    olm.properties: '[{"type":"from-csv-annotations-key", "value":"from-csv-annotations-value"}]'
+spec:
+  installModes:
+    - type: AllNamespaces
+      supported: true`)},
+	}
+
+	// required for unmockable call to util.ManifestObjects
+	validManifest = `apiVersion: v1
+kind: Service
+metadata:
+  name: service-a
+  namespace: ns-a
+spec:
+  clusterIP: None
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: service-b
+  namespace: ns-b
+spec:
+  clusterIP: 0.0.0.0`
+
+	testCE            = &v1.ClusterExtension{}
+	testObjectLabels  = map[string]string{"object": "label"}
+	testStorageLabels = map[string]string{"storage": "label"}
+)
+
+func TestApply_Base(t *testing.T) {
+	t.Run("fails converting content FS to helm chart", func(t *testing.T) {
+		helmApplier := applier.Helm{}
+
+		objs, state, err := helmApplier.Apply(context.TODO(), os.DirFS("/"), testCE, testObjectLabels, testStorageLabels)
+		require.Error(t, err)
+		require.Nil(t, objs)
+		require.Empty(t, state)
+	})
+
+	t.Run("fails trying to obtain an action client", func(t *testing.T) {
+		mockAcg := &mockActionGetter{actionClientForErr: errors.New("failed getting action client")}
+		helmApplier := applier.Helm{ActionClientGetter: mockAcg}
+
+		objs, state, err := helmApplier.Apply(context.TODO(), validFS, testCE, testObjectLabels, testStorageLabels)
+		require.Error(t, err)
+		require.ErrorContains(t, err, "getting action client")
+		require.Nil(t, objs)
+		require.Empty(t, state)
+	})
+
+	t.Run("fails getting current release and !driver.ErrReleaseNotFound", func(t *testing.T) {
+		mockAcg := &mockActionGetter{getClientErr: errors.New("failed getting current release")}
+		helmApplier := applier.Helm{ActionClientGetter: mockAcg}
+
+		objs, state, err := helmApplier.Apply(context.TODO(), validFS, testCE, testObjectLabels, testStorageLabels)
+		require.Error(t, err)
+		require.ErrorContains(t, err, "getting current release")
+		require.Nil(t, objs)
+		require.Empty(t, state)
+	})
+}
+
+func TestApply_Installation(t *testing.T) {
+	t.Run("fails during dry-run installation", func(t *testing.T) {
+		mockAcg := &mockActionGetter{
+			getClientErr:     driver.ErrReleaseNotFound,
+			dryRunInstallErr: errors.New("failed attempting to dry-run install chart"),
+		}
+		helmApplier := applier.Helm{ActionClientGetter: mockAcg}
+
+		objs, state, err := helmApplier.Apply(context.TODO(), validFS, testCE, testObjectLabels, testStorageLabels)
+		require.Error(t, err)
+		require.ErrorContains(t, err, "attempting to dry-run install chart")
+		require.Nil(t, objs)
+		require.Empty(t, state)
+	})
+
+	t.Run("fails during pre-flight installation", func(t *testing.T) {
+		mockAcg := &mockActionGetter{
+			getClientErr: driver.ErrReleaseNotFound,
+			installErr:   errors.New("failed installing chart"),
+		}
+		mockPf := &mockPreflight{installErr: errors.New("failed during install pre-flight check")}
+		helmApplier := applier.Helm{ActionClientGetter: mockAcg, Preflights: []applier.Preflight{mockPf}}
+
+		objs, state, err := helmApplier.Apply(context.TODO(), validFS, testCE, testObjectLabels, testStorageLabels)
+		require.Error(t, err)
+		require.ErrorContains(t, err, "install pre-flight check")
+		require.Equal(t, applier.StateNeedsInstall, state)
+		require.Nil(t, objs)
+	})
+
+	t.Run("fails during installation", func(t *testing.T) {
+		mockAcg := &mockActionGetter{
+			getClientErr: driver.ErrReleaseNotFound,
+			installErr:   errors.New("failed installing chart"),
+		}
+		helmApplier := applier.Helm{ActionClientGetter: mockAcg}
+
+		objs, state, err := helmApplier.Apply(context.TODO(), validFS, testCE, testObjectLabels, testStorageLabels)
+		require.Error(t, err)
+		require.ErrorContains(t, err, "installing chart")
+		require.Equal(t, applier.StateNeedsInstall, state)
+		require.Nil(t, objs)
+	})
+
+	t.Run("successful installation", func(t *testing.T) {
+		mockAcg := &mockActionGetter{
+			getClientErr: driver.ErrReleaseNotFound,
+			desiredRel: &release.Release{
+				Info:     &release.Info{Status: release.StatusDeployed},
+				Manifest: validManifest,
+			},
+		}
+		helmApplier := applier.Helm{ActionClientGetter: mockAcg}
+
+		objs, state, err := helmApplier.Apply(context.TODO(), validFS, testCE, testObjectLabels, testStorageLabels)
+		require.NoError(t, err)
+		require.Equal(t, applier.StateNeedsInstall, state)
+		require.NotNil(t, objs)
+		assert.Equal(t, "service-a", objs[0].GetName())
+		assert.Equal(t, "service-b", objs[1].GetName())
+	})
+}
+
+func TestApply_Upgrade(t *testing.T) {
+	testCurrentRelease := &release.Release{
+		Info: &release.Info{Status: release.StatusDeployed},
+	}
+
+	t.Run("fails during dry-run upgrade", func(t *testing.T) {
+		mockAcg := &mockActionGetter{
+			dryRunUpgradeErr: errors.New("failed attempting to dry-run upgrade chart"),
+		}
+		helmApplier := applier.Helm{ActionClientGetter: mockAcg}
+
+		objs, state, err := helmApplier.Apply(context.TODO(), validFS, testCE, testObjectLabels, testStorageLabels)
+		require.Error(t, err)
+		require.ErrorContains(t, err, "attempting to dry-run upgrade chart")
+		require.Nil(t, objs)
+		require.Empty(t, state)
+	})
+
+	t.Run("fails during pre-flight upgrade", func(t *testing.T) {
+		testDesiredRelease := *testCurrentRelease
+		testDesiredRelease.Manifest = "do-not-match-current"
+
+		mockAcg := &mockActionGetter{
+			upgradeErr: errors.New("failed upgrading chart"),
+			currentRel: testCurrentRelease,
+			desiredRel: &testDesiredRelease,
+		}
+		mockPf := &mockPreflight{upgradeErr: errors.New("failed during upgrade pre-flight check")}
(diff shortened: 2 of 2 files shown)


internal/applier/helm.go
```diff
@@ -153,13 +153,6 @@ func (h *Helm) Apply(ctx context.Context, contentFS fs.FS, ext *ocv1.ClusterExte
 
 func (h *Helm) getReleaseState(cl helmclient.ActionInterface, ext *ocv1.ClusterExtension, chrt *chart.Chart, values chartutil.Values, post postrender.PostRenderer) (*release.Release, *release.Release, string, error) {
 	currentRelease, err := cl.Get(ext.GetName())
-	if err != nil && !errors.Is(err, driver.ErrReleaseNotFound) {
-		return nil, nil, StateError, err
-	}
-	if errors.Is(err, driver.ErrReleaseNotFound) {
-		return nil, nil, StateNeedsInstall, nil
-	}
```
camilamacedo86, line 161: Hey!

You mentioned:

> contained two if blocks that handled driver.ErrReleaseNotFound with the first one returning, which meant the second block would never run.

From what I see, the logic is set up like this:

- If the error is **not** `driver.ErrReleaseNotFound`, it returns a StateError.
- If the error is `driver.ErrReleaseNotFound`, it returns a StateNeedsInstall status.

Since the conditions are mutually exclusive, the second block does get executed when the error is specifically `driver.ErrReleaseNotFound`. **So, I don’t think it’s accurate to say the second block would never run.**

Given that, I wonder if the proposed change is really necessary. Could there be another reason behind it, like ensuring the desired state is properly set up for Preflight validation? Let me know what you think!

internal/applier/helm.go
```diff
@@ -171,6 +164,10 @@ func (h *Helm) getReleaseState(cl helmclient.ActionInterface, ext *ocv1.ClusterE
 		}
 		return nil, desiredRelease, StateNeedsInstall, nil
 	}
+	if err != nil {
+		return nil, nil, StateError, err
```
camilamacedo86, line 168: Should we really return as a StateError when the error is `driver.ErrReleaseNotFound` which means that the release was not found? Should not this scenario be return as `StateNeedsInstall` as described and in the original code?

What is the problem that you are facing?
Can you please describe What you did? What did you face, and what would you expect instead?

hack/tools/catalogs/download-catalog
```diff
@@ -8,7 +8,7 @@ assert-commands kubectl jq wget
```
camilamacedo86, line 8: Hi @azych,

It looks like you’ve combined three changes/proposed fixes into a single PR. Wouldn’t it make more sense to separate these into three PRs instead? That way, you could clearly describe for each one:

- Why the change is needed,
- What was done,
- What behavior you expected, and
- What issue you encountered.

IHMO: This would make it easier for us to understand the individual scenarios, review them more effectively, and move forward with the proposals.

Let me know your thoughts!

hack/tools/catalogs/download-catalog
```diff
@@ -8,7 +8,7 @@ assert-commands kubectl jq wget
 
 # ClusterCatalog coordinates
 : "${CATALOGD_CATALOGD_SERVICE_NAMESPACE:=olmv1-system}"
-: "${CATALOGD_SERVICE_NAME:=catalogd-server}"
+: "${CATALOGD_SERVICE_NAME:=catalogd-service}"
```
camilamacedo86, line 11: 👍 LGTM this one for me

hack/tools/catalogs/list-compatible-bundles
```diff
@@ -90,7 +90,7 @@ function filter-by-regex-if-necessary() {
         end'
 }
 
-cat - | select-bundle-documents | that-support-allnamespace-install-mode | that-dont-have-dependencies | group-versions-by-package-name | filter-by-regex-if-necessary
+cat - | select-bundle-documents | that-support-allnamespace-install-mode | that-dont-have-dependencies | extract-olm-package-property | group-versions-by-package-name | filter-by-regex-if-necessary
 
 echo "NOTE: OLM v1 currently only supports bundles that support AllNamespaces install model, don't have dependencies, and don't contain webhooks" >&2
 echo "WARNING: These results may include bundles with webhooks, which are incompatible" >&2
```
camilamacedo86, line 96: See the symbol that means we need a space at the end of the file.
Could you please ensure that?

hack/tools/catalogs/list-compatible-bundles
```diff
@@ -77,7 +77,7 @@ function extract-olm-package-property() {
 
 # Group packages by name and collect versions into an array
 function group-versions-by-package-name() {
-    jq -s 'group_by(.packageName) | map({packageName: .[0].packageName, versions: map(.version)})' | regex_it
+    jq -s 'group_by(.packageName) | map({packageName: .[0].packageName, versions: map(.version)})'
```
camilamacedo86, line 80: WDYT about in the description of this change  (assuming that we have a PR to address this specific need), you add the result before the change and after so that we can easily follow up on what you are trying to do/fix here?

internal/applier/helm.go
azych: Hey, thanks for looking at the PR.

So the current state of the code in `getReleaseState` looks like that:

    if errors.Is(err, driver.ErrReleaseNotFound) {
		return nil, nil, StateNeedsInstall, nil
	}

	if errors.Is(err, driver.ErrReleaseNotFound) {
		desiredRelease, err := cl.Install(ext.GetName(), ext.Spec.Namespace, chrt, values, func(i *action.Install) error {
			i.DryRun = true
			i.DryRunOption = "server"
			return nil
		}, helmclient.AppendInstallPostRenderer(post))
		if err != nil {
			return nil, nil, StateError, err
		}
		return nil, desiredRelease, StateNeedsInstall, nil
	}

with both blocks handling the same case (`errors.Is(err, driver.ErrReleaseNotFound)`) only the second one never actually has the chance to do any handling, because the first one `return`s.
My PR removes the first (~no-op) block.

internal/applier/helm.go
azych: `StateError` is not actually being returned for the case when the error is `driver.ErrReleaseNotFound`, only when it is not that specific error type and I am not changing current code behaviour in that respect.
The only thing this PR changes within the applier and `getReleaseState` is to remove the first `errors.Is(err, driver.ErrReleaseNotFound)` block and makes sure the second one runs (please see my response above).

hack/tools/catalogs/download-catalog
azych: I honestly thought that the changes were minor enough to be included in a single PR and was hoping that clearly separating each change into its own commit and explaining it in the description would be enough, which I've also seen in some past PR's.

If you think it's necessary to create a PR for each change, please let me know and I'll be happy to do that.

hack/tools/catalogs/list-compatible-bundles
azych: let me add it to current description to potentially re-use in case of separating PR's

hack/tools/catalogs/download-catalog
camilamacedo86: The advantage of having a PR for each scope here is:
- we can decide move with PR A but not B
- the motivation, reasons and etc are clear
- it makes easier the review.

internal/applier/helm.go
camilamacedo86: The first condition is **!**`errors.Is(err, driver.ErrReleaseNotFound)` (see **!**) which means: if the error is **not** `ErrReleaseNotFound`, it should return `StateError`. 

The second condition applies when the error **is** `ErrReleaseNotFound` (without **!**).

Could you share how you are testing this?  
What do you think about creating an issue to describe:  
- What you did  
- What the result was  
- What you expected in that case  

Also, we are making these changes, but it seems that we are not covering this scenario with tests.  
We probably should add tests for it as well.

camilamacedo86: Sorry,  I might have misunderstood something or did not follow your comment. See: 

You say that `with both blocks handling the same case (errors.Is(err, driver.ErrReleaseNotFound))` but it does not seem to be the case:

- The first condition is **!**`errors.Is(err, driver.ErrReleaseNotFound)` (see **!**), which means: if the error is **not** `ErrReleaseNotFound`, it should return `StateError`. 
- The second condition applies when the error **is** `ErrReleaseNotFound` (without **!**).

Could you share how you are testing this?  
What do you think about creating an issue to describe:  
- What you did  
- What the result was  
- What do you expect in that case  

We should also add tests for it. IHMO, if you still think it should be changed, you should submit a new PR, and we must ensure that the test covers it.

internal/applier/helm.go
camilamacedo86: See; https://github.com/operator-framework/operator-controller/pull/1539#discussion_r1903235394

hack/tools/catalogs/download-catalog
camilamacedo86: Hi @azych,

Thank you for the effort in this PR! IMO, combining three different changes here might make it harder to review, discuss, and move forward efficiently. It could be more productive to split this into separate PRs, each focused on a specific scope.

The benefits of doing so are:

- **Easier Reviews**: Each PR would be smaller and easier to assess.
- **Clearer Context**: The description for each PR could clearly explain:
  - What was changed
  - Why the change is needed
  - How it was tested
  - What was expected vs. the actual behavior observed
- **Independent Progress**: For instance, we could merge PR A while holding off on PR B if it needs more time or discussion.

I appreciate your effort to organize the commits and provide detailed explanations here. However, separating them into PRs would be a minimal additional effort and could significantly benefit the review process.

internal/applier/helm.go
azych: @camilamacedo86 could you please take a look at the actual code? I even pasted it in my previous comment to show that there are two `if` blocks handling the same exact case. What you mention with `!` is a different case for the error is non nil _and_ not `driver.ErrReleaseNotFound`.

I have not done any testing on this and I am not sure if this is something testable at this point.
As per description - the desired state eventually goes into the only currently used preflight validator - `crdupgradesafety` which validates the desired state against the current state, but for `driver.ErrReleaseNotFound`, the current state will not be present and the check will obviously not go through. This means there will not be any effective difference in passing a non nil desired state vs passing a nil one.
The only way I can think of potentially testing this might be to check that the dry run happens (with my fix) vs dry run not happening (current code state), but I'm not sure if that's sensible and what type of effort it might require.

hack/tools/catalogs/download-catalog
azych: sure, I will split this PR into two additional ones.

All of the benefits you mentioned are of course true, but I think it's also good to keep in mind some proportions and try to be more flexible if the scope of the changes (limited to fixes/improvements) is minor/trivial and we can ensure a good separation and clear context. Otherwise we might just be introducing red tape without too much benefit.

internal/applier/helm.go
joelanford: >> We should also add tests for it.

> I have not done any testing on this and I am not sure if this is something testable at this point.

Somewhat convoluted, but I maybe a scenario-like e2e test for this would be:
1. Create a CRD outside of OLM (might have to give it the helm labels that tell helm it is allowed to adopt it; not sure if that is necessary for a helm server-side dry-run install)
2. Install a ClusterExtension with that CRD, where the ClusterExtension's CRD introduces a breaking change according to the pre-flight check.
3. Verify that the CRD pre-flight check fails.

I'm guessing that test would fail on the current code, but pass with your fix?

joelanford: Might be just as reasonable to do a unit test of `getReleaseState`. I'm kinda surprised we don't already have some unit tests in this package.

azych: sure, I'll look into it and see what I can come up with to test this

internal/applier/helm.go
camilamacedo86, line 168: 👍 I see here the return all fine,
I just think would be great that ensure that is covered via tests

azych: just added unit tests

bentito: we have tests now. the new tests for the logic flow in this file seem like pretty good coverage now. thanks!

internal/applier/helm_test.go
```diff
@@ -0,0 +1,315 @@
...
+
+	t.Run("fails getting current release and !driver.ErrReleaseNotFound", func(t *testing.T) {
+		mockAcg := &mockActionGetter{getClientErr: errors.New("failed getting current release")}
+		helmApplier := applier.Helm{ActionClientGetter: mockAcg}
+
+		objs, state, err := helmApplier.Apply(context.TODO(), validFS, testCE, testObjectLabels, testStorageLabels)
+		require.Error(t, err)
+		require.ErrorContains(t, err, "getting current release")
+		require.Nil(t, objs)
+		require.Empty(t, state)
```
camilamacedo86, line 151: Should we not validate that the state is the expected state?

camilamacedo86, line 151: Should not, in this case, we validate that the state is the expected one?

internal/applier/helm_test.go
```diff
@@ -0,0 +1,315 @@
...
+
+	t.Run("fails trying to obtain an action client", func(t *testing.T) {
+		mockAcg := &mockActionGetter{actionClientForErr: errors.New("failed getting action client")}
+		helmApplier := applier.Helm{ActionClientGetter: mockAcg}
+
+		objs, state, err := helmApplier.Apply(context.TODO(), validFS, testCE, testObjectLabels, testStorageLabels)
+		require.Error(t, err)
+		require.ErrorContains(t, err, "getting action client")
+		require.Nil(t, objs)
+		require.Empty(t, state)
```
camilamacedo86, line 140: Should not, in this case, we validate that the state is the expected one?

internal/applier/helm_test.go
```diff
@@ -0,0 +1,315 @@
...
+)
+
+func TestApply_Base(t *testing.T) {
+	t.Run("fails converting content FS to helm chart", func(t *testing.T) {
+		helmApplier := applier.Helm{}
+
+		objs, state, err := helmApplier.Apply(context.TODO(), os.DirFS("/"), testCE, testObjectLabels, testStorageLabels)
+		require.Error(t, err)
+		require.Nil(t, objs)
+		require.Empty(t, state)
```
camilamacedo86, line 129: Should not, in this case, we validate that the state is the expected one?

internal/applier/helm_test.go
```diff
@@ -0,0 +1,315 @@
...
+			getClientErr: driver.ErrReleaseNotFound,
+			installErr:   errors.New("failed attempting to dry-run install chart"),
+		}
+		helmApplier := applier.Helm{ActionClientGetter: mockAcg}
+
+		objs, state, err := helmApplier.Apply(context.TODO(), validFS, testCE, testObjectLabels, testStorageLabels)
+		require.Error(t, err)
+		require.ErrorContains(t, err, "attempting to dry-run install chart")
+		require.Nil(t, objs)
+		require.Empty(t, state)
```
camilamacedo86, line 167: Should not, in this case, we validate that the state is the expected one?

internal/applier/helm_test.go
```diff
@@ -0,0 +1,315 @@
...
+			installErr:      errors.New("failed installing chart"),
+			dryRunInstallOK: true, // let the dry-run installation pass
+		}
+		mockPf := &mockPreflight{installErr: errors.New("failed during install pre-flight check")}
+		helmApplier := applier.Helm{ActionClientGetter: mockAcg, Preflights: []applier.Preflight{mockPf}}
+
+		objs, state, err := helmApplier.Apply(context.TODO(), validFS, testCE, testObjectLabels, testStorageLabels)
+		require.Error(t, err)
+		require.ErrorContains(t, err, "install pre-flight check")
+		require.Equal(t, applier.StateNeedsInstall, state)
```
camilamacedo86, line 182: 👍

internal/applier/helm_test.go
```diff
@@ -0,0 +1,315 @@
...
+			getClientErr:    driver.ErrReleaseNotFound,
+			installErr:      errors.New("failed installing chart"),
+			dryRunInstallOK: true, // let the dry-run installation pass
+		}
+		helmApplier := applier.Helm{ActionClientGetter: mockAcg}
+
+		objs, state, err := helmApplier.Apply(context.TODO(), validFS, testCE, testObjectLabels, testStorageLabels)
+		require.Error(t, err)
+		require.ErrorContains(t, err, "installing chart")
+		require.Equal(t, applier.StateNeedsInstall, state)
```
camilamacedo86, line 197: 👍

internal/applier/helm_test.go
```diff
@@ -0,0 +1,315 @@
...
+			desiredRel: &release.Release{
+				Info:     &release.Info{Status: release.StatusDeployed},
+				Manifest: validManifest,
+			},
+		}
+		helmApplier := applier.Helm{ActionClientGetter: mockAcg}
+
+		objs, state, err := helmApplier.Apply(context.TODO(), validFS, testCE, testObjectLabels, testStorageLabels)
+		require.NoError(t, err)
+		require.Equal(t, applier.StateNeedsInstall, state)
```
camilamacedo86, line 213: Is that the expected state when we have a successful installation?

internal/applier/helm_test.go
```diff
@@ -0,0 +1,315 @@
...
+		mockAcg := &mockActionGetter{
+			upgradeErr: errors.New("failed attempting to dry-run upgrade chart"),
+		}
+		helmApplier := applier.Helm{ActionClientGetter: mockAcg}
+
+		objs, state, err := helmApplier.Apply(context.TODO(), validFS, testCE, testObjectLabels, testStorageLabels)
+		require.Error(t, err)
+		require.ErrorContains(t, err, "attempting to dry-run upgrade chart")
+		require.Nil(t, objs)
+		require.Empty(t, state)
```
camilamacedo86, line 235: Should not, in this case, we validate that the state is the expected one?

internal/applier/helm_test.go
```diff
@@ -0,0 +1,315 @@
...
+			currentRel:      testCurrentRelease,
+			desiredRel:      &testDesiredRelease,
+		}
+		mockPf := &mockPreflight{upgradeErr: errors.New("failed during upgrade pre-flight check")}
+		helmApplier := applier.Helm{ActionClientGetter: mockAcg, Preflights: []applier.Preflight{mockPf}}
+
+		objs, state, err := helmApplier.Apply(context.TODO(), validFS, testCE, testObjectLabels, testStorageLabels)
+		require.Error(t, err)
+		require.ErrorContains(t, err, "upgrade pre-flight check")
+		require.Equal(t, applier.StateNeedsUpgrade, state)
```
camilamacedo86, line 254: 👍

internal/applier/helm_test.go
```diff
@@ -0,0 +1,315 @@
...
+			currentRel:      testCurrentRelease,
+			desiredRel:      &testDesiredRelease,
+		}
+		mockPf := &mockPreflight{}
+		helmApplier := applier.Helm{ActionClientGetter: mockAcg, Preflights: []applier.Preflight{mockPf}}
+
+		objs, state, err := helmApplier.Apply(context.TODO(), validFS, testCE, testObjectLabels, testStorageLabels)
+		require.Error(t, err)
+		require.ErrorContains(t, err, "upgrading chart")
+		require.Equal(t, applier.StateNeedsUpgrade, state)
```
camilamacedo86, line 274: 👍

internal/applier/helm_test.go
```diff
@@ -0,0 +1,315 @@
...
+			currentRel:      testCurrentRelease,
+			desiredRel:      &testDesiredRelease,
+		}
+		mockPf := &mockPreflight{}
+		helmApplier := applier.Helm{ActionClientGetter: mockAcg, Preflights: []applier.Preflight{mockPf}}
+
+		objs, state, err := helmApplier.Apply(context.TODO(), validFS, testCE, testObjectLabels, testStorageLabels)
+		require.Error(t, err)
+		require.ErrorContains(t, err, "reconciling charts")
+		require.Equal(t, applier.StateUnchanged, state)
```
camilamacedo86, line 294: 👍

internal/applier/helm_test.go
```diff
@@ -0,0 +1,315 @@
...
+
+		mockAcg := &mockActionGetter{
+			currentRel: testCurrentRelease,
+			desiredRel: &testDesiredRelease,
+		}
+		helmApplier := applier.Helm{ActionClientGetter: mockAcg}
+
+		objs, state, err := helmApplier.Apply(context.TODO(), validFS, testCE, testObjectLabels, testStorageLabels)
+		require.NoError(t, err)
+		require.Equal(t, applier.StateNeedsUpgrade, state)
```
camilamacedo86, line 310: Is that the correct state expected when the upgrade occurs with success?

internal/applier/helm_test.go
azych: it is being validated. The expected valid state here and in other places like that is an empty string

internal/applier/helm_test.go
```diff
@@ -0,0 +1,315 @@
...
+
+func (mag *mockActionGetter) History(name string, opts ...helmclient.HistoryOption) ([]*release.Release, error) {
+	return nil, mag.getClientErr
+}
+
+func (mag *mockActionGetter) Install(name, namespace string, chrt *chart.Chart, vals map[string]interface{}, opts ...helmclient.InstallOption) (*release.Release, error) {
+	if mag.dryRunInstallOK {
+		mag.dryRunInstallOK = false
+		return mag.desiredRel, nil
+	}
```
joelanford, line 64: This only works because the current code always calls:
- Install(dryRun=true)
- Install(dryRun=false)

In that order.

I worry that we could potentially change that guarantee in the future, but in a way where our tests continue to pass when they shouldn't.

It might be worth it to handle the functional `opts` and explicitly check to see if this call is a dry run.

```suggestion
	i := action.Install{}
	for _, opt := range opts {
		if err := opt(&i); err != nil {
			return nil, err
		}
	}
	if i.DryRun && mag.dryRunInstallOK {
		return mag.desiredRel, nil
	}
```

joelanford: Same for upgrade.

internal/applier/helm_test.go
azych: Short answer: yes.
Longer answer: applier only defines 4 different states now:
> 	StateNeedsInstall     string = "NeedsInstall"
	StateNeedsUpgrade     string = "NeedsUpgrade"
	StateUnchanged        string = "Unchanged"
	StateError            string = "Error"

If you're asking because one could potentially expect 'Installed' instead of 'NeedsInstall' after successful installation, I am not sure how the transition looks down the call stack and if it might be something that might be needed.
Something for the future maybe.

internal/applier/helm_test.go
azych: please see https://github.com/operator-framework/operator-controller/pull/1539#discussion_r1906034448

internal/applier/helm_test.go
azych: good idea, the flag with order was a little hacky ;) I'll make the changes for both `Install()` and `Upgrade()`

internal/applier/helm_test.go
camilamacedo86: Yes, having the status NeedsInstall when it was successfully installed is strange, right?
But I agree that is out of the scope. 

@joelanford @bentito, is that a bug we should track/address in a follow-up? 
OR do we want to return that? Is not it odd?

internal/applier/helm_test.go
camilamacedo86: @joelanford @bentito 
Same here. Is it not a bug? 
Should we have the state `StateNeedsUpgrade ` when the upgrade was successful?

azych: so just my 2c after looking at the code - `Apply` is only ever called by the cluster extension controller and the status it returns is currently being ignored altogether:

> managedObjs, _, err := r.Applier.Apply(ctx, unpackResult.Bundle, ext, objLbls, storeLbls)

In case there was no error, the extension status will be set as installed in status conditions, so there is no differentiation between reconcilier having to internally upgrade an existing release or just do a fresh install.
IMHO, that probably makes sense as the final status info to pass (current bundle status being 'installed' rather than 'upgraded').

azych:
@camilamacedo86 @joelanford @bentito 
I've just squashed the changes and the fix should be ready to go. Please take a look and approve.

I think any potential discussion about https://github.com/operator-framework/operator-controller/pull/1539#discussion_r1906146507 should not be a blocker and can be had outside (maybe on an issue?) if necessary, WDYT?

internal/applier/helm_test.go
camilamacedo86: Yes, I think we do not update it after that,
It is out of the scope anyway.

I will put it in my TODO list to do a few tests and see how is the UX 
If I think that is confusing, then I open an issue for us to discuss. 

Thank you a lot. Marking as solved

camilamacedo86 (approved):
Great work 🥇 
/lgtm

perdasilva (approved):
looks awesome! Thanks for the thorough unit testing ^^

PR #1548, opened 2025-01-07
🌱 Increase timeout for upgrade-e2e test to avoid flakes

camilamacedo86:
Increasing the timeout to avoid flakes

PS.: I notice that it sometimes fails in GitHub action; therefore, I am just increasing to avoid flakes. 
Example: https://github.com/operator-framework/operator-controller/actions/runs/12653708386/job/35259771016?pr=1542

(I think it occurs when the same action is running and we push more changes )


joelanford:
One thing I think all reviewers should ask when a PR comes in for bumping a timeout (so I'll ask): what exactly is taking more than a minute when the flakes happen? Is there a different change we could make such that the test would execute within the existing time budget?

To me, there's a bit of a smell if it takes 3 minutes for an upgrade to happen.

camilamacedo86:
IMO, 1 minute is too low for this test.
I stopped to see the flake with 3, but 2 seems reasonable. 
However, I closed this PR as not accepted and tracked an issue with the discussion summary so we can plan any further investigation since it is appropriately required.

PR #1554, opened 2025-01-08
🐛 Fix operator-controller not using updated registries configuration

dtfranz:
~NOTE: Same issue applies to catalogd, I'll make a similar change there.~ Done

With this change, operator-controller will be able to invalidate the configuration cache of `containers/image` and accept new configuration at runtime when changes occur.

~For code simplicity's sake, there's a few things to take note of that can be addressed if deemed necessary:~
* ~The cache always invalidates on the first unpack, as we've not yet read the `registries.conf` file. This is easily addressed but I wanted to keep `main()` clear of the file reading op.~
* ~We only check the `registries.conf` file. Checking the entire directory is more involved since we have to traverse the folder tree along with the symlinks dropped in there by k8s (`WalkDir` does not support symlinks , but it might [get easier](https://github.com/golang/go/issues/49580) in `go` 1.25?)~

### DIFF
--- a/internal/rukpak/source/containers_image.go
+++ b/internal/rukpak/source/containers_image.go
@@ -17,6 +17,7 @@ import (
 	"github.com/containers/image/v5/oci/layout"
 	"github.com/containers/image/v5/pkg/blobinfocache/none"
 	"github.com/containers/image/v5/pkg/compression"
+	"github.com/containers/image/v5/pkg/sysregistriesv2"
 	"github.com/containers/image/v5/signature"
 	"github.com/containers/image/v5/types"
 	"github.com/go-logr/logr"
@@ -43,10 +44,14 @@ func (i *ContainersImageRegistry) Unpack(ctx context.Context, bundle *BundleSour
 		return nil, reconcile.TerminalError(fmt.Errorf("error parsing bundle, bundle %s has a nil image source", bundle.Name))
 	}
 
+	// Reload registries cache in case of configuration update
+	sysregistriesv2.InvalidateCache()
+
 	srcCtx, err := i.SourceContextFunc(l)
 	if err != nil {
 		return nil, err
 	}
+
 	//////////////////////////////////////////////////////
 	//
 	// Resolve a canonical reference for the image.
@@ -254,6 +259,11 @@ func (i *ContainersImageRegistry) unpackImage(ctx context.Context, unpackPath st
 	if err != nil {
 		return fmt.Errorf("error creating image source: %w", err)
 	}
+	defer func() {
+		if err := layoutSrc.Close(); err != nil {
+			panic(err)
+		}
+	}()
 
 	if err := os.MkdirAll(unpackPath, 0700); err != nil {
 		return fmt.Errorf("error creating unpack directory: %w", err)
--- a/catalogd/internal/source/containers_image.go
+++ b/catalogd/internal/source/containers_image.go
@@ -20,6 +20,7 @@ import (
 	"github.com/containers/image/v5/oci/layout"
 	"github.com/containers/image/v5/pkg/blobinfocache/none"
 	"github.com/containers/image/v5/pkg/compression"
+	"github.com/containers/image/v5/pkg/sysregistriesv2"
 	"github.com/containers/image/v5/signature"
 	"github.com/containers/image/v5/types"
 	"github.com/go-logr/logr"
@@ -50,6 +51,9 @@ func (i *ContainersImageRegistry) Unpack(ctx context.Context, catalog *catalogdv
 	if catalog.Spec.Source.Image == nil {
 		return nil, reconcile.TerminalError(fmt.Errorf("error parsing catalog, catalog %s has a nil image source", catalog.Name))
 	}
+
+	// Reload registries cache in case of configuration update
+	sysregistriesv2.InvalidateCache()
 
 	srcCtx, err := i.SourceContextFunc(l)
 	if err != nil {
--- a/cmd/operator-controller/main.go
+++ b/cmd/operator-controller/main.go
@@ -306,7 +306,8 @@ func main() {
 				return nil, fmt.Errorf("could not stat auth file, error: %w", err)
 			}
 			return srcContext, nil
-		}}
+		},
+	}
 
 	clusterExtensionFinalizers := crfinalizer.NewFinalizers()
 	if err := clusterExtensionFinalizers.Register(controllers.ClusterExtensionCleanupUnpackCacheFinalizer, finalizers.FinalizerFunc(func(ctx context.Context, obj client.Object) (crfinalizer.Result, error) {
--- a/test/e2e/cluster_extension_install_test.go
+++ b/test/e2e/cluster_extension_install_test.go
@@ -357,6 +357,82 @@ func TestClusterExtensionInstallRegistry(t *testing.T) {
 	}
 }
 
+func TestClusterExtensionInstallRegistryDynamic(t *testing.T) {
+	// NOTE: Like 'TestClusterExtensionInstallRegistry', this test also requires extra configuration in /etc/containers/registries.conf
+	packageName := "dynamic"
+
+	t.Log("When a cluster extension is installed from a catalog")
+	t.Log("When the extension bundle format is registry+v1")
+
+	clusterExtension, extensionCatalog, sa, ns := testInit(t)
+	defer testCleanup(t, extensionCatalog, clusterExtension, sa, ns)
+	defer getArtifactsOutput(t)
+
+	clusterExtension.Spec = ocv1.ClusterExtensionSpec{
+		Source: ocv1.SourceConfig{
+			SourceType: "Catalog",
+			Catalog: &ocv1.CatalogSource{
+				PackageName: packageName,
+				Selector: &metav1.LabelSelector{
+					MatchLabels: map[string]string{"olm.operatorframework.io/metadata.name": extensionCatalog.Name},
+				},
+			},
+		},
+		Namespace: ns.Name,
+		ServiceAccount: ocv1.ServiceAccountReference{
+			Name: sa.Name,
+		},
+	}
+	t.Log("It updates the registries.conf file contents")
+	cm := corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "e2e-registries-conf",
+			Namespace: "olmv1-system",
+		},
+		Data: map[string]string{
+			"registries.conf": `[[registry]]
+prefix = "dynamic-registry.operator-controller-e2e.svc.cluster.local:5000"
+location = "docker-registry.operator-controller-e2e.svc.cluster.local:5000"`,
+		},
+	}
+	require.NoError(t, c.Update(context.Background(), &cm))
+
+	t.Log("It resolves the specified package with correct bundle path")
+	t.Log("By creating the ClusterExtension resource")
+	require.NoError(t, c.Create(context.Background(), clusterExtension))
+
+	t.Log("By eventually reporting a successful resolution and bundle path")
+	require.EventuallyWithT(t, func(ct *assert.CollectT) {
+		assert.NoError(ct, c.Get(context.Background(), types.NamespacedName{Name: clusterExtension.Name}, clusterExtension))
+	}, 2*time.Minute, pollInterval)
+
+	// Give the check 2 minutes instead of the typical 1 for the pod's
+	// files to update from the configmap change.
+	// The theoretical max time is the kubelet sync period of 1 minute +
+	// ConfigMap cache TTL of 1 minute = 2 minutes
+	t.Log("By eventually reporting progressing as True")
+	require.EventuallyWithT(t, func(ct *assert.CollectT) {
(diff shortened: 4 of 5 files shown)


dtfranz:
Working on tests
/hold

cmd/manager/main.go
```diff
@@ -115,6 +116,7 @@ func main() {
 	flag.BoolVar(&operatorControllerVersion, "version", false, "Prints operator-controller version information")
 	flag.StringVar(&systemNamespace, "system-namespace", "", "Configures the namespace that gets used to deploy system resources.")
 	flag.StringVar(&globalPullSecret, "global-pull-secret", "", "The <namespace>/<name> of the global pull secret that is going to be used to pull bundle images.")
+	flag.StringVar(&registriesConfig, "registries-config", "/etc/containers/registries.conf", "The local file path of the registries configuration file")
```
azych, line 119: hey @dtfranz, just a question - could you explain how this file would be initially created and then tracked?

In case of globalPullSecret, we set up a reconciler that is able to watch on the secret changes and then record changes in data in local authFile, but I don't see which part of the code would be responsible for working with the registries file.

also, NIT: maybe `registriesConfigPath` similar to how there's `certFile` and `keyFile`?

internal/rukpak/source/containers_image.go
```diff
@@ -45,6 +48,22 @@ func (i *ContainersImageRegistry) Unpack(ctx context.Context, bundle *BundleSour
...
+	//////////////////////////////////////////////////////
+	//
+	// Verify that registries configuration is up-to-date
+	//
+	//////////////////////////////////////////////////////
+	data, err := os.ReadFile(srcCtx.SystemRegistriesConfPath)
+	if err != nil {
+		return nil, fmt.Errorf("could not read registries configuration file, error: %w", err)
+	}
+	registriesConfHash := sha256.Sum256(data)
```
joelanford, line 61: It would be more memory efficient to `os.Open` + `defer file.Close(), and then `io.Copy` from the file into an instance of the hash object,. If you choose to make that change, I'd also suggest extracting that logic into a separate function since it is slightly more complex.

cmd/manager/main.go
```diff
@@ -294,8 +296,9 @@ func main() {
 		BaseCachePath: filepath.Join(cachePath, "unpack"),
 		SourceContextFunc: func(logger logr.Logger) (*types.SystemContext, error) {
 			srcContext := &types.SystemContext{
-				DockerCertPath: caCertDir,
-				OCICertPath:    caCertDir,
+				DockerCertPath:           caCertDir,
+				OCICertPath:              caCertDir,
+				SystemRegistriesConfPath: registriesConfig,
```
joelanford, line 301: I suspect this may result in other subtle bugs if there are files in `/etc/containers/registry.conf.d/`, but I don't know for sure. See: https://github.com/containers/image/blob/main/docs/containers-registries.conf.d.5.md

Either way, I think we may need to compute a hash over the entire `/etc/containers` directory in order to guarantee that we catch all relevant file changes. Maybe with `filepath.Walk` that writes every file path and file content to a hasher?

cmd/manager/main.go
joelanford: The entire `/etc/containers` directory can be mounted into the pod. `/etc/containers` is the default configuration directory for `containers/image`

Given my comment about the likely need to read/hash the entire directory content, I would suggest dropping this flag and hardcoding the assumption of `/etc/containers` (Maybe there's a constant in the `containers/image` library we could use?)

internal/rukpak/source/containers_image.go
```diff
@@ -45,6 +48,22 @@ func (i *ContainersImageRegistry) Unpack(ctx context.Context, bundle *BundleSour
...
+	if registriesConfHash != i.registriesConfHash {
+		i.registriesConfHash = registriesConfHash
```
joelanford, line 63: Pretty sure the `ContainersImageRegistry` struct is thread-safe and meant to be usable concurrently. With that in mind, we would need the entire "compute hash + update hash field" logic to be guarded by a mutex.

cmd/manager/main.go
azych: this may also be naive, but if it _can_ (optionally) be mounted then should we always assume it has been mounted in `Unpack`? Maybe it would be better to check if the file/directory exists first?

cmd/manager/main.go
tmshort: +1 on the `.d` config directory

dtfranz: Unfortunately `filepath.Walk` does not handle the symlinks that get mounted in the pod, so we'll need to write our own thing unless there's another good library out there somewhere. Agreed here as well though that we should probably just check the entire directory.

From what I understand though @joelanford, K8s updates a symlink in the mounted folder named with a timestamp each time the mounted volume is updated. Could we not just check the filename and see if the timestamp has changed since our last `Unpack`?

internal/rukpak/source/containers_image.go
dtfranz: Absolutely, that was a miss - I'll add that!

cmd/manager/main.go
joelanford: Oh yeah, if that works, that would be much easier and more efficient.

internal/rukpak/source/containers_image.go
```diff
@@ -171,6 +194,35 @@ func (i *ContainersImageRegistry) Cleanup(_ context.Context, bundle *BundleSourc
...
+		// Ignore not found errors
+		return fmt.Errorf("could not read registries configuration directory, error: %w", err)
+	}
+	for _, entry := range dirEntries {
+		if entry.Type().IsRegular() || !strings.HasPrefix(entry.Name(), "..") {
+			// Skip 'normal' entries
+			continue
+		}
+		// Trim the leading '..' for symlinks
+		trimmed := strings.TrimLeft(entry.Name(), "..")
```
azych, line 211: `TrimLeft` would actually strip all of the leading `.` chars instead of `..`. You might use `TrimPrefix` instead.

internal/rukpak/source/containers_image.go
```diff
@@ -171,6 +194,35 @@ func (i *ContainersImageRegistry) Cleanup(_ context.Context, bundle *BundleSourc
...
+		t, err := time.Parse(mountedSymlinkTimestampFormat, trimmed)
+		if err != nil {
+			continue
+		} else if t != i.RegistriesConfTimestamp {
```
azych, line 215: it's generally safer to use `t.Equal()` rather than `==`, see: https://pkg.go.dev/time#hdr-Monotonic_Clocks

internal/rukpak/source/containers_image.go
```diff
@@ -45,6 +57,17 @@ func (i *ContainersImageRegistry) Unpack(ctx context.Context, bundle *BundleSour
...
 	}
+
+	//////////////////////////////////////////////////////
+	//
+	// Verify that registries configuration is up-to-date
+	//
+	//////////////////////////////////////////////////////
+	err = i.checkUpdatedConfiguration()
+	if err != nil {
+		return nil, err
```
azych, line 68: might be a good idea to wrap the error with some context, eg.   `fmt.Errorf("error checking registries configuration: %w", err)`, WDYT?

internal/rukpak/source/containers_image.go
```diff
@@ -250,6 +302,11 @@ func (i *ContainersImageRegistry) unpackImage(ctx context.Context, unpackPath st
 	if err != nil {
 		return fmt.Errorf("error creating image source: %w", err)
 	}
+	defer func() {
+		if err := layoutSrc.Close(); err != nil {
+			panic(err)
```
azych, line 307: just a question - imho panicking here seems a little excessive, but I see it in multiple places, so I think it's getting caught

internal/rukpak/source/containers_image.go
```diff
@@ -171,6 +194,35 @@ func (i *ContainersImageRegistry) Cleanup(_ context.Context, bundle *BundleSourc
...
+	for _, entry := range dirEntries {
+		if entry.Type().IsRegular() || !strings.HasPrefix(entry.Name(), "..") {
+			// Skip 'normal' entries
+			continue
+		}
+		// Trim the leading '..' for symlinks
+		trimmed := strings.TrimLeft(entry.Name(), "..")
+		t, err := time.Parse(mountedSymlinkTimestampFormat, trimmed)
+		if err != nil {
+			continue
```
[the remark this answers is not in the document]
dtfranz: I'd argue yes, they should be ignored. If we were to log this then every time we `Unpack` we would get `n` logs with `time.Parse` errors where `n` is the number of files we checked before we encountered the timestamped file. I don't think those logs would be of any value and could potentially be spammy depending on how many files are in the mounted directory.

internal/rukpak/source/containers_image.go
dtfranz: Just following convention here :) Not sure it's the right way to handle it either but I just noticed the source close was missing so I wanted to catch that while I was in here.

internal/rukpak/source/containers_image.go
```diff
@@ -171,6 +195,35 @@ func (i *ContainersImageRegistry) Cleanup(_ context.Context, bundle *BundleSourc
...
+			// Skip 'normal' entries
+			continue
+		}
+		// Trim the leading '..' for symlinks
+		trimmed := strings.TrimPrefix(entry.Name(), "..")
+		t, err := time.Parse(mountedSymlinkTimestampFormat, trimmed)
+		if err != nil {
+			continue
+		} else if !t.Equal(i.RegistriesConfTimestamp) {
+			// We've found the timestamped symlink and it's been updated
```
azych, line 217: @dtfranz @joelanford 

turns out there might be an issue with relying on created symlinks: if you use `subPath` to mount a specific file from ConfigMap, the symlinks are not created at all. See https://stackoverflow.com/a/50687707 and https://github.com/kubernetes/kubernetes/issues/50345 for context.

Note: _can't seem to have the powers to unresolve a previous discussion, so adding it here._

dtfranz: Given that, an `fsnotify` solution is starting to seem more attractive to me now :sweat_smile:

internal/rukpak/source/containers_image.go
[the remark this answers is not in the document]
azych: you're probably right, there may be mounts with multiple symlinks which would not be ideal.
I was just trying to figure out a way so that issues with potential formatting of timestamp wouldn't be a silent issue in the future.

Anyway, this seems no longer valid given https://github.com/operator-framework/operator-controller/pull/1554#discussion_r1909384994

internal/rukpak/source/containers_image.go
joelanford: Going back to a question I saw @perdasilva ask.. What is the cost of invalidating every single time? I assume that would mean that `Unpack` would be re-reading files and re-loading essentially the entire configuration at every pass.

WDYT about starting out with "invalidate every time" to keep code complexity low, with a comment that explains that we pay this reload cost unnecessarily, and that if it becomes a performance issue, we should follow up later and optimize to invalidate only on changes.

joelanford:
Now that `catalogd` is included in this repo, those changes can be incorporated in this PR as well.

dtfranz:
> Now that `catalogd` is included in this repo, those changes can be incorporated in this PR as well.

Absolutely - I'll include the change in there as well.

internal/rukpak/source/containers_image.go
dtfranz: I'm fine with that personally. Have we encountered any performance issues otherwise so far? Just so we aren't adding on top of anything else. AFAIK we aren't, so I'd be OK with invalidating on each pass with a follow up issue to track it.

dtfranz: @perdasilva now kicking myself because I completely misinterpreted that question you asked - my bad!! I thought you were saying my first change was invalidating on each pass and didn't read it as a recommendation :sweat_smile: :sweat_smile:

dtfranz: Running an extremely rudimentary time elapsed check on a local e2e run, the cache invalidation can take between 500ns and 3µs. I think that's acceptable though it may be worth mentioning every unpack will have to wait on the config mutex from any prior unpack.

perdasilva: all good ^^

perdasilva:
@dtfranz @joelanford one thing I was thinking about is that (and maybe this has changed since) the unpack flow blocks the reconciliation thread. I wonder if we should just spawn the unpack process, update the status and re-enqueue after 30 seconds (or whatever), check the status of the process, update status, repeat until the process reaches a terminal state.

internal/rukpak/source/containers_image.go
perdasilva: maybe let's proceed with the convention for this PR, but create a discussion in upstream slack? I'm also a bit :thinking: about these panics. Maybe we should level set on a tenet for when to use it and why?

perdasilva (approved):
lgtm =D great work! thank you!!!

PR #1558, opened 2025-01-09
📖  release: update the doc since it is no longer in a initial development

camilamacedo86:
After cut 1.0.0, we should follow semver
Therefore, this PR only removes the outdated and no relevant info


LalatenduMohanty (approved):
/lgtm

PR #1564, opened 2025-01-10
✨ improve logging: catalog http server, op-con resolver

joelanford:
# Description

This adds more logging to help administrators troubleshoot issues resolving bundles from catalogs. Specifically, it:
- Adds logging to the catalogd http server, so that all requests are logged with useful metadata about the request and response
- Adds logging in the resolution code path to:
    - report with catalogs are included in the resolution
    - report whether the desired package is found in each catalog
    - report the total count of bundles and the matched count of bundles in each catalog

This should help administrators narrow down which stage of resolution should be investigated when unexpected resolution errors occur.

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/catalogd/internal/serverutil/serverutil.go
+++ b/catalogd/internal/serverutil/serverutil.go
@@ -3,10 +3,13 @@ package serverutil
 import (
 	"crypto/tls"
 	"fmt"
+	"io"
 	"net"
 	"net/http"
 	"time"
 
+	"github.com/go-logr/logr"
+	"github.com/gorilla/handlers"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/certwatcher"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
@@ -40,12 +43,16 @@ func AddCatalogServerToManager(mgr ctrl.Manager, cfg CatalogServerConfig, tlsFil
 
 	shutdownTimeout := 30 * time.Second
 
+	l := mgr.GetLogger().WithName("catalogd-http-server")
+	handler := catalogdmetrics.AddMetricsToHandler(cfg.LocalStorage.StorageServerHandler())
+	handler = logrLoggingHandler(l, handler)
+
 	catalogServer := manager.Server{
 		Name:                "catalogs",
 		OnlyServeWhenLeader: true,
 		Server: &http.Server{
 			Addr:        cfg.CatalogAddr,
-			Handler:     catalogdmetrics.AddMetricsToHandler(cfg.LocalStorage.StorageServerHandler()),
+			Handler:     handler,
 			ReadTimeout: 5 * time.Second,
 			// TODO: Revert this to 10 seconds if/when the API
 			// evolves to have significantly smaller responses
@@ -61,4 +68,32 @@ func AddCatalogServerToManager(mgr ctrl.Manager, cfg CatalogServerConfig, tlsFil
 	}
 
 	return nil
+}
+
+func logrLoggingHandler(l logr.Logger, handler http.Handler) http.Handler {
+	return handlers.CustomLoggingHandler(nil, handler, func(_ io.Writer, params handlers.LogFormatterParams) {
+		// extract parameters used in apache common log format, but then log using `logr` to remain consistent
+		// with other loggers used in this codebase.
+		username := "-"
+		if params.URL.User != nil {
+			if name := params.URL.User.Username(); name != "" {
+				username = name
+			}
+		}
+
+		host, _, err := net.SplitHostPort(params.Request.RemoteAddr)
+		if err != nil {
+			host = params.Request.RemoteAddr
+		}
+
+		uri := params.Request.RequestURI
+		if params.Request.ProtoMajor == 2 && params.Request.Method == http.MethodConnect {
+			uri = params.Request.Host
+		}
+		if uri == "" {
+			uri = params.URL.RequestURI()
+		}
+
+		l.Info("handled request", "host", host, "username", username, "method", params.Request.Method, "uri", uri, "protocol", params.Request.Proto, "status", params.StatusCode, "size", params.Size)
+	})
 }
--- a/internal/resolve/catalog.go
+++ b/internal/resolve/catalog.go
@@ -39,6 +39,7 @@ type foundBundle struct {
 
 // Resolve returns a Bundle from a catalog that needs to get installed on the cluster.
 func (r *CatalogResolver) Resolve(ctx context.Context, ext *ocv1.ClusterExtension, installedBundle *ocv1.BundleMetadata) (*declcfg.Bundle, *bsemver.Version, *declcfg.Deprecation, error) {
+	l := log.FromContext(ctx)
 	packageName := ext.Spec.Source.Catalog.PackageName
 	versionRange := ext.Spec.Source.Catalog.Version
 	channels := ext.Spec.Source.Catalog.Channels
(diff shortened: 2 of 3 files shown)


camilamacedo86 (dismissed):
/lgtm

catalogd/internal/serverutil/serverutil.go
```diff
@@ -61,3 +68,29 @@ func AddCatalogServerToManager(mgr ctrl.Manager, cfg CatalogServerConfig, tlsFil
...
+
+		uri := params.Request.RequestURI
+		if params.Request.ProtoMajor == 2 && params.Request.Method == http.MethodConnect {
+			uri = params.Request.Host
+		}
+		if uri == "" {
+			uri = params.URL.RequestURI()
+		}
+
+		l.Info("handled request", "host", host, "username", username, "method", params.Request.Method, "uri", uri, "protocol", params.Request.Proto, "status", params.StatusCode, "size", params.Size)
```
azych, line 94: wouldn't we get the same output from the standard `LoggingHandler` middleware that gorilla provides? As per docs, it logs using the Apache Common Log Format. Looking at its internal `buildCommonLogLine`:
> 	buf := make([]byte, 0, 3*(len(host)+len(username)+len(req.Method)+len(uri)+len(req.Proto)+50)/2)
	buf = append(buf, host...)
	buf = append(buf, " - "...)
	buf = append(buf, username...)
	buf = append(buf, " ["...)
	buf = append(buf, ts.Format("02/Jan/2006:15:04:05 -0700")...)
	buf = append(buf, `] "`...)
	buf = append(buf, req.Method...)
	buf = append(buf, " "...)
	buf = appendQuoted(buf, uri)
	buf = append(buf, " "...)
	buf = append(buf, req.Proto...)
	buf = append(buf, `" `...)
	buf = append(buf, strconv.Itoa(status)...)
	buf = append(buf, " "...)
	buf = append(buf, strconv.Itoa(size)...)

config/samples/olm_v1_clusterextension.yaml
```diff
@@ -281,5 +281,5 @@ spec:
   source:
     sourceType: Catalog
     catalog:
-      packageName: argocd-operator
+      packageName: foobar
```
azych, line 284: nit: looks like dev leftover?

internal/resolve/catalog.go
```diff
@@ -65,6 +66,15 @@ func (r *CatalogResolver) Resolve(ctx context.Context, ext *ocv1.ClusterExtensio
 		}
 	}
 
+	type catStat struct {
```
azych, line 69: WDYT about extracting this to package-level and making attributes private?

catalogd/internal/serverutil/serverutil.go
joelanford: That would result in non-structured log lines. Since we use logr throughout the catalogd codebase, I opted to use logic from `buildCommonLogLine` to derive each field, but then use a `logr.Logger` instance to actually log the line. That way log output is remains consistent.

config/samples/olm_v1_clusterextension.yaml
joelanford: Oops! Yep.

internal/resolve/catalog.go
joelanford: I started out with private attributes, but then when these are eventually passed into `l.Info()`, logr only outputs the exported attributes. We could write more code to specifically log each unexported field, but that would be brittle if we ever add new fields.

Not highly opposed to extracting to package level, but I didn't see any use case where this would be used outside the function, so I figured the more local, the better.

azych: you're right, though it's actually the `zap` backend sink that controller-runtime/log uses which requires public fields - see https://github.com/go-logr/logr/issues/142

Regarding the declaration, sure it can be extracted in the future if there's a need. Personally I just find it more readable with type declarations done on the outside and functions/methods kept smaller.

config/samples/olm_v1_clusterextension.yaml
camilamacedo86: We do nothing related the name of the pkg right
Otherwise, it would fail. But good catcher 
It is good keep the name so we know the source (from where the mock data came from)

internal/resolve/catalog.go
joelanford: Hmm, I thought I had switched everything over to the klog backend a month or so ago.

internal/resolve/catalog.go
```diff
@@ -76,6 +86,16 @@ func (r *CatalogResolver) Resolve(ctx context.Context, ext *ocv1.ClusterExtensio
 			return fmt.Errorf("error getting package %q from catalog %q: %w", packageName, cat.Name, err)
 		}
 
+		cs := catStat{CatalogName: cat.Name}
+		catStats = append(catStats, &cs)
+
+		if isFBCEmpty(packageFBC) {
+			return nil
+		}
```
tmshort, line 94: This early exit did not exist before. Are we worried about a change in behavior?

internal/resolve/catalog.go
```diff
@@ -158,6 +179,7 @@ func (r *CatalogResolver) Resolve(ctx context.Context, ext *ocv1.ClusterExtensio
 
 	// Check for ambiguity
 	if len(resolvedBundles) != 1 {
+		l.Info("resolution failed", "stats", catStats)
```
tmshort, line 182: Do we want this before all returns from this point forward?

internal/resolve/catalog.go
azych: I think we should also ensure that the log output is lowercased, which I am not sure happens right now.

With `textlogger` it can be done by making the `catStat` implement `slog.LogValuer` for example, which would also solve the problem of private attributes - ie. we would be free to make them private.

Alternatively, adding tags with lowercase keys to the struct would also work.

internal/resolve/catalog.go
joelanford: Before, if `len(packageFBC.Bundles)` was 0, we would evaluate/run the predicates against a 0-length slice, which would return a 0-length slice, and then we'd `return nil` in line 103 (old) / line 124 (new)

So I don't think there's a behavior change here other than skipping some unnecessary predicate setup.

internal/resolve/catalog.go
joelanford: In the log line, the structured log _keys_ are lowercase. However, the `catStat` value string contains field names that match the casing of the struct fields (hence they are currently capitalized).

I don't personally think we _need_ to downcase the first letter of those field names, but I'm also not really against it. I'll generally agree that the logs would look more consistent if we did.

I'll take a look at what it would take to do that. It might just be json tags.

internal/resolve/catalog.go
joelanford: We could? I felt like those other returns include context in the error message that would make it possible to figure out what happened, so that's why I only included it here.

internal/resolve/catalog.go
```diff
@@ -271,3 +299,18 @@ func CatalogWalker(
 		return nil
 	}
 }
+
+func isFBCEmpty(fbc *declcfg.DeclarativeConfig) bool {
+	if fbc == nil {
+		return true
+	}
+	return len(fbc.Packages) == 0 && len(fbc.Channels) == 0 && len(fbc.Bundles) == 0 && len(fbc.Deprecations) == 0 && len(fbc.Others) == 0
```
camilamacedo86, line 307: Hi @joelanford 

IHMO IsEmpty should be implemented in DeclarativeConfig itself. However, that is not a blocker for us to move forward here.

camilamacedo86 (approved):
I think it is good enough for us move forward

/lgtm

internal/resolve/catalog.go
joelanford: Adding JSON tags fixed it!

tmshort (approved):
LGTM

PR #1565, opened 2025-01-10
:sparkles: use single semver library

azych:
There are currently two libraries being used directly in the code to deal with semver: `github.com/Masterminds/semver/v3` and `github.com/blang/semver/v4`.
This PR tries to elimitate one of them (blang/semver), since the second library seems to be covering all use cases where blang was being used.
This has a direct impact on security as well as readability and general development, because we won't be needing to deal with different types that ultimately handle the same thing.

At this stage, this is meant to be a draft looking for opinions and any considerations which would explain why this shouldn't be done. I have not looked into masterminds `NewVersion()` vs `StrictNewVersion()` for example and if there would be a need to use the latter.

Rationale for choosing to go with mastermind's lib in this draft was that we only used two methods from blang that mm covered and used a lot more functionality of mm that blang didin't cover (or at least doesn't seem to cover).

Note: blang/semver still remains as an indirect dependency, potentially because of use in `declcfg`.

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/catalogd/internal/version/version.go
+++ b/catalogd/internal/version/version.go
@@ -5,7 +5,7 @@ import (
 	"runtime"
 	"strings"
 
-	"github.com/blang/semver/v4"
+	mmsemver "github.com/Masterminds/semver/v3"
 	genericversion "k8s.io/apimachinery/pkg/version"
 )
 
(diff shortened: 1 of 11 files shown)


azych:
It was brought up that there are changes in how `github.com/blang/semver/v4` and `github.com/Masterminds/semver/v3` handles version ranges.

Because OLM v0 used `github.com/blang/semver/v4`, package+v1 bundles and their authors might still rely on those specific differences in handling. This means that for as long as package+v1 bundles are fully supported by OLM v1 (or we can confirm it otherwise somehow), `github.com/blang/semver/v4` needs to remain alongside any other semver handling lib to ensure version ranges are handled in a compatible and expected way.
For additional context, see [OLM v1 Version Ranges RFC](https://docs.google.com/document/d/1Cf4vw6kdadNp3bHFxh5AMcq1xq0vsEUgC-JNEaVtU6g/edit?tab=t.0#heading=h.x3tfh25grvnv).

I am closing this PR and will be adding comments with context to the code in a separate one.

@joelanford @bentito @thetechnick

Issue #1567, opened 2024-12-19
[tech debt] migrate catalogd command handling to cobra


grokspawn:
Maintenance/update of https://github.com/operator-framework/catalogd/blob/main/cmd/manager/main.go flag handling would be greatly simplified by adopting https://github.com/spf13/cobra

LalatenduMohanty:
https://github.com/operator-framework/operator-controller/pull/1598 fixed this.

PR #1598, opened 2025-01-10
 ✨ Migrate Command Handling to Cobra for Simplified Flag Management

Horiodino:
### **This PR fixes the issue described in operator-framework/operator-controller#1567.**

---
### **Description:**

This PR migrates the command-line handling logic in the `cmd/manager/main.go` file to utilize the [[Cobra CLI framework](https://github.com/spf13/cobra)]. The migration simplifies flag handling, improves readability, and enhances the extensibility of the CLI for future development.

---

### **Key Changes:**

1. **Cobra Root Command**:
   - Introduced the `rootCmd` for top-level command handling, providing a clean and standardized entry point for the CLI.

2. **Flag Migration**:
   - Migrated all existing flags from the standard `flag` package to Cobra’s `cmd.Flags()` and `cmd.PersistentFlags()`.

3. **Subcommand Restructuring**:
   - Refactored and restructured any existing subcommands under Cobra’s subcommand system.

4. **Helper Functions**:
   - Refined the main function by extracting logic into dedicated helper functions.

---

### DIFF
--- a/catalogd/cmd/catalogd/main.go
+++ b/catalogd/cmd/catalogd/main.go
@@ -30,7 +30,7 @@ import (
 
 	"github.com/containers/image/v5/types"
 	"github.com/sirupsen/logrus"
-	"github.com/spf13/pflag"
+	"github.com/spf13/cobra"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	k8slabels "k8s.io/apimachinery/pkg/labels"
@@ -71,6 +71,7 @@ import (
 var (
 	scheme   = runtime.NewScheme()
 	setupLog = ctrl.Log.WithName("setup")
+	cfg      = &config{}
 )
 
 const (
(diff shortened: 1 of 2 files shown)


kevinrizza:
👋 Thanks for the contribution! Quick question, if we're going to do this for the catalogd binary, should we do the same for the operator-controller binary? I.e. https://github.com/operator-framework/operator-controller/blob/main/cmd/operator-controller/main.go

I suspect there will be some opinions on the functional refactoring as the controllers are set up (isn't some of that handled by controller runtime and kubebuilder?) and it would probably be good for us to match in those places. I know that's outside the scope of that github issue, but I think that's an artifact of these two binaries being part of separate repositories until very recently.

grokspawn:
Hey @Horiodino!  Thanks for your PR (especially since you noticed catalogd's new home and reproduced it here).  
Could you do a `make verify` in your local and commit the dependency updates please?

catalogd/cmd/catalogd/main.go
```diff
@@ -74,144 +73,232 @@ const (
...
-
-	// Combine both flagsets and parse them
-	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
-	features.CatalogdFeatureGate.AddFlag(pflag.CommandLine)
-	pflag.Parse()
-
-	if catalogdVersion {
-		fmt.Printf("%#v\n", version.Version())
-		os.Exit(0)
+	cfg := &config{}
```
joelanford, line 99: Nit: since `config` is used entirely inside `newRootCmd`, it seems like we could declare it there instead?

catalogd/cmd/catalogd/main.go
```diff
@@ -74,144 +73,232 @@ const (
...
+func newRootCmd(cfg *config) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "catalogd",
+		Short: "Catalogd is a Kubernetes operator for managing operator catalogs",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return run(cfg)
+		},
 	}
 
+	flags := cmd.PersistentFlags()
```
joelanford, line 116: Persistent flags apply to all subcommands of this command. Since this command doesn't have any subcommands, I think it would make more sense to use `cmd.Flags()` here.

catalogd/cmd/catalogd/main.go
```diff
@@ -74,144 +73,232 @@ const (
...
+	flags.StringVar(&cfg.cacheDir, "cache-dir", "/var/cache/", "Directory for file based caching")
+	flags.DurationVar(&cfg.gcInterval, "gc-interval", 12*time.Hour, "Garbage collection interval")
+	flags.StringVar(&cfg.certFile, "tls-cert", "", "Certificate file for TLS")
+	flags.StringVar(&cfg.keyFile, "tls-key", "", "Key file for TLS")
+	flags.IntVar(&cfg.webhookPort, "webhook-server-port", 9443, "Webhook server port")
+	flags.StringVar(&cfg.caCertDir, "ca-certs-dir", "", "Directory of CA certificates")
+	flags.StringVar(&cfg.globalPullSecret, "global-pull-secret", "", "Global pull secret (<namespace>/<name>)")
+
+	cmd.AddCommand(newVersionCmd())
+	klog.InitFlags(nil)
```
joelanford, line 133: Maybe it is just me, but I prefer explicitly passing `flag.CommandLine` here instead of relying on the defaulting that happens within `InitFlags`.

catalogd/cmd/catalogd/main.go
```diff
@@ -74,144 +73,232 @@ const (
...
+
+	if err := setupGarbageCollector(mgr, k8sCfg, cfg, unpackCacheBasePath); err != nil {
+		return err
+	}
+
+	if err := setupWebhook(mgr); err != nil {
+		return err
+	}
+
+	if err := mgr.Start(ctrl.SetupSignalHandler()); err != nil {
```
joelanford, line 246: Might be nice to call `ctrl.SetupSignalHandler()` from the inlined `Run` command, and then update this `run` function to accept a `context.Context`.

catalogd/cmd/catalogd/main.go
```diff
@@ -74,144 +73,232 @@ const (
...
+	}
+}
+
+func newRootCmd(cfg *config) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "catalogd",
+		Short: "Catalogd is a Kubernetes operator for managing operator catalogs",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return run(cfg)
+		},
```
joelanford, line 113: I've never been a fan of the `RunE` abstraction. In almost all cases, there are two steps to a CLI program that need different error handling
1. Parse and validate the CLI args, env, etc. If a failure occurs here, report the error and print the usage
2. Execute the command. If a failure occurs here, report the error, but DO NOT print the usage.

WDYT about some tweaks to align with that?

I don't recall the exact behavior of `RunE`, but maybe we'd get reasonable behavior for now if we set `cmd.SilenceUsage`, so that only the returned errors strings are ever output (but never usage)?

Horiodino: the `run` function serves as the main entry point, where all other components are initialized and invoked. Including an error statements.The  cobra `RunE`  returns err,ensures proper error handling and clarity.

catalogd/cmd/catalogd/main.go
```diff
@@ -147,7 +153,7 @@ func newVersionCmd() *cobra.Command {
 	}
 }
 
-func run(cfg *config) error {
+func run(cfg *config, ctx context.Context) error {
```
joelanford, line 156: By convention `context.Context` should always be the first argument in the function signature.

catalogd/cmd/catalogd/main.go
LalatenduMohanty: @joelanford Do you mean you would like to have a `func init() {` for the flags?  More info at https://github.com/spf13/cobra/blob/main/site/content/user_guide.md

LalatenduMohanty: I would prefer if we can move the flags to init() method. Because it is easier to maintain in long term.

LalatenduMohanty (dismissed):

catalogd/cmd/catalogd/main.go
Horiodino: Could someone please give a final review so we can make the necessary changes? Everyone seems to have different opinions.

grokspawn (approved):

grokspawn:
Thanks @Horiodino!  We can make additional adjustments in subsequent PRs, but I'd like to keep momentum in this area. I think this achieves the original goals, so approving it.

grokspawn:
Hi @Horiodino .  Sorry about the delays here, but it looks like this PR needs a rebase.  Can you take care of that?

Horiodino:
Folks, please! I've already rebased five times. If it's not good, I'll close the PR—but please don’t leave it hanging.

grokspawn:
Sorry @Horiodino.  We've had a bunch of motion in this project very recently, as we've consolidated multiple repositories.  
I'll check this PR for issues this evening and periodically tomorrow, US Central Standard time.
I've enabled auto-merge so that if it passes CI then it'll go in immediately.

tmshort:
I rebased this and consolidated the code structure to match #1717

grokspawn (approved):

PR #1600, opened 2025-01-13
:book: chore: doc comment the need to keep using blang for version ranges

azych:
Adds doc comment about the need to keep using `github.com/blang/semver` library for handling version ranges.
Follow up from https://github.com/operator-framework/operator-controller/pull/1565 

## Reviewer Checklist

- [ ] API Go Documentation


perdasilva (dismissed):
<3 thank you!

camilamacedo86:
/hold

Just a small note before we proceed:
We won't be able to fully support the registry+v1 bundles due to fundamental differences in concepts and design choices. We've outlined the primary limitations in detail here: [OLMv1 Limitations](https://operator-framework.github.io/operator-controller/project/olmv1_limitations/).

azych:
> /hold
> 
> Hi @azych and @perdasilva
> 
> Sorry, I remove this one from the Queue.
> 
> Just a small note before we proceed:
> 
> We won't be able to fully support the registry+v1 bundles due to fundamental differences in concepts and design choices. We've outlined the primary limitations in detail here: [OLMv1 Limitations](https://operator-framework.github.io/operator-controller/project/olmv1_limitations/).
> 
> So we cannot add **`Because OLM v1 fully supports registry+v1 bundles`** (it is not true and we never will be to support all features of OLM V0, with registry+v1 bundles). Can you please refine this one?

Would `Because OLM v1 supports registry+v1 bundles` be alright?

internal/catalogmetadata/filter/successors.go
```diff
@@ -60,6 +60,11 @@ func legacySuccessor(installedBundle ocv1.BundleMetadata, channels ...declcfg.Ch
 			}
 		}
 		if candidateBundleEntry.SkipRange != "" {
+			// There are differences between how "github.com/blang/semver/v4" and "github.com/Masterminds/semver/v3"
+			// handle version ranges. OLM v0 used blang and there might still be registry+v1 bundles that rely
+			// on those specific differences. Because OLM v1 fully supports registry+v1 bundles,
```
perdasilva, line 65:
```suggestion
			// on those specific differences. Because OLM v1 supports a sub-set of registry+v1 bundles,
```

azych: I think people might wonder what sub-set would that be exactly and if we already only support a sub-set, then why not the subset of bundles that expects versioning handled the way 'masterminds' and not 'blang' handles it?

Additional ideas:
1. ~"Because OLM v1 fully supports registry+v1 bundles versioning" - I think that's true and is the specific reason we are keeping 'blang' as I understood.
2. add ~"limited" and link to the limitations @camilamacedo86 linked to

Open to additional opinions of course, but I don't think we should be overthinking this here :)

camilamacedo86: Anyway that we do not write that we fully support in OLM v1 is OK for me 👍

azych: I went with just removing 'fully' - the remaining statement is true and basically taken verbatim from out internal discussion

While 'sub-set' might be good to add in other contexts, it didin't make sense for me in this one - it would essentially be something like (paraphrasing) "because olm v1 supports a sub-set of bundles, we need to keep using blang", which to me would imply that if we didin't use blang we would be supporting not a sub-set but a full range of bundles.

azych:
@camilamacedo86 @perdasilva 

PR should be good to go, can you remove the hold and merge it?

camilamacedo86 (approved):
Hi @azych 


I would prefer if we added support for a "superset of" as suggested by @perdasilva, since it seems more accurate and aligns with the terminology we use in the documentation and other places.

That said, since we're no longer claiming full support for OLM v1 here, I’m okay with the current approach.

Just to share, the registry-v1 for OLM v1 is, in some way, deprecated. Until we are able to move with the Roadmap, it will remain, but we’re still moving forward in directions like adopting HelmChart. There’s a possibility of designing a registry-v2 in the future. This new design would aim to address the issues identified with registry-v1 and be more suited for OLM v1.

/lgtm
/approve

PR #1618, opened 2025-01-15
:seedling: add vendor dir to .gitignore

perdasilva:
# Description

Adds `vendor` dir to .gitignore to avoid adding them by mistake

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,9 +23,8 @@ operator-controller.yaml
 install.sh
 catalogd.yaml
 
-# Kubernetes Generated files - skip generated files, except for vendored files
-
-!vendor/**/zz_generated.*
+# vendored files
+vendor/
 
 # editor and IDE paraphernalia
 .idea/


.gitignore
```diff
@@ -25,6 +25,7 @@ catalogd.yaml
 
 # Kubernetes Generated files - skip generated files, except for vendored files
 
+vendor/
 !vendor/**/zz_generated.*
```
perdasilva, line 29: Y'all reckon we need to keep this?  Won't it be generated anyway?

azych (approved):

.gitignore
camilamacedo86: We can test it out to see if I do not know this one.
I will defer this to others

joelanford: I don't think we need to keep `!vendor/**/zz_generated.*`. Looks like something that was cargo-culled from somewhere else.

camilamacedo86: I do not see why we would as well.
So, if we have not an historical context here

/lgtm
/approve

joelanford (approved):

Issue #1550, opened 2025-01-07
 Investigate Flakiness in Post-Upgrade Test Timing


camilamacedo86:
The post-upgrade E2E test at [post_upgrade_test.go#L101-L112](https://github.com/operator-framework/operator-controller/blob/6755e94463650f8ac719736a8143c8f20eae0d1b/test/upgrade-e2e/post_upgrade_test.go#L101-L112) appears to be experiencing flakiness due to timing issues. Currently, the test has a 1-minute timeout, which should theoretically be sufficient for the operations it validates. However, under certain conditions, it seems the test may fail due to delays in reconciliation or related operations.

## Expected Behavior
The test should complete successfully within the allocated 1-minute timeout.

## Observed Behavior
The test sometimes fails, likely due to timing issues during the post-upgrade process.

## Analysis
Without additional context, it appears that the 1-minute timeout should be sufficient to cover the following steps:
1. **Catalog cache repopulation** (potentially async, initiated after OLM upgrade).
2. **Reconciliation of the cluster extension ** may involve exponential backoff while the catalog cache repopulates.
3. **Upgrade process**:
   - Resolving and running the upgrade after cache repopulation.
   - Dry-running the upgrade to generate the desired manifest.
   - Running preflight checks.
   - If the checks pass, the actual upgrade will be run (similar to a `helm upgrade` without waiting for object health/readiness).

A potential root cause is hitting exponential backoff on cache repopulation, where an extra reconciliation attempt increases the backoff time, possibly exceeding the 1-minute threshold.

## Steps to Reproduce
1.  Push a PR to trigger the test
2. Before finishing, push another change. Then, when more than 1 GitHub action is running, the tests often fail, and we can check the flake. 

## Suggested Next Steps
- Investigate if exponential backoff during cache repopulation is causing the delay.
- Add logging to identify where the test is spending time and which step(s) contribute most to the timeout.
- Consider increasing the timeout to accommodate edge cases or optimizing the underlying operations.
- Determine if additional retries or adjustments to reconciliation backoff logic are needed.
- Check if GitHub Actions share the resources across the many VMs created (Maybe)

More info: https://github.com/operator-framework/operator-controller/pull/1548

perdasilva:
[...]

PR #1619, opened 2025-01-15
:bug: increase eventually timeout to reduce false negatives

perdasilva:
# Description

I've been noticing a few false negatives on the upgrade-e2e. Increasing the eventually timeout to see if that helps reduce the incidence.

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/test/upgrade-e2e/post_upgrade_test.go
+++ b/test/upgrade-e2e/post_upgrade_test.go
@@ -22,6 +22,11 @@ import (
 	catalogd "github.com/operator-framework/operator-controller/catalogd/api/v1"
 )
 
+const (
+	eventuallyTimeout    = 3 * time.Minute
+	eventuallyPollPeriod = time.Second
+)
+
 func TestClusterExtensionAfterOLMUpgrade(t *testing.T) {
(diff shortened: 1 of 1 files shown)


perdasilva:
> From what I can see there was actually already a PR created by @camilamacedo86 to do this but it was closed because the consensus was to do a deeper investigation and try to fix the underlying problem. This is the issue which also links to previous PR - #1550
> 
> Having said that, I notice that those failures are getting more and more common, so from my POV it might be a good idea to have the timeout increased until proper fix is in place.

Thanks for that! Let's see what the team reckons. Maybe we could create a tracking issue, though then we lose the forcing function to RC...

perdasilva:
Moving to draft to block merging until we have a wider discussion with the community/team

camilamacedo86:
Hi @perdasilva 

Yep, we have the issue for it already: https://github.com/operator-framework/operator-controller/issues/1550
The only test that fails is this one. The issue only occurs if we have the e2e test running for more than 1 github action at the time. 

So, I would suggest we close this one until the analysis/issue be addressed. WDYT?

Issue #1456, opened 2024-11-13
Constant reconciliation due to changing error message on CRD validation


m1kola:
**Steps to reproduce**

1. Create pre-requisites such as RBAC and service account. [This](https://github.com/operator-framework/operator-controller/blob/6d79092c8921f888b125e084b494576e87d5706e/config/samples/olm_v1_clusterextension.yaml#L1-L268) can be used as a reference.
2. Deploy `argocd-operator` of version `0.5.0` from `operatorhubio` catalog.
    ```yaml
    apiVersion: olm.operatorframework.io/v1
    kind: ClusterExtension
    metadata:
      name: argocd
    spec:
      namespace: argocd
      serviceAccount:
        name: argocd-installer
      source:
        sourceType: Catalog
        catalog:
          packageName: argocd-operator
          version: 0.5.0
    ```
3. Upgrade to 0.6.0
    ```bash
    kubectl patch clusterextension argocd --type='merge' -p '{"spec": {"source": {"catalog": {"version": "0.6.0"}}}}'
    ```
4. Watch for changes:
    ```bash
    kubectl get clusterextension argocd -o yaml -w
    ```

**Result**

Error message seems to be ordered differently on each reconciliation which probably results in useless reconciliation runs.

E.g. I see the condition alternating few variations such as:

```yaml
  - lastTransitionTime: "2024-11-13T15:48:14Z"
    message: |-
      validating upgrade for CRD "argocds.argoproj.io" failed: CustomResourceDefinition argocds.argoproj.io failed upgrade safety validation. "ChangeValidator" validation failed: version "v1alpha1", field "^.status.dex" has unknown change, refusing to determine that change is safe
      version "v1alpha1", field "^.status.redis" has unknown change, refusing to determine that change is safe
      version "v1alpha1", field "^.status.phase" has unknown change, refusing to determine that change is safe
      version "v1alpha1", field "^.status.notificationsController" has unknown change, refusing to determine that change is safe
      version "v1alpha1", field "^.status.server" has unknown change, refusing to determine that change is safe
      version "v1alpha1", field "^.status.repo" has unknown change, refusing to determine that change is safe
      version "v1alpha1", field "^.status.ssoConfig" has unknown change, refusing to determine that change is safe
      version "v1alpha1", field "^.status.applicationController" has unknown change, refusing to determine that change is safe
[...]

PR #1640, opened 2025-01-20
:bug: add orderKappsValidateErr in crdupgradesafety preflight

azych:
orderKappsValidateErr() is meant as a temporary solution to an external (ie. dependency) problem. carvel.dev/kapp/pkg/kapp/crdupgradesafety Validate() can return a multi-line error message which comes in random order. Until that is changed upstream, we need to fix this on our side to avoid falling into cycle of constantly trying to reconcile ClusterExtension's status due to random error message we set in its conditions.

An upstream PR is already created, but it might be some time before it makes it into the carvel's codebase and then to ours, hence this change. 

For full context please see https://github.com/operator-framework/operator-controller/issues/1456 and https://github.com/carvel-dev/kapp/pull/1047, especially https://github.com/carvel-dev/kapp/pull/1047#discussion_r1922017781

# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/rukpak/preflights/crdupgradesafety/crdupgradesafety.go
+++ b/internal/rukpak/preflights/crdupgradesafety/crdupgradesafety.go
@@ -1,9 +1,11 @@
 package crdupgradesafety
 
 import (
+	"cmp"
 	"context"
 	"errors"
 	"fmt"
+	"slices"
 	"strings"
 
 	kappcus "carvel.dev/kapp/pkg/kapp/crdupgradesafety"
@@ -84,7 +86,7 @@ func (p *Preflight) runPreflight(ctx context.Context, rel *release.Release) erro
 		return fmt.Errorf("parsing release %q objects: %w", rel.Name, err)
 	}
 
-	validateErrors := []error{}
+	validateErrors := make([]error, 0, len(relObjects))
 	for _, obj := range relObjects {
 		if obj.GetObjectKind().GroupVersionKind() != apiextensionsv1.SchemeGroupVersion.WithKind("CustomResourceDefinition") {
 			continue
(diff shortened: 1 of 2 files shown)


internal/rukpak/preflights/crdupgradesafety/crdupgradesafety.go
```diff
@@ -112,9 +114,71 @@ func (p *Preflight) runPreflight(ctx context.Context, rel *release.Release) erro
...
+// This helper attempts to do that and falls back to original unchanged error message
+// in case of any unforeseen issues which likely mean that the internals of validator.Validate
+// have changed.
+//
+// For full context see:
+// github.com/operator-framework/operator-controller/issues/1456 (original issue and comments)
+// github.com/carvel-dev/kapp/pull/1047 (PR to ensure order in upstream)
+//
+// TODO: remove this once ordering has been handled by the upstream.
+func orderKappsValidateErr(err error) error {
```
perdasilva, line 143: Well written (as with the unit tests). Thank you! Qq: is it worth surfacing whether the function had to fallback on the original? My guess is that if we fallback, we may run into the continuous reconcile issue. Wondering if we should make that noisy somehow...?

azych: thank you @perdasilva I thought about it too and it's a tricky situation, because while the first two steps in the unwrapping logic are true to all internal validators (as of now at least), the last one (most nested error.Join) might only be valid for `CheckValidator` internal validator. Since fallback might as well be a false negative here, I'd normally log this via `Debug`, but I noticed `Debug` isn't being used anywhere in the code so ultimately thought against it.
In case of a non-empty error here and regardless of fallback, we are still logging the full reconciliation error message every time we try to reconcile and that should make spotting the random order possible, though I agree an added hint from failed fallback might definitely help, so maybe it would be ok.

WDYT?

perdasilva: yeah. Ultimately, I guess, it becomes a debug artifice. Which, while it undoubtedly helps with debugging, we'd still be reconciling repeatedly. If we have exp. backoff, then maybe it's not a big issue, and it's also temporary. I was thinking if we should panic (since the library shouldn't change out from under us during runtime), but then we'd need to be very careful about testing each error condition the underlying library checks for, which might not be a good use of time atm. Let's proceed ^^ thanks, dude =D

perdasilva (approved):

PR #1643, opened 2025-01-21
⚠ (feat) Introduce new feature-gated query endpoint 

anik120:
# Description

Implementation of [Query Endpoint API RFC](https://docs.google.com/document/d/1s6_9IFEKGQLNh3ueH7SF4Yrx4PW9NSiNFqFIJx0pU-8/edit?usp=sharing)

- Adds a new feature-gated endpoint, that can be enabled by passing the flag `--feature-gates=APIV1QueryHandler=true` to the binary 
- Once the feature gate is enabled, separate indexes are created while storing unpacked data from a catalog image, besides the existing `all` data. 
- Query endpoint can be queried at `<base.url>/api/vi/query?` for each catalog 
- Server uses the indexes created previously to serve the query endpoint to send a filtered list of FBC meta blobs.
  

## Reviewer Checklist

- [ ] API Go Documentation


catalogd/internal/storage/multireadseeker.go
```diff
@@ -0,0 +1,118 @@
+package storage
```
joelanford, line 1: Oh, we should discuss this. I added this because the `http` package documentation recommended using `http.ServeContent` over `io.Copy` for a number of reasons, and `http.ServeContent` needs an `io.ReadSeeker`. The main benefit for us would be support for the `Range` http headers.

I found this multiReadSeaker implementation in a Golang issue proposing that it be added to the standard library, so there may be license concerns. We may need to implement our own or look into implementing the range functionality another way.

catalogd/internal/serverutil/serverutil.go
```diff
@@ -42,17 +45,22 @@ func AddCatalogServerToManager(mgr ctrl.Manager, cfg CatalogServerConfig, tlsFil
...
 		Name:                "catalogs",
 		OnlyServeWhenLeader: true,
 		Server: &http.Server{
-			Addr:        cfg.CatalogAddr,
-			Handler:     handler,
+			Addr:    cfg.CatalogAddr,
+			Handler: handler,
+			BaseContext: func(_ net.Listener) context.Context {
+				return log.IntoContext(context.Background(), mgr.GetLogger().WithName("http.catalogs"))
+			},
```
joelanford, line 63: I _think_ we can either drop this whole `BaseContext` or alternately, simply `return l`

joelanford: Either way, I don't think the code uses the logger from the context. This was a holdover from the way that the original logging middleware worked.

catalogd/internal/serverutil/serverutil_tests.go
```diff
@@ -0,0 +1,254 @@
...
+// 	setupStore: func() error {
+// 		return writeFile(filepath.Join(store.RootDir, "test-catalog2", "catalog.jsonl"), []byte(`{"foo":"bar"}`), 0600)
+// 	},
+// 	expectStatusOK:  true,
+// 	expectedContent: `{"foo":"bar"}`,
+// 	URLPath:         "/catalogs/test-catalog2/api/v1/all",
+// },
+
+// by default the compressor will only trigger for files larger than 1400 bytes
+const testCompressableJSON = `{
```
joelanford, line 76: If all this is stuff is to test the gzip handler in isolation, I think we can simplify.

If the goal is to test the entire stack of handlers, I figure there's a better way to organize and test to reduce duplication and fragility. For example, if we make the `CatalogServerConfig` accept an arbitrary `http.Handler` instead of a `storage.Instance`, then we can:
- test the `storage.Instance` handler in isolation
- test the `serverutil` setup with dummy handlers that are easier to manipulate for test purposes.

catalogd/internal/serverutil/serverutil_tests.go
```diff
@@ -0,0 +1,254 @@
...
+// 			return err
+// 		}
+// 		return err
+// 	},
+// },
+
+// {
+// 	name: "Ignores accept-encoding for the path /catalogs/test-catalog/api/v1/all with size < 1400 bytes",
+// 	setupStore: func() error {
+// 		return writeFile(filepath.Join(store.RootDir, "test-catalog2", "catalog.jsonl"), []byte(`{"foo":"bar"}`), 0600)
```
joelanford, line 68: I know existing tests violate this rule, but one thing I think I would fix is to avoid writing to the underlying filesystem, because it is pretty fragile. I think we can probably test as a black box:
1. Check `ContentExists(foo)` is false
1. `Store(foo)`
2. Query various endpoints with positive and negative tests for `foo`, `bar` catalogs, but also just different path prefixes, etc.
3. Check `ContentExists(foo)` is true
4. `Delete(foo)`
5. Query `foo` endpoints and expect 404
6. Check `ContentExists(foo)` is false.

azych:
What's the relation of this PR to https://github.com/operator-framework/operator-controller/pull/1642?
This one seems to include two additional commits, with the first 4 being the same. 

Which PR is a duplicate and which should be reviewed?
If this is the one to review (and also not a Draft), can you add a description of the main changes this introduces?

catalogd/internal/serverutil/serverutil_tests.go
anik120: Or, the other relevant question is, do we need to test the gzip handler explicitly in our unit tests? We're importing `"github.com/klauspost/compress/gzhttp"` to provided compressed content via our server. Are unit tests that say "tests if the library is doing its job" really that useful?

catalogd/internal/serverutil/serverutil_tests.go
anik120: +1, I've added this test in `localdir_test.go`, and avoided writing directly to the file system (using `store.Store` instead)

anik120:
> What's the relation of this PR to #1642? This one seems to include two additional commits, with the first 4 being the same.
> 
> Which PR is a duplicate and which should be reviewed? If this is the one to review (and also not a Draft), can you add a description of the main changes this introduces?

@azych apologies for the confusion. You're just seeing the result of rapid experimentation. This is the one to review, but it's still a WIP. I've converted this PR to a draft.

catalogd/internal/storage/localdir_test.go
```diff
@@ -13,216 +11,384 @@ import (
...
+		// 		if err != nil {
+		// 			t.Fatal(err)
+		// 		}
+
+		// 		if !s.ContentExists("test-catalog") {
+		// 			t.Error("content should exist after store")
+		// 		}
+
+		// 		// Verify index file was created
+		// 		indexPath := filepath.Join(s.RootDir, "test-catalog", "index.json")
```
joelanford, line 90: I would avoid the hardcoded path because it is pretty fragile. Maybe use:
```
indexPath := catalogIndexFilePath(s.catalogDir("test-catalog"))
```

catalogd/internal/serverutil/serverutil_tests.go
grokspawn: IMHO, any external interaction is a good test.  And unit tests are better than e2e (smaller, faster, usually more tightly-scoped).
We /were/ testing to ensure that we would respond accordingly for `accept-encoding` request headers that could control/enable this functionality.

catalogd/internal/serverutil/serverutil.go
anik120: Dropped it.

catalogd/internal/storage/multireadseeker.go
anik120: Discussion topic for when we get on a call tomorrow: why deal with all of these ourselves, when others have already done it for us? 

https://github.com/operator-framework/operator-controller/pull/1654

catalogd/internal/storage/localdir.go
```diff
@@ -20,95 +26,266 @@ import (
...
+	}
+	idx, err := s.getIndex(catalog)
+	if err != nil {
+		httpError(w, err)
+		return
+	}
+	indexReader, ok := idx.Get(catalogFile, schema, pkg, name)
+	if !ok {
+		httpError(w, fs.ErrNotExist)
+		return
```
joelanford, line 237: This should change to write a 200 status and write no content to the response body.

catalogd/internal/storage/multireadseeker.go
joelanford: Let's just delete this file, stop wrapping the query readers with this, and use `io.Copy` instead of `http.ServeContent`

catalogd/internal/storage/localdir.go
```diff
@@ -20,95 +26,266 @@ import (
...
+	http.Error(w, fmt.Sprintf("%d %s", code, http.StatusText(code)), code)
+}
+
+func serveJsonLines(w http.ResponseWriter, r *http.Request, modTime time.Time, rs io.ReadSeeker) {
+	w.Header().Add("Content-Type", "application/jsonl")
+	http.ServeContent(w, r, "", modTime, rs)
+}
+
+func (s *LocalDirV1) getIndex(catalog string) (*index, error) {
+	idx, err, _ := s.sf.Do(catalog, func() (interface{}, error) {
```
joelanford, line 275: Let's add some comments here about the purpose of the single flight structure.

catalogd/internal/storage/localdir.go
```diff
@@ -20,95 +26,266 @@ import (
...
-	defer os.Remove(tempFile.Name())
-	if err := declcfg.WalkMetasFS(ctx, fsys, func(path string, meta *declcfg.Meta, err error) error {
+	defer os.RemoveAll(tmpCatalogDir)
+
+	storeMetaFuncs := []storeMetasFunc{storeCatalogData}
+	if s.EnableQueryHandler {
+		storeMetaFuncs = append(storeMetaFuncs, storeIndexData)
+	}
+
+	var (
```
joelanford, line 60: Throughout these lines where we setup and run concurrent channel writers and readers, we should add some comments to explain what is happening.

catalogd/internal/storage/localdir.go
```diff
@@ -20,95 +26,266 @@ import (
...
+	default:
+		code = http.StatusInternalServerError
+	}
+	http.Error(w, fmt.Sprintf("%d %s", code, http.StatusText(code)), code)
+}
+
+func serveJsonLines(w http.ResponseWriter, r *http.Request, modTime time.Time, rs io.ReadSeeker) {
+	w.Header().Add("Content-Type", "application/jsonl")
+	http.ServeContent(w, r, "", modTime, rs)
+}
```
joelanford, line 272: With my suggestion of backing away from the multi-read-seeker code, we'll have to have a different version of this function that the query handler can use (the reader returned by the index will no longer be a io.ReadSeeker, it'll just be a reader)

Unfortunately, if we don't use `http.ServeContent` for the query handler, we also do not get the `Last-Modified`/`If-Modified-Since` header handling. Does the standard library have a way for us to implement this caching logic so that we avoid implementing bugs in our own implementation? If we copy code from the Go stdlib, let's just make sure it's license allows that and we do whatever attribution is necessary.

catalogd/internal/serverutil/serverutil_test.go
```diff
@@ -0,0 +1,254 @@
...
+  `
+
+// mockStorageInstance implements storage.Instance interface for testing
+type mockStorageInstance struct {
+	content string
+}
+
+func (m *mockStorageInstance) StorageServerHandler() http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Write([]byte(m.content))
```
joelanford, line 237: Do we actually care what this content is, or do we just care about its length? If the latter, I'd suggest removing the `testCompressableJSON` variable and replacing with with something like `strings.Rand(sizeThreshold+1)`

catalogd/internal/storage/index.go
```diff
@@ -0,0 +1,133 @@
...
+	"fmt"
+	"io"
+	"slices"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+
+	"github.com/operator-framework/operator-registry/alpha/declcfg"
+)
+
+type index struct {
```
joelanford, line 15: We should add some GoDoc to `index` to help our future selves understand how the indexing works.
```suggestion
// index is a simple index of sections of an FBC file used to lookup FBC blobs that
// match any combination of their schema, package, and name fields. A section is the
// byte offset and length of an FBC blob within the file.
//
// This index strikes a balance between space and performance. It indexes each field
// separately, and performs logical set intersections at lookup time in order to implement
// a multi-parameter query.
//
// Note: it is permissible to change the indexing algorithm later if it is necessary to
// tune the space / performance tradeoff. However care should be taken to ensure
// that the actual content returned by the index remains identical, as users of the index
// may be sensitive to differences introduced by index algorithm changes (e.g. if the
// order of the returned sections changes).

type index struct {
```

catalogd/internal/storage/index.go
```diff
@@ -0,0 +1,133 @@
...
+
+	srs := make([]io.ReadSeeker, 0, len(sections))
+	for _, s := range sections {
+		sr := io.NewSectionReader(r, s.offset, s.length)
+		srs = append(srs, sr)
+	}
+	return newMultiReadSeeker(srs...), true
+}
+
+func (i *index) getSectionSet(schema, packageName, name string) sets.Set[section] {
```
joelanford, line 70: One thing that occurs to me is that the indexed sections are always guaranteed to be sorted by the offset. We could likely take advantage of that fact and implement a more performant algorithm that can step directly through each slice rather than building a set for each slice, performing a set intersection, the converting back into a slice and resorting.

But this is something we can capture as a follow-up.

catalogd/internal/storage/localdir.go
```diff
@@ -20,95 +26,266 @@ import (
...
+
+func httpError(w http.ResponseWriter, err error) {
+	var code int
+	switch {
+	case errors.Is(err, fs.ErrNotExist):
+		code = http.StatusNotFound
+	case errors.Is(err, fs.ErrPermission):
+		code = http.StatusForbidden
+	case errors.Is(err, ErrInvalidParams):
+		code = http.StatusBadRequest
```
joelanford, line 262: I don't see any calls to `httpError`, where we pass `ErrInvalidParams` in. But I think this is definitely something we should keep. Just need to make sure we actually check for unknown parameters pass it in.

catalogd/internal/storage/localdir.go
```diff
@@ -20,95 +26,266 @@ import (
...
+	m  sync.RWMutex
+	sf singleflight.Group
 }
 
-const (
-	v1ApiPath = "api/v1"
-	v1ApiData = "all"
+var (
+	_                Instance = &LocalDirV1{}
+	ErrInvalidParams          = errors.New("invalid parameters")
```
joelanford, line 39: Since this is only used inside this package, it should be unexported:
```suggestion
	errInvalidParams          = errors.New("invalid parameters")
```

catalogd/internal/storage/localdir.go
```diff
@@ -20,95 +26,266 @@ import (
...
+
+	m  sync.RWMutex
+	sf singleflight.Group
 }
 
-const (
-	v1ApiPath = "api/v1"
-	v1ApiData = "all"
+var (
+	_                Instance = &LocalDirV1{}
```
joelanford, line 38: Nit: it is _slightly_ better to do the compile-time interface check this way to avoid an allocation:
```suggestion
	_                Instance = (*LocalDirV1)(nil)
```

catalogd/internal/storage/localdir.go
joelanford: The TL;DR is that we expect `getIndex` to be called concurrently if there are concurrent HTTP requests being handled by the query handler. The single flight instance gives us a way to load the index from disk exactly once per concurrent group of callers, and then let every concurrent caller have access to the loaded index. This avoids lots of unnecessary open/decode/close cycles when concurrent requests are being handled, which improves overall performance and decreases response latency.

catalogd/internal/storage/localdir.go
anik120: Was able to find a way to `http.ServeContent` with `io.Reader` https://github.com/operator-framework/operator-controller/pull/1643/commits/79da4de47b2c40799b9fcea7dec481faf62b04b0

catalogd/internal/storage/localdir.go
```diff
@@ -20,95 +26,353 @@ import (
...
-	typeHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		w.Header().Add("Content-Type", "application/jsonl")
-		gzHandler.ServeHTTP(w, r)
-	})
-	mux.Handle(s.RootURL.Path, typeHandler)
+	mux.HandleFunc(s.RootURL.JoinPath("{catalog}", "api", "v1", "all").Path, s.handleV1All)
+	if s.EnableQueryHandler {
+		mux.HandleFunc(s.RootURL.JoinPath("{catalog}", "api", "v1", "query").Path, s.handleV1Query)
+	}
 	return mux
```
joelanford, line 192: We should wrap this mix with a middleware hander that checks for HEAD or GET.

catalogd/internal/storage/localdir.go
```diff
@@ -20,95 +26,353 @@ import (
...
 }
 
-// Open opens a named file from the underlying filesystem. If the file
-// is not a regular file, it return os.ErrNotExists. Callers are resposible
-// for closing the file returned.
-func (f *filesOnlyFilesystem) Open(name string) (fs.File, error) {
-	file, err := f.FS.Open(name)
+func serveJsonLinesQuery(w http.ResponseWriter, rs io.Reader) {
+	w.Header().Add("Content-Type", "application/jsonl")
+	_, err := io.Copy(w, rs)
```
joelanford, line 286: We should only copy data if it is a GET request. a HEAD request should be identical to GET except without the response body.

catalogd/internal/storage/localdir.go
```diff
@@ -20,95 +26,353 @@ import (
...
+	}
+	http.Error(w, fmt.Sprintf("%d %s", code, http.StatusText(code)), code)
 }
 
-// filesOnlyFilesystem is a file system that can open only regular
-// files from the underlying filesystem. All other file types result
-// in os.ErrNotExists
-type filesOnlyFilesystem struct {
-	FS fs.FS
+func serveJsonLines(w http.ResponseWriter, r *http.Request, modTime time.Time, rs io.ReadSeeker) {
```
joelanford, line 279: Now that we have a separate function to serve data for the query handler, I think we should have the all handler directly call `http.ServeFile` (after explicitly setting the "Content-Type")

catalogd/internal/storage/localdir.go
joelanford: Something like:

```go
allowedMethodsHandler := func(next http.Handler, allowedMethods ...string) http.Handler {
	allowedMethodSet := sets.New[string](allowedMethods...)
	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		if !allowedMethodSet.Has(r.Method) {
			http.Error(w, http.StatusText(http.StatusMethodNotAllowed), http.StatusMethodNotAllowed)
			return
		}
		next.ServeHTTP(w, r)
	})
}
return allowedMethodsHandler(mux, http.MethodGet, http.MethodHead)
```

joelanford: There are other ways to do it too. For example:
```go
mux.HandleFunc("HEAD "+s.RootURL.JoinPath("{catalog}", "api", "v1", "all").Path, s.handleV1All)
mux.HandleFunc("GET "+s.RootURL.JoinPath("{catalog}", "api", "v1", "all").Path, s.handleV1All)
```

And then the same for the query endpoint

catalogd/internal/storage/index.go
anik120: I realized that we'd want robust test cases for the indexing algorithm too, but that too is probably a follow up in the interest of time.

catalogd/internal/storage/localdir.go
anik120: Added the comment in the `LocalDirV1` structure definition

catalogd/internal/storage/localdir.go
anik120: Copied the `checkPreconditions` function and all the code associated with it from the net/http package in the `http_preconditions_check.go` file. I did have to make modifications to the copied code too. Imho it's smells funky to me, it's a lot of copied code that we'll have to hope won't carry any bugs, especially with the modifications I did. But, for now it's in there.

catalogd/internal/storage/localdir.go
anik120: I noticed that `index.Get()` was only ever returning `true` (so this code path was never actually being hit). Something else solved the test case failure though, but I went ahead and cleaned up  `Get()`'s signature https://github.com/operator-framework/operator-controller/pull/1643/commits/53692a97707d315c84ce3b592bc77dcf5fc04003

catalogd/internal/serverutil/serverutil_tests.go
anik120: Added these tests in `serverutil_tests.go`

catalogd/internal/storage/localdir.go
anik120: How about something like https://github.com/operator-framework/operator-controller/pull/1643/commits/046c66c1377af86c55c843993cd32477eb0dd156. We get to retain the benefits of `serverContent` that way.

catalogd/internal/storage/localdir.go
```diff
@@ -2,113 +2,317 @@ package storage
...
+		httpError(w, err)
+		return
+	}
+	defer catalogFile.Close()
+
+	w.Header().Set("Last-Modified", catalogStat.ModTime().UTC().Format(TimeFormat))
+	if checkPreconditions(w, r, catalogStat.ModTime()) {
+		w.WriteHeader(http.StatusNotModified)
+		return
+	}
```
grokspawn, line 239: Isn't this insufficient for last-modified caching, because a former query and this may have been with different parameters, couldn't they? 
For e.g., is this scenario reasonable, where a follow-on query for channels/bundles is erroneously conflated with the package-level predecessor query:
```
A --> B query?schema=olm.package&name=foo
A <-- B response ([]Metas, last-mod-time=$TIME)
A --> B query?package=foo (if-modified-since=$TIME)
A <-- B response (200, empty)
```
?

catalogd/internal/storage/localdir.go
```diff
@@ -2,113 +2,317 @@ package storage
...
+	defer s.m.RUnlock()
+
+	catalog := r.PathValue("catalog")
+	catalogFile, catalogStat, err := s.catalogData(catalog)
+	if err != nil {
+		httpError(w, err)
+		return
+	}
+	defer catalogFile.Close()
+
```
grokspawn, line 234: Think we also need to handle unsupported/unsupportable Range requests here:
```golang
       rtype, haveType := w.Header()["Ranges"]
       if haveType {
               httpError(w, fmt.Errorf("unexpected Range header: %q", rtype))
       }
```

catalogd/internal/storage/localdir.go
grokspawn: Just proved this
1. `curl -vvv -k 'https://localhost:8080/catalogs/operatorhubio/api/v1/query?package=postgres-operator'`
2. (pulled the `Last-Modified` timestamp)
3. replay the query, with the timestamp populated in If-Modified-Since: `curl -vvv -k -H "If-Modified-Since: Mon, 03 Feb 2025 16:04:58 GMT" 'https://localhost:8080/catalogs/operatorhubio/api/v1/query?package=postgres-operator'`  ✅ 
4. run an unrelated query with the same timestamp populated in If-Modified-Since: `curl -vvv -k -H "If-Modified-Since: Mon, 03 Feb 2025 16:04:58 GMT" 'https://localhost:8080/catalogs/operatorhubio/api/v1/query?schema=olm.package&name=postgres-operator'` ❌ : 
```sh
> GET /catalogs/operatorhubio/api/v1/query?schema=olm.package&name=postgres-operator HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/8.7.1
> Accept: */*
> If-Modified-Since: Mon, 03 Feb 2025 16:04:58 GMT
>
* Request completely sent off
< HTTP/1.1 304 Not Modified
< Last-Modified: Mon, 03 Feb 2025 16:04:58 GMT
< Vary: Accept-Encoding
< Date: Mon, 03 Feb 2025 16:14:10 GMT
```

catalogd/internal/storage/localdir.go
grokspawn: To be fair, we do not respond with any kind of `Accept-Range` header, so clients are not supposed to send us `Range` requests, but they might, unless we unconditionally send them a `Accept-Ranges: none` response header.
This can be deferred.  I just wanted to record the note while I was exploring the space.

catalogd/internal/storage/localdir.go
anik120: @grokspawn looks like that's the expected behavior. I wrote this test case using `http.ServeContent` https://go.dev/play/p/isOSaY-cWUO : 

1. query with parms A, gets 200, caches mod-time
2. query with parms B, sets if-mod-since, expects 200, 
3. query with parms B, sets if-mod-since, expects 304

and it fails in 2 because the response is 304. 

Clients are expected to use the "if-modified-since" header with same set of parameters only for subsequent queries.

catalogd/internal/storage/localdir.go
anik120: Created https://github.com/operator-framework/operator-controller/issues/1696 to capture follow up

grokspawn (approved):
All follow ups are being tracked in other issues to try not to distract this effort any further.

catalogd/internal/serverutil/serverutil.go
```diff
@@ -97,3 +93,12 @@ func logrLoggingHandler(l logr.Logger, handler http.Handler) http.Handler {
 		l.Info("handled request", "host", host, "username", username, "method", params.Request.Method, "uri", uri, "protocol", params.Request.Proto, "status", params.StatusCode, "size", params.Size)
 	})
 }
+
+func storageServerHandlerWrapped(l logr.Logger, cfg CatalogServerConfig) http.Handler {
```
joelanford, line 97: Nit (follow-up): We could probably change `CatalogServerConfig` to have an `http.Handler` field instead of the entire `LocalStorage` field?

catalogd/internal/storage/http_precoditions_check.go
```diff
@@ -0,0 +1,226 @@
...
+		buf = remain
+	}
+	return condTrue
+}
+
+// ParseTime parses a time header (such as the Date: header),
+// trying each of the three formats allowed by HTTP/1.1:
+// [TimeFormat], [time.RFC850], and [time.ANSIC].
+// nolint:nonamedreturns
+func ParseTime(text string) (t time.Time, err error) {
```
joelanford, line 162: I would unexpert everything from this copied code.
```suggestion
func parseTime(text string) (t time.Time, err error) {
```

catalogd/internal/storage/index.go
joelanford: We may have a bunch of coverage of this in the higher level tests, so maybe unnecessary.

catalogd/internal/storage/localdir.go
```diff
@@ -2,113 +2,316 @@ package storage
...
+	}
+	return allowedMethodsHandler(mux, http.MethodGet, http.MethodHead)
+}
+
+func (s *LocalDirV1) handleV1All(w http.ResponseWriter, r *http.Request) {
+	s.m.RLock()
+	defer s.m.RUnlock()
+
+	catalog := r.PathValue("catalog")
+	catalogFile, catalogStat, err := s.catalogData(catalog)
```
joelanford, line 213: Need to `defer catalogFile.Close()` here.

Even better, now that we've got separate handling for all and query, maybe we go all the way to simply `http.ServeFile`?

catalogd/internal/storage/localdir.go
```diff
@@ -2,113 +2,316 @@ package storage
...
-	FS fs.FS
+func httpError(w http.ResponseWriter, err error) {
+	var code int
+	switch {
+	case errors.Is(err, fs.ErrNotExist):
+		code = http.StatusNotFound
+	case errors.Is(err, fs.ErrPermission):
+		code = http.StatusForbidden
+	case errors.Is(err, errInvalidParams):
+		code = http.StatusBadRequest
```
joelanford, line 279: I don't think I see any where that we call `httpError` with `errInvalidParams`. Seems we should add a check in the query handler for unexpected parameters?

catalogd/internal/storage/localdir.go
```diff
@@ -2,113 +2,316 @@ package storage
...
+	catalogFile, catalogStat, err := s.catalogData(catalog)
+	if err != nil {
+		httpError(w, err)
+		return
+	}
+	defer catalogFile.Close()
+
+	w.Header().Set("Last-Modified", catalogStat.ModTime().UTC().Format(TimeFormat))
+	if checkPreconditions(w, r, catalogStat.ModTime()) {
+		w.WriteHeader(http.StatusNotModified)
```
joelanford, line 236: It looks like `checkPreconditions` already sets the appropriate status code, so no need to do that here?

(This may be a bug because `checkPreconditions` can also write the `http.StatusPreconditionFailed` status code)

catalogd/internal/storage/localdir_test.go
```diff
@@ -13,219 +10,498 @@ import (
...
+			},
+			test: func(t *testing.T, s *LocalDirV1, fsys fs.FS) {
+				const catalog = "test-catalog"
+				var wg sync.WaitGroup
+
+				// Start multiple concurrent readers
+				for i := 0; i < 10; i++ {
+					wg.Add(1)
+					go func() {
+						defer wg.Add(-1)
```
joelanford, line 109:
```suggestion
						defer wg.Done()
```

catalogd/internal/storage/localdir_test.go
```diff
@@ -13,219 +10,498 @@ import (
...
+	}
+	if store.Store(context.Background(), "test-catalog", createTestFS(t)) != nil {
+		t.Fatal("failed to store test catalog")
+	}
+	testServer := httptest.NewServer(store.StorageServerHandler())
+
+	testCases := []struct {
+		name               string
+		setupStore         func() (*httptest.Server, error)
+		queryParams        string
```
joelanford, line 285: Nit: might be nicer to use `url.Values` for more structure setup of the query params.

catalogd/internal/storage/localdir_test.go
```diff
@@ -13,219 +10,498 @@ import (
...
+		expectedStatusCode int
+		expectedContent    string
+	}{
+		{
+			name:               "valid query with package schema",
+			queryParams:        "?schema=olm.package",
+			expectedStatusCode: http.StatusOK,
+			expectedContent:    `{"defaultChannel":"preview_test","name":"webhook_operator_test","schema":"olm.package"}`,
+		},
```
joelanford, line 294: I don't see any test that verifies that a query that matches multiple blobs returns all the blobs.

catalogd/internal/storage/localdir_test.go
```diff
@@ -13,219 +10,498 @@ import (
...
+			name:               "cached response with If-Modified-Since",
+			queryParams:        "?schema=olm.package",
+			expectedStatusCode: http.StatusNotModified,
+			expectedContent:    "",
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			req, err := http.NewRequest(http.MethodGet, fmt.Sprintf("%s/catalogs/test-catalog/api/v1/query%s", testServer.URL, tc.queryParams), nil)
```
joelanford, line 341: We should parameters the http method and test `HEAD` (which should work, but with no response body), as well as `POST` and `DELETE`, and `PUT` which should all fail with the unsupported method status code.

catalogd/internal/storage/localdir_test.go
```diff
@@ -13,219 +10,498 @@ import (
...
+			queryParams:        "?schema=non_existent_schema",
+			expectedStatusCode: http.StatusOK,
+			expectedContent:    "",
+		},
+		{
+			name:               "cached response with If-Modified-Since",
+			queryParams:        "?schema=olm.package",
+			expectedStatusCode: http.StatusNotModified,
+			expectedContent:    "",
+		},
```
joelanford, line 336: Add another test cases with an unknown parameter, expected to return BadRequest

grokspawn: And another when we get duplicate parameters? (i.e. 'query?package=foo&package=foo')

catalogd/internal/storage/index.go
anik120: We do have sufficient coverage with the higher level tests, still writing these tests helped me with grasping the indexing algorithm more https://github.com/operator-framework/operator-controller/pull/1702

ie my argument for the utility of these tests are "better code documentation with unit tests that help understand behavior"

catalogd/internal/storage/http_precoditions_check.go
anik120: Sneaked this in with this bug fix https://github.com/operator-framework/operator-controller/pull/1710

PR #1666, opened 2025-01-29
✨ Add feature gate for preflight permissions

bentito:
# Description
Allows this work to be done behind a feature gate, allowing for testing and for the code to settle in before it starts affecting operations.

Name of the feature gate is, `PreflightPermissions`, see code and tests for usage.

Should be a no-op even if enabled, as there is currently no code behind it.

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/applier/helm.go
+++ b/internal/applier/helm.go
@@ -162,6 +162,7 @@ func (h *Helm) getReleaseState(cl helmclient.ActionInterface, ext *ocv1.ClusterE
 		}, helmclient.AppendInstallPostRenderer(post))
 		if err != nil {
 			if features.OperatorControllerFeatureGate.Enabled(features.PreflightPermissions) {
+				_ = struct{}{} // minimal no-op to satisfy linter
 				// probably need to break out this error as it's the one for helm dry-run as opposed to any returned later
 			}
 			return nil, nil, StateError, err
--- a/internal/applier/helm.go
+++ b/internal/applier/helm.go
@@ -24,6 +24,7 @@ import (
 	helmclient "github.com/operator-framework/helm-operator-plugins/pkg/client"
 
 	ocv1 "github.com/operator-framework/operator-controller/api/v1"
+	"github.com/operator-framework/operator-controller/internal/features"
 	"github.com/operator-framework/operator-controller/internal/rukpak/convert"
 	"github.com/operator-framework/operator-controller/internal/rukpak/preflights/crdupgradesafety"
 	"github.com/operator-framework/operator-controller/internal/rukpak/util"
@@ -160,6 +161,9 @@ func (h *Helm) getReleaseState(cl helmclient.ActionInterface, ext *ocv1.ClusterE
 			return nil
 		}, helmclient.AppendInstallPostRenderer(post))
 		if err != nil {
+			if features.OperatorControllerFeatureGate.Enabled(features.PreflightPermissions) {
+				// probably need to break out this error as it's the one for helm dry-run as opposed to any returned later
+			}
 			return nil, nil, StateError, err
 		}
 		return nil, desiredRelease, StateNeedsInstall, nil
--- a/internal/features/features.go
+++ b/internal/features/features.go
@@ -13,11 +13,10 @@ const (
 
 var operatorControllerFeatureGates = map[featuregate.Feature]featuregate.FeatureSpec{
 	// Add new feature gate definitions
-	// if you're adding a feature gate for fellow developers to use in ongoing development, use PreAlpha
 	// Ex: SomeFeature: {...}
 	PreflightPermissions: {
 		Default:       false,
-		PreRelease:    featuregate.PreAlpha, // keep this PreAlpha until done with feature development
+		PreRelease:    featuregate.Alpha,
 		LockToDefault: false,
 	},
 }
--- a/internal/features/features.go
+++ b/internal/features/features.go
@@ -6,13 +6,15 @@ import (
 )
 
 const (
-// Add new feature gates constants (strings)
-// Ex: SomeFeature featuregate.Feature = "SomeFeature"
+	// Add new feature gates constants (strings)
+	// Ex: SomeFeature featuregate.Feature = "SomeFeature"
+	PreflightPermissions featuregate.Feature = "PreflightPermissions"
 )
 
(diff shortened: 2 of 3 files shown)


internal/applier/helm.go
```diff
@@ -24,6 +24,7 @@ import (
 	helmclient "github.com/operator-framework/helm-operator-plugins/pkg/client"
 
 	ocv1 "github.com/operator-framework/operator-controller/api/v1"
+	"github.com/operator-framework/operator-controller/internal/features"
```
camilamacedo86, line 27: Here’s a polished version of your message with improved clarity and flow:

Hi @bentito,

I don’t think we should merge only the preflight.

If we need to do a release, we would have the gate but not the actual code—meaning the feature isn’t complete.

We should merge the gate along with at least a minimal version of the feature that provides value to end users. That way, we avoid having incomplete work on main.

Otherwise, we risk blocking future releases by merging incomplete changes that don’t yet add value for end users.

I am adding a 

/hold

c/c @thetechnick

bentito: There is nothing wrong with merging this. As a matter of fact it would be nice to merge, so that other contributors can use this for hiding their own code. It's a no-op even if a user happens to enable it. What exactly is the downside? "having incomplete work on main" isn't really a reason.

camilamacedo86 (changes requested):
/hold cancel

IMO, we shouldn't add a flag to enable a feature unless there's at least a partial initial implementation. As a user of this project, I would consider that a bug. There needs to be some functional value; otherwise, if we release just the flag and someone tries to use it, one of two things will happen:

a) They may assume a preflight check was performed, leading to a false impression.
b) They will consider it a bug since the flag is enabled but does nothing.

To ensure we maintain our DX goals, I think we should add at least one check here, as discussed in today’s meeting.

[Comment link](https://github.com/operator-framework/operator-controller/pull/1666#discussion_r1935172236)

internal/applier/helm.go
camilamacedo86: One of our goals is to ensure a good UX, so I don’t think we should merge PRs that add no value or could be misleading for users. We can do a release anytime if needed, but releasing with a flag that does nothing doesn’t seem ideal.

IMO, we shouldn’t add a flag to enable a feature unless there’s at least a partial initial implementation. As a user of this project, I would consider that a bug. There needs to be some functional value; otherwise, if we release just the flag and someone tries to use it, one of two things will happen:

a) They may assume a preflight check was performed, leading to a false impression.
b) They will consider it a bug since the flag is enabled but does nothing.

We discussed this today—I thought some people were okay with this approach, but I might be wrong. My takeaway was that adding the flag alone doesn’t make much sense. It’s just a couple of lines that could be included in the first PR to introduce an initial check—one of many that will follow through incremental work. I could be mistaken, but that was my understanding of our discussion.

bentito:
> To ensure we maintain our DX goals, I think we should add at least one check here.

What is DX? Dev Experience? I don't see how that's relevant.

For this to be a problem for a user, they have to 1) find out how to enable a feature gated item, 2) have some expectation of what enabling that feature would do. That seems unlikely without documenting the flag, which will come later.

The reason this PR is helpful to merge is that it allows other contributing devs to use the flag. There are lots of other ways that other devs could learn about and use this flag, but merging this PR is a straightforward manner, one that doesn't assume how the devs will communicate and share info. 

As soon as a dev uses the flag there will be at least one check hidden behind the flag.

internal/applier/helm.go
camilamacedo86: I think we can table this one since we agreed to use the option `LockToDefault: true  ` and that address the concern.

camilamacedo86:
Hi @bentito 

As discussed, we can add the option `LockToDefault: true`  than, my concern https://github.com/operator-framework/operator-controller/pull/1666#discussion_r1935172236 is sorted out! Thank you !

internal/features/features.go
```diff
@@ -6,13 +6,15 @@ import (
...
-// Ex: SomeFeature featuregate.Feature = "SomeFeature"
+	// Add new feature gates constants (strings)
+	// Ex: SomeFeature featuregate.Feature = "SomeFeature"
+	PreflightPermissions featuregate.Feature = "PreflightPermissions"
 )
 
 var operatorControllerFeatureGates = map[featuregate.Feature]featuregate.FeatureSpec{
 	// Add new feature gate definitions
 	// Ex: SomeFeature: {...}
+	PreflightPermissions: {Default: false, PreRelease: featuregate.Alpha},
```
camilamacedo86, line 17: It is missing `LockToDefault: true`

bentito: making it PreAlpha to settle all (most?) concerns

internal/features/features.go
```diff
@@ -13,8 +13,13 @@ const (
 
 var operatorControllerFeatureGates = map[featuregate.Feature]featuregate.FeatureSpec{
 	// Add new feature gate definitions
+	// if you're adding a feature gate for fellow developers to use in ongoing development, use PreAlpha
 	// Ex: SomeFeature: {...}
-	PreflightPermissions: {Default: false, PreRelease: featuregate.Alpha},
+	PreflightPermissions: {
+		Default:       false,
+		PreRelease:    featuregate.PreAlpha, // keep this PreAlpha until done with feature development
+		LockToDefault: false,
```
camilamacedo86, line 21: Should not be true?

bentito: If we set `LockToDefault` to `true`, @ankitathomas reports that you cannot access the feature-gated code for testing by then setting the emulated version. So LockToDefault true would render it untestable.

camilamacedo86: If we do not set it true:
- users will be able to opt-in / use a flag that do nothing
- then, we came back to the same concerns: https://github.com/operator-framework/operator-controller/pull/1666#discussion_r1935172236

internal/features/features.go
bentito: and back to Alpha b/c PreAlpha does not allow for toggling.

internal/features/features.go
bentito: I think this is resolved?

camilamacedo86 (approved):
/lgtm
/approve

PR #1668, opened 2025-01-30
🌱 [Monorepo] Fix make verify-crd-compatibility for catalogd

camilamacedo86:
To proceed, we needed to skip the check when the previous commit is not found. However, this results in the `verify-crd-compatibility` check always being skipped for catalogd. Example: https://github.com/operator-framework/operator-controller/actions/runs/13016547026/job/36307072381?pr=1653#step:4:68

Simply reverting the `verify-crd-compatibility` target and fixing the path to include catalogd will not work because the check must be executed from the root directory; otherwise, it fails to correctly resolve to commit SHAs for comparison.

To address this, we have combined the checks for both CRDs into a single verification step, ensuring correctness while eliminating duplicate file references.


LalatenduMohanty (approved):
/lgtm

PR #1669, opened 2025-01-30
🐛 Ensure catalog caches are valid after restart

oceanc80:
# Description
This addresses an intermittent failure we've seen where operator-controller appears to not resolve catalogs that exist. We identified the issue coming from the catalog caches not being valid after a restart.

This PR adds two things to address this issue:
- On startup of the binary, always clear the catalog cache directory

The PR also includes some small refactoring and fixes:
 - moves certain unpack functions to an util package and are now shared by catalog and bundle image unpacker
 - if unpack path happens to be an existing file, we delete it and unpack the image

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/catalogd/cmd/catalogd/main.go
+++ b/catalogd/cmd/catalogd/main.go
@@ -63,6 +63,7 @@ import (
 	"github.com/operator-framework/operator-controller/catalogd/internal/storage"
 	"github.com/operator-framework/operator-controller/catalogd/internal/version"
 	"github.com/operator-framework/operator-controller/catalogd/internal/webhook"
+	"github.com/operator-framework/operator-controller/internal/util"
 )
 
 var (
@@ -257,8 +258,8 @@ func main() {
 		systemNamespace = podNamespace()
 	}
 
-	if err := os.MkdirAll(cacheDir, 0700); err != nil {
-		setupLog.Error(err, "unable to create cache directory")
+	if err := util.EnsureEmptyDirectory(cacheDir, 0700); err != nil {
+		setupLog.Error(err, "unable to ensure empty cache directory")
 		os.Exit(1)
 	}
 
--- a/catalogd/internal/source/containers_image.go
+++ b/catalogd/internal/source/containers_image.go
@@ -30,6 +30,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	catalogdv1 "github.com/operator-framework/operator-controller/catalogd/api/v1"
+	"github.com/operator-framework/operator-controller/internal/rukpak/source"
+	"github.com/operator-framework/operator-controller/internal/util"
 )
 
 const ConfigDirLabel = "operators.operatorframework.io.index.configs.v1"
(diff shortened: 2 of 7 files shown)


catalogd/cmd/catalogd/main.go
```diff
@@ -263,6 +263,10 @@ func main() {
 	}
 
 	unpackCacheBasePath := filepath.Join(cacheDir, source.UnpackCacheDir)
+	if err := os.RemoveAll(unpackCacheBasePath); err != nil {
```
perdasilva, line 266: would it make sense to nuke `cacheDir` (above) instead? And, should we also make a similar change to o-c for the bundle cache?

catalogd/internal/source/containers_image.go
```diff
@@ -296,11 +296,17 @@ func (i *ContainersImageRegistry) unpackImage(ctx context.Context, unpackPath st
 		return wrapTerminal(fmt.Errorf("catalog image is missing the required label %q", ConfigDirLabel), specIsCanonical)
 	}
 
-	if err := os.MkdirAll(unpackPath, 0700); err != nil {
-		return fmt.Errorf("error creating unpack directory: %w", err)
-	}
 	l := log.FromContext(ctx)
-	l.Info("unpacking image", "path", unpackPath)
+	l.Info("unpacking image", "path", unpackPath, "temp path", tempUnpackPath)
```
perdasilva, line 300: careful, `tempUnpackPath` isn't define yet

catalogd/internal/source/containers_image.go
```diff
@@ -309,15 +315,21 @@ func (i *ContainersImageRegistry) unpackImage(ctx context.Context, unpackPath st
...
 			return nil
 		}(); err != nil {
-			return errors.Join(err, deleteRecursive(unpackPath))
+			return errors.Join(err, deleteRecursive(tempUnpackPath))
 		}
 	}
+	if err := os.MkdirAll(unpackPath, 0700); err != nil {
+		return fmt.Errorf("error creating unpack directory: %w", err)
+	}
+	if err := os.Rename(tempUnpackPath, unpackPath); err != nil {
```
perdasilva, line 330: I think there are some limitations with `Rename` renaming across filesystems (not sure if on the Pod /tmp and /var are different fs's though)

catalogd/cmd/catalogd/main.go
joelanford: Yes, but the cache dir itself is a volumeMount. So we likely need to nuke all of the contents of the cache dir without actually nuking the cache dit itself.

catalogd/cmd/catalogd/main.go
```diff
@@ -392,3 +396,16 @@ func podNamespace() string {
 	}
 	return string(namespace)
 }
+
+func clearCacheDir(cacheDirPath string) error {
```
joelanford, line 400: May as well rename this to "ensureEmptyDirectory" and throw the `os.MkdirAll` call in there as well?

catalogd/internal/source/containers_image.go
```diff
@@ -296,6 +293,10 @@ func (i *ContainersImageRegistry) unpackImage(ctx context.Context, unpackPath st
 		return wrapTerminal(fmt.Errorf("catalog image is missing the required label %q", ConfigDirLabel), specIsCanonical)
 	}
 
+	// ensure unpack directory is empty
+	if err := os.RemoveAll(unpackPath); err != nil {
+		return fmt.Errorf("error removing unpacked path: %w", err)
+	}
 	if err := os.MkdirAll(unpackPath, 0700); err != nil {
 		return fmt.Errorf("error creating unpack directory: %w", err)
 	}
```
joelanford, line 302: Another place for "ensureEmptyDirectory"? So maybe make that function a utility that we can reuse in a few places?

catalogd/internal/source/containers_image.go
```diff
@@ -431,3 +432,13 @@ func wrapTerminal(err error, isTerminal bool) error {
...
 	return reconcile.TerminalError(err)
 }
+
+func isImageUnpacked(unpackPath string) (bool, time.Time) {
+	if unpackStat, err := os.Stat(unpackPath); err == nil {
+		if !unpackStat.IsDir() {
+			panic(fmt.Sprintf("unexpected file at unpack path %q: expected a directory", unpackPath))
```
joelanford, line 439: I know this is a slight tangent from this PR, but I wonder if it would be better for us to just delete the file and unpack over top of where it was.

joelanford (commented):
First pass looks good to me.

One more overall comment: we should put this same set of changes in operator-controller's `main.go` and `containers_image.go` files.

joelanford (approved):

PR #1672, opened 2025-01-30
:seedling: Add support for SSL env vars to cert pool watcher

tmshort:
The SystemRoot store looks at the SSL_CERT_DIR and SSL_CERT_FILE environment variables for certificate locations. Because these variables are under control of the user, we should assume that the user wants to control the contents of the SystemRoot, and subsequently that those contents could change (as compared to certs located in the default /etc/pki location).

Thus, we should watch those locations if they exist.

### DIFF
--- a/internal/httputil/certpoolwatcher.go
+++ b/internal/httputil/certpoolwatcher.go
@@ -4,6 +4,8 @@ import (
 	"crypto/x509"
 	"fmt"
 	"os"
+	"slices"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


camilamacedo86 (approved):
/lgtm

internal/httputil/certpoolwatcher.go
```diff
@@ -44,8 +45,38 @@ func NewCertPoolWatcher(caDir string, log logr.Logger) (*CertPoolWatcher, error)
...
+				return nil, err
+			}
+		}
+	}
+
+	// If the SSL_CERT_DIR or SSL_CERT_FILE environment variables are
+	// specified, this means that we have some control over the system root
+	// location, thus they may change, thus we should watch those locations.
+	if d := os.Getenv("SSL_CERT_DIR"); d != "" {
+		dirs := strings.Split(d, ":")
```
joelanford, line 62:
```suggestion
		dirs := filepath.SplitList(d)
```

internal/httputil/certpoolwatcher.go
```diff
@@ -44,8 +45,38 @@ func NewCertPoolWatcher(caDir string, log logr.Logger) (*CertPoolWatcher, error)
...
+		}
+	}
+
+	if f := os.Getenv("SSL_CERT_FILE"); f != "" {
+		// only watch if we can Stat() it
+		if _, err := os.Stat(f); err == nil {
+			if err = watcher.Add(f); err != nil {
+				return nil, err
+			}
+		}
```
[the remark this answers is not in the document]
joelanford:
```go
watchPaths := []string{caDir, os.Getenv("SSL_CERT_DIR"), os.Getenv("SSL_CERT_FILE")}

slices.DeleteFunc(watchPaths, func(p string) bool { /* do stat here */})

for _, p := range watchPaths {
    if err := watcher.Add(p); err != nil {
        return nil, err
    }
}

internal/httputil/certpoolwatcher.go
tmshort: The separater is explicitly `":"`, and this is exactly how the golang library crypto/x509 does it, so I'm a bit reluctant to change it.

joelanford: Ok, if `crypto/x509` does it with simple `strings.Split`, I'm good.

joelanford (approved):

PR #1676, opened 2025-01-31
🌱 fix(e2e): wait for leader election

camilamacedo86:
TestClusterExtensionAfterOLMUpgrade was failing due to increased leader election timeouts, causing reconciliation checks to run before leadership was acquired.

This fix ensures the test explicitly waits for leader election logs (`"successfully acquired lease"`) before verifying reconciliation.

Example: https://github.com/operator-framework/operator-controller/actions/runs/13047935813/job/36401741998

Logs from operator-controller; 

```
I0130 08:12:51.290826       1 main.go:427] "starting manager" logger="setup"
I0130 08:12:51.291050       1 server.go:208] "Starting metrics server" logger="controller-runtime.metrics"
I0130 08:12:51.291109       1 server.go:83] "starting server" name="health probe" addr="[::]:8081"
I0130 08:12:51.291121       1 server.go:247] "Serving metrics server" logger="controller-runtime.metrics" bindAddress=":8443" secure=true
I0130 08:12:51.291209       1 leaderelection.go:257] attempting to acquire leader lease olmv1-system/9c4404e7.operatorframework.io...


```

### DIFF
--- a/test/upgrade-e2e/post_upgrade_test.go
+++ b/test/upgrade-e2e/post_upgrade_test.go
@@ -40,6 +40,17 @@ func TestClusterExtensionAfterOLMUpgrade(t *testing.T) {
 	t.Log("Wait for operator-controller deployment to be ready")
 	managerPod := waitForDeployment(t, ctx, "operator-controller-controller-manager")
 
+	t.Log("Wait for acquired leader election")
+	// Average case is under 1 minute but in the worst case: (previous leader crashed)
+	// we could have LeaseDuration (137s) + RetryPeriod (26s) +/- 163s
+	leaderCtx, leaderCancel := context.WithTimeout(ctx, 3*time.Minute)
+	defer leaderCancel()
+
+	leaderSubstrings := []string{"successfully acquired lease"}
+	leaderElected, err := watchPodLogsForSubstring(leaderCtx, managerPod, "manager", leaderSubstrings...)
+	require.NoError(t, err)
+	require.True(t, leaderElected)
+
 	t.Log("Reading logs to make sure that ClusterExtension was reconciled by operator-controller before we update it")
 	// Make sure that after we upgrade OLM itself we can still reconcile old objects without any changes
 	logCtx, cancel := context.WithTimeout(ctx, time.Minute)


camilamacedo86:
@joelanford @thetechnick

camilamacedo86:
@LalatenduMohanty @joelanford @thetechnick

test/upgrade-e2e/post_upgrade_test.go
```diff
@@ -40,6 +40,18 @@ func TestClusterExtensionAfterOLMUpgrade(t *testing.T) {
 	t.Log("Wait for operator-controller deployment to be ready")
 	managerPod := waitForDeployment(t, ctx, "operator-controller-controller-manager")
 
+	t.Log("Start measuring leader election time")
+	leaderStartTime := time.Now()
+	leaderElectionCtx, leaderCancel := context.WithTimeout(ctx, 2*time.Minute)
```
camilamacedo86, line 45: 2 minutes because the times show intermittently.

Observations:
- The first PR/commit worked fine with a 1-minute timeout for leader election.
- After adjusting log messages and pushing again, the test failed—1 minute was no longer enough.
- Increasing the timeout to 2 minutes allowed the test to pass. The leader election toke 33.54 seconds.
- Running the same test locally, leader election took 81.60 seconds.

cmd/operator-controller/main.go
```diff
@@ -235,9 +235,9 @@ func main() {
 		LeaderElectionID:       "9c4404e7.operatorframework.io",
 		// Recommended Leader Election values
 		// https://github.com/openshift/enhancements/blob/61581dcd985130357d6e4b0e72b87ee35394bf6e/CONVENTIONS.md#handling-kube-apiserver-disruption
-		LeaseDuration: ptr.To(137 * time.Second),
-		RenewDeadline: ptr.To(107 * time.Second),
-		RetryPeriod:   ptr.To(26 * time.Second),
+		LeaseDuration: ptr.To(137 * time.Second), // Default: 15s
```
bentito, line 238: are you just commenting them so we see how long they are compared to the defaults?

test/upgrade-e2e/post_upgrade_test.go
```diff
@@ -40,6 +40,20 @@ func TestClusterExtensionAfterOLMUpgrade(t *testing.T) {
 	t.Log("Wait for operator-controller deployment to be ready")
 	managerPod := waitForDeployment(t, ctx, "operator-controller-controller-manager")
 
+	t.Log("Start measuring leader election time")
```
perdasilva, line 43: I think we need to be careful about how we measure the timing here. What we are measuring right now is the amount of time between:
-  the test detecting that the operator-controller deployment is finished, and
- how it takes for `watchPodLogsForSubstring(leaderElectionCtx, managerPod, "manager", leaderSubstrings...)` to return

This may correlate with the time taken for leader election, but it won't _necessarily_ correlate with it. E.g. let's say I upgrade the deployments, go out for lunch for 1h, come back and run the post upgrade test.

Maybe it would be better to extract the timestamp in the first and leader election log lines instead?

cmd/operator-controller/main.go
camilamacedo86: It only makes it easier for us to know the default.
We do not need to check the source code. 
We can remove it. Just a comment to try make easier

camilamacedo86: Removed 👍

perdasilva (approved):
lgtm ^^

test/upgrade-e2e/post_upgrade_test.go
```diff
@@ -40,6 +40,17 @@ func TestClusterExtensionAfterOLMUpgrade(t *testing.T) {
 	t.Log("Wait for operator-controller deployment to be ready")
 	managerPod := waitForDeployment(t, ctx, "operator-controller-controller-manager")
 
+	t.Log("Wait for acquired leader election")
+	// Average case is under 1 minute but in the worst case: (previous leader crashed)
+	// we could have LeaseDuration (137s) + RetryPeriod (26s) +/- 163s
+	leaderCtx, leaderCancel := context.WithTimeout(ctx, 3*time.Minute)
```
LalatenduMohanty, line 46: I am assuming 3 minutes is the worst case scenario. I am not familiar with `context.WithTimeout` , does it return if we acquire the lease before 163s?

test/upgrade-e2e/post_upgrade_test.go
```diff
@@ -40,6 +40,17 @@ func TestClusterExtensionAfterOLMUpgrade(t *testing.T) {
...
+	defer leaderCancel()
+
+	leaderSubstrings := []string{"successfully acquired lease"}
+	leaderElected, err := watchPodLogsForSubstring(leaderCtx, managerPod, "manager", leaderSubstrings...)
```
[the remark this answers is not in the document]
bentito: I realize it's also longer and more code, but the upside is it reacts right away, like watching for the pod, but without caring if strings change at some point, and break our tests out of our control.

test/upgrade-e2e/post_upgrade_test.go
perdasilva: `context.WithTimeout` just gives you a context that timesout (gets cancelled) after then timeout period.
This means that the call to `watchPodLogsForSubstring(leaderCtx, managerPod, "manager", leaderSubstrings...)` will return with an error if it hasn't already after 3 minutes.

test/upgrade-e2e/post_upgrade_test.go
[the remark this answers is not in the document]
perdasilva: that's a good idea! If this work is blocking CI, I'd say merge it as it is, then follow up with the watch ^^

test/upgrade-e2e/post_upgrade_test.go
LalatenduMohanty: ok, looks like it is a straight forward timeout method.

test/upgrade-e2e/post_upgrade_test.go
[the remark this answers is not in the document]
camilamacedo86: Yes, I agree that we could do something more fancy
But we check the logs in many places, indeed below. 
We can see if we improve after, but there is no reason for us to face the pain of the flak.

PR #1688, opened 2025-01-31
🐛 gracefully shutdown reconcilers and catalogd FBC server

joelanford:
# Description

This PR adds a new `context` utility package with a new kind of context that is cancelled some duration _after_ its parent context is cancelled. This is useful for reconcilers and servers where having some extra time to complete can decrease the chance of interruptions.

The particular motivation for this is a case we've seen in e2e tests where the operator-controller pod is cancelled while it is in the middle of an installation or upgrade. If this process is interrupted, it leaves the underlying installation in an intermediate state that requires human intervention to recover.

While this change does not specifically try to make it possible to progress from this intermediate state, it _does_ make it less likely for the interruption to happen during this critical section.

It is an open question of the design philosophy of the project how an unfinished installation/upgrade should be treated. The current stance is "let a human decide what to do", but that is not a reasonable answer at scale, so I think we'll need to think about further improvements to help OLM know when a roll forward (or back) is safe.

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -66,6 +66,8 @@ linters-settings:
       alias: ctrl
     - pkg: github.com/blang/semver/v4
       alias: bsemver
+    - pkg: "^github.com/operator-framework/operator-controller/internal/util/([^/]+)$"
+      alias: "${1}util"
 
 output:
   formats:
--- a/catalogd/config/base/manager/manager.yaml
+++ b/catalogd/config/base/manager/manager.yaml
@@ -85,7 +85,7 @@ spec:
         imagePullPolicy: IfNotPresent
         terminationMessagePolicy: FallbackToLogsOnError
       serviceAccountName: controller-manager
-      terminationGracePeriodSeconds: 10
+      terminationGracePeriodSeconds: 60
       volumes:
         - name: cache
           emptyDir: {}
--- a/catalogd/internal/controllers/core/clustercatalog_controller.go
+++ b/catalogd/internal/controllers/core/clustercatalog_controller.go
@@ -40,6 +40,7 @@ import (
 	catalogdv1 "github.com/operator-framework/operator-controller/catalogd/api/v1"
 	"github.com/operator-framework/operator-controller/catalogd/internal/source"
 	"github.com/operator-framework/operator-controller/catalogd/internal/storage"
+	contextutil "github.com/operator-framework/operator-controller/internal/util/context"
 )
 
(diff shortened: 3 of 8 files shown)


joelanford:
I added the `do-not-merge/work-in-progress` label for two reasons:
1. I still need to add tests
2. I think maintainers should discuss whether this is a change we want to make and if there is a deeper problem we still want to solve around recovering from interruptions.

joelanford:
Yeah, things could get out of hand if these delayed contexts exist all over the place, I agree.

But I'm not sure cleanup logic is the right concept either. For one, we would generally want cleanup logic related to a particular CR object to run as part of reconcile for better recovery mechanics. But also, in the case of an interrupted helm release, there is no safe cleanup after the helm release has been interrupted unless there is some signal from the chart author, the user, or both.

So I'm trying to think through a mechanic where:
1. we stop pull items off the controller queues
2. we allow any in-flight reconciles to continue for some period of time. it would be interesting if we had some metrics history for a histogram of reconcile duration. I know we have a metric for that, but I don't think we have any historical data.

Also, I'm not at all attached to this PR, and I can definitely see how it has downsides. Another downside is that it could make issues due to interruptions even harder to find and reproduce.

This PR is mainly a lever for getting a conversation going, so I'm interesting to see where this leads!

joelanford:
Also of note, I just looked at our helm usage, and tracing through helm applier -> helm-operator-plugins -> helm, we are using the helm install/upgrade variants that use `context.Background()`, so I suspect that existing code is running up to `terminationGracePeriodSeconds` after the stop signal is received by the process.

Maybe we should switch over to Helm's `RunWithContext` methods in helm-operator-plugins. But that would risk making things worse if we don't have some other mitigation in place.

azych:
I think it might be a good start to go through all of the reconciliation logic and try to identify places where an interruption could mean entering into this intermediate state. Once we have that we should better understand the scope of this and think about what can we do in each specific case to potentially mitigate/cleanup any sideffects to a working state or if there is anything we could actually do. Also, fully agree about taking into account any metrics we might have, or adding new ones to have a better understanding if need be.

Just thinking out loud, but one rough idea could be to keep track of how many reconciliations are in those critical places and communicate this outside of reconciliation loop and specific controller, so that upon receiving the interruption signal we could for example adjust the behavior/timeout on a more "global" level, which should be potentially easier to reason about and debug.

PR #1689, opened 2025-01-31
✨ release leader election lease on manager cancellation

### LINKED ISSUE #1687: Release leader election lease on shutdown
In https://github.com/operator-framework/operator-controller/pull/1683, we fixed an issue related to flaky upgrade e2e tests that were a result of longer election times caused by https://github.com/operator-framework/operator-controller/pull/1663 (which was an intentional change to be tolerant to longer periods of apiserver downtime or heavy load)

As part of 1683, we bumped a test timeout to account for the new leader election settings.

My spidey sense always goes off when I see a timeout get bumped up in our tests. After some disucssion with other maintainers, I propose that we:
1. Revert the change to the upgrade e2e test from 1683.
2. Set `LeaderElectionReleaseOnCancel: true` in our manager options for both `catalogd` and `operator-controller`.

joelanford:
# Description

Fixes #1687 

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/cmd/operator-controller/main.go
+++ b/cmd/operator-controller/main.go
@@ -228,11 +228,12 @@ func main() {
 	}
 
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
-		Scheme:                 scheme.Scheme,
-		Metrics:                metricsServerOptions,
-		HealthProbeBindAddress: probeAddr,
-		LeaderElection:         enableLeaderElection,
-		LeaderElectionID:       "9c4404e7.operatorframework.io",
+		Scheme:                        scheme.Scheme,
+		Metrics:                       metricsServerOptions,
+		HealthProbeBindAddress:        probeAddr,
+		LeaderElection:                enableLeaderElection,
+		LeaderElectionID:              "9c4404e7.operatorframework.io",
+		LeaderElectionReleaseOnCancel: true,
 		// Recommended Leader Election values
 		// https://github.com/openshift/enhancements/blob/61581dcd985130357d6e4b0e72b87ee35394bf6e/CONVENTIONS.md#handling-kube-apiserver-disruption
 		LeaseDuration: ptr.To(137 * time.Second),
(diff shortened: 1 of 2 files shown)


joelanford:
I added the `do-not-merge/work-in-progress` label. Let's get consensus that this is a safe change before merge. I know @camilamacedo86 wanted to go over the details more closely.

joelanford:
Ha. The latest release doesn't release the lease. So we'll have to make the changes in main.go's, then release again, then revert the e2e changes.

LalatenduMohanty (approved):
/lgtm

joelanford:
> This means that if the pod is shutdown properly, it will give up its lease. Is that right? What are the risks?

Correct, and if it is not shut down properly, it likely won't give up its lease because giving up its lease is the last thing it does.

The main risk is that we go out of our way to run a separate goroutine that does leader-y things whose lifetime goes beyond that of the manager. If we do that, then the manager will release the lease, some other process could assume leadership, and the our other goroutine and the new leader will be duking it out.

I don't think that's a big risk though.

camilamacedo86 (approved):
👍

cmd/operator-controller/main.go
```diff
@@ -228,11 +228,12 @@ func main() {
...
-		Scheme:                 scheme.Scheme,
-		Metrics:                metricsServerOptions,
-		HealthProbeBindAddress: probeAddr,
-		LeaderElection:         enableLeaderElection,
-		LeaderElectionID:       "9c4404e7.operatorframework.io",
+		Scheme:                        scheme.Scheme,
+		Metrics:                       metricsServerOptions,
+		HealthProbeBindAddress:        probeAddr,
+		LeaderElection:                enableLeaderElection,
+		LeaderElectionID:              "9c4404e7.operatorframework.io",
```
camilamacedo86, line 235: Nit: Could we not use operator-controller-lock here as well like we have in catalogd? 
Maybe for another PR

joelanford: I don't think we can do that. Because then old code and new code could both get elected leader under different leader election IDs.

PR #1690, opened 2025-02-02
🌱 consolidate image layer handling; move fs utils

### LINKED ISSUE #1711: Consolidate image layer handling; move fs utils
Now that we have a monorepo, we can begin consolidating various bits of code that had identical or substantially similar implementations in catalogd and operator-controller. 

P.S: Created the issue retroactively after @joelanford created the PR https://github.com/operator-framework/operator-controller/pull/1690

joelanford:
# Description

Now that we have a monorepo, we can begin consolidating various bits of code that had identical or substantially similar implementations in catalogd and operator-controller.

This PR adds an `image` utility package, for now with common code that handles applying an image to an unpack directory. Of note is that catalogd and operator-controller had separate sets of filters used when applying the image layers. This PR, extracts the separate filter functions into the image utility library, and then updates the (still mostly duplicate) unpacker implementations to use the new shared code.

In follow-up PRs, we can further de-dup the containers/image unpacker implementations (and any other duplicate code we find).

NOTE: To avoid an import cycle, I needed to also make a new `fs` utility package for `internal/rukpak/source/helpers.go` (which are all fs-related helpers). I also took the opportunity to move `internal/fsutil/* into the new `fs` utility package.

Fixes #1711

## Reviewer Checklist

- [ ] API Go Documentation


internal/util/fs/fs.go
```diff
@@ -33,21 +34,42 @@ func DeleteReadOnlyRecursive(root string) error {
...
 
-// IsImageUnpacked checks whether an image has been unpacked in `unpackPath`.
-// If true, time of unpack will also be returned. If false unpack time is gibberish (zero/epoch time).
-// If `unpackPath` is a file, it will be deleted and false will be returned without an error.
-func IsImageUnpacked(unpackPath string) (bool, time.Time, error) {
-	unpackStat, err := os.Stat(unpackPath)
+// IsDirectoryPresent checks whether a directory exists. If true, directory mode time will also be returned.
+// If false, mod time is gibberish (zero/epoch time).
+// If `dir` is a file, false will be returned. The file can be deleted by setting `deleteConflictingFile` to true.
+func IsDirectoryPresent(dir string, deleteConflictingFile bool) (bool, time.Time, error) {
```
perdasilva, line 40: This one is kinda bothering me, as it did as `IsImageUnpacked` as well. We're naming it as a boolean function (Is_), but we returning time as well, and there's the side-effect of deleting the path if it's not a directory.
 I think you tried to address that here through the parameter, but chances are we probably won't event set it to false anyway. I could live with something like IsImageUnpacked, or checkImageUnpacked, because that encapsulates that process, which we could then spruce up with CRC checks or whatever. The fact that we just check for a directory is an implementation detail (even through the side-effect still irked me a little XD). I'm thinking, maybe we could either:
a) remove this function entirely and go back to just using os.Stat in the Unpack function: this makes the whole isFile handling more transparent, and gives us the unpack time if we want it. We would duplicate the code a little across catd and o-c, but my guess is we'll eventually want to merge containers_image at some point anyway.
b) move this to the image utils package and call it something like `GetImageUnpackState`, or `EnsureUnpackPath`, or something that conveys everything the function is trying to do.
c) just split the behavior: have one function that checks if the image has already been unpacked (e.g. directory is present) and another that ensures that unpackPath doesn't point to a file (and if it does nukes it).

wdyt?

internal/util/fs/fs.go
```diff
@@ -1,28 +1,29 @@
...
 	"path/filepath"
 	"time"
 )
 
 const (
-	OwnerWritableFileMode os.FileMode = 0700
-	OwnerWritableDirMode  os.FileMode = 0700
-	OwnerReadOnlyFileMode os.FileMode = 0400
-	OwnerReadOnlyDirMode  os.FileMode = 0500
+	ownerWritableFileMode os.FileMode = 0700
```
perdasilva, line 13: I wonder if we should export these and use them in the image utils. Maybe we could even add a comment here to explain why we don't give group perms.

perdasilva:
I'm wondering if we should call 

```
if err := i.deleteOtherImages(bundle.Name, canonicalRef.Digest()); err != nil {
		return nil, fmt.Errorf("error deleting old images: %w", err)
}
```

On every successful run of unpack. E.g. always ensure that the other images previously unpacked are gone, and we don't have some garbage hanging around because the process got killed mid deletion (though I guess the cache clear at pod restart should handle that...)

joelanford:
/hold

There's a problem we discovered with EnsureEmptyDirectory, when it tries to delete directories with read-only bits set. Need to fix that first, then rebase this.

internal/util/image/layers.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/containerd/containerd/archive"
+	"github.com/containers/image/v5/pkg/blobinfocache/none"
+	"github.com/containers/image/v5/pkg/compression"
+	"github.com/containers/image/v5/types"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	fsutil "github.com/operator-framework/operator-controller/internal/util/fs"
+)
+
+func ForceOwnershipRWX() archive.Filter {
```
camilamacedo86, line 22: Could we doc the funcs ?

internal/util/fs/fs.go
joelanford: I didn't see any usage outside of this package, so opted to unexport them. If there's a strong use case that comes up later, we can always export them again. This is all internal, so it doesn't matter either way from a public API standpoint, but it can be nice to keep even internal packages nice and tidy to avoid leaky abstractions.

joelanford:
> On every successful run of unpack. E.g. always ensure that the other images previously unpacked are gone, and we don't have some garbage hanging around because the process got killed mid deletion

You're thinking to move it up right after we determine the canonical ref, but before we check to see if we already have the canonical ref cached? I think that makes sense to me. Right now this is only called in the case that we actually pull a new image, but not when we return an existing one.

> (though I guess the cache clear at pod restart should handle that...)

Yes, but it's better to keep packages self-contained and avoid relying on hidden side-effects like that. I feel like that is one of OLMv0's downfalls.

internal/util/image/layers.go
```diff
@@ -0,0 +1,99 @@
...
+			return nil
+		}(); err != nil {
+			return errors.Join(err, fsutil.DeleteReadOnlyRecursive(dest))
+		}
+	}
+	if err := fsutil.SetReadOnlyRecursive(dest); err != nil {
+		return fmt.Errorf("error making unpack directory read-only: %w", err)
+	}
+	return nil
+}
```
azych, line 99: to me all of those wrapper functions above are very specific to the logic required in `source` and it might be a good idea to defer placing them in an external dedicated util/<x> package until there is a need to share them, until them it would be better to have them close to where they're being used imho

azych: nvm, my bad - I noticed they are actually shared between op-con and catalogd ;)

azych (approved):
/lgtm

internal/rukpak/source/containers_image.go
```diff
@@ -71,11 +67,15 @@ func (i *ContainersImageRegistry) Unpack(ctx context.Context, bundle *BundleSour
...
-	if isUnpacked, _, err := IsImageUnpacked(unpackPath); isUnpacked && err == nil {
+	if _, err := fsutil.GetDirectoryModTime(unpackPath); err == nil {
 		l.Info("image already unpacked", "ref", imgRef.String(), "digest", canonicalRef.Digest().String())
 		return successResult(bundle.Name, unpackPath, canonicalRef), nil
-	} else if err != nil {
-		return nil, fmt.Errorf("error checking bundle already unpacked: %w", err)
+	} else if errors.Is(err, fsutil.ErrNotDirectory) {
+		if err := fsutil.DeleteReadOnlyRecursive(unpackPath); err != nil {
+			return nil, err
+		}
```
LalatenduMohanty, line 76: This seems to be a new code i.e. we are deleting the file if it is not a directory, is that right?

joelanford: It was originally in `IsDirectoryPresent`, but and I both @perdasilva thought that was a terrible idea given the name and location of that helper function. So I renamed that function to `GetDirectoryModTime` and moved the delete call to the caller side.

internal/util/image/layers.go
LalatenduMohanty: +1 we should add go docs for these functions.

joelanford:
I'm going to hold off on moving `deleteOtherImages`. I'm thinking about this scenario:
1. I'm using image `foobar:v1`. It's unpacked and working great
2. I decide I want `foobar:v2`, so I update my image ref.
3. Oops, for some reason I can't pull that image (typo, auth problem, etc.)
4. Set my ref back to `foobar:v1`. Phew still unpacked so I don't have to pull it again.

camilamacedo86 (approved):
It seems good for me 👍

PR #1716, opened 2025-02-05
✨ Check known required permissions for install before installing with the helm applier

trgeiger:
Adds a check that makes sure the installer service account has the necessary get permissions to get all the resources it will need to inspect for a server-connected dry-run and returns errors detailing all the missing get permissions. This prevents a hung server-connected dry-run getting caught on individual missing get permissions.

# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/operator-controller/applier/helm.go
+++ b/internal/operator-controller/applier/helm.go
@@ -18,12 +18,12 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	apimachyaml "k8s.io/apimachinery/pkg/util/yaml"
+	authorizationv1client "k8s.io/client-go/kubernetes/typed/authorization/v1"
 	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
 	helmclient "github.com/operator-framework/helm-operator-plugins/pkg/client"
-	authorizationv1client "k8s.io/client-go/kubernetes/typed/authorization/v1"
 
 	ocv1 "github.com/operator-framework/operator-controller/api/v1"
 	"github.com/operator-framework/operator-controller/internal/operator-controller/authorization"
@@ -111,7 +111,6 @@ func shouldSkipPreflight(ctx context.Context, preflight Preflight, ext *ocv1.Clu
 }
 
 func (h *Helm) Apply(ctx context.Context, contentFS fs.FS, ext *ocv1.ClusterExtension, objectLabels map[string]string, storageLabels map[string]string) ([]client.Object, string, error) {
-
 	if features.OperatorControllerFeatureGate.Enabled(features.PreflightPermissions) {
 		authclient, err := h.AuthClientMapper.GetAuthenticationClient(ctx, ext)
 		if err != nil {
--- a/internal/operator-controller/applier/helm.go
+++ b/internal/operator-controller/applier/helm.go
@@ -18,12 +18,15 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	apimachyaml "k8s.io/apimachinery/pkg/util/yaml"
+	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
 	helmclient "github.com/operator-framework/helm-operator-plugins/pkg/client"
+	authorizationv1client "k8s.io/client-go/kubernetes/typed/authorization/v1"
 
 	ocv1 "github.com/operator-framework/operator-controller/api/v1"
+	"github.com/operator-framework/operator-controller/internal/operator-controller/authorization"
 	"github.com/operator-framework/operator-controller/internal/operator-controller/features"
 	"github.com/operator-framework/operator-controller/internal/operator-controller/rukpak/convert"
 	"github.com/operator-framework/operator-controller/internal/operator-controller/rukpak/preflights/crdupgradesafety"
@@ -53,9 +56,38 @@ type Preflight interface {
 	Upgrade(context.Context, *release.Release) error
 }
 
+type RestConfigMapper func(context.Context, client.Object, *rest.Config) (*rest.Config, error)
+
+type AuthClientMapper struct {
+	rcm     RestConfigMapper
+	baseCfg *rest.Config
+}
+
+func (acm *AuthClientMapper) GetAuthenticationClient(ctx context.Context, ext *ocv1.ClusterExtension) (*authorizationv1client.AuthorizationV1Client, error) {
+	authcfg, err := acm.rcm(ctx, ext, acm.baseCfg)
+	if err != nil {
+		return nil, err
+	}
+
+	authclient, err := authorizationv1client.NewForConfig(authcfg)
+	if err != nil {
+		return nil, err
+	}
+
+	return authclient, nil
+}
+
 type Helm struct {
 	ActionClientGetter helmclient.ActionClientGetter
 	Preflights         []Preflight
+	AuthClientMapper   AuthClientMapper
+}
+
+func NewAuthClientMapper(rcm RestConfigMapper, baseCfg *rest.Config) AuthClientMapper {
+	return AuthClientMapper{
+		rcm:     rcm,
+		baseCfg: baseCfg,
+	}
 }
 
 // shouldSkipPreflight is a helper to determine if the preflight check is CRDUpgradeSafety AND
@@ -79,7 +111,21 @@ func shouldSkipPreflight(ctx context.Context, preflight Preflight, ext *ocv1.Clu
 }
 
 func (h *Helm) Apply(ctx context.Context, contentFS fs.FS, ext *ocv1.ClusterExtension, objectLabels map[string]string, storageLabels map[string]string) ([]client.Object, string, error) {
+
+	if features.OperatorControllerFeatureGate.Enabled(features.PreflightPermissions) {
+		authclient, err := h.AuthClientMapper.GetAuthenticationClient(ctx, ext)
+		if err != nil {
+			return nil, "", err
+		}
+
+		err = h.checkContentPermissions(ctx, contentFS, authclient, ext)
+		if err != nil {
+			return nil, "", err
+		}
+	}
+
 	chrt, err := convert.RegistryV1ToHelmChart(ctx, contentFS, ext.Spec.Namespace, []string{corev1.NamespaceAll})
+
 	if err != nil {
 		return nil, "", err
 	}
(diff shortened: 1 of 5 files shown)


internal/operator-controller/applier/helm.go
```diff
@@ -53,9 +56,36 @@ type Preflight interface {
...
+
+type NewForConfigFunc func(*rest.Config) (authorizationv1client.AuthorizationV1Interface, error)
+
+type AuthClientMapper struct {
+	rcm          RestConfigMapper
+	baseCfg      *rest.Config
+	NewForConfig NewForConfigFunc
+}
+
+func (acm *AuthClientMapper) GetAuthenticationClient(ctx context.Context, ext *ocv1.ClusterExtension) (authorizationv1client.AuthorizationV1Interface, error) {
```
joelanford, line 69:
```suggestion
func (acm *AuthClientMapper) GetAuthorizationClient(ctx context.Context, ext *ocv1.ClusterExtension) (authorizationv1client.AuthorizationV1Interface, error) {
```

internal/operator-controller/applier/helm.go
```diff
@@ -53,9 +56,36 @@ type Preflight interface {
 	Upgrade(context.Context, *release.Release) error
 }
 
+type RestConfigMapper func(context.Context, client.Object, *rest.Config) (*rest.Config, error)
+
+type NewForConfigFunc func(*rest.Config) (authorizationv1client.AuthorizationV1Interface, error)
+
+type AuthClientMapper struct {
```
joelanford, line 63:
```suggestion
type AuthorizationClientMapper struct {
```

trgeiger: Haha, yeah I have that in a local change I need to push up, got myself mixed up

internal/operator-controller/authorization/authorization.go
```diff
@@ -0,0 +1,112 @@
...
+			NonResourceURLs: rule.NonResourceURLs,
+		})
+	}
+
+	resAttrs := []authorizationv1.ResourceAttributes{}
+	namespacedErrs := []error{}
+	clusterScopedErrs := []error{}
+	requiredVerbs := []string{"get", "create", "update", "list", "watch", "delete", "patch"}
+
+	for _, o := range objects {
```
joelanford, line 58: If the object is a ClusterRole or Role, we need to check that we have _its_ rules at the appropriate scope (or the `escalate` verb on clusterroles or roles)

If the object is a ClusterRoleBinding or a RoleBinding, we need to check that we have the rules present in the referenced ClusterRole/Role (or the `bind` verb on clusterrolesbindings or rolebindings)

Right?

internal/operator-controller/authorization/authorization.go
joelanford, line 1: I feel like I've seen libraries in k/k for this. Looking now.

internal/operator-controller/authorization/authorization.go
trgeiger: That does sound right, yes. We can try do to that work here or have that be a future iteration/improvement in another PR. I'm fine with either. This PR started as just implementing the quick "get" check but along the way realized we can and should just go ahead and check all the verbs above while we check for get so that's why nothing more complicated is happening yet.

internal/operator-controller/authorization/authorization.go
joelanford: https://github.com/kubernetes/kubernetes/blob/f0077a3689669018557de723a0416fda267d132f/plugin/pkg/auth/authorizer/rbac/rbac.go

Maybe this could help?

trgeiger: Oh yeah I actually was using the RulesAllow in an earlier version of this but it was causing some import issue Brett was noticing

trgeiger: I'll fix that up and use those functions again

joelanford: Looking around, I also see:
- https://github.com/kubernetes/kubernetes/blob/f0077a3689669018557de723a0416fda267d132f/staging/src/k8s.io/component-helpers/auth/rbac/validation/policy_comparator.go#L27
- https://github.com/kubernetes/kubernetes/blob/f0077a3689669018557de723a0416fda267d132f/pkg/registry/rbac/validation/rule.go#L53

Ultimately though, it seems like the RBACAuthorizer encapsulates everything we'd want to do and is the actual authorizer the apiserver uses. The catch is that we'd have to give ourselves permission to list/watch all RBAC objects in the cluster. Is that a dealbreaker? 

Maybe something we should go ask the sig-auth folks: what's better for best effort authz checks? use an in-process RBAC authorizer with cached RBAC from informers, or a flood of SSAR requests?

internal/operator-controller/authorization/authorization.go
joelanford: I think we'll have to have this basically immediately because every single bundle we support is going to have RBAC objects in the manifest by virtue of all of this stuff having CRDs and controllers.

internal/operator-controller/authorization/authorization.go
```diff
@@ -0,0 +1,112 @@
...
+			canI = true
+			break
+		}
+	}
+	return canI
+}
+
+// SelfSubjectRulesReview formats the resource names as lowercase and plural
+func sanitizeResourceName(resourceName string) string {
+	return strings.ToLower(resourceName) + "s"
```
joelanford, line 111: This is not a safe way to handle this because pluralization is complicated, and CRDs get to define their resource name however they want. Seems like we would need to use a rest mapper to get the actual resource name for the object.

internal/operator-controller/authorization/authorization.go
trgeiger: By give ourselves, do you mean operator-controller's manager would need list/watch on all RBAC?

internal/operator-controller/applier/helm.go
```diff
@@ -53,9 +56,36 @@ type Preflight interface {
 	Upgrade(context.Context, *release.Release) error
 }
 
+type RestConfigMapper func(context.Context, client.Object, *rest.Config) (*rest.Config, error)
+
+type NewForConfigFunc func(*rest.Config) (authorizationv1client.AuthorizationV1Interface, error)
+
+type AuthorizationClientMapper struct {
```
azych, line 63: would it make sense for `AuthorizationClientMapper` and its related types and methods to be part of authorization package? 
To me they don't seem helm specific and we'll be encapsulating authorization into a separate package.

Also, to avoid situation where we the caller needs to know about `authorizationv1client.AuthorizationV1Interface` and first obtain the authorization client and then pass it to `CheckObjectPermissions` I was thinking we could have `CheckContentPermissions` be a public method of the custom authorization client that we return, which would internally call more generic `checkObjectPermissions`.

For that, the client in authorization package would have to wrap `authorizationv1client.AuthorizationV1Interface` and then in this package, we'd just define an interface of the custom authorization client with whatever we need to use here similar to this:

```go
type authorizationClient interface {
    CheckContentPermissions(...)
}
```

With that we would be encapsulating internals of authorization within a dedicated single package and exposing a clean interface to the callers with the added benefit of having a simple interface to mock on the callers side.

WDYT?

Obviously those are just suggestions, the second one especially.

internal/operator-controller/applier/helm.go
```diff
@@ -79,7 +109,20 @@ func shouldSkipPreflight(ctx context.Context, preflight Preflight, ext *ocv1.Clu
...
 func (h *Helm) Apply(ctx context.Context, contentFS fs.FS, ext *ocv1.ClusterExtension, objectLabels map[string]string, storageLabels map[string]string) ([]client.Object, string, error) {
+	if features.OperatorControllerFeatureGate.Enabled(features.PreflightPermissions) {
+		authclient, err := h.AuthorizationClientMapper.GetAuthorizationClient(ctx, ext)
+		if err != nil {
+			return nil, "", err
+		}
+
+		err = h.checkContentPermissions(ctx, contentFS, authclient, ext)
+		if err != nil {
+			return nil, "", err
```
azych, line 120: might be a good idea to wrap with context, ie. `fmt.Errorf(err, "failed checking content permissions") for better debuggability. Here and other places.

internal/operator-controller/authorization/authorization.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	for _, rule := range ssrr.Status.NonResourceRules {
+		rules = append(rules, rbacv1.PolicyRule{
+			Verbs:           rule.Verbs,
+			NonResourceURLs: rule.NonResourceURLs,
+		})
+	}
+
+	resAttrs := []authorizationv1.ResourceAttributes{}
```
azych, line 53: nit: you could probably initialize it with specific length or at least capacity equal to len(requiredVerbs) * len(objects)

internal/operator-controller/authorization/authorization.go
```diff
@@ -0,0 +1,112 @@
...
+				Group:     o.GetObjectKind().GroupVersionKind().Group,
+				Name:      o.GetName(),
+			})
+		}
+	}
+
+	for _, resAttr := range resAttrs {
+		if !canI(resAttr, rules) {
+			if resAttr.Namespace != "" {
+				namespacedErrs = append(namespacedErrs, fmt.Errorf("cannot %s %s %s in namespace %s",
```
azych, line 73: nit: maybe it would make sense to put the name of the resource and namespace or even all of those string values in quotes by using `%q`?

internal/operator-controller/authorization/authorization.go
```diff
@@ -0,0 +1,112 @@
...
+			}
+			clusterScopedErrs = append(clusterScopedErrs, fmt.Errorf("cannot %s %s %s",
+				resAttr.Verb,
+				strings.TrimSuffix(resAttr.Resource, "s"),
+				resAttr.Name))
+		}
+	}
+	errs := append(namespacedErrs, clusterScopedErrs...)
+	if len(errs) > 0 {
+		errs = append([]error{fmt.Errorf("installer service account %s is missing required permissions", ext.Spec.ServiceAccount.Name)}, errs...)
```
azych, line 88: nit: that's not ideal from performance perspective, maybe just do return fmt.Errorf("installer service..., %s, errors.Join(errs...))?

internal/operator-controller/authorization/authorization.go
bentito: @joelanford @trgeiger was getting at that the point that the import adding `RBACAuthorizer` brings in seems to normally be `k8s.io/kubernetes` which is undesirable b/c that depends on several staging-only deps at v0.0.0. I am trying to figure out a more granular import for it to avoid the kitchen sink import?

internal/operator-controller/authorization/authorization.go
trgeiger: Wait, can you re-explain this suggestion? I thought I understood at first glance. Are you suggesting I sub in the errs into the "installer service account %s..." error message with Errorf? Like nested `fmt.Errorf(fmt.Sprintf("installer SA %s is missing perms", ext.Spec.ServiceAccountName)) + " %s", errs...)`? or something like that

internal/operator-controller/applier/helm.go
trgeiger: I'll make this change in the bits I introduce in this PR and we can maybe edit the other error returns in helm.go in a separate enhancement PR

internal/operator-controller/authorization/authorization.go
trgeiger: In considering RBACAuthorizer, I think we should probably get more team input on the addition of all the list/watch on all rbac objects in the cluster since that would be a pretty significant expansion of what OLMv1 sees, right? I'll link this thread to the olm-dev chat as well

joelanford: The more I dive into the RBACAuthorizer, the more I understand the complexity of doing this check ourselves. There are _a lot_ of edge cases. I think we may need to at least start out with using SelfSubjectAccessReview so that we don't have to take on the complexity of handling all of the nuance.

joelanford: And it looks like SSAR is not sufficient for checking bindings and role escalations... So for RBAC objects we probably need to do a dry-run create.

internal/operator-controller/applier/helm.go
azych: sure, that's actually what I meant - other places introduced by this PR

internal/operator-controller/authorization/authorization.go
azych: sure, I meant something like fmt.Errorf("installer SA... perms, %w", ext.Spec.ServiceAccountName, errors.Join(errs...))

Alternatively you could start with errs slice already having this "install service account %s" error at first position and then at the end when you see there were no actual errors, simply return a nil error

trgeiger:
Closing as we are going to implement this functionality differently

PR #1723, opened 2025-02-06
🌱 (catalogd) serveFile for all endpoint instead of serveContent

anik120:
# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/catalogd/internal/storage/localdir.go
+++ b/catalogd/internal/storage/localdir.go
@@ -210,13 +210,8 @@ func (s *LocalDirV1) handleV1All(w http.ResponseWriter, r *http.Request) {
 	defer s.m.RUnlock()
 
 	catalog := r.PathValue("catalog")
-	catalogFile, catalogStat, err := s.catalogData(catalog)
-	if err != nil {
-		httpError(w, err)
-		return
-	}
 	w.Header().Add("Content-Type", "application/jsonl")
-	http.ServeContent(w, r, "", catalogStat.ModTime(), catalogFile)
+	http.ServeFile(w, r, catalogFilePath(s.catalogDir(catalog)))
 }
 
 func (s *LocalDirV1) handleV1Query(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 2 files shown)


catalogd/internal/storage/localdir_test.go
```diff
@@ -238,7 +238,7 @@ func TestLocalDirServerHandler(t *testing.T) {
 		{
 			name:               "Server returns 404 when non-existent catalog is queried",
 			expectedStatusCode: http.StatusNotFound,
-			expectedContent:    "404 Not Found",
+			expectedContent:    "404 page not found",
```
joelanford, line 241: Oh this is interesting. I actually don't love this (because we aren't serving pages), but I also think it doesn't actually matter at all. It's just my OCD.

anik120: Right. I guess it's something we have to live with if we want to `serveFile`

catalogd/internal/storage/localdir.go
```diff
@@ -210,13 +210,8 @@ func (s *LocalDirV1) handleV1All(w http.ResponseWriter, r *http.Request) {
...
 
 	catalog := r.PathValue("catalog")
-	catalogFile, catalogStat, err := s.catalogData(catalog)
-	if err != nil {
-		httpError(w, err)
-		return
-	}
 	w.Header().Add("Content-Type", "application/jsonl")
-	http.ServeContent(w, r, "", catalogStat.ModTime(), catalogFile)
+	http.ServeFile(w, r, catalogFilePath(s.catalogDir(catalog)))
```
camilamacedo86, line 214: 👍 We are doing the same. Just cleanup.

catalogd/internal/storage/localdir_test.go
camilamacedo86: What is the motivation for using serveFile instead of ServeContent as it is? Is there any specific reason to prefer one over the other?

joelanford: Might be worth looking at the implementations to see if there really is a significant difference. If it's just a difference of who opens the file, then maybe sticking with `ServeContent` is better because then 404 errors are consistent.

joelanford: I took a look. `ServeFile` has a bunch of handling for directories, and then eventually calls `serveContent`. I think we close this and keep things as they are?

Sorry for the noise @anik120 I know I was the one that originally suggested this change.

PR #1724, opened 2025-02-06
[WIP] :sparkles: Single own namespace

perdasilva:
# Description

## Reviewer Checklist

- [ ] API Go Documentation


joelanford:
Wow, I am impressed with how simple that is with the existing code we have.

internal/util/clusterextension/clusterextension.go
```diff
@@ -0,0 +1,24 @@
...
+)
+
+// GetWatchNamespace determines the watch namespace the ClusterExtension should use
+// Note: this is a temporary artifice to enable gated use of single/own namespace install modes
+// for registry+v1 bundles. This will go away once the ClusterExtension API is updated to include
+// (opaque) runtime configuration.
+func GetWatchNamespace(ext *ocv1.ClusterExtension) string {
+	if features.OperatorControllerFeatureGate.Enabled(features.SingleOwnNamespaceInstallSupport) {
+		if ext != nil && ext.Annotations[clusterExtensionWatchNamespaceAnnotation] != "" {
+			return ext.Annotations[clusterExtensionWatchNamespaceAnnotation]
```
joelanford, line 20: Should we fail if the annotation value contains a comma so that we don't sign ourselves up for MultiNamespace support?

LalatenduMohanty:
@joelanford @perdasilva Are we planning to write a RFC for this feature `Single own namespace`. Also Is this PR related to https://github.com/operator-framework/operator-controller/issues/593 in some way. In that case, can we update the epic description too.

internal/util/clusterextension/clusterextension.go
perdasilva: yeah - I was thinking about that. Ultimately, this will go away, though. And all the user will be saying is that it wants the operator to watch namespace "n1,ns2,ns3,ns4" XD. This was just the first iteration to get some tangible experience with the whole thing. I also wanted to run through going from own to single (it worked insofar as the generated role went to the new watch namespace =D)

hack/demo/own-namespace-demo.sh
```diff
@@ -0,0 +1,43 @@
...
+
+# check for argocd-operator rbac in watch namespace
+kubectl get roles,rolebindings -n argocd-system -o name
+
+# get controller service-account name
+kubectl get deployments -n argocd-system argocd-operator-controller-manager -o jsonpath="{.spec.template.spec.serviceAccount}"
+
+# check service account for role binding is the same as controller service-account
+rolebinding=$(kubectl get rolebindings -n argocd-system -o name | grep 'argocd-operator' | head -n 1)
+kubectl get -n argocd-system $rolebinding -o jsonpath='{.subjects}' | jq .[0]
```
camilamacedo86, line 43: WDYT about we have a doc for it, maybe for now under the drafts, and we add the demo?

internal/operator-controller/applier/helm_test.go
```diff
@@ -106,6 +110,10 @@ metadata:
     olm.properties: '[{"type":"from-csv-annotations-key", "value":"from-csv-annotations-value"}]'
 spec:
   installModes:
+    - type: SingleNamespace
+      supported: true
+    - type: OwnNamespace
+      supported: true
```
camilamacedo86, line 116: We have the demos, but would adding e2e tests within the use cases be nice? 
Then, we could either validate that the permissions are limited in a second moment.

perdasilva: I do want to add e2es for sure. But I'm not sure how we want to handle feature-gated e2e tests. I thought I would follow this PR up with the tests once we've had this discussion with the team.

hack/demo/own-namespace-demo.sh
perdasilva: done =D

hack/demo/resources/single-namespace-demo.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: olm.operatorframework.io/v1
+kind: ClusterExtension
+metadata:
+  name: argocd-operator
+  annotations:
+    # watch-namespace is different from install namespace
+    olm.operatorframework.io/watch-namespace: argocd
```
camilamacedo86, line 7: One question: Today, if the Operator is using a single-namespace mode, the namespace is already defined in the bundle-registry-v1 as an environment variable (e.g., [falcon-operator](https://github.com/redhat-openshift-ecosystem/redhat-marketplace-operators/blob/main/operators/falcon-operator-rhmp/v0.8.1/manifests/falcon-operator.clusterserviceversion.yaml#L659-L663)).

Could we not simply retrieve the ENV VAR and use it in those modes, or does it still fail if MultiNamespace is used instead? Why do we need a new annotation? Is it to allow users to switch between modes dynamically?

By the way, I recall that we previously had a check where installation would fail if targetNamespace was not AllNamespace, but I can’t find that logic now. I checked [this search](https://github.com/search?q=repo%3Aoperator-framework%2Foperator-controller%20InstallModes&type=code), but it looks like we’re no longer failing in that case. Are we handling this differently now? Where do we fail if is MultiNamespace now?

perdasilva: For the second point: in a way, yes. But the way to frame it is that we're treating this as a kind of an install-time configuration and don't want to necessarily block the user from doing it. I still think this needs extra discussion, since allowing configuration change at the API level != the runtime allowing the installation to continue with a given configuration (and the check may or may not take into account the current state of the installation).

perdasilva: Some more info on install modes and targetNamespaces (https://olm.operatorframework.io/docs/advanced-tasks/operator-scoping-with-operatorgroups/#targetnamespaces-and-their-relationship-to-installmodes)

docs/draft/howto/single-ownnamespace-install.md
```diff
@@ -0,0 +1,105 @@
...
+This feature is still in *alpha* the `SingleOwnNamespaceInstallSupport` feature-gate must be enabled to make use of it.
+See the instructions below on how to enable it.
+
+---
+
+A component of OLMv0's multi-tenancy feature is its support of four [*installModes*](https://olm.operatorframework.io/docs/advanced-tasks/operator-scoping-with-operatorgroups/#targetnamespaces-and-their-relationship-to-installmodes):
+for operator installation:
+
+ - *OwnNamespace*: If supported, the operator can be configured to watch for events in the namespace it is deployed in.
+ - *SingleNamespace*: If supported, the operator can be configured to watch for events in a single namespace that the operator is not deployed in.
```
camilamacedo86, line 13: I was wondering if we need to support this one.
Since this is the MultiNamespace case behind the scenes, the config here will watch its own namespace plus one extra namespace.
Or did I miss something?

hack/demo/generate-asciidemo.sh
```diff
@@ -3,6 +3,7 @@
 trap cleanup SIGINT SIGTERM EXIT
 
 SCRIPTPATH="$( cd -- "$(dirname "$0")" > /dev/null 2>&1 ; pwd -P )"
+export DEMO_RESOURCE_DIR="${SCRIPTPATH}/resources"
```
camilamacedo86, line 6: We have a target to update the demo: [link](https://github.com/operator-framework/operator-controller/blob/main/Makefile#L388-L391).

When we run this, will it update all demos?

internal/operator-controller/applier/helm.go
```diff
@@ -79,7 +80,7 @@ func shouldSkipPreflight(ctx context.Context, preflight Preflight, ext *ocv1.Clu
 }
 
 func (h *Helm) Apply(ctx context.Context, contentFS fs.FS, ext *ocv1.ClusterExtension, objectLabels map[string]string, storageLabels map[string]string) ([]client.Object, string, error) {
-	chrt, err := convert.RegistryV1ToHelmChart(ctx, contentFS, ext.Spec.Namespace, []string{corev1.NamespaceAll})
+	chrt, err := h.buildHelmChart(contentFS, ext)
```
camilamacedo86, line 83: 👍 I see what you're doing here.

We now call RegistryV1ToHelmChart to handle the conversion and set it in the structure.
Then, after the check/convention, we override the watch namespace based on the annotation.

However, if the CVS is set to single namespace or own namespace, the bundle will still fail, right?
We're only allowing the switch to single/own namespace at runtime via the feature flag, but we still cannot manage/applying a bundle when the CVS is explicitly set to single/own namespace.

docs/draft/howto/single-ownnamespace-install.md
```diff
@@ -0,0 +1,105 @@
...
+### Examples
+
+``` terminal title="SingleNamespace install mode example"
+kubectl apply -f - <<EOF
+apiVersion: olm.operatorframework.io/v1
+kind: ClusterExtension
+metadata:
+  name: argocd
+  annotations:
+    olm.operatorframework.io/watch-namespace: argocd-watch
```
camilamacedo86, line 74: @perdasilva based on the comment it seems that we want the same annotation name that we have in OLMV0, see: https://github.com/operator-framework/operator-controller/issues/593#issuecomment-1919858334

hack/demo/resources/single-namespace-demo.yaml
perdasilva: But the watch namespace is a parameter for the user to choose. I don't quite understand the suggestion. You might be thinking about the default (or suggested) namespace. That's some metadata on the CSV that can be used by console as the value for the install namespace to save the user having to input something. But, the user can still set the value. 

In v0 to install an operator you need to:
1. create the (install) Namespace
2. create the OperatorGroup
3. create the Subscription

The console basically automates those steps and uses the default namespace by default.

The OperatorGroup's [.spec.targetNamespaces](https://github.com/operator-framework/api/blob/master/pkg/operators/v1/operatorgroup_types.go#L60)  will then go into the `olm.targetNamespaces` annotation on the controller pod. The env var is then set using the [downward api](https://kubernetes.io/docs/concepts/workloads/pods/downward-api/).

My guess is that the console also gives you different installation options depending on the install mode support. If you want to install it in SingleNamespace, you'll need to name the namespace (not sure if the console will also create it if it doesn't exist - I'd say so though). If the user wants OwnNamespace, then console will use the install namespace. If AllNamespace, then it will leave .spec.targetNamespace empty. If MultiNamespaces, then the user will probably have to enter in a list (or select from the list of existing namespace, I'm not sure).

The point is, both install and watch (or target) namespace(s) are user inputs. Does that help clarify everything?

hack/demo/generate-asciidemo.sh
perdasilva: only the catd at the moment. I don't know that we want to update each an every demo though. Might take a long time to do. But I'd also leave that out of the scope of this PR.

docs/draft/howto/single-ownnamespace-install.md
perdasilva: I don't think we have to - the `olm.targetNamespaces` is an internal OLMv0 implementation detail. I don't think it's user facing. I also think there's value in using something different from v0 to keep both version as separate as possible the users' mind. Additionally, because it's in the plural it might give the wrong impression (i.e. that we can reference more than one). Also, this will go away once we change the `ClusterExtension` API.

internal/operator-controller/applier/helm.go
perdasilva: The install mode [validation](https://github.com/operator-framework/operator-controller/blob/main/internal/operator-controller/rukpak/convert/registryv1.go#L203) is based on the install and watch namespace inputs. If the user tries to install in `SingleNamespace` mode but only `OwnNamespace` mode is supported, validation will fail.

The logic is: 
1. take registry+v1 bundle filesystem (with CSV and all other bundle resources), install namespace, and watch namespace
2. generate a `RegistryV1` representation
3. feed that to `Convert` along with the install and watch namespace. This method generates all the plain manifest resources needed to install the bundle in the install namespace and watching the given install namespace ("" -> watch all namespaces). This includes parsing the CSV into the controller deployment(s), service account(s), and RBAC scoped depending on the install mode. This method also does all the validation to ensure that the install mode is supported by the bundle, and that it doesn't include any (currently) unsupported resources (e.g. Webhooks).
4. the set of plain manifests is transformed into a Helm Chart
5. Helm Chart is installed/upgraded.

The `Apply` method doesn't have any concept of watch namespaces. It just takes the `ClusterExtension`. Right now, we take the watch namespace input from the annotation. In the near future we'll take it directly from the `ClusterExtension` spec.

The CSV isn't set to one install mode or another. It just declares which install modes it supports. The value of watch namespace is what decides the install mode at install time. Does that make sense?

docs/draft/howto/single-ownnamespace-install.md
perdasilva: In `SingleNamespace` mode that operator will still only watch for events in a single namespace (not its own namespace + another). I.e. it's "configured to watch for events in a _single_ namespace that the operator is **not** deployed in. While under the hood the watch namespace parameter is indeed a `[]string`:

If len(watch_namespace) == 0 then AllNamespaces
If len(watch_namespace) > 1 then MultiNamespaces
If len(watch_namespace) == 1 && watch_namespace[0] == install namespace then OwnNamespace
If len(watch_namespace) == 1 && watch_namespace[0] != install namespace then SingleNamespace

perdasilva:
@camilamacedo86 the comments I don't mark as resolved I'm leaving open to make sure you understand what's going on. If you're good, please mark them as resolved.

docs/draft/howto/single-ownnamespace-install.md
camilamacedo86: Thank you for the reminder!

I was trying to recall whether the SingleNamespace mode in OLM would result in the operator watching both the installed namespace and the explicitly informed namespace to ensure the operator itself functions properly.

However, as discussed, OLM is responsible for deploying and managing the operator, so the operator does not need to watch its own namespace. Instead, it only watches the namespace specified in the OperatorGroup, and OLM ensures the operator itself is running correctly.

This clarifies that in SingleNamespace mode, the operator will watch only the informed namespace, and the installed namespace is not included unless explicitly configured.

We can close this one

internal/operator-controller/applier/helm_test.go
```diff
@@ -386,3 +443,76 @@ func TestApply_Upgrade(t *testing.T) {
...
+		testExt := &v1.ClusterExtension{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "testExt",
+				Annotations: map[string]string{
+					applier.AnnotationClusterExtensionWatchNamespace: expectedWatchNamespace,
+				},
+			},
+		}
+
+		_, _, _ = helmApplier.Apply(context.TODO(), validFS, testExt, testObjectLabels, testStorageLabels)
```
camilamacedo86, line 476: Can we add here a test to ensure that if it is MultiNamespace, then it fails when the flag is enabled?

internal/operator-controller/applier/helm_test.go
camilamacedo86, line 116: Should not those still fail when the flag is not used and only work when the feature flag is used?

docs/draft/howto/single-ownnamespace-install.md
camilamacedo86: Make sense 👍

hack/demo/resources/single-namespace-demo.yaml
camilamacedo86: If we do not modify [this part of the code](https://github.com/operator-framework/operator-controller/blob/main/internal/operator-controller/rukpak/convert/registryv1.go#L203), then I understand that:

- If we continue using the same index that includes multiple operators targeting **SingleNamespace** or **OwnNamespace** modes,  
- And I attempt to install those operators, they will **fail**,  
- The current workaround would be to **manually add the annotation** in the Extension to make it work.

Given this, I believe we also need to modify this validation, right?  
We should introduce a change that allows support for **both SingleNamespace and OwnNamespace modes**, gated under a **feature flag** to ensure compatibility such as follows 


```go
func validateTargetNamespaces(supportedInstallModes sets.Set[string], installNamespace string, targetNamespaces []string) error {
	set := sets.New[string](targetNamespaces...)
	switch {
	case set.Len() == 0 || (set.Len() == 1 && set.Has("")):
		// Default behavior: Only supports InstallModeTypeAllNamespaces
		if supportedInstallModes.Has(string(v1alpha1.InstallModeTypeAllNamespaces)) {
			return nil
		}

		// If feature flag is enabled, also support SingleNamespace and OwnNamespace
		if features.OperatorControllerFeatureGate.Enabled(features.SingleOwnNamespaceInstallSupport) {
			if supportedInstallModes.Has(string(v1alpha1.InstallModeTypeSingleNamespace)) ||
				supportedInstallModes.Has(string(v1alpha1.InstallModeTypeOwnNamespace)) {
				return nil
			}
		}

		return fmt.Errorf("supported install modes %v do not support targeting all namespaces", sets.List(supportedInstallModes))

	case set.Len() == 1 && !set.Has(""):
		if supportedInstallModes.Has(string(v1alpha1.InstallModeTypeSingleNamespace)) {
			return nil
		}
		if supportedInstallModes.Has(string(v1alpha1.InstallModeTypeOwnNamespace)) && targetNamespaces[0] == installNamespace {
			return nil
		}
	default:
		if supportedInstallModes.Has(string(v1alpha1.InstallModeTypeMultiNamespace)) && !set.Has("") {
			return nil
		}
	}

	// Default error handling if feature flag is disabled
	if !features.OperatorControllerFeatureGate.Enabled(features.SingleOwnNamespaceInstallSupport) {
		return fmt.Errorf("supported install modes %v do not support target namespaces %v", sets.List[string](supportedInstallModes), targetNamespaces)
	}

	// If feature flag is enabled, allow SingleNamespace and OwnNamespace
	return nil
}


```

LalatenduMohanty:
> The user can install a registry+v1 bundle in SingleNamespace or OwnNamespace install mode by setting the watch namespace through the olm.operatorframework.io/watch-namespace annotation on the ClusterExtension, as long as the following criteria is met:
>    bundle supports Single- and/or OwnNamespace install modes
    SingleOwnNamespaceInstallSupport feature-gate is enabled on the controller
    olm.operatorframework.io/watch-namespace annotation points to a single existing namespace


@perdasilva Does OLM version update impacts this i.e. OLM v1.1 has  SingleOwnNamespaceInstallSupport feature-gate and user is updating to OLM v1.2 ?

hack/demo/resources/single-namespace-demo.yaml
perdasilva: I don't think so. What I'm understanding from the code above is: if targetNamespaces is empty or "" and the operator doesn't support AllNamespaces, don't error if the FG is enabled and the bundle supports Single- or OwnNamespace modes. I don't see why that should be the case. 

I think the mental model of:
- targetNamespaces empty or "" implies AllNamespaces
- targetNamespace == install namespace implies SingleNamespace
- targeNamespace != installNamespace implies OwnNamespace

much clearer than:
- if targetNamespace is empty or "" and AllNamespaces is not supported implies AllNamespaces
- if targetNamespace is empty or "" and AllNamespaces is not supported and Single- or OwnNamespaces is supported implies either Single -or OwnNamespaces (with the values of targetNamespace coming from the suggested default in the CSV). What happens if it supports both Single- and OwnNamespace? Which one are we using? Or does it then become, if you set the installNamespace to the default in the CSV then it's OwnNamespace if you don't it's Single? I think this makes things more complicated with now clear added benefit. No? Or maybe I'm not seeing something?

internal/operator-controller/applier/helm_test.go
perdasilva: We kinda handle that [here](https://github.com/operator-framework/operator-controller/pull/1724/files#diff-2939d8b5276ecbb3620b1f3a113be76a4de4d85ba41087abe6cf28db93710f19R50. I've also restricted it at the API level with the [RegistryV1ToHelmChart](https://github.com/operator-framework/operator-controller/pull/1724/files#diff-4e1161250663230583e5c1839b774ef8ee35d18cc9fc5801b68516a12bd3dc5eR42) only accepting a single watchNamespace. There's no way to initiate installation in MultiNamespace mode. 

Additionally, for the applier suite I'd really only want to test integration between the applier and the converter. I.e. is the applier calling the converter correctly and is it floating up errors. For deeper testing I think it's better to focus on the Rukpak package.

internal/operator-controller/applier/helm_test.go
perdasilva: Only if you try to install a bundle in Single- OwnNamespace mode. It's ok if a bundle supports those install modes. As it is now for v1, AllNamespaces MUST be supported (but it doesn't need to be the _only_ supported install mode).

I've added another unit test for the GetWatchNamespace function behavior when the FG is off.

perdasilva:
> > The user can install a registry+v1 bundle in SingleNamespace or OwnNamespace install mode by setting the watch namespace through the olm.operatorframework.io/watch-namespace annotation on the ClusterExtension, as long as the following criteria is met:
> > bundle supports Single- and/or OwnNamespace install modes
> > SingleOwnNamespaceInstallSupport feature-gate is enabled on the controller
> > olm.operatorframework.io/watch-namespace annotation points to a single existing namespace
> 
> @perdasilva Does OLM version update impacts this i.e. OLM v1.1 has SingleOwnNamespaceInstallSupport feature-gate and user is updating to OLM v1.2 ?

No, I don't think it's the case. This PR introduces the FG.

hack/demo/resources/single-namespace-demo.yaml
camilamacedo86: The goal of this one is allow the content in OLMV0 works with OLMV1 right?
See the test: https://github.com/operator-framework/operator-controller/pull/1801
That will fails with
>  E0224 08:46:49.374502       1 controller.go:341] "Reconciler error" err="supported install modes [OwnNamespace SingleNamespace] do not support targeting all namespaces"

The above error should still occur without the feature flag but should not occur when we set it. Does that make sense? Or did I miss something?

thetechnick (approved):
lgtm

camilamacedo86 (approved):
Thank you for all the amazing explanations! Your work is truly outstanding—terrific job!
🚀

hack/demo/resources/single-namespace-demo.yaml
camilamacedo86: We checked that all fine !!!

camilamacedo86 (dismissed):
/lgtm

thetechnick (approved):

docs/draft/howto/single-ownnamespace-install.md
```diff
@@ -0,0 +1,105 @@
...
+
+ - *OwnNamespace*: If supported, the operator can be configured to watch for events in the namespace it is deployed in.
+ - *SingleNamespace*: If supported, the operator can be configured to watch for events in a single namespace that the operator is not deployed in.
+ - *MultiNamespace*: If supported, the operator can be configured to watch for events in more than one namespace.
+ - *AllNamespaces*: If supported, the operator can be configured to watch for events in all namespaces.
+
+OLMv1 will not attempt multi-tenancy (see [design decisions document](../../project/olmv1_design_decisions.md)) and will think of operators
+as globally installed, i.e. in OLMv0 parlance, as installed in *AllNamespaces* mode. However, there are operators that
+were intended only for the *SingleNamespace* and *OwnNamespace* install modes. In order to make these operators installable in v1 while they
+transition to the new model, v1 is adding support for these two new *installModes*. It should be noted that, in line with v1's no multi-tenancy policy,
```
pgodowski, line 20: As discussed in the separate venue, @joelanford , there is set of use cases, where it is expected for OLMv1 to allow multiple installation of the same operator with different scopes each, on the same cluster. Just want to make sure this PR is an incremental step towards this and not the final decision.

Issue #1712, opened 2025-02-05
Consolidate E2E Tests Between Catalogd and Operator-Controller


camilamacedo86:
After migrating catalogd to operator-controller, we need to consolidate the end-to-end (E2E) tests to ensure comprehensive coverage and maintain test reliability. This involves integrating catalogd’s existing tests into operator-controller’s E2E suite, eliminating redundancy, and ensuring consistency across components.

PR #1726, opened 2025-02-07
🌱 [Monorepo]: Move e2e tests from catalogd to operator-controller

camilamacedo86:
Motivated by: https://github.com/operator-framework/operator-controller/issues/1712


test/catalogd-e2e/unpack_test.go
```diff
@@ -79,7 +79,7 @@ var _ = Describe("ClusterCatalog Unpacking", func() {
 			actualFBC, err := ReadTestCatalogServerContents(ctx, catalog, c, kubeClient)
 			Expect(err).To(Not(HaveOccurred()))
 
-			expectedFBC, err := os.ReadFile("../../testdata/catalogs/test-catalog/expected_all.json")
+			expectedFBC, err := os.ReadFile("../../catalogd/testdata/catalogs/test-catalog/expected_all.json")
```
camilamacedo86, line 82: We cannot move now the testdata/catalogs to the root because it will broke the sync:
- https://github.com/openshift/release/blob/master/ci-operator/config/openshift/operator-framework-operator-controller/openshift-operator-framework-operator-controller-main.yaml#L26-L27

catalogd/Makefile
```diff
@@ -78,17 +78,9 @@ FOCUS := $(if $(TEST),-v -focus "$(TEST)")
...
-
-e2e: KIND_CLUSTER_NAME := catalogd-e2e
-e2e: ISSUER_KIND := Issuer
-e2e: ISSUER_NAME := selfsigned-issuer
-e2e: KUSTOMIZE_OVERLAY := config/overlays/e2e
-e2e: run image-registry test-e2e kind-cluster-cleanup ## Run e2e test suite on local kind cluster
 
 image-registry: ## Setup in-cluster image registry
-	./test/tools/imageregistry/registry.sh $(ISSUER_KIND) $(ISSUER_NAME)
+	./../test/tools/imageregistry/registry.sh $(ISSUER_KIND) $(ISSUER_NAME)
```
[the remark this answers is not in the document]
camilamacedo86: In a follow-up, we can probably combine both. But for now, I felt better keep in this away because they are not the same code implementation. 

See

```

.PHONY: image-registry
E2E_REGISTRY_IMAGE=localhost/e2e-test-registry:devel
image-registry: export GOOS=linux
image-registry: export GOARCH=amd64
image-registry: ## Build the testdata catalog used for e2e tests and push it to the image registry
	go build $(GO_BUILD_FLAGS) -tags '$(GO_BUILD_TAGS)' -ldflags '$(GO_BUILD_LDFLAGS)' -gcflags '$(GO_BUILD_GCFLAGS)' -asmflags '$(GO_BUILD_ASMFLAGS)' -o ./testdata/registry/bin/registry ./testdata/registry/registry.go
	go build $(GO_BUILD_FLAGS) -tags '$(GO_BUILD_TAGS)' -ldflags '$(GO_BUILD_LDFLAGS)' -gcflags '$(GO_BUILD_GCFLAGS)' -asmflags '$(GO_BUILD_ASMFLAGS)' -o ./testdata/push/bin/push         ./testdata/push/push.go
	$(CONTAINER_RUNTIME) build -f ./testdata/Dockerfile -t $(E2E_REGISTRY_IMAGE) ./testdata
	$(CONTAINER_RUNTIME) save $(E2E_REGISTRY_IMAGE) | $(KIND) load image-archive /dev/stdin --name $(KIND_CLUSTER_NAME)
	./testdata/build-test-registry.sh $(E2E_REGISTRY_NAMESPACE) $(E2E_REGISTRY_NAME) $(E2E_REGISTRY_IMAGE) <- Call another implementation 
```` 


```
catalogd-image-registry: ## Setup in-cluster image registry
./test/tools/imageregistry/registry.sh $(ISSUER_KIND) $(ISSUER_NAME)
```

File: Makefile [hunk 3]
```diff
@@ -302,7 +322,13 @@ go-build-linux: export GOARCH=amd64
 go-build-linux: $(BINARIES)
 
 .PHONY: run
-run: docker-build kind-cluster kind-load kind-deploy #HELP Build the operator-controller then deploy it into a new kind cluster.
+run: generate docker-build kind-cluster kind-load kind-deploy wait #HELP Build the operator-controller then deploy it into a new kind cluster.
```
LalatenduMohanty, line 341: I am curious why `generate` was not part of the `run` before?

camilamacedo86: It was for catalogd but not for oper-con
But, I think we should right have that

dtfranz: It does actually run already as a dependency, it was hard to find it but:
`docker-build` runs `build-linux`, which runs `build-deps`, which runs `manifests` and `generate`.

Makefile
```diff
@@ -302,7 +338,12 @@ go-build-linux: export GOARCH=amd64
...
```
camilamacedo86, line 341:
```suggestion
run: docker-build kind-cluster kind-load kind-deploy wait #HELP Build the operator-controller then deploy it into a new kind cluster.
```

Makefile
camilamacedo86: I removed

Makefile
```diff
@@ -209,6 +213,39 @@ test-e2e: KUSTOMIZE_BUILD_DIR := config/overlays/e2e
 test-e2e: GO_BUILD_FLAGS := -cover
 test-e2e: run image-registry e2e e2e-coverage kind-clean #HELP Run e2e test suite on local kind cluster
```
dtfranz, line 214: If I'm reading the Makefile correctly, we have:
`test-e2e` - runs `operator-controller` e2e only
`catalogd-e2e` - runs `catalogd` e2e only

I think we need some more clear separation between the e2e for the two components. I'd prefer this as a follow-up if we do change it, but IMO we should have:
`test-e2e` - runs **both** `operator-controller` and `catalogd` e2e
`catalogd-e2e` - runs`catalogd` e2e only
`operator-controller-e2e` - runs `operator-controller` e2e only

We could then have the github actions run the `catalogd-e2e` and `operator-controller-e2e` in two separate jobs as it does now, but with slightly more clear job names. WDYT?

Again though, don't do that on this PR if you agree, I think this one is big enough already :)

dtfranz (approved):
Left a comment on the Makefile above but otherwise /lgtm!

Makefile
camilamacedo86: HI @dtfranz 

I also want :-) That is what I expected we have at some point:

> test-e2e - runs both operator-controller and catalogd e2e
> upgrade-e2e runs for both

But that needs to be follow ups.
We need to do step by step

- We need change the tests to no longer use Omega Ginkgo
- Then, we can start to began to combine
- We will need to combine as well the image registry and going on

We need to split. Otherwise, it is too much for a single go

perdasilva (dismissed):
The PR takes another step in the direction of merging operator-controller and catalogd by moving the catalogd tools and e2e tests under the top-level test and tools directories. Only the required changes to the catalogd Makefile and e2e GHA were made in order to ensure the tests are still being correctly executed by CI. Follow-ups will continue the work of merging the projects.

Awesome work <3 thank you!

grokspawn:
apidiff is an e2e test change, unavoidable as we merge the directory hierarchies.

Issue #1707, opened 2025-02-04
Consolidate code from catalogd/inernal  in to operataor-controller/internal


LalatenduMohanty:
Currently the code of catalogd/internal and operataor-controller/internal exists separately. We need to look in to what we can consolidate in one place i.e. operataor-controller/internal

[1] https://github.com/operator-framework/operator-controller/tree/main/catalogd/internal
[2] https://github.com/operator-framework/operator-controller/tree/main/internal

camilamacedo86:
HI @azych 

The idea here is to move the pkg code from catalogd to the root (operator-controller)
we still need to use internal because we do not want to expose our code. Otherwise, we need to support all related to our internal code as well since someone could start to use it.

LalatenduMohanty:
Let me look in to this and get back here with my suggestion , you can review and let me know if that sounds good.

camilamacedo86:
It seems to me that it will end up being exactly the same as what we have in the catalog, just in the root directory. However, if we encounter any issues, we'll need to discuss them as they arise. At this point, it seems difficult to come up with anything more before we try it out (IHMO)

joelanford:
Putting aside the question of how to consolidate catalogd and operator-controller, I've had a few more fundamental feelings recently:
- Do we have too many packages directly under `internal`?
- Are we inconsistent where some directories in `internal` are packages, but some are just parent directories that have packages somewhere further down?
- Do we have good names for our packages? Are they well-organized.

I agree with @azych that it might be helpful to look at everything holistically and decide if we can drive some consistency as part of this exercise:

Current listing of our internal trees (just directories):
```
internal
├── action
│   └── error
├── applier
├── authentication
├── bundleutil
├── catalogd
├── catalogmetadata
│   ├── cache
│   ├── client
│   ├── compare
│   └── filter
├── conditionsets
├── contentmanager
│   ├── cache
│   └── source
│       └── internal
├── controllers
├── features
├── finalizers
├── httputil
├── labels
├── resolve
├── rukpak
│   ├── convert
│   │   └── testdata
│   │       └── combine-properties-bundle
│   │           ├── manifests
│   │           └── metadata
│   ├── operator-registry
│   ├── preflights
│   │   └── crdupgradesafety
│   ├── source
│   └── util
├── scheme
├── shared
├── util
│   ├── fs
│   └── image
└── version
catalogd/internal
├── controllers
│   └── core
├── features
├── garbagecollection
├── metrics
├── serverutil
[...]

PR #1728, opened 2025-02-07
🌱 (cleanup): remove duplication of version implementation

camilamacedo86:
Before catalogd, the operator-controller requires a code implementation to set the Version. However, now both projects are the same and have the same version. Therefore, there is no longer a reason to keep both.

Motivated by: https://github.com/operator-framework/operator-controller/issues/1707


camilamacedo86:
@joelanford 

WDYT? Can we move already with this one 
OR do we want create a subdir to add shared libs between both?

camilamacedo86:
@joelanford

WDYT? Can we move forward with this one already?

An alternative solution would be to create a sub-directory like shared, as @azych suggested, and place the version there. However, IMHO, all libraries might be used for both.

The only component-specific part is the controllers. In fact, the feature gates could be shared as well. Given that, I’d argue for centralizing everything rather than introducing a more complex structure.

That said, I’ll defer to you since you might have a different long-term vision.

c/c @LalatenduMohanty

joelanford (approved):

PR #1729, opened 2025-02-07
WIP: Enhance catalogd cache by using sync.Map and adds logger

camilamacedo86:
- Adds logger to track cache updates, misses, and removals
- Replaced map-based cache tracking with `sync.Map` to improve thread safety, reduce lock contention, and optimize concurrent access
- Added new test TestFilesystemCacheConcurrentAccess verifies that the catalogd cache can handle concurrency
- Prevents unnecessary deletion attempts when the cache dir doesn’t exist

### DIFF
--- a/test/e2e/metrics_test.go
+++ b/test/e2e/metrics_test.go
@@ -1,234 +1,186 @@
+// Package e2e contains end-to-end tests to verify that the metrics endpoints
+// for both components. Metrics are exported and accessible by authorized users through
+// RBAC and ServiceAccount tokens.
+//
+// These tests perform the following steps:
+// 1. Create a ClusterRoleBinding to grant necessary permissions for accessing metrics.
+// 2. Generate a ServiceAccount token for authentication.
+// 3. Deploy a curl pod to interact with the metrics endpoint.
+// 4. Wait for the curl pod to become ready.
+// 5. Execute a curl command from the pod to validate the metrics endpoint.
+// 6. Clean up all resources created during the test, such as the ClusterRoleBinding and curl pod.
+//
+//nolint:gosec
 package e2e
 
 import (
 	"bytes"
 	"io"
 	"os/exec"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/require"
+
+	"github.com/operator-framework/operator-controller/test/utils"
 )
 
-// nolint:gosec
 // TestOperatorControllerMetricsExportedEndpoint verifies that the metrics endpoint for the operator controller
-// is exported correctly and accessible by authorized users through RBAC and a ServiceAccount token.
-// The test performs the following steps:
-// 1. Creates a ClusterRoleBinding to grant necessary permissions for accessing metrics.
-// 2. Generates a ServiceAccount token for authentication.
-// 3. Deploys a curl pod to interact with the metrics endpoint.
-// 4. Waits for the curl pod to become ready.
-// 5. Executes a curl command from the pod to validate the metrics endpoint.
-// 6. Cleans up all resources created during the test, such as the ClusterRoleBinding and curl pod.
 func TestOperatorControllerMetricsExportedEndpoint(t *testing.T) {
-	var (
-		token   string
-		curlPod = "curl-metrics"
-		client  = ""
-		clients = []string{"kubectl", "oc"}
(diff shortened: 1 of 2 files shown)


internal/catalogmetadata/cache/cache.go
```diff
@@ -8,18 +8,21 @@ import (
...
 
 var _ client.Cache = &filesystemCache{}
 
-func NewFilesystemCache(cachePath string) *filesystemCache {
+func NewFilesystemCache(cachePath string, logger logr.Logger) *filesystemCache {
 	return &filesystemCache{
 		cachePath:              cachePath,
 		mutex:                  sync.RWMutex{},
-		cacheDataByCatalogName: map[string]cacheData{},
+		cacheDataByCatalogName: sync.Map{},
```
joelanford, line 24: `sync.Map`'s lack of type safety is a concern. Could we hold off on this change until `sync/v2` package lands with a generic `sync.Map`?

internal/catalogmetadata/cache/cache.go
```diff
@@ -104,13 +114,16 @@ func (fsc *filesystemCache) writeFS(catalogName string, source io.Reader) (fs.FS
 		return nil, err
 	}
 
-	if err := os.RemoveAll(cacheDir); err != nil {
-		return nil, fmt.Errorf("error removing old cache directory: %v", err)
+	if _, err := os.Stat(cacheDir); err == nil {
```
joelanford, line 117: This seems unnecessary since `os.RemoveAll` already tolerates the case that the cache dir doesn't exist.

internal/catalogmetadata/cache/cache.go
joelanford: Also it looks like the existing mutex is still in place (for other reasons?), so the extra locking if the sync.Map doesn't buy us anything.

internal/catalogmetadata/cache/cache.go
```diff
@@ -64,12 +68,18 @@ func (fsc *filesystemCache) Put(catalogName, resolvedRef string, source io.Reade
...
 		cacheFS, errToCache = fsc.writeFS(catalogName, source)
 	}
-	fsc.cacheDataByCatalogName[catalogName] = cacheData{
+	fsc.cacheDataByCatalogName.Store(catalogName, cacheData{
 		Ref:   resolvedRef,
 		Error: errToCache,
+	})
+
+	if errToCache != nil {
+		fsc.logger.Error(errToCache, "Cache update failed", "catalog", catalogName, "ref", resolvedRef)
```
joelanford, line 77: Would this error ultimately propagate back up to to the reconciler and be reported back to the user? If so, seems duplicative to log it since it will show up in our logs anyway. Maybe add context to the returned error instead?

internal/catalogmetadata/cache/cache.go
```diff
@@ -64,12 +68,18 @@ func (fsc *filesystemCache) Put(catalogName, resolvedRef string, source io.Reade
...
+		return nil, errToCache
 	}
 
-	return cacheFS, errToCache
+	fsc.logger.Info("Cache updated successfully", "catalog", catalogName, "ref", resolvedRef)
```
joelanford, line 81: Do we have any particular concerns with the need for more visibility into the inner workings of what is going on inside this cache that motivates the extra logging?

This feels like it could start to make our logs pretty noisy. I'd suggest setting this to log at level 4 to keep our standard logging a little bit leaner.

internal/catalogmetadata/cache/cache.go
camilamacedo86: It seems that os.RemoveAll costs more than  os.Stat(. 
That was the motivation.

joelanford: Is there a benchmark that can show the difference? Even if os.RemoveAll is significantly slower:
1. How often will we be running this function?
2. When this function runs, how often will cacheDir not exist?

camilamacedo86: You are totally right, it will depends on how much times we call writeFS.
I have nothing of that for now, I am just looking to see how we can improve the cache. 
It is just an experimental staff.

camilamacedo86:
It seems that not so much here that will be valid
So closing it out.

PR #1731, opened 2025-02-07
🌱 Monorepo pt2: fully consolidate image pull/cache implementations

joelanford:
# Description

This PR is a follow-up from #1690. It completely removes the `source` packages that existed in both `catalogd` and `operator-controller`, eliminates the `Unpacker` abstraction that is unused and leftover from rukpak's support for multiple remote fetch protocols, adds new common abstractions for an image puller and an image cache, adds a new `containers/image` puller implementation, and adds a new disk-based cache implementation.

This is still a work-in-progress because the unit tests need to be consolidated and de-duplicated. Because the tests were written with the unpacker interface in mind, they also couple the pull and cache concepts together, so more work can be done to tease those tests apart and make them more appropriate for the new interfaces and implementations.

## Reviewer Checklist

- [ ] API Go Documentation


catalogd/internal/controllers/core/clustercatalog_controller.go
```diff
@@ -216,50 +222,56 @@ func (r *ClusterCatalogReconciler) reconcile(ctx context.Context, catalog *catal
...
 	r.storedCatalogs[catalog.Name] = storedCatalogData{
-		unpackResult:       *unpackResult,
+		ref:                canonicalRef,
+		lastUnpack:         unpackTime,
+		lastSuccessfulPoll: lastSuccessfulPoll,
 		observedGeneration: catalog.GetGeneration(),
 	}
 	r.storedCatalogsMu.Unlock()
-	return nextPollResult(unpackResult.LastSuccessfulPollAttempt.Time, catalog), nil
+	return nextPollResult(time.Now(), catalog), nil
```
joelanford, line 274:
```suggestion
	return nextPollResult(lastSuccessfulPoll, catalog), nil
```

catalogd/internal/controllers/core/clustercatalog_controller_test.go
```diff
@@ -818,16 +818,10 @@ func TestPollingRequeue(t *testing.T) {
...
 				Client: nil,
-				Unpacker: &MockSource{result: &source.Result{
-					State: source.StateUnpacked,
-					FS:    &fstest.MapFS{},
-					ResolvedSource: &catalogdv1.ResolvedCatalogSource{
-						Image: &catalogdv1.ResolvedImageSource{
-							Ref: "my.org/someImage@someSHA256Digest",
-						},
-					},
-					LastSuccessfulPollAttempt: tc.lastPollTime,
```
joelanford, line 829: Hmm, this was not carried forward, but it didn't seem to cause tests to fail. Need to look more into this.

catalogd/internal/controllers/core/clustercatalog_controller_test.go
```diff
@@ -875,14 +869,13 @@ func TestPollingReconcilerUnpack(t *testing.T) {
...
+	successfulStoredCatalogData := func(lastPoll time.Time) map[string]storedCatalogData {
 		return map[string]storedCatalogData{
 			"test-catalog": {
 				observedGeneration: successfulObservedGeneration,
-				unpackResult: source.Result{
-					ResolvedSource:            successfulUnpackStatus().ResolvedSource,
-					LastSuccessfulPollAttempt: lastPoll,
-				},
+				ref:                mustRef(t, "my.org/someimage@sha256:"+oldDigest),
+				lastUnpack:         time.Time{},
```
joelanford, line 877: These values came from `successfulUnpackStatus().ResolvedSource` (and now I have them hardcoded here). I should check and see if changing them in that function causes this test to fail.

catalogd/internal/controllers/core/clustercatalog_controller_test.go
```diff
@@ -927,7 +920,7 @@ func TestPollingReconcilerUnpack(t *testing.T) {
 				},
 				Status: successfulUnpackStatus(),
 			},
-			storedCatalogData: successfulStoredCatalogData(metav1.Now()),
+			storedCatalogData: successfulStoredCatalogData(time.Now()),
```
joelanford, line 923: These `metav1.Now()` to `time.Now()` changes are there because our image handling libraries should not be coupled to kubernetes-isms. I moved conversion to `metav1.Time` into the reconciler code, iirc.

cmd/operator-controller/main.go
```diff
@@ -306,13 +306,14 @@ func main() {
...
-		BaseCachePath: filepath.Join(cachePath, "unpack"),
-		SourceContextFunc: func(logger logr.Logger) (*types.SystemContext, error) {
+	imageCache := imageutil.BundleCache(filepath.Join(cachePath, "unpack"))
+	imagePuller := &imageutil.ContainersImagePuller{
+		SourceCtxFunc: func(ctx context.Context) (*types.SystemContext, error) {
 			srcContext := &types.SystemContext{
 				DockerCertPath: pullCasDir,
 				OCICertPath:    pullCasDir,
 			}
+			logger := logr.FromContextOrDiscard(ctx)
```
joelanford, line 316: Need to change this to be consistent with other `log.FromContext` (which we get from controller-runtime). This seems like a really easy thing to mess up. I wonder if we can implement another linter for this somehow.

internal/controllers/clusterextension_controller_test.go
```diff
@@ -196,70 +195,10 @@ func TestClusterExtensionResolutionSuccessfulUnpackFails(t *testing.T) {
 	}
 }
 
-func TestClusterExtensionUnpackUnexpectedState(t *testing.T) {
-	cl, reconciler := newClientAndReconciler(t)
-	reconciler.Unpacker = &MockUnpacker{
-		result: &source.Result{
-			State: "unexpected",
```
joelanford, line 203: Because the rukpak `source.Result` struct is gone and there is no concept of unpack state anymore, this entire test is no longer relevant.

internal/controllers/suite_test.go
```diff
@@ -39,26 +41,21 @@ import (
...
 
-// MockUnpacker is a mock of Unpacker interface
-type MockUnpacker struct {
-	err    error
-	result *source.Result
+type MockImagePuller struct {
+	err     error
+	fsys    fs.FS
+	modTime time.Time
+	ref     reference.Canonical
```
joelanford, line 51: There may still be some duplicate mock structs floating around. I fixed up tests individually in catalogd and operator-controller and then consolidated, so I may have accidentally left duplicate puller and cache mocks. Will double check.

internal/resolve/catalog.go
```diff
@@ -117,7 +117,7 @@ func (r *CatalogResolver) Resolve(ctx context.Context, ext *ocv1.ClusterExtensio
 			predicates = append(predicates, successorPredicate)
 		}
 
-		// Apply the predicates to get the candidate bundles
+		// Store the predicates to get the candidate bundles
```
joelanford, line 120: IDE refactor gone wrong:
```suggestion
		// Apply the predicates to get the candidate bundles
```

internal/util/image/disk_cache.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+func BundleCache(basePath string) Cache {
+	return &diskCache{
+		basePath: basePath,
+		filterFunc: func(_ context.Context, _ reference.Named, _ ocispecv1.Image) (archive.Filter, error) {
+			return forceOwnershipRWX(), nil
+		},
+	}
+}
```
joelanford, line 54: These filter funcs are literally the only difference between catalog and bundle image unpacking. :)

internal/util/image/disk_cache.go
```diff
@@ -0,0 +1,144 @@
...
+
+func (a *diskCache) idPath(id string) string {
+	return filepath.Join(a.basePath, id)
+}
+
+func (a *diskCache) unpackPath(id string, digest digest.Digest) string {
+	return filepath.Join(a.idPath(id), digest.String())
+}
+
+func (a *diskCache) Store(ctx context.Context, id string, srcRef reference.Named, canonicalRef reference.Canonical, imgCfg ocispecv1.Image, layers iter.Seq[LayerData]) (fs.FS, time.Time, error) {
```
joelanford, line 86: First use of a go1.23 iterator!

I implemented this function this way in order to eliminate the cache's knowledge of the puller's internals. The puller provides the `layers` iterator to the cache. And the cache simply needs to iterate the layers.

internal/util/image/filters.go
joelanford, line 1: These are not new. They existed before in `layers.go`. Only change in this PR was to unexport them.

internal/util/image/pull.go
```diff
@@ -0,0 +1,265 @@
...
+	GarbageCollect(context.Context, string, reference.Canonical) error
+}
+
+var insecurePolicy = []byte(`{"default":[{"type":"insecureAcceptAnything"}]}`)
+
+type ContainersImagePuller struct {
+	SourceCtxFunc func(context.Context) (*types.SystemContext, error)
+}
+
+func (p *ContainersImagePuller) Pull(ctx context.Context, id string, ref string, cache Cache) (fs.FS, reference.Canonical, time.Time, error) {
```
joelanford, line 53: Should probably rename `id` to `owner` everywhere it gets passed around. If two different owners request the same image, we'll end up pulling the image separately. This is by design so that if/when we introduce more configuration in our APIs around image pulling, we can make sure that each ClusterExtension is sandboxed from the other and is unable to re-use an image that another ClusterExtension pulled. This is especially important, for example, if two ClusterExtensions ask for the same private image, but only one ClusterExtension has permission to pull it.


```suggestion
func (p *ContainersImagePuller) Pull(ctx context.Context, owner string, ref string, cache Cache) (fs.FS, reference.Canonical, time.Time, error) {
```

internal/util/image/pull_bundle_test.go
joelanford, line 1: This file still needs to be consolidated with `pull_catalog_test.go`. We can likely re-write all the tests now that the abstractions are better separated.

joelanford: This is essentially a move/update of the `containers_image_test.go` from operator-controller's original `rukpak/source` package.

internal/util/image/pull_catalog_test.go
joelanford, line 1: This is essentially a move/update of the containers_image_test.go from catalogd's original rukpak/source package. Thankfully, it looks like it didn't change enough for GH to decide it was a different file.

testdata/push/push.go
```diff
@@ -64,7 +64,7 @@ func buildBundles(path string) (map[string]v1.Image, error) {
 		return nil, err
 	}
 	mutatedMap := make(map[string]v1.Image, 0)
-	// Apply required bundle labels
+	// Store required bundle labels
```
joelanford, line 67:
```suggestion
	// Apply required bundle labels
```

catalogd/internal/controllers/core/clustercatalog_controller_test.go
joelanford: Fixed up the test scenario to ensure the correct code path was actually being tested (and added another important test case to prove it!)

catalogd/internal/controllers/core/clustercatalog_controller_test.go
joelanford: Tweaked this code to share the same value used in the function it originally came from.

catalogd/internal/controllers/core/clustercatalog_controller.go
```diff
@@ -216,50 +222,58 @@ func (r *ClusterCatalogReconciler) reconcile(ctx context.Context, catalog *catal
...
-		}
-		baseURL := r.Storage.BaseURL(catalog.Name)
-
-		updateStatusProgressing(&catalog.Status, catalog.GetGeneration(), nil)
-		updateStatusServing(&catalog.Status, *unpackResult, baseURL, catalog.GetGeneration())
-	default:
-		panic(fmt.Sprintf("unknown unpack state %q", unpackResult.State))
+	// TODO: We should check to see if the unpacked result has the same content
+	//   as the already unpacked content. If it does, we should skip this rest
+	//   of the unpacking steps.
```
camilamacedo86, line 256: Would that be a follow-up?
Are you trying to reduce the changes?

catalogd/internal/controllers/core/clustercatalog_controller.go
```diff
@@ -216,50 +222,56 @@ func (r *ClusterCatalogReconciler) reconcile(ctx context.Context, catalog *catal
...
-	unpackResult, err := r.Unpacker.Unpack(ctx, catalog)
+	fsys, canonicalRef, unpackTime, err := r.ImagePuller.Pull(ctx, catalog.Name, catalog.Spec.Source.Image.Ref, r.ImageCache)
 	if err != nil {
 		unpackErr := fmt.Errorf("source catalog content: %w", err)
 		updateStatusProgressing(&catalog.Status, catalog.GetGeneration(), unpackErr)
 		return ctrl.Result{}, unpackErr
 	}
 
-	switch unpackResult.State {
-	case source.StateUnpacked:
```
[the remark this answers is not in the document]
camilamacedo86: I see that we are moving from ``r.Unpacker.Unpack` to `ImagePuller.Pull`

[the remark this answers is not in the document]
camilamacedo86: @joelanford 

I am looking at https://github.com/search?q=repo%3Aoperator-framework%2Foperator-controller%20StateUnpacked&type=code to see if we covered all places and it seems fine 👍

catalogd/internal/controllers/core/clustercatalog_controller.go
```diff
@@ -272,7 +286,7 @@ func (r *ClusterCatalogReconciler) getCurrentState(catalog *catalogdv1.ClusterCa
 	// Set expected status based on what we see in the stored catalog
 	clearUnknownConditions(expectedStatus)
 	if hasStoredCatalog && r.Storage.ContentExists(catalog.Name) {
-		updateStatusServing(expectedStatus, storedCatalog.unpackResult, r.Storage.BaseURL(catalog.Name), storedCatalog.observedGeneration)
+		updateStatusServing(expectedStatus, storedCatalog.ref, storedCatalog.lastUnpack, r.Storage.BaseURL(catalog.Name), storedCatalog.observedGeneration)
```
camilamacedo86, line 289: It seems that we covered all places where have before `unpackResult` https://github.com/search?q=repo%3Aoperator-framework%2Foperator-controller+unpackResult&type=code 👍

camilamacedo86 (dismissed):
/lgtm

internal/shared/util/image/pull.go
```diff
@@ -0,0 +1,268 @@
...
+		// We use the OCI layout as a temporary storage and
+		// pushing signatures for OCI images is not supported
+		// so we remove the source signatures when copying.
+		// Signature validation will still be performed
+		// accordingly to a provided policy context.
+		RemoveSignatures: true,
+	}); err != nil {
+		return nil, nil, time.Time{}, fmt.Errorf("error copying image: %w", err)
+	}
+	l.Info("pulled image")
```
tmshort, line 158: At first I thought this was missing the image ref, but it's done way up on lines 51/52.

catalogd/internal/controllers/core/clustercatalog_controller.go
```diff
@@ -216,50 +222,58 @@ func (r *ClusterCatalogReconciler) reconcile(ctx context.Context, catalog *catal
...
 		// that it is set correctly. Otherwise, we'd be unpacking again.
-		return nextPollResult(storedCatalog.unpackResult.LastSuccessfulPollAttempt.Time, catalog), nil
+		return nextPollResult(storedCatalog.lastSuccessfulPoll, catalog), nil
+	}
+
+	if catalog.Spec.Source.Type != catalogdv1.SourceTypeImage {
+		err := reconcile.TerminalError(fmt.Errorf("unknown source type %q", catalog.Spec.Source.Type))
+		updateStatusProgressing(&catalog.Status, catalog.GetGeneration(), err)
+		return ctrl.Result{}, err
+	}
```
LalatenduMohanty, line 240: So we never checked the `catalog.Spec.Source.Type` or did we move this from other place?

internal/shared/util/image/pull.go
```diff
@@ -0,0 +1,268 @@
...
+
+	fsys, modTime, err = p.applyImage(ctx, ownerID, dockerRef, canonicalRef, layoutImgRef, cache, srcCtx)
+	if err != nil {
+		return nil, nil, time.Time{}, fmt.Errorf("error applying image: %w", err)
+	}
```
tmshort, line 163: This seems like another stage, where a large comment might be useful; the old code indicated:
```
        //////////////////////////////////////////////////////
        //
        // Mount the image we just pulled
        //
        //////////////////////////////////////////////////////

**Review by tmshort — COMMENTED** [review]
Generally looks fine.

File: catalogd/cmd/catalogd/main.go [hunk 19]
```diff
@@ -177,7 +177,8 @@ func main() {
 
 	cw, err := certwatcher.New(certFile, keyFile)
 	if err != nil {
-		log.Fatalf("Failed to initialize certificate watcher: %v", err)
+		setupLog.Error(err, "failed to initialize certificate watcher")
+		os.Exit(1)
```
LalatenduMohanty, line 181: This `os.Exit` is new.

cmd/operator-controller/main.go
```diff
@@ -315,13 +315,14 @@ func main() {
 		os.Exit(1)
 	}
 
-	unpacker := &source.ContainersImageRegistry{
-		BaseCachePath: filepath.Join(cachePath, "unpack"),
-		SourceContextFunc: func(logger logr.Logger) (*types.SystemContext, error) {
+	imageCache := imageutil.BundleCache(filepath.Join(cachePath, "unpack"))
```
LalatenduMohanty, line 318: We are using `unpackCacheBasePath := filepath.Join(cacheDir, "unpack")` also in catalogd main.go . we should use a constant or variable for `unpack`.

internal/shared/util/image/filters.go
```diff
@@ -0,0 +1,65 @@
...
+func allFilters(filters ...archive.Filter) archive.Filter {
+	return func(h *tar.Header) (bool, error) {
+		for _, filter := range filters {
+			keep, err := filter(h)
+			if err != nil {
+				return false, err
+			}
+			if !keep {
+				return false, nil
+			}
```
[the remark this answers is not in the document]
joelanford: I can personally go either way on this. The condensed version, IMO, is ever so slightly harder to reason about because I would assume that most go developers will have their eye drawn primarily to `err != nil` and `return false, err` and initialize their memory model with "this code path returns an error", only then to see `!keep` and have to reason their way through to understand that `err` may or may not be `nil`.

I think I'll leave as is for clarity unless there is broad consensus to condense.

internal/shared/util/image/pull.go
joelanford: Yeah, this is a double-edged sword probably. We can:
1. Incrementally add k/v pairs as the context fills in, meaning later `Info` and `Error` calls don't have to keep repeating that context over and over (and potentially differently)
2. Leave the incremental k/v pairs out, and repeat them every time (potentially forgetting important k/v pairs in certain places).

(1) has the problem that we may forget about the logger already being setup with k/v pairs, which might lead us to duplicating the same information with a slightly different key. (e.g. `ref` and `reference` both showing up in a log line with the same value)

I'm almost convincing myself that the tests should include tests of the structured fields of the logger, but maybe we can add that later if we discover that we aren't being consistent with our logging.

internal/shared/util/image/pull.go
```diff
@@ -0,0 +1,268 @@
...
+	//
+	// Create an OCI layout reference for the destination,
+	// where we will temporarily store the image in order
+	// to unpack it.
+	//
+	// We use the OCI layout as a temporary storage because
+	// copy.Image can concurrently pull all the layers.
+	//
+	//////////////////////////////////////////////////////
+	layoutDir, err := os.MkdirTemp("", "oci-layout-*")
```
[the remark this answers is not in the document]
joelanford: It was random before as well (the `*` just lets the caller be explicit about where the random name piece should go). By default the randomness of the name is appended, so me adding this `*` here is essentailly a no-op.

I notice that we used to include what is now the `ownerID` in the name. I'll add that back in.

catalogd/cmd/catalogd/main.go
```diff
@@ -273,14 +274,16 @@ func main() {
 		os.Exit(1)
 	}
 
-	unpackCacheBasePath := filepath.Join(cacheDir, source.UnpackCacheDir)
+	unpackCacheBasePath := filepath.Join(cacheDir, "unpack")
```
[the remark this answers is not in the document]
joelanford: This is the only place that constant was used. Nothing in the `source` package used it, nothing in the new imageutil package uses it.

catalogd/internal/controllers/core/clustercatalog_controller.go
joelanford: That used to be part of the `source.ContainersImage` unpacker code: https://github.com/operator-framework/operator-controller/blob/becde514411344e71d51e0fcb8ea8a840e6fb1a1/internal/catalogd/source/containers_image.go#L43-L49

cmd/operator-controller/main.go
joelanford: These unpack paths are decided/defined/used purely in the context of each `main.go`. It's just a coincidence/convention that they are the same. But they don't have to be.

LalatenduMohanty (approved):
/lgtm

PR #1734, opened 2025-02-10
:seedling: Refactor filter package

perdasilva:
# Description

Refactors the `catalogmetadata/filter` package by:
 - moving the predicates to `internal/util/filter`
 - renaming the `Filter` function to `RemoveInPlace` to make it clear that function mutates the input
 - adding a non-input-mutating `Filter` function
 - moving `Filter` and `RemoveInPlace` to `internal/util/slices`
 - generalizing and complementing the slice filtering/removing unit tests

This was done as slice (and other) filtering is, generally, a useful thing to have and can be used throughout the code-base. I've been using it as part of go'fied rbac rendering tool (`hacks/tools/catalog/generate-manifests`) that I made to help me with the single/own namespace testing and RBAC scoping.

## Reviewer Checklist

- [ ] API Go Documentation


camilamacedo86 (approved):
/lgtm

internal/util/slices/slices.go
```diff
@@ -0,0 +1,24 @@
...
+		}
+	}
+	return slices.Clip(out)
+}
+
+// RemoveInPlace removes all elements from s for which test returns true.
+// Elements between new length and original length are zeroed out.
+func RemoveInPlace[T any](s []T, test filter.Predicate[T]) []T {
+	return slices.DeleteFunc(s, filter.Not(test))
+}
```
joelanford, line 24: WDYT?

```suggestion
func RemoveInPlace[T any](s []T, test filter.Predicate[T]) {
	slices.DeleteFunc(s, filter.Not(test))
}
```

internal/util/slices/slices.go
```diff
@@ -0,0 +1,24 @@
...
+	for i := 0; i < len(s); i++ {
+		if test(s[i]) {
+			out = append(out, s[i])
+		}
+	}
+	return slices.Clip(out)
+}
+
+// RemoveInPlace removes all elements from s for which test returns true.
+// Elements between new length and original length are zeroed out.
```
azych, line 21: > Elements between new length and original length are zeroed out.

This is only true if you don't use the return value as you're doing in the unit tests, no? Otherwise you'd get a clipped result similar to what `Filter` would return?
If that's the case and `RemoveInPlace`/`Filter` would be all we need, should it be extracted to its own function and package?

I'm also wondering about the organization.
We're creating `util/slices` which imports `util/filter` and requires users to import `util/filter` (because of `filter.Predicate`). To me this shows that those are all tied together, so maybe it would make sense to extract them to a dedicated `internal/filter` (or just `util/filter`, though IMO if it makes sense it's still best to avoid nesting too much stuff under util) package which would encapsulate filtering logic and could be then reused where needed? 
WDYT?

azych (approved):

camilamacedo86 (approved):
/lgtm

PR #1749, opened 2025-02-12
:seedling: Move catalogd controllers and webhooks to internal/catalogd

LalatenduMohanty:
# Description

## Reviewer Checklist

- [ ] API Go Documentation


Makefile
```diff
@@ -115,16 +115,26 @@ tidy: #HELP Update dependencies.
 	# Force tidy to use the version already in go.mod
 	$(Q)go mod tidy -go=$(GOLANG_VERSION)
 
+.PHONY: catalogd-manifests #EXHELP Generate catalogd code and catalogd manifests.
+KUSTOMIZE_CRDS_DIR := catalogd/config/base/crd/bases
+KUSTOMIZE_RBAC_DIR := catalogd/config/base/rbac
+KUSTOMIZE_WEBHOOKS_DIR := catalogd/config/base/manager/webhook
+catalogd-manifests: $(CONTROLLER_GEN) 
+	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
```
joelanford, line 123: This codegen using `hack/boilderplate.go.txt` should move to the `generate` target. I'm guessing we actually only need to call this once for the entire repo, so we may be able to just delete this. 

You can verify by deleting the catalogd generated deepcopy code and re-running `make generate` from the root Makefile.

Makefile
```diff
@@ -115,16 +115,26 @@ tidy: #HELP Update dependencies.
 	# Force tidy to use the version already in go.mod
 	$(Q)go mod tidy -go=$(GOLANG_VERSION)
 
+.PHONY: catalogd-manifests #EXHELP Generate catalogd code and catalogd manifests.
+KUSTOMIZE_CRDS_DIR := catalogd/config/base/crd/bases
+KUSTOMIZE_RBAC_DIR := catalogd/config/base/rbac
+KUSTOMIZE_WEBHOOKS_DIR := catalogd/config/base/manager/webhook
+catalogd-manifests: $(CONTROLLER_GEN)
```
joelanford, line 122: I would suggest sticking with just `manifests` and adding catalogd's controller-gen commands to that existing recipe instead of splitting out separate component-based manifests targets. As we consolidate further, and start looking toward kustomize consolidation, we'll be able to refactor some of the controller-gen commands to generate catalogd and operator-controller manifests at the same time (e.g. the CRDs)

Makefile
```diff
@@ -115,16 +115,26 @@ tidy: #HELP Update dependencies.
 	# Force tidy to use the version already in go.mod
 	$(Q)go mod tidy -go=$(GOLANG_VERSION)
 
+.PHONY: catalogd-manifests #EXHELP Generate catalogd code and catalogd manifests.
+KUSTOMIZE_CRDS_DIR := catalogd/config/base/crd/bases
+KUSTOMIZE_RBAC_DIR := catalogd/config/base/rbac
+KUSTOMIZE_WEBHOOKS_DIR := catalogd/config/base/manager/webhook
```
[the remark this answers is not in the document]
LalatenduMohanty: Sounds good to me. Will fix it.

Makefile
LalatenduMohanty: Either way is fine with me. Will fix it

LalatenduMohanty: Fixed in commit https://github.com/operator-framework/operator-controller/commit/d06df7ff15ac685729983f5fc8890a923da1aeea

Makefile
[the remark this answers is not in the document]
LalatenduMohanty: Fixed in https://github.com/operator-framework/operator-controller/commit/d06df7ff15ac685729983f5fc8890a923da1aeea

Makefile
LalatenduMohanty: Fixed in https://github.com/operator-framework/operator-controller/commit/d06df7ff15ac685729983f5fc8890a923da1aeea

LalatenduMohanty:
I had to remove the `build-deps` target from the catalogd Makefile as well because it had dependancy on the manifest generation target. Next I would move the `build` target of catalogd Makefile.

camilamacedo86 (approved):
/lgtm

LalatenduMohanty:
@joelanford PTAL

catalogd/Makefile
```diff
@@ -98,18 +88,15 @@ export GO_BUILD_TAGS     := containers_image_openpgp
 
 BUILDCMD = go build -tags '$(GO_BUILD_TAGS)' -ldflags '$(GO_BUILD_LDFLAGS)' -gcflags '$(GO_BUILD_GCFLAGS)' -asmflags '$(GO_BUILD_ASMFLAGS)' -o $(BUILDBIN)/$(notdir $@) ./cmd/$(notdir $@)
 
-.PHONY: build-deps
-build-deps: generate
-
 .PHONY: build go-build-local $(BINARIES)
-build: build-deps go-build-local ## Build binaries for current GOOS and GOARCH.
+build: go-build-local ## Build binaries for current GOOS and GOARCH.
```
joelanford, line 92: This is a little bit unfortunate, as it requires that the root Makefile has already run it's `generate` target. But since this is just temporary, and we don't have any CI or dev flows where we would be running `make` tasks directly in the `catalogd` Makefile, I think its fine.

LalatenduMohanty: Agree. We should move rest of the Makefile soon and that would remove the gap.

PR #1773, opened 2025-02-13
🌱 Add an overlay for local development on Tilt

trgeiger:
Adds an overlay that removes the liveness and readiness probes as well as the --leader-elect flag from the operator-controller-controller-manager deployment since those will cause container restarts when the go debugger is stopped at breakpoints. Also edits the Tiltfile to use this new overlay target.

Also adds our 1 existing feature flag to the new overlay.

# Description

## Reviewer Checklist

- [ ] API Go Documentation


config/overlays/tilt-local-dev/patches/dev-deployment.yaml
```diff
@@ -0,0 +1,13 @@
...
+  value: null
+- op: replace
+  path: /spec/template/spec/containers/0/readinessProbe
+  value: null
+- op: remove
+  # remove --leader-elect so container doesn't restart during breakpoints
+  path: /spec/template/spec/containers/0/args/2
+- op: add
+  path: /spec/template/spec/containers/0/args/-
+  value: --feature-gates=PreflightPermissions=true
```
perdasilva, line 13: probably ok for now, but we may want to start to think about how to handle this when we have multiple fgs. Maybe one patch / fg?

perdasilva (approved):

config/overlays/tilt-local-dev/patches/dev-deployment.yaml
camilamacedo86: It is a good thought

camilamacedo86 (approved):

config/overlays/tilt-local-dev/patches/dev-deployment.yaml
trgeiger: Yeah that's a good idea. I also just wasn't sure if we wanted to auto enable the feature flags for dev env or have individual devs manually write patches for them locally but thought i'd just throw this one in and see. I can separate it out into its own patch now before merge if you want, but you also approved it so can hold off until we have another one in the future.

grokspawn: Saw this yesterday afternoon, but it was too late in the day for me to say "hey please also add APIV1MetasHandler for catalogd". :) 

IMHO merge this and then I'll add that separately.  Would be good for me to learn more about overlays anyway.

PR #1774, opened 2025-02-14
:sparkles: Add installer policy rule generators

perdasilva:
# Description

Adds functions to the rukpak/convert package to generate policy rules that can be used by the installer service account / user to manage a cluster extension.

## Reviewer Checklist

- [ ] API Go Documentation


internal/operator-controller/rukpak/convert/installer_rbac.go
```diff
@@ -0,0 +1,166 @@
...
+		out = append(out, fn(k, v))
+	}
+	return out
+}
+
+func isClusterScopedResource(o client.Object) bool {
+	return slices.Contains(clusterScopedResources, o.GetObjectKind().GroupVersionKind().Kind)
+}
+
+func isNamespaceScopedResource(o client.Object) bool {
```
perdasilva, line 138: since we expect the rendered manifests as input, I'm wondering if we could just check whether `.metadata.namespace` is set as the discriminant....

internal/operator-controller/rukpak/convert/registryv1.go
```diff
@@ -29,6 +29,10 @@ import (
 	"github.com/operator-framework/operator-controller/internal/operator-controller/rukpak/util"
 )
 
+const (
+	AnnotationRegistryV1GeneratedManifest = "io.operatorframework.olm.generated-manifest"
```
camilamacedo86, line 33: So it is a new annotation
Could you please help me understand ( low verbose overview ) 
So, if we add this I am as an Cluster extension User will be able to ...

perdasilva: This is something we can talk about. This annotation exists only to inform the client of `convert.Convert` which of the resources where generated by it as opposed to coming directly from the bundle. Other than that, it does nothing. It's really just a signal to say "this resource was generated when rendering the bundle - it is not part of the bundle itself"

internal/operator-controller/rukpak/convert/installer_rbac.go
```diff
@@ -0,0 +1,166 @@
...
+		"ConfigMap",
+		"ServiceAccount",
+		"Service",
+		"Role",
+		"RoleBinding",
+		"PrometheusRule",
+		"ServiceMonitor",
+		"PodDisruptionBudget",
+		"VerticalPodAutoscaler",
+		"Deployment",
```
camilamacedo86, line 44: Hi @perdasilva 👋

Thanks for the PR! I just want to make sure I fully understand the implications of this change.

With this update, it looks like we're now generating the RBAC automatically, rather than asking authors to manually add it—am I reading that right? If so, are we aligned on this direction?

More specifically:

Are we comfortable with the system creating permissions automatically?

Do we currently auto-generate any permissions (e.g., for Secrets, ServiceAccounts, etc.)?

If yes, then this seems like a natural extension of what we already do.

If not, has this approach been discussed or agreed upon somewhere?

If the team is aligned and this was already considered, then I’m totally fine with it and happy to move forward with merging 🚀

Thanks again!

perdasilva:
closing as stale

PR #1776, opened 2025-02-14
[Draft] Handle interrupted helm releases in applier

azych:
# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/operator-controller/applier/helm.go
+++ b/internal/operator-controller/applier/helm.go
@@ -154,6 +154,19 @@ func (h *Helm) Apply(ctx context.Context, contentFS fs.FS, ext *ocv1.ClusterExte
 
 func (h *Helm) getReleaseState(cl helmclient.ActionInterface, ext *ocv1.ClusterExtension, chrt *chart.Chart, values chartutil.Values, post postrender.PostRenderer) (*release.Release, *release.Release, string, error) {
 	currentRelease, err := cl.Get(ext.GetName())
+	// if release is pending at this point, that should mean we were likely
+	// interruped and would normally enter into reconciliation error loop
+	// try to avoid that by attempting to uninstall the pending release
+	// normal reconciliation should be able to successfully pick up from here next time
+	//
+	// TODO: potentially move this to avoid failing at Upgrade() and having to go
+	// through another reconciliation attempt
+	if err == nil && currentRelease.Info.Status.IsPending() {
+		_, err = cl.Uninstall(ext.GetName())
+		if err != nil {
+			return nil, nil, StateError, fmt.Errorf("failed attempting to uninstall pending release: %w", err)
+		}
+	}
 	if errors.Is(err, driver.ErrReleaseNotFound) {
 		desiredRelease, err := cl.Install(ext.GetName(), ext.Spec.Namespace, chrt, values, func(i *action.Install) error {
 			i.DryRun = true


azych:
@joelanford WDYT? are there any additional considerations with that approach?

joelanford:
@azych, my main concern is whether helm allows any objects from the release manifest to be applied while still in the pending phase. If the answer is "yes", I'm concerned that rolling back could mean that some workload from the `to` version may have run some migration that makes it impossible to successfully go back to the `from` version. 

Maybe the better option would be to:
1. Leave the existing half-rolled-out release manifests objects that have already been applied in place on the cluster.
2. Directly delete the release secret(s) (not `helm uninstall/rollback`) that says it is pending. (We have a custom helm release storage driver that shards a release across multiple secrets if necessary to overcome etcd size limits).
3. Try again, the idea being that helm is able to essentially pick up where it left off without rolling back to the previous revision.

azych:
@joelanford I agree that this might be a valid concern in non-installation scenarios, though I also wonder that if we'd ever end up in an interrupted state that would prevent doing a successful rollback, couldn't the same/similar issue prevent us from actually progressing with a smooth upgrade when we remove release state info via secret(s) deletion instead?
I honestly don't feel knowledgeable enough about helm internals to be able to tell one way or the other at this point.

However, I followed your idea and so far just doing some basic testing it seems to be able to resolve the interruption regardless of the specific action that was interrupted (installation/upgrade).

internal/operator-controller/applier/helm.go
```diff
@@ -220,3 +262,20 @@ func (p *postrenderer) Run(renderedManifests *bytes.Buffer) (*bytes.Buffer, erro
 	}
 	return &buf, nil
 }
+
+func (h *Helm) deleteReleaseSecrets(ctx context.Context, releaseName string) error {
+	return h.secretsClientGetter.Secrets(h.systemNamespace).DeleteCollection(
```
joelanford, line 267: This makes assumptions about the underlying storage driver that we should probably avoid. I'm not sure everything is plumbed through in a usable way from the `ActionClientGetter` so we may need a change in `helm-operator-plugins`, but what would be ideal would be to find a way to get access to the `action.Configuration` created here: https://github.com/operator-framework/helm-operator-plugins/blob/3d11ac3f0107f9553dbb3c0060a54f9684c95b2d/pkg/client/actionclient.go#L167

And then use it's `.Releases.Delete()` to delete the the underlying release

azych: I agree that this would be preferable and less error-prone since even though we control the storage driver we initialize the `ActionConfigGetter` with first, we'd have to track those changes in two places if they ever happen.

With getting access to `action.Configuration` I think we should avoid exposing it via `actionClient` as a whole, as it's currently private in both `helm-operator-plugins` and `helm` making it immutable and predictable in the context of `ActionClientFor` and all the actual actions. 
Because `helm-operator-plugins` action client already handles `Release` objects and has an `Uninstall`, adding something like `DeleteMetadata(releaseName)` that calls `.Releases.Delete()` would make the most sense to me.
WDYT?


Sidenote:
Theoretically we also have access to `ActionConfigGetter` which we initialize in `main` and then use to initialize the `actionClient`, but I don't it's a viable option here because it would also mean making assumptions that we do not use the `Memory` driver for example, in which case our storages would be different each time we call `ActionConfigFor` and that is the first call `ActionClientFor` makes.

azych:
> > Couldn't the same/similar state issue prevent us from actually progressing with a smooth upgrade when we remove release state info via secret(s) deletion instead?
> 
> I think there is always the possibility of failing to move forward. The primary cases I can think of for that are:

true, but what I meant was that if an interruption leaves an in-flight action in such a broken state that we wouldn't be able to rollback because of it, there is a chance the same broken state would prevent moving forward when use a different workaround. I'd differentiate a 'broken state' caused by interruption and a 'failed' state that can definitely happen

internal/operator-controller/applier/helm.go
joelanford: > something like DeleteMetadata(releaseName) that calls .Releases.Delete()

I kinda feel like that particular interface method would put us on a slippery slope of stuffing more and more into the ActionClient interface. Maybe I'm wrong and this will be the last one, but we did also recently add `History` method.

I'm kinda tempted to put a `Config() (*action.Configuration)` escape hatch method in instead, which would definitely prevent the slippery slope. But that does somewhat have the downside that it returns a mutable `action.Configuration`.

But! If you trace `ActionClientGetter -> ActionConfigGetter -> ActionConfigFor`, you'll see that we build a new `action.Configuration` struct during every reconcile anyway. So yes, we would return a mutable `action.Configuration`, but we throw _that_ one away when we are finished with `reconcile`.

joelanford:
> true, but what I meant was that if an interruption leaves an in-flight action in such a broken state that we wouldn't be able to rollback because of it

I see what you mean. My unsubstantiated hunch is that failures during upgrades will largely fall into two buckets
1. Failure because everything was going smoothly and the process was interrupted (e.g. context was cancelled, connection with API server was lost, etc.).
2. Failure because of anything else

IMO, (1) isn't _really_ a failure. It's an indefinite pause to the progression of the upgrade that the helm client doesn't deal with well. I'm primarily concerned with recovering gracefully where we are able to simply pick up where we left off, and what I'm hoping is that there's a way to convince helm to do that. 

My original proposal of "if the most recent release secret is pending, delete it" would rely on helm's adoption logic. What I would envision happening is that the next time through (after deleting the latest pending release), we would:
1. See the "installed" release
2. Do a dry-run upgrade, and detect a diff because the previous upgrade attempt did not complete its updates/patches
3. Start a real upgrade
   - Any objects with the same name/namespace in existing/upgrade releases that we applied successfully during the first attempt would likely see a no-op update/patch
   - Any objects that are _new_ in the upgrade release that we applied successfully during the first attempt would have the correct labels that would allow helm to "adopt" the object into the release
   - Any objects that we didn't get to before the interruption would be created/updated as usual.

internal/operator-controller/applier/helm.go
azych: To me this still has that drawback of exposing a whole mutable config which comes from the fact that we'll be making it available via library method. While I agree it doesn't really matter for us, it might for other users of this lib.

Having said that I'll go with it. At this point every option has some drawback(s) and I don't think we are going to find a perfect solution.

joelanford:
One question that came to mind for me after we wrapped our discussion today: Will we be able to distinguish between "actual failure" and "resumable interruption"? I think that distinction matters because we may want to handle these scenarios differently.

Another thing that came to mind (going back to the assumption I was talking about in the discussion). If we see a pending upgrade, would it make sense to try to complete that particular upgrade to that particular bundle before we attempt to resolve a new bundle?

Consider this scenario:
- `v1` is "deployed"
- `v2` upgrade started, but didn't finish
- `v3` is now available in the catalog, there is an upgrade edge from `v1`, but not `v2`

Since we partially rolled out `v2` (or maybe we fully rolled it out, but our network connection failed right as we were about to set the release secret to "deployed"), it seems like we should _not_ proceed to `v3` because we have to assume that whatever it is that makes `v2` not upgradeable to `v3` was included in the partial rollout.

What if the scenario changes where `v3` upgrades from both `v1` and `v2`? Would it be safe _then_ to "resume" the upgrade but instead go to `v3`? I feel like there are scenarios there where the answer is still no. 

For example: suppose there are two datasets managed by the operator's manifests, where:
1. v1 has two workloads that individually migrate their respective datasets to level 1
2. v2 has two workloads that individually migrate their respective datasets to level 2
3. v3 expects the datasets to be at the same migration level, either both datasets at level 1 or both datasets at level 2, but not a mix.

If v2 partially rolls out such that one dataset gets migrated, but the other doesn't, then we can't migrate to `v3`. It seems reasonable that a `v2 -> v3` upgrade edge has the implicit pre-condition that `v2` is healthy (by our best possible definition of healthy).

Therefore, it seems like the only way to do this safely is to complete the upgrade to `v2` before we try to do anything else. In theory, the pending release secret has all of the bookkeeping necessary to actually try to resume `v2`, even if the `v2` bundle is no longer available.

Thoughts?

azych:
> One question that came to mind for me after we wrapped our discussion today: Will we be able to distinguish between "actual failure" and "resumable interruption"? I think that distinction matters because we may want to handle these scenarios differently.

Looking at what actually happens within `helm/pkg/action/upgrade.go:Upgrade()`, once the in-flight release is created in storage with a 'pending' state (end of `performUpgrade()`) and the actual upgrade logic starts, in case there was an error at any step, that 'pending' state will be replaced by `failRelease()` and error will also be returned to the caller. 

So while resuming the interrupted release does not mean it will succeed (or succeeded already), it should mean that it did not fail up to the point of interruption, although theoretically the interruption could still happen just before the 'pending' state is replaced and recorded, in which case failure will be 'masked' by 'pending' state.

> 
> Another thing that came to mind (going back to the assumption I was talking about in the discussion). If we see a pending upgrade, would it make sense to try to complete that particular upgrade to that particular bundle before we attempt to resolve a new bundle?
> ...
> Therefore, it seems like the only way to do this safely is to complete the upgrade to `v2` before we try to do anything else. In theory, the pending release secret has all of the bookkeeping necessary to actually try to resume `v2`, even if the `v2` bundle is no longer available.

Theoretically and if we can (and want) to detect those new upgrade paths I could also see attempting to do a rollback as a valid option. That way, if there were upgrade path changes, we can do a fresh start taking those into consideration. If not and assuming a rollback will be successful and is properly set up (not sure how concerned we should be if it is) we simply retry. 

> 
> Thoughts?

My general thoughts about this go back to "available" workarounds for this issue that are present in the community (eg. reading those threads I linked to in the doc comment). What's repeatedly being mentioned there is:
* uninstall/rollback (my initial approach)
* pending release secret deletion
With some people also bringing up manually editing 'pending' status to 'failure' or 'success' (if we can detect it)

At this point I don't really have a 'clear winner' here which I could say is the absolute best way to approach it with. All of those are really 'hacks' (maybe uninstall/rollback less so) that just work around the problem and the more we discuss it it seems we can come up with more and more scenarios where each of those workarounds might out perfectly and we might not be able to resolve the situation to a 'sane' state.

My question would be do we really need to ensure it? Should we dig deeper and hope we find a bulletproof solution (which I don't think there is given that the main problem isn't really on our side)? Maybe with the effort so far, we should just consider picking one of those "established" ideas to provide a "best effort" workarounds on our side?
To be fair, I thought we kinda did that already with secret deletion https://github.com/operator-framework/operator-controller/pull/1776#discussion_r1961163821 

BTW. There has been _some_ activity around resolving this on the helm side (either in 3.x or 4), see: https://github.com/helm/helm/issues/11863 and https://github.com/helm/community/pull/354

internal/operator-controller/applier/helm.go
```diff
@@ -164,9 +165,30 @@ func (h *Helm) buildHelmChart(bundleFS fs.FS, ext *ocv1.ClusterExtension) (*char
...
+	// that pending release which should 'reset' its state communicated to helm
+	// and the next reconciliation should be able to successfully pick up from here
+	// for context see: https://github.com/helm/helm/issues/5595 and https://github.com/helm/helm/issues/7476
+	// and the discussion in https://github.com/operator-framework/operator-controller/pull/1776
+	if err == nil && currentRelease.Info.Status.IsPending() {
+		if _, err = cl.Config().Releases.Delete(currentRelease.Name, currentRelease.Version); err != nil {
+			return nil, nil, StateError, fmt.Errorf("failed removing interrupted release %q version %d metadata: %w", currentRelease.Name, currentRelease.Version, err)
+		}
+		// return error to try to detect proper state (installation/upgrade) at next reconciliation
+		return nil, nil, StateError, fmt.Errorf("removed interrupted release %q version %d metadata", currentRelease.Name, currentRelease.Version)
```
camilamacedo86, line 187: That is the way that we all discussed in the team, right? 
So, we delete the latest release and try again 👍

internal/operator-controller/applier/helm.go
```diff
@@ -164,9 +165,30 @@ func (h *Helm) buildHelmChart(bundleFS fs.FS, ext *ocv1.ClusterExtension) (*char
...
+	// (installation/upgrade) we were attempting was likely interrupted in-flight.
+	// Pending release would leave us in reconciliation error loop because helm
+	// wouldn't be able to progress automatically from it.
+	//
+	// one of the workarounds is to try and remove helm metadata relating to
+	// that pending release which should 'reset' its state communicated to helm
+	// and the next reconciliation should be able to successfully pick up from here
+	// for context see: https://github.com/helm/helm/issues/5595 and https://github.com/helm/helm/issues/7476
+	// and the discussion in https://github.com/operator-framework/operator-controller/pull/1776
+	if err == nil && currentRelease.Info.Status.IsPending() {
```
camilamacedo86, line 182: I have only one question:
Should we like to see it pending for X time, or if it's pending, then we are 100% sure it fails in the scenario?

cmd/operator-controller/main.go
```diff
@@ -411,6 +411,7 @@ func run() error {
 		ActionClientGetter:  acg,
 		Preflights:          preflights,
 		BundleToHelmChartFn: convert.RegistryV1ToHelmChart,
+		SystemNamespace:     cfg.systemNamespace,
```
camilamacedo86, line 414: Just a tiny question: Why we must you pass the NS now?
I do not see us using it? Is that required so we can delete the release?
If so, all fine 👍 just to confirm. 
I i

camilamacedo86 (commented):
Hi @azych 

I have just two small questions to confirm my understanding.
Otherwise, all shows great here for me 👍

internal/operator-controller/applier/helm.go
azych: yes, we discussed this in the team among the option to rollback and manually change the release state.
The idea to choose to go with deleting latest release comes mostly from this thread: https://github.com/operator-framework/operator-controller/pull/1776#discussion_r1958491722 and the changes we did in `helm-operator-plugins` to be able to facilitate this solution (https://github.com/operator-framework/helm-operator-plugins/pull/424)

internal/operator-controller/applier/helm.go
azych: if it's pending, we should be 100% sure it's has failed before and is in the interrupted/transitioning state.

This is the code for`IsPending()`
```go
// IsPending determines if this status is a state or a transition.
func (x Status) IsPending() bool {
	return x == StatusPendingInstall || x == StatusPendingUpgrade || x == StatusPendingRollback
}
```

and example status:
```go
	// StatusPendingInstall indicates that an install operation is underway.
	StatusPendingInstall Status = "pending-install"
```

cmd/operator-controller/main.go
azych: great spot, this is a leftover from one of the previous implementations, before we made the changes to `helm-operator-plugins` - I'll remove it

azych:
thanks for the review @camilamacedo86, I'd still like to hear from @joelanford before potentially moving forward with those changes 

/hold

internal/operator-controller/applier/helm.go
camilamacedo86: Great so that is safe 👍

internal/operator-controller/applier/helm.go
camilamacedo86: IHMO that is a good / safe option 👍 
I am OK with

camilamacedo86 (approved):
I am OK with the changes.
It seems that addressed all 
So,

/lgtm
/approve

and sure, wait for @joelanford LGTM since it is his baby :-)

camilamacedo86 (changes requested):
My apologies—I just noticed that the logic to **delete** the latest release is inside `getReleaseState`. This doesn’t seem ideal, as a method named `getReleaseState` should **only** retrieve the state, not modify or delete anything.  

Would it make sense to:  
- Move the deletion logic to a separate method, e.g., `handlePendingRelease`.  
- Add a brief comment to `handlePendingRelease` explaining its purpose.  
- Call `handlePendingRelease` **after `getReleaseState`**, but only if the status is **pending**.  

The expected flow:  
1. Check the release state using `getReleaseState`.  
2. If the status is pending, call `handlePendingRelease`.  

This keeps `getReleaseState` focused on retrieval while handling pending releases separately, improving clarity and maintainability.  

Let me know what you think—appreciate your help! 😊  

/approve cancel  
/lgtm cancel

camilamacedo86 (approved):
I saw that you addressed: https://github.com/operator-framework/operator-controller/pull/1776#pullrequestreview-2669289171

Thank you a lot for the great work 🥇 

So all fine for me

/lgtm
/approve

joelanford:
@camilamacedo86 Thanks for the review! For now I'm going to back out the approval so that another maintainer doesn't come and merge this before we have a chance to have a conversation about how to proceed. Our decisions here have broad impact on the safety guarantees and automation that OLM can provide, so I just want to make sure that we all agree with the ultimate direction.

The outcome of a discussion should be enshrined in an RFC to make sure we can go back and find our reference point for the ultimate decision we make since this is so central to OLM behavior.

camilamacedo86:
Thank you @joelanford 

Yep, I would update this one again with

/lgtm cancel
/approve cancel 

Based on the discussion:

I was thinking that if the chart always ends up in 'Pending' for an unknown reason, we might have a problem if we don't have the ability to simply retry X times and after that set Failed. :-(

azych:
> I was thinking that if the chart always ends up in 'Pending' for an unknown reason, we might have a problem if we don't have the ability to simply retry X times and after that set Failed. :-(

For something like that to happen (ie. release constantly ends up in 'pending-x' state), the process (generally our pod) **would have to constantly keep being interrupted (eg. by deletion) during the same very specific narrow window** of when we started helm action (installation/upgrade) and it did not finish because of the interruption, or more precisely helm did not manage to set a new status on the release other than pending - see https://github.com/operator-framework/operator-controller/pull/1776#issuecomment-2671051469).

Also per that linked comment and traced helm library code, whenever an error occurs during an action, pending state will be replaced with that error state and returned to the caller as a failed action with release status being set accordingly.
There is a theoretical possibility of a failure happening and then an immediate interruption, which could cause relase to remain in 'pending' state even though there was a failure. With approach of deleting the in-flight release, we'd run into the same failed state during next reconciliation and won't lose anything.

I think it's also important to differentiate between helm release (k8s secret in our case) state and clusterextension state, which are two different things. Whenever a clusterextension ends up with an error during reconciliation, it will be retried and an error will be placed in 'Progressing' condition type in '.Status.Conditions'. This would be same with an error which comes from a pending helm action.

perdasilva:
closing in favor of rebased version #1987 (I couldn't push the changes directly here =S)

PR #1789, opened 2025-02-18
🌱 revert temporaly changes at catalogd/testdata/

camilamacedo86:
To allow to pass: https://github.com/openshift/operator-framework-operator-controller/pull/271


tmshort:
Alternatively, for something that doesn’t need to be reverted, simply delete `COPY test-catalog /configs` from `test-catalog.Dockerfile`, and not have to restore then revert these files. That being said, it's the same number of commits

catalogd/testdata/catalogs/test-catalog.Dockerfile
```diff
@@ -1,5 +1,4 @@
 FROM scratch
-COPY test-catalog /configs
```
camilamacedo86, line 2: @tmshort that works as well :-P
I was silly. Good point

LalatenduMohanty (approved):
/lgtm

PR #1794, opened 2025-02-19
🌱 (cleanup): remove catalogd/testdata

camilamacedo86:
We no longer use it.
Now, catalogd e2e tests are executed with operator-controller one and we are using its testdata instead.


LalatenduMohanty (approved):
/lgtm

Issue #1781, opened 2025-02-15
The E2E test TestCatalogdMetricsExportedEndpoint is flaking


camilamacedo86:
The E2E test TestCatalogdMetricsExportedEndpoint is intermittently failing.

**GitHub Actions Run:**
[Failed Run](https://github.com/operator-framework/operator-controller/actions/runs/13336997934/job/37254318982)

**Error Log Snippet:**

```sh
=== RUN   TestCatalogdMetricsExportedEndpoint
    utils.go:21: Finding kubectl client
    utils.go:26: Using "kubectl" as k8s client
    metrics_test.go:100: Creating ClusterRoleBinding catalogd-metrics-binding in namespace olmv1-system
    metrics_test.go:110: Generating ServiceAccount token at namespace olmv1-system
    metrics_test.go:119: Creating curl pod (olmv1-system/catalogd-curl-metrics) to validate the metrics endpoint
    metrics_test.go:146: Waiting for the curl pod to be ready
    metrics_test.go:151: Validating the metrics endpoint
    metrics_test.go:161: Cleaning up resources
panic: test timed out after 10m0s
```

**Potential Causes:**

- The curl pod may not be reaching a ready state within the expected time.
- Cleanup of resources might be blocking test completion. (race condition)

**Steps to Reproduce:**

- Run make e2e multiple times.
- Observe that the test intermittently times out.

PR #1795, opened 2025-02-19
:bug: Add checks to catch race-driven panics in metrics endpoint e2e

bentito:
# Description

There’s potential for flakiness due to timing issues. For example, the TestOperatorControllerMetricsExportedEndpoint assumes that:
•	The ClusterRoleBinding and token are immediately effective.
•	The curl pod becomes ready promptly.
•	The metrics endpoint is available as soon as the pod is ready.

In slower or variable environments, these assumptions might not hold, leading to intermittent failures. Adding explicit waits or retry logic for each resource’s readiness could help mitigate race conditions.

To fix this, this PR:

•	Adds a generic retry helper that attempts a function up to 5 times with a 5-second delay between tries.
•	Wraps key steps in the metrics endpoint tests with the retry logic:
	•	Creating the ClusterRoleBinding.
	•	Generating the ServiceAccount token.
	•	Creating the curl pod.
	•	Validating the metrics endpoint via curl.
•	Refactored these steps to return errors instead of calling require.NoError immediately, so transient issues are retried before ultimately failing.
•	This change minimizes test flakiness due to race conditions or temporary resource unavailability.

Currently on main, on my local machine, I ran this script:

```bash
#!/bin/bash

pass_count=0
total=10

for i in $(seq 1 $total); do
    echo "Running test iteration $i/$total..."
    if make test-e2e > /dev/null 2>&1; then
        echo "Iteration $i: PASS"
        ((pass_count++))
    else
        echo "Iteration $i: FAIL"
    fi
done

echo "Summary: Passed $pass_count/$total times."
```
Failed on the 2nd try on main:
```bash
$ ./scripts/test_e2e_10_times.sh
Running test iteration 1/10...
Iteration 1: PASS
Running test iteration 2/10...
Iteration 2: FAIL
Running test iteration 3/10...
^C
```
I'm running the script now on this branch and will update with my findings.
I gotta make the script report more, to make sure this is the e2e actually causing the fail, but looks like no joy :-(

run script on branch:
```bash
Running test iteration 1/10...
Iteration 1: PASS
Running test iteration 2/10...
Iteration 2: PASS
Running test iteration 3/10...
Iteration 3: FAIL
Running test iteration 4/10...
Iteration 4: PASS
Running test iteration 5/10...
Iteration 5: FAIL
Running test iteration 6/10...
```

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/test/e2e/metrics_test.go
+++ b/test/e2e/metrics_test.go
@@ -15,10 +15,13 @@ package e2e
 
 import (
 	"bytes"
+	"context"
+	"fmt"
 	"io"
 	"os/exec"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/require"
 
@@ -156,11 +159,43 @@ func (c *MetricsTestConfig) validate(token string) {
 	require.Contains(c.t, string(output), "200 OK", "Metrics endpoint did not return 200 OK")
 }
 
-// cleanup created resources
+// cleanup removes the created resources. Uses a context with timeout to prevent hangs.
 func (c *MetricsTestConfig) cleanup() {
 	c.t.Log("Cleaning up resources")
-	_ = exec.Command(c.client, "delete", "clusterrolebinding", c.clusterBinding, "--ignore-not-found=true").Run()
-	_ = exec.Command(c.client, "delete", "pod", c.curlPodName, "-n", c.namespace, "--ignore-not-found=true").Run()
+	_ = exec.Command(c.client, "delete", "clusterrolebinding", c.clusterBinding, "--ignore-not-found=true", "--force").Run()
+	_ = exec.Command(c.client, "delete", "pod", c.curlPodName, "-n", c.namespace, "--ignore-not-found=true", "--force").Run()
+
+	// Create a context with a 60-second timeout.
+	ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
+	defer cancel()
+
+	// Wait for the ClusterRoleBinding to be deleted.
+	if err := waitForDeletion(ctx, c.client, "clusterrolebinding", c.clusterBinding); err != nil {
+		c.t.Logf("Error waiting for clusterrolebinding deletion: %v", err)
+	} else {
+		c.t.Log("ClusterRoleBinding deleted")
+	}
+
+	// Wait for the Pod to be deleted.
+	if err := waitForDeletion(ctx, c.client, "pod", c.curlPodName, "-n", c.namespace); err != nil {
+		c.t.Logf("Error waiting for pod deletion: %v", err)
+	} else {
+		c.t.Log("Pod deleted")
+	}
+}
+
+// waitForDeletion uses "kubectl wait" to block until the specified resource is deleted
+// or until the 60-second timeout is reached.
+func waitForDeletion(ctx context.Context, client, resourceType, resourceName string, extraArgs ...string) error {
+	args := []string{"wait", "--for=delete", resourceType, resourceName}
+	args = append(args, extraArgs...)
+	args = append(args, "--timeout=60s")
+	cmd := exec.CommandContext(ctx, client, args...)
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("error waiting for deletion of %s %s: %v, output: %s", resourceType, resourceName, err, string(output))
+	}
+	return nil
 }
 
 // getComponentNamespace returns the namespace where operator-controller or catalogd is running


bentito:
Tries to fix #1781

test/e2e/metrics_test.go
```diff
@@ -15,16 +15,33 @@ package e2e
...
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/require"
 
 	"github.com/operator-framework/operator-controller/test/utils"
 )
 
+// retry attempts to run fn up to attempts times with delay between attempts.
+func retry(fn func() error) error {
```
joelanford, line 31: There are helper functions in a kubernetes `wait` package that we should use (if we actually need this). I think it is in apimachinery.

test/e2e/metrics_test.go
```diff
@@ -86,36 +103,63 @@ func NewMetricsTestConfig(t *testing.T, client, selector, clusterRole, clusterBi
...
+	err = retry(func() error {
+		return c.createCurlMetricsPod()
+	})
+	require.NoError(c.t, err, "Failed to create curl pod after retries")
+
+	err = retry(func() error {
+		return c.validate(token)
+	})
+	require.NoError(c.t, err, "Failed to validate metrics endpoint after retries")
+
```
joelanford, line 133: This feels very brute force and makes me think we don't actually understand the root cause of the problem yet.

I would not expect to need to retry really _any_ of this so long as we validate that the correct preconditions exist before proceeding to the next step.

Do we know where the flake failure is actually happening? Is that flake failure consistent?

joelanford:
> The ClusterRoleBinding and token are immediately effective.
> The curl pod becomes ready promptly.
> The metrics endpoint is available as soon as the pod is ready.

Are there examples in the flakes for each of these assumptions not holding? If so, I think we should look at each one to understand what's happening and then decide what the correct fix is.

test/e2e/metrics_test.go
bentito: I'll add more detail to my mult-e2e-test run script to find out.

bentito: now logging fails on my multi-run-e2e script:

https://pastebin.com/WMjR9nyV

bentito: bad news is, it's still panic failing even with the heavy handed attempts to not

camilamacedo86: I think the way that we set up things is that we start, and we run the tests,
then the metrics run in parallel. Sometimes, it seems that we clean up the cluster before the metrics tests finish.

bentito: after the most recent commit, with the wait on cleanup:

```bash
Running test iteration 1/10...
Iteration 1: PASS
Running test iteration 2/10...
Iteration 2: PASS
Running test iteration 3/10...
Iteration 3: PASS
Running test iteration 4/10...
Iteration 4: PASS
Running test iteration 5/10...
Iteration 5: PASS
Running test iteration 6/10...
Iteration 6: PASS
Running test iteration 7/10...
Iteration 7: PASS
Running test iteration 8/10...
Iteration 8: PASS
Running test iteration 9/10...
Iteration 9: PASS
Running test iteration 10/10...
Iteration 10: PASS
Summary: Passed 10/10 times.
```
I am going to revise the PR to just do what the last commit does.

test/e2e/metrics_test.go
bentito: I backed these changes out in most recent commit

test/e2e/metrics_test.go
bentito: I backed out these timing items in the most recent commit.

test/e2e/metrics_test.go
```diff
@@ -156,11 +158,21 @@ func (c *MetricsTestConfig) validate(token string) {
 	require.Contains(c.t, string(output), "200 OK", "Metrics endpoint did not return 200 OK")
 }
 
-// cleanup created resources
+// cleanup removes the created resources. Uses a context with timeout to prevent hangs.
 func (c *MetricsTestConfig) cleanup() {
 	c.t.Log("Cleaning up resources")
-	_ = exec.Command(c.client, "delete", "clusterrolebinding", c.clusterBinding, "--ignore-not-found=true").Run()
-	_ = exec.Command(c.client, "delete", "pod", c.curlPodName, "-n", c.namespace, "--ignore-not-found=true").Run()
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
LalatenduMohanty, line 164: This is interesting. What about adding a check to see if the clusterbinding and pod is actually deleted and then move forward.

bentito: yeah that might be more direct fix. I'll try it after this one shows 10/10 pass rate locally.

test/e2e/metrics_test.go
```diff
@@ -156,11 +158,21 @@ func (c *MetricsTestConfig) validate(token string) {
...
+	cmd := exec.CommandContext(ctx, c.client, "delete", "clusterrolebinding", c.clusterBinding, "--ignore-not-found=true")
+	if output, err := cmd.CombinedOutput(); err != nil {
+		c.t.Logf("Error cleaning up clusterrolebinding: %s", string(output))
+	}
+
+	cmd = exec.CommandContext(ctx, c.client, "delete", "pod", c.curlPodName, "-n", c.namespace, "--ignore-not-found=true")
+	if output, err := cmd.CombinedOutput(); err != nil {
+		c.t.Logf("Error cleaning up pod: %s", string(output))
+	}
 }
```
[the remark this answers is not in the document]
grokspawn: This repo uses merge queues with `squash and merge` mechanism, so there's currently no need for a manual squash here.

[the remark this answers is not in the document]
camilamacedo86: Hi @grokspawn 

The problem is that the git history will be very bad because will result as above 
It will combine all commit message
The message will not be clear at all about is the actual change

test/e2e/metrics_test.go
bentito: I added the checks in the latest commit, could resolve faster this way.

test/e2e/metrics_test.go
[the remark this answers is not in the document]
bentito: I think they'll get squashed on commit, til then, it's nice to see the commits as responses to raised problems?

test/e2e/metrics_test.go
```diff
@@ -86,36 +101,63 @@ func NewMetricsTestConfig(t *testing.T, client, selector, clusterRole, clusterBi
...
 
-// run will execute all steps of those tests
+// run will execute all steps of those tests with retries.
 func (c *MetricsTestConfig) run() {
-	c.createMetricsClusterRoleBinding()
-	token := c.getServiceAccountToken()
-	c.createCurlMetricsPod()
-	c.validate(token)
+	err := retry(5, 5*time.Second, func() error {
```
LalatenduMohanty: I think the retry stuff is way more than what we need for  test. Is it not working without the retry logic?

test/e2e/metrics_test.go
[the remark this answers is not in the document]
LalatenduMohanty: +1 to squash the commits, because the way current commit are , I got confused as some code got added and they were removed in the intermediate commits.

test/e2e/metrics_test.go
bentito: it's gone now

test/e2e/metrics_test.go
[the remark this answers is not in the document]
camilamacedo86: See that we have in the checklist:

Screenshot 2025-02-21 at 08 43 16

If we do not squash this one it will NOT end up with a comprehensive commit message ^

test/e2e/metrics_test.go
```diff
@@ -156,11 +159,43 @@ func (c *MetricsTestConfig) validate(token string) {
...
 
-// cleanup created resources
+// cleanup removes the created resources. Uses a context with timeout to prevent hangs.
 func (c *MetricsTestConfig) cleanup() {
 	c.t.Log("Cleaning up resources")
 	_ = exec.Command(c.client, "delete", "clusterrolebinding", c.clusterBinding, "--ignore-not-found=true").Run()
 	_ = exec.Command(c.client, "delete", "pod", c.curlPodName, "-n", c.namespace, "--ignore-not-found=true").Run()
+
+	// Create a context with a 60-second timeout.
+	ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
```
azych, line 169: 60 seconds seems like a lot. If we wanted to minimize the waiting times, running delete commands with `--force` or ``grace-period=1` might be a good idea.
It shouldn't make any difference, because at the point of cleanup we're already run the tests we needed to run.

camilamacedo86: > 60 seconds seems like a lot. If we wanted to minimize the waiting times, running delete commands with --force or grace-period=1 might be a good idea.

The timeout is set to 1 minute, but it won’t necessarily wait for 60 seconds. If the command completes successfully in 1 second, it will proceed immediately. Since the default timeout for a kubectl delete --force command is 30 seconds, I don’t think 60 seconds—being the standard in most test cases—is excessive.

> It shouldn't make any difference, because at the point of cleanup we've already run the tests we needed to run.

The tests run in parallel, right? This means that as we add more tests in the future, we can't predict the exact order in which they will be executed. Failing to clean up properly could lead to conflicts and issues with other tests.

test/e2e/metrics_test.go
```diff
@@ -156,11 +159,43 @@ func (c *MetricsTestConfig) validate(token string) {
...
+		c.t.Log("Pod deleted")
+	}
+}
+
+// waitForDeletion uses "kubectl wait" to block until the specified resource is deleted
+// or until the 30-second timeout is reached.
+func waitForDeletion(ctx context.Context, client, resourceType, resourceName string, extraArgs ...string) error {
+	args := []string{"wait", "--for=delete", resourceType, resourceName}
+	args = append(args, extraArgs...)
+	args = append(args, "--timeout=30s")
```
camilamacedo86, line 192: Since the default timeout for a kubectl delete --force command is 30 seconds, I don’t think wait here just 30 seconds is enough. Could we change it for 1 minute since mainly every place on the tests we give the grace of 1 minute?

test/e2e/metrics_test.go
azych: > > 60 seconds seems like a lot. If we wanted to minimize the waiting times, running delete commands with --force or grace-period=1 might be a good idea.
> 
> The timeout is set to 1 minute, but it won’t necessarily wait for 60 seconds. If the command completes successfully in 1 second, it will proceed immediately. Since the default timeout for a kubectl delete --force command is 30 seconds, I don’t think 60 seconds—being the standard in most test cases—is excessive.

I am aware we won't have to wait 60 seconds every time. It's still a lot if we need to just to reach the timeout.

Regarding what you said about kubectl delete options I mentioned, following kubernetes docs:
```
--force
If true, immediately remove resources from API and bypass graceful deletion. Note that immediate deletion of some resources may result in inconsistency or data loss and requires confirmation.

--grace-period int     Default: -1
Period of time in seconds given to the resource to terminate gracefully. Ignored if negative. Set to 1 for immediate shutdown. Can only be set to 0 when --force is true (force deletion).
```

My point was that we don't need to wait for a graceful shutdown when we're cleaning up and with using either of those option we should be terminating the resource(s) we want to delete almost immediately.

> > It shouldn't make any difference, because at the point of cleanup we've already run the tests we needed to run.
> 
> The tests run in parallel, right? This means that as we add more tests in the future, we can't predict the exact order in which they will be executed. Failing to clean up properly could lead to conflicts and issues with other tests.

I did not say we shouldn't do a proper cleanup. 
My point is that we can make sure that we wait as little as possible with additional options passed to delete and circumventing graceful period, which is especially important for deleting pods

camilamacedo86 (commented):
Hi @bentito,

Terrific work! 🚀 This is looking great.

Just a couple of minor suggestions:

The PR description and commit history seem to be a bit out of sync with the latest changes. It would be great to update the description accordingly and squash the commits to maintain a clean history.
The 30s timeout might be a bit short, especially considering that the --force option is used here: [link]. Since the standard timeout across tests is 1 minute, it might be better to align with that for consistency.
Other than that, everything looks good to me! 👍

Thanks for your work on this. 🎉

test/e2e/metrics_test.go
camilamacedo86: I am OK with use force and grace-period here 👍 
I just would keep context max timeout 1 minute

LalatenduMohanty: I am not a fan of using --force unless we need to use it for some reason. I would prefer to test to fail if we can not delete an object within certain time-frame. I am fine with 60 seconds as the timeout value it does not significantly changing the test duration.

test/e2e/metrics_test.go
```diff
@@ -156,11 +159,43 @@ func (c *MetricsTestConfig) validate(token string) {
...
+	} else {
+		c.t.Log("ClusterRoleBinding deleted")
+	}
+
+	// Wait for the Pod to be deleted.
+	if err := waitForDeletion(ctx, c.client, "pod", c.curlPodName, "-n", c.namespace); err != nil {
+		c.t.Logf("Error waiting for pod deletion: %v", err)
+	} else {
+		c.t.Log("Pod deleted")
+	}
```
joelanford, line 184: Do we really need to even wait for these to be deleted?

test/e2e/metrics_test.go
LalatenduMohanty: The intent of the fix is to fix the the flakiness of the test. I am not entirely sure if --force would solve that. @bentito Can you test with --force and let us know. 

I guess folks favoring the --force is confident that our test wont cause any issues which would cause deleting of the `curlPodName` pod to fail.

azych: Just for the record, my suggestion is meant to ensure that we wait as little as possible for the resources we're cleaning up (ie. deleting), as there is no need to wait for graceful deletion of any of them. 
I am _not_ saying this is going to fix the entire issue.

AFAIK the difference between `kubectl delete` and delete with force/grace-period is simply that (in case of a pod for example) the process will be killed instead of sending a standard interruption signal and having to wait for graceful shutdown.

Regarding the issue itself I am not sure about the root cause at this point.
- We are creating resources with different names for each test, so whether or not any of those haven't been deleted before we move on to next test shouldn't be a problem.
- Similarly, curl pod is run with command to sleep for 60 minutes, but when delete is called (even right now, without force/grace-period shortened) afair the default grace period should be 30 seconds before it is killed, so again, that shouldn't be the problem

Looking at the linked issue, there seems to be synchronization problems with potential deadlock. Maybe replacing `cmd.Run()` with the k8s client that we obtain when we initialize e2e test suite (available at package level as `client` could be a path worth exploring?
That would also be valuable from consistency perspective, because all of the other e2e tests seem to be using this pattern when deleting resources:

```go
{
  t.Logf("By deleting ClusterCatalog %q", cat.Name)
  require.NoError(t, c.Delete(context.Background(), cat))
  require.Eventually(t, func() bool {
  err := c.Get(context.Background(), types.NamespacedName{Name: cat.Name}, &catalogd.ClusterCatalog{})
  return errors.IsNotFound(err)
}
```

tmshort:
It happened here: 
https://github.com/operator-framework/operator-controller/actions/runs/13524917676/job/37792916935?pr=1810

test/e2e/metrics_test.go
bentito: The point is to avoid flakes. So both setting a 60 second timeout and --force make sense. 60 seconds because if we succeed fast, you'll never notice the extra time. Why fail the test due to slow responses for some reason? -- better not to. So I would like to leave that as is at 60s. I am trying --force for the same reason, get done fast and hopefully avoid flakes, testing now.

test/e2e/metrics_test.go
bentito: done

test/e2e/metrics_test.go
bentito: to be clear: I have pretty quick test -- run e2e 10 times. This test always fails on my local box when run on main (takes about ~50m. With the changes on this PR, that same test passes. I think @camilamacedo86 is right, this wait is key to the fix and we log instead of fail. I guess I could try not waiting, but I'd rather move on as something is fixing the flake.

test/e2e/metrics_test.go
bentito: Testing with force seems fine as well:

```bash
$ ./scripts/test_e2e_10_times.sh
Running test iteration 1/10...
Iteration 1: PASS
Running test iteration 2/10...
Iteration 2: PASS
Running test iteration 3/10...
Iteration 3: PASS
Running test iteration 4/10...
Iteration 4: PASS
Running test iteration 5/10...
Iteration 5: PASS
Running test iteration 6/10...
Iteration 6: PASS
Running test iteration 7/10...
Iteration 7: PASS
Running test iteration 8/10...
Iteration 8: PASS
Running test iteration 9/10...
Iteration 9: PASS
Running test iteration 10/10...
Iteration 10: PASS
Summary: Passed 10/10 times.
```

LalatenduMohanty (approved):
/lgtm

test/e2e/metrics_test.go
camilamacedo86: Previously, the code would fail when deletion was not possible, but now it simply logs the issue.

It seems that when we attempt to delete a Pod or a ClusterRoleBinding, one of the following issues occurs:

- The Pod gets stuck in the "Terminating" state, or (If so the usage of `--force` suggested by @azych could be a good solution) 
OR
- The cleanup process from another end-to-end (e2e) test runs before the deletion request, causing inconsistencies in the cluster. If so:

**This suggests a problem with how we are setting things up. In the test suite, we should properly structure the setup and teardown process.** Specifically, the suite test should be responsible for installing operator-con and catalogd, then executing all tests, and finally performing cleanup. My guess here is that the other e2e test is doing that instead of the suite test. 

I believe this is the root cause of these issues, and we may encounter even more problems as additional tests are added.

test/e2e/metrics_test.go
```diff
@@ -156,11 +159,43 @@ func (c *MetricsTestConfig) validate(token string) {
...
 }
 
-// cleanup created resources
+// cleanup removes the created resources. Uses a context with timeout to prevent hangs.
 func (c *MetricsTestConfig) cleanup() {
 	c.t.Log("Cleaning up resources")
-	_ = exec.Command(c.client, "delete", "clusterrolebinding", c.clusterBinding, "--ignore-not-found=true").Run()
-	_ = exec.Command(c.client, "delete", "pod", c.curlPodName, "-n", c.namespace, "--ignore-not-found=true").Run()
+	_ = exec.Command(c.client, "delete", "clusterrolebinding", c.clusterBinding, "--ignore-not-found=true", "--force").Run()
+	_ = exec.Command(c.client, "delete", "pod", c.curlPodName, "-n", c.namespace, "--ignore-not-found=true", "--force").Run()
```
camilamacedo86, line 166: Hi @bentito 

After we add the --force in your 10 runner tests what is happening. 
Do we still facing the issue and logging ( it was not possible delete the pod or clusterbinding ) or now the issue is no longer faced?

camilamacedo86: Anyway, I would be OK with it for now, if works does not impact us /lgtm
Just for we keep in mind in the future.

camilamacedo86 (approved):
Hi @bentito,

This might not fully address the root cause—why was it not possible to delete? The fix here mainly adds logging instead of failing, as discussed in [this comment](https://github.com/operator-framework/operator-controller/pull/1795/files#r1970467026).

That said, I don’t see much value in spending more effort on it at this point. **You did a great job in addressing the intermittent scenario**, and if needed, we can always revisit this in the future.

/lgtm
/approve

azych:
I share the sentiment - I'm still unsure about the root cause, but if this works and other people are good with it, I'm OK too.
We can always revisit.

/lgtm

test/e2e/metrics_test.go
azych: https://github.com/operator-framework/operator-controller/pull/1795#issuecomment-2684181052 - feel free to resolve this, as I can't with my permission level :/

azych (approved):

test/e2e/metrics_test.go
bentito: 👍

Issue #1744, opened 2025-02-11
Combine cmd directories for operator-controller and catalogd


tmshort:
Combine the `main.go` files for operator-controller and catalogd into a single directory structure.

tmshort:
This complains of "internal" packages when accessing `catalogd/internal`, so those must be moved first.

camilamacedo86:
I am closing this one.
It seems done already.
However, if not please feel free to re-open.

PR #1800, opened 2025-02-20
:sparkles: Consolidate api directory

tmshort:
Move operator-controller APIs into apis/operator-controller Move catalogd APIs into apis/catalogd

There's a conflict in types, both have a CatalogSource type that we may not be able to easily rename, so the simplest solution was to keep the two APIs serparate

# Description

## Reviewer Checklist

- [ ] API Go Documentation


Makefile
```diff
@@ -120,10 +120,10 @@ KUSTOMIZE_OPCON_CRDS_DIR := config/base/operator-controller/crd/bases
 KUSTOMIZE_OPCON_RBAC_DIR := config/base/operator-controller/rbac
 manifests: $(CONTROLLER_GEN) #EXHELP Generate WebhookConfiguration, ClusterRole, and CustomResourceDefinition objects.
 	# Generate the operator-controller manifests
-	rm -rf $(KUSTOMIZE_OPCON_CRDS_DIR) && $(CONTROLLER_GEN) crd paths=./api/... output:crd:artifacts:config=$(KUSTOMIZE_OPCON_CRDS_DIR)
+	rm -rf $(KUSTOMIZE_OPCON_CRDS_DIR) && $(CONTROLLER_GEN) crd paths=./api/operator-controller/... output:crd:artifacts:config=$(KUSTOMIZE_OPCON_CRDS_DIR)
 	rm -f $(KUSTOMIZE_OPCON_RBAC_DIR)/role.yaml && $(CONTROLLER_GEN) rbac:roleName=manager-role paths=./internal/operator-controller/... output:rbac:artifacts:config=$(KUSTOMIZE_OPCON_RBAC_DIR)
 	# Generate the catalogd manifests
-	rm -rf $(KUSTOMIZE_CATD_CRDS_DIR) && $(CONTROLLER_GEN) crd paths="./catalogd/api/..." output:crd:artifacts:config=$(KUSTOMIZE_CATD_CRDS_DIR)
+	rm -rf $(KUSTOMIZE_CATD_CRDS_DIR) && $(CONTROLLER_GEN) crd paths=./api/catalogd/... output:crd:artifacts:config=$(KUSTOMIZE_CATD_CRDS_DIR)
```
camilamacedo86, line 126: By consolidating the API, I thunk we no longer need to generate rules webhooks and crd spllited 
We can just 

$(CONTROLLER_GEN) rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases

Also, I think we need to stop to rm the files.
We should run the commands and the make verify should fail if has a diff instead.

Makefile
```diff
@@ -358,12 +358,12 @@ API_REFERENCE_DIR := $(ROOT_DIR)/docs/api-reference
...
 crd-ref-docs: $(CRD_REF_DOCS) #EXHELP Generate the API Reference Documents.
 	rm -f $(API_REFERENCE_DIR)/$(OPERATOR_CONTROLLER_API_REFERENCE_FILENAME)
-	$(CRD_REF_DOCS) --source-path=$(ROOT_DIR)/api \
+	$(CRD_REF_DOCS) --source-path=$(ROOT_DIR)/api/operator-controller \
 	--config=$(API_REFERENCE_DIR)/crd-ref-docs-gen-config.yaml \
 	--renderer=markdown --output-path=$(API_REFERENCE_DIR)/$(OPERATOR_CONTROLLER_API_REFERENCE_FILENAME);
 
 	rm -f $(API_REFERENCE_DIR)/$(CATALOGD_API_REFERENCE_FILENAME)
-	$(CRD_REF_DOCS) --source-path=$(ROOT_DIR)/catalogd/api \
+	$(CRD_REF_DOCS) --source-path=$(ROOT_DIR)/api/catalogd \
```
camilamacedo86, line 366: Can we not simplify here as well ./...

tmshort: Not if the API docs are in different locations, if we wish to consolidate the documentation, then that's a possibility.

Makefile
camilamacedo86: We can either to do that in a follow up.
So, has no reason to block this one

camilamacedo86 (approved):
/lgtm

camilamacedo86 (changes requested):
/lgtm cancel

We need to discuss the path for the API

tmshort:
> /lgtm cancel
> 
> We need to discuss the path for the API

As I said, I'm working on a fixup to combine it all into `api/v1`, but it requires extensive renaming and changes. One of the requirements is that the CRD outputs are placed into separate kustomize directories, associated with each manager (i.e. ClusterExtension with operator-controller, and ClusterCatalog with catalogd). This is necessary for downstream.

azych:
Given some decisions about the code layout from here: https://github.com/operator-framework/operator-controller/issues/1707#issuecomment-2643264470
I'm wondering what code specific to `catalogd/` will be left in `catalogd/` once this and then also https://github.com/operator-framework/operator-controller/issues/1744 are done?

It seems like `catalogd/` would become redundant at that point and we would have instead moved to a nested structure with 
- cmd/<project>
- api/<project>
- internal/<project>
combined Makefile etc.

camilamacedo86:
Hi @azych 

>I'm wondering what code specific to catalogd/ will be left in catalogd/ once this and then also https://github.com/operator-framework/operator-controller/issues/1744 are done?

The catalogd/ dir will no longer exist when we finish the whole work,

grokspawn (approved):

Issue #1340, opened 2024-10-03
Combine catalogd and operator-controller Makefiles


OchiengEd:
Combine the `Makefiles` in the root operator-controller repository and the one in the catalogd directory.

jsm84:
Linked to internal Jira issue EET-4418.

bentito:
Couple things I thought of when looking at this story: 
- Don't forget to update docs for renamed targets
- Don't forget to update Github actions for renamed targets
- Don't forget to update hack/helper scripts for renamed targets

camilamacedo86:
It is done I am closing

PR #1802, opened 2025-02-24
:sparkles: Move catalogd entrypoint

OchiengEd:
# Description

This pull request aims to move the catalogd entrypoint from `operator-controller/catalogd/cmd/catalogd` to `operator-controller/cmd/catalogd`.

The above action is a continuation of the monorepo work detailed in issue https://github.com/operator-framework/operator-controller/issues/1340

While this PR achieves the high level task, more work is ongoing to optimize the makefile to re-use for instance the build command.

## Reviewer Checklist

- [ ] API Go Documentation

Related to https://github.com/operator-framework/operator-controller/issues/1340


Makefile
```diff
@@ -254,7 +254,7 @@ e2e-coverage:
 .PHONY: kind-load
 kind-load: $(KIND) #EXHELP Loads the currently constructed images into the KIND cluster.
 	$(CONTAINER_RUNTIME) save $(IMG) | $(KIND) load image-archive /dev/stdin --name $(KIND_CLUSTER_NAME)
-	IMAGE_REPO=$(CATALOG_IMAGE_REPO) KIND_CLUSTER_NAME=$(KIND_CLUSTER_NAME) $(MAKE) -C catalogd kind-load
+	IMAGE_REPO=$(CATALOG_IMAGE_REPO) KIND_CLUSTER_NAME=$(KIND_CLUSTER_NAME) $(MAKE) -f Makefile.catalogd kind-load
```
camilamacedo86, line 257:
```suggestion
	$(CONTAINER_RUNTIME) save $(CATALOG_IMAGE_REPO) | $(KIND) load image-archive /dev/stdin --name $(KIND_CLUSTER_NAME)
```
We no longer need to call kind-load from the Makefile inside catalogd.
Both make run and the tests run from the root file now.

Therefore, we can follow the same approach as we did for the operator-controller image and remove the related targets from the Makefile in catalogd.

The goal is to remove the Makefile under the catalogd directory entirely.

Makefile
```diff
@@ -305,27 +305,26 @@ build-deps: manifests generate fmt
 .PHONY: build go-build-local
 build: build-deps go-build-local #HELP Build manager binary for current GOOS and GOARCH. Default target.
 go-build-local: BUILDBIN := bin
-go-build-local: $(BINARIES)
+go-build-local: $(BINARIES) go-build-catalogd-local
```
camilamacedo86, line 308: I don’t think we need the new target.
Instead, we can either build the catalogd binary in the existing target used for building the operator-controller binary.
We would follow the same process—just applying it to catalogd as well.

Makefile
```diff
@@ -305,27 +305,26 @@ build-deps: manifests generate fmt
...
 
 .PHONY: build-linux go-build-linux
 build-linux: build-deps go-build-linux #EXHELP Build manager binary for GOOS=linux and local GOARCH.
 go-build-linux: BUILDBIN := bin/linux
 go-build-linux: export GOOS=linux
 go-build-linux: export GOARCH=amd64
-go-build-linux: $(BINARIES)
+go-build-linux: $(BINARIES) go-build-catalogd-linux
```
camilamacedo86, line 315: same here

Makefile
```diff
@@ -305,27 +305,26 @@ build-deps: manifests generate fmt
...
 
 .PHONY: run
-run: docker-build kind-cluster kind-load kind-deploy wait #HELP Build the operator-controller then deploy it into a new kind cluster.
+run: docker-build kind-cluster kind-load-catalogd kind-load kind-deploy wait #HELP Build the operator-controller then deploy it into a new kind cluster.
```
camilamacedo86, line 318: we do not need new a new target
 kind-load will load both images

Makefile
```diff
@@ -389,4 +388,44 @@ deploy-docs: venv
 demo-update:
 	./hack/demo/generate-asciidemo.sh -u -n catalogd-demo catalogd-demo-script.sh
 
+# Catalogd targets
```
camilamacedo86, line 391: Instead to add the targets here
We can just add the commands under the the current targets
also, not that we have the `CATALOG_IMAGE_REPO` in this Makefile already. 
So, we should not duplicate it.

OchiengEd: Thanks. This should be resolved.

Makefile
OchiengEd: This is combined.

Makefile
OchiengEd: That should be resolved. That was just a remnnant that needed to be cleaned

Makefile
LalatenduMohanty: I would prefer separate target for catalogd but also need a target which will build both operator-controller and catalogd both. But I am not too opinionated on this. As I want to change some catalogd specific code and do not want to build operator-controller.

camilamacedo86: If you want to have the other target and it pass in the CI
I am OK with it. 👍 All fine for me.

tmshort: You can build either `catalogd` or `operator-controller` by specifying that as the target. No need for an additional target.

Makefile
```diff
@@ -24,6 +24,11 @@ IMAGE_TAG := devel
 endif
 export IMAGE_TAG
 
+CATALOGD_BINARIES=catalogd
+LINUX_BINARIES=$(join $(addprefix linux/,$(CATALOGD_BINARIES)), )
+
+BUILDCMD = go build -tags '$(GO_BUILD_TAGS)' -ldflags '$(GO_BUILD_LDFLAGS)' -gcflags '$(GO_BUILD_GCFLAGS)' -asmflags '$(GO_BUILD_ASMFLAGS)' -o $(BUILDBIN)/$(notdir $@) ./cmd/$(notdir $@)
+
```
tmshort, line 31: This shouldn't be necessary, as catalogd and operator-controller binaries should be able to use the same set of build rules/recipes.

Makefile
```diff
@@ -340,7 +345,7 @@ export GORELEASER_ARGS
 
 .PHONY: release
 release: $(GORELEASER) #EXHELP Runs goreleaser for the operator-controller. By default, this will run only as a snapshot and will not publish any artifacts unless it is run with different arguments. To override the arguments, run with "GORELEASER_ARGS=...". When run as a github action from a tag, this target will publish a full release.
-	OPERATOR_CONTROLLER_IMAGE_REPO=$(IMAGE_REPO) CATALOGD_IMAGE_REPO=$(CATALOG_IMAGE_REPO) $(GORELEASER) $(GORELEASER_ARGS)
+	OPERATOR_CONTROLLER_IMAGE_REPO=$(IMAGE_REPO) CATALOG_IMAGE_REPO=$(CATALOG_IMAGE_REPO) $(GORELEASER) $(GORELEASER_ARGS)
```
tmshort, line 348: Why this change?

Makefile
tmshort: On line 302 (297 in the original file), you should just be able to add `catalogd` to the list of binaries, so that they build with the same rules/recipies.

Makefile
OchiengEd: This was an error and has been reverted.

Makefile
OchiengEd: done.

.goreleaser.yml
```diff
@@ -20,7 +20,7 @@ builds:
       - ppc64le
```
camilamacedo86, line 20: Hi @OchiengEd 

That seems great now.
Could you please just squash the commits so that we can ensure a nice log for it when we run `git log` after it get merged? It does not seems that requires 9/10 commits 

Also, I think you can move forward with this one .. it does not need to be a draft anymore it seems good to fly 🚀

OchiengEd: Sounds good. Working on it

.goreleaser.yml
joelanford, line 1: We should rename `Dockerfile` to `Dockerfile.operator-controller` for symmetry.

Makefile
```diff
@@ -264,7 +269,7 @@ e2e-coverage:
 .PHONY: kind-load
 kind-load: $(KIND) #EXHELP Loads the currently constructed images into the KIND cluster.
 	$(CONTAINER_RUNTIME) save $(IMG) | $(KIND) load image-archive /dev/stdin --name $(KIND_CLUSTER_NAME)
```
joelanford, line 271: `IMG` should be renamed to `OPCON_IMG` or `OPERATOR_CONTROLLER_IMG` for symmetry.

OchiengEd: It is possible to introduce the `IMG_NAMESPACE` variable however, it seems that the `OPERATOR_CONTROLLER_IMAGE_REPO` and `CATALOGD_IMAGE_REPO` variables can not be removed as they would be passed to the release target in the make file.

.goreleaser.yml
OchiengEd: done

.goreleaser.yml
OchiengEd: done

Makefile
OchiengEd: done. Let me know if the current  solution works

Makefile
OchiengEd: those have been cleaned up.

Makefile
```diff
@@ -9,23 +9,28 @@ export ROOT_DIR := $(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
 
 GOLANG_VERSION := $(shell sed -En 's/^go (.*)$$/\1/p' "go.mod")
 # Image URL to use all building/pushing image targets
-ifeq ($(origin IMAGE_REPO), undefined)
-IMAGE_REPO := quay.io/operator-framework/operator-controller
+ifeq ($(origin IMG_NAMESPACE), undefined)
+IMG_NAMESPACE := quay.io/operator-framework
```
camilamacedo86, line 13: I would prefer REGISTRY
instead since we publish the image into the registry
but I do not think that it is a blocker for this one.

camilamacedo86 (approved):
/lgtm

@LalatenduMohanty @tmshort ?

tmshort:
/hold
A quick hold while I review

Makefile
tmshort: I like REGISTRY better, but we can update later

tmshort (approved):
/unhold
/lgtm

I personally prefer OPCON and CATD for some of these variable names, to make them shorter and closer in length to each other, but it's not a blocker.

Makefile
LalatenduMohanty: +1 agree with @camilamacedo86

LalatenduMohanty:
Looks like we left the dockerignore file in catalogd/ https://github.com/operator-framework/operator-controller/blob/main/catalogd/.dockerignore , sending a PR for this.

LalatenduMohanty:
PR to move the dockerignore fine https://github.com/operator-framework/operator-controller/pull/1813

PR #1803, opened 2025-02-24
:sparkles: Consolidate APIs into the same directory

tmshort:
This required renaming op-con's CatalogSource to CatalogFilter to avoid a name conflict with catd's CatalogSource

Update source files to refer to the proper API directory, and name the import reference consistently.

The docs/api-reference is now combined as well.

# Description

## Reviewer Checklist

- [ ] API Go Documentation


Makefile
```diff
@@ -118,14 +118,20 @@ KUSTOMIZE_CATD_RBAC_DIR := config/base/catalogd/rbac
...
+	mkdir $(CRD_WORKING_DIR)
+	$(CONTROLLER_GEN) crd paths="./api/v1/..." output:crd:artifacts:config=$(CRD_WORKING_DIR)
+	mv $(CRD_WORKING_DIR)/olm.operatorframework.io_clusterextensions.yaml $(KUSTOMIZE_OPCON_CRDS_DIR)
+	mv $(CRD_WORKING_DIR)/olm.operatorframework.io_clustercatalogs.yaml $(KUSTOMIZE_CATD_CRDS_DIR)
+	rmdir $(CRD_WORKING_DIR)
+	# Generate the remaining operator-controller manifests
+	$(CONTROLLER_GEN) rbac:roleName=manager-role paths="./internal/operator-controller/..." output:rbac:artifacts:config=$(KUSTOMIZE_OPCON_RBAC_DIR)
+	# Generate the remaining catalogd manifests
+	$(CONTROLLER_GEN) rbac:roleName=manager-role paths="./internal/catalogd/..." output:rbac:artifacts:config=$(KUSTOMIZE_CATD_RBAC_DIR)
+	$(CONTROLLER_GEN) webhook paths="./internal/catalogd/..." output:webhook:artifacts:config=$(KUSTOMIZE_CATD_WEBHOOKS_DIR)
```
camilamacedo86, line 134: We do not need ./api/v1..

All of this can be replaced by saying `$(CONTROLLER_GEN)` and to generate all things `rbac:roleName=manager-role crd webhook` with path `./...`

Such as originally it was: 

https://github.com/operator-framework/operator-controller/blob/d0b7376933b71eb42a2f6ca474803952a77a8f95/Makefile#L102-L104

api/v1/clusterextension_types.go
```diff
@@ -138,8 +138,8 @@ type ClusterExtensionInstallConfig struct {
 	Preflight *PreflightConfig `json:"preflight,omitempty"`
 }
 
-// CatalogSource defines the attributes used to identify and filter content from a catalog.
-type CatalogSource struct {
+// CatalogFilter defines the attributes used to identify and filter content from a catalog.
+type CatalogFilter struct {
```
camilamacedo86, line 142: Do we need to change the name of the type?
Would not it be a real breaking change in the API/end users CRD?

tmshort: Yes we do. Otherwise, there is a conflict with catalogd CatalogSource golang data structure. The first line of the commit comment says:

>This required renaming op-con's CatalogSource to CatalogFilter to avoid a name conflict with catd's CatalogSource

There is no change to the CRD, as the CRD does not reference golang data structures. You will notice that the CRDs in `config/...` are not modified.

Makefile
tmshort: True, but specifying the API directory doesn't really hurt.
However, since the RBAC and webhook are generated separately now (due to the manifest locations), it probably makes some sense to keep the specific directory.

tmshort: We can't go back to the one-liner.

camilamacedo86 (approved):
/lgtm

tmshort:
/hold
for @LalatenduMohanty

Makefile
```diff
@@ -118,14 +118,20 @@ KUSTOMIZE_CATD_RBAC_DIR := config/base/catalogd/rbac
...
-	rm -rf $(KUSTOMIZE_OPCON_CRDS_DIR) && $(CONTROLLER_GEN) crd paths=./api/operator-controller/... output:crd:artifacts:config=$(KUSTOMIZE_OPCON_CRDS_DIR)
-	rm -f $(KUSTOMIZE_OPCON_RBAC_DIR)/role.yaml && $(CONTROLLER_GEN) rbac:roleName=manager-role paths=./internal/operator-controller/... output:rbac:artifacts:config=$(KUSTOMIZE_OPCON_RBAC_DIR)
-	# Generate the catalogd manifests
-	rm -rf $(KUSTOMIZE_CATD_CRDS_DIR) && $(CONTROLLER_GEN) crd paths=./api/catalogd/... output:crd:artifacts:config=$(KUSTOMIZE_CATD_CRDS_DIR)
-	rm -f $(KUSTOMIZE_CATD_RBAC_DIR)/role.yaml && $(CONTROLLER_GEN) rbac:roleName=manager-role paths="./internal/catalogd/..." output:rbac:artifacts:config=$(KUSTOMIZE_CATD_RBAC_DIR)
-	rm -f $(KUSTOMIZE_CATD_WEBHOOKS_DIR)/manifests.yaml && $(CONTROLLER_GEN) webhook paths="./internal/catalogd/..." output:webhook:artifacts:config=$(KUSTOMIZE_CATD_WEBHOOKS_DIR)
+	mkdir $(CRD_WORKING_DIR)
+	$(CONTROLLER_GEN) crd paths="./api/v1/..." output:crd:artifacts:config=$(CRD_WORKING_DIR)
+	mv $(CRD_WORKING_DIR)/olm.operatorframework.io_clusterextensions.yaml $(KUSTOMIZE_OPCON_CRDS_DIR)
+	mv $(CRD_WORKING_DIR)/olm.operatorframework.io_clustercatalogs.yaml $(KUSTOMIZE_CATD_CRDS_DIR)
```
LalatenduMohanty, line 128: We are not deleting the older files in KUSTOMIZE_OPCON_CRDS_DIR and expecting to override the olderfiles with the new ones. It should work but it is not a full proof method. Also assuming we will not change the file names in future.

api/v1/common_types.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2022.
+Copyright 2024.
```
LalatenduMohanty, line 2: Is it necessary to change the Copyright year?

LalatenduMohanty: Found https://stackoverflow.com/questions/2390230/do-copyright-dates-need-to-be-updated, So I think we should change it to 2025?

tmshort: It's actually a new file, but git thinks it's a rename of `api/catalogd/doc.go`...
And you're right, it ought to be 2025

tmshort: We don't have a copyright policy... something we may want to talk about in one of our meetings.

PR #1811, opened 2025-02-25
🐛bugfix for operator-controller not outputting the right commit ID in the version

rashmigottipati:
# Description
The `--version` doesn't output the source code commit ID. It refers to "vcs-ref" in the image instead of "io.openshift.build.commit.id". 

Setting the `GIT_COMMIT` value to `SOURCE_GIT_COMMIT` if empty (which is a corresponding downstream variable) will ensure that the version outputs the right commit ID as the code from this repo gets copied to another repo for building purposes. 

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -308,6 +308,8 @@ VERSION := $(shell git describe --tags --always --dirty)
 endif
 export VERSION
 
+GIT_COMMIT := $(if $(SOURCE_GIT_COMMIT),$(SOURCE_GIT_COMMIT),$(shell git rev-parse HEAD))
+
 ifeq ($(origin CGO_ENABLED), undefined)
 CGO_ENABLED := 0
 endif
@@ -321,6 +323,7 @@ export GO_BUILD_GCFLAGS := all=-trimpath=$(PWD)
 export GO_BUILD_FLAGS :=
 export GO_BUILD_LDFLAGS := -s -w \
     -X '$(VERSION_PATH).version=$(VERSION)' \
+-X "$(VERSION_PATH).gitCommit=$(GIT_COMMIT)" \
 
 BINARIES=operator-controller catalogd
 
--- a/Makefile
+++ b/Makefile
@@ -308,8 +308,6 @@ VERSION := $(shell git describe --tags --always --dirty)
 endif
 export VERSION
 
-GIT_COMMIT := $(if $(SOURCE_GIT_COMMIT),$(SOURCE_GIT_COMMIT),$(shell git rev-parse HEAD))
-
 ifeq ($(origin CGO_ENABLED), undefined)
 CGO_ENABLED := 0
 endif
@@ -323,7 +321,7 @@ export GO_BUILD_GCFLAGS := all=-trimpath=$(PWD)
 export GO_BUILD_FLAGS :=
 export GO_BUILD_LDFLAGS := -s -w \
     -X '$(VERSION_PATH).version=$(VERSION)' \
--X "$(VERSION_PATH).gitCommit=$(GIT_COMMIT)" \
+    -X '$(VERSION_PATH).gitCommit=$(GIT_COMMIT)' \
 
 BINARIES=operator-controller catalogd
 
--- a/internal/shared/version/version.go
+++ b/internal/shared/version/version.go
@@ -6,7 +6,7 @@ import (
 )
 
 var (
-	gitCommit  = "unknown"
+	gitCommit  = ""
 	commitDate = "unknown"
 	repoState  = "unknown"
 	version    = "unknown"
@@ -29,7 +29,9 @@ func init() {
 	for _, setting := range info.Settings {
 		switch setting.Key {
 		case "vcs.revision":
-			gitCommit = setting.Value
+			if gitCommit == "" {
+				gitCommit = setting.Value
+			}
 		case "vcs.time":
 			commitDate = setting.Value
 		case "vcs.modified":
--- a/internal/shared/version/version.go
+++ b/internal/shared/version/version.go
@@ -6,7 +6,7 @@ import (
 )
 
 var (
-	gitCommit  = ""
+	gitCommit  = "unknown"
 	commitDate = "unknown"
 	repoState  = "unknown"
 	version    = "unknown"
@@ -29,7 +29,7 @@ func init() {
 	for _, setting := range info.Settings {
 		switch setting.Key {
 		case "vcs.revision":
-			if gitCommit == "" {
+			if gitCommit == "unknown" {
 				gitCommit = setting.Value
 			}
 		case "vcs.time":


joelanford:
If this is a downstream-only concern, is it possible to make a change in the downstream repo instead?

I'm kinda curious why the existing upstream implementation doesn't "just work" in downstreams since it should be pulling commit refs for the downstream git repo in the same way.

camilamacedo86 (dismissed):
Before this change, **I was unable to reproduce the issue scenario**. As a result, I’m struggling to understand what we are actually trying to solve a problem that does not occur. OR how are you able to reproduce the issue scenario with?  

## Steps I Took to Test Using the `main` Branch  

1. I ran `make build` to generate the binaries.  
2. I checked the latest commit using `git log`
3. I verified the version information:  

### Operator Controller  

```
$ git log -1  
commit 6cf1853f61916399cde65ef9971f66103abd6ff3 (HEAD -> main, upstream/main)  
Author: Todd Short <tmshort@users.noreply.github.com>  
Date:   Tue Feb 25 20:19:43 2025 -0500  

    Tidy Makefile (#1814)  

    Use OPCON and CATD abbreviations  

```

```
$ ./bin/operator-controller version  
version: "v1.2.0-rc4-82-g6cf1853", commit: "6cf1853f61916399cde65ef9971f66103abd6ff3", date: "2025-02-26T01:19:43Z", state: "dirty"  
```

### Catalogd  

```
$ ./bin/catalogd version  
"version: \"v1.2.0-rc4-82-g6cf1853\", commit: \"6cf1853f61916399cde65ef9971f66103abd6ff3\", date: \"2025-02-26T01:19:43Z\", state: \"dirty\""  
```

✅ **Both worked fine.**  

## Testing `main` Downstream  https://github.com/openshift/operator-framework-operator-controller

```
$ git log -1  
commit 9c4952e8d2ad4b2cd502f718786b3df3a68b5ab3 (HEAD -> main, upstream/main, origin/main, origin/HEAD)  
Merge: beab3350 3c576e57  
Author: openshift-merge-bot[bot] <148852131+openshift-merge-bot[bot]@users.noreply.github.com>  
Date:   Wed Feb 19 02:28:20 2025 +0000  

    Merge pull request #271 from openshift-bot/synchronize-upstream  

    NO-ISSUE: Synchronize From Upstream Repositories  
```  

✅ **It worked:**  

```
$ ./bin/operator-controller version  
version: "v1.2.0-rc1-2810-g4635b1c7", commit: "4635b1c79696e203d45ed89b1c8a661e0e60f931", date: "2025-02-25T01:08:03Z", state: "clean"  
```

```
$ ./bin/catalogd version  
"version: \"(devel)\", commit: \"4635b1c79696e203d45ed89b1c8a661e0e60f931\", date: \"2025-02-25T01:08:03Z\", state: \"clean\""  
```

## Next Steps (IHMO)
If the issue still occurs, we need to:  
- **(a)** Identify **where and when** it happens. Does it only affect version 4.18 but is already resolved in `main` for both upstream and downstream?  **Or is it also solved for 4.18 downstream already?** 
- **(b)** If it is fixed in `main`, we should determine **how and when** it was resolved and cherry-pick the fix or create an adapt solution based on it to fit the legacy 4.18 for catalogd repo at https://github.com/openshift/operator-framework-catalogd into older release branches downstream (**if necessary only**).  

## Conclusion  
✅ **The issue appears to be already solved in `main`. for upstream and downstream**   If not, How are you reproducing that? How are you checking that the issue occurs downstream? (We can discuss that via slack) 
🚫 **No additional changes are needed here or main downstream or any place/version that we are unable to reproduce the issue scenario**
🔍 **We only need to address the issue where it still occurs. So we need to find out how to reproduce the scenario and in what version**  

As discussed on Slack: [Slack Thread](https://redhat-internal.slack.com/archives/C0880TP2KC7/p1739962776071669?thread_ts=1739949517.452449&cid=C0880TP2KC7).

rashmigottipati:
@joelanford yeah, this is a downstream only concern. And I think it's fine when we build from the repo downstream too, but it's something to do with the way images get generated using ART. That's when it's using the env var `SOURCE_GIT_COMMIT` which is the downstream equivalent of `GIT_COMMIT`. 

We can fix this in downstream repo directly as well without fixing it here, but I thought fixing it here and sync'ing it downstream would be better to have both repos as close as possible for maintainability purposes.  Let me know if you disagree.

Makefile
```diff
@@ -303,6 +305,7 @@ export GO_BUILD_GCFLAGS := all=-trimpath=$(PWD)
 export GO_BUILD_FLAGS :=
 export GO_BUILD_LDFLAGS := -s -w \
     -X '$(VERSION_PATH).version=$(VERSION)' \
+	-X "$(VERSION_PATH).gitCommit=$(GIT_COMMIT)" \
```
camilamacedo86, line 308:
```suggestion
    -X "$(VERSION_PATH).gitCommit=$(GIT_COMMIT)" \
```

camilamacedo86: Fix the space

rashmigottipati: Fixed

Makefile
```diff
@@ -308,6 +310,7 @@ export GO_BUILD_GCFLAGS := all=-trimpath=$(PWD)
 export GO_BUILD_FLAGS :=
 export GO_BUILD_LDFLAGS := -s -w \
     -X '$(VERSION_PATH).version=$(VERSION)' \
+-X "$(VERSION_PATH).gitCommit=$(GIT_COMMIT)" \
```
camilamacedo86, line 313: Can you just ensure that it is alined with above?
Otherwise, it seems that we agreed with this one 👍

joelanford: Also nit: use single quotes rather than double quotes for consistency.

Makefile
```diff
@@ -295,6 +295,8 @@ VERSION := $(shell git describe --tags --always --dirty)
 endif
 export VERSION
 
+GIT_COMMIT := $(if $(SOURCE_GIT_COMMIT),$(SOURCE_GIT_COMMIT),$(shell git rev-parse HEAD))
+
```
joelanford, line 299: I don't think we need this line. If we (or a fork) want to build binaries with a specific GIT_COMMIT reference, we/they should explicitly set `GIT_COMMIT`

When `GIT_COMMIT` is left unset, we will also set `$(VERSION_PATH).gitCommit` to empty string. So we should:
1. Initialize `gitCommit` as `""` rather than `"unknown"` [here](https://github.com/operator-framework/operator-controller/blob/a6de9f94e085989a086459f2b5325a5cd27424b0/internal/shared/version/version.go#L9).
2. Put [this](https://github.com/operator-framework/operator-controller/blob/a6de9f94e085989a086459f2b5325a5cd27424b0/internal/shared/version/version.go#L32) in an if statement that only sets from the build data when `gitCommit == ""`.

Makefile
rashmigottipati: Done

Makefile
rashmigottipati: Addressed based on this feedback in the latest commit.

This approach seems better as it eliminates the need to introduce an openshift specific envvar upstream.

internal/shared/version/version.go
```diff
@@ -29,7 +29,9 @@ func init() {
 	for _, setting := range info.Settings {
 		switch setting.Key {
 		case "vcs.revision":
-			gitCommit = setting.Value
+			if gitCommit == "" {
```
tmshort, line 32: Couldn't this be a check against `"unknown"`, and leave the default value as-is?

tmshort: i.e. is there a _need_ to change the default?

rashmigottipati: @tmshort I am okay with leaving the default as it is, also `"unknown"` is _slightly better_ than empty string as it can be a bit ambiguous IMO.

Fixed in latest commit.

camilamacedo86 (approved):
I am OK with 👍 

@tmshort WDYT?

tmshort (approved):
/lgtm

PR #1816, opened 2025-02-26
✨ auth: use synthetic user/group when service account is not defined

joelanford:
# Description

Today at a meeting among maintainers of OLMv1, we discussed an idea that @thetechnick proposed awhile back. That is: stop using service accounts and service account tokens. Instead use synthetic names with impersonation.

While we are now 1.0.0 with support for service accounts, we can deprecate that feature and recommend attaching permissions to synthetic users/groups instead.

This PR demonstrates how we might do this. But with the API change, we should write up a detailed design and gain consensus.

This PR uses:
- User: `"olmv1:clusterextensions:<clusterExtensionName>:admin"`
- Groups: `["system:authenticated", "olmv1:clusterextensions:admin"]`

But I'm not sure this is the best setup. There's more discussion to be had around what sythentic names/groups we could derive from a cluster extension.

## Reviewer Checklist

- [ ] API Go Documentation


thetechnick:
I'd like to have a discussion on the `olmv1` prefix vs just `olm` and the `:admin` suffix before this gets merged, but otherwise this looks pretty much how I'd wanted it to look. Thanks for the work!

joelanford:
> I'd like to have a discussion on the olmv1 prefix vs just olm and the :admin suffix before this gets merged, but otherwise this looks pretty much how I'd wanted it to look. Thanks for the work!

I put next to no critical thought into the name and group names. @thetechnick you propose the following?

- User: `olm:clusterextensions:<ceName>`
- Group:`olm:clusterextensions` (and still keep `system:authenticated`)

thetechnick:
@joelanford precisely. :+1:

hack/demo/resources/synthetic-user-perms/argocd-clusterextension.yaml
```diff
@@ -0,0 +1,13 @@
...
+  name: argocd-operator
+spec:
+  namespace: argocd-system
+  serviceAccount:
+    name: "olm.synthetic-user"
+  source:
+    sourceType: Catalog
+    catalog:
+      packageName: argocd-operator
+      version: 0.6.0
```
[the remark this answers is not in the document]
perdasilva: we should probably have them ^^ added

internal/operator-controller/action/restconfig.go
```diff
@@ -6,13 +6,52 @@ import (
...
+	return func(ctx context.Context, o client.Object, c *rest.Config) (*rest.Config, error) {
+		cExt := o.(*ocv1.ClusterExtension)
+		if synthAuthFeatureEnabled && cExt.Spec.ServiceAccount.Name == syntheticServiceAccountName {
+			return m.synthUserRestConfigMapper(ctx, o, c)
+		}
+		return m.saRestConfigMapper(ctx, o, c)
+	}
+}
+
+func ClusterExtensionUserRestConfigMapper(tokenGetter *authentication.TokenGetter) func(ctx context.Context, o client.Object, c *rest.Config) (*rest.Config, error) {
```
joelanford, line 35: Should we pass a `enableSyntheticUserAuthentication` function parameter so that we can reference the feature gate only in `main.go`?

perdasilva: is that a general goal that we have that FGs should only be referenced in main? I might need to update the Single-OwnNamespace FG if that's the case. My only worry about it is that if it's only needed down the stack, we end up having to thread things all the way down, which could be painful. What is the value of having it in main? In the end I end up searching the code for usages of the FG anyway. Is it helpful in other contexts as well?

joelanford: My main thinking for this was to avoid functionality controlled by global state, which generally causes pain and can turn things into spaghetti if we aren't careful.

If we want to library-ify some of this underlying functionality, where/how to we envision the feature gates being implemented and setup? I guess I view the feature gates as attributes of the main binary, not of the libraries that the main binary uses.

cmd/operator-controller/main.go
```diff
@@ -304,7 +304,7 @@ func run() error {
 		return err
 	}
 	tokenGetter := authentication.NewTokenGetter(coreClient, authentication.WithExpirationDuration(1*time.Hour))
-	clientRestConfigMapper := action.ServiceAccountRestConfigMapper(tokenGetter)
+	clientRestConfigMapper := action.ClusterExtensionUserRestConfigMapper(tokenGetter)
```
camilamacedo86, line 307: We use the SA to generate the names at: https://github.com/operator-framework/operator-controller/blob/main/internal/operator-controller/rukpak/render/generators/generators.go#L84

How would that be after this change?

joelanford: That SA is the name of an SA that is provided in the permissions section of the CSV. This SA is the SA in the ClusterExtension spec.

Different SAs :)

internal/operator-controller/action/restconfig.go
perdasilva: then maybe the right way is to separate the implementations entirely and switch implementations depending on the FG. I'll try that.

perdasilva: Fixed ^^

perdasilva: ok - now it's good =D

config/overlays/featuregate/synthetic-user-permissions/patches/enable-featuregate.yaml
```diff
@@ -0,0 +1,4 @@
+# enable synthetic-user feature gate
+- op: add
+  path: /spec/template/spec/containers/0/args/-
+  value: "--feature-gates=SyntheticPermissions=true"
```
camilamacedo86, line 4: 👍 Cool

hack/demo/synthetic-user-cluster-admin-demo.sh
```diff
@@ -0,0 +1,30 @@
...
+kubectl apply -f ${DEMO_RESOURCE_DIR}/synthetic-user-perms/cegroup-admin-binding.yaml
+
+# install cluster extension - for now .spec.serviceAccount = "olm.synthetic-user"
+bat --style=plain ${DEMO_RESOURCE_DIR}/synthetic-user-perms/argocd-clusterextension.yaml
+
+# apply cluster extension
+kubectl apply -f ${DEMO_RESOURCE_DIR}/synthetic-user-perms/argocd-clusterextension.yaml
+
+# wait for cluster extension installation to succeed
+kubectl wait --for=condition=Installed clusterextension/argocd-operator --timeout="60s"
```
camilamacedo86, line 30: WDYT about we add something like

```
# List ClusterRoleBindings for the synthetic group
kubectl get clusterrolebindings -o json | jq -r '
  .items[] |
  select(.subjects[]? | .kind == "Group" and .name == "olm:clusterextensions") |
  "\(.metadata.name) → \(.roleRef.kind)/\(.roleRef.name)"
'

# List RoleBindings for the synthetic user
kubectl get rolebindings --all-namespaces -o json | jq -r '
  .items[] |
  select(.subjects[]? | .kind == "User" and .name == "olm:clusterextensions:argocd-operator") |
  "\(.metadata.namespace)/\(.metadata.name) → \(.roleRef.kind)/\(.roleRef.name)"
'
```

camilamacedo86: Maybe would be nice add in the demo:

> Have we here any step we could add in the demo to prove and show that with the synthetic-user-permissions the workload will have the required permission to run BUT the logged user will not able to take advantage of it to scalate the permissions?

However, it might not too be too simple. So, I am fine without it.

camilamacedo86 (approved):
Just a few nits. ( not mandatory )
It is protected behind a feature gate, so 👍 

/lgtm
/approve

I could not find any reason for not move forward with protected as it is.

PR #1817, opened 2025-02-27
🌱 e2e: make pods exit faster with `terminationGracePeriodSeconds: 0`

joelanford:
# Description

We can make our e2e's significantly faster by making sure pods stop immediately. Setting `terminationGracePeriodSeconds: 0` means that kubelet will send the kill signal immediately.

This is okay in our tests because:
1. For test bundles, we're setting it in the bundle (this is not something OLM itself is doing)
2. For the metrics pod, we are fine to kill it immediately because we already have the info we need from the `kubectl exec` call that has already completed.

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/test/e2e/metrics_test.go
+++ b/test/e2e/metrics_test.go
@@ -125,6 +125,7 @@ func (c *MetricsTestConfig) createCurlMetricsPod() {
 		"--restart=Never",
 		"--overrides", `{
 			"spec": {
+				"terminationGradePeriodSeconds": 0,
 				"containers": [{
 					"name": "curl",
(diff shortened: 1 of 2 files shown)


joelanford:
On this PR, `e2e-kind` finished in 8m.

Previous runs on `main` are typically 13-15m: https://github.com/operator-framework/operator-controller/actions/workflows/e2e.yaml?query=branch%3Amain+event%3Apush

camilamacedo86 (approved):
Good Call !!!

/lgtm

PR #1821, opened 2025-02-27
:bug: Client go version metrics endpoint test

bentito:
# Description
This PR rewrites the metrics endpoint e2e test using client-go. It maintains the recent change to the kubectl exec version by still "force" deleting the Pod (0 time grace period). Seems to pass a local 10 of 10 in a row e2e test. 

Added an optional TEST_FILTER env var to the e2e run in the Makefile. Not really related but sure is better when you're iterating on one e2e test change.

## Reviewer Checklist

- [ ] API Go Documentation


bentito:
Hmm, I pulled `test/utils.go` because I thought it had no used methods once I moved FindK8sClient to the test file and made it local. There was another seemingly unused, but now I'm realizing possibly externally used function there. Should I just put utils.go back as-was to alleviate the go-apidiff check fail, or...?

Makefile
```diff
@@ -178,7 +178,7 @@ test: manifests generate fmt lint test-unit test-e2e #HELP Run all tests.
 
 .PHONY: e2e
 e2e: #EXHELP Run the e2e tests.
-	go test -count=1 -v ./test/e2e/...
+	go test -count=1 -v -run "$(if $(TEST_FILTER),$(TEST_FILTER),.)" ./test/e2e/...
```
camilamacedo86, line 181: Why we need TEST_FILTER?

LalatenduMohanty: +1, I am also curious about why we need TEST_FILTER?

bentito: I thought about leaving it out, but it was very handy for repeatedly running only one of the e2e test while developing.

camilamacedo86: IHMO, that adds more complexities, could we remove? That is not clear when/how to use
If we want to run only 1 test, we can use go test manually and pass the test right?

test/e2e/metrics_test.go
```diff
@@ -89,134 +129,265 @@ func NewMetricsTestConfig(t *testing.T, client, selector, clusterRole, clusterBi
...
-	cmd.Stderr = &outAndErr
-	err := cmd.Run()
-	return outOnly.Bytes(), outAndErr.Bytes(), err
+// Helpers for pointers
+func boolPtr(b bool) *bool {
+	return &b
+}
+
+func int64Ptr(i int64) *int64 {
+	return &i
```
camilamacedo86, line 392: Could we remove it?
K8s provides the helper already.
The import is `"k8s.io/utils/ptr"`
And how to use it is:

```
RunAsNonRoot: ptr.To(true),

OR

RunAsUser:                ptr.To(int64(1001)),

```

test/e2e/metrics_test.go
```diff
@@ -89,134 +129,265 @@ func NewMetricsTestConfig(t *testing.T, client, selector, clusterRole, clusterBi
...
 }
 
-// cleanup removes the created resources. Uses a context with timeout to prevent hangs.
-func (c *MetricsTestConfig) cleanup() {
+// cleanup deletes the test resources
+func (c *MetricsTestConfig) cleanup(ctx context.Context) {
 	c.t.Log("Cleaning up resources")
-	_ = exec.Command(c.client, "delete", "clusterrolebinding", c.clusterBinding, "--ignore-not-found=true", "--force").Run()
-	_ = exec.Command(c.client, "delete", "pod", c.curlPodName, "-n", c.namespace, "--ignore-not-found=true", "--force").Run()
+	policy := metav1.DeletePropagationForeground
```
camilamacedo86, line 303: Why we are adding it?

test/e2e/metrics_test.go
```diff
@@ -89,134 +129,265 @@ func NewMetricsTestConfig(t *testing.T, client, selector, clusterRole, clusterBi
...
 
-	c.t.Log("Validating the metrics endpoint")
-	curlCmd := exec.Command(c.client, "exec", c.curlPodName, "-n", c.namespace, "--",
-		"curl", "-v", "-k", "-H", "Authorization: Bearer "+token, c.metricsURL)
-	output, err := curlCmd.CombinedOutput()
-	require.NoError(c.t, err, "Error calling metrics endpoint: %s", string(output))
-	require.Contains(c.t, string(output), "200 OK", "Metrics endpoint did not return 200 OK")
+	// Combine stdout + stderr
+	combined := stdout.String() + stderr.String()
+	require.Contains(c.t, combined, "200 OK", "Metrics endpoint did not return 200 OK")
```
camilamacedo86, line 297: Is it not much more complex than before?
I think before it was very clear what it does, we could easily copy and paste to validate the things.
But anyone, maybe it is a matter of preference.

test/e2e/metrics_test.go
```diff
@@ -41,14 +49,16 @@ func TestOperatorControllerMetricsExportedEndpoint(t *testing.T) {
...
-	config.run()
+	mtc.run()
 }
 
 // TestCatalogdMetricsExportedEndpoint verifies that the metrics endpoint for catalogd
 func TestCatalogdMetricsExportedEndpoint(t *testing.T) {
-	client := utils.FindK8sClient(t)
-	config := NewMetricsTestConfig(
-		t, client,
+	kubeClient, restConfig := findK8sClient(t)
```
azych, line 57: I was thinking that since metrics endpoint tests are part of the general e2e suite, we could also use the `client.Client` that we initialize as part of `TestMain` entrypoint:
```go
func TestMain(m *testing.M) {
	cfg = ctrl.GetConfigOrDie()

	var err error
	utilruntime.Must(apiextensionsv1.AddToScheme(scheme.Scheme))
	c, err = client.New(cfg, client.Options{Scheme: scheme.Scheme})
	utilruntime.Must(err)

	os.Exit(m.Run())
}
```

For that to work, we'd have to rename it, as it's currently available on a package-level as `c` and would be shadowed by `MetricsTestConfig` receiver name, which is also called `c` atm. We could also rename the receiver ofc ;)

WDYT?

test/e2e/metrics_test.go
```diff
@@ -57,13 +67,29 @@ func TestCatalogdMetricsExportedEndpoint(t *testing.T) {
 		"https://catalogd-service.NAMESPACE.svc.cluster.local:7443/metrics",
 	)
 
-	config.run()
+	mtc.run()
+}
+
+func findK8sClient(t *testing.T) (kubernetes.Interface, *rest.Config) {
+	cfg, err := config.GetConfig()
+	if err != nil {
```
azych, line 75: nit: maybe just `require.NoError(f)` instead of an `if`?

test/e2e/metrics_test.go
```diff
@@ -89,134 +129,265 @@ func NewMetricsTestConfig(t *testing.T, client, selector, clusterRole, clusterBi
...
-// createCurlMetricsPod creates the Pod with curl image to allow check if the metrics are working
-func (c *MetricsTestConfig) createCurlMetricsPod() {
+	ctx := context.Background()
+	c.createMetricsClusterRoleBinding(ctx)
+	token := c.getServiceAccountToken(ctx)
+	c.createCurlMetricsPod(ctx)
+	c.waitForPodReady(ctx)
+	// Exec `curl` in the Pod to validate the metrics
+	c.validateMetricsEndpoint(ctx, token)
+	defer c.cleanup(ctx)
```
azych, line 141: nit: I know that wasn't introduced by your PR, but `defer` looks redundant, as it's already being called as the last call

test/e2e/metrics_test.go
```diff
@@ -89,134 +129,265 @@ func NewMetricsTestConfig(t *testing.T, client, selector, clusterRole, clusterBi
...
 
-	// Wait for the Pod to be deleted.
-	if err := waitForDeletion(ctx, c.client, "pod", c.curlPodName, "-n", c.namespace); err != nil {
-		c.t.Logf("Error waiting for pod deletion: %v", err)
+// waitForClusterRoleBindingDeletion polls until the named ClusterRoleBinding no longer exists
+func waitForClusterRoleBindingDeletion(ctx context.Context, t *testing.T, kubeClient kubernetes.Interface, name string) {
+	err := wait.PollUntilContextTimeout(ctx, 2*time.Second, 60*time.Second, false, func(ctx context.Context) (bool, error) {
+		_, err := kubeClient.RbacV1().ClusterRoleBindings().Get(ctx, name, metav1.GetOptions{})
+		if err != nil {
+			if strings.Contains(err.Error(), "not found") {
```
azych, line 325: nit: you can use `k8s.io/apimachinery/pkg/api/errors:IsNotFound(err)`. Here and potentially other places

test/e2e/metrics_test.go
```diff
@@ -89,134 +129,265 @@ func NewMetricsTestConfig(t *testing.T, client, selector, clusterRole, clusterBi
...
+		}
+		return false, nil
+	})
+	if err != nil {
+		// If the context timed out, the test should fail with a more direct message
+		if errors.Is(err, context.DeadlineExceeded) {
+			c.t.Fatal("Timed out waiting for the curl pod to become Ready")
+		}
+		require.NoError(c.t, err, "Error waiting for curl pod to become Ready")
+	}
```
azych, line 250: super nit: you could get rid of one nesting level, by doing:
```go
if errors.Is(err, context.DeadlineExceeded) {
	c.t.Fatal("Timed out waiting for the curl pod to become Ready")
}
if err != nil {
    ....
}
```

as errors.Is() won't do anything if the error is nil

azych (approved):
LGTM, just added a few nits

Makefile
perdasilva: Personally, I'm a fan of it. What's the worry in terms of complexity? It stays mostly hidden anyway...?
If anything I'd update the help to call it out XDD but I'm easy either way.

test/e2e/metrics_test.go
bentito: yeah Idk, it was certainly harder to implement, `kubectl exec` was doing several things for us. I think this PR mostly just answers @joelanford concern that that help from `kubectl` also opened more avenues for things to go subtly wrong. And @azych desire that this test should use `go-client` like the other e2e.

Makefile
bentito: seem okay to keep it? I've been using it like:

```bash
make test-e2e TEST_FILTER="Test.*MetricsExportedEndpoint"
```

test/e2e/metrics_test.go
bentito: yeah, makes sense to me, my solution is to rename the main test one: `globalClient`

test/e2e/metrics_test.go
bentito: Yeah I think it inadvertently got moved down. Moving back to the top of the function since point was to cleanup no matter what (pretty sure)

test/e2e/metrics_test.go
bentito: done. and the one below too.

test/e2e/metrics_test.go
bentito: done in two places in next commit

Makefile
joelanford: I like it too! Could we make sure that the unit test recipe uses this same variable?

test/e2e/metrics_test.go
```diff
@@ -57,13 +69,25 @@ func TestCatalogdMetricsExportedEndpoint(t *testing.T) {
 		"https://catalogd-service.NAMESPACE.svc.cluster.local:7443/metrics",
 	)
 
-	config.run()
+	mtc.run()
+}
+
+func findK8sClient(t *testing.T) (kubernetes.Interface, *rest.Config) {
+	cfg, err := config.GetConfig()
+	require.NoError(t, err, "Failed to get Kubernetes config")
```
joelanford, line 77: Do we already have `cfg` from `TestMain` as well?

test/e2e/metrics_test.go
joelanford: I could see wrapping this in a more general purpose `execInPod` kind of utility function that could be re-used. But mostly a nit since there's not another place that we need this at the moment.

test/e2e/metrics_test.go
bentito: yah, same, fixed in recent commit

Makefile
bentito: @perdasilva having trouble seeing where to add for docs, do we mention make targets in help now?

bentito: @joelanford to be equivalent for test unit we need to change like:
```bash
make test-unit TEST_PKG=./internal/operator-controller/applier TEST_FILTER=TestApply_Installation
```

I am committing this change now.

Makefile
```diff
@@ -211,7 +211,8 @@ test-unit: $(SETUP_ENVTEST) envtest-k8s-bins #HELP Run the unit tests
                 -tags '$(GO_BUILD_TAGS)' \
                 -cover -coverprofile ${ROOT_DIR}/coverage/unit.out \
                 -count=1 -race -short \
-                $(UNIT_TEST_DIRS) \
+                -run "$(if $(TEST_FILTER),$(TEST_FILTER),.)" \
+                $(if $(TEST_PKG),$(TEST_PKG),$(UNIT_TEST_DIRS)) \
```
joelanford, line 215: Rather than both `UNIT_TEST_DIRS` and `TEST_PKG`, I think we should consolidate to avoid confusion about what to set. How about this?
```make
ifeq ($(origin TEST_PKGS), undefined)
TEST_PKGS := $(shell go list ./... | grep -v /test/ | grep -v /catalogd/test/)
endif
```

Then, this line can stay simple:
```suggestion
                $(TEST_PKGS) \
```

bentito: yeah that makes sense. done.

test/e2e/metrics_test.go
azych: I agree, good point

Makefile
```diff
@@ -208,7 +208,11 @@ test-ext-dev-e2e: $(OPERATOR_SDK) $(KUSTOMIZE) $(KIND) #HELP Run extension creat
 	test/extension-developer-e2e/setup.sh $(OPERATOR_SDK) $(CONTAINER_RUNTIME) $(KUSTOMIZE) $(KIND) $(KIND_CLUSTER_NAME) $(E2E_REGISTRY_NAMESPACE)
 	go test -count=1 -v ./test/extension-developer-e2e/...
 
-UNIT_TEST_DIRS := $(shell go list ./... | grep -v /test/)
+ENVTEST_VERSION := $(shell go list -m k8s.io/client-go | cut -d" " -f2 | sed 's/^v0\.\([[:digit:]]\{1,\}\)\.[[:digit:]]\{1,\}$$/1.\1.x/')
```
camilamacedo86, line 211: We have this value already.
See; https://github.com/operator-framework/operator-controller/blob/7fc18c64660c97c70e4c6704147a746f657543f4/Makefile#L35-L48

Could you please rebase with main and ensure use that instead

bentito: ah yep, nothing to rebase as it was caused by my error rebase merging. I just removed the dupe in new commit. TY!

joelanford (approved):

PR #1823, opened 2025-02-28
🐛 util/image: unset PAXRecords and Xattrs when applying files

joelanford:
# Description

If a catalog or bundle image contains files with xattrs, that can cause the controller to fail to unpack the image. This PR unsets those attributes to avoid permission errors when writing files to disk.

## Reviewer Checklist

- [ ] API Go Documentation


camilamacedo86 (approved):
I think the proposed solution is a good approach.
We address the problem when unpacking the image, and from that point on, it's up to the solution itself to determine whether it has the necessary capabilities to run whatever it needs to. We are not modifying it at all.

So, if my understanding is correct:

/lgtm
/approve

PR #1827, opened 2025-03-03
:seedling: Update containers/image to v5.34.1

tmshort:
However, due to some incompatibility between controller-gen and
github.com/proglottis/gpgme@v0.1.4, we need to move the kubebuilder
comments so controller-gen doesn't try to access gpgme.

This meant moving the kubebuilder:rbac comments into their own
package, and focusing the webhook generation the just one directory.

# Description

## Reviewer Checklist

- [ ] API Go Documentation


Makefile
```diff
@@ -133,14 +133,15 @@ manifests: $(CONTROLLER_GEN) #EXHELP Generate WebhookConfiguration, ClusterRole,
 	mv $(CRD_WORKING_DIR)/olm.operatorframework.io_clustercatalogs.yaml $(KUSTOMIZE_CATD_CRDS_DIR)
 	rmdir $(CRD_WORKING_DIR)
 	# Generate the remaining operator-controller manifests
-	$(CONTROLLER_GEN) rbac:roleName=manager-role paths="./internal/operator-controller/..." output:rbac:artifacts:config=$(KUSTOMIZE_OPCON_RBAC_DIR)
+	$(CONTROLLER_GEN) rbac:roleName=manager-role paths="./internal/operator-controller/controllers/..." output:rbac:artifacts:config=$(KUSTOMIZE_OPCON_RBAC_DIR)
```
camilamacedo86, line 136: The problematic dep is called in the controller :-( 
I thought in extra in a utils and then do something like that
it might works

tmshort: That might just be my crazy idea I mentioned earlier

tmshort:
Replacement for #1818 and #1692

joelanford:
I think we need to be going to the controller-tools repo to get this fixed there.

tmshort:
> I think we need to be going to the controller-tools repo to get this fixed there.

Problem is that it doesn't seem to be controller-tools, _specifically_. It works fine locally, this seems to be a problem with gpgme _somehow_.

perdasilva:
> > I think we need to be going to the controller-tools repo to get this fixed there.
> 
> Problem is that it doesn't seem to be controller-tools, _specifically_. It works fine locally, this seems to be a problem with gpgme _somehow_.

Could it be working locally because CGO_ENABLED=1? 

The issue is the build tags don't get passed down to controller-gen when its loading the packages (somewhere around [here](https://github.com/kubernetes-sigs/controller-tools/blob/main/pkg/genall/genall.go#L221)). I've done local test hard-coding the `containers_image_openpgp` [here](https://github.com/kubernetes-sigs/controller-tools/blob/main/pkg/loader/loader.go#L378) and it seemed to work. 

We need a way to configure controller-gen to pass build tags to the loader =S

tmshort:
@perdasilva 
At first it was failing for me. After I did a `go clean -modcache`, it started working. Doing that in CI doesn't seem to work, and it also seems fragile.

To be blunt, this PR _does_ work, because it avoids pulling in gpgme code by isolating the RBAC directives. Admittedly, fixing controler-gen would be great, but how long would that take and when would that happen?

internal/catalogd/webhook/cluster_catalog_webhook.go
```diff
@@ -13,8 +13,6 @@ import (
 
 // +kubebuilder:webhook:admissionReviewVersions={v1},failurePolicy=Fail,groups=olm.operatorframework.io,mutating=true,name=inject-metadata-name.olm.operatorframework.io,path=/mutate-olm-operatorframework-io-v1-clustercatalog,resources=clustercatalogs,verbs=create;update,versions=v1,sideEffects=None,timeoutSeconds=10
 
-// +kubebuilder:rbac:groups=olm.operatorframework.io,resources=clustercatalogs,verbs=get;list;watch;patch;update
```
camilamacedo86, line 16: That is interesting, we fail to get the RBAC but not to generate the webhook config based on the markers?

tmshort: `controller-gen` also has a limitation that it cannot use multiple directories (as discovered in my earlier merging of the API directories).

So, I pulled all the RBAC into one place.
The webhook generation still looks at this one directory.

joelanford:
Done in #1892

PR #1828, opened 2025-03-03
:bug: (fix): update PSA versions to match Kubernetes API version

camilamacedo86:
Derive Kubernetes minor version from client-go for PSA label and combine the calls required to get the k8s version from go.mod

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,6 +32,20 @@ export IMAGE_TAG
 OPCON_IMG := $(OPCON_IMAGE_REPO):$(IMAGE_TAG)
 CATD_IMG := $(CATD_IMAGE_REPO):$(IMAGE_TAG)
 
+# Extract Kubernetes client-go version used to set the version to the PSA labels, for ENVTEST and KIND
+ifeq ($(origin K8S_VERSION), undefined)
+K8S_VERSION := $(shell go list -m k8s.io/client-go | cut -d" " -f2 | sed -E 's/^v0\.([0-9]+)\.[0-9]+$$/1.\1/')
+endif
+
+# Ensure ENVTEST_VERSION follows correct "X.Y.x" format
+ENVTEST_VERSION := $(K8S_VERSION).x
+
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -143,9 +143,18 @@ generate: $(CONTROLLER_GEN) #EXHELP Generate code containing DeepCopy, DeepCopyI
 	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
 
 .PHONY: verify
-verify: tidy fmt generate manifests crd-ref-docs #HELP Verify all generated code is up-to-date.
+verify: tidy fmt generate manifests crd-ref-docs update-k8s-values #HELP Verify all generated code is up-to-date.
 	git diff --exit-code
 
+# Minor Kubernetes version to build against derived from the client-go dependency version
+KUBE_MINOR ?= $(shell go list -m k8s.io/client-go | cut -d" " -f2 | sed 's/^v0\.\([[:digit:]]\{1,\}\)\.[[:digit:]]\{1,\}$$/1.\1/')
```
joelanford, line 150: We already have `KIND_NODE_VERSION` that is nearly the same. Maybe we should consolidate so that we don't have to shell out to `go list` multiple times?

Also: don't use `?=` here. Use `:=`. This value should not be overridable.

Also, also for future reference: `?=` and `=` causes make to re-evaluate the variable every time it is referenced. We should not use these at all for variables derived from shell commands, because it can make `make` take _much_ longer unnecessary. If we truly need overridable variables which defaults from a shell command: use this syntax:

```make
ifeq ($(origin VAR_NAME), undefined)
VAR_NAME := $(shell do-a-thing)
endif

camilamacedo86: I thought on that as well, but I imagine a little out of scope for the other side.
So, I was planning on doing something like in a follow-up anyway, it is done now.
Thank you a lot for your time and review !!!!

Let me know wdyt..

perdasilva (approved):

PR #1829, opened 2025-03-03
🐛 (fix) PSA enforcement: Move from baseline to restricted

camilamacedo86:
Namespaces should be restricted by default instead of have granted additional not required permissions

# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/config/base/common/namespace.yaml
+++ b/config/base/common/namespace.yaml
@@ -3,6 +3,6 @@ kind: Namespace
 metadata:
   labels:
     app.kubernetes.io/part-of: olm
-    pod-security.kubernetes.io/enforce: baseline
+    pod-security.kubernetes.io/enforce: restricted
     pod-security.kubernetes.io/enforce-version: "v1.32"
   name: system


config/base/common/namespace.yaml
```diff
@@ -3,6 +3,6 @@ kind: Namespace
 metadata:
   labels:
     app.kubernetes.io/part-of: olm
-    pod-security.kubernetes.io/enforce: baseline
+    pod-security.kubernetes.io/enforce: restricted
```
camilamacedo86, line 6: @joelanford ^ JFY

bentito: @camilamacedo86 can you identify which, if any, current e2e test might be able to verify this doesn't cause a problem? Maybe just?:

in `./kind-config.yaml` Should we turn on PodSecurity?
```diff
    kubeadmConfigPatches:
      - |
        kind: ClusterConfiguration
        apiServer:
            extraArgs:
-              enable-admission-plugins: OwnerReferencesPermissionEnforcement
+              enable-admission-plugins: OwnerReferencesPermissionEnforcement,PodSecurity
```

bentito: e2e seems to pass for me with this set, so maybe it's all good?

bentito: just trying to verify it's really used if I set it there

camilamacedo86: When we install OLMv1 the ns will be set within right
So, the tests running on that. But do you still want any extra step? if so, can you please clarify? Sorry, I could not follow what are you asking for I do in this case.

bentito: nvm, seems like at our current version enforced for Kind, we've got PodSecurity by default. Verified it bounced a non-compliant pod trying to be created in an NS set restricted. Since our current e2e do deployment, they must be okay.

perdasilva (dismissed):

camilamacedo86:
I need to rebase this one,
Could we get the approval again :-) ?
@perdasilva

PR #1840, opened 2025-03-04
🌱 Add catalogd metas service demo

grokspawn:
# Description
- adds a demo for the metas catalog service endpoint (but does not make this part of regular CI, since it is currently dependent on deploying from local artifacts
- Fixes a bug in the Makefile where the VERSION attribute would be empty instead of undefined, and succeeds now only with valid (non-empty) input

# Demo
[asciicast](https://asciinema.org/a/706817)

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -302,7 +302,7 @@ kind-clean: $(KIND) #EXHELP Delete the kind cluster.
 #SECTION Build
 
 # attempt to generate the VERSION attribute for certificates
-# fail if it is unset afterwards, since the sideeffects are indirect
+# fail if it is unset afterwards, since the side effects are indirect
 ifeq ($(strip $(VERSION)),)
 VERSION := $(shell git describe --tags --always --dirty)
 endif
--- a/Makefile
+++ b/Makefile
@@ -301,10 +301,15 @@ kind-clean: $(KIND) #EXHELP Delete the kind cluster.
 
 #SECTION Build
 
-ifeq ($(origin VERSION), undefined)
+# attempt to generate the VERSION attribute for certificates
+# fail if it is unset afterwards, since the sideeffects are indirect
+ifeq ($(strip $(VERSION)),)
 VERSION := $(shell git describe --tags --always --dirty)
 endif
 export VERSION
+ifeq ($(strip $(VERSION)),)
+	$(error undefined VERSION; resulting certs will be invalid)
+endif
 
 ifeq ($(origin CGO_ENABLED), undefined)
 CGO_ENABLED := 0
--- a/hack/demo/catalogd-metas-demo-script.sh
+++ b/hack/demo/catalogd-metas-demo-script.sh
@@ -1,0 +1,44 @@
+#!/usr/bin/env bash
+#
+# Welcome to the catalogd metas API endpoint demo
+#
+trap 'trap - SIGTERM && kill -- -"$$"' SIGINT SIGTERM EXIT
+
+kind delete cluster
+kind create cluster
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -275,10 +275,15 @@ kind-clean: $(KIND) #EXHELP Delete the kind cluster.
 
 #SECTION Build
 
-ifeq ($(origin VERSION), undefined)
+# attempt to generate the VERSION attribute for certificates
+# fail if it is unset afterwards, since the sideeffects are indirect
```
tmshort, line 279:
```suggestion
# fail if it is unset afterwards, since the side effects are indirect
```

hack/demo/catalogd-metas-demo-script.sh
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env bash
+#
+# Welcome to the catalogd metas API endpoint demo
+#
+trap 'trap - SIGTERM && kill -- -"$$"' SIGINT SIGTERM EXIT
+
+## until we cut a release containing the new API, we need to deploy local artifacts
```
tmshort, line 7: Why are some of these "double" commented?

grokspawn:
[This](https://github.com/operator-framework/operator-controller/pull/1840/files#diff-a8f769dc3839ebffbf5dcfd667b49eb1509d50f31df746dd7d04cb90b2a85109R32) is not OK.  The problem here appears to be that catalogd's deployment conditions are serving := True with progressing := True... but the api is not actually serving.  

I need to diagnose this more, but it looks like an architectural communications gap rather than a bug.

hack/demo/catalogd-metas-demo-script.sh
azych, line 7: > until we cut a release containing the new API

Looks like 1.2.0 already has it

grokspawn: The linter said "use single quotes so they are not expanded now, but are expanded later", and the general syntax is "quote to capture strings with spaces / special characters".

perdasilva:
> [This](https://github.com/operator-framework/operator-controller/pull/1840/files#diff-a8f769dc3839ebffbf5dcfd667b49eb1509d50f31df746dd7d04cb90b2a85109R32) is not OK. The problem here appears to be that catalogd's deployment conditions are serving := True with progressing := True... but the api is not actually serving.
> 
> I need to diagnose this more, but it looks like an architectural communications gap rather than a bug.

should we hold until this is resolved? Or can we still merge and resolve out of band?

azych (approved):

grokspawn:
> > [This](https://github.com/operator-framework/operator-controller/pull/1840/files#diff-a8f769dc3839ebffbf5dcfd667b49eb1509d50f31df746dd7d04cb90b2a85109R32) is not OK. The problem here appears to be that catalogd's deployment conditions are serving := True with progressing := True... but the api is not actually serving.
> > I need to diagnose this more, but it looks like an architectural communications gap rather than a bug.
> 
> should we hold until this is resolved? Or can we still merge and resolve out of band?

IMO we can merge this and then look into what's going on with catd under the hood.  It doesn't look like a demo bug, but the demo could be useful in diagnosing what's going on in catd.

perdasilva (approved):
lgtm ^^ thanks for adding the curl retries =D

perdasilva:
we should probably standardize the w:h for the demo window when generating it, if that's possible. Nothing that needs doing in the PR tho...

PR #1855, opened 2025-03-07
✨ Enable pprof to allow memory profiling

anik120:
# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/config/base/catalogd/manager/manager.yaml
+++ b/config/base/catalogd/manager/manager.yaml
@@ -46,6 +46,7 @@ spec:
         args:
         - --leader-elect
         - --metrics-bind-address=:7443
+        - --pprof-bind-address=:6060
         - --external-address=catalogd-service.olmv1-system.svc
         image: controller:latest
         name: manager
--- a/config/base/operator-controller/manager/manager.yaml
+++ b/config/base/operator-controller/manager/manager.yaml
@@ -45,6 +45,7 @@ spec:
         args:
         - "--health-probe-bind-address=:8081"
         - "--metrics-bind-address=:8443"
+        - --pprof-bind-address=:6060
         - "--leader-elect"
         image: controller:latest
         imagePullPolicy: IfNotPresent


config/base/catalogd/manager/manager.yaml
```diff
@@ -46,6 +46,7 @@ spec:
         args:
         - --leader-elect
         - --metrics-bind-address=:7443
+        - --pprof-bind-address=:6060
```
camilamacedo86, line 49: This feature is handy for testing and debugging, **but I’m wondering if it should be enabled by default**. I have concerns about **performance impact** and **security risks**. That’s why we added a warning in Kubebuilder’s [docs](https://book.kubebuilder.io/reference/pprof-tutorial#not-recommended-for-production).  

**I think it's fine to enable `pprof` in production when needed, do the profiling, and then disable it. My concern is leaving it enabled indefinitely.** That seems to be the general approach others take as well (enable/disable), see: [Google Groups discussion](https://groups.google.com/g/golang-nuts/c/e6lB8ENbIw8/m/azeTCGj7AgAJ).  

⚠️ Some references highlight security risks:  
- **Amazon CodeGuru** warns about exposed `pprof` endpoints: https://docs.aws.amazon.com/codeguru/detector-library/go/pprof-endpoint/  
- **Potential data leaks**: https://beaglesecurity.com/blog/vulnerability/jira-exposed-pprof.html and  https://www.aquasec.com/blog/300000-prometheus-servers-and-exporters-exposed-to-dos-attacks/

Also, I found this issue: https://github.com/golang/go/issues/65208

camilamacedo86: But let's also see what the other thinks @tmshort @joelanford, what are your thoughts? 
Are you OK with?

anik120: That's super helpful information @camilamacedo86. We'd already discussed this on slack and I decided to close the PR, but this detailed information is really helpful, so thank you!

PR #1859, opened 2025-03-11
🐛(fix) Remove "Serving" condition type from ConditionSets

anik120:
conditionset.ConditionTypes was being used by the CluterExtension controller to `ensureAllConditionsWithReason` whenever a particular reason needed to be set for all of the ClusterExention's conditions. However, the `ConditionTypes` included a Condition Type `Serving`, which is not a ClusterExtension Condition(it is a ClusterCatalog condition).

This is causing the `Serving` condition to show up when a resolution fails, which is incorrect to begin with, and is then never cleared when the resolution succeeds at a later stage.

```
try to upgrade ClusterExtension to a non-exist version
$ kubectl patch ClusterExtension extension-77972 -p '{"spec":{"source":{"catalog":{"version":"0.2.0"}}}}' --type=merge
clusterextension.olm.operatorframework.io/extension-77972 patched

- lastTransitionTime: "2025-03-04T07:16:27Z"
    message: 'error upgrading from currently installed version "0.1.0": no bundles
      found for package "nginx77972" matching version "0.2.0"'
    observedGeneration: 2
    reason: Retrying
    status: "True"
    type: Progressing
  - lastTransitionTime: "2025-03-04T07:16:35Z"
    message: 'error upgrading from currently installed version "0.1.0": no bundles
      found for package "nginx77972" matching version "0.2.0"'
    observedGeneration: 2
    reason: Failed
    status: "False"
    type: Serving
  install:
    bundle:
      name: nginx77972.v0.1.0
      version: 0.1.0
```

This PR removes the `Serving` condition type from `conditonSets.ConditionType`

# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/api/v1/clustercatalog_types.go
+++ b/api/v1/clustercatalog_types.go
@@ -34,6 +34,14 @@ const (
 
 	AvailabilityModeAvailable   AvailabilityMode = "Available"
 	AvailabilityModeUnavailable AvailabilityMode = "Unavailable"
+
+	// Condition types
+	TypeServing = "Serving"
+
+	// Serving Reasons
+	ReasonAvailable                = "Available"
+	ReasonUnavailable              = "Unavailable"
+	ReasonUserSpecifiedUnavailable = "UserSpecifiedUnavailable"
 )
 
 //+kubebuilder:object:root=true
--- a/api/v1/clusterextension_types_test.go
+++ b/api/v1/clusterextension_types_test.go
@@ -5,7 +5,6 @@ import (
 	"go/ast"
 	"go/parser"
 	"go/token"
-	"io/fs"
 	"strconv"
 	"strings"
 	"testing"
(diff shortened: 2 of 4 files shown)


api/v1/clustercatalog_types.go
```diff
@@ -34,6 +34,14 @@ const (
 
 	AvailabilityModeAvailable   AvailabilityMode = "Available"
 	AvailabilityModeUnavailable AvailabilityMode = "Unavailable"
+
+	// Serving condition types
```
azych, line 38: nit: ~Serving~ Condition types?

azych (approved):
/lgtm

api/v1/clustercatalog_types.go
anik120: Done 👍🏽 
I pushed a change to address Joe's feedback too can I get a re-lgtm please 🙏🏽

api/v1/clusterextension_types_test.go
```diff
@@ -61,7 +61,7 @@ func parseConstants(prefix string) ([]string, error) {
 	// ParseDir returns a map of package name to package ASTs. An AST is a representation of the source code
 	// that can be traversed to extract information. The map is keyed by the package name.
 	pkgs, err := parser.ParseDir(fset, ".", func(info fs.FileInfo) bool {
-		return !strings.HasSuffix(info.Name(), "_test.go")
+		return !strings.HasSuffix(info.Name(), "_test.go") && !(info.Name() == "clustercatalog_types.go")
```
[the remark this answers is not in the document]
anik120: That's a good point. I actually changed the function to 
```
// parseConstants parses the values of the top-level constants that start with the given prefix,
// in the files clusterextension_types.go and common_types.go.
func parseConstants(prefix string) ([]string, error)
```

so that this function never sees any movement if new types are added. It also helped to get rid of the 3 nest for-loops and replace it with a nicer looking 2 nested one. 

All this made possible by the replacement of `parseDir` with `parseFile`

joelanford:
Other then the nit on the comment, changes look good to me!

joelanford (approved):

Issue #1607, opened 2025-01-13
catalogd web api benchmarking


grokspawn:
1. Establish a common access pattern for assessing performance of the catalogd query web API for e.g. 
   - initial catalog data unpack (index preparation)
   - catalog data update & re-unpack
   - query[-no-parameters-], 
   - query&name=$somePackage&schema=olm.package,
   - query&package=$somePackage  
3. Collect pprof profiles for CPU and memory when evaluating the pattern
4. Find the most resource intensive code paths. Provide documentation and recommendations related to making improvements in those areas.
5. Coordinate with OLM maintainers to make improvements in areas deemed to provide the most significant performance and scale gain.
6. Implement automated performance and scale regression tests in the existing upstream CI test suite.

grokspawn:
We want to include this scope in #920.  I'd initially argued that we would want to scope 920 over existing APIs and performance evaluations and use this for just the query API, but in discussion now this really ought to be considered part of the scope associated with 920.  I'll adjust the body there to include it.

Issue #1040, opened 2024-07-12
Add tests to generate pprof files


JudeNiroshan:
There are no automated tests around profiling that can run in the CI pipeline. New tests should cover general workflows and generate the pprof files. The primary focus areas are:

- CPU usage
- Memory usage

Relates to https://github.com/operator-framework/operator-controller/issues/920

camilamacedo86:
We have a WIP Brief to start checking the test performance. 
I am not sure if we will use pprof, @dtfranz just for let you know

camilamacedo86:
Could we not close this one as duplicated of: https://github.com/operator-framework/operator-controller/issues/1607?

PR #1876, opened 2025-03-20
📖 (doc): Add under the docs/draft how to profilling with pprof

camilamacedo86:
This one as just a doc to show we can use pprof with the components
The usage of pprof can help us solve bugs and analyse issues
It is blocked by #1875 because we need to add a flag.

It is related to (Motivated by):

- https://github.com/operator-framework/operator-controller/issues/1607
- https://github.com/operator-framework/operator-controller/issues/1040
- https://github.com/operator-framework/operator-controller/issues/920

### DIFF
--- a/docs/draft/howto/profiling_with_pprof.md
+++ b/docs/draft/howto/profiling_with_pprof.md
@@ -1,0 +1,297 @@
+# Profiling with Pprof 
+
+!!! warning
+Pprof bind port flag are available as an alpha release and are subject to change in future versions.
+
+[Pprof][pprof] is a useful tool for analyzing memory and CPU usage profiles. However, it is not
+recommended to enable it by default in production environments. While it is great for troubleshooting,
+keeping it enabled can introduce performance concerns and potential information leaks.
+
+Both components allow you to enable pprof by specifying the port it should bind to using the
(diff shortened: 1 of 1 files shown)


anik120 (changes requested):
I don't think this is needed in the user docs. 

The `docs` folder hosts the OLMv1 user docs, and this is a dev doc.

camilamacedo86:
Hi @anik120 

Thank you for your input and help. As far I understand our documentation as a whole is valid for both end users and maintainers.

IHMO: I believe it is important to have documentation that explains how to enable and use it with the components. This will be helpful for end users when they encounter issues and need gather further details, as well as for us (maintainers) when we are looking to debug an issue

camilamacedo86:
/hold cancel

We can move forward now

grokspawn (approved):

anik120 (dismissed):
Yea we probably need to have a conversation about where we want dev docs to be. Usually it's not in the same place because that causes confusion. We don't want our non-technical users to be flooded with dev docs. 

Eg: k8s has user docs [here](https://github.com/kubernetes/kubernetes/tree/master#to-start-using-k8s) , and separate dev docs [here](https://github.com/kubernetes/kubernetes/tree/master#to-start-developing-k8s). 

I've never run into any dev docs while using kubernetes.io.

PR #1880, opened 2025-03-27
✨ contentmanager: use namespace-scoped informers

joelanford:
# Description

This PR changes operator-controller's contentmanager, which is used to establish watches for objects it is lifecycling, so that it uses namespace-scoped informers for namespace-scoped watches. Prior to this PR, contentmanager is using a cluster-scoped watch for all resources, regardless of their scoping.

This is beneficial because it means that the installer user can change its current cluster-wide list/watch permissions into namespace-scoped permissions for each resource in their extension's manifest.

It should be noted that the content manager manages a separate set of informers for each cluster extension due to the use of the cluster extension's service account. So switching from cluster-scoped informers to namespace-scoped informers, in most cases, will not increase the total number of informers that operator-controller has running at a given time. The only scenario where the number of informers increases is if the manifest for a particular cluster extension contains namespace-scoped resources in multiple namespaces.

On main, the number of informers for a cluster extension is: `<number of unique resource types>`
After this PR merges, the number of informers for a cluster extension is: `<number of unique resource type/namespace pairs>`.

In almost all cases, these numbers will be the same because extensions almost always install cluster-scoped objects and namespace-scoped objects in a single namespace.

The only case that this _would_ be an issue is if we supported MultiNamespace install mode because that would cause roles and role bindings to be included in the manifest for all of the watched namespaces. We do no support that install mode, and we do not plan to support that install mode, essentially for this exact reason.

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/operator-controller/contentmanager/cache/cache.go
+++ b/internal/operator-controller/contentmanager/cache/cache.go
@@ -10,6 +10,8 @@ import (
 	"sync"
 	"time"
 
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -41,27 +43,29 @@ type CloserSyncingSource interface {
 }
 
 type sourcerer interface {
-	// Source returns a CloserSyncingSource for the provided
-	// GroupVersionKind. If the CloserSyncingSource encounters an
+	// Source returns a CloserSyncingSource for the provided namespace
+	// and GroupVersionKind. If the CloserSyncingSource encounters an
 	// error after having initially synced, it should requeue the
 	// provided client.Object and call the provided callback function
-	Source(schema.GroupVersionKind, client.Object, func(context.Context)) (CloserSyncingSource, error)
+	Source(string, schema.GroupVersionKind, client.Object, func(context.Context)) (CloserSyncingSource, error)
 }
 
 type cache struct {
-	sources     map[schema.GroupVersionKind]CloserSyncingSource
+	sources     map[sourceKey]CloserSyncingSource
 	sourcerer   sourcerer
 	owner       client.Object
 	syncTimeout time.Duration
 	mu          sync.Mutex
+	restMapper  meta.RESTMapper
 }
 
-func NewCache(sourcerer sourcerer, owner client.Object, syncTimeout time.Duration) Cache {
+func NewCache(sourcerer sourcerer, owner client.Object, syncTimeout time.Duration, rm meta.RESTMapper) Cache {
 	return &cache{
-		sources:     make(map[schema.GroupVersionKind]CloserSyncingSource),
+		sources:     make(map[sourceKey]CloserSyncingSource),
 		sourcerer:   sourcerer,
 		owner:       owner,
 		syncTimeout: syncTimeout,
+		restMapper:  rm,
 	}
 }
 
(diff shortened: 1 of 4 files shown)


internal/operator-controller/contentmanager/cache/cache.go
```diff
@@ -207,23 +212,23 @@ func (c *cache) removeStaleSources(gvks sets.Set[schema.GroupVersionKind]) error
...
-	delete(c.sources, gvk)
+	delete(c.sources, srcKey)
 	return nil
 }
 
-func (c *cache) getCacheGVKs() sets.Set[schema.GroupVersionKind] {
-	cacheGvks := sets.New[schema.GroupVersionKind]()
-	for gvk := range c.sources {
-		cacheGvks.Insert(gvk)
+func (c *cache) getCacheGVKs() sets.Set[sourceKey] {
```
everettraven, line 226: Because this is no longer returning a set of the GVKs, rename this to something like:
```suggestion
func (c *cache) getCacheSourceKeys() sets.Set[sourceKey] {
```
?

internal/operator-controller/contentmanager/cache/cache.go
```diff
@@ -233,9 +238,10 @@ func (c *cache) getCacheGVKs() sets.Set[schema.GroupVersionKind] {
 //
 // An empty Group is assumed to be the "core" Kubernetes
 // API group.
-func gvksForObjects(objs ...client.Object) (sets.Set[schema.GroupVersionKind], error) {
-	gvkSet := sets.New[schema.GroupVersionKind]()
+func sourceKeysForObjects(objs ...client.Object) (sets.Set[sourceKey], error) {
+	sourceKeys := sets.New[sourceKey]()
 	for _, obj := range objs {
+		ns := obj.GetNamespace()
```
everettraven, line 244: I would recommend gating this in a check for the scope of the resource (namespace vs cluster). Don't set a namespace for cluster scoped resource informers.

For example, if I were to provide you a `ClusterRole` defined like:
```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: role-grantor
  namespace: default
rules:
- apiGroups: ["rbac.authorization.k8s.io"]
  resources: ["rolebindings"]
  verbs: ["create"]
- apiGroups: ["rbac.authorization.k8s.io"]
  resources: ["clusterroles"]
  verbs: ["bind"]
  # omit resourceNames to allow binding any ClusterRole
  resourceNames: ["admin","edit","view"]
```

The informer you create here would not be valid for monitoring clusterroles as they are cluster scoped resources. Even though specifying a namespace here does nothing, it is accepted by the Kube apiserver.

If you get these objects from parsing manifests that may result in you having a namespace set for a cluster scoped resource I would perform this check. If you are only ever going to get these objects from the cluster, it _might_ be OK to just statically set this, but it doesn't hurt to have the extra layer of protection anyways.

internal/operator-controller/contentmanager/cache/cache_test.go
```diff
@@ -69,11 +70,11 @@ func TestCacheWatch(t *testing.T) {
...
 
 	pod := &corev1.Pod{}
-	podGvk := corev1.SchemeGroupVersion.WithKind("Pod")
-	pod.SetGroupVersionKind(podGvk)
+	pod.SetGroupVersionKind(corev1.SchemeGroupVersion.WithKind("Pod"))
+	pod.SetNamespace(rand.String(8))
 
 	require.NoError(t, c.Watch(context.Background(), &mockWatcher{}, pod))
-	require.Contains(t, c.(*cache).sources, podGvk, "sources", c.(*cache).sources)
+	require.Contains(t, c.(*cache).sources, sourceKey{pod.Namespace, pod.GroupVersionKind()}, "sources", c.(*cache).sources)
```
everettraven, line 77: Probably would require a bit of test refactoring (or a new test function), but I would recommend adding a test case to cover a cluster scoped resource being provided with a namespace set to ensure you are appropriately creating a cluster scoped source instead of a namespace scoped one.

internal/operator-controller/contentmanager/cache/cache.go
joelanford: Done!

internal/operator-controller/contentmanager/cache/cache.go
joelanford: Good catch! This input _could_ be coming directly from user-provided manifests, so best to sanitize it. For now, I coded it to silently ignore the namespace setting in the manifest for cluster-scoped objects.

It is maybe debatable if we should consider that an error. I guess since the API server allows it, we should too?

internal/operator-controller/contentmanager/cache/cache_test.go
joelanford: Added a new test!

everettraven (dismissed):
/lgtm

joelanford:
Closed. We can revive if/when we need it.

PR #1893, opened 2025-04-02
[WIP] :sparkles: add render package

perdasilva:
# Description

## Reviewer Checklist

- [ ] API Go Documentation


internal/operator-controller/rukpak/convert/render/generate_test.go
```diff
@@ -0,0 +1,430 @@
...
+                        Labels: map[string]string{
+                            "bar": "foo",
+                        },
+                    },
+                    Spec: appsv1.DeploymentSpec{
+                        RevisionHistoryLimit: ptr.To(int32(1)),
+                        Template: corev1.PodTemplateSpec{
+                            ObjectMeta: metav1.ObjectMeta{
+                                Annotations: map[string]string{
+                                    "csv":                  "annotation",
```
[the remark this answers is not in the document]
perdasilva: I had the same questions in my mind. This is probably a game of telephone, but I just wanted to make sure we had what we currently have in the convert function. I don't know the motivation behind this =S

internal/operator-controller/rukpak/convert/render/generate.go
```diff
@@ -0,0 +1,191 @@
...
+		objs = append(objs, obj)
+	}
+	return objs, nil
+}
+
+func saNameOrDefault(saName string) string {
+	if saName == "" {
+		return "default"
+	}
+	return saName
```
joelanford, line 190: Nit: New-ish stdlib function that can simplify this.
```suggestion
	return cmp.Or(saName, "default")
```

internal/operator-controller/rukpak/convert/generate.go
```diff
@@ -0,0 +1,203 @@
...
+	if rv1 == nil {
+		return nil, fmt.Errorf("bundle cannot be nil")
+	}
+	objs := make([]client.Object, 0, len(rv1.CSV.Spec.InstallStrategy.StrategySpec.DeploymentSpecs))
+	for _, depSpec := range rv1.CSV.Spec.InstallStrategy.StrategySpec.DeploymentSpecs {
+		annotations := util.MergeMaps(rv1.CSV.Annotations, depSpec.Spec.Template.Annotations)
+		annotations["olm.targetNamespaces"] = strings.Join(opts.TargetNamespaces, ",")
+		depSpec.Spec.Template.Annotations = annotations
+
+		// Hardcode the deployment with RevisionHistoryLimit=1 (something OLMv0 does, not sure why)
```
anik120, line 58: Looks like we have the answer for this [here](https://github.com/operator-framework/operator-lifecycle-manager/blob/dfd0b2bea85038d3c0d65348bc812d297f16b8d2/pkg/controller/install/deployment.go#L177-L180) right?

anik120 (commented):
The test cases are *chef's kiss*

internal/operator-controller/rukpak/convert/generate.go
perdasilva: Fixed =D

anik120 (commented):
I realized that while testing is being added to test the `convert` functionalities, the testing is exclusively on the new components (ResourceGenerators, BundleRenders etc). 

Maybe a way to break this PR would have to be add tests in the first PR that tests the conversion exhaustively. 

Then the new PR would refractor and add the new components, with minimal changes to the tests, while the pieces being called would obviously need to change, the behavior would not change in the existing tests. 

That way we can be sure that we're not regressing on any functionality while refactoring and introducing the new components.

perdasilva:
> I realized that while testing is being added to test the `convert` functionalities, the testing is exclusively on the new components (ResourceGenerators, BundleRenders etc).
> 
> Maybe a way to break this PR would have to be add tests in the first PR that tests the conversion exhaustively.
> 
> Then the new PR would refractor and add the new components, with minimal changes to the tests, while the pieces being called would obviously need to change, the behavior would not change in the existing tests.
> 
> That way we can be sure that we're not regressing on any functionality while refactoring and introducing the new components.

I started with that. But, because the convert function was so monolithic it meant each test would need to set itself up in a way that it can get through to the logic it wants to test, e.g. the csv would need to support AllNamespaces (which is not super important when you want to check that strategy deployment spec is properly converted to a deployment resource, etc. It also meant that, for the permissions tests, the resulting resource names would carry the hash (which is also not so important when you want to check that spec.permissions -> rbav1.Role and rbacv1.RoleBinding). Also, the convert function returns _all_ generated resources, which means we need to try and find the expected ones in the pile. Not the end of the world, but adds complexity to write and in case of failure. So, I pivoted away towards adding the regression tests (#1895) , which, tbf might not be as thorough as unit testing,  I thought it still gave me enough of an indication that everything would be ok after refactor. Having said all that, the generator unit tests were portable enough that I brought them over to the current rukpak/convert package (#1912). Maybe as we move through, their short livedness will outweigh their drawbacks and give us extra security. Wdyt?

internal/operator-controller/rukpak/convert/registryv1.go
```diff
@@ -224,22 +212,23 @@ func validateTargetNamespaces(supportedInstallModes sets.Set[string], installNam
...
+			generators.BundleAdditionalResourcesGenerator,
+			generators.BundleCSVDeploymentGenerator,
+		},
+	},
 }
 
 type Converter struct {
-	BundleValidator BundleValidator
+	render.BundleRenderer
 }
```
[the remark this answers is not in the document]
perdasilva: Eventually I'd like to do away with the convert package all together in favor of the renderer. I just didn't want to also carry over all the logic checking the install and target namespaces, and validating install modes, and conversion support, etc. to the renderer (yet). In hindsight, I wish I had hidden everything inside the Convert() function for now to also reduce touch points with other parts of the code (leaving that for later).

I don't quite get the code suggesting. Seems its to rename PlainConverter to RegistryV1ToPlainBundleRenderer?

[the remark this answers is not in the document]
anik120: > Eventually I'd like to do away with the convert package all together in favor of the renderer.

I was talking about this - getting rid of the `Converter` in favor of using `render.BundleRenderer` directly, instead of wrapping `renderer.BundleRender` inside a `Converter` and then using `Converter`. That'd mean 

```
package render 

type BundleRenderer struct{
.
.
}

func (r BundleRenderer) Convert (rv1 RegistryV1, installNamespace string, targetNamespaces []string) (*Plain, error)}
.
.
}
``` 
would have to be defined. 

But that would also mean 

>carry over all the logic checking the install and target namespaces, and validating install modes, and conversion support, etc. to the renderer (yet)

has to be done. It didn't look like this'd be too much of a heavy lift to me, but maybe I'm missing something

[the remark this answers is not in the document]
perdasilva: it's not about the effort, it's just about the scope of the change in the PR - and I'd also like to consider what to do with the Helm converter as part of this, and that will definitely influence everything up from that. I'd also like to think over whether that logic should live in the renderer itself or outside of it, or just in the Plain renderer, etc. It might have made things tidier, but I don't think it's necessary for this PR (though I do think its necessary).

internal/operator-controller/rukpak/render/generators/resources.go
```diff
@@ -0,0 +1,177 @@
...
+			opt(obj)
+		}
+	}
+	return obj
+}
+
+// WithSubjects applies rbac subjects to ClusterRoleBinding and RoleBinding resources
+func WithSubjects(subjects ...rbacv1.Subject) func(client.Object) {
+	return func(obj client.Object) {
+		switch o := obj.(type) {
```
joelanford, line 29: Add a `default` case to panic if the type doesn't support setting a subject?

joelanford: Same question for the other `With*` funcs.

joelanford (dismissed):
Looks great! One nit is all.

thetechnick (approved):

PR #1897, opened 2025-04-03
:bug: Fix conversion regression test

perdasilva:
# Description

In #1895 we added a regression test for the registry+v1 bundle to plain manifest converter that generates plain manifests from a bundle into a directory then uses git diff --exit-code to fail if it detects changes (similar to how verify works with the generated code). Currently, the manifest files are generated in the order and the filename for each manifest includes an index based on that order. E.g. 05_clusterrole_something.yaml. 

To avoid failing on correctly generated manifests that happened to be out of order, this PR sorts the output manifests by kind/namespace/name before writing them out to maintain order over consecutive independent executions. For the test, the order of manifest is not important. What is important is that the same manifests are generated with the same content.

## Reviewer Checklist

- [ ] API Go Documentation


anik120 (commented):
Code changes look pretty minimal. I am just a little lost on the `why` of this PR from just this message 

```
Fixes bundle -> plain manifest generator to sort generated manifest before writing them out to avoid false negatives
```
😅  I am unaware of what these false negatives are, so it'll be helpful for these PRs to have a little more context. For current reviewers who are trying to catch up to the context and for future maintainers who might stumble upon this and need to do something with it. 

There are also a LOT of file renames that I'm not sure I understand the "why it was needed" part of.

perdasilva:
> Code changes look pretty minimal. I am just a little lost on the `why` of this PR from just this message
> 
> ```
> Fixes bundle -> plain manifest generator to sort generated manifest before writing them out to avoid false negatives
> ```
> 
> 😅 I am unaware of what these false negatives are, so it'll be helpful for these PRs to have a little more context. For current reviewers who are trying to catch up to the context and for future maintainers who might stumble upon this and need to do something with it.
> 
> There are also a LOT of file renames that I'm not sure I understand the "why it was needed" part of.

@anik120 fair call - sorry about that. I was rushing because I needed to sign off and wanted this up to help not block any other PRs.

I'll add more description to the context.

thetechnick (approved):

PR #1506, opened 2024-11-27
WIP: ✨ Add support for registry+v1 bundles that include webhooks

joelanford:
# Description

This PR adds support for bundles that define webhooks via:
- A new alpha feature gate called `RegistryV1WebhookSupport` that is disabled by default
- A new `--registry-v1-cert-provider` flag for the manager binary (only present when the feature gate is enabled) that allows someone installing OLMv1 to specify a certificate provider to use for registry+v1 webhooks.
- A cert-manager certificate provider
- An OpenShift Service CA certificate provider

The PR also has a small demo program (`cmd/registryv1-to-helm`) and some example input/output (using cloudnative-pg) that is a light wrapper around the `internal/rukpak/convert` library. I intend to remove all of this before the PR merges.

## Reviewer Checklist

- [ ] API Go Documentation
- [ ] Tests: Unit Tests (and E2E Tests, if appropriate)
- [ ] Comprehensive Commit Messages
- [ ] Links to related GitHub Issue(s)

PR #1914, opened 2025-04-11
Add wh support

perdasilva:
# Description

Depends on #1893

Adds webhook support to bundle renderer with certificate management. Based on Joe's PoC #1506 

## Reviewer Checklist

- [ ] API Go Documentation


internal/operator-controller/rukpak/render/generators/resources.go
```diff
@@ -183,3 +216,46 @@ func CreateDeploymentResource(name string, namespace string, opts ...ResourceCre
...
+
+func CreateValidatingWebhookConfigurationResource(generateName string, namespace string, opts ...ResourceCreatorOption) *admissionregistrationv1.ValidatingWebhookConfiguration {
+	return ResourceCreatorOptions(opts).ApplyTo(
+		&admissionregistrationv1.ValidatingWebhookConfiguration{
+			TypeMeta: metav1.TypeMeta{
+				Kind:       "ValidatingWebhookConfiguration",
+				APIVersion: admissionregistrationv1.SchemeGroupVersion.String(),
+			},
+			ObjectMeta: metav1.ObjectMeta{
+				GenerateName: fmt.Sprintf("%s-", generateName),
```
joelanford, line 228: Using `GenerateName` is what OLMv0 does, correct? I think it does this so that multiple validating/mutating webhook configuration objects can be created when own/single namespace bundles are installed multiple times. 

That is an anti-goal for OLMv1, so perhaps it makes sense to deviate from OLMv0's use of `GenerateName` in this case.

Unless there is a really strong reason not to, I'd prefer to use a deterministic name here instead, as that would ultimately make the set of on-cluster objects deterministic as well. I'm sure there are many benefits, but one that comes to mind is that ClusterExtension SAs would be able to specify the deterministic name in advance when setting up RBAC for `get/update/patch/delete` verbs with the correct `resourceName`.

joelanford: Another benefit: using a deterministic name here means that this webhook can only be installed once, regardless of the install modes supported by the extension. That is directly in line with one of the stated invariants of OLMv1 (no two cluster extensions can manage the same object)

internal/operator-controller/rukpak/util/util.go
```diff
@@ -17,6 +19,21 @@ func ObjectNameForBaseAndSuffix(base string, suffix string) string {
 	return fmt.Sprintf("%s-%s", base, suffix)
 }
 
+func ToUnstructured(obj client.Object) (*unstructured.Unstructured, error) {
+	gvk := obj.GetObjectKind().GroupVersionKind()
```
joelanford, line 23: A `client.Object` is not guaranteed to have a populated GVK. If we need guarantees about setting the GVK on the unstructured type, we should also pass in a scheme so that we can lookup the GVK based on the object type.

joelanford: See https://github.com/kubernetes-sigs/controller-runtime/blob/6ad5c1dd4418489606d19dfb87bf38905b440561/pkg/client/apiutil/apimachinery.go#L95

internal/operator-controller/features/features.go
```diff
@@ -13,6 +13,7 @@ const (
 	// Ex: SomeFeature featuregate.Feature = "SomeFeature"
 	PreflightPermissions             featuregate.Feature = "PreflightPermissions"
 	SingleOwnNamespaceInstallSupport featuregate.Feature = "SingleOwnNamespaceInstallSupport"
+	WebhookSupport                   featuregate.Feature = "WebhookSupport"
```
joelanford, line 16: If we want to add support for new providers later, how do you envision that looking from a feature gate standpoint?

joelanford (approved):
This is great! My only open question (that I already left in a code comment) is how we would introduce more certificate providers in the future?

Presumably we'd want a feature gate per provider, so that each provider could mature indepentently? If so, should we have `CertManagerWebhookSupport` for the current feature gate name?

I don't think we have to answer that question or change what we've got in the PR _right now_, so approving!

internal/operator-controller/rukpak/render/generators/resources.go
perdasilva: I've changed it to use `Name` instead. I'm trying to think if there's foot gun potential. I might be a little more defensive here and ensure that any `-` suffix is removed from what is defined in the CSV to avoid any bad name errors.

We might already, but I'll double check and ensure that all webhook names defined in the CSV (for a particular type of webhook) are unique.

I guess if there is a naming clash, i.e. two different bundles call their webhook "my-webhook", the user can just install the bundle in another namespace as an escape hatch...wdyt?

internal/operator-controller/rukpak/util/util.go
perdasilva: For now, I've just hardened it by checking if version and kind are empty. I don't think we need to add the complexity of passing a scheme. At least not for now =D

camilamacedo86: If we want the GKV and use it is the right approach then, I think we should use the controller-runtime implementation to do so: https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/client/apiutil/apimachinery.go#L95C6-L95C18

internal/operator-controller/rukpak/render/validators/validator.go
```diff
@@ -86,3 +95,144 @@ func CheckPackageNameNotEmpty(rv1 *render.RegistryV1) []error {
...
+		return nil
+	}
+
+	ownedCRDNames := sets.Set[string]{}
+	for _, crd := range rv1.CSV.Spec.CustomResourceDefinitions.Owned {
+		ownedCRDNames.Insert(crd.Name)
+	}
+
+	slices.SortFunc(conversionWebhooks, func(a, b v1alpha1.WebhookDescription) int {
+		return cmp.Compare(a.GenerateName, b.GenerateName)
```
camilamacedo86, line 197: Hi @perdasilva 

I am not finding where we are setting the GenerateName 
Can you help me out and link it ?

perdasilva: It used to be [here](https://github.com/operator-framework/operator-controller/pull/1914/commits/7861571eb73dbda49b27b0c1e0f8c193d5870202#diff-867fc86b4cf60bb0357a2d94f2c61a76d3cf9f0b52d2aa4b7b32f4941cded7c3L312)

perdasilva: For the RBAC resources we create, we use [this](https://github.com/operator-framework/operator-controller/blob/main/internal/operator-controller/rukpak/render/render.go#L120) function, which uses [the original rukpak fn](https://github.com/operator-framework/operator-controller/blob/main/internal/operator-controller/rukpak/util/util.go#L13). This currently keeps names below 63 characters.

camilamacedo86: If we will do == OLV0 them I think that would be right:

OLM v0: Webhooks utilize metadata.generateName from csv.spec.webhookdefinitions.generateName as a prefix. Corresponding services are named <generateName>-service.
- Example: Refer to [webhook.go#L91](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/controller/operators/webhook.go#L91) and [webhook.go#L254](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/controller/operators/webhook.go#L254).

camilamacedo86: If we will do == OLV0 them I think that would be right:

OLM v0: Webhooks utilize metadata.generateName from csv.spec.webhookdefinitions.generateName as a prefix. Corresponding services are named <generateName>-service.
- Example:

https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/controller/install/webhook.go#L91

https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/controller/install/webhook.go#L254

So are we doing the same?

perdasilva: That's essentially what we had as the first pass and Joe thought maybe this would be something where we could break the rule of doing exactly what olmv1 does ([comment](https://github.com/operator-framework/operator-controller/pull/1914#discussion_r2064234391)).

The service name remains the same ([i.e.](https://github.com/operator-framework/operator-controller/pull/1914/files#diff-95654e48627996e345270b32ee6275d5210a580c2d627c29af690027fde9a6dfR66)

internal/operator-controller/rukpak/util/util.go
perdasilva: I don't think we need it right now. The precondition for the function is that the client.Object will carry the GVK. If not, then error.

internal/operator-controller/rukpak/render/validators/validator.go
```diff
@@ -86,3 +95,144 @@ func CheckPackageNameNotEmpty(rv1 *render.RegistryV1) []error {
...
+// has a unique name. Webhooks of different types can have the same name. Errors are sorted by webhook type
+// and name.
+func CheckWebhookNameUniqueness(rv1 *render.RegistryV1) []error {
+	webhookNameSetByType := map[v1alpha1.WebhookAdmissionType]sets.Set[string]{}
+	duplicateWebhooksByType := map[v1alpha1.WebhookAdmissionType]sets.Set[string]{}
+	for _, wh := range rv1.CSV.Spec.WebhookDefinitions {
+		if _, ok := webhookNameSetByType[wh.Type]; !ok {
+			webhookNameSetByType[wh.Type] = sets.Set[string]{}
+		}
+		if webhookNameSetByType[wh.Type].Has(wh.GenerateName) {
```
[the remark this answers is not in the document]
perdasilva: I can add it. But, I think a better approach would be (if it's possible) to validate the csv against its openapi schema. But I don't know that there's a simple way to do that =S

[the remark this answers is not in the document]
perdasilva: what did we do in operator-sdk? Do we have any bundle validation code/library we can rip out from that?

internal/operator-controller/rukpak/render/generators/resources.go
camilamacedo86: So, now I understand better the @joelanford recommendation/idea: https://redhat-internal.slack.com/archives/C0881N26CGP/p1745960594662879?thread_ts=1745942335.360109&cid=C0881N26CGP 

I am OK with since in OLMv1 we should have one instance of each project only the cluster

👍

internal/operator-controller/features/features.go
camilamacedo86: @joelanford could you help me understand what do you mean with `new providers` ?

internal/operator-controller/rukpak/render/validators/validator.go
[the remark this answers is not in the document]
camilamacedo86: > I can add it. But, I think a better approach would be (if it's possible) to validate the csv against its openapi schema. But I don't know that there's a simple way to do that =S

That is definitely a nice idea. 👍 

We’d need to try using the markers for validation: https://book.kubebuilder.io/reference/markers/crd-validation
Add the annotations and re-generate the CRDs where the types are defined.

I think it would be something like

```go
// +kubebuilder:validation:Pattern=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`
// +kubebuilder:validation:MaxLength=253
// +kubebuilder:validation:MinLength=1
```

> what did we do in operator-sdk? Do we have any bundle validation code/library we can rip out from that?

I think we should consider checking this at runtime, since currently it isn’t validated. **But, maybe it's a bit of an overstatement**, though—**if the name is invalid, the same check used by k8s.io/apimachinery/pkg/util/validation (IsDNS1123Subdomain) will be enforced by Kubernetes itself.**

In the past, we used IsDNS1123Subdomain for scaffolding in the SDK (see: https://github.com/operator-framework/operator-sdk/pull/953/files). The SDK uses Kubebuilder, and in Kubebuilder we also rely on the same validation logic (we just copied it to avoid adding a dependency).

If we decide to add the check then, IMO the best approach is to use the checks from k8s.io/apimachinery/pkg/util/validation instead any internal lib.

camilamacedo86 (commented):
I think this PR looks good 👍 
Great work 🥇  

## Only: 

### E2E Test

> **Could we just add an e2e test before merging, or at plan to do so in a follow-up?**  

We might be able to reuse something from here:  
https://github.com/search?q=repo%3Ak8s-operatorhub%2Fcommunity-operators%20generateName&type=code  
(Since we’re using real operators) — this could help validate that the webhooks are working correctly.

##  Otherwise, it shows great and I think we could move forward 🚀

- The new feature is gated behind a feature flag, which ensures safe rollout.
- All tests are passing, so 🚀 
- There’s a potential discussion around renaming the feature ([link](https://github.com/operator-framework/operator-controller/pull/1914#pullrequestreview-2804460601)), but it does not seem a blocker since we could circle back/change it in a follow up
- There's also a minor suggestion to add a validation check for the name ([link](https://github.com/operator-framework/operator-controller/pull/1914/files#r2069561766)). **I’d consider this a nice-to-have that we could tackle in a follow-up. It’s definitely not a blocker for this PR.**

internal/operator-controller/rukpak/util/util.go
camilamacedo86: Do you think that's acceptable since we only use it with CertManager? That way, we'll have the GKV. **Is that what you were thinking?**

internal/operator-controller/features/features.go
camilamacedo86: From what I understand, downstream will use a different provider, but the implementation remains the same.

I don’t see a strong need to support multiple providers upstream for now — downstream will likely stick with one as well. Since this is under an alpha flag, we have flexibility to adjust later.

So I think this is fine as-is and no need to optimize prematurely. We can revisit if we decide to support more providers in the future.

camilamacedo86 (dismissed):
@perdasilva

Everything looks great to me! 👍
Awesome work 🥇

Just one thought: I think it would be valuable to have some e2e tests covering install/upgrade scenarios with webhooks.

**Do you think we could add those before merging? Or, since the feature is behind a flag, maybe we could make sure to cover that in a follow-up?**

Other small nits (nice-to-haves): https://github.com/operator-framework/operator-controller/pull/1914#discussion_r2069561766

Aside from that, I'm all good with it.

/lgtm
/approved

internal/operator-controller/rukpak/render/validators/validator.go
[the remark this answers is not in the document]
perdasilva: I've added the validation functions that use the IsDNS1123Subdomain under the hood. I'm also cooking up a scheme where we generate a validator from based on an imported ClusterServiceVersion CRD. This will give us the ultimate security that the CSV is valid (though additional validators around referential integrity, uniqueness, etc. will still be required)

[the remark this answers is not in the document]
perdasilva: But this validator will come in a follow up PR

perdasilva:
> Just one thought: I think it would be valuable to have some e2e tests covering install/upgrade scenarios with webhooks.
> 
> **Do you think we could add those before merging? Or, since the feature is behind a flag, maybe we could make sure to cover that in a follow-up?**

I think this falls under the discussion around having e2e tests for feature-gated code. I don't know that we've landed on a solid answer yet, and I think we may want a separate e2e suite (maybe?) for FGs? Because of this, I think it will come in a follow-up.

camilamacedo86:
>I think this falls under the discussion around having e2e tests for feature-gated code. 

Without E2E tests, we can't be sure the feature works properly. It also makes it easier to miss bugs or introduce regressions.
I believe we should have E2E tests no matter if the feature is GA, tech-preview, or alpha. In downstream, we already require E2E tests even for tech-preview features.
So I'm not sure why we shouldn't have them just because the feature is alpha. Can you help me understand the reasoning?

internal/operator-controller/rukpak/render/certprovider_test.go
```diff
@@ -0,0 +1,122 @@
+package render_test
```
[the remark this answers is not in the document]
perdasilva: lets do this as a follow up since we have some work we need to do before we can enable e2es for FGs

internal/operator-controller/features/features.go
joelanford: I think you are missing my point (which I should have explained a bit better 😅 ). Let's say we implement webhook support now with cert-manager, but then later decide we want to add a separate implementation (e.g. `BuiltIn` or OPA's cert-controller)?

My point is, I think there is a difference between "we support webhooks" and "we support a certain certificate provider".

I would suggest calling this `WebhookProviderCertManager` so that if/when we introduce other cert providers upstream, we don't have a strange naming convention where `WebhookSupport` _actually_ means cert-manager webhook support and `WebhookSupportFooBar` means "webhook support with the foobar provider".

And then once at least one provider's feature gate graduates to GA, _then_ we can say that we support bundles with webhooks.

I also think we should explore implementing OCP's Service CA provider in our upstream.

That would benefit Red Hat: if Red Hat implements it only downstream, they (we) would have to carry patches in the downstream codebase that could invite conflicts when syncing. There is no precedent yet for carrying a patch in the downstream Go source code, and I'm not convinced it makes sense to change that for this feature.

It would also benefit other users of OCP. For example, if an OCP user wants to disable OCP's OLM and use upstream OLM instead, they would still be able to use the service-ca provider. We've talked about this possibility with SRE organizations that provide managed OpenShift and that want a consistent OLM surface across a fleet of OCP clusters with varying OCP versions.

internal/operator-controller/rukpak/util/util.go
joelanford: Yeah, if we check/fail if GVK is unset, then I think that's sufficient.

internal/operator-controller/features/features.go
joelanford: Another general benefit: If we use a naming convention like `WebhookProvider<ProviderName>` for the feature gates, we can directly translate the set of enabled webhook providers to a `--webhook-provider` flag's presence and the set of allowed values (i.e. the set of `<ProviderName>` suffixes).

Then, distro maintainers retain the choice of:
- no webhook support (don't set the flag)
- webhook support with whichever provider they prefer (they set the flag to that provider)

All this driven by the feature gate naming convention.

perdasilva: I've updated the featuregate name

joelanford (approved):

PR #1921, opened 2025-04-15
:seedling: Change how systemroot is created when SSL_CERT environment is set

tmshort:
# Description

The x509.SystemCertPool() looks at the SSL_CERT_FILE and SSL_CERT_DIR
environment variables to generate the pool. However, if the contents of
the referenced file (singular) or directories (multiple) change, there
is no guarantee that x509.SystemCertPool() will be updated. Since we
are watching these locations (defined by the environment) via fsnotify,
we want to ensure that when those files are updated that the cert pool
we use is also updated.

So, if SSL_CERT_FILE or SSL_CERT_DIR are defined, create our cert pool
from those variable _only_, ignoring the x509.SystemCertPool().

This is how the x509.SystemCertPool() would be created, so we do it
explicitly instead. This allows us to properly refresh the pool when
fsnotify tells us there are changes to our watches.

This does not impact images/containers (i.e. impage pulling) directly,
since that still uses x509.SystemCertPool(), so it may get a stale pool,
but the catalogd client will have an up-to-date pool.

See: https://pkg.go.dev/crypto/x509#SystemCertPool

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/shared/util/http/certpoolwatcher_test.go
+++ b/internal/shared/util/http/certpoolwatcher_test.go
@@ -91,6 +91,7 @@ func TestCertPoolWatcher(t *testing.T) {
 	t.Logf("Create cert file at %q\n", certName)
 	createCert(t, certName)
 
+	// Eventually, the pools updates
 	require.Eventually(t, func() bool {
 		secondPool, secondGen, err := cpw.Get()
 		if err != nil {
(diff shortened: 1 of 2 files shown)


tmshort:
closing, not needed.

Issue #1903, opened 2025-04-04
Collect metrics summary after e2e testing


dtfranz:
Upon completion of e2e test run, collect and output a summary of our metrics results.

Please see [here](https://github.com/camilamacedo86/operator-controller/blob/main/docs/draft/howto/consuming-metrics.md#integrating-the-metrics-endpoints-with-third-party-solutions) for documented steps on how to get started with prometheus scraping for OLMv1. This should be set up before the e2e tests run, so that once they finish the metrics should be available for query via PromQL from prometheus.

dtfranz:
Hi @Horiodino , my apologies, this issue should have been closed as of #1928 but clearly the issue closing automation doesn't work when you string together multiple issues like I did 🤦 . The remaining tasks for this epic are assigned to me and I'll be pushing up a PR for them very shortly if you're interested in following that.

PR #1928, opened 2025-04-17
🌱 Add prometheus to e2e workflow

### LINKED ISSUE #1902: Install Prometheus on e2e cluster
As part of e2e cluster setup, install prometheus to gather metrics from operator-controller and catalogd when using the test-e2e-metrics makefile target.

When e2e tests are run through our makefile targets, there are a number of setup jobs that run to create the runtime environment for e2e, including: kind-cluster, kind-load, kind-deploy, etc. Once operator-controller and catalogd have been installed on the kind cluster, a makefile target should be run to install prometheus. To keep this overhead separate from the normal maintainer testing workflow of running test-e2e/e2e targets, this environment should only be included when running the test-e2e-metrics target.

dtfranz:
# Description

Closes #1902 and #1903

Adds prometheus to the `test-e2e` Makefile target, which stands up a barebones prometheus scraper to gather metrics from the operator-controller and catalogd pods during the e2e test run. When finished, the prometheus server is queried for a raw output of the metrics and stores it in `ARTIFACTS_PATH/metrics.out`. These metrics will be analyzed in a later PR.

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -256,7 +256,24 @@ image-registry: ## Build the testdata catalog used for e2e tests and push it to 
 test-e2e: KIND_CLUSTER_NAME := operator-controller-e2e
 test-e2e: KUSTOMIZE_BUILD_DIR := config/overlays/e2e
 test-e2e: GO_BUILD_EXTRA_FLAGS := -cover
-test-e2e: run image-registry e2e e2e-coverage kind-clean #HELP Run e2e test suite on local kind cluster
+test-e2e: run image-registry prometheus e2e e2e-metrics e2e-coverage kind-clean #HELP Run e2e test suite on local kind cluster
+
+.PHONY: prometheus
+prometheus: PROMETHEUS_NAMESPACE := olmv1-system
+prometheus: PROMETHEUS_VERSION := v0.83.0
+prometheus: #HELP Deploy Prometheus into specified namespace
+	./hack/test/setup-monitoring.sh $(PROMETHEUS_NAMESPACE) $(PROMETHEUS_VERSION) $(KUSTOMIZE)
+
+# The metrics.out file contains raw json data of the metrics collected during a test run.
+# In an upcoming PR, this query will be replaced with one that checks for alerts from
+# prometheus. Prometheus will gather metrics we currently query for over the test run, 
+# and provide alerts from the metrics based on the rules that we set.
+.PHONY: e2e-metrics
+e2e-metrics: #HELP Request metrics from prometheus; place in ARTIFACT_PATH if set
+	curl -X POST \
+	-H "Content-Type: application/x-www-form-urlencoded" \
+	--data 'query={pod=~"operator-controller-controller-manager-.*|catalogd-controller-manager-.*"}' \
+	http://localhost:30900/api/v1/query > $(if $(ARTIFACT_PATH),$(ARTIFACT_PATH),.)/metrics.out
 
 .PHONY: extension-developer-e2e
 extension-developer-e2e: KUSTOMIZE_BUILD_DIR := config/overlays/cert-manager
--- a/hack/test/setup-monitoring.sh
+++ b/hack/test/setup-monitoring.sh
@@ -1,0 +1,222 @@
+#!/bin/bash
+
+set -euo pipefail
+
+help="setup-monitoring.sh is used to set up prometheus monitoring for e2e testing.
+
+Usage:
+  setup-monitoring.sh [PROMETHEUS_NAMESPACE] [PROMETHEUS_VERSION] [KUSTOMIZE]
+"
+
+if [[ "$#" -ne 3 ]]; then
+  echo "Illegal number of arguments passed"
+  echo "${help}"
+  exit 1
+fi
+
+NAMESPACE=$1
+PROMETHEUS_VERSION=$2
+KUSTOMIZE=$3
+
+TMPDIR=$(mktemp -d)
+trap 'echo "Cleaning up ${TMPDIR}"; rm -rf "${TMPDIR}"' EXIT
+curl -s "https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/refs/tags/${PROMETHEUS_VERSION}/kustomization.yaml" > "${TMPDIR}/kustomization.yaml"
+curl -s "https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/refs/tags/${PROMETHEUS_VERSION}/bundle.yaml" > "${TMPDIR}/bundle.yaml"
+(cd ${TMPDIR} && ${KUSTOMIZE} edit set namespace ${NAMESPACE}) && kubectl create -k "${TMPDIR}"
+kubectl wait --for=condition=Ready pods -n ${NAMESPACE} -l app.kubernetes.io/name=prometheus-operator
+
+kubectl apply -f - << EOF
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: prometheus
+  namespace: ${NAMESPACE}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: prometheus
+rules:
+- apiGroups: [""]
+  resources:
+  - nodes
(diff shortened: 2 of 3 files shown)


tmshort:
So, this adds a new e2e that isn't run downstream? There's no workflow to run it in upstream CI?

Makefile
```diff
@@ -252,6 +252,23 @@ test-e2e: KUSTOMIZE_BUILD_DIR := config/overlays/e2e
...
 
+.PHONY: test-e2e-metrics
+test-e2e-metrics: KIND_CLUSTER_NAME := operator-controller-e2e
+test-e2e-metrics: KUSTOMIZE_BUILD_DIR := config/overlays/e2e
+test-e2e-metrics: GO_BUILD_EXTRA_FLAGS := -cover
+test-e2e-metrics: run image-registry prometheus e2e e2e-coverage kind-clean #HELP Run e2e test suite on local kind cluster with prometheus installed
+
+.PHONY: prometheus
+prometheus: PROMETHEUS_NAMESPACE := prometheus
+prometheus: TMPDIR := $(shell mktemp -d)
```
tmshort, line 263: Do you delete this `$(TMPDIR)`?

dtfranz: `mktemp` creates a directory in `/tmp` - if you think it's necessary to explicitly delete the directory in there we can do that easily :)

bentito: yeah, it's bad to not delete, if you want to make sure it goes away in the same target, you could add:
`trap 'echo "Cleaning up $$tmpdir for $@"; rm -rf "$$tmpdir"' EXIT;`

dtfranz: Done! Please feel free to un-resolve if the solution needs work.

Makefile
```diff
@@ -252,6 +252,24 @@ test-e2e: KUSTOMIZE_BUILD_DIR := config/overlays/e2e
...
+prometheus: LATEST := $(shell curl -s https://api.github.com/repos/prometheus-operator/prometheus-operator/releases/latest | jq -cr .tag_name)
+prometheus: ## Deploy Prometheus into 'prometheus' namespace
+	trap 'echo "Cleaning up $(TMPDIR)"; rm -rf "$(TMPDIR)"' EXIT; \
+	kubectl create ns $(PROMETHEUS_NAMESPACE); \
+	curl -s "https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/refs/tags/$(LATEST)/kustomization.yaml" > "$(TMPDIR)/kustomization.yaml" ; \
+	curl -s "https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/refs/tags/$(LATEST)/bundle.yaml" > "$(TMPDIR)/bundle.yaml" ; \
+	(cd $(TMPDIR) && $(KUSTOMIZE) edit set namespace $(PROMETHEUS_NAMESPACE)) && kubectl create -k "$(TMPDIR)" ; \
+	kubectl wait --for=condition=Ready pods -n $(PROMETHEUS_NAMESPACE) -l app.kubernetes.io/name=prometheus-operator
```
joelanford, line 271: Did you consider adding the prometheus setup/teardown into the e2e's `test_main.go`? https://github.com/openshift/operator-framework-operator-controller/blob/355dcf40d5fcb4d9e93ab312a376c4e80d46ba76/test/e2e/e2e_suite_test.go#L31-L40

That would keep the Makefile a bit cleaner, and it would be a bit easier to deal with the tmpdir cleanup.

dtfranz: I think this would go against our convention of setting up and tearing down everything in the Makefile. If people want to make changes to the e2e environment, they're going to be looking for things in here, and not in the go test files IMO.

Wouldn't that also add prometheus to every test run? I was hoping not to do that. I assume there's probably ways to turn it off/on in the `go test` calls, but then that adds more complication, and then we'll need to add additional targets in the Makefile for that too.

If your goal is just to keep the Makefile clean, I would be happy to compromise by adding a script instead and calling it from the Makefile. It's mainly just that putting it into the go test files doesn't feel right to me.

dtfranz:
Finally getting back to this. @joelanford @tmshort Do you still have any remaining concerns?

Makefile
dtfranz: I've updated the PR to make this functional - I'll set this thread as resolved unless you feel it's still relevant.

Makefile
```diff
@@ -256,7 +256,17 @@ image-registry: ## Build the testdata catalog used for e2e tests and push it to
...
+
+.PHONY: prometheus
+prometheus: PROMETHEUS_NAMESPACE := olmv1-system
+prometheus: PROMETHEUS_VERSION := v0.83.0
+prometheus: #HELP Deploy Prometheus into specified namespace
+	./hack/test/setup-monitoring.sh $(PROMETHEUS_NAMESPACE) $(PROMETHEUS_VERSION) $(KUSTOMIZE)
+
+.PHONY: e2e-metrics
+e2e-metrics: #HELP Request metrics from prometheus; place in ARTIFACT_PATH if set
+	curl http://localhost:30900/metrics > $(if $(ARTIFACT_PATH),$(ARTIFACT_PATH),.)/metrics.out
```
camilamacedo86: Hey, just trying to understand a bit better — in this test, are we getting the output from Prometheus itself? Or is the test directly calling the metrics endpoint on catalogd and operator-controller and do not add this complexities at all?

I was wondering what the difference is between the output we get here and the output we’d see if we just hit the catalogd or operator-controller metrics endpoint directly (like with a curl).

If it’s not too much trouble,  could you add/attach the metrics.out in the description? 

How will we use it?

Thanks a lot — just trying to learn and follow along!

dtfranz: If we directly query `/metrics` from catalogd and operator-controller, we will just get a snapshot of raw metrics data. By using and querying prometheus, we get the added benefits of a single endpoint, advanced content and time-based filtration, output formatted in json, and the ability to create Alerts which is what we will be using initially to look for issues during our e2e runs. Once we have alerts established, we can simply query `/alerts` and fail a test run if any are encountered. These could be performance or functionality related.

That being said, I'm currently hitting the wrong endpoint here 😅 I'll fix up this query. It'll still just be raw data though until our alerts are set up.

dtfranz: I've fixed up the query. This now outputs metrics from the catalogd and operator-controller pods into `metrics.out`, formatted in json, but otherwise raw data from the whole test run. As I said before, I'll create alert rules for this data so we can get more meaningful information from it.

camilamacedo86: 👍 That is great we just need also

a) do we want to commit the metrics.out? if not then we need to add in the gitignore
b) Could we add a comment on top of metrics.out makefile target explain how, why that is required? How it is used ?

Then I am 1000% fine with this one

hack/test/setup-monitoring.sh
```diff
@@ -0,0 +1,222 @@
...
+  podSelector:
+    matchLabels:
+      app.kubernetes.io/name: prometheus
+  policyTypes:
+    - Egress
+    - Ingress
+  egress:
+    - {}  # Allows all egress traffic for metrics requests
+  ingress:
+    - {}  # Allows us to query prometheus
```
camilamacedo86, line 114: I understand that we need this NP to allow us to get data from Prometheus integrate it with Grafana and etc. So, I am OK with 👍

camilamacedo86 (approved):
@dtfranz 

Thank you for the nice work. 🎉 

I understand that we will need to improve it in the long run and that is just teh first step for we have the targets and quick start.

So, I just added a nit: https://github.com/operator-framework/operator-controller/pull/1928#discussion_r2141987384 but otherwise I am OK with it. LGTM

We might also want to have a second look and a help from @tmshort too

Makefile
dtfranz: The metrics output file uses the `.out` extension, which is covered by our .gitignore already, so there's no need to update it :)

And yes, I can add a comment here!

dtfranz:
I reran it and it passed, seems a test flaked 🫤

Makefile
dtfranz: Updated!

PR #1934, opened 2025-04-21
🐛 OCPBUGS-55165 Permissions preflight error output fixes

trgeiger:
# Description
This PR contains 2 changes:
- Adds an errors.Join() to properly expand a list of errors that might include nil errors
- Adds explicit rule examples to 2 of the test cases

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/operator-controller/applier/helm.go
+++ b/internal/operator-controller/applier/helm.go
@@ -109,7 +109,7 @@ func (h *Helm) runPreAuthorizationChecks(ctx context.Context, ext *ocv1.ClusterE
 		preAuthErrors = append(preAuthErrors, fmt.Errorf("authorization evaluation error: %w", authErr))
 	}
 	if len(preAuthErrors) > 0 {
-		return fmt.Errorf("pre-authorization failed: %v", preAuthErrors)
+		return fmt.Errorf("pre-authorization failed: %v", errors.Join(preAuthErrors...))
 	}
 	return nil
 }
--- a/internal/operator-controller/authorization/rbac_test.go
+++ b/internal/operator-controller/authorization/rbac_test.go
@@ -214,6 +214,195 @@ subjects:
 			},
 		},
 	}
+
+	expectedSingleNamespaceMissingRules = []ScopedPolicyRules{
+		{
+			Namespace: "",
+			MissingRules: []rbacv1.PolicyRule{
+				{
+					Verbs:           []string{"list", "watch"},
+					APIGroups:       []string{""},
+					Resources:       []string{"services"},
+					ResourceNames:   []string(nil),
+					NonResourceURLs: []string(nil)},
+				{
+					Verbs:           []string{"list", "watch"},
+					APIGroups:       []string{"rbac.authorization.k8s.io"},
+					Resources:       []string{"rolebindings"},
+					ResourceNames:   []string(nil),
+					NonResourceURLs: []string(nil)},
+				{
+					Verbs:           []string{"list", "watch"},
+					APIGroups:       []string{"rbac.authorization.k8s.io"},
+					Resources:       []string{"roles"},
+					ResourceNames:   []string(nil),
+					NonResourceURLs: []string(nil)},
+				{
+					Verbs:           []string{"update"},
+					APIGroups:       []string{""},
+					Resources:       []string{"clusterextensions/finalizers"},
+					ResourceNames:   []string{"test-cluster-extension"},
+					NonResourceURLs: []string(nil),
+				},
+			},
+		},
+		{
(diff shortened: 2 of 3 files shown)


internal/operator-controller/applier/helm.go
```diff
@@ -103,7 +103,7 @@ func (h *Helm) runPreAuthorizationChecks(ctx context.Context, ext *ocv1.ClusterE
 			}
 		}
 		slices.Sort(missingRuleDescriptions)
-		preAuthErrors = append(preAuthErrors, fmt.Errorf("service account requires the following permissions to manage cluster extension:\n  %s", strings.Join(missingRuleDescriptions, "\n  ")))
+		preAuthErrors = append(preAuthErrors, fmt.Errorf("service account requires the following permissions to manage cluster extension:\n  %s\n", strings.Join(missingRuleDescriptions, "\n  "))) //nolint:stylecheck
```
bentito, line 106: We should probably add a comment here saying why this newline is needed.

But I thought about it a bit more and I wonder if we should try to not put the newlines in the error, make the tests assert on the error contents without the newline and then post-process newlines and maybe bullets for each point, into the rendered return:

so like:
```golang
       // join descriptions with commas; formatting (newlines/bullets) will be applied at error render time
       preAuthErrors = append(preAuthErrors,
           fmt.Errorf("service account requires the following permissions to manage cluster extension: %s",
               strings.Join(missingRuleDescriptions, ", ")))
   }
	if authErr != nil {
		preAuthErrors = append(preAuthErrors, fmt.Errorf("authorization evaluation error: %w", authErr))
	}
   if len(preAuthErrors) > 0 {
       // Render the collected errors with line breaks for clarity
       var b strings.Builder
       b.WriteString("pre-authorization failed:")
       for _, e := range preAuthErrors {
           b.WriteString("\n  - ")
           b.WriteString(e.Error())
       }
       return errors.New(b.String())
   }
   return nil
```

internal/operator-controller/authorization/rbac_test.go
```diff
@@ -241,7 +241,7 @@ func TestPreAuthorize_Failure(t *testing.T) {
 		fakeClient := setupFakeClient(limitedClusterRole)
 		preAuth := NewRBACPreAuthorizer(fakeClient)
 		missingRules, err := preAuth.PreAuthorize(context.TODO(), &exampleClusterExtension, strings.NewReader(testManifest))
-		require.Error(t, err)
+		require.NoError(t, err)
```
bentito, line 244: but aren't these failure type tests that should require an error?

bentito: oh, I see we're handling the nil properly now

bentito: but then maybe this and the next test should change names and we should test for what is in them and not what's not? Name change like "preauthorize finds the missing RBAC rules"

internal/operator-controller/applier/helm.go
trgeiger: Changing the missing rules join is going to make the rules output into one giant long string. It ends up as one single error in preAuthErrors so your loop is only adding a newline at the last rule.

bentito: Hmm, not my intention, now I gotta go try it maybe not just suggest it ;-) Was hoping for that `range` with comma separated errors would solve the nolint and allow for building the errors into human readable form with a return to separate them from the context.

internal/operator-controller/authorization/rbac_test.go
joelanford: +1. We should change `require.NotEqual` to a `require.Equal` with the expected missing rules.

internal/operator-controller/applier/helm.go
joelanford: Maybe we simply need to `errors.Join(preAuthErrors...)` with no other changes?

```go
	if len(preAuthErrors) > 0 {
		return fmt.Errorf("pre-authorization failed: %v", errors.Join(preAuthErrors...))
	}
```

trgeiger: Change made, it seems to work

internal/operator-controller/authorization/rbac_test.go
trgeiger: Done

bentito (approved):
LGTM

internal/operator-controller/applier/helm.go
bentito: 👍

camilamacedo86 (approved):
It seems fine for me 
Thank you for the nice work 🥇 
/lgtm

PR #1938, opened 2025-04-25
✨ Reformat error wrapping for resolved bundle status

trgeiger:
Adds another test case to helm_test.go for when PreAuthorize() returns missing RBAC rules.

# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/operator-controller/applier/helm_test.go
+++ b/internal/operator-controller/applier/helm_test.go
@@ -16,6 +16,7 @@ import (
 	"helm.sh/helm/v3/pkg/release"
 	"helm.sh/helm/v3/pkg/storage/driver"
 	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	featuregatetesting "k8s.io/component-base/featuregate/testing"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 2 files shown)


internal/operator-controller/applier/helm_test.go
```diff
@@ -163,6 +155,33 @@ spec:
...
+					Verbs:     []string{"create"},
+					APIGroups: []string{"*"},
+					Resources: []string{"certificates"}},
+			},
+		},
+	}
+
+	errMissingRBAC = "pre-authorization failed: service account requires the following permissions to manage cluster extension:\n" +
+		"  Namespace:\"\" APIGroups:[] Resources:[services] Verbs:[list,watch]\n" +
+		"  Namespace:\"test-namespace\" APIGroups:[*] Resources:[certificates] Verbs:[create]\n"
```
joelanford, line 184: I would suggest using backticks to format multi-line strings and/or strings containing quotation marks:
```suggestion
	errMissingRBAC = `pre-authorization failed: service account requires the following permissions to manage cluster extension:
  Namespace:"" APIGroups:[] Resources:[services] Verbs:[list,watch]
  Namespace:"test-namespace" APIGroups:[*] Resources:[certificates] Verbs:[create]
`
```

internal/operator-controller/applier/helm.go
```diff
@@ -109,7 +109,9 @@ func (h *Helm) runPreAuthorizationChecks(ctx context.Context, ext *ocv1.ClusterE
 		preAuthErrors = append(preAuthErrors, fmt.Errorf("authorization evaluation error: %w", authErr))
 	}
 	if len(preAuthErrors) > 0 {
-		return fmt.Errorf("pre-authorization failed: %v", errors.Join(preAuthErrors...))
+		// End the error in a newline since this is a heavily formatted collection of errors such as missing RBAC rules
+		// and we don't want it to run along into any following errors, i.e. the status error wrapping in clusterextension_controller
```
joelanford, line 113: Is there an example of where this actually happens? If so, that's probably because we are not using `errors.Join` in _that_ location?

(i.e. if we used errors.Join all the way up the stack, we'd get newlines between all the errors, but no extraneous newlines otherwise)

joelanford (approved):

PR #1939, opened 2025-04-27
:bug: set the SElinux type

jianzhangbjz:
Address https://github.com/openshift/operator-framework-operator-controller/pull/329

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/config/components/tls/operator-controller/patches/manager_deployment_cert.yaml
+++ b/config/components/tls/operator-controller/patches/manager_deployment_cert.yaml
@@ -16,3 +16,6 @@
 - op: add
   path: /spec/template/spec/containers/0/args/-
   value: "--tls-key=/var/certs/tls.key"
+- op: add
+  path: /spec/template/spec/securityContext/seLinuxOptions
+  value: {"type":"spc_t"}


config/components/tls/operator-controller/patches/manager_deployment_cert.yaml
```diff
@@ -16,3 +16,6 @@
 - op: add
   path: /spec/template/spec/containers/0/args/-
   value: "--tls-key=/var/certs/tls.key"
+- op: add
+  path: /spec/template/spec/securityContext/seLinuxOptions
+  value: {"type":"spc_t"}
```
camilamacedo86, line 21: Hey 👋

I wanted to share a quick note about this part of the change:

```
seLinuxOptions:
  type: spc_t
```

Just a heads up: in SELinux, `spc_t` stands for **"Super Privileged Container"**. It grants elevated privileges and relaxes some of the default confinement SELinux provides — which might not be something we want to apply by default, especially if it's not strictly necessary.

That said, I totally get that you're trying to address an issue here — but I think the root cause may not be related to SELinux permissions. The error message you're seeing appears to be TLS-related:

```
error creating image source: pinging container registry https://my-route-jian.apps.heli-419.qe.devcluster.openshift.com/v2/: tls: failed to verify certificate: x509: certificate signed by unknown authority
```

This looks more like a trust/handshake problem with the registry's certificate rather than a container-level permissions issue. I'm not sure how granting additional SELinux privileges would help in this scenario. We might still unable to get the right cert, but how could that solve it out? What are your thoughts about this one? 

Would you mind explaining a bit more about why you think this is the right solution? Totally open to learning more if there’s something I’m missing. Moreover, it is an issue that does not occur in upstream,, so I do not think that, if required, we should add the path to grant permissions here. 

Thanks! 🙏

camilamacedo86: Could we close this one as deferred?
@tmshort @jianzhangbjz WDYT?

jianzhangbjz: Thanks for reviewing! The root cause of this issue is `Permission denied,` which means the container's default process cannot access the `/etc/docker`. 
```go
jiazha-mac:~ jiazha$ oc exec catalogd-controller-manager-5c768bcf55-fk9ws -- ls -R /etc/docker
ls: cannot open directory '/etc/docker': Permission denied
```
I closed it since it is an issue that does not occur in upstream.

PR #1942, opened 2025-04-29
✨ add NetworkPolicy objects for catalogd and operator-controller

joelanford:
# Description

This PR adds `NetworkPolicy` objects, 1 each for operator-controller and catalogd.

Both network policies allow:
- all egress traffic (for communication with arbitrary image registries)
- ingress traffic to the respective metrics ports

In addition, catalogd's network policy also allows
- ingress traffic to the mutating webhook server
- ingress traffic to the catalogd http server

## Reviewer Checklist

- [ ] API Go Documentation


joelanford:
I did verify that the NetworkPolicy objects have an effect in `kind` clusters. For example, I:
- changed the egress policy to be empty (which means allow no egress). This broke both catalogd and operator-controller.
- changed the ingress policy of catalogd such that the catalogd http server was denied. This broke operator-controller.

config/base/catalogd/manager/network_policy.yaml
```diff
@@ -0,0 +1,22 @@
...
+  ingress:
+    - ports:
+        - protocol: TCP
+          port: 7443 # metrics
+        - protocol: TCP
+          port: 8443 # catalogd http server
+        - protocol: TCP
+          port: 9443 # webhook
+  egress:
+    - {}  # Allows all egress traffic (needed to pull catalog images from arbitrary image registries)
```
camilamacedo86, line 22: Nit: Add space at the end of the file.

config/base/catalogd/manager/network_policy.yaml
```diff
@@ -0,0 +1,22 @@
...
+metadata:
+  name: controller-manager
+  namespace: system
+spec:
+  podSelector:
+    matchLabels:
+      control-plane: catalogd-controller-manager
+  policyTypes:
+    - Ingress
+    - Egress
```
camilamacedo86, line 12: I think we jyst need Ingress right?

config/base/operator-controller/manager/network_policy.yaml
```diff
@@ -0,0 +1,18 @@
...
+      control-plane: operator-controller-controller-manager
+  policyTypes:
+    - Ingress
+    - Egress
+  ingress:
+    - ports:
+        - protocol: TCP
+          port: 8443 # metrics
+  egress:
+    - {}  # Allows all egress traffic (needed to pull bundle images from arbitrary image registries)
```
camilamacedo86, line 18: At the time, when we added it Kubebuilder, came a request from the community for we use labels 
So, that if we have a Pod that want to consume the metrics for example then requires the labels, you might want to give a look on it: https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v4/config/network-policy/allow-metrics-traffic.yaml#L19-L24

config/base/catalogd/manager/network_policy.yaml
camilamacedo86: First I did like you there, and I received this feedback from community: https://github.com/kubernetes-sigs/kubebuilder/pull/3853#discussion_r1685714616

So, we might need to reconsider.

joelanford:
Installing a separate CNI is no longer necessary. See:
- https://github.com/operator-framework/operator-controller/pull/1942#issuecomment-2838949442
- https://github.com/kubernetes-sigs/kind/pull/3612

config/base/catalogd/manager/network_policy.yaml
joelanford: We explicitly want to allow all egress traffic both catalogd and operator-controller because they need to be able to communicate with unknown-in-advance image registries. If we are not explicitly granting that access, then another NetworkPolicy that matches our pod that is more restrictive could break us.

config/base/operator-controller/manager/network_policy.yaml
joelanford: Given that we are fully open now, I think keeping it simple and slightly more open is better for this first pass. If we find later that we can ratchet down access to metrics and the catalogd webhook port in a general way, we can do that later.

config/base/catalogd/manager/network_policy.yaml
joelanford: Done!

config/base/operator-controller/manager/network_policy.yaml
camilamacedo86: OK, we can do a follow-up after checking that this config works on the downstream as well.
It is fine.

config/base/catalogd/manager/network_policy.yaml
camilamacedo86: I think we'll need to be more restrictive to make sense.
But I get your point. 
We can also enhance it on follow ups. So, 👍

camilamacedo86 (dismissed):
/lgtm
/approved

bentito:
Should this have a new e2e specific to validating the reduced network access are working as intended?

joelanford:
> Should this have a new e2e specific to validating the reduced network access are working as intended?

Brett and I discussed this a bit offline and couldn't immediately figure out a concrete test to add. 
- What we _can_ say is that the current NetworkPolicy objects are working because the e2e's are still passing.
- But, we _cannot_ say that the current NetworkPolicy objects are the minimum necessary for the e2e's to pass.

It would be nice to be able to introduce a higher barrier for changing network policy because it is a security-sensitive change. We would not want to accidentally open up network access unnecessarily. 

From a technical perspective, I'm not sure tests that probing blocked ports would work all that well because we may need to have something actually listening on the blocked port and matched by the network policy selector in order to ever see something other than a network failure (would the client see a difference between "NetworkPolicy blocked the traffic" and "nothing is listening there"?)

Perhaps we could have a meta-test like "make sure the NetworkPolicy looks like $this", where "$this" is some sort of annotated list of rules where a justification is provided. If we had that, then:
- PR authors would have to change "$this" if they wanted to change the NetworkPolicy.
- Reviewers would be confronted with the change and would have a chance to see the justification for the change, which would theoretically provide more context to a review than an opaque two-line YAML change in the NetworkPolicy definition.

camilamacedo86 (approved):
/lgtm

PR #1955, opened 2025-05-07
:bug: Ease CI requirements for go-verdiff and codecov

tmshort:
1) Remove path from now-required go-verdiff CI

We're experiencing a weird behavior with this CI being required, and running only on certain paths.

2) Add a 2% threshold for codecov

# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/.github/workflows/go-verdiff.yaml
+++ b/.github/workflows/go-verdiff.yaml
@@ -1,9 +1,6 @@
 name: go-verdiff
 on:
   pull_request:
-    paths:
-    - '**.mod'
-    - '.github/workflows/go-verdiff.yaml'
     branches:
     - main
 jobs:
--- a/codecov.yml
+++ b/codecov.yml
@@ -5,6 +5,11 @@ codecov:
 # Configure the paths to include in coverage reports.
 # Exclude documentation, YAML configurations, and test files.
 coverage:
+  status:
+    project:
+      default:
+        target: auto
+        threshold: 2%
   paths:
     - "api/"
     - "cmd/"


.github/workflows/go-verdiff.yaml
```diff
@@ -1,9 +1,6 @@
 name: go-verdiff
 on:
   pull_request:
-    paths:
-    - '**.mod'
-    - '.github/workflows/go-verdiff.yaml'
```
camilamacedo86, line 6: We should only run it when we change the go,mod and the action itself right?

grokspawn: The problem is that GHA conditional stuff by path doesn't seem to align with prow/tide's idea of optional. So when there aren't changes in the path tripwires, tide waits forever for status that will never come (since the GHA doesn't run).

grokspawn: I would personally prefer that we find a way to make prow better-informed about conditionality so that we could continue doing targeted CI upstream, which generally makes sense.

tmshort: It's causing problems when it's a required test; it doesn't provide any results. So, we need to have it run all the time if we want to make it a required test.

tmshort: In other words, "This is a required test, but oh, I'm only supposed to run it when these files are changed." So, unless those file are changed, it doesn't run, doesn't generate any results, and thus the CI hangs waiting for it.

grokspawn (approved):

PR #1956, opened 2025-05-07
:sparkles: (hack) Make service-account optional

perdasilva:
# Description

* makes the olmv1 service account admin
* makes the .spec.serviceAccount optional
* sets create namespace to true in the helm applier

```
cat <<EOF | kubectl create -f -
apiVersion: olm.operatorframework.io/v1
kind: ClusterExtension
metadata:
  name: zookeeper-operator
spec:
  namespace: zookeeper-operator
  source:
    sourceType: Catalog
    catalog:
      packageName: zookeeper-operator
      version: 0.17.0
```

```
kubectl wait clusterextension zookeeper-operator --for=condition=Installed=true
```

just_works

## Demo

[asciicast](https://asciinema.org/a/xoIwqFGpEiLgwvdnUIHqf7S0V)

## Reviewer Checklist

- [ ] API Go Documentation


internal/operator-controller/controllers/clusterextension_controller.go
```diff
@@ -97,6 +97,7 @@ type InstalledBundleGetter interface {
 //+kubebuilder:rbac:groups=core,resources=serviceaccounts/token,verbs=create
 //+kubebuilder:rbac:groups=apiextensions.k8s.io,resources=customresourcedefinitions,verbs=get
 //+kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=clusterroles;clusterrolebindings;roles;rolebindings,verbs=list;watch
+//+kubebuilder:rbac:groups=*,resources=*,verbs=*
```
camilamacedo86, line 100: Would not it result in OLM have all possible permissions, do we want that?

internal/operator-controller/action/restconfig.go
```diff
@@ -15,6 +15,9 @@ import (
 func ServiceAccountRestConfigMapper(tokenGetter *authentication.TokenGetter) func(ctx context.Context, o client.Object, c *rest.Config) (*rest.Config, error) {
 	return func(ctx context.Context, o client.Object, c *rest.Config) (*rest.Config, error) {
 		cExt := o.(*ocv1.ClusterExtension)
+		if cExt.Spec.ServiceAccount == nil {
+			return rest.CopyConfig(c), nil
+		}
```
camilamacedo86, line 20: So, can we create one to bind if the user does not provide an SA? 
Is that?

perdasilva: the point of this PR was just to demo the kind of UX I'm after. It just made it so that you could stamp out a ClusterExtension and everything would "just work". I didn't really think about all the odds and ends. I think the idea here was: olm is admin, if user doesn't specify an sa, just use olm's.

internal/operator-controller/controllers/clusterextension_controller.go
perdasilva: that's an open question at the moment. But given that we can create tokens for any service account, that's essentially what we have already XDD

PR #1964, opened 2025-05-13
:seedling: Allow override of go-verdiff result via label

tmshort:
Rather than disabling the go-verdiff CI to allow it to pass when there is a legitimate golang version change, use a label to override the test results.

The label is `(override-go-verdiff)`.

# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/hack/tools/check-go-version.sh
+++ b/hack/tools/check-go-version.sh
@@ -1,7 +1,39 @@
 #!/bin/bash
 
-BASE_REF=${1:-main}
-GO_VER=$(sed -En 's/^go (.*)$/\1/p' "go.mod")
+U_FLAG='false'
+B_FLAG=''
+
+usage() {
+    cat <<EOF
+Usage:
+  $0 [-b <git-ref>] [-h] [-u]
+
+Reports on golang mod file version updates, returns an error when a go.mod
+file exceeds the root go.mod file (used as a threshold).
+
+Options:
+  -b <git-ref>  git reference (branch or SHA) to use as a baseline.
+                Defaults to 'main'.
+  -h            Help (this text).
+  -u            Error on any update, even below the threshold.
+EOF
+}
+
+while getopts 'b:hu' f; do
+    case "${f}" in
+        b) B_FLAG="${OPTARG}" ;;
+        h) usage
+           exit 0 ;;
+        u) U_FLAG='true' ;;
+        *) echo "Unknown flag ${f}"
+           usage
+           exit 1 ;;
+    esac
+done
+
+BASE_REF=${B_FLAG:-main}
+ROOT_GO_MOD="./go.mod"
+GO_VER=$(sed -En 's/^go (.*)$/\1/p' "${ROOT_GO_MOD}")
 OLDIFS="${IFS}"
 IFS='.' MAX_VER=(${GO_VER})
 IFS="${OLDIFS}"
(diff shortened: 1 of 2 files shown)


grokspawn:
Why? 
We already made this unconditionally executed and required.  
Is there a reason we need to provide an override capability as well? 

Is the issue [here|https://github.com/operator-framework/operator-controller/pull/1963] that the golang version went up, not that it threatened to exceed our high-water mark?

tmshort:
> Why? We already made this unconditionally executed and required. Is there a reason we need to provide an override capability as well?
> 
> Is the issue [here](https://github.com/operator-framework/operator-controller/pull/1963) that the golang version went up, not that it threatened to exceed our high-water mark? (in other words, is the check enforcing the wrong thing?)

There _are_ times we want to be able to override this check. For example, we _can_ go up to golang 1.23.6 as of right now without any downstreaming challenges. But we can't do that if the test is required and always fails. That would mean making the test non-required, and I'm not sure we want to do that and allow something we _don't_ want to slip through.

But yes, there may be bingo tools and other updates that can be permitted, and we should allow that EXPLICIT override.

grokspawn:
>There are times we want to be able to override this check. For example, we can go up to golang 1.23.6 as of right now without any downstreaming challenges. But we can't do that if the test is required and always fails. That would mean making the test non-required, and I'm not sure we want to do that and allow something we don't want to slip through.

Aside from vendoring challenges, this project is concerned with coordinating the golang version (including bingo'd tools) to not grow.  That isn't really the right goal, IMO, since growth under a threshold should not be problematic.  It's when this project (and projects which include its packages) advances a version collectively.  

So maybe it makes more sense for this to obtain a threshold version, and allow anything which doesn't exceed it.  We could follow the pattern of actions/setup-go's `go-version-file` and still use the underlying concept that the project's root `go.mod` is the "do not exceed" threshold.

grokspawn:
I'd delete lines 63-78.  The failures we're seeing in the kustomize bump are where the golang version changed but still lies at/below the threshold.  
Does it matter if the version changed, if it's below the threshold?

grokspawn:
> We already use the root go.mod as the threshold, so we're good there, and yes we ought to be fine if nothing exceeds that.
I'd prefer to keep those lines, and instead put it under a flag option; I'd rather not discard that code.

It's not discarded if it was previously committed.  
But my point is really that we _need_ the check to keep us from advancing beyond a specific version.  If we can use the go.mod in the root here as the sentinel version, fine.  
But why would we want to alert if any go version changes in future commits? 
IMO, this should be something which is enabled with a flag and not in the common path, if we want to keep it.

tmshort:
That's what I meant.
Please review the updated code.
_However_, we need to check if the threshold version has changed, so that's kept as an error, otherwise, it could mask a problem!

hack/tools/check-go-version.sh
```diff
@@ -1,10 +1,42 @@
...
+           exit 1 ;;
+    esac
+done
+
+BASE_REF=${B_FLAG:-main}
 GO_VER=$(sed -En 's/^go (.*)$/\1/p' "go.mod")
 OLDIFS="${IFS}"
 IFS='.' MAX_VER=(${GO_VER})
 IFS="${OLDIFS}"
+ROOT_GO_MOD="./go.mod"
```
grokspawn, line 39: nit:  define GO_VER after ROOT_GO_MOD, and change it to 
```
GO_VER=$(sed -En 's/^go (.*)$/\1/p' ${ROOT_GO_MOD})
```

grokspawn (dismissed):

grokspawn (approved):

PR #1968, opened 2025-05-14
🌱 Refactor rukpack package for configurable registry+v1 rendering behavior

perdasilva:
# Description

This PR refactors the rukpack/convert package to make it easier to configure registry+v1 conversion features like webhook and apiservice support by:

1. creating a canonical registry+v1 renderer that uses the right set of validators and generators into its own package
2. removing the PlainConverter from rukpack/convert (the registryv1.Renderer takes its place)
3. refactoring the RegistryV1ToHelmChart function into a fully fledged BundleToHelmChartConverter and using that in the Helm applier
4. in main configuring the BundleToHelmChartConverter used by the Helm Applier appropriately based on the feature flags

I decided not to add a controller-manager flag for toggling between cert providers. For now, let's rely on the FeatureGate values to do that. Once we have coalesced on default behavior we can start to think about this, I think. No need to add the complexity right now. 

## Reviewer Checklist

- [ ] API Go Documentation


internal/operator-controller/rukpak/render/registryv1/generators/resources_test.go
```diff
@@ -12,7 +12,7 @@ import (
 	rbacv1 "k8s.io/api/rbac/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
-	"github.com/operator-framework/operator-controller/internal/operator-controller/rukpak/render/generators"
+	"github.com/operator-framework/operator-controller/internal/operator-controller/rukpak/render/registryv1/generators"
```
camilamacedo86, line 15: 👍 I think it is nice to allow more types

cmd/operator-controller/main.go
```diff
@@ -421,12 +424,25 @@ func run() error {
 		preAuth = authorization.NewRBACPreAuthorizer(mgr.GetClient())
 	}
 
+	// determine if a certificate provider should be set in the bundle renderer and feature support for the provider
+	// based on the feature flag
+	var certProvider render.CertificateProvider
+	var isWebhookSupportEnabled bool
+	if features.OperatorControllerFeatureGate.Enabled(features.WebhookProviderCertManager) {
+		certProvider = certproviders.CertManagerCertificateProvider{}
+		isWebhookSupportEnabled = true
```
camilamacedo86, line 433: 👍

internal/operator-controller/applier/helm.go
```diff
@@ -53,13 +54,15 @@ type Preflight interface {
 	Upgrade(context.Context, *release.Release) error
 }
 
-type BundleToHelmChartFn func(rv1 fs.FS, installNamespace string, watchNamespace string) (*chart.Chart, error)
+type BundleToHelmChartConverter interface {
+	ToHelmChart(bundle source.BundleSource, installNamespace string, watchNamespace string) (*chart.Chart, error)
```
camilamacedo86, line 58: OK, so we are moving away to have a function to define a contract/interface to convert any BundleSource to a HelmChart. 

For HelmChart, we’d skip the convert call anyway, right?
But for something like a registry/v2 source, having the interface could help.
Is that what you had in mind?

internal/operator-controller/rukpak/convert/testdata/combine-properties-bundle/manifests/csv.yaml
```diff
@@ -1,10 +0,0 @@
-apiVersion: operators.operatorframework.io/v1alpha1
```
camilamacedo86, line 1: Why did we remove the testdata samples?
Are we mocking in the tests directly?

internal/operator-controller/rukpak/bundle/source/source_test.go
```diff
@@ -0,0 +1,111 @@
...
+    - type: AllNamespaces
+      supported: true
+`
+
+	return fstest.MapFS{
+		bundlePathAnnotations: &fstest.MapFile{Data: []byte(strings.Trim(annotationsYml, "\n"))},
+		bundlePathProperties:  &fstest.MapFile{Data: []byte(strings.Trim(propertiesYml, "\n"))},
+		bundlePathCSV:         &fstest.MapFile{Data: []byte(strings.Trim(csvYml, "\n"))},
+	}
+}
```
camilamacedo86, line 100: Here, I found the testdata removed 👍

internal/operator-controller/rukpak/convert/testdata/combine-properties-bundle/manifests/csv.yaml
camilamacedo86: I found it 👍

camilamacedo86 (approved):
Hi @perdasilva 👋

Thanks for the PR!
The bulk of the changes seem to be moving things around to better support other bundle source types — that makes sense 👍

I did notice the addition of the flag for CertManager provider and how it's combined. Would it maybe make sense to split that into a separate PR? That way we could give it more focused attention, and if anything breaks downstream, it’s easier to revert.
(That said, I’m not blocking this PR because of it.)

Overall, I don’t have any objections — nothing here looks like a blocker from our side, so I’m OK with moving forward.

/lgtm
/approve

Issue #962, opened 2024-06-21
[epic] Add support for handling helm charts


joelanford:
In OLMv1, we plan to support the `registry+v1` bundle format in order to provide a means for migration from OLMv0. However, we recognize the many pain points and limitations of that format (to name a few: limited types of supported objects, lack of templating, too opinionated out how RBAC is generated)

In order to provide operator authors and cluster admins with more flexibility and control, this epic tracks the work necessary to support managing native helm charts in OLMv1.

Some items to consider:
- Not all helm charts are conductive to being used in a declarative, always reconciling controller. Charts the use template functions that cause non-deterministic output will cause a reconciler to continuously detect and rollout changes.
- Helm charts that use hooks may be problematic in a reconciler because they introduce phases that require waiting. If we need/want to support helm hooks, we may need to figure out how to make the helm install/upgrade/uninstall processes asynchronous.

bentito:
Have you thought about which libraries might be good choices for helping with the validating of Helm charts to ensure we can work with them? We need to rule out chart with non-deterministic template functions and hooks, so something like:

```go
customChecks := []*config.Check{
		{
			Name:        "detect-non-deterministic-functions",
			Description: "Detect non-deterministic template functions in Helm charts",
			Template:    "template-name",
			Params: map[string]interface{}{
				"functions": []string{"randAlphaNum", "randAlpha", "randNumeric", "randAlphaSpecial", "now"},
			},
		},
		{
			Name:        "detect-helm-hooks",
			Description: "Detect Helm hooks in Helm charts",
			Template:    "template-name",
			Params: map[string]interface{}{
				"hooks": []string{"pre-install", "post-install", "pre-delete", "post-delete", "pre-upgrade", "post-upgrade", "pre-rollback", "post-rollback"},
			},
		},
```
in Stackrox's [kube-linter](golang.stackrox.io/kube-linter) terms, though I can't quite make a working example yet. Maybe there are other libs that would work better?

joelanford:
Checks for non-deterministic functions would probably require use of the https://pkg.go.dev/text/template/parse package, since those functions disappear post-templating.

Checks for helm hooks should be straightforward because we can render a manifest, parse as normal kube objects, then search for the annotations. If we end up with a handful of checks like that, I could definitely see the value in using an off-the-shelf library.

LalatenduMohanty:
This epic need more discussion to finalize what we want to deliver as part of this. Also what helm chart support means in OLM 1.0.0+

[...]

PR #1971, opened 2025-05-15
:sparkles: Add support for deploying OCI helm charts in OLM v1

OchiengEd:
* added support for deploying OCI helm charts which sits behind the HelmChartSupport feature gate
* extended the Cache interface to allow storing of Helm charts

# Description

This pull request aims to add logic to OLM v1 for handling OCI Helm chart support. We expect more work to go into this feature as further discussion on this occurs on issue https://github.com/operator-framework/operator-controller/issues/962 and the [Arbitrary Configuration RFC](https://docs.google.com/document/d/1IW733_9vroFFti08KMLOpwSKLUSlYtJbpbBkPHvwWsc/edit?tab=t.0#heading=h.x3tfh25grvnv) which may inform how `values.yml` would be passed to Helm charts.

## Reviewer Checklist

- [ ] API Go Documentation


internal/operator-controller/applier/helm.go
```diff
@@ -209,7 +213,119 @@ func (h *Helm) buildHelmChart(bundleFS fs.FS, ext *ocv1.ClusterExtension) (*char
 	if err != nil {
 		return nil, err
 	}
-	return h.BundleToHelmChartConverter.ToHelmChart(source.FromFS(bundleFS), ext.Spec.Namespace, watchNamespace)
+	var result *chart.Chart
+	contentType := bundleContentType(bundleFS)
+	switch contentType {
+	case HelmContent:
+		if !features.OperatorControllerFeatureGate.Enabled(features.HelmChartSupport) {
+			return nil, fmt.Errorf("helm chart support is not enabled")
```
trgeiger, line 221: Since this will be an alpha feature gate, we probably want to word this differently--this current sentence makes it sound like helm support could be enabled as a supported feature. Changing it to something like "installing helm charts is not supported" or "cannot install helm chart content type" would still communicate that we properly detected the content as a helm chart without implying that the user can/should toggle support for it.

kevinrizza: Should the content negotiation code also be wrapped by the feature gate instead of just the execution of the implementation?

camilamacedo86: That is a great point. We might want to wrap the whole code under the feature gate so the `switch contentType {` is just valid if the feature flag is used.

internal/operator-controller/applier/helm.go
```diff
@@ -209,7 +213,119 @@ func (h *Helm) buildHelmChart(bundleFS fs.FS, ext *ocv1.ClusterExtension) (*char
...
+func bundleContentType(bundleFS fs.FS) string {
+	var contentType string
+	_ = fs.WalkDir(bundleFS, ".", func(path string, f fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+
+		if !f.IsDir() {
+			// Check if Helm chart
+			if filepath.Dir(path) == "charts" &&
```
camilamacedo86, line 254: The charts as far I understand is optional.
what is mandatory for any chart is 

Chart.yaml and dir templates/

So I think we want something like

```
func IsBundleSourceHelm(path string) bool {
	chartFile := filepath.Join(path, "Chart.yaml")
	templatesDir := filepath.Join(path, "templates")

	// Check if Chart.yaml file exists
	if _, err := os.Stat(chartFile); os.IsNotExist(err) {
		return false
	}

	// Check if templates directory exists and is a directory
	if stat, err := os.Stat(templatesDir); os.IsNotExist(err) || !stat.IsDir() {
		return false
	}

	return true
}
```

internal/operator-controller/applier/helm.go
```diff
@@ -209,7 +213,119 @@ func (h *Helm) buildHelmChart(bundleFS fs.FS, ext *ocv1.ClusterExtension) (*char
...
+				filepath.Ext(f.Name()) == ".tgz" {
+				contentType = HelmContent
+				return fs.SkipAll
+			}
+
+			// Check if registry/v1 bundle
+			if filepath.Dir(path) == "metadata" &&
+				f.Name() == "annotations.yaml" {
+				contentType = BundleContent
+				return fs.SkipAll
```
camilamacedo86, line 264:
```suggestion
```



Also, I do not think that we need to check if it is registry/v1
We can do something like a more simplified version, see the comment: https://github.com/operator-framework/operator-controller/pull/1971/files#r2091973534

internal/operator-controller/applier/helm.go
```diff
@@ -209,7 +213,119 @@ func (h *Helm) buildHelmChart(bundleFS fs.FS, ext *ocv1.ClusterExtension) (*char
...
+	}
+
+	for _, f := range options {
+		f(chrt)
+	}
+
+	return chrt, nil
+}
+
+func loadPulledHelmChart(bundleFS fs.FS) (*chart.Chart, error) {
```
camilamacedo86, line 303: Why do we need to do this? 
What was your motivation?

internal/operator-controller/applier/helm.go
OchiengEd: The current implementation does not entail unpacking the contents of the Helm chart to the cache filesystem. However, if this is a hard requirement, then we can rewrite the implementation. For this reason, we have not taken the approach you have recommended above.

internal/operator-controller/applier/helm.go
camilamacedo86: I think I got, I saw above loadChartWithEnrichments you can ignore this one thx.
Closing to facilitate the review for others.

internal/operator-controller/applier/helm.go
OchiengEd: Indeed there is room for simplifying the solution. 

With that being said, what is happening in the code today is, if we were pulling an imaginary Helm chart `registry.io/water-melon:1.0.0` from an OCI registry. It would stored in the bundleFS filesystem as `charts/water-melon-1.0.0.tgz`.

Which I can see was a poor naming choice. Going forward,  we have two options:

1) Change the name of the directory from `charts` to something else. e.g. `artifacts`
2) Alternatively, we can unpack the contents of the Helm chart archive to the filesystem to be able to inspect its contents.

internal/shared/util/image/pull.go
```diff
@@ -224,6 +224,11 @@ func (p *ContainersImagePuller) applyImage(ctx context.Context, ownerID string,
 		}
 	}()
 
+	if hasChart(img) {
+		return pullChart(ctx, ownerID, imgSrc, canonicalRef, cache, layoutDir)
+	}
+
+	// Helm charts would error when getting OCI config
```
joelanford, line 231: This tells me that our cache `Store` interface method is too specific. We need to make that generic enough to accommodate registry+v1 bundles and OCI helm charts as a first step, and rebase the feature-gated helm support on top.

internal/operator-controller/applier/helm.go
OchiengEd: The code has been wrapped by the feature gate per the recommendations.

internal/operator-controller/applier/helm.go
OchiengEd: Done.

internal/operator-controller/applier/helm.go
```diff
@@ -209,9 +213,83 @@ func (h *Helm) buildHelmChart(bundleFS fs.FS, ext *ocv1.ClusterExtension) (*char
...
+	ch, err := readChartFS(bundleFS, filename)
+	if err != nil {
+		return false
+	}
+	*chart = *ch
+
+	return chart.Metadata != nil &&
+		chart.Metadata.Name != "" &&
+		chart.Metadata.Version != "" &&
+		len(chart.Templates) > 0
```
camilamacedo86, line 256: If a directory contains a Chart.yaml and templates, it qualifies as a Helm chart, right?
I mean, even if the name or version fields are missing from Chart.yaml, it’s still technically a chart — just not a valid one according to Helm’s requirements.

So maybe we should distinguish between two concepts:

- IsHelmChart: checks for the presence of Chart.yaml and templates (basic structure).
- IsValidHelmChart: additionally ensures required fields like name and version are set. ( as any other that we might to consider in the future )

Does that distinction make sense?

camilamacedo86 (commented):
OH. Great work 🥇

internal/shared/util/image/helm.go
```diff
@@ -0,0 +1,90 @@
...
+	var chartDataLayerDigest digest.Digest
+	if len(chartManifest.Layers) == 1 &&
+		(chartManifest.Layers[0].MediaType == registry.ChartLayerMediaType) {
+		chartDataLayerDigest = chartManifest.Layers[0].Digest
+	}
+
+	filename := fmt.Sprintf("%s-%s.tgz",
+		chartManifest.Annotations["org.opencontainers.image.title"],
+		chartManifest.Annotations["org.opencontainers.image.version"],
+	)
```
camilamacedo86, line 77: I think we need:
- a) **IsValidHelmChart():** (ForOLM) it ensures that required fields like name and version are set. ( as any other that we might consider in the future )
- b) The name and version must come from Chart.yaml ( where authors are expected to set ). Those annotations are optional and not necessary for the author of the content set. Furthermore, they are less intuitive than a Chart—ymal spec for them.

internal/operator-controller/applier/helm.go
```diff
@@ -209,9 +213,83 @@ func (h *Helm) buildHelmChart(bundleFS fs.FS, ext *ocv1.ClusterExtension) (*char
 	if err != nil {
 		return nil, err
 	}
+	if features.OperatorControllerFeatureGate.Enabled(features.HelmChartSupport) {
+		chart := new(chart.Chart)
+		if IsBundleSourceHelmChart(bundleFS, chart) {
+			return enrichChart(chart, WithInstallNamespace(ext.Spec.Namespace))
+		}
+	}
```
camilamacedo86, line 221: Could we add the unit tests for this code? WDYT?

OchiengEd: Sure. Working on tests right now.

internal/shared/util/image/helm.go
```diff
@@ -0,0 +1,90 @@
...
+
+	chartManifest := specsv1.Manifest{}
+	if err := json.Unmarshal(raw, &chartManifest); err != nil {
+		return nil, time.Time{}, fmt.Errorf("unmarshaling chart manifest; %w", err)
+	}
+
+	var chartDataLayerDigest digest.Digest
+	if len(chartManifest.Layers) == 1 &&
+		(chartManifest.Layers[0].MediaType == registry.ChartLayerMediaType) {
+		chartDataLayerDigest = chartManifest.Layers[0].Digest
```
camilamacedo86, line 71: Would we not need to do further verifications like:

```go
var chartDataLayerDigest digest.Digest

if len(chartManifest.Layers) == 0 {
	return nil, time.Time{}, fmt.Errorf("manifest has no layers; expected one chart layer")
}

if len(chartManifest.Layers) != 1 {
	return nil, time.Time{}, fmt.Errorf("expected exactly one chart layer, got %d", len(chartManifest.Layers))
}

layer := chartManifest.Layers[0]
if layer.MediaType != registry.ChartLayerMediaType {
	return nil, time.Time{}, fmt.Errorf("unexpected media type %q for chart layer", layer.MediaType)
}

chartDataLayerDigest = layer.Digest
```

internal/shared/util/image/helm.go
```diff
@@ -0,0 +1,90 @@
...
+	}
```
camilamacedo86, line 72: Would we not need to do further verifications? 
Also, I think we can have more than 1 layer.

So maybe something like:

```go
var chartDataLayerDigest digest.Digest
found := false

if len(chartManifest.Layers) == 0 {
	return nil, time.Time{}, fmt.Errorf("manifest has no layers; expected at least one chart layer")
}

for i, layer := range chartManifest.Layers {
	if layer.MediaType == registry.ChartLayerMediaType {
		chartDataLayerDigest = layer.Digest
		found = true
		break
	}
}

if !found {
	return nil, time.Time{}, fmt.Errorf(
		"no layer with media type %q found in manifest (total layers: %d)",
		registry.ChartLayerMediaType,
		len(chartManifest.Layers),
	)
}
```

internal/shared/util/image/cache.go
```diff
@@ -29,6 +29,7 @@ type LayerData struct {
 }
 
 type Cache interface {
+	ExtendCache
```
camilamacedo86, line 32: Is that valid for any or only Helm
Would be better we rename it like ChartCache, something like?

internal/shared/util/image/helm.go
```diff
@@ -0,0 +1,90 @@
...
+	if err != nil {
+		return nil, time.Time{}, fmt.Errorf("error getting mod time of unpack directory: %w", err)
+	}
+	return os.DirFS(filepath.Dir(dest)), modTime, nil
+}
+
+func pullChart(ctx context.Context, ownerID string, img types.ImageSource, canonicalRef reference.Canonical, cache Cache, imgRef types.ImageReference) (fs.FS, time.Time, error) {
+	raw, _, err := img.GetManifest(ctx, nil)
+	if err != nil {
+		return nil, time.Time{}, fmt.Errorf("get OCI helm chart manifest; %w", err)
```
camilamacedo86, line 60:
```suggestion
		return nil, time.Time{}, fmt.Errorf("copying chart to %s: %w", filename, err)

```

camilamacedo86, line 60: Why are we returning the Time?
Do we return it for registry/v1 as well?

internal/operator-controller/applier/helm.go
OchiengEd: What you are saying certainly make sense. However, what I am working around is the chart would be stored on a filesystem  of type `fs.FS`. Therefore, `loader.LoadDir()` is unable to find a Helm chart. The only option to proceed with this approach is to implement the [ChartLoader interface](https://pkg.go.dev/helm.sh/helm/v3/pkg/chart/loader#ChartLoader) and all other tools that error when working with files in an `fs.FS` filesystem.

Since the contents of the Chart.yaml would be loaded into a `&chart.Manifest{}` object. I think it should suffice to check that the chart.Metadata field is not nil and we have templates.

internal/shared/util/image/helm.go
OchiengEd: The return values are dictated by the `Store()`  and `StoreChart()` functions in the [Cache interface](https://github.com/operator-framework/operator-controller/blob/ea05e44a204ae268b13fc59d80d8576c6866af81/internal/shared/util/image/cache.go#L31-L36). 

In addition, the `pullChart()` method is called by the  [applyImage() function](https://github.com/operator-framework/operator-controller/blob/ea05e44a204ae268b13fc59d80d8576c6866af81/internal/shared/util/image/pull.go#L209) when the Helm chart MediaTypes are detected.

internal/operator-controller/applier/helm.go
camilamacedo86: OK, we can keep doing both checks. But would it make sense to define both functions explicitly and call them, rather than assuming something is a HelmChart just because it has a name and version set? I mean, can we just split in 2 funcs and call them out instead of have both logics inside of IsHelmChart?

Sorry if I’m being a bit nitpicky — I just think this approach could give us more flexibility to grow and evolve the logic in the future.

internal/shared/util/image/helm.go
OchiengEd: @camilamacedo86 The lines of code you are referring to are dealing with the OCI image that represents the Helm Chart. Therefore, we can not readily review the `Chart.yaml`  contents. However, the helm library generates [OCI annotations](https://github.com/helm/helm/blob/334045d002024a36cec36624b3302af87de950b3/pkg/registry/util.go#L150-L196) using the values in the `Chart.yaml`; specifically, the `Name` and `Version` are used to populate `org.opencontainers.image.title` and `org.opencontainers.image.version` respectively.

This can be seen in action in the registry client Push() method which starts at [line 630](https://github.com/helm/helm/blob/1b21f04b3b6960fd48bfee99ae2cfb37c0be708a/pkg/registry/client.go#L630) but more specifically the OCI annotations are added in between [lines 687 and 693](https://github.com/helm/helm/blob/1b21f04b3b6960fd48bfee99ae2cfb37c0be708a/pkg/registry/client.go#L687-L693).

internal/shared/util/image/cache.go
OchiengEd: Only for Helm charts at this time. We would need to  see if it is possible to refactor the `Store()` method in the Cache interface to be able to store both Helm charts and OCI images to the cache.

internal/shared/util/image/pull.go
OchiengEd: Making the `Cache` interface generic enough to be able to store both Helm charts and OCI images is inevitable. Will be looking at that at some point

internal/shared/util/image/helm.go
OchiengEd: Good catch

internal/shared/util/image/pull.go
OchiengEd: This is should be resolved. As can be seen in the code block below specifically on line 72, we are no longer using the StoreChart() method.

  https://github.com/operator-framework/operator-controller/blob/059008d7032205974784e75f6b011655c1359dc4/internal/shared/util/image/helm.go#L35-L73

OchiengEd:
When pulling a Helm chart with a provenance file, at this time we have chosen to skip pulling the layer to the cache filesystem since we have no logic in place at this time to verify the chart integrity.

https://github.com/operator-framework/operator-controller/blob/059008d7032205974784e75f6b011655c1359dc4/internal/shared/util/image/helm.go#L62-L65

OchiengEd:
Also added logic to inspect the contents of a `.tgz` or `.tar.gz` file to verify it contains of a Helm archive.
 

https://github.com/operator-framework/operator-controller/blob/059008d7032205974784e75f6b011655c1359dc4/internal/shared/util/image/helm.go#L97-L135

internal/operator-controller/applier/helm.go
```diff
@@ -209,6 +211,17 @@ func (h *Helm) buildHelmChart(bundleFS fs.FS, ext *ocv1.ClusterExtension) (*char
...
+	if features.OperatorControllerFeatureGate.Enabled(features.HelmChartSupport) {
+		meta := new(chart.Metadata)
+		if ok, _ := imageutil.IsBundleSourceChart(bundleFS, meta); ok {
+			return imageutil.LoadChartFSWithOptions(
+				bundleFS,
+				fmt.Sprintf("%s-%s.tgz", meta.Name, meta.Version),
+				imageutil.WithInstallNamespace(ext.Spec.Namespace),
+			)
+		}
+	}
```
camilamacedo86, line 223: If we are using OCI we would not have  tgz right.
So could we not improve this code to accept both formats
Example

```
helm chart save ./mychart oci://registry.example.com/charts
helm chart push oci://registry.example.com/charts/mychart:1.2.3
```

internal/shared/util/image/cache_test.go
```diff
@@ -211,6 +213,22 @@ func TestDiskCacheStore(t *testing.T) {
...
+			layers: func() iter.Seq[LayerData] {
+				sampleChart := filepath.Join("../../../../", "testdata", "charts", "sample-chart-0.1.0.tgz")
+				data, _ := os.ReadFile(sampleChart)
+				return func(yield func(LayerData) bool) {
+					yield(LayerData{Reader: bytes.NewBuffer(data), MediaType: registry.ChartLayerMediaType})
+				}
+			}(),
+			expect: func(t *testing.T, cache *diskCache, fsys fs.FS, modTime time.Time, err error) {
+				require.NoError(t, err)
+			},
```
camilamacedo86: Hey OchiengEd,

Thanks for the great work pushing this forward. I have a few suggestions to help make this more maintainable and easier to extend.

- 1  - We currently have 3 mocks under `testdata/charts` but only one test case using them. It seems like we might be missing additional test scenarios (e.g. invalid charts, missing templates, etc.). Could we ensure that we cover all cases?

- 2 - Also, I think it would be better to mock the chart archive in memory instead of adding static `.tgz` files under `testdata`. This keeps the tests self-contained and avoids the need to manage test artifacts in the repo. It would also make it easier to test different edge cases. Here's an example of how we could do that:

```go
func mockHelmChartTgz(t *testing.T) []byte {
	var buf bytes.Buffer
	tw := tar.NewWriter(&buf)

	// Add Chart.yaml
	chartYaml := []byte("apiVersion: v2\nname: testchart\nversion: 0.1.0")
	require.NoError(t, tw.WriteHeader(&tar.Header{
		Name: "testchart/Chart.yaml",
		Mode: 0600,
		Size: int64(len(chartYaml)),
	}))
	_, _ = tw.Write(chartYaml)

	// Add templates/deployment.yaml
	tmpl := []byte("kind: Deployment\napiVersion: apps/v1")
	require.NoError(t, tw.WriteHeader(&tar.Header{
		Name: "testchart/templates/deployment.yaml",
		Mode: 0600,
		Size: int64(len(tmpl)),
	}))
	_, _ = tw.Write(tmpl)

	require.NoError(t, tw.Close())

	var gzBuf bytes.Buffer
	gz := gzip.NewWriter(&gzBuf)
	_, err := gz.Write(buf.Bytes())
	require.NoError(t, err)
	require.NoError(t, gz.Close())

	return gzBuf.Bytes()
}
```

Also, see how we mock things for tests in the operator-framework/api that might be helpful. You will see that there, we start to add mocks under testdata dirs, but then it becomes hard to maintain. 

3. Last, I think the Helm-specific test is currently inserted in the middle of `diskCache` test cases, which **mixes different concerns.** Ideally, we should ensure a good SoC(https://en.wikipedia.org/wiki/Separation_of_concerns). Could we to move Helm-specific cases into their own test set (e.g. `TestDiskCacheStore_HelmChart`) for better readability and structure?

Let me know what you think.

internal/shared/util/image/helm.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+
+	return nil
+}
+
+func IsBundleSourceChart(bundleFS fs.FS, metadata *chart.Metadata) (bool, error) {
+	var chartPath string
+	files, _ := fs.ReadDir(bundleFS, ".")
+	for _, file := range files {
+		if slices.Contains([]string{".tar.gz", ".tgz"}, filepath.Ext(file.Name())) {
```
camilamacedo86:
```suggestion
		if strings.HasSuffix(file.Name(), ".tgz") || strings.HasSuffix(file.Name(), ".tar.gz")
```

I think it should be like above right?

internal/shared/util/image/helm.go
```diff
@@ -0,0 +1,222 @@
...
+	templatesExist bool
+	// chartfileExists is set to true if the Chart.yaml
+	// file exists in the chart archive
+	chartfileExists bool
+}
+
+func inspectChart(data []byte, metadata *chart.Metadata) (chartInspectionResult, error) {
+	gzReader, err := gzip.NewReader(bytes.NewReader(data))
+	if err != nil {
+		return chartInspectionResult{}, err
```
camilamacedo86: Can we wrap the errors to make it easier to debug? like 

return chartInspectionResult{}, fmt.Errorf("creating gzip reader: %w", err)

internal/shared/util/image/helm.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+
+	chartData, err := fs.ReadFile(bundleFS, chartPath)
+	if err != nil {
+		return false, err
+	}
+
+	result, err := inspectChart(chartData, metadata)
+	if err != nil {
+		return false, err
```
camilamacedo86: same in the above cases could we wrapper the error like

```
return false, fmt.Errorf("failed to read %s from fs: %w", chartPath, err)
```

internal/shared/util/image/helm.go
```diff
@@ -0,0 +1,222 @@
...
+	if len(chartManifest.Layers) == 0 {
+		return nil, time.Time{}, fmt.Errorf("manifest has no layers; expected at least one chart layer")
+	}
+
+	layerIter := iter.Seq[LayerData](func(yield func(LayerData) bool) {
+		for i, layer := range chartManifest.Layers {
+			ld := LayerData{Index: i, MediaType: layer.MediaType}
+			if layer.MediaType == registry.ChartLayerMediaType {
+				var contents []byte
+				contents, ld.Err = os.ReadFile(filepath.Join(
```
camilamacedo86: Could we avoid to use is.ReadFile in favor of something like 

```
rc, _, err := imgSrc.GetBlob(ctx, layer, none.NoCache)
if err != nil {
	ld.Err = err
} else {
	ld.Reader = rc
}
```

internal/operator-controller/applier/helm.go
camilamacedo86: Would that make sense?
If so, and we can change now that would be great. WDYT?

OchiengEd: You can not push a helm chart directory to an OCI registry. It has to be packaged into a chart archive(.tgz)

See example below:
```
$ helm push ./random-chart oci://registry.local/charts         
Error: cannot push directory, must provide chart archive (.tgz)
$
```

The chart archive actually forms a layer in the Helm image and if present the provenance file would make a different layer.

internal/shared/util/image/helm.go
OchiengEd: Both solutions accomplish the same thing. I will switch over to use your statements for simplicity

internal/operator-controller/applier/helm.go
camilamacedo86: I see good point. Thank you for the clarification and sharing 👍 
Marking as solved

internal/shared/util/image/helm.go
camilamacedo86: I do not think that they are the same.

The first approach:
- Uses filepath.Ext(...) to get the file extension (after the last dot).
- Then checks if it's in the list [".tar.gz", ".tgz"].

Then, it seems that we have a problem with filepath.Ext("chart-0.1.0.tar.gz") returns .gz, not .tar.gz.
So this check won’t work for .tar.gz files — it will only match .tgz.

OchiengEd: Thanks  for the clarification. I overlooked the .tar.gz scenario. Good catch

internal/shared/util/image/cache_test.go
OchiengEd: 1. The mocks are in use in the `TestIsBundleSourceChart`  test. In addition, the test cases cover the scenarios you mention. Let me know if you had something different in mind.

```bash
operator-controller : helm_explorations ✘ :✹ ᐅ  go test -v -run ^TestIsBundleSourceChart$ ./internal/shared/util/image/...
=== RUN   TestIsBundleSourceChart
=== RUN   TestIsBundleSourceChart/complete_helm_chart_with_nil_*chart.Metadata
=== RUN   TestIsBundleSourceChart/complete_helm_chart
=== RUN   TestIsBundleSourceChart/helm_chart_without_templates
=== RUN   TestIsBundleSourceChart/helm_chart_without_a_Chart.yaml
--- PASS: TestIsBundleSourceChart (0.00s)
    --- PASS: TestIsBundleSourceChart/complete_helm_chart_with_nil_*chart.Metadata (0.00s)
    --- PASS: TestIsBundleSourceChart/complete_helm_chart (0.00s)
    --- PASS: TestIsBundleSourceChart/helm_chart_without_templates (0.00s)
    --- PASS: TestIsBundleSourceChart/helm_chart_without_a_Chart.yaml (0.00s)
PASS
ok  	github.com/operator-framework/operator-controller/internal/shared/util/image	0.016s
operator-controller : helm_explorations ✘ :✹ ᐅ
```

2. I did explore this briefly and it was getting too lengthy. I can revisit if need be. You have only captured one scenario in the example above.
 
3. I understand the idea of separation of concerns. However, in this case the concern is the test cases all aim to ensure that the cache's Store() method is able to store the desired content (indiscriminate of whether it is an OCI image or Helm image) to the cache. My  concern with splitting the test is if making a change intending to only affect OCI images introduces an issue  in the logic for storing  Helm charts or vice versa.

camilamacedo86: But that is an alpha feature.
Then, imagine if we need to remove all alpha features tomorrow, how would we do that? 
So, in POV I still have the feeling that this one should be like it would be better if it were in a new chunk. WDYT?

internal/shared/util/image/helm.go
OchiengEd: Thanks for the explanation. I clearly overlooked that.

internal/shared/util/image/cache_test.go
OchiengEd: Alright. I will create a separate `TestDiskCacheStore_HelmChart` block.

internal/shared/util/image/pull.go
OchiengEd: @camilamacedo86 all concerns raised should be resolved.

internal/shared/util/image/pull.go
```diff
@@ -224,14 +224,19 @@ func (p *ContainersImagePuller) applyImage(ctx context.Context, ownerID string,
 		}
 	}()
 
+	if hasChart(img) {
+		return pullChart(ctx, ownerID, srcRef, canonicalRef, imgSrc, cache)
+	}
```
camilamacedo86, line 232: Hi @OchiengEd 

I missed this one.
It also should be protected by a feature flag.
Any code related to the helmchart should not be outside othe featurere flag and should not be executed when we do not enable the flag, Can you please add the feature flag

internal/shared/util/image/cache_test.go
```diff
@@ -585,6 +648,7 @@ func TestDiskCacheGarbageCollection(t *testing.T) {
 	}
 }
 
+//nolint:unparam
```
camilamacedo86: Could we not solve the lint issue here instead of ignore the check?

internal/shared/util/image/pull.go
```diff
@@ -224,14 +225,20 @@ func (p *ContainersImagePuller) applyImage(ctx context.Context, ownerID string,
...
+
 	ociImg, err := img.OCIConfig(ctx)
 	if err != nil {
 		return nil, time.Time{}, err
 	}
 
 	layerIter := iter.Seq[LayerData](func(yield func(LayerData) bool) {
 		for i, layerInfo := range img.LayerInfos() {
-			ld := LayerData{Index: i}
+			ld := LayerData{Index: i, MediaType: layerInfo.MediaType}
```
camilamacedo86, line 241: It is added with the helm chart, but I think it is OK to exist on the default code as well.
@tmshort @thetechnick WDYT?

internal/shared/util/image/pull.go
camilamacedo86: Thank you to solve that marking as close

internal/shared/util/image/cache_test.go
camilamacedo86: @OchiengEd ^ could we fix it so then I can LGTM this one.
But after that would also be good get an approval from @joelanford or @thetechnick in this one probably

internal/shared/util/image/cache.go
```diff
@@ -128,8 +131,15 @@ func (a *diskCache) Store(ctx context.Context, ownerID string, srcRef reference.
...
-				return fmt.Errorf("error applying layer[%d]: %w", layer.Index, err)
+			switch layer.MediaType {
+			case registry.ChartLayerMediaType:
+				if err := storeChartLayer(dest, layer); err != nil {
+					return err
+				}
+			default:
+				if _, err := archive.Apply(ctx, dest, layer.Reader, applyOpts...); err != nil {
+					return fmt.Errorf("error applying layer[%d]: %w", layer.Index, err)
+				}
```
camilamacedo86: Should it not be protected behind the feature flag as well?

internal/shared/util/image/cache_test.go
```diff
@@ -144,6 +146,67 @@ func TestDiskCacheFetch(t *testing.T) {
...
+		srcRef       reference.Named
+		canonicalRef reference.Canonical
+		imgConfig    ocispecv1.Image
+		layers       iter.Seq[LayerData]
+		filterFunc   func(context.Context, reference.Named, ocispecv1.Image) (archive.Filter, error)
+		setup        func(*testing.T, *diskCache)
+		expect       func(*testing.T, *diskCache, fs.FS, time.Time, error)
+	}{
+		{
+			name:         "returns no error if layer read contains helm chart",
```
camilamacedo86, line 168: Have we not any scenario where it should fail?
Could we add those as well?

internal/operator-controller/applier/helm.go
```diff
@@ -26,9 +26,11 @@ import (
 
 	ocv1 "github.com/operator-framework/operator-controller/api/v1"
 	"github.com/operator-framework/operator-controller/internal/operator-controller/authorization"
+	"github.com/operator-framework/operator-controller/internal/operator-controller/features"
 	"github.com/operator-framework/operator-controller/internal/operator-controller/rukpak/bundle/source"
 	"github.com/operator-framework/operator-controller/internal/operator-controller/rukpak/preflights/crdupgradesafety"
 	"github.com/operator-framework/operator-controller/internal/operator-controller/rukpak/util"
+	imageutil "github.com/operator-framework/operator-controller/internal/shared/util/image"
```
camilamacedo86, line 33: @OchiengEd 

Here: https://github.com/operator-framework/operator-controller/pull/1724/files

@perdasilva added a demo and the doc under the docs/draft for another alpha feature
It would be very nice if we could do your demo within and add the doc for that. Such as it was done in this PR. However, I am okay with it being a follow-up.

internal/shared/util/image/cache.go
OchiengEd: The code will be unreachable as a result of the `pullChart()` function being wrapped around a feature gate. However, if we need to wrap that code as well, feel tree to let me know

https://github.com/operator-framework/operator-controller/blob/9b366d34a83ac582ddfa11ef3e89fb9ccf4dd75f/internal/shared/util/image/pull.go#L228-L232

camilamacedo86: I believe that all code should be categorised under a flag, allowing us to easily identify the purpose of each specific code for each alpha/beta feature. However, I think we can wait for others' input to see what they think about. My concern is: 

What happens if we decide not to use the alpha/beta feature and want to delete it? If we decide not to promote feature A or B. How hard will it be if not all related code is not under the feature flag condition? But others might think that is OK

@perdasilva @tmshort WDYT?

internal/shared/util/image/cache_test.go
OchiengEd: The `TestDiskCacheStore_HelmChart` is an extensions of the `TestDiskCacheStore` tests. The errors are already covered in the `TestDiskCacheStore` tests. Therefore, I think it would  be unnecessary to cover all scenarios here.

internal/operator-controller/applier/helm.go
OchiengEd: I would be for adding documentation and a demo as a follow up.

internal/shared/util/image/cache_test.go
camilamacedo86: So, we have no other scenarios that are specific to Helm charts that could be added here?
If so it is fine 👍

internal/shared/util/image/cache_test.go
```diff
@@ -619,3 +800,11 @@ func fsTarReader(fsys fs.FS) io.ReadCloser {
 	}()
 	return pr
 }
+
+type dummyReader struct{}
+
+var _ io.Reader = &dummyReader{}
+
+func (r *dummyReader) Read(p []byte) (n int, err error) {
```
camilamacedo86: @OchiengEd we have a lint issue here to sort out.

internal/shared/util/image/cache_test.go
camilamacedo86: Done thx

internal/shared/util/image/cache_test.go
camilamacedo86:
```
/home/runner/go/bin/golangci-lint-v1.64.6 run --build-tags containers_image_openpgp --out-format colored-line-number
Error: internal/shared/util/image/cache_test.go:808:1: named return "n" with type "int" found (nonamedreturns)
func (r *dummyReader) Read(p []byte) (n int, err error) {
^
make: *** [Makefile:122: lint] Error 1
Error: Process completed with exit code 2.
```

internal/operator-controller/applier/helm.go
```diff
@@ -209,6 +211,17 @@ func (h *Helm) buildHelmChart(bundleFS fs.FS, ext *ocv1.ClusterExtension) (*char
 	if err != nil {
 		return nil, err
 	}
+	if features.OperatorControllerFeatureGate.Enabled(features.HelmChartSupport) {
```
perdasilva, line 214: We've been trying to put the feature checks in main. Wdyt about refactoring to have another implementation of the BundleToHelmChart converter that accepts chart bundles, or maybe that can route between different bundle types? Then if the feature gate is enabled, use that one?

internal/shared/util/image/cache.go
perdasilva: Similarly to my comment above about moving the feature flag checks to main, should we refactor this in a way that make the cache configurable as to the different layer media types and how they are handled?

perdasilva: actually - the comment is below =P

internal/operator-controller/applier/helm.go
OchiengEd: I do not have any concerns with having feature checks in main. I think that would be the ideal scenario as it would allow for feature-gated functionality to be tested at build time. 

Trying to look at the issue from afar, our challenge in this case is we are trying to run tests on a feature gated function, `pullChart()` while we are not allowing a portion of the code in the cache's `Store()` method not to run since the feature gate is not enabled. The big question therefore is how do we navigate this quandary?

internal/shared/util/image/cache.go
OchiengEd: Learning moment for me here. How have we removed feature gated code that we have decided not to land in the past? Is using the git history to undo code that is associated with a feature gate sufficient or acceptable?

camilamacedo86: I think it’d be helpful to keep all changes behind the feature gate clearly wrapped and centralized. That way, it’s easy to track what’s gated just by reading the code—since we’ll likely be iterating over multiple PRs, relying on Git history isn’t ideal.

Totally open to any approach that makes the boundary clear. The primary goal is to maintain cohesion and ensure we all share a common understanding of what lies behind the gate. 😊

OchiengEd: Understood. With that being said, from my perspective our options might be:

1. Execute tests with feature gates enabled 
2. Extend the cache to allow tests to run even when the feature gate is disabled
3. Don't wrap the Helm logic in the cache `Store()` method which will allow tests to run even with feature gate disabled

Open to hear any other ideas or if we might have a preference for one of these options over others.

OchiengEd: @camilamacedo86 I was able to get a way to wrap the code and allow tests to run without issue. Hopefully that could suffice for now as we deliberate if to land the feature or not.

https://github.com/operator-framework/operator-controller/blob/fb98756685b11535f407d128ab7276e6947035f6/internal/shared/util/image/cache.go#L137-L141

internal/shared/util/image/cache.go
```diff
@@ -128,8 +133,16 @@ func (a *diskCache) Store(ctx context.Context, ownerID string, srcRef reference.
...
+			if layer.MediaType == registry.ChartLayerMediaType {
+				if features.OperatorControllerFeatureGate.Enabled(features.HelmChartSupport) || testing.Testing() {
+					if err := storeChartLayer(dest, layer); err != nil {
+						return err
+					}
+				}
+			} else {
+				if _, err := archive.Apply(ctx, dest, layer.Reader, applyOpts...); err != nil {
+					return fmt.Errorf("error applying layer[%d]: %w", layer.Index, err)
+				}
```
joelanford: I may have mentioned this elsewhere. If I did, apologies for repeating myself...

I feel like a better way of handling this would be to register handlers for media types. Then this section of the code would lookup the media type of the layer and then call whatever handler is found (or error if there isn't one).

With that setup, the registration of the helm chart media type/handler would be feature gated, but this code would not.

internal/shared/util/image/cache.go
```diff
@@ -147,6 +160,33 @@ func (a *diskCache) Store(ctx context.Context, ownerID string, srcRef reference.
...
+	)
+	fmt.Println(filename)
+	chart, err := os.Create(filename)
+	if err != nil {
+		return fmt.Errorf("writing chart layer: %w", err)
+	}
+	defer chart.Close()
+
+	_, err = chart.Write(data)
+	return err
```
joelanford: Have you seen https://pkg.go.dev/helm.sh/helm/v3/pkg/chart/loader#LoadArchive? Seems like we should use that over our own custom implementation in `inspectChart`.

I'd suggest a setup with `renameio.PendingFile`, where we:
1. `io.Copy` the reader to the temp file
2. Rewind the temp file with `tempFile.Seek(0, io.SeekStart)`
3. Use `loader.LoadArchive` to make sure we can load a chart from the archive
4. Call https://pkg.go.dev/github.com/google/renameio/v2#PendingFile.CloseAtomicallyReplace to "commit" the file.

internal/shared/util/image/cache.go
```diff
@@ -147,6 +160,33 @@ func (a *diskCache) Store(ctx context.Context, ownerID string, srcRef reference.
...
+	}
+	meta := new(chart.Metadata)
+	_, err = inspectChart(data, meta)
+	if err != nil {
+		return fmt.Errorf("inspecting chart layer: %w", err)
+	}
+	filename := filepath.Join(path,
+		fmt.Sprintf("%s-%s.tgz", meta.Name, meta.Version),
+	)
+	fmt.Println(filename)
```
joelanford: Remove the print statement?

OchiengEd: absolutely. That was not supposed to be committed.

internal/shared/util/image/cache.go
OchiengEd: Code has been refactored. Hopefully I got your suggestion right.

internal/operator-controller/applier/helm.go
camilamacedo86: >I do not have any concerns with having feature checks in main. I think that would be the ideal scenario as it would allow for feature-gated functionality to be tested at build time.

We are working towards to it
See; https://github.com/operator-framework/operator-controller/pull/1980

We will use the experimental CRD and have all feature flags enabled for our e2e tests.

camilamacedo86 (commented):
Thank you a lot for your contribution 🥇 
I’m OK with the changes 👍
/lgtm

IHMO: It is an initial implementation, and as follow I would expected:

- Documentation (docs/draft) and a demo automated for this one
- Full end-to-end (e2e) test coverage
- A discussion around the remaining caveats, particularly how we plan to store the Helm chart bundle in the catalog via OPM

Would be great to get a PTAL from or at least one of them @joelanford, @thetechnick, and @perdasilva before get this one merged. 

Thank you 🙌

camilamacedo86:
Hi @OchiengEd,

Just to clarify:

In the F2F, we talked about using a single internal type and converting everything to it. We also discussed having only one runtime—Buxcutter. PoC: https://github.com/operator-framework/operator-controller/pull/1946

**Supporting Helm fully will take time, and it is a big journey as we have been discussing**. There are still open questions and blockers. If we go with this approach (internal format), we probably want OPM/OLM to support the internal type first before anything else. I created an internal [Slack](https://redhat.enterprise.slack.com/archives/C093EJ8EY13) channel so that we can discuss all details and caveats.

I’m OK merging this since **it’s behind a feature gate—safe to experiment and doesn’t break anything**. But it may change later.

_PS:  If we get it merged, it would be great to have a small doc with automated demo steps (as done for others feature gates), so people can understand what was done and iterate_

Leaving final call to @joelanford, @thetechnick, and @perdasilva. 🔨

tmshort:
@OchiengEd @itroyano 
Regarding feature-gated APIs, you don't change any of the API files, so that makes things easier.
You need to create a kustomize component to enable this feature gate for experimental manifests/runs/testing.
For additional information:
https://github.com/operator-framework/operator-controller/blob/main/config/README.md
For an example:
https://github.com/operator-framework/operator-controller/tree/main/config/components/features/webhook-provider-certmanager
Once this is done, then your feature will be available in the `make run-experimental` target, and you can add e2e tests in to the `test/experimental-e2e` directory, which will be run via `make test-experimental-e2e`
https://github.com/operator-framework/operator-controller/tree/main/test/experimental-e2e

OchiengEd:
@camilamacedo86 @tmshort The documentation and feature-gated APIs concerns should be resolved. Kindly review and advice if anything else is needed at this time.

docs/draft/howto/enable-helm-chart-support.md
```diff
@@ -0,0 +1,415 @@
...
+    $ kubectl patch deployment -n olmv1-system operator-controller-controller-manager --type='json' -p='[{"op": "add", "path": "/spec/template/spec/containers/0/args/-", "value": "--feature-gates=HelmChartSupport=true"}]'
+    ```
+
+2.  **Wait for the controller manager pods to be ready:**
+
+    ```bash
+    $ kubectl -n olmv1-system wait --for condition=ready pods -l control-plane=operator-controller-controller-manager
+    ```
+
+Once the above wait condition is met, the `HelmChartSupport` feature gate should be enabled in operator controller.
```
tmshort, line 30: There is a `run-experimental` make target that can be used to run this (of course it includes all other experimental features as well).
You might want to mention this alternative method.

tmshort:
The changes to the manifests look good!

docs/draft/howto/enable-helm-chart-support.md
OchiengEd: Good catch. Can we do a small revision after the PR is merged? I'm hesitant to add a change and kick out the one approval we have at this time.

joelanford: With the new `run-experimental`, I'm templated to say that our docs should _only_ talk about the experimental feature set, and not get into the details of enabling/disabling individual features.

internal/shared/util/image/cache.go
joelanford: It still isn't exactly what I have in mind. What I have in mind is that `ContainersImagePuller` would have a new field named `ImageHandlers` that would represent the kinds of images that can be pulled. So we'd have an `ImageHandler` for registry+v1 and one for helm charts, and we would register those appliers in `main.go` (where the helm image applier would be registered behind a feature gate).

The `ImageHandler` could be an interface something like this:
```go
type ImageHandler interface {
    Match(ctx context.Context, img types.ImageCloser) (bool, error) // likely just needs `img.Manifest()`
    Apply(ctx context.Context, ownerID string, srcRef reference.Named, canonicalRef reference.Canonical, imgSrc types.ImageSource, cache Cache, sourceContext *types.SystemContext) (fs.FS, time.Time, error)
```

With that, `ContainersImagePuller.applyImage` could iterate the handlers looking for the first matching handler, and then call _that_ handler's `Apply` method. And then the `ContainersImagePuller` implementation wouldn't have to concern itself with the details of every single image type we ever want to support.

internal/operator-controller/applier/helm.go
perdasilva: @OchiengEd I think in order to do it properly, we'll need some architectural changes to the applier. As it is now, it assumes it will be receiving a registry+v1 bundle FS. As the code is now, we are expanding the scope of the applier to also identify the bundle type and produce a chart accordingly. I think we should pivot and instead of the applier having a `BundleToHelmChartConverter` it should rather have something like a `HelmChartGetter` (or something mode sophisticated) that takes an FS + config and returns a helm chart object.

The `HelmChartGetter` could use a pattern similar to what Joe suggested for the cache layer, e.g. have some kind of interface that can identify the bundle type from the FS and apply and build the chart.

In the main function, we can then configure the HelmChartGetter to either include (or not) the the interface implementation that deals with Helm bundles.

Alternatively, we compose have two HelmChartGetters, one with the standard implementation for registry+v1 bundles and another that uses the standard implementation and extends it by basically the code you have here (e.g. first check whether its a helm chart fs, if not, then just use the standard implementation).

The same pattern could be used for the Cache. If the flag is set, configure it one way, if not, configure it another.

Breaking it down like this makes it easy to test independently and you don't even need to rely on flags being set or not for (unit) testing.

perdasilva: To keep momentum, I'll approve this for now and maybe we could work together on a refactor as a fast follow.

PR #1973, opened 2025-05-16
📖 Add NetworkPolicy doc

anik120:
# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/docs/draft/api-reference/network-policies.md
+++ b/docs/draft/api-reference/network-policies.md
@@ -1,0 +1,94 @@
+# NetworkPolicy in OLMv1
+
+## Overview
+
+OLMv1 uses [Kubernetes NetworkPolicy](https://kubernetes.io/docs/concepts/services-networking/network-policies/) to secure communication between components, restricting network traffic to only what's necessary for proper functionality. 
+
+* The catalogd NetworkPolicy is implemented [here](https://github.com/operator-framework/operator-controller/blob/main/config/base/catalogd/manager/network_policy.yaml).
+* The operator-controller is implemented [here](https://github.com/operator-framework/operator-controller/blob/main/config/base/operator-controller/manager/network_policy.yaml).
+
+This document explains the details of `NetworkPolicy` implementation for the core components.
+
+
+## Implementation Overview
+
+NetworkPolicy is implemented for both catalogd and operator-controller components to:
+
+* Restrict incoming (ingress) traffic to only required ports and services
+* Control outgoing (egress) traffic patterns
+
+Each component has a dedicated NetworkPolicy that applies to its respective pod through label selectors:
+
+* For catalogd: `control-plane=catalogd-controller-manager`
+* For operator-controller: `control-plane=operator-controller-controller-manager`
+
+### Catalogd NetworkPolicy
+
+- Ingress Rules
+Catalogd exposes three services, and its NetworkPolicy allows ingress traffic to the following TCP ports:
+
+* 7443: Metrics server for Prometheus metrics
+* 8443: Catalogd HTTPS server for catalog metadata API
+* 9443: Webhook server for Mutating Admission Webhook implementation
+
+All other ingress traffic to the catalogd pod is blocked.
+
+- Egress Rules
+Catalogd needs to communicate with:
+
+* The Kubernetes API server
+* Image registries specified in ClusterCatalog objects
+
+Currently, all egress traffic from catalogd is allowed, to support communication with arbitrary image registries that aren't known at install time.
+
+### Operator-Controller NetworkPolicy
+
+- Ingress Rules
+Operator-controller exposes one service, and its NetworkPolicy allows ingress traffic to:
+
+* 8443: Metrics server for Prometheus metrics
+
+All other ingress traffic to the operator-controller pod is blocked.
+
+- Egress Rules
+Operator-controller needs to communicate with:
+
+* The Kubernetes API server
+* Catalogd's HTTPS server (on port 8443)
+* Image registries specified in bundle metadata
+
+Currently, all egress traffic from operator-controller is allowed to support communication with arbitrary image registries that aren't known at install time.
+
+## Security Considerations
+
+The current implementation focuses on securing ingress traffic while allowing all egress traffic. This approach:
+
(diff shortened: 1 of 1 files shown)


docs/draft/api-reference/network-policies.md
```diff
@@ -0,0 +1,89 @@
...
+* Allows communication with arbitrary image registries
+* Establishes a foundation for future refinements to egress rules
+
+While allowing all egress does present some security risks, this implementation provides significant security improvements over having no network policies at all.
+
+## Troubleshooting Network Issues
+
+If you encounter network connectivity issues after deploying OLMv1, consider the following:
+
+* Verify NetworkPolicy support: Ensure your cluster has a CNI plugin that supports NetworkPolicy
```
camilamacedo86, line 71: Nit: If the CNI does not support NP, does this not mean that the NPs applied have no effect at all? So, I mean, if my k8s is using a CNI that does not provide this support is like it does not make any difference and no network connectivity issues could be faced. Am I right?

docs/draft/api-reference/network-policies.md
```diff
@@ -0,0 +1,89 @@
...
+* Check pod labels: Confirm that catalogd and operator-controller pods have the correct labels for NetworkPolicy selection
```
camilamacedo86, line 72: Could we be more specific here? 
For example, could we give the kubectl commands to help the user ensure that the labels used to do the match are present?

docs/draft/api-reference/network-policies.md
```diff
@@ -0,0 +1,89 @@
...
+* Inspect logs: Check component logs for connection errors
+* Test connectivity: Run test pods that attempt to communicate with OLMv1 components
```
camilamacedo86, line 74: Could we add an example here with the commands to help the user and us to know how to do those checks?

docs/draft/api-reference/network-policies.md
```diff
@@ -0,0 +1,89 @@
...
+- How NetworkPolicy is implemented with [network plugins](https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/network-plugins/) via the Container Network Interface (CNI)
+- Installing [Network Policy Providers](https://kubernetes.io/docs/tasks/administer-cluster/network-policy-provider/) documentation.
+
+## Future Enhancements
+
+The operator-framework team plan to revisit improvements to network policies in the future, such as:
+
+* More restrictive egress rules based on configured catalog image references
+* Further securing metrics and webhook server access
+* Dynamic network policy updates based on configured bundle image references
```
camilamacedo86, line 87: I think we do nto need add the FE part, when we enhance/if we enhance then we update this doc with the changes applied

docs/draft/api-reference/network-policies.md
```diff
@@ -0,0 +1,89 @@
...
+All other ingress traffic to the operator-controller pod is blocked.
+
+- Egress Rules
+Operator-controller needs to communicate with:
+
+* The Kubernetes API server
+* Catalogd's HTTPS server (on port 8443)
+* Image registries specified in bundle metadata
+
+Currently, all egress traffic from operator-controller is allowed to support communication with arbitrary image registries that aren't known at install time.
```
camilamacedo86, line 55: What do you think about linking the code from the repo?
So, those looking can easily check the NPs by clicking on them.

docs/draft/api-reference/network-policies.md
```diff
@@ -0,0 +1,89 @@
...
+
+All other ingress traffic to the catalogd pod is blocked.
+
+- Egress Rules
+Catalogd needs to communicate with:
+
+* The Kubernetes API server
+* Image registries specified in ClusterCatalog objects
+
+Currently, all egress traffic from catalogd is allowed, to support communication with arbitrary image registries that aren't known at install time.
```
camilamacedo86, line 37: What do you think about linking the code from the repo?
So, those looking can easily check the NPs by clicking on them.

camilamacedo86 (commented):
Good work, just some small fix and improvements. 🥇

docs/draft/api-reference/network-policies.md
anik120: Yup that's correct, I've added a sentence at the end to call this out. Thanks for the call out!

docs/draft/api-reference/network-policies.md
anik120: Done!

docs/draft/api-reference/network-policies.md
anik120: I'd actually started doing this, but it got very complicated very soon. That made me lean towards removing this line altogether, to keep this doc simple for now. We could either 

1. keep this out till we see the need to add this point with enhanced examples because end users are getting into this situation a lot, or 
2. add this section in a follow up PR. 

I'm definitely leaning toward 1 over 2.

docs/draft/api-reference/network-policies.md
anik120: Fair point, removed it.

docs/draft/api-reference/network-policies.md
anik120: I don't know how I feel about that tbh. If we just link the yaml file, that's self documenting 😄. 

This doc is to expand on that yaml file, in words. 

I feel like users who're adept enough at reading yaml files will find that file and know how to read it anyway, and this is for those users who will need help understanding that yaml file. 

Which is why I didn't include the link to the yaml file in the first place.

camilamacedo86: >I don't know how I feel about that tbh. If we just link the yaml file, that's self documenting 😄.
Yes, can we link the YAMLs?

For me, for example, it would be much easier to check the NP directly rather than reading the documentation.
As a new contributor and user, I might not even know where they live in the repo.

**So the suggestion here is simply to add a direct link to the YAML files.**

anik120: Added links to the `NetworkPolicy` yaml files from the repo in the overview section 👍🏽

camilamacedo86 (approved):
Great 👍 🥇 

It seems great for me to 🪰 
Thank you for your amazing contribution 🥇

PR #1979, opened 2025-05-20
🐛 fix: don't template registry+v1 manifests

joelanford:
# Description

Some registry+v1 manifests may actually contain Go Template strings that are meant to survive and actually persist into etcd (e.g. to be used as a templated configuration for another component). In order to avoid applying templating logic to registry+v1's static manifests, we create the manifests as Files, and then template those files via simple Templates.

## Reviewer Checklist

- [ ] API Go Documentation


grokspawn:
I wouldn't worry about the upgrade-e2e CI, but it seems weird to me that we are only now seeing the sanity/lint errors, and those look easy enough to resolve.

grokspawn (approved):

PR #1980, opened 2025-05-20
:warning: OPRUN-3954: Update CRD generator to support experimental CRDs

### LINKED ISSUE #2014: Introduce New CRD Generation Tooling
Introduce the new kububuilder CRD generation tool, based on the gateway-api code. This will change how CRDs are generated. A standard CRD and a new experimental CRD for both ClusterExtension and ClusterCatalog will be introduced. The standard CRDs will be generated and replace the current CRD. With the exception of a new annotation, the newly generated standard CRDs will be unchanged from the originals. This is one of the acceptance criteria for this step.

tmshort:
Fix #2014

Add support for a new CRD generator. This generator will use `opcon`
tags to identify experimental features. A script runs the CRD
generation, and creates all the CRDs.

New experimental CRDs are now created.

# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -138,20 +138,14 @@ tidy:
 	go mod tidy
 
 .PHONY: manifests
-KUSTOMIZE_CATD_CRDS_DIR := config/base/catalogd/crd/bases
 KUSTOMIZE_CATD_RBAC_DIR := config/base/catalogd/rbac
 KUSTOMIZE_CATD_WEBHOOKS_DIR := config/base/catalogd/manager/webhook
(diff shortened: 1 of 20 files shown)


tmshort:
Yes, `crd-diff` is failing, because I moved the location of the CRD file.

tmshort:
> It would be nice if we have the experimental fields to ensure that it works well as expected. However, I think we can change and improve in follow ups if you need to.

The unit test does try to use some experimental fields, but we don't have any in the main codebase.

PR #1996, opened 2025-05-28
:seedling: Add feature-gate kustomize files, docs, and demo for webhook support feature

perdasilva:
# Description

Adds docs and demo for webhook support feature

## Reviewer Checklist

- [ ] API Go Documentation


config/overlays/featuregate/webhook-provider-certmanager/kustomization.yaml
```diff
@@ -0,0 +1,15 @@
+# kustomization file for secure OLMv1
```
camilamacedo86, line 1: That's great, but could we change the title to clarify that we're adding more than just the documents?
Something like adding kustomize files to enable feature gates for the webhook feature and documentation.
WDYT?

perdasilva: fixed

hack/demo/webhook-provider-certmanager-demo.sh
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
camilamacedo86: Thanks for your comment – you made some good points! 😊
Just to share another view:
- We tested one today and it worked fine, no flakes.
- These tests check real features, like E2E, If something breaks, it’s probably the feature, not the docs. 
**(However, +1 for a possible duplication since we can end up have e2e tests and demos doing the same)**
- Performance might be a problem later, but we can wait and see.

**I think your opinion makes sense too, and it’s probably not something we need to fix now. 
All good from my side!**

camilamacedo86 (approved):
Hey @perdasilva 👋

I had another look at this one:

- We’re just adding the kustomize files, so in theory, it shouldn’t break anything downstream 👍
- Regarding the re-organisation in https://github.com/operator-framework/operator-controller/pull/1949 – I don’t think we should block this PR because of it. That effort might still evolve, and we can adjust later if needed.

So for me, everything looks good – LGTM 🚀

PR #2005, opened 2025-06-03
:sparkles: Support serviceaccount pull secrets

tmshort:
Serviceaccounts reference pull secrets!

* Determine our serviceaccount (via the new internal/shared/util/sa package).
* Use a common pull_secret_controller
* Update the pull_secret_controller to know about the service account
* Update the pull_secret_controller to watch the namespace-local secrets
* Update caching to include sa, and use filters for additional secrets
* Add RBAC to access these secrets and sa
* Update writing the auth.json file to handle dockercfg and dockerconfigjson
* Update writing the auth.json file to include multiple secrets

# Description

## Reviewer Checklist

- [ ] API Go Documentation


tmshort:
/hold
Until I do some additional manual testing.

config/base/catalogd/rbac/role.yaml
```diff
@@ -4,6 +4,14 @@ kind: ClusterRole
...
   name: manager-role
 rules:
+- apiGroups:
+  - ""
+  resources:
+  - secrets
+  verbs:
+  - get
+  - list
+  - watch
```
joelanford: Can this set of permissions be moved to a Role? Looks like that's how we have it setup for operator-controller. The image pull secrets attached to the service account are in our pod's namespace.

internal/shared/controllers/pull_secret_controller.go
```diff
@@ -0,0 +1,228 @@
...
+		// Add the configured pull secret to the list of secrets
+		if err := r.Get(ctx, *r.SecretKey, secret); err != nil {
+			if apierrors.IsNotFound(err) {
+				logger.Info("secret not found", "name", r.SecretKey)
+			} else {
+				logger.Error(err, "failed to get Secret", "name", r.SecretKey)
+				return ctrl.Result{}, err
+			}
+		} else {
+			logger.Info("global pull secret", "name", *r.SecretKey)
```
joelanford: Separate out name and namespace?
```suggestion
			logger.Info("global pull secret", "name", r.SecretKey.Name, "namespace", r.SecretKey.Namespace)
```

internal/shared/controllers/pull_secret_controller.go
```diff
@@ -0,0 +1,228 @@
...
+func (r *PullSecretReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	logger := log.FromContext(ctx).WithName("pull-secret-reconciler")
+
+	logger.Info("processing event", "name", req.NamespacedName)
+	defer logger.Info("processed event", "name", req.NamespacedName)
+
+	secrets := []*corev1.Secret{}
+	secret := &corev1.Secret{}
+
+	if r.SecretKey != nil { //nolint:nestif
```
joelanford: Might be nice to put this block in a separate function like `secret, err := r.getSecret(ctx, secretKey)`

Then we could reuse that function for both the global pull secret and the service account secrets, and it would probably no longer run afoul of the nestif linter.

internal/shared/controllers/pull_secret_controller.go
```diff
@@ -0,0 +1,228 @@
...
+	}
+
+	data, err := json.Marshal(jsonData)
+	if err != nil {
+		return fmt.Errorf("failed to marshal secret data: %w", err)
+	}
+	err = os.WriteFile(r.AuthFilePath, data, 0600)
+	if err != nil {
+		return fmt.Errorf("failed to write secret data to file: %w", err)
+	}
```
joelanford: This could be racy because writing this file could potentially happen at the same time that our other controllers are trying to read it. Maybe use `github.com/google/renameio/v2` to get atomic writes?

internal/shared/controllers/pull_secret_controller.go
tmshort: Did you want that for all the related logs?

internal/shared/controllers/pull_secret_controller.go
tmshort: I'll look into it.

config/base/catalogd/rbac/role.yaml
tmshort: Hmm.... I'm not 100% sure it can, because it needs to be able to access the global pull secret which could be in any namespace. But I suppose that's more of a downstream issue, as the default deployment upstream does not supply a global pull secret, and users can always add the proper RBAC.

internal/shared/controllers/pull_secret_controller.go
tmshort: I thought of that, but there's still the same set of failure modes that need to be checked, so I'm not sure it would reduce the complexity. But I can look.

config/base/catalogd/rbac/role.yaml
joelanford: > But I suppose that's more of a downstream issue, as the default deployment upstream does not supply a global pull secret, and users can always add the proper RBAC.

+100

internal/shared/controllers/pull_secret_controller.go
joelanford: Same overall complexity, but at the function scope the complexity would be reduced because `r.Get` would happen in the main flow of the function and the subsequent error checks would be in the first/second levels rather than second/third.

Plus there's still the re-use angle to it.

internal/shared/controllers/pull_secret_controller.go
joelanford: Yeah, and changing `name` to `key` would probably work too?

internal/shared/controllers/pull_secret_controller.go
```diff
@@ -0,0 +1,228 @@
...
+}
+
+func (r *PullSecretReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	logger := log.FromContext(ctx).WithName("pull-secret-reconciler")
+
+	logger.Info("processing event", "name", req.NamespacedName)
+	defer logger.Info("processed event", "name", req.NamespacedName)
+
+	secrets := []*corev1.Secret{}
+	secret := &corev1.Secret{}
```
joelanford: Doesn't look like we are with the current code, but we shouldn't reuse this in multiple calls to `Get` because it does not get zero'd out.

I'd suggest moving this directly before the `r.Get` call to make it less likely to accidentally be re-used if there are changes that happen in this function later.  If we add a `getSecret` function to encapsulate, that would be a good place to initialize it.

tmshort: Adding the `getSecret` function fixes this issue.

internal/shared/controllers/pull_secret_controller.go
```diff
@@ -0,0 +1,228 @@
...
+// that contains global pull secrets for pulling Catalog images
+type PullSecretReconciler struct {
+	client.Client
+	SecretKey                 *types.NamespacedName
+	ServiceAccountKey         types.NamespacedName
+	ServiceAccountPullSecrets []types.NamespacedName
+	AuthFilePath              string
+}
+
+func (r *PullSecretReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
joelanford: Is `req` used anywhere other than logging? If not, I think I'd suggest dropping it (e.g. rename to `_`). And changing the logging to just generically say something like "reconciling pull secrets".

The name/namespace of the request without the type info might be a little confusing to show up in the log. But maybe we could log the events that pass our predicate in our predicate where we _do_ have type information. That way there's still detail about what is causing our reconciler to be triggered?

internal/shared/controllers/pull_secret_controller.go
```diff
@@ -0,0 +1,228 @@
...
+		Build(r)
+	if err != nil {
+		return err
+	}
+
+	_, err = ctrl.NewControllerManagedBy(mgr).
+		For(&corev1.ServiceAccount{}).
+		Named("service-account-controller").
+		WithEventFilter(newNamespacedPredicate(r.ServiceAccountKey)).
+		Build(r)
```
joelanford, line 141: I don't think I'd setup two separate controllers. IIRC, there's a way to have a single controller with multiple watches. You may need to drop down to the lower-level controller package though (can't remember if `For` is required).

tmshort:
lint doesn't like my logging trick...

internal/shared/controllers/pull_secret_controller.go
tmshort: Interestingly... the logger has a lot of additional information added to it, so you _can_ tell what type of resource triggered the event, I wrapped the log below to make it easier to read:

```
I0606 18:43:42.641045       1 pull_secret_controller.go:113] "found secret"
  logger="pull-secret-reconciler"
  controller="service-account-controller"
  controllerGroup=""
  controllerKind="ServiceAccount"
  ServiceAccount="olmv1-system/operator-controller-controller-manager"
  reconcileID="26404229-a2d3-495d-86ee-0da390a1e8f4"
  name="pull-dockercfg"
  namespace="olmv1-system"

Comment (tmshort) [reply internal/shared/controllers/pull_secret_controller.go]: And when a pull secret is modified:
```
I0606 18:58:10.831129       1 pull_secret_controller.go:113] "found secret"
  logger="pull-secret-reconciler"
  controller="pull-secret-controller"
  controllerGroup=""
  controllerKind="Secret"
  Secret="olmv1-system/pull-dockercfg"
  reconcileID="a783ca58-0213-4fe3-a19d-3ed7ca21f5ae"
  name="pull-dockercfg"
  namespace="olmv1-system"

internal/shared/controllers/pull_secret_controller.go
tmshort: TBH, having two controllers actually seems the cleanest, in that we get different invocations that recognize ServiceAccount vs. Secret, and it's clear what triggered the reconcile (i.e. the logger clearly indicates what resource is the trigger). With a single controller, we have to somehow map the ServiceAccount to a set of Secrets (which we may not yet know about yet), which feels kinda kludgy. Otherwise, the Secret controller is getting a ServiceAccount for the request, rather than a Secret.

cmd/catalogd/main.go
```diff
@@ -246,18 +248,40 @@ func run(ctx context.Context) error {
...
-					LabelSelector: k8slabels.Everything(),
-					FieldSelector: fields.SelectorFromSet(map[string]string{
-						"metadata.name": cfg.globalPullSecretKey.Name,
-					}),
-				},
+
+	saKey, err := sautil.GetServiceAccount()
+	if err != nil {
+		setupLog.Error(err, "Unable to get pod namesapce and serviceaccount")
+		return err
```
camilamacedo86:
```suggestion
		setupLog.Error(err, "Failed to extract namespace/serviceaccount from JWT token")
		return fmt.Errorf("failed to get service account identity from token: %w", err)
```

cmd/catalogd/main.go
```diff
@@ -246,18 +248,40 @@ func run(ctx context.Context) error {
...
+	}
+
+	setupLog.Info("Read token", "serviceaccount", saKey)
```
camilamacedo86:
```suggestion
	if saKey.Namespace == "" || saKey.Name == "" {
		setupLog.Error(nil, "Extracted service account name or namespace is empty", "saKey", saKey)
		return fmt.Errorf("invalid service account identity extracted: %v", saKey)
	}

	setupLog.Info("Successfully extracted serviceaccount identity from token", "namespace", saKey.Namespace, "name", saKey.Name)
```

cmd/operator-controller/main.go
```diff
@@ -217,18 +219,40 @@ func run() error {
...
-					}),
-				},
+
+	saKey, err := sautil.GetServiceAccount()
+	if err != nil {
+		setupLog.Error(err, "Unable to get pod namesapce and serviceaccount")
+		return err
+	}
+
+	setupLog.Info("Read token", "serviceaccount", saKey)
```
camilamacedo86: The block below appears identical in both cmd/catalogd/main.go and cmd/operator-controller/main.go.
Could we move this logic into a helper function under pkg/shared/util/sa and call it from both main.go files? That would reduce duplication and help with future changes

camilamacedo86 (commented):
Just a few suggestions on possible improvements, mostly around structuring and reusability — nothing blocking though. 

The overall implementation looks good for me.
Great work! 🎉 

Otherwise, all good from my side — LGTM

cmd/catalogd/main.go
tmshort: There's no point in checking for an empty name or namespace, as the `sautil.GetServiceAccount()` function should (will) error out if both can't be retrieved.

cmd/operator-controller/main.go
tmshort: Yup, I can try to consolidate a bit.

cmd/catalogd/main.go
tmshort: It's a bit redundant to wrap the error, since it's also logged.

joelanford (approved):

camilamacedo86 (approved):

PR #2007, opened 2025-06-03
🐛 set GIT_COMMIT in Makefile to populate version info 

rashmigottipati:
# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,9 @@ SHELL := /usr/bin/env bash -o pipefail
 .SHELLFLAGS := -ec
 export ROOT_DIR := $(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
 
+# use CI-provided SOURCE_GIT_COMMIT if available, fallback to local git commit ID
+export GIT_COMMIT := $(or $(SOURCE_GIT_COMMIT),$(shell git rev-parse HEAD))
+
 GOLANG_VERSION := $(shell sed -En 's/^go (.*)$$/\1/p' "go.mod")
 # Image URL to use all building/pushing image targets
 ifeq ($(origin IMAGE_REGISTRY), undefined)


joelanford:
This still feels off to me from a supply chain security perspective.

Under what circumstances does it make sense for a CI system to build commit `a` but say that it is commit `b`?

joelanford:
/hold

Can we discuss this a bit more? Besides my concern about supply chain security, this still feels like a downstream-only concern.

rashmigottipati:
@joelanford  Thanks for raising this. You're right—if the build is for commit A, we absolutely shouldn't be marking it as commit B and this is certainly a concern from a supply chain security standpoint. 

This is a downstream only concern. Previously I've created https://github.com/operator-framework/operator-controller/pull/1811 but that didn't resolve the bug (downstream) as confirmed by QE in the report (which says that with the latest change it's returning empty commit info).
 
The purpose of this change is to ensure that GIT_COMMIT gets correctly set in environments where the actual git metadata isn't available at build time—specifically in some downstream OpenShift build pipelines. It looks like in those cases, the build system injects the commit SHA via the SOURCE_GIT_COMMIT env var, probably due to not having access to .git directly (this is my assumption - I am not very familiar with the build process downstream). Without it, the resulting binaries end up with an empty gitCommit field.

Upstream, we expect `git rev-parse HEAD` to suffice, and that’s what is used as a fallback here.

If the concern is around upstream builds incorrectly using SOURCE_GIT_COMMIT, maybe we could explicitly scope the override to CI? Or, we could leave GIT_COMMIT empty unless it's explicitly set by the caller. Or we could make this change downstream and leave upstream as it is. 

Happy to adjust based on what we agree is the cleanest path forward.

joelanford:
I'm guessing that https://github.com/operator-framework/operator-controller/pull/1811 is the upstream plumbing that was needed such that we could add some downstream plumbing to set `GIT_COMMIT` to whatever downstream CI has set in its environment for `SOURCE_GIT_COMMIT`. I think I'd expect our downstream Dockerfiles to do something like (not sure this syntax is correct off the top of my head):
```dockerfile
ENV GIT_COMMIT=$SOURCE_GIT_COMMIT
```
prior to calling `make go-build-local`. For example, just before here: https://github.com/openshift/operator-framework-operator-controller/blob/6bb442f94f89eba1c2c05b72f711087397706b54/openshift/operator-controller.Dockerfile#L4

perdasilva:
closing this as stale - we should reopen when there's more bandwidth for it

PR #2013, opened 2025-06-05
✨OPRUN-3873: Add e2e tests for NetworkPolicies

anik120:
# Description

This PR introduces a new e2e test to continuously validate the NetworkPolicy objects for the catalogd and operator-controller components. 

The new test, `TestNetworkPolicyJustifications`, achieves the following:

* Documents NetworkPolicies deployed: A registry of all allowed network policies (`allowedNetworkPolicies`) is defined using the [RFC: OLMv1 Static Network Policies](https://docs.google.com/document/d/1bHEEWzA65u-kjJFQRUY1iBuMIIM1HbPy4MeDLX4NI3o/edit?tab=t.0#heading=h.x3tfh25grvnv). Every rule within these policies is explicitly registered in the registry with a clear, written justification of a minimum length. This ensures that every rule's purpose is documented and understood.

* Prevents Regressions: The test validates that the NetworkPolicy objects deployed in the cluster exactly match the definitions in a registry (`allowedNetworkPolicies`). This prevents undocumented or accidental changes to our network rules.

* Improves Security Posture: Enforces thoughtful consideration for any change to existing network rules, improving overall security posture. Also ensures that other, unrelated network policies won't inadvertently break existing functionalities.

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/test/e2e/network_policy_test.go
+++ b/test/e2e/network_policy_test.go
@@ -1,0 +1,337 @@
+package e2e
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	networkingv1 "k8s.io/api/networking/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/operator-framework/operator-controller/test/utils"
+)
+
+const (
+	minJustificationLength        = 40
+	catalogdManagerSelector       = "control-plane=catalogd-controller-manager"
+	operatorManagerSelector       = "control-plane=operator-controller-controller-manager"
+	catalogdMetricsPort           = 7443
+	catalogdWebhookPort           = 9443
+	catalogServerPort             = 8443
+	operatorControllerMetricsPort = 8443
+)
+
+type portWithJustification struct {
+	port          []networkingv1.NetworkPolicyPort
+	justification string
+}
+
+// ingressRule defines a k8s IngressRule, along with a justification.
+type ingressRule struct {
+	ports []portWithJustification
+	from  []networkingv1.NetworkPolicyPeer
+}
+
+// egressRule defines a k8s egressRule, along with a justification.
+type egressRule struct {
+	ports []portWithJustification
+	to    []networkingv1.NetworkPolicyPeer
+}
+
+// AllowedPolicyDefinition defines the expected structure and justifications for a NetworkPolicy.
+type allowedPolicyDefinition struct {
+	selector                    metav1.LabelSelector
+	policyTypes                 []networkingv1.PolicyType
+	ingressRule                 ingressRule
+	egressRule                  egressRule
+	denyAllIngressJustification string // Justification if Ingress is in PolicyTypes and IngressRules is empty
+	denyAllEgressJustification  string // Justification if Egress is in PolicyTypes and EgressRules is empty
+}
+
+// Ref: https://docs.google.com/document/d/1bHEEWzA65u-kjJFQRUY1iBuMIIM1HbPy4MeDLX4NI3o/edit?usp=sharing
+var allowedNetworkPolicies = map[string]allowedPolicyDefinition{
+	"catalogd-controller-manager": {
+		selector:    metav1.LabelSelector{MatchLabels: map[string]string{"control-plane": "catalogd-controller-manager"}},
+		policyTypes: []networkingv1.PolicyType{networkingv1.PolicyTypeIngress, networkingv1.PolicyTypeEgress},
+		ingressRule: ingressRule{
+			ports: []portWithJustification{
+				{
+					port:          []networkingv1.NetworkPolicyPort{{Protocol: ptr.To(corev1.ProtocolTCP), Port: &intstr.IntOrString{Type: intstr.Int, IntVal: catalogdMetricsPort}}},
+					justification: "Allows Prometheus to scrape metrics from catalogd, which is essential for monitoring its performance and health.",
+				},
+				{
+					port:          []networkingv1.NetworkPolicyPort{{Protocol: ptr.To(corev1.ProtocolTCP), Port: &intstr.IntOrString{Type: intstr.Int, IntVal: catalogdWebhookPort}}},
+					justification: "Permits Kubernetes API server to reach catalogd's mutating admission webhook, ensuring integrity of catalog resources.",
+				},
+				{
+					port:          []networkingv1.NetworkPolicyPort{{Protocol: ptr.To(corev1.ProtocolTCP), Port: &intstr.IntOrString{Type: intstr.Int, IntVal: catalogServerPort}}},
+					justification: "Enables clients (eg. operator-controller) to query catalog metadata from catalogd, which is a core function for bundle resolution and operator discovery.",
+				},
+			},
+		},
+		egressRule: egressRule{
+			ports: []portWithJustification{
+				{
+					port:          nil, // Empty Ports means allow all egress
+					justification: "Permits catalogd to fetch catalog images from arbitrary container registries and communicate with the Kubernetes API server for its operational needs.",
+				},
+			},
+		},
+	},
+	"operator-controller-controller-manager": {
+		selector:    metav1.LabelSelector{MatchLabels: map[string]string{"control-plane": "operator-controller-controller-manager"}},
+		policyTypes: []networkingv1.PolicyType{networkingv1.PolicyTypeIngress, networkingv1.PolicyTypeEgress},
+		ingressRule: ingressRule{
+			ports: []portWithJustification{
+				{
+					port:          []networkingv1.NetworkPolicyPort{{Protocol: ptr.To(corev1.ProtocolTCP), Port: &intstr.IntOrString{Type: intstr.Int, IntVal: operatorControllerMetricsPort}}},
+					justification: "Allows Prometheus to scrape metrics from operator-controller, which is crucial for monitoring its activity, reconciliations, and overall health.",
+				},
+			},
+		},
+		egressRule: egressRule{
+			ports: []portWithJustification{
+				{
+					port:          nil, // Empty Ports means allow all egress
+					justification: "Enables operator-controller to pull bundle images from arbitrary image registries, connect to catalogd's HTTPS server for metadata, and interact with the Kubernetes API server.",
+				},
+			},
+		},
+	},
+	"default-deny-all-traffic": {
+		selector:    metav1.LabelSelector{}, // Empty selector, matches all pods
+		policyTypes: []networkingv1.PolicyType{networkingv1.PolicyTypeIngress, networkingv1.PolicyTypeEgress},
+		// No IngressRules means deny all ingress if PolicyTypeIngress is present
+		// No EgressRules means deny all egress if PolicyTypeEgress is present
+		denyAllIngressJustification: "Denies all ingress traffic to pods selected by this policy by default, unless explicitly allowed by other policy rules, ensuring a baseline secure posture.",
+		denyAllEgressJustification:  "Denies all egress traffic from pods selected by this policy by default, unless explicitly allowed by other policy rules, minimizing potential exfiltration paths.",
+	},
+}
+
+func TestNetworkPolicyJustifications(t *testing.T) {
+	ctx := context.Background()
+
+	// Validate justifications have min length in the allowedNetworkPolicies definition
+	for name, policyDef := range allowedNetworkPolicies {
+		for i, pwj := range policyDef.ingressRule.ports {
+			assert.GreaterOrEqualf(t, len(pwj.justification), minJustificationLength,
+				"Justification for ingress PortWithJustification entry %d in policy %q is too short: %q", i, name, pwj.justification)
+		}
+		for i, pwj := range policyDef.egressRule.ports { // Corrected variable name from 'rule' to 'pwj'
+			assert.GreaterOrEqualf(t, len(pwj.justification), minJustificationLength,
+				"Justification for egress PortWithJustification entry %d in policy %q is too short: %q", i, name, pwj.justification)
+		}
+		if policyDef.denyAllIngressJustification != "" {
+			assert.GreaterOrEqualf(t, len(policyDef.denyAllIngressJustification), minJustificationLength,
+				"DenyAllIngressJustification for policy %q is too short: %q", name, policyDef.denyAllIngressJustification)
+		}
+		if policyDef.denyAllEgressJustification != "" {
+			assert.GreaterOrEqualf(t, len(policyDef.denyAllEgressJustification), minJustificationLength,
+				"DenyAllEgressJustification for policy %q is too short: %q", name, policyDef.denyAllEgressJustification)
+		}
+	}
+
+	clientForComponent := utils.FindK8sClient(t)
+	componentNamespace := getComponentNamespace(t, clientForComponent, operatorManagerSelector)
+	clusterPolicies := &networkingv1.NetworkPolicyList{}
+	err := c.List(ctx, clusterPolicies, client.InNamespace(componentNamespace))
+	require.NoError(t, err, "Failed to list NetworkPolicies in namespace %q", componentNamespace)
+
+	validatedRegistryPolicies := make(map[string]bool)
+
+	for _, policy := range clusterPolicies.Items {
+		t.Run(fmt.Sprintf("Policy_%s", strings.ReplaceAll(policy.Name, "-", "_")), func(t *testing.T) {
+			expectedPolicy, found := allowedNetworkPolicies[policy.Name]
+			require.Truef(t, found, "NetworkPolicy %q found in cluster but not in allowed registry. Namespace: %s", policy.Name, policy.Namespace)
+			validatedRegistryPolicies[policy.Name] = true
+
+			// 1. Compare PodSelector
+			assert.True(t, equality.Semantic.DeepEqual(expectedPolicy.selector, policy.Spec.PodSelector),
+				"PodSelector mismatch for policy %q. Expected: %+v, Got: %+v", policy.Name, expectedPolicy.selector, policy.Spec.PodSelector)
+
+			// 2. Compare PolicyTypes
+			require.ElementsMatchf(t, expectedPolicy.policyTypes, policy.Spec.PolicyTypes,
+				"PolicyTypes mismatch for policy %q.", policy.Name)
+
+			// 3. Validate Ingress Rules
+			hasIngressPolicyType := false
+			for _, pt := range policy.Spec.PolicyTypes {
+				if pt == networkingv1.PolicyTypeIngress {
+					hasIngressPolicyType = true
+					break
+				}
+			}
+
+			if hasIngressPolicyType {
+				switch len(policy.Spec.Ingress) {
+				case 0:
+					validateDenyAllIngress(t, policy.Name, expectedPolicy)
+				case 1:
+					validateSingleIngressRule(t, policy.Name, policy.Spec.Ingress[0], expectedPolicy)
+				default:
+					assert.Failf(t, "Policy %q in cluster has %d ingress rules. Allowed definition supports at most 1 explicit ingress rule.", policy.Name, len(policy.Spec.Ingress))
+				}
+			} else {
+				validateNoIngress(t, policy.Name, policy, expectedPolicy)
+			}
+
+			// 4. Validate Egress Rules
+			hasEgressPolicyType := false
+			for _, pt := range policy.Spec.PolicyTypes {
+				if pt == networkingv1.PolicyTypeEgress {
+					hasEgressPolicyType = true
+					break
+				}
+			}
+
+			if hasEgressPolicyType {
+				switch len(policy.Spec.Egress) {
+				case 0:
+					validateDenyAllEgress(t, policy.Name, expectedPolicy)
+				case 1:
+					validateSingleEgressRule(t, policy.Name, policy.Spec.Egress[0], expectedPolicy)
+				default:
+					assert.Failf(t, "Policy %q in cluster has %d egress rules. Allowed definition supports at most 1 explicit egress rule.", policy.Name, len(policy.Spec.Egress))
+				}
+			} else {
+				validateNoEgress(t, policy, expectedPolicy)
+			}
+		})
+	}
+
+	// 5. Ensure all policies in the registry were found in the cluster
+	assert.Equal(t, len(allowedNetworkPolicies), len(validatedRegistryPolicies),
+		"Mismatch between number of expected policies in registry (%d) and number of policies found & validated in cluster (%d). Missing policies from registry: %v", len(allowedNetworkPolicies), len(validatedRegistryPolicies), missingPolicies(allowedNetworkPolicies, validatedRegistryPolicies))
+}
+
+func missingPolicies(expected map[string]allowedPolicyDefinition, actual map[string]bool) []string {
+	missing := []string{}
+	for k := range expected {
+		if !actual[k] {
+			missing = append(missing, k)
+		}
+	}
+	return missing
+}
+
+// validateNoEgress confirms that a policy which does not have spec.PolicyType=Egress specified
+// has no corresponding egress rules or expectations defined.
+func validateNoEgress(t *testing.T, policy networkingv1.NetworkPolicy, expectedPolicy allowedPolicyDefinition) {
+	// Policy is NOT expected to affect Egress traffic (no Egress in PolicyTypes)
+	// Expected: Cluster has no egress rules; Registry has no DenyAllEgressJustification and empty EgressRule.
+	require.Emptyf(t, policy.Spec.Egress,
+		"Policy %q: Cluster does not have Egress PolicyType, but has Egress rules defined.", policy.Name)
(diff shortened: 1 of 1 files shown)


test/e2e/network_policy_test.go
```diff
@@ -0,0 +1,247 @@
...
+				Ports:         []networkingv1.NetworkPolicyPort{{Protocol: ptr.To(corev1.ProtocolTCP), Port: intOrStrPtr(7443)}},
+				Justification: "Allows Prometheus to scrape metrics from catalogd, providing observability into its performance and health. This is essential for monitoring.",
+			},
+			{
+				Ports:         []networkingv1.NetworkPolicyPort{{Protocol: ptr.To(corev1.ProtocolTCP), Port: intOrStrPtr(8443)}},
+				Justification: "Enables operator-controller to query catalog metadata from catalogd. This is a core function for bundle resolution and operator discovery.",
+			},
+			{
+				Ports:         []networkingv1.NetworkPolicyPort{{Protocol: ptr.To(corev1.ProtocolTCP), Port: intOrStrPtr(9443)}},
+				Justification: "Permits Kubernetes API server to reach catalogd's admission webhook for CRD validation, ensuring integrity of catalog resources.",
```
joelanford: It's a mutating webhook that sets a label for the metadata.name that can be used by selectors in ClusterExtension objects to choose specific catalogs by name.
```suggestion
				Justification: "Permits Kubernetes API server to reach catalogd's mutating admission webhook, ensuring integrity of catalog resources.",
```

test/e2e/network_policy_test.go
```diff
@@ -0,0 +1,247 @@
...
+		Selector:    metav1.LabelSelector{MatchLabels: map[string]string{"control-plane": "catalogd-controller-manager"}},
+		PolicyTypes: []networkingv1.PolicyType{networkingv1.PolicyTypeIngress, networkingv1.PolicyTypeEgress},
+		IngressRules: []IngressRule{
+			{
+				Ports:         []networkingv1.NetworkPolicyPort{{Protocol: ptr.To(corev1.ProtocolTCP), Port: intOrStrPtr(7443)}},
+				Justification: "Allows Prometheus to scrape metrics from catalogd, providing observability into its performance and health. This is essential for monitoring.",
+			},
+			{
+				Ports:         []networkingv1.NetworkPolicyPort{{Protocol: ptr.To(corev1.ProtocolTCP), Port: intOrStrPtr(8443)}},
+				Justification: "Enables operator-controller to query catalog metadata from catalogd. This is a core function for bundle resolution and operator discovery.",
```
joelanford:
```suggestion
				Justification: "Enables clients (e.g. operator-controller) to query catalog metadata from catalogd. This is a core function for bundle resolution and operator discovery.",
```

test/e2e/network_policy_test.go
```diff
@@ -0,0 +1,247 @@
...
+			},
+			{
+				Ports:         []networkingv1.NetworkPolicyPort{{Protocol: ptr.To(corev1.ProtocolTCP), Port: intOrStrPtr(9443)}},
+				Justification: "Permits Kubernetes API server to reach catalogd's admission webhook for CRD validation, ensuring integrity of catalog resources.",
+			},
+		},
+		EgressRules: []EgressRule{
+			{
+				// Empty Ports and To means allow all egress
+				Justification: "Permits catalogd to fetch catalog images from various container registries and communicate with the Kubernetes API server for its operational needs.",
```
joelanford:
```suggestion
				Justification: "Permits catalogd to fetch catalog images from arbitrary container registries and communicate with the Kubernetes API server for its operational needs.",
```

test/e2e/network_policy_test.go
```diff
@@ -0,0 +1,247 @@
...
+		IngressRules: []IngressRule{
+			{
+				Ports:         []networkingv1.NetworkPolicyPort{{Protocol: ptr.To(corev1.ProtocolTCP), Port: intOrStrPtr(8443)}},
+				Justification: "Allows Prometheus to scrape metrics from operator-controller, crucial for monitoring its activity, reconciliations, and overall health.",
+			},
+		},
+		EgressRules: []EgressRule{
+			{
+				// Empty Ports and To means allow all egress
+				Justification: "Enables operator-controller to pull bundle images, connect to catalogd's HTTPS server for metadata, and interact with the Kubernetes API server.",
```
joelanford:
```suggestion
				Justification: "Enables operator-controller to pull bundle images from arbitrary image registries, connect to catalogd's HTTPS server for metadata, and interact with the Kubernetes API server.",
```

anik120:
@joelanford incorporated your suggestions..also fyi had to change the struct definitions a bit because I realized the policies are defined as a single ingress/egress policy with multiple ports, instead of multiple ingress/egress policies, for each NetworkPolicy object. PTAL, thanks!

test/e2e/network_policy_test.go
```diff
@@ -0,0 +1,336 @@
...
+				{
+					Port:          []networkingv1.NetworkPolicyPort{{Protocol: ptr.To(corev1.ProtocolTCP), Port: intOrStrPtr(8443)}},
+					Justification: "Allows Prometheus to scrape metrics from catalogd, which is essential for monitoring its performance and health.",
+				},
+				{
+					Port:          []networkingv1.NetworkPolicyPort{{Protocol: ptr.To(corev1.ProtocolTCP), Port: intOrStrPtr(9443)}},
+					Justification: "Permits Kubernetes API server to reach catalogd's mutating admission webhook, ensuring integrity of catalog resources.",
+				},
+				{
+					Port:          []networkingv1.NetworkPolicyPort{{Protocol: ptr.To(corev1.ProtocolTCP), Port: intOrStrPtr(7443)}},
```
dtfranz: I believe this should be swapped with 8443, since 7443 is the metrics port on catalogd (and 8443 is metrics for operator-controller). Referencing [this](https://github.com/operator-framework/operator-controller/blob/44de6f2e03c2f02845c9d7362ade607a0b0c501d/config/base/catalogd/manager/manager.yaml#L48)

test/e2e/network_policy_test.go
```diff
@@ -0,0 +1,336 @@
...
+					Justification: "Enables clients (eg. operator-controller) to query catalog metadata from catalogd, which is a core function for bundle resolution and operator discovery.",
+				},
```
dtfranz:
```suggestion
				{
					Port:          []networkingv1.NetworkPolicyPort{{Protocol: ptr.To(corev1.ProtocolTCP), Port: intOrStrPtr(7443)}},
					Justification: "Allows Prometheus to scrape metrics from catalogd, which is essential for monitoring its performance and health.",
				},
				{
					Port:          []networkingv1.NetworkPolicyPort{{Protocol: ptr.To(corev1.ProtocolTCP), Port: intOrStrPtr(9443)}},
					Justification: "Permits Kubernetes API server to reach catalogd's mutating admission webhook, ensuring integrity of catalog resources.",
				},
				{
					Port:          []networkingv1.NetworkPolicyPort{{Protocol: ptr.To(corev1.ProtocolTCP), Port: intOrStrPtr(8443)}},
					Justification: "Enables clients (eg. operator-controller) to query catalog metadata from catalogd, which is a core function for bundle resolution and operator discovery.",
				},
```

test/e2e/network_policy_test.go
perdasilva: nit: maybe move the ports to consts?

test/e2e/network_policy_test.go
```diff
@@ -0,0 +1,336 @@
...
+)
+
+const (
+	olmSystemNamespace      = "olmv1-system"
+	minJustificationLength  = 40
+	catalogdManagerSelector = "control-plane=catalogd-controller-manager"
+	operatorManagerSelector = "control-plane=operator-controller-controller-manager"
+)
+
+type PortWithJustification struct {
```
perdasilva: nit: do these need to be exported? (PortWithJustification, IngressRule, EgressRule, etc.)

test/e2e/network_policy_test.go
```diff
@@ -0,0 +1,336 @@
...
+		PolicyTypes: []networkingv1.PolicyType{networkingv1.PolicyTypeIngress, networkingv1.PolicyTypeEgress},
+		IngressRule: IngressRule{
+			Ports: []PortWithJustification{
+				{
+					Port:          []networkingv1.NetworkPolicyPort{{Protocol: ptr.To(corev1.ProtocolTCP), Port: intOrStrPtr(8443)}},
+					Justification: "Allows Prometheus to scrape metrics from catalogd, which is essential for monitoring its performance and health.",
+				},
+				{
+					Port:          []networkingv1.NetworkPolicyPort{{Protocol: ptr.To(corev1.ProtocolTCP), Port: intOrStrPtr(9443)}},
+					Justification: "Permits Kubernetes API server to reach catalogd's mutating admission webhook, ensuring integrity of catalog resources.",
```
perdasilva: kube apiserver and dns ports could change downstream (or in different installations). We might want a way to configure those =S

test/e2e/network_policy_test.go
anik120: Good point, un-exported them

test/e2e/network_policy_test.go
anik120: 👍🏽 moved them to constants

test/e2e/network_policy_test.go
anik120: that's a great catch thank you @dtfranz!! Updated the PR, PTAL

test/e2e/network_policy_test.go
anik120: hhhmm really good point. 

So I was looking at the manifests downstream: looks like we have the same ports configured for now (see [catalogd's network policy](https://github.com/openshift/operator-framework-operator-controller/blob/main/config/base/catalogd/manager/network_policy.yaml#L15-L20) and [operator-controller's network policy](https://github.com/openshift/operator-framework-operator-controller/blob/main/config/base/operator-controller/manager/network_policy.yaml#L13-L18)).


We could configure this using a env vars: 

```
# Makefile
export E2E_CATALOGD_METRICS_PORT ?= 7443
export E2E_SERVER_PORT ?= 8443
```

and then read those env vars in the test, which would enable us to configure these variable in the Makefile downstream (or anywhere else). 

But it doesn't look like we have to do that for now? 


This comment did make me look into my usage of 

```
const olmv1systemNamespace = olmv1-system
```

and now I switched to an existing helper function `getComponentNamespace` from the `metrics_test.go` file, so thank you for that :)

test/e2e/network_policy_test.go
```diff
@@ -0,0 +1,343 @@
...
+	// 5. Ensure all policies in the registry were found in the cluster
+	assert.Equal(t, len(allowedNetworkPolicies), len(validatedRegistryPolicies),
+		"Mismatch between number of expected policies in registry (%d) and number of policies found & validated in cluster (%d). Missing policies from registry: %v", len(allowedNetworkPolicies), len(validatedRegistryPolicies), missingPolicies(allowedNetworkPolicies, validatedRegistryPolicies))
+}
+
+// Helper function to create a pointer to intstr.IntOrString from an int
+func intOrStrPtr(port int32) *intstr.IntOrString {
+	val := intstr.FromInt(int(port))
+	return &val
+}
```
camilamacedo86:
```suggestion
```

We do not need the helper.
We can use k8s.io/apimachinery/pkg/util/intstr  (since we import this lib already)
and then something like `Port: intstr.FromInt(NUMBER)`

test/e2e/network_policy_test.go
```diff
@@ -0,0 +1,343 @@
...
+// Ref: https://docs.google.com/document/d/1bHEEWzA65u-kjJFQRUY1iBuMIIM1HbPy4MeDLX4NI3o/edit?usp=sharing
+var allowedNetworkPolicies = map[string]allowedPolicyDefinition{
+	"catalogd-controller-manager": {
+		selector:    metav1.LabelSelector{MatchLabels: map[string]string{"control-plane": "catalogd-controller-manager"}},
+		policyTypes: []networkingv1.PolicyType{networkingv1.PolicyTypeIngress, networkingv1.PolicyTypeEgress},
+		ingressRule: ingressRule{
+			ports: []portWithJustification{
+				{
+					port:          []networkingv1.NetworkPolicyPort{{Protocol: ptr.To(corev1.ProtocolTCP), Port: intOrStrPtr(catalogdMetricsPort)}},
+					justification: "Allows Prometheus to scrape metrics from catalogd, which is essential for monitoring its performance and health.",
```
camilamacedo86, line 68: Is it really working?
@dtfranz did you not need to apply the NP to scrape the metrics?

test/e2e/network_policy_test.go
```diff
@@ -0,0 +1,343 @@
+package e2e
```
camilamacedo86, line 1: This test suite does a good job verifying the structure of the `NetworkPolicy` definitions,That level of validation is helpful for maintaining consistency and catching configuration issues.

However, I think the tests could be improved by structuring them around specific, real-world use case scenarios. This would make the test intent clearer and more aligned with how these policies are expected to behave in practice.

---

### Why use scenario-based tests?

Instead of only validating the structure of the policies, we can write tests that describe the actual expected behavior in the cluster. This helps ensure not just that the policies exist, but that they are being enforced correctly.

Based on the current policies, here are the main scenarios we should be validating:

- Prometheus should be able to scrape metrics from catalogd (port 7443)
- The Kubernetes API server should be able to call catalogd’s webhook (port 9443)
- Operator-controller should be able to query catalogd (port 8443)
- Default deny should block traffic between pods that are not explicitly allowed

Each of these behaviors can be tested by deploying a pod, attempting a connection, and then removing the policy to confirm the connection is blocked.

---

### Example: Prometheus can scrape metrics from catalogd

Below is a complete example test that validates this behavior. It confirms that access works with the policy in place, and fails once the policy is removed.

```go
func TestPrometheusCanScrapeCatalogdMetrics(t *testing.T) {
	ctx := context.Background()
	client := utils.FindK8sClient(t)
	namespace := getComponentNamespace(t, client, catalogdManagerSelector)

	// Step 1: Create a curl test pod to simulate Prometheus
	testPod := buildCurlTestPod("prometheus-client", namespace)
	require.NoError(t, client.Create(ctx, testPod))
	waitForPodReady(t, client, namespace, "prometheus-client")

	targetURL := fmt.Sprintf("https://catalogd.%s.svc:%d", namespace, catalogdMetricsPort)

	t.Run("with_network_policy", func(t *testing.T) {
		output, err := execCurl(t, "prometheus-client", namespace, targetURL)
		require.NoError(t, err, "Expected access to be allowed with NetworkPolicy")
		assert.Contains(t, output, "metrics", "Expected metrics output from catalogd")
	})

	t.Run("after_removing_policy", func(t *testing.T) {
		// Remove the NetworkPolicy
		err := client.Delete(ctx, &networkingv1.NetworkPolicy{
				Name:      "catalogd-controller-manager",
				Namespace: namespace,
			},
		})
		require.NoError(t, err)
		time.Sleep(5 * time.Second) // Allow CNI time to apply changes

		_, err = execCurl(t, "prometheus-client", namespace, targetURL)
		assert.Error(t, err, "Expected access to be denied after policy removal")
	})
}

// Helper to build a test pod that can perform curl requests
func buildCurlTestPod(name, namespace string) *corev1.Pod {
	return &corev1.Pod{
			Name:      name,
			Namespace: namespace,
			Labels:    map[string]string{"app": name},
		},
		Spec: corev1.PodSpec{
			RestartPolicy: corev1.RestartPolicyNever,
			Containers: []corev1.Container{
				{
					Name:    "curl",
					Image:   "curlimages/curl",
					Command: []string{"sleep", "3600"},
				},
			},
		},
	}
}

// Helper to exec into a pod and run curl
func execCurl(t *testing.T, podName, namespace, url string) (string, error) {
	cmd := []string{"curl", "-k", "--max-time", "5", url}
	stdout, stderr, err := utils.ExecCommandInPod(podName, namespace, cmd)
	t.Logf("curl stderr: %s", stderr)
	return stdout, err
}
```

---

### Recommendation

Restructuring the tests around these scenarios would make the purpose of each test clearer, and also help validate that the policies are being enforced as intended. It would also make it easier for future reviewers and maintainers to understand why each rule exists.

Let me know what you think and if you'd need any help.

camilamacedo86, line 1: Hi @anik120,

This test does a good job checking that the `NetworkPolicy` definitions are great. I think that’s really helpful for making sure everything stays consistent and avoids mistakes.

That said, I think the test could be easier to follow if it were structured around real use cases — the actual things these policies are meant to allow or block.

Instead of only checking the structure, we could test what’s supposed to happen in the cluster. 

**Example**

- Prometheus should be able to scrape metrics from catalogd (port 7443)
- The Kubernetes API server should be able to call catalogd’s webhook (port 9443)
- Operator-controller should be able to query catalogd (port 8443)
- Pods without rules should be blocked by default deny

**Each of these could be tested by:**
1. Deploying a pod (like a curl container)
2. Trying the connection
3. Removing the policy
4. Confirming the connection fails

**Example: Prometheus can scrape catalogd metrics**

I mean we could do the tests, such as we do in other places, like the metrics: https://github.com/operator-framework/operator-controller/blob/main/test/e2e/metrics_test.go

Not as follows, but in this style-ish to illustrate only ( as a reference ) 

```go
func TestPrometheusCanScrapeCatalogdMetrics(t *testing.T) {
	ctx := context.Background()
	client := utils.FindK8sClient(t)
	namespace := getComponentNamespace(t, client, catalogdManagerSelector)

	// Create curl pod to simulate Prometheus
	testPod := buildCurlTestPod("prometheus-client", namespace)
	require.NoError(t, client.Create(ctx, testPod))
	waitForPodReady(t, client, namespace, "prometheus-client")

	targetURL := fmt.Sprintf("https://catalogd.%s.svc:%d", namespace, catalogdMetricsPort)

	t.Run("with_network_policy", func(t *testing.T) {
		output, err := execCurl(t, "prometheus-client", namespace, targetURL)
		require.NoError(t, err)
		assert.Contains(t, output, "metrics")
	})

	t.Run("after_removing_policy", func(t *testing.T) {
		err := client.Delete(ctx, &networkingv1.NetworkPolicy{
				Name:      "catalogd-controller-manager",
				Namespace: namespace,
			},
		})
		require.NoError(t, err)
		time.Sleep(5 * time.Second)

		_, err = execCurl(t, "prometheus-client", namespace, targetURL)
		assert.Error(t, err)
	})
}

func buildCurlTestPod(name, namespace string) *corev1.Pod {
	return &corev1.Pod{
			Name:      name,
			Namespace: namespace,
			Labels:    map[string]string{"app": name},
		},
		Spec: corev1.PodSpec{
			RestartPolicy: corev1.RestartPolicyNever,
			Containers: []corev1.Container{
				{
					Name:    "curl",
					Image:   "curlimages/curl",
					Command: []string{"sleep", "3600"},
				},
			},
		},
	}
}

func execCurl(t *testing.T, podName, namespace, url string) (string, error) {
	cmd := []string{"curl", "-k", "--max-time", "5", url}
	stdout, stderr, err := utils.ExecCommandInPod(podName, namespace, cmd)
	t.Logf("curl stderr: %s", stderr)
	return stdout, err
}
```

IHMO: writing the tests like this would make them easier to read and understand. It also clearly shows what behavior each policy is protecting.

WDYT?

camilamacedo86: @dtfranz, that is your expertise :-) , WDYT?

test/e2e/network_policy_test.go
dtfranz: I'm pretty sure it will work without the additional NP if prometheus lives in another namespace.

test/e2e/network_policy_test.go
dtfranz: I can definitely understand why you'd want to test that functionally, but it does also feel a little bit like testing k8s and not our own code 🫤 I think it's fine to just enforce the content of the policies. 

If for instance this were something like an RBAC definition, I think it would be unnecessary for us to spin up a pod just to make sure that the pod couldn't perform an action not allowed by the RBAC. Let me know if you don't think that's a fair comparison, though!

camilamacedo86: Just to clarify: (to ensure that we are all on the same page and I could properly share my thoughts). As a maintainer, what I would really like to see in the tests is:

- A clear definition of the **use cases** we want to cover.
- When should a request to these endpoints **work**, and when should it **fail**?
- A complete **end-to-end test** that covers these cases.

Right now, the tests are mostly checking the structure of the `NetworkPolicy` — which is useful — but they don’t verify whether the policies actually **work** or **behave as expected** in practice. (e2e tests)

On top of that, without the use case context, it becomes harder for us to understand what the rules are trying to achieve, and even harder to keep the test logic well maintained over time.

I think having the use cases clearly defined and tested end-to-end would make the tests much easier to follow and more reliable as the system evolves.

However, I am interested in your thoughts, and if you disagree with this case and prefer to proceed with the tests as they are, that's fine. However, it is unclear to me what the use cases are and whether we are properly covering them. I would appreciate it if a better clarification of those were at least added in the code

perdasilva: I totally get what you mean, and I've had similar concerns. I believe there have been some conversations around this decision already, which concluded with this approach. Maybe @anik120 can elucidate this a bit more . But, here's my 2c...

My bias with e2es is to _not_ let them become like with v0 (slow and bloated) by advocating (as much as possible) that we avoid testing the same thing multiple ways. With that in mind, I think the first three points are already implicitly tested by the e2es (with the possible exception of prometheus, which will be exercised by the benchmarking stuff that's coming). I.e. we stamp out a catalog (exercising the webhook) and create an extension (exercising the connection b/w o-c and the catalog service). The connection b/w the controllers and the kube apiserver is also implicitly tested by cache creation.

It could be worth adding a sanity check for the deny all, maybe ensuring the correct NP spec seems most sensible given that:
- it's hard to test deny all (i.e. we wouldn't be making sure that nothing could come in or out, but rather that an arbitrary pod couldn't reach to a particular host), and
- pod networking is not a core behavior of OLM (without the NP API we probably wouldn't be doing this)

I do get your reservations, and it does feel like we're testing implementation details instead of behavior. However,  the main risk I see is whether its possible for the correct NP spec to change our from under us (silently) in some future update of the NP API, which I think is highly unlikely (given API guarantees and so on).

These are the reasons I think this is one of those rare cases where its ok to test implementation and not behavior. It still does leave a bad taste tho...XD

anik120: Fyi updated the description as per our synchronous discussion about this thread: https://github.com/operator-framework/operator-controller/pull/2013#issue-3121587285

test/e2e/network_policy_test.go
anik120: @camilamacedo86 I did try that first but fyi `Port` is of type `*intstr.IntOrString` but `intstr.FromInt(Number)` returns a `intstr.IntOrString`, and you can't really do a `&intstr.FromInt(Number)`. Which is why I decided to store the `intstr.IntOrString` returned by `FromInt` in a variable, and then return the address of that variable from the helper function, so that that can be used for setting `Port`

anik120:
@joelanford @perdasilva regarding the discussion about possibly switching this to a unit test, I see the Quickstart target in the Makefile, and ack the discussion about using steps from that target (separated out) to generate the manifests and before starting the unit tests, and then consuming the generated manifests from the unit test. 

I don't see a natural home for the unit test though. Maybe that's actually a good signal that this should be a e2e test instead. 

Thoughts?

test/e2e/network_policy_test.go
camilamacedo86: Thank you for give attention to this suggestion and all clarifications.
Closing as sorted out .

test/e2e/network_policy_test.go
camilamacedo86: Hi @anik120 

We are using it here:

> port:          []networkingv1.NetworkPolicyPort{{Protocol: ptr.To(corev1.ProtocolTCP), Port: intOrStrPtr(catalogdMetricsPort)}},

We do not need to create a helper for this one. We can

```
&intstr.IntOrString{Type: intstr.Int, IntVal: PORT_NUMBER}
```

Like

```
port: []networkingv1.NetworkPolicyPort{{Protocol: ptr.To(corev1.ProtocolTCP), Port: &intstr.IntOrString{Type: intstr.Int, IntVal: catalogdMetricsPort}}},
```

anik120: Nice that works thank you for that.

camilamacedo86:
Hi @anik120 


I see that you asked for their input here as well; https://github.com/operator-framework/operator-controller/pull/2013#issuecomment-2963847723, so we can wait for their thoughts

However, IHMO:
- Yes, it would be better to do the tests by checking the YAML files directly, as discussed yesterday
- However, IMHO:

Yes — it would indeed be better to write the tests by checking the YAML files directly, as we discussed yesterday. That said, the current implementation is already acceptable. Also, considering our earlier conversation, we’ll likely have the YAML merged in the repo in the future, so an update would be necessary at that point anyway if we want to move toward the ideal approach.

So, IHMO: it’s really up to you. If you’re able to make the change now, that would be great 🥇 — but if not, I think this PR still accomplishes its goal effectively. From my side, LGTM.

anik120:
@camilamacedo86 PR #2025 gives me the manifests locally that I can consume and write the test with. 

However the question of "which package is a natural home for that test, and if we cannot think of a natural home package for that test is that a strong signal that this is better suited as an e2e test", still remains. 

Maybe we wait a little to see if anyone has a better idea, if not I'm happy with this PR and we can merge it in 👍🏽

tmshort:
Make sure you rebase...

perdasilva:
@anik120 I've pushed up a rebase to try and get this down if its green =D

perdasilva:
@anik120 I'll wait to approve until you give the final go ahead. I don't know if you want to make some changes still...

anik120:
Looks like there's no opposition to adding this test as an e2e, let's ship it 🚀

PR #2020, opened 2025-06-10
🌱 Update target and GitHub Action to run and upload all demos daily

camilamacedo86:
- Renamed Makefile target and GitHub Action to 'update-demos' for clarity
- Iterates over all *-demo-script.sh files and uploads casts
- Replaced CI trigger with a daily scheduled cron job at 03:00 UTC
- Retained workflow_dispatch for manual triggering

## Motivation

We've been adding a lot of demos lately — which is awesome! :tada:
To make sure we don't break them over time, it's important to keep them up to date.

However, running them in CI on every change would significantly increase build times.
We already had a GitHub Action in place to update the catalogd demo.

Therefore, this PR updates that logic to run all demos once a day instead, via a scheduled job — and renames the Makefile target accordingly.


perdasilva:
should we rename the other scripts to append _script?

camilamacedo86:
Hi @perdasilva 

> should we rename the other scripts to append _script?

I did that. If you have a better idea, please feel free to share.
But also, I think we can improve in the long run.

grokspawn:
Is this in response to an actual problem, or a perceived future one?  
The benefit of the current approach is that it prevents a PR which breaks a demo from merging. This would be delayed and possibly manual under the new regime.

camilamacedo86:
Hi @grokspawn 

> Is this in response to an actual problem, or a perceived future one?
The benefit of the current approach is that it prevents a PR which breaks a demo from merging. This would be delayed and possibly manual under the new regime.

Today **we don’t run all demos in CI—**just the ones listed in the catalogd catalog.

**We can run all of them via GitHub action in CI,** but over time that might slow things down or get expensive. **IMO agree it’s better to wait for real problems before adding more work. That is my preferred choice**

Personally, I’d keep running demos in CI only when related code changes. They kind of act like end-to-end tests, especially for feature-gated stuff where we don’t have full test coverage yet.

We moved to a cronjob because of @perdasilva’s concerns about CI time.

If @perdasilva is okay with it, I’m happy to move the demo call back to CI—and only switch to cronjob if it becomes a problem again. That feels like the best approach to me 👍

perdasilva:
@grokspawn yeah, I'm not in favor of running a 15 minute job (which will only grow) on a PR basis for something that seems more like stale documentation to me. Don't get me wrong, it matters. I'm just not sure we want to hold up PRs because of it, or run it with this level of frequency. We may want to hold a release back, but PRs, probably not. 

The unit tests and e2e suites should already give us pretty good signal that things are working as they should. I don't think we should use demos for that. In any case, the biggest vector for demo failures would probably be changes in API surface, which should occur fairly seldomly. The failure rate of the catalogd-demo GHA is extremely low (and in the failures I've seen, it's not even exactly clear that the RC was something wrong with the demo). 

Tbh, thinking a bit more about it, I'd start by making this a non-blocking step of the release pipeline and use fewer resources.

If we feel strongly about running this more frequently than once before release, maybe a good compromise would be to only run it when there are relevant manifest/API changes?

cc @camilamacedo86

perdasilva:
> should we rename the other scripts to append _script?

my bad if I missed it. I was looking in the output for the job and didn't see them (which doesn't mean they weren't there XDD) - thank you!!!

camilamacedo86:
Hi @grokspawn @perdasilva 

Thank you for your support and inputs 🎉 

> If we feel strongly about running this more frequently than once before release, maybe a good compromise would be to only run it when there are relevant manifest/API changes?

The way that we configure it here now is:
- If we change the action, add a new demo script, or modify any part of the hack/demo, we execute the action in the CI. 
- Otherwise, it will run once a day as a cron job. So, if it fails, we have the opportunity to identify the latest change and fix it as soon as possible. 
- Also, I add the paths when there are relevant manifest/API changes; we call it in the CI when we make changes there. 

I think it is a good compromise that can make everybody happy 👍 
Let me know wdyt?

grokspawn:
Sorry to necro here, but wanted to add that the paths are a really good approach which targets the scenarios I was concerned about, without impacting other merges.  Thanks for that!

PR #2024, opened 2025-06-12
📖 Metrics Docs Maintenance

dtfranz:
Updates the docs around metrics gathering to include necessary NetworkPolicy, and fixes some errors in the ServiceMonitor yaml for securityContext and catalogd labels.

# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/docs/draft/howto/consuming-metrics.md
+++ b/docs/draft/howto/consuming-metrics.md
@@ -6,7 +6,7 @@ The following procedure is provided as an example for testing purposes. Do not d
 
 In OLM v1, you can use the provided metrics with tools such as the [Prometheus Operator][prometheus-operator]. By default, Operator Controller and catalogd export metrics to the `/metrics` endpoint of each service.
 
-You must grant the necessary permissions to access the metrics by using [role-based access control (RBAC) polices][rbac-k8s-docs].
+You must grant the necessary permissions to access the metrics by using [role-based access control (RBAC) polices][rbac-k8s-docs]. You will also need to create a `NetworkPolicy` to allow egress traffic from your scraper pod, as the OLM namespace by default allows only `catalogd` and `operator-controller` to send and receive traffic.
 Because the metrics are exposed over HTTPS by default, you need valid certificates to use the metrics with services such as Prometheus.
 The following sections cover enabling metrics, validating access, and provide a reference of a `ServiceMonitor`
 to illustrate how you might integrate the metrics with the [Prometheus Operator][prometheus-operator] or other third-part solutions.
@@ -23,6 +23,25 @@ kubectl create clusterrolebinding operator-controller-metrics-binding \
    --serviceaccount=olmv1-system:operator-controller-controller-manager
 ```
 
+2. Next, create a `NetworkPolicy` to allow the scraper pods to send their scrape requests:
+
+```shell
+kubectl apply -f - << EOF
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: scraper-policy
+  namespace: olmv1-system
+spec:
+  podSelector:
+    matchLabels:
+      metrics: scraper
+  policyTypes:
+    - Egress
+  egress:
+    - {}  # Allows all egress traffic for metrics requests
+EOF
+```
 ### Validating Access Manually
 
 1. Generate a token for the service account and extract the required certificates:
@@ -41,6 +60,8 @@ kind: Pod
 metadata:
   name: curl-metrics
   namespace: olmv1-system
+  labels:
+    metrics: scraper
 spec:
   serviceAccountName: operator-controller-controller-manager
   containers:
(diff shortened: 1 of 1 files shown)


docs/draft/howto/consuming-metrics.md
```diff
@@ -6,7 +6,7 @@ The following procedure is provided as an example for testing purposes. Do not d
 
 In OLM v1, you can use the provided metrics with tools such as the [Prometheus Operator][prometheus-operator]. By default, Operator Controller and catalogd export metrics to the `/metrics` endpoint of each service.
 
-You must grant the necessary permissions to access the metrics by using [role-based access control (RBAC) polices][rbac-k8s-docs].
+You must grant the necessary permissions to access the metrics by using [role-based access control (RBAC) polices][rbac-k8s-docs]. You will also need to create a `NetworkPolicy` to allow egress traffic from your scraper pod, as the OLM namespace by default allows only `catalogd` and `operator-controller` to send and receive traffic.
```
camilamacedo86, line 9: I do not think that is required. 
See that the NPs that we have should allow already we scrap the metrics
Also, note that we are calling the metrics endpoint at: https://github.com/operator-framework/operator-controller/blob/main/test/e2e/metrics_test.go and we do not create any new NP

If we break it, then in the downstream we would no longer be able to get the metrics, and that is why we have a test to ensure it.

docs/draft/howto/consuming-metrics.md
```diff
@@ -41,6 +60,8 @@ kind: Pod
 metadata:
   name: curl-metrics
   namespace: olmv1-system
+  labels:
+    metrics: scraper
```
camilamacedo86, line 64: Why would we need the label?
see: https://github.com/operator-framework/operator-controller/blob/main/test/e2e/metrics_test.go#L129-L154
We do not need the label

docs/draft/howto/consuming-metrics.md
```diff
@@ -69,6 +90,9 @@ spec:
       secretName: olmv1-cert
   securityContext:
     runAsNonRoot: true
+    runAsUser: 65534
```
camilamacedo86:
```suggestion
    runAsUser: 1000
```
to be == our e2e tests

docs/draft/howto/consuming-metrics.md
```diff
@@ -131,6 +155,8 @@ kind: Pod
 metadata:
   name: curl-metrics-catalogd
   namespace: olmv1-system
+  labels:
+    metrics: scraper
```
camilamacedo86, line 155: we do not need either

docs/draft/howto/consuming-metrics.md
```diff
@@ -159,6 +185,9 @@ spec:
       secretName: $OLM_SECRET
   securityContext:
     runAsNonRoot: true
+    runAsUser: 65534
```
camilamacedo86:
```suggestion
    runAsUser: 1000
```

docs/draft/howto/consuming-metrics.md
```diff
@@ -253,7 +282,7 @@ metadata:
 spec:
   endpoints:
     - path: /metrics
-      port: https
+      port: metrics
```
camilamacedo86, line 277: We need to ensure that is == https://github.com/openshift/operator-framework-operator-controller/blob/0b44e8ce155a7c32660eeff6d862025c9d015880/openshift/catalogd/manifests/17-servicemonitor-openshift-catalogd-catalogd-metrics-monitor.yml#L11-L22

So 👍

docs/draft/howto/consuming-metrics.md
dtfranz: I just tried the guide again without the NetworkPolicy and it does not work. After I apply the NetworkPolicy, it works again.

The reason that the e2e test works is that it puts the curl pod into a random namespace, outside of olmv1-system. If you were to create the pod inside olmv1-system, the tests would fail.

dtfranz: Oh, and to your point on downstream metrics, the reason that also works fine is because the metrics scraper pod does not live in the same namespace as catalogd or operator-controller.

camilamacedo86: I see. 👍 
Thank you for the clarification

docs/draft/howto/consuming-metrics.md
dtfranz: Updated 👍

docs/draft/howto/consuming-metrics.md
dtfranz: Updated 👍

docs/draft/howto/consuming-metrics.md
camilamacedo86: I see: https://github.com/operator-framework/operator-controller/pull/2024/files#diff-8b584e069a03bf4a3bee2d6bdfb8d9414ab502a720f939dda7e80273e2973d4aR35-R38

The NP added is matching with the label 👍

PR #2034, opened 2025-06-23
🐛 (e2e) fix namespaces network-policy test searches in

anik120:
The test was previously listing networkpolicies to validate only in the namespace operator-controller is deployed in.
Upstream, that is valid since the namespace that operator-controller is deployed in, is the namespace that all other components are deployed in (and therefore all the network polices we expect to find).

A different distribution of olmv1 (downstream) could decide to deploy the policies in different namespaces, in which case the test will not look into all the namespaces it should look into.

This PR fixes the issue by searching in the namespaces both operator-controller and catalogd is deployed in.

# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/test/e2e/network_policy_test.go
+++ b/test/e2e/network_policy_test.go
@@ -140,14 +140,26 @@ func TestNetworkPolicyJustifications(t *testing.T) {
 	}
 
 	clientForComponent := utils.FindK8sClient(t)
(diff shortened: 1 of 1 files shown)


tmshort (commented):
/lgtm
/approve

PR #2037, opened 2025-06-23
🌱 Upgrade bingo deps 

camilamacedo86:


.bingo/controller-gen.mod
```diff
@@ -1,7 +1,7 @@
 module _ // Auto generated by https://github.com/bwplotka/bingo. DO NOT EDIT
 
-go 1.23.0
+go 1.24.0
```
grokspawn, line 3: I really wish that this is what our root go.mod resolved to as well (in terms of go/toolchain). 😞

tmshort:
Note that #2047 updates kind.

.bingo/controller-gen.mod
camilamacedo86: Not sure if I follow it
Do you think that is wrong?
Is it to be changed?
How?

tmshort (approved):
/lgtm
/spprove

.bingo/controller-gen.mod
grokspawn: Nothing to do here.  Just wishing that each of our components expressed a dependency on the 0th patchlevel for the golang version, and the toolchain (if needed) expressed the higher version for additional dependencies.

PR #2047, opened 2025-06-25
:bug: Update kindest/node image to v1.33.1

tmshort:
The Makefile calculates the version of kindest/node image to use. However, it appears that kindest/node:v1.33.0 has issues with some systems (e.g. Fedora).
Using kindest/node:v1.33.1 fixes this issue.

Also note that kind never used kindest/node:v1.33.0 as a default image, kind v0.28.0/v0.29.0 use kindest/node:v1.33.1.

# Description

## Reviewer Checklist

- [ ] API Go Documentation


Makefile
```diff
@@ -44,7 +44,12 @@ ENVTEST_VERSION := $(K8S_VERSION).x
 # The K8S_VERSION is set by getting the version of the k8s.io/client-go dependency from the go.mod
 # and sets major version to "1" and the patch version to "0". For example, a client-go version of v0.28.5
 # will map to a K8S_VERSION of 1.28.0
+ifeq ($(K8S_VERSION),1.33)
```
grokspawn: Why would we pin to the 0-th patch version of the aligned kube release?   That feels like a false dependency, where the real solution is to ensure that our kube/kind dependencies were always aligned (for e.g. as go-verdiff evals that we sync golang bumps, we should also bump kind when we bump kube). 

Please
/hold
while we discuss

camilamacedo86: **Clarification on KIND and node image versions:**

Just to clarify, **`kind` is the tool**, while the **node image refers to the Kubernetes version/image used**—these are distinct.

✅ The node image is **automatically updated based on the `go.mod`**, which is why we’re already using Kubernetes `v1.33`.  
So what you're asking about is actually already handled.

📌 We have a separate PR to update `kind` along with the other Bingo-managed tools: https://github.com/operator-framework/operator-controller/pull/2037  
We usually avoid combining all updates in one PR to keep reviews focused and manageable.

So, for me it is great if that solves the specific issue faced with Fedora.

/lgtm
/approved

grokspawn: Disagree. 
The node image is getting pinned here to the `$kube-version.0` (0 patch version) for the underlying kube version. 
This effectively prevents us from using any updated node images released by kind, unless they re-release them under the same patch version, which isn't sane. 
I think @joelanford participated in the initial discussion on this issue, and I think the argument had been that we would continue support for kube bump minor releases as long as we stayed on that kube version... but I admit that it's been a minute and my recollection might not be exact.  
But in this case we are overriding the defaults of a version of the tool that we deliberately chose, which feels a bit like needlessly fighting ourselves. 
I am not convinced that the expediency of this solution necessitates haste here.

trgeiger: I will just add that changing the kind node version from 1.33.0 to 1.33.1 makes Makefile targets work for me again on f42.

tmshort:
/retile :bug: Update kindest/node image to v1.33.1 via kind v0.29.0

trgeiger (approved):

Makefile
```diff
@@ -320,8 +314,9 @@ kind-deploy: manifests
 
 .PHONY: kind-cluster
 kind-cluster: $(KIND) #EXHELP Standup a kind cluster.
+	env K8S_VERSION=v$(K8S_VERSION) KIND=$(KIND) GOBIN=$(GOBIN) hack/tools/validate_kindest_node.sh
```
grokspawn, line 317: Didn't get to this yesterday, but would've preferred to keep this check separate from the cluster creation step, so we could include it in the verify target.

grokspawn: Followed up with https://github.com/operator-framework/operator-controller/pull/2052

PR #2010, opened 2025-06-04
:seedling: migrate from github.com/everettraven/crd-diff to sigs.k8s.io/crdify

everettraven:
# Description

This updates the usage of `github.com/everettraven/crd-diff` in CI to `sigs.k8s.io/crdify` as the project has now been donated under the umbrella of sig-api-machinery.

## Reviewer Checklist

- [ ] API Go Documentation
- [ ] Tests: Unit Tests (and E2E Tests, if appropriate)
- [ ] Comprehensive Commit Messages
- [ ] Links to related GitHub Issue(s)

PR #2054, opened 2025-06-26
✨ (deps): import kubernetes-sigs/crdify for performing CRD upgrade safety checks

everettraven:
# Description

Removes all custom CRD upgrade pre-flight check validations in favor of importing validation logic from [kubernetes-sigs/crdify](https://github.com/kubernetes-sigs/crdify).

This allows OLM to stay consistent with what is deemed "safe" by an more formal library that performs validation checks. As was the previous behavior, OLM configures the validation to be as strict as reasonably possible and to continue failing closed on unhandled changes.

While it is still possible to implement custom validations with this approach (by adding them to the registry), it would be awesome to see OLM maintainers get involved in discussions with the maintainers of kubernetes-sigs/crdify and contribute new validations there first and pull them in through dependency updates.

## Reviewer Checklist

- [ ] API Go Documentation


go.mod
```diff
@@ -44,6 +44,7 @@ require (
 	k8s.io/utils v0.0.0-20250604170112-4c0f3b243397
 	sigs.k8s.io/controller-runtime v0.21.0
 	sigs.k8s.io/controller-tools v0.18.0
+	sigs.k8s.io/crdify v0.4.0
```
everettraven: Note: I believe that this particular import version removes the behavior added in #2023 

https://github.com/kubernetes-sigs/crdify/pull/16 merged recently so pulling the latest commit in should fix this

everettraven: I pulled in the latest commit in the latest push, but I'll need to update the description change validation configuration to be set to `Ignore` because it was decided in #2023 that OLM should not block on description changes.

tmshort:
/approve
Looks OK to me, but I'm not too familiar with the code in question.

The primary purpose seems to replace github.com/openshift/crd-schema-checker with k8s-sigs/crdify?

grokspawn:
Hi @everettraven !
This looks like it supplants #2010. Confirm/deny?

everettraven:
> Hi @everettraven ! This looks like it supplants #2010. Confirm/deny?

Deny. #2010 is for the check in CI that runs against the OLMv1 APIs. This PR imports it for use _within_ OLMv1 for OLMv1's CRD upgrade safety check validations

everettraven:
> The primary purpose seems to replace github.com/openshift/crd-schema-checker with k8s-sigs/crdify?

@tmshort The removal of `github.com/openshift/crd-schema-checker` is a side effect of removing the internal validation checks that we had that imported that library. `k8s-sigs/crdify` is a new sig-api-machinery subproject that got created to centralize the work that a few folks were doing related to CRD schema change validations.

The primary purpose is to have OLM stop using its own checks and instead use this centralized place where there is active development and discussion with sig-api-machinery folks on what rules should be codified when evaluating CRD schema changes for compatibility.

tmshort:
Clicking the button to override codecov/project...

PR #2056, opened 2025-06-26
:sparkles: Add standard-e2e manifest

tmshort:
This renames two overlays:
* cert-manager -> standard
* e2e -> standard-e2e

This also adds a new manifest:
* manifests/standard-e2e.yaml

This fixes the issue with the current test-e2e modifying the standard manifest with e2e parameters, leaving a dirty workspace. This changes the manifest used by the e2e test to be it's own standard-e2e manifest.

This is part of the feature-gated API functionality.

# Description

## Reviewer Checklist

- [ ] API Go Documentation


grokspawn:
Had a side-channel convo with @tmshort about possible impacts to plans to open up the PKI model to accept other providers (for instance, currently we rely very heavily on certmanager, but others could be used like openshift/serviceca). 

It sounds like we need to resolve that question separately, and that any change to that would inescapably require new changes to this area anyway, so no opportunity to "leave room" for it here. 

/lgtm

tmshort:
@grokspawn openshift's serviceca is already handled by separate downstream overlays; so it certainly is possible. By keeping the cert-manager items separate (they are in the `component/tls` directory), they are easily replaceable.

PR #2060, opened 2025-06-27
:seedling: Rename tls component to cert-manager and consolidate

tmshort:
This renames the tls component to cert-manager to reflect what it really is. A kustomization.yaml file is added to cert-manager to make it behave as a single component. This will make it easier to manage this component, and allow for easier future replacement.

This is part of the feature-gated API functionality.

# Description

## Reviewer Checklist

- [ ] API Go Documentation


config/overlays/standard-e2e/kustomization.yaml
```diff
@@ -7,8 +7,6 @@ resources:
 - ../../base/operator-controller
 - ../../base/common
 components:
-- ../../components/tls/catalogd
-- ../../components/tls/operator-controller
 - ../../components/e2e
-# ca must be last or other components will overwrite the namespaces
-- ../../components/tls/ca
+# This must be last due to namespace overwrite issues of the ca
+- ../../components/cert-manager
```
camilamacedo86, line 12: That make sense if we want add the OCP CA in upstream as well.
I understood that we were looking for that so we could have another componet for it too

/lgtm

tmshort: Yes, that's the idea, if we ever want to swap certificate providers, this will make it easier.
I may try to update this to avoid this weird namespace mechanism... @joelanford had something in older refactoring PR.

PR #2073, opened 2025-07-03
:sparkles: Add feature-set annotation to all manifest resources

tmshort:
The feature-set annotation was originally applied by the crd-generator to the generated CRDs. Now, that annotaion is being added by the overlay, and is being added to all resources in the manifest.

Each manifest has an annotation value based its name.

The crd-generator is now adding a generated annotation indicating what the CRD was generated for.

# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/config/base/catalogd/crd/experimental/olm.operatorframework.io_clustercatalogs.yaml
+++ b/config/base/catalogd/crd/experimental/olm.operatorframework.io_clustercatalogs.yaml
@@ -4,7 +4,7 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.18.0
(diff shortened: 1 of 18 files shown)


tmshort:
The **go-apidiff** issue with with the crd-generator code, and is innocuous.

camilamacedo86:
We did not release, and that is internal, so we can ignore go-apidiff
Also, it is in the hack.

But @tmshort, could we please move the GoLang code in the hack to an internal directory to avoid it? 
That would ensure safely that we do not export this staff.

I will not consider it a block but seems very nice to have 
/lgtm

tmshort:
I'm not sure moving it to "internal" will help, as it's not part of the operator-controller or catalogd, since it's a separate tool. All other such tools are located in hack.

PR #2076, opened 2025-07-07
:sparkles: Add NamespaceSelector to generated webhook configs

perdasilva:
# Description

Sets the NamespaceSelector in generated Validating and MutatingWebhookConfiguration resources. This ensures that webhooks for operators installed in single/own namespace mode are (at least for namespaced APIs) restricted to the target namespace of the installation.

The implementation diverges from that of OLMv0. In OLM v0 the namespace selector is derived from OperatorGroup labels OLM sets on the target namespaces. In OLMv1 we don't have OperatorGroups and there's probably no need to add any labels to target namespaces. So, we use the kubernetes provided "kubernetes.io/metadata.name" label. This may not work in older versions of kubernetes.

Note: the renderer itself supports rendering the bundle manifests for multi-namespace mode. OLMv1 just doesn't support MultiNamespace mode. Therefore, the namespace selector uses a match expression with the `In` operator.

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/operator-controller/rukpak/render/registryv1/generators/generators.go
+++ b/internal/operator-controller/rukpak/render/registryv1/generators/generators.go
@@ -13,6 +13,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/utils/ptr"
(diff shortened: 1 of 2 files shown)


internal/operator-controller/rukpak/render/registryv1/generators/generators.go
```diff
@@ -29,6 +30,8 @@ import (
 const (
 	tlsCrtPath = "tls.crt"
 	tlsKeyPath = "tls.key"
+
+	labelKubernetesNamespaceMetadataName = "kubernetes.io/metadata.name"
```
tmshort, line 34: Is this right? Should it be `metadata.namespace`? Because this ought to be the *name* of the resource, not the namespace?

tmshort: (Unless you're looking at namespaces?)

perdasilva: I think so. We're selecting for the name of namespace. In more recent kube versions, the label "kubernetes.io/metadata.name" started to be stamped out to allow for label selection on the resource name. So, if targetNamespaces = ["ns1"], it will limit the webhook to namespaces with label "kubernetes.io/metadata.name" in ["ns1"] => the namespace with name ns1. Or, am I misunderstanding your question?

tmshort: I think i may have been confused, but you're looking at namespace resources for this selector, so I think we're good.

PR #2078, opened 2025-07-07
✨ Add Network Policy to e2e test bundles

trgeiger:
Adds a Netwok Policy to the end-to-end test bundles and adds a check to the tests that the Network Policy resources are created. Closes OPRUN-3967

# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/hack/test/pre-upgrade-setup.sh
+++ b/hack/test/pre-upgrade-setup.sh
@@ -103,6 +103,16 @@ rules:
     - "bind"
     - "escalate"
   - apiGroups:
+    - networking.k8s.io
+    resources:
+    - networkpolicies
+    verbs:
+    - get
+    - list
+    - create
+    - update
+    - delete
+  - apiGroups:
     - "olm.operatorframework.io"
     resources:
     - "clusterextensions/finalizers"
--- a/testdata/images/bundles/test-operator/v1.0.0/manifests/testoperator.clusterserviceversion.yaml
+++ b/testdata/images/bundles/test-operator/v1.0.0/manifests/testoperator.clusterserviceversion.yaml
@@ -97,6 +97,16 @@ spec:
           - patch
           - delete
         - apiGroups:
+          - networking.k8s.io
+          resources:
+          - networkpolicies
+          verbs:
+          - get
+          - list
+          - create
+          - update
+          - delete
+        - apiGroups:
           - coordination.k8s.io
           resources:
           - leases
--- a/testdata/images/bundles/test-operator/v1.0.0/manifests/testoperator.networkpolicy.yaml
+++ b/testdata/images/bundles/test-operator/v1.0.0/manifests/testoperator.networkpolicy.yaml
@@ -1,0 +1,8 @@
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: test-operator-network-policy
+spec:
+  podSelector: {}
+  policyTypes:
+  - Ingress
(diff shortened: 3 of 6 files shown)


testdata/images/bundles/test-operator/v2.0.0/manifests/testoperator.networkpolicy.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: test-operator-network-policy
+spec:
+  podSelector: {}
+  policyTypes:
+  - Ingress
```
camilamacedo86: Could you please ensure that we have a line at the end?
We should not see the above simble

testdata/images/bundles/test-operator/v1.0.0/manifests/testoperator.clusterserviceversion.yaml
```diff
@@ -96,6 +96,16 @@ spec:
...
+        - apiGroups:
+          - networking.k8s.io
+          resources:
+          - networkpolicies
+          verbs:
+          - get
+          - list
+          - create
+          - update
+          - delete
```
camilamacedo86, line 108: @perdasilva for example here, if we use the tools to generate the sample to ensure that we have a mock with networkpolicy we should just enable it out: https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v4/config/default/kustomization.yaml#L30-L34

Therefore, our automation would only uncomment this line after the scaffold and run the make targets to ensure that the manifests were generated correctly.

testdata/images/bundles/test-operator/v1.0.0/manifests/testoperator.networkpolicy.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: test-operator-network-policy
+spec:
+  podSelector: {}
+  policyTypes:
+  - Ingress
```
camilamacedo86: NIT: Could we add space as well

testdata/images/bundles/test-operator/v1.0.0/manifests/testoperator.clusterserviceversion.yaml
perdasilva: I get that. But it only works for supported flags. Once we are out of the happy path, we need to start doing esoteric things.

hack/test/pre-upgrade-setup.sh
```diff
@@ -102,6 +102,16 @@ rules:
...
+  - apiGroups:
+    - networking.k8s.io
+    resources:
+    - networkpolicies
+    verbs:
+    - get
+    - list
+    - create
+    - update
+    - delete
```
camilamacedo86, line 114: Sorry, now I see that we have the NP in both sample versions
Can we have it in one version only, so we ensure that it works with both cases? (with and without NP)

trgeiger: Did you not mean to approve it? It already merged from your review. Also, not having the NP in one version isn't going to be any different, we always want to be able to support installing NPs so we always want the e2e to try to install one so the CI fails if it can't. There's no value-add from running the same test without an NP.

camilamacedo86: It was a nit idea, I think we may could cover both scenarios 
But it is all fine. I think it is OK as it is no worry.

camilamacedo86: Btw, thank you a lot for taking care of that !!!

PR #2081, opened 2025-07-08
✨ Performance Alerting

### LINKED ISSUE #1904: Create/Modify upstream CI job
To enforce our performance goals, we need to create a CI job or modify an existing one which enforces certain performance thresholds.

The e2e job in the github CI is a good first candidate. It allows us to avoid creating another CI job and use the near-real-world runtime environment to look for performance issues. If we find that the overhead of having prometheus running on our e2e tests is too high, we can instead add a new CI job and perhaps make it periodic.

### LINKED ISSUE #1905: Select detection criteria for CI failure of e2e metrics job
Look through the data made available from our scraped metrics during e2e, and select criteria that may qualify as CI blocking.

For now, we need only ensure that the metrics collected during the e2e run do not contain any alerts or critical metrics. A critical metric could be, for instance, a count of controller panics greater than zero - in other words, unhandled events that should not happen under any circumstances. These are the kinds of things we want to be looking for.

dtfranz:
# Description

Introduces an early-warning series of prometheus alerts to attempt to catch issues with performance at an early stage in development.

As the e2e tests run, the installed prometheus instance is scraping metrics from catalogd and operator-controller, and will fire alerts based on rules introduced in this PR. Since we're running these tests on the github runners which do not have consistent performance, our alerts must be based on platform-independent metrics and are therefore limited. Any other ideas for metrics to check on this PR are appreciated!

Once the e2e tests finish, prometheus is queried for active alerts. Any alerts found in `pending` state will result in a warning being set on the e2e workflow. Any alerts in `firing` state will give an error. These errors do not (at the moment) fail the run, but are visible when the workflow details are viewed.

For instance:
```
Prometheus Alert Pending
operator-controller-memory-growth: operator-controller pod memory usage growing at a high rate for 5 minutes: 72.86kB/sec
```

I am not making this a required check until we have a pretty good idea of an approximate baseline.

Potential Enhancements:
* Additional alerts, if any
* Fine-tune the alerts and fail runs when they fire
* ~~Remove yaml from script and organize into an additional kustomization component~~ Done
* Output metrics as a mermaid XY plot in the workflow summary

Closes #1904 
Closes #1905 

## Reviewer Checklist

- [ ] API Go Documentation


Makefile
```diff
@@ -285,11 +285,9 @@ prometheus: #EXHELP Deploy Prometheus into specified namespace
...
 # and provide alerts from the metrics based on the rules that we set.
 .PHONY: e2e-metrics
-e2e-metrics: #EXHELP Request metrics from prometheus; place in ARTIFACT_PATH if set
-	curl -X POST \
-	-H "Content-Type: application/x-www-form-urlencoded" \
-	--data 'query={pod=~"operator-controller-controller-manager-.*|catalogd-controller-manager-.*"}' \
-	http://localhost:30900/api/v1/query > $(if $(ARTIFACT_PATH),$(ARTIFACT_PATH),.)/metrics.out
+e2e-metrics: #EXHELP Request metrics from prometheus; select only actively firing alerts; place in ARTIFACT_PATH if set
+	ALERTS_FILE_PATH=$(if $(ARTIFACT_PATH),$(ARTIFACT_PATH),.)/alerts.out
+	curl -X GET http://localhost:30900/api/v1/alerts | jq 'if (.data.alerts | length) > 0 then .data.alerts.[] else empty end' > $(ALERTS_FILE_PATH)
```
tmshort: This may be complaining because ALERTS_FILE_PATH is indented, so it's not considered a make variable. So you need braces `{}` (and some more sh magic) or unindent the line and use `:=`

dtfranz: Ah whoops, yes I actually had this line duplicated where the other line was doing as you suggested, and I removed the wrong one right before I pushed 🤦

dtfranz: Fixed, thanks for pointing that out 💜

trgeiger:
I think it looks good but does it really close #1905? I would think that issue is more specific to a future iteration of this feature where we do make the job fail if it hits certain thresholds.

trgeiger:
One other thing is I don't have context for the thresholds you chose in the alerts. I see you mention that you don't want this to be required until we have a firmer idea of good baselines--are the current ones based on some previous work or did you just pick some decent-seeming thresholds for all the checks based on your experience? Do we need to queue up additional work to fine-tune these?

dtfranz:
Thanks for taking a look @trgeiger ! 

For your first point, I agree that the issue definitely indicates that we should fail the CI but I'm hesitant to do that at the moment without larger group buy-in. I'm happy to keep the issue open and close it after we turn on ci blocking, or close it with this PR and track a follow-up issue. As long as it's tracked I'm happy either way.

On your second point, these values are based on my experience running the workflow many times over and checking the metrics. Up till this point, nobody (to my knowledge) has run a more thorough study of v1 performance, not counting @jianzhangbjz and his work on the downstream version of this. These changes will enable to us to quickly get a better understanding and make any necessary adjustments.

jianzhangbjz:
Yeah, I haven’t collected the data for the OLMv1 performance baseline yet. I’m planning to reuse https://github.com/cloud-bulldozer/orion to help identify performance issues. The current metrics are being discussed on Slack: [link](https://redhat-internal.slack.com/archives/C06KP34REFJ/p1751444567049439), and progress is being tracked here: https://issues.redhat.com/browse/OCPQE-28161

trgeiger:
Cool, that's exactly what I wanted to know re: thresholds. And as for the issue tracking, either solution works--I just wanted to make sure the next iteration was tracked, as you stated. Keeping that issue or opening a new one both sound good to me.

trgeiger (approved):

config/prometheus/prometheus_rule.yaml
```diff
@@ -0,0 +1,59 @@
...
+      annotations:
+        description: "controller webhook of pod {{ $labels.pod }} experienced panic(s); count={{ $value }}"
+  - name: resource-usage
+    rules:
+    - alert: oom-events
+      expr: container_oom_events_total > 0
+      annotations:
+        description: "container {{ $labels.container }} of pod {{ $labels.pod }} experienced OOM event(s); count={{ $value }}"
+    - alert: operator-controller-memory-growth
+      expr: deriv(sum(container_memory_working_set_bytes{pod=~"operator-controller.*",container="manager"})[5m:]) > 50_000
```
camilamacedo86: @dtfranz so we are manually defining the trashholders here? 
Could we doc how it works in the https://github.com/operator-framework/operator-controller/blob/main/docs/contribute/developer.md ? WDYT?
Not a blocker for this one for sure

Makefile
```diff
@@ -277,19 +277,23 @@ test-experimental-e2e: run image-registry prometheus experimental-e2e e2e e2e-me
...
+	trap 'echo "Cleaning up $(TMPDIR)"; rm -rf "$(TMPDIR)"' EXIT; \
+	curl -s "https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/refs/tags/$(PROMETHEUS_VERSION)/kustomization.yaml" > "$(TMPDIR)/kustomization.yaml"; \
+	curl -s "https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/refs/tags/$(PROMETHEUS_VERSION)/bundle.yaml" > "$(TMPDIR)/bundle.yaml"; \
+	(cd $(TMPDIR) && $(KUSTOMIZE) edit set namespace $(PROMETHEUS_NAMESPACE)) && kubectl create -k "$(TMPDIR)"
+	kubectl wait --for=condition=Ready pods -n $(PROMETHEUS_NAMESPACE) -l app.kubernetes.io/name=prometheus-operator
+	$(KUSTOMIZE) build config/prometheus | CATALOGD_SERVICE_CERT=$(shell kubectl get certificate -n olmv1-system catalogd-service-cert -o jsonpath={.spec.secretName}) envsubst '$$CATALOGD_SERVICE_CERT' | kubectl apply -f -
+	kubectl wait --for=condition=Ready pods -n $(PROMETHEUS_NAMESPACE) -l app.kubernetes.io/name=prometheus-operator --timeout=60s
+	kubectl wait --for=create pods -n $(PROMETHEUS_NAMESPACE) prometheus-prometheus-0 --timeout=60s
+	kubectl wait --for=condition=Ready pods -n $(PROMETHEUS_NAMESPACE) prometheus-prometheus-0 --timeout=120s
```
camilamacedo86, line 290: Wouldn't it be better to centralise the Prometheus installation and related configurations in the hack directory? It might help keep things more organised and easier to understand

camilamacedo86, line 290: Wouldn't it be better to centralise the Prometheus installation and related configurations in the hack directory? It might help keep things more organised and easier to understand.

.github/workflows/e2e.yaml
tmshort, line 1: This file has a number of trailing whitespaces you should clean up.

Makefile
```diff
@@ -277,19 +277,23 @@ test-experimental-e2e: run image-registry prometheus experimental-e2e e2e e2e-me
...
+
+# The output alerts.out file contains any alerts, pending or firing, collected during a test run in json format.
```
tmshort: This line has trailing whitespace

tmshort:
/hold
I'm going to ask you to move `config/prometheus` into `config/overlays/prometheus` or to make it a component that is built as part of the e2e manifests. This later option means moving the files into `config/components/e2e/prometheus` and then including into the `config/components/e2e/kustomization.yaml` file. However, this may be tricky to do until #2088 is done. 
My preference would be for this to all be included as part of the e2e manifests, rather than something done separately, but again, it might have to wait until #2088 is done.

Makefile
tmshort: Actually, I would prefer it to be part of the existing e2e manifests, since this is something we are planning to do for our e2e's.

Makefile
```diff
@@ -277,19 +277,23 @@ test-experimental-e2e: run image-registry prometheus experimental-e2e e2e e2e-me
...
-# The metrics.out file contains raw json data of the metrics collected during a test run.
-# In an upcoming PR, this query will be replaced with one that checks for alerts from
-# prometheus. Prometheus will gather metrics we currently query for over the test run, 
-# and provide alerts from the metrics based on the rules that we set.
+	trap 'echo "Cleaning up $(TMPDIR)"; rm -rf "$(TMPDIR)"' EXIT; \
+	curl -s "https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/refs/tags/$(PROMETHEUS_VERSION)/kustomization.yaml" > "$(TMPDIR)/kustomization.yaml"; \
+	curl -s "https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/refs/tags/$(PROMETHEUS_VERSION)/bundle.yaml" > "$(TMPDIR)/bundle.yaml"; \
+	(cd $(TMPDIR) && $(KUSTOMIZE) edit set namespace $(PROMETHEUS_NAMESPACE)) && kubectl create -k "$(TMPDIR)"
+	kubectl wait --for=condition=Ready pods -n $(PROMETHEUS_NAMESPACE) -l app.kubernetes.io/name=prometheus-operator
+	$(KUSTOMIZE) build config/prometheus | CATALOGD_SERVICE_CERT=$(shell kubectl get certificate -n olmv1-system catalogd-service-cert -o jsonpath={.spec.secretName}) envsubst '$$CATALOGD_SERVICE_CERT' | kubectl apply -f -
```
tmshort: The name of this secret ought to be fixed, so you shouldn't have to extract it?

config/prometheus/prometheus_rule.yaml
dtfranz: Are you talking about adding notes to explain these specific queries? I'm happy to do that somewhere if so, but if you mean explaining how to adjust/create rules I'd prefer to link to the official prometheus docs.

Makefile
dtfranz: ~~This secret is generated by certmanager at runtime, after installation, so it can't be predetermined (unless you know of a way).~~ EDIT: I've noticed the name does seem to stay as `catalogd-service-cert-v1.3.0-68-g7cd03f1-dirty` for me, but without knowing how exactly that name is generated I'd rather not set it as such here.

Makefile
dtfranz: The script was getting way too big, and only had one or two operations that justified it as a script. The prometheus yaml is way more readable and maintainable in a kustomization manifest, IMO.

@tmshort I initially wanted to add these manifests to the e2e collection as you mentioned, but the catalogd certificate generated by certmanager is named `catalogd-service-cert-v1.3.0-68-g7cd03f1-dirty` (at least, for me), which doesn't give me confidence that I can just hard-code it and hope it keeps working. Unless you know what exactly that name comes from?

dtfranz: Also @tmshort if this sufficiently explains why I can't do as you mentioned [here](https://github.com/operator-framework/operator-controller/pull/2081#issuecomment-3063649513), I'd appreciate if we could drop the hold, but if you can think of a way around the issue I'd more than happy to give it a try!

.github/workflows/e2e.yaml
dtfranz: Fixed

Makefile
dtfranz: Fixed

Makefile
dtfranz: NVM, I found it, I'll try and work that in.

dtfranz: OK so it's possible to do the same `sed "s/cert-git-version/cert-$(VERSION)/g"` for these manifests, but the next issue we face putting the manifests in `config/components/e2e` is that it's not possible to put the prometheus CRDs and CRs into the same manifest that gets installed by `install.tpl.sh`, otherwise you get `no matches for kind "Prometheus"` errors. The prometheus-operator install and metrics yaml need to happen at different steps. That probably means adding additional content to the `install.tpl.sh` script.

Unless you know a better way, I'm thinking I could add this to `install.tpl.sh`:
```
if [[ -n "$prometheus_version" ]]; then
  trap 'echo "Cleaning up $(TMPDIR)"; rm -rf "$(TMPDIR)"' EXIT; \
  curl -s "https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/refs/tags/$(PROMETHEUS_VERSION)/kustomization.yaml" > "$(TMPDIR)/kustomization.yaml"; \
  curl -s "https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/refs/tags/$(PROMETHEUS_VERSION)/bundle.yaml" > "$(TMPDIR)/bundle.yaml"; \
  (cd $(TMPDIR) && $(KUSTOMIZE) edit set namespace $(PROMETHEUS_NAMESPACE)) && kubectl create -k "$(TMPDIR)"
  kubectl wait --for=condition=Ready pods -n $(PROMETHEUS_NAMESPACE) -l app.kubernetes.io/name=prometheus-operator
fi
```

...then have the Makefile set the variable in `test-e2e` here:
```
test-e2e: PROMETHEUS_VERSION := v0.83.0
```

I kind of hesitate to add anything to the install script that's for test only, though 🫤

Makefile
dtfranz: Figured out the secret name stuff but ran into other issues; see [here](https://github.com/operator-framework/operator-controller/pull/2081#discussion_r2203977577)

Makefile
tmshort: That name is based on `$(VERSION)` as you point out, and it's for rotation purposes, and I think it's not something we ought to be doing, but it is what it is.

I agree, the install script should not have test stuff in it. It *will* cause problems for users, as they don't necessarily have access to the kustomize configuration files.

tmshort:
At this point, due to the $(VERISON) naming issue with certs, I think that putting this as part of the e2e manifest is not realistic *now*. Perhaps if we remove that from the name of the certificate, we can re-look at that.

My only hold objection is to move `config/prometheus` into `config/overlays/prometheus` and to update the `config/README.md` file to indicate the use of this manifest.

dtfranz:
> At this point, due to the $(VERISON) naming issue with certs, I think that putting this as part of the e2e manifest is not realistic _now_. Perhaps if we remove that from the name of the certificate, we can re-look at that.
> 
> My only hold objection is to move `config/prometheus` into `config/overlays/prometheus` and to update the `config/README.md` file to indicate the use of this manifest.

I'm a little bit confused by your suggestion here. 

Are you asking that I create a separate overlay for prometheus which includes nothing except for the prometheus manifests? If so, I'm not sure I see the utility of that. It would also break the convention of that folder, since the rest of those folders are meant to be applied on top of the standard manifests to create the monolithic manifest files in `manifests/`. We would also gain nothing in terms of Makefile simplicity since prometheus still needs to be installed in two steps (unless we add `helm` to our install tools...).

Or, are you suggesting that I create an additional overlay which adds the prometheus CRDs and CRs on top of the standard manifests, and puts them all into the big manifest file? If so, we will still run into the problem I mentioned [here](https://github.com/operator-framework/operator-controller/pull/2081#discussion_r2203977577), since the CRDs and CRs cannot be applied from the same file in the same step. It also causes the installation to run into [this issue](https://github.com/prometheus-operator/prometheus-operator/issues/4355), since `install.tpl.sh` uses `kubectl apply` instead of `create`, and if we switch to `create` we lose the ability to run `make run` subsequent times for updates without a complete uninstall of v1. We can get around that by doing SSA, but again we'd be changing the install script for our tests (though adding SSA is probably less of a big deal, admittedly, but we still run into the CRD+CR issue).

Unless I'm still misunderstanding something, or unless you can think of solutions to the above, would you mind dropping your hold?

tmshort:
>I'm a little bit confused by your suggestion here.

I'm asking you to move the directory `config/prometheus` to be `config/overlays/prometheus` and update any references.

So, it's not a *new* overlay, it's just moving the existing one. 

The reason is that the root `config` directory does not contain any valid kustomizations _inside it_ (except samples). I want to keep the `config` directory as clean as possible.

Makefile
camilamacedo86: Could we change it for something like:

Makefile

```
.PHONY: prometheus
PROMETHEUS_NAMESPACE := olmv1-system
PROMETHEUS_VERSION := v0.83.0

prometheus:
	./hack/promotheus/deploy-prometheus.sh $(PROMETHEUS_NAMESPACE) $(PROMETHEUS_VERSION)
```

```
#!/bin/bash

set -euo pipefail

PROMETHEUS_NAMESPACE="$1"
PROMETHEUS_VERSION="$2"
KUSTOMIZE="${KUSTOMIZE:-kustomize}"
VERSION="${VERSION:-latest}"
TMPDIR="$(mktemp -d)"

trap 'echo "Cleaning up $TMPDIR"; rm -rf "$TMPDIR"' EXIT

echo "Downloading Prometheus resources..."
curl -s "https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/refs/tags/${PROMETHEUS_VERSION}/kustomization.yaml" > "${TMPDIR}/kustomization.yaml"
curl -s "https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/refs/tags/${PROMETHEUS_VERSION}/bundle.yaml" > "${TMPDIR}/bundle.yaml"

echo "Patching namespace to ${PROMETHEUS_NAMESPACE}..."
(cd "$TMPDIR" && $KUSTOMIZE edit set namespace "$PROMETHEUS_NAMESPACE")

echo "Applying Prometheus base..."
kubectl create -k "$TMPDIR"

echo "Waiting for Prometheus Operator pods to become ready..."
kubectl wait --for=condition=Ready pods -n "$PROMETHEUS_NAMESPACE" -l app.kubernetes.io/name=prometheus-operator

echo "Applying overlay config..."
$KUSTOMIZE build config/overlays/prometheus | sed "s/cert-git-version/cert-${VERSION}/g" | kubectl apply -f -

echo "Rechecking readiness..."
kubectl wait --for=condition=Ready pods -n "$PROMETHEUS_NAMESPACE" -l app.kubernetes.io/name=prometheus-operator --timeout=60s
kubectl wait --for=create pods -n "$PROMETHEUS_NAMESPACE" prometheus-prometheus-0 --timeout=60s
kubectl wait --for=condition=Ready pods -n "$PROMETHEUS_NAMESPACE" prometheus-prometheus-0 --timeout=120s

echo "Prometheus deployment completed successfully."
```

There is too much in the Makefile, which is starting to become hard for us to understand and maintain.
IHMO the logic to install and check would better fit in a shell

config/README.md
```diff
@@ -27,6 +27,12 @@ This provides additional configuration support for end-to-end testing, including
 
 This configuration is used to generate `manifests/standard-e2e.yaml`.
 
+## config/overlays/prometheus
+
+Overlay containing manifest files which enable prometheus scraping of the catalogd and operator-controller pods. Used during e2e runs to measure performance over the lifetime of the test. 
+
+These manifests will not end up in the `manifests/` folder, as they must be applied in two distinct steps to avoid issues with applying prometheus CRDs and CRs simultaneously.
+
```
camilamacedo86, line 35: Could we add here something like:

Note that the values defined for the performance checks are fixed and defined in : config/overlays/prometheus/prometheus_rule.yaml

camilamacedo86: I am just trying to have it write down in some place to make easier understand it if we need afterwords

camilamacedo86 (approved):
As we spoke this, and to make things easier, I’m OK with going incrementally here. We can keep the scope smaller for now and make improvements in a follow-up to this PR.

1. I think we need a shell script to deploy and verify the Prometheus installation.  
   I shared an example here to better explain the idea:  
   https://github.com/operator-framework/operator-controller/pull/2081/files#r2209548237

2. Just a small suggestion: maybe add a note in the docs (or somewhere else) to explain that the rule values are fixed and where they come from.  
   Reference: https://github.com/operator-framework/operator-controller/pull/2081/files#r2209551143

Let’s just make sure we don’t forget to follow up later. 😉

Also, based on @tmshort’s comment  
(https://github.com/operator-framework/operator-controller/pull/2081#issuecomment-3063649513),  
we’ll probably need another follow-up to improve the manifests too.
Since @tmshort unhold it seems fine we move forward with it now. 

We should ensure that only e2e installs has Prometheus.
However,  now we are generting the manifests: https://github.com/operator-framework/operator-controller/tree/main/manifests and I cannot see any required changes there. So, it seems fine 👍

PR #2085, opened 2025-07-09
🌱 Add test-operator with custom controller

perdasilva:
# Description

Re-opening #2070 and moving test-operator to Helm chart to facilitate the bundle manifest generation.

Needs #2093 to reduce the size and scope of change for this one.

## Reviewer Checklist

- [ ] API Go Documentation


Makefile
```diff
@@ -246,11 +246,41 @@ test-unit: $(SETUP_ENVTEST) envtest-k8s-bins #HELP Run the unit tests
...
+    $(CONTROLLER_GEN) --load-build-tags=$(GO_BUILD_TAGS) rbac:roleName=test-operator-manager-role webhook paths=".$(TEST_OPERATOR_DIR)/v2/..." output:dir=$(TEST_OPERATOR_DIR)/charts/v2/templates
+
+.PHONY: test-operator-generate-bundles
+test-operator-generate-bundles: test-operator-generate-manifests $(OPERATOR_SDK)
+	# generate bundles
+	helm template --include-crds $(TEST_OPERATOR_DIR)/charts/v1 | $(OPERATOR_SDK) generate bundle -q --channels beta --version 1.0.0 --package test --output-dir ./testdata/images/bundles/test-operator/v1.0.0
+	helm template --include-crds --set=configmap.shouldNotTemplate=true $(TEST_OPERATOR_DIR)/charts/v1 | $(OPERATOR_SDK) generate bundle -q --channels beta  --version 1.3.0 --package test --output-dir ./testdata/images/bundles/test-operator/v1.3.0
+	helm template --include-crds $(TEST_OPERATOR_DIR)/charts/v2 | $(OPERATOR_SDK) generate bundle -q --channels beta --version 2.0.0 --package test --output-dir ./testdata/images/bundles/test-operator/v2.0.0
+	rm -rf bundle.Dockerfile
```
camilamacedo86: All above should be part of the sample 
If we use the tooling we do not need this code

perdasilva: In the end the only difference between both approaches is how you deal with changes to the underlying operator code:

kubebuilder approach:

some bash script:

kubebuilder init
kubebuilder create ...
kubebuilder ...
commmands to mutate generated code: cp, sed, patch, etc.
operator-sdk generate bundle

this approach:

build the operator how you want once in-tree as a helm chart modify and render at will using a familiar methodology (the exact same we use for o-c). Modifications can be done directly to the fixture, no need to figure out how modify generated code the right way to get the things you need in there. Basically, use kubebuilder/operator-sdk to scaffold a helm project, remove cruft, modify directly when needed.

Ultimately, that's the issue for me with the automated approach, you end up transferring complexity to how the modifications are done to save on resource files that can be generated. Except, once we're closer to the helm world, we'll need to keep those resources around anyway...so I really don't get what we get out of simulating the operator developer flow...=S

camilamacedo86: The problem with is:

goes against important separation-of-concerns principles.
It introduces coupling that will make these components harder to maintain over time.
Keeping this logic self-contained in the sample is key to preserving clarity and long-term maintainability.

testdata/images/bundles/test-operator/v2.0.0/manifests/testoperator.clusterserviceversion.yaml
```diff
@@ -1,141 +0,0 @@
-apiVersion: operators.coreos.com/v1alpha1
```
camilamacedo86, line 1: Why were the manifests removed?

.github/workflows/testdata.yaml
```diff
@@ -0,0 +1,23 @@
+name: testdata
```
camilamacedo86, line 1: Can we discuss the case scenarios:

The test-operator supports AllNamespaces, SingleNamespace, and OwnNamespace install modes.

- v1.0.0: is just the crd and the existing configmap resource
- v1.3.0: has the same crd as v1.0.0 and the updated configmap previously part of v2.0.0
- v2.0.0: has validating, mutating and conversion webhooks

The DELTA between v1.0.0 and v2.0.0 is clear
V2.0.0 should introduce webhooks 

But what is the DELTA between v1.0.0 and v1.3.0
What do we want to cover with v1.3.0? 
Do we want to add any changes that would not be a breaking change?
What scenarios do we need to cover?
 
IHMO it should be:

- v1.0.0 - Initial basic project ( no metrics. no networkpolicies )
- v.1.1.0 - Enable network policy, metrics, etc (not a breaking change)
- v2.0.0 - Has webhooks and v2 CRD -> migration from v1 to V2 using spoke

perdasilva: 1.3.0 is just to harmonize with what's already there. It's something that can definitely be revisited. Right now, 2.0.0 includes this configmap. We can probably just collapse it to 1.0.0 and do the check. I think it's just a sanity check to make sure that annotations that include template syntax don't get affected.

Makefile
perdasilva: There's coupling in any case. In the automated case, we are coupled to the tools that we use to generate the foundation on top of which we will build out the operators. If the foundation changes, all layers on top of that will be affected, and that to me that, afaics, will be far more of a maintenance headache. If there are any changes in the generated scaffolding, it will be like doing a rebase. And in any case, the fixtures we will build will be in function of the project. So, there's coupling there too.

Given that the goals are the same in both approaches, I don't think they differ very much when it comes to separation of concerns. 

To me, the core of the difference between both approaches are how in how we make changes to preexisting test-operator code/resources. In one case, we generate the scaffolding once, and make changes directly to the files and produce the artifacts from that source of truth. In the other, we generate the scaffolding and replay changes on top of it to get it to the desired state and produce the artifacts from that source of truth.

This is the only dimension in which I see there's any material difference between the approaches.

.github/workflows/testdata.yaml
perdasilva: I think the list you put there is fine. I don't even think we need the 1.1.0 version. We only care about what is in the payload insofar as it interacts with OLM features/behavior. I.e. whether it has network policies, exposes metrics, uses HA, etc. isn't important if it doesn't matter to OLM. We don't need to add the complexity. 

When we get to things like probing the state of the installation, or want to test some boxcutter features etc, then I think this will become more necessary. For now, if it stamps out a configmap (or any other resource) just to say it stamped out a resource from the bundle, that's enough imo. I think v1.0.0 and v2.0.0 give us what we need for now.

testdata/images/bundles/test-operator/v2.0.0/manifests/testoperator.clusterserviceversion.yaml
perdasilva: They are back now. I was playing with just generating them when needed.

.github/workflows/testdata.yaml
camilamacedo86: >  I.e. whether it has network policies, exposes metrics, uses HA, etc. isn't important if it doesn't matter to OLM

IHMO: It is metter because we must support those configs
We must ensure that OLM works well with NP, and etc

perdasilva: > We must ensure that OLM works well with NP, and etc

I agree. But if working well means those resources are included when we create a bundle, I think unit tests would be sufficient to verify that a NetworkPolicy in the bundle gets properly rendered and that things that get rendered get applied to the cluster. 

If it means that OLM needs to interact with specific resources in some kind of special way (e.g. probing for status - or waiting for some other kind of signal), then I agree we'd probably want an e2e for that.

perdasilva:
closing as stale

PR #2088, opened 2025-07-10
:seedling: Move Registry YAML and remove use of namespace in kustomizations

tmshort:
Move e2e registry resources into config

Currently, the e2e registry resources are located in a script `testdata/build-test-registry.sh`

The resources in this script can be moved the the e2e config and manifests that now exist.

The only thing that remains is the job, that kinda-sorta needs to wait until the registry is up and working.

This requires removing the use of `namespace` in kustomize, because it behaves oddly when components and overlays use it.

# Description

## Reviewer Checklist

- [ ] API Go Documentation


internal/operator-controller/controllers/clusterextension_controller.go
```diff
@@ -93,9 +93,9 @@ type InstalledBundleGetter interface {
 //+kubebuilder:rbac:groups=olm.operatorframework.io,resources=clusterextensions,verbs=get;list;watch;update;patch
 //+kubebuilder:rbac:groups=olm.operatorframework.io,resources=clusterextensions/status,verbs=update;patch
 //+kubebuilder:rbac:groups=olm.operatorframework.io,resources=clusterextensions/finalizers,verbs=update
-//+kubebuilder:rbac:namespace=system,groups=core,resources=secrets,verbs=create;update;patch;delete;deletecollection;get;list;watch
+//+kubebuilder:rbac:namespace=olmv1-system,groups=core,resources=secrets,verbs=create;update;patch;delete;deletecollection;get;list;watch
 //+kubebuilder:rbac:groups=core,resources=serviceaccounts/token,verbs=create
-//+kubebuilder:rbac:namespace=system,groups=core,resources=serviceaccounts,verbs=get;list;watch
+//+kubebuilder:rbac:namespace=olmv1-system,groups=core,resources=serviceaccounts,verbs=get;list;watch
```
camilamacedo86, line 98: We no longer transform with the Kustomize system for the olmv1-system, right?

config/base/catalogd/kustomization.yaml
```diff
@@ -1,7 +1,6 @@
 # Does not include the CRD, which must be added separately (it's non-namespaced)
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
-namespace: olmv1-system
```
camilamacedo86, line 4: The problem with it is:
If we need to change the name, then we need to change it in all files, right?
Now we change just in one place.

camilamacedo86, line 4: So, if we need to change the name, do we have to update all the files after this change?
Also, I haven't checked, but in the downstream setup, are we using a shell script to make the change, or are we properly setting the Kustomize value and then building?

camilamacedo86: Could we not do the other changes and keep the namespace definition on the kustomize?

tmshort: This is making the kustomization more robust.

Downstream, the kustomization is done with a completely different overlay, and thus, sets the namespace appropriately according to openshift.

It's non-trivial to change the namespace via kustomize; there are also odd interdependencies based on ordering of components that may cause the namespaces of components outside of `olmv1-system` to be overwritten. For example, the CA components need(ed/s) to be last, or their `cert-manager` namespace will be overwritten with `olmv1-system`. Downstream doesn't have to deal with cert-manager, so there's no issues with namespace overwriting.

This change is something @joelanford had been looking at, in his reorganization of kustomize (tagging him as he has some interest here).

We never really gave people an easy way to change the namespace anyways, as the installation was done via the quickstart-generated files, and those don't have options to change the namespace.

If we want people to change the namespace, perhaps we should add functionality to the resulting `install.sh` script?

camilamacedo86: This is my only concern in the PR. That said, if everyone is aligned with the change, I’m okay moving forward with it.

If we want to provide this flexibility, we’ll need to replace the namespace (ns) in all places eventually—**regardless of whether we use Kustomize to build it or not.** 

**The only downside I see is that this information isn't currently centralised**, so we’ll need to remember to update every reference manually. That said, a simple find-and-replace should be enough to handle it if changes are needed later. We can either add validation to ensure/ci if we wish to do so

tmshort: There has been some discussion on moving the configuration from kustomize to helm... but that's a project in-and-of itself...

tmshort: I'm having problems with the upgrade-e2e, at this point, I want to split this into two PRs, the namespace change and the registry change.

tmshort:
Closing to separate the PR into two parts.

PR #2092, opened 2025-07-14
:seedling: Remove synthetic auth from experimental manifests

tmshort:
We are not planning on doing the synthetic auth in light of the boxcutter effort.

# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/config/components/base/experimental/kustomization.yaml
+++ b/config/components/base/experimental/kustomization.yaml
@@ -8,7 +8,6 @@ resources:
 components:
 - ../common
 # EXPERIMENTAL FEATURES ARE LISTED HERE
-- ../../features/synthetic-user-permissions
 - ../../features/webhook-provider-certmanager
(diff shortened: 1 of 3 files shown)


config/components/base/experimental/kustomization.yaml
```diff
@@ -8,7 +8,6 @@ resources:
 components:
 - ../common
 # EXPERIMENTAL FEATURES ARE LISTED HERE
-- ../../features/synthetic-user-permissions
```
camilamacedo86, line 11: Could you please add a description about why we should remove it?

perdasilva:
/lgtm

But I agree with Camila, might be worth adding some words in the description for the motivation

config/components/base/experimental/kustomization.yaml
tmshort: I can add it to the description of the PR.

PR #2103, opened 2025-07-16
🌱 test: Improve registry+v1 render regression test

camilamacedo86:
## Simplify and fix render regression test

- Old test made it easy to overwrite expected files via `make verify`.
- In all Operator Framework projects, `make verify` is used for codegen (`go fmt`, `controller-gen`, etc), **not** for regenerating testdata that should never change.  
  See [context](https://github.com/operator-framework/operator-controller/pull/1897)
- Expected manifests should only change when renderer logic changes on purpose.

This PR:

- Moves bundles and expected output under `testdata/`, following Go test conventions.
- Makes it clear the ArgoCD bundle is used only for this regression test.
- Adds a proper test that runs the renderer and checks the output matches the expected YAMLs.

Safer, clearer, and avoids mistakes in reviews.


internal/operator-controller/rukpak/render/render_test.go
```diff
@@ -267,3 +271,190 @@ func Test_BundleValidatorCallsAllValidationFnsInOrder(t *testing.T) {
...
+
+// Test_Render_ValidatesOutputForAllInstallModes verifies that the BundleRenderer
+// generates the correct set of Kubernetes resources (client.Objects) for each supported
+// install mode: AllNamespaces, SingleNamespace, and OwnNamespace.
+//
+// For each mode, it checks that:
+// - All expected resources are returned.
+// - The full content of each resource matches the expected values
+// It validates that the rendered objects are correctly rendered.
+func Test_Render_ValidatesOutputForAllInstallModes(t *testing.T) {
```
camilamacedo86: @perdasilva I could improve this one. 
I hope that you liked it

@anik120 and @thetechnick were involved in the introduction of this one, so I think it's good to ping them to help us PTAL this one and get it merged.

perdasilva: I think you've raised some valid concerns in the description, but missed the main point of this test. What we want to verify is that the registry+v1 bundle renderer consistently behaves in the expected way. That is, given the same bundle (+ config) the renderer always outputs the same manifests. 

The way we decided to do this is by taking an existing bundle that we are confident is correct (since we use the ArgoCD operator in our config sample - we took that one) and using the registry+v1 bundle renderer to generate manifests with different configurations (right now watchNamespace, but will also be subscriptionConfig in the near future) and make sure that the output doesn't change (i.e. the renderer's behavior has not changed).

The way I'm reading the test here is we create our own renderer which returns the mocked resources and we verify that implementation. Does that make sense?

I'm also not a huge fan of carrying so many manifests and I hear your concerns about the whats and whys. Maybe we could find a different way to address those concerns? Maybe adding a README, etc.?

What we want to be certain of is that we consistently render registry+v1 bundles in the same way. The most important part of that is whether the ClusterServiceVersion gets transpiled in to the right Deployment, RBAC, Webhook resources. Other resource kinds will just get tacked on. So, things like NetworkPolicy are not so important - or are as important as any other (service, configmap, etc.).

camilamacedo86: I changed the implementation to keep everything working as before and only address the main concerns raised.  
Closing this PR so we can continue with a clean one and allow a fresh round of reviews.

test/convert/README.md
```diff
@@ -1,7 +0,0 @@
-## registry+v1 bundle generation regression tests
```
camilamacedo86, line 1: we do not need the readme anymore
now, the code is documented and we can clarify on it directly to simplify

Makefile
```diff
@@ -252,6 +246,10 @@ test-unit: $(SETUP_ENVTEST) envtest-k8s-bins #HELP Run the unit tests
                 $(UNIT_TEST_DIRS) \
                 -test.gocoverdir=$(COVERAGE_UNIT_DIR)
 
+.PHONY: test-regression
+test-regression: #HELP Run regression test
+	go test -count=1 -v ./test/regression/...
```
camilamacedo86: We now have a dedicated directory for regression tests, so it's easy to add more cases in the future.  
It also makes it very clear what this test does and why it's there.

Makefile
```diff
@@ -172,15 +172,9 @@ generate: $(CONTROLLER_GEN) #EXHELP Generate code containing DeepCopy, DeepCopyI
 	$(CONTROLLER_GEN) --load-build-tags=$(GO_BUILD_TAGS) object:headerFile="hack/boilerplate.go.txt" paths="./..."
 
 .PHONY: verify
-verify: k8s-pin kind-verify-versions fmt generate manifests crd-ref-docs generate-test-data #HELP Verify all generated code is up-to-date. Runs k8s-pin instead of just tidy.
```
camilamacedo86, line 175: If `make verify` failed in CI, most people would just run it locally and push the changes — that’s normal since `verify` usually runs things like `go fmt`, `vet`, and `controller-gen` (as in all OF projects). But it was also re-generating regression test output — which should **never change or at least very unlikely change**.  Then:
- If that output changed because of a bug, we could accidentally miss it and merge a regression.
- It also wasn’t clear to reviewers that files in `testdata/expected-manifests/` should never be edited or auto-regenerated.
- Expecting everyone to remember to check that manually is risky and easy to miss.

test/regression/convert/covert_test.go
```diff
@@ -0,0 +1,86 @@
...
+	cmd.Stdout = os.Stdout
+
+	err := cmd.Run()
+	require.NoError(t, err, "failed to generate manifests")
+
+	// Compare structure + contents
+	err = compareDirs(expectedRoot, tmpRoot)
+	require.NoError(t, err, "rendered output differs from expected")
+}
+func compareDirs(expectedRoot, actualRoot string) error {
```
perdasilva: As it is, this might be missing the case where actual contains all expected files + some additional ones. So, we might want to run in twice (once in each direction), or compare the number of files generated in each case.

test/regression/convert/generate-manifests.go
```diff
@@ -16,11 +29,26 @@ import (
...
+//	./testdata/tmp/generate/
+//
+// If you want to update the expected output, run it with:
+//
+//	go run generate-manifests.go -output-dir=./testdata/expected-manifests/
+//
+// Note: Expected output should never change unless the renderer changes which is unlikely.
+// If the covert_test.go test fails, it likely means a regression was introduced in the renderer.
 func main() {
 	bundleRootDir := "testdata/bundles/"
```
perdasilva, line 46: should we move testdata/bundle and testdata/expected-manifests under testdata/regression-test or something like that? Just keep those two directories grouped under the same heading/purpose

camilamacedo86: I think we should keep this test data in test/regression/convert/.
This way, it's clear the data is only for the convert test.

If one day we have another regression test that also needs this data (which seems unlikely; it is very specific use case), then we can move it to the main regression/ folder.
That would show it's shared by more than one test.

test/regression/convert/covert_test.go
camilamacedo86: Good idea !!

camilamacedo86: I changed see wdyt

test/regression/convert/covert_test.go
```diff
@@ -0,0 +1,115 @@
...
+		if !bytes.Equal(expectedBytes, actualBytes) {
+			return fmt.Errorf("file content mismatch at: %s", relPath)
+		}
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+
+	// Step 1: Ensure actual does not contain unexpected files
```
perdasilva:
```suggestion
	// Step 2: Ensure actual does not contain unexpected files
```

test/regression/convert/generate-manifests.go
perdasilva: it's all good - I thought it was in the <root>/testdata. This is perfect!

test/regression/convert/covert_test.go
```diff
@@ -0,0 +1,115 @@
+/*
```
bentito: I think this file is named wrong, should be `convert_test.go` right?

test/regression/convert/covert_test.go
```diff
@@ -0,0 +1,115 @@
...
+)
+
+// Test_RenderedOutputMatchesExpected runs generate-manifests.go,
+// then compares the generated files against expected-manifests/.
+//
+// It fails if any file differs or is missing.
+//
+// TMP dir is cleaned up after test ends.
+func Test_RenderedOutput_MatchesExpected(t *testing.T) {
+	tmpRoot := "./testdata/tmp/rendered/"
```
bentito: Hard coding `tmpRoot := "./testdata/tmp/rendered"` writes into your source tree. Consider using `t.TempDir()` (or at least `os.MkdirTemp`) so failures don’t leave junk under testdata/ and so tests can run in parallel without colliding.

test/regression/convert/covert_test.go
```diff
@@ -0,0 +1,115 @@
...
+	})
+
+	// Call the generate-manifests.go script to generate the manifests
+	// in the temporary directory.
+	cmd := exec.Command("go", "run", "generate-manifests.go", "-output-dir="+tmpRoot)
+	cmd.Stderr = os.Stderr
+	cmd.Stdout = os.Stdout
+
+	err := cmd.Run()
+	require.NoError(t, err, "failed to generate manifests")
```
bentito: Spawning `go run generate-manifests.go` inside your test will recompile the whole helper and is fairly slow. You could factor the logic in `generate-manifests.go` into a package and call that from the test binary. This will surface compile-time errors directly in the test binary.

test/regression/convert/covert_test.go
```diff
@@ -0,0 +1,115 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to read expected file: %s", expectedPath)
+		}
+		actualBytes, err := os.ReadFile(actualPath)
+		if err != nil {
+			return fmt.Errorf("missing file: %s", relPath)
+		}
+
+		if !bytes.Equal(expectedBytes, actualBytes) {
+			return fmt.Errorf("file content mismatch at: %s", relPath)
```
bentito: nit: When you detect a content mismatch you return only `file content mismatch at: ""`. It would be more helpful to surface a small diff (like with `github.com/google/go-cmp/cmp.Diff`, etc) so people can see what actually changed.

test/regression/convert/generate-manifests.go
```diff
@@ -1,7 +1,20 @@
+// generate-manifests.go
+//
+// Renders registry+v1 bundles into YAML manifests for regression testing.
+// Used by tests to make sure output from the BundleRenderer stays consistent.
+//
+// By default, writes to ./testdata/tmp/generate/.
```
bentito, line 6: You say “By default, writes to ./testdata/tmp/generate/” but the test uses `./testdata/tmp/rendered/`. Either unify on one directory name or drop the default entirely and require the flag, so there’s no confusion.

test/regression/convert/generate-manifests.go
```diff
@@ -16,11 +29,26 @@ import (
...
 func main() {
 	bundleRootDir := "testdata/bundles/"
-	outputRootDir := "test/convert/expected-manifests/"
+	defaultOutputDir := "./testdata/tmp/generate/"
+	outputRootDir := flag.String("output-dir", defaultOutputDir, "path to write rendered manifests to")
+	flag.Parse()
 
-	if err := os.RemoveAll(outputRootDir); err != nil {
+	if err := os.RemoveAll(*outputRootDir); err != nil {
 		fmt.Printf("error removing output directory: %v\n", err)
```
bentito, line 52: should this direct to stderr so it doesn't mix with stdout at some point, or maybe even import `log` and `log.Fatalf`

test/regression/convert/covert_test.go
camilamacedo86: Thanks for the feedback!

You're right — using `go run generate-manifests.go` is a bit slower since it recompiles the helper. I timed it, and it's around **~500ms to 2 seconds faster**, but feels fine for this test since it only runs once, not per test case.

**We now have a stronger check in place for renderer output**, and **I think it's solid enough to move forward with this version**. I do agree that may have hall for improvements, but it might be more effort than it's worth at the moment.

That said, I’d be happy if someone wants to improve it later!

**WDYT — okay to move forward for now?**

test/regression/convert/generate-manifests.go
camilamacedo86: The test was failing because it couldn't remove a temporary directory.

I don’t think that should be a hard failure — if this happens in CI, the container will be cleaned up anyway when the job ends. It’s not critical to the test result itself.

I also added the directory to `.gitignore` to make sure we don’t accidentally commit it in any case.

test/regression/convert/covert_test.go
camilamacedo86: Done

test/regression/convert/covert_test.go
camilamacedo86: Done :thanks:

test/regression/convert/covert_test.go
camilamacedo86: Done :thanks:

test/regression/convert/covert_test.go
camilamacedo86: Done :thanks:

Makefile
tmshort: This does not appear to be invoked in any CI, specifically, this is only included in `test`, which is something we only do locally.

camilamacedo86: I did not realise that we did not call `make test` in the CI :-( 
Good catcher 💯 

Since we have a specific action for unit tests and can, in the future, expand the regression tests to cover other scenarios, I added a new action to call this one.

camilamacedo86: It is here now: https://github.com/operator-framework/operator-controller/actions/runs/16507728332/job/46682560856?pr=2103

camilamacedo86: @tmshort could we move forward with this one now?

.github/workflows/test-regression.yaml
```diff
@@ -0,0 +1,31 @@
...
+    - name: Run basic unit tests
+      run: |
+        make test-regression
+
+    - uses: codecov/codecov-action@v5.4.3
+      with:
+        disable_search: true
+        files: coverage/unit.out
+        flags: unit
+        token: ${{ secrets.CODECOV_TOKEN }}
```
tmshort: I don't think we need the codecov action here. This could inadvertently trigger codecov to think it's received all its data before the (real) tests complete. If you want this here, you should increase the value here:
https://github.com/operator-framework/operator-controller/blob/e9c5b6e3be4f90d793bb705e544ec98f92a9f3d0/codecov.yml#L3

tmshort: The number 3 corresponds to test-unit, test-e2e and test-experimental-e2e.

camilamacedo86: Interesting now, I think I understand better why codev instead of the old coveralls :-)

camilamacedo86: I think we should either use it if we are covering more within

Screenshot 2025-07-31 at 07 25 17

All done !

tmshort: Actually interesting, the experimental-e2e isn't creating it's own directory, it's sharing it with e2e, so I there' still 3 files, which I'm not sure if it corresponds to three pushes...

However, I'm not sure it's a good idea to code cover the test harness, as that gives a false impression of code coverage (either positive or negative).
Regardless, I'll let this pass.

camilamacedo86: If we add more tests here, will we not be covering more scenarios and more areas of the code each time?
I think we need to fix the experimental-e2e. I can look at that in a follow-up.
Could you LGTM this one for we get it merged?

PR #2116, opened 2025-07-24
🌱 Remove kaniko / extension-developer-e2e cleanup

dtfranz:
Removes use of kaniko, which is now an archived project, in favor of using tools we already have at our disposal. Also attempted to clean up the Makefile and setup script a bit, as well as a fail-early check in the test to make it clearer that some env is required.

The combination of env + args for this test is still not great but hopefully it's at least a bit easier to understand.

# Description

## Reviewer Checklist

- [ ] API Go Documentation


Makefile
```diff
@@ -223,14 +223,13 @@ E2E_REGISTRY_NAME := docker-registry
 E2E_REGISTRY_NAMESPACE := operator-controller-e2e
 
 export REG_PKG_NAME := registry-operator
-export LOCAL_REGISTRY_HOST := $(E2E_REGISTRY_NAME).$(E2E_REGISTRY_NAMESPACE).svc:5000
-export CLUSTER_REGISTRY_HOST := localhost:30000
-export E2E_TEST_CATALOG_V1 := e2e/test-catalog:v1
-export E2E_TEST_CATALOG_V2 := e2e/test-catalog:v2
-export CATALOG_IMG := $(LOCAL_REGISTRY_HOST)/$(E2E_TEST_CATALOG_V1)
+export CLUSTER_REGISTRY_HOST := $(E2E_REGISTRY_NAME).$(E2E_REGISTRY_NAMESPACE).svc:5000
```
dtfranz, line 226: The naming of `CLUSTER` vs `LOCAL` for registry host was backwards IMO, so I switched it around.

test/extension-developer-e2e/extension_developer_test.go
```diff
@@ -32,6 +32,9 @@ func TestExtensionDeveloper(t *testing.T) {
 	require.NoError(t, corev1.AddToScheme(scheme))
 	require.NoError(t, rbacv1.AddToScheme(scheme))
 
+	require.NotEmpty(t, os.Getenv("CATALOG_IMG"), "environment variable CATALOG_IMG must be set")
```
dtfranz, line 35: These fields are required to be set for the test, and if not it will fail in an unclear way.

dtfranz:
Having issues with podman vs docker environments 😑 
/hold

dtfranz:
Should be fixed now
/unhold

Makefile
```diff
@@ -223,14 +223,14 @@ E2E_REGISTRY_NAME := docker-registry
...
+export LOCAL_REGISTRY_HOST := localhost:30000
 export E2E_TEST_CATALOG_V1 := e2e/test-catalog:v1
 export E2E_TEST_CATALOG_V2 := e2e/test-catalog:v2
-export CATALOG_IMG := $(LOCAL_REGISTRY_HOST)/$(E2E_TEST_CATALOG_V1)
+export CATALOG_IMG := $(CLUSTER_REGISTRY_HOST)/$(E2E_TEST_CATALOG_V1)
 .PHONY: test-ext-dev-e2e
-test-ext-dev-e2e: $(OPERATOR_SDK) $(KUSTOMIZE) $(KIND) #HELP Run extension create, upgrade and delete tests.
-	test/extension-developer-e2e/setup.sh $(OPERATOR_SDK) $(CONTAINER_RUNTIME) $(KUSTOMIZE) $(KIND) $(KIND_CLUSTER_NAME) $(E2E_REGISTRY_NAMESPACE)
+test-ext-dev-e2e: $(OPERATOR_SDK) $(KUSTOMIZE) #HELP Run extension create, upgrade and delete tests.
```
camilamacedo86, line 232: Should we not keep the  $(KIND) here: `test-ext-dev-e2e: $(OPERATOR_SDK) $(KUSTOMIZE)`
We removed it from the shell `test/extension-developer-e2e/setup.sh` but I understand that the above will ensure that we have it installed.

test/extension-developer-e2e/setup.sh
```diff
@@ -149,107 +153,5 @@ cat <<EOF > "${TMP_ROOT}"/catalog/index.yaml
...
-        - name: manifests
-          emptyDir: {}
-        - name: manifests-data
-          configMap:
-            name: operator-controller-${reg_pkg_name}.manifests
-EOF
-
-kubectl wait --for=condition=Complete -n "${namespace}" jobs/kaniko-${reg_pkg_name} --timeout=60s
+${container_tool} build -f "${TMP_ROOT}/catalog.Dockerfile" -t "${catalog_push_tag}" "${TMP_ROOT}/"
+${container_tool} push ${catalog_push_tag} ${tls_flag}
```
camilamacedo86, line 157: Just to ensure that I understand it, do we need to push the catalog image for a registry build locally, because in OCP we cannot do a kind load image, am I right?

camilamacedo86 (commented):
Nice 🚀 
I loved it when we were able to delete dependencies with third parties!!!!
Just a nit and a question. Otherwise, I am OK with.

I think @tmshort should give a look as well since I think he has been working relate to it.

Makefile
dtfranz: IMO we should limit dependency declaration like that to targets that actually use the dependency. In my mind this would be a little bit like adding `$(KIND)` to the `e2e` target; it's not necessary because `e2e` can't be run without the prerequisites from `test-e2e` already being called, and if they weren't then having `$(KIND)` declared as a dependency for it wouldn't help at all.

test/extension-developer-e2e/setup.sh
dtfranz: Using a registry vs `kind load` is theoretically faster since `kind load` will transfer all image layers each time it's run, while the registry can reuse them. It's still an option, but I think the registry is probably more feature-rich.

Makefile
camilamacedo86: make sense 👍

Makefile
tmshort: The original naming was cluster-centric... I do wonder if there are better names... e.g.
INSIDE_CLUSTER ?
OUTSIDE_CLUSTER ?

tmshort:
Looks reasonable
/approve

PR #2134, opened 2025-08-04
✨ Metrics Summary

dtfranz:
# Description

Adds a util to the e2e suite which queries prometheus at the end of the test run for alerts and metrics data. This data is then processed into markdown and displayed to the contributor at the end of their test runs.

The principal idea here is that if we are to fail a test run based on performance results, then those results need to be easily visible to the contributor to avoid a potentially frustrating development experience. 

## Reviewer Checklist

- [ ] API Go Documentation


config/overlays/prometheus/prometheus_rule.yaml
```diff
@@ -22,13 +22,13 @@ spec:
       annotations:
         description: "container {{ $labels.container }} of pod {{ $labels.pod }} experienced OOM event(s); count={{ $value }}"
     - alert: operator-controller-memory-growth
-      expr: deriv(sum(container_memory_working_set_bytes{pod=~"operator-controller.*",container="manager"})[5m:]) > 50_000
+      expr: deriv(sum(container_memory_working_set_bytes{pod=~"operator-controller.*",container="manager"})[5m:]) > 100_000
```
dtfranz, line 25: These values were too sensitive.

test/e2e/metrics_test.go
```diff
@@ -129,15 +129,15 @@ func (c *MetricsTestConfig) getServiceAccountToken(t *testing.T) string {
 func (c *MetricsTestConfig) createCurlMetricsPod(t *testing.T) {
 	t.Logf("Creating curl pod (%s/%s) to validate the metrics endpoint", c.namespace, c.curlPodName)
 	cmd := exec.Command(c.client, "run", c.curlPodName,
-		"--image=curlimages/curl",
+		"--image=curlimages/curl:8.15.0",
```
dtfranz, line 132: So we don't force `kind` to pull this image every time we create the pod.

testdata/images/bundles/test-operator/v1.0.0/manifests/testoperator.clusterserviceversion.yaml
```diff
@@ -58,7 +58,7 @@ spec:
               terminationGracePeriodSeconds: 0
               containers:
               - name: busybox
-                image: busybox
+                image: busybox:1.36
```
dtfranz, line 61: Same logic as with the `curl` pod.

config/overlays/prometheus/prometheus_rule.yaml
trgeiger: I might just note this in the commit message or put in another commit

trgeiger:
This is very cool

config/overlays/prometheus/prometheus_rule.yaml
dtfranz: Good call out 👍 I'll add a note to the commit message.

dtfranz: I've added the following to the commit and the PR description:

> Extra: Tuned prometheus alerts to be less sensitive to memory growth. The tests will naturally cause an additional memory footprint at the beginning of the e2e, so we need to account for that somehow. Also tagged a couple of images we were implicitly using 'latest' versions of so nodes won't have to pull them on every test run.

jianzhangbjz:
That's cool! 
/lgtm

PR #2136, opened 2025-08-05
:seedling: Update e2e tests for boxcutter

tmshort:
Modify the `RecoversFromInitialInstallFailedWhenFailureFixed` test to instead not create the namespace and service account, instead of having bad service account. Boxcutter will ignore the service account.

Add `RecoversFromExsitingDeploymentWhenFailureFixed` test to ensure we don't adopt, and instead fail when there's an existing resource that matches the bundle (i.e. don't adopt),

# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/test/e2e/cluster_extension_install_test.go
+++ b/test/e2e/cluster_extension_install_test.go
@@ -10,6 +10,7 @@ import (
 	"github.com/google/go-containerregistry/pkg/crane"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	networkingv1 "k8s.io/api/networking/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
@@ -20,6 +21,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/rand"
+	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	ocv1 "github.com/operator-framework/operator-controller/api/v1"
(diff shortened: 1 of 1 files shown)


test/e2e/cluster_extension_install_test.go
```diff
@@ -882,21 +901,95 @@ func TestClusterExtensionInstallReResolvesWhenManagedContentChanged(t *testing.T
...
-	require.NoError(t, err)
+
+	t.Log("It resolves the specified package with correct bundle path")
+	t.Log("By creating the ClusterExtension resource")
+	require.NoError(t, c.Create(context.Background(), clusterExtension))
+
+	t.Log("By eventually reporting a successful resolution and bundle path")
+	require.EventuallyWithT(t, func(ct *assert.CollectT) {
+		require.NoError(ct, c.Get(context.Background(), types.NamespacedName{Name: clusterExtension.Name}, clusterExtension))
+	}, pollDuration, pollInterval)
```
joelanford: This says we're reporting a successful resolution and bundle path, but then all we actually do is get the CE without any assertions? Seems like the log needs to change or we need to add some assertions to match the log?

joelanford: Or maybe we just delete this whole block since the next block is calling Get for its own purposes?

test/e2e/cluster_extension_install_test.go
```diff
@@ -933,18 +1061,18 @@ func TestClusterExtensionRecoversFromInitialInstallFailedWhenFailureFixed(t *tes
...
+		require.Contains(ct, cond.Message, "No bundle installed")
 	}, pollDuration, pollInterval)
 
-	t.Log("By fixing the ServiceAccount permissions")
-	require.NoError(t, createClusterRoleAndBindingForSA(context.Background(), name, sa, clusterExtension.Name))
+	t.Log("By deleting the new Deployment")
+	require.NoError(t, c.Delete(context.Background(), newDeployment))
 
 	// NOTE: In order to ensure predictable results we need to ensure we have a single
 	// known failure with a singular fix operation. Additionally, due to the exponential
```
joelanford, line 1068: This NOTE text should be updated to say that all that we should need to do is delete the deployment.

test/e2e/cluster_extension_install_test.go
```diff
@@ -882,21 +901,95 @@ func TestClusterExtensionInstallReResolvesWhenManagedContentChanged(t *testing.T
...
+	}, pollDuration, pollInterval)
+
+	t.Log("By creating the Namespace and ServiceAccount")
+	sa, ns := testInitServiceAccountNamespace(t, clusterExtension.Name)
+	defer testCleanup(t, nil, nil, sa, ns)
+
+	// NOTE: In order to ensure predictable results we need to ensure we have a single
+	// known failure with a singular fix operation. Additionally, due to the exponential
+	// backoff of this eventually check we MUST ensure we do not touch the ClusterExtension
+	// after creating and binding the needed permissions to the ServiceAccount.
```
joelanford:
```suggestion
	// after creating the namespace, SA, and RBAC.
```

PR #2139, opened 2025-08-08
✨ API Call Alerts

dtfranz:
# Description

Adds prometheus alerts for excessive API calls from operator-controller or catalogd, as well as summary graphs to match.

Do not fail e2e run when issues with summary generation are encountered or when the output isn't specified.

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/test/e2e/e2e_suite_test.go
+++ b/test/e2e/e2e_suite_test.go
@@ -25,7 +25,7 @@ var (
 )
 
 const (
-	testSummaryOutputEnvVar = "GITHUB_STEP_SUMMARY"
+	testSummaryOutputEnvVar = "E2E_SUMMARY_OUTPUT"
 	testCatalogRefEnvVar    = "CATALOG_IMG"
 	testCatalogName         = "test-catalog"
 	latestImageTag          = "latest"
@@ -40,9 +40,16 @@ func TestMain(m *testing.M) {
 	utilruntime.Must(err)
 
 	res := m.Run()
-	err = utils.PrintSummary(testSummaryOutputEnvVar)
-	if err != nil {
-		fmt.Println("PrintSummary error", err)
+	path := os.Getenv(testSummaryOutputEnvVar)
+	if path == "" {
+		fmt.Printf("Note: E2E_SUMMARY_OUTPUT is unset; skipping summary generation")
+	} else {
+		err = utils.PrintSummary(path)
+		if err != nil {
+			// Fail the run if alerts are found
+			fmt.Printf("%v", err)
+			os.Exit(1)
+		}
 	}
 	os.Exit(res)
 }
--- a/test/utils/summary.go
+++ b/test/utils/summary.go
@@ -42,14 +42,16 @@ type xychart struct {
 }
 
 type githubSummary struct {
-	client api.Client
-	Pods   []string
+	client       api.Client
+	Pods         []string
+	alertsFiring bool
 }
 
 func NewSummary(c api.Client, pods ...string) githubSummary {
 	return githubSummary{
-		client: c,
-		Pods:   pods,
+		client:       c,
+		Pods:         pods,
+		alertsFiring: false,
 	}
 }
 
@@ -60,7 +62,7 @@ func NewSummary(c api.Client, pods ...string) githubSummary {
 // yLabel - Label of the Y axis i.e. "KB/s", "MB", etc.
 // scaler - Constant by which to scale the results. For instance, cpu usage is more human-readable
 // as "mCPU" vs "CPU", so we scale the results by a factor of 1,000.
-func (s githubSummary) PerformanceQuery(title, pod, query string, yLabel string, scaler float64) (string, error) {
+func (s *githubSummary) PerformanceQuery(title, pod, query, yLabel string, scaler float64) (string, error) {
 	v1api := v1.NewAPI(s.client)
 	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
 	defer cancel()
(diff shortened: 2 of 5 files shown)


test/e2e/e2e_suite_test.go
```diff
@@ -24,7 +25,7 @@ var (
 )
 
 const (
-	testSummaryOutputEnvVar = "GITHUB_STEP_SUMMARY"
+	testSummaryOutputEnvVar = "E2E_SUMMARY_OUTPUT"
```
camilamacedo86, line 28: Nice 👍 
I think is better keep generic since we can run the same with or not GITHUB
Very cool

test/utils/summary.go
```diff
@@ -136,6 +143,8 @@ func (s githubSummary) Alerts() (string, error) {
 			switch a.State {
 			case v1.AlertStateFiring:
 				firingAlerts = append(firingAlerts, aConv)
+				// TODO see note in PrintSummary()
+				//s.firingAlerts = true
```
camilamacedo86: Could we remove the commented code?

test/utils/summary.go
```diff
@@ -172,28 +181,35 @@ func executeTemplate(templateFile string, obj any) (string, error) {
...
+	err = os.WriteFile(path, []byte(summaryMarkdown), 0o600)
+	if err != nil {
+		fmt.Printf("warning: failed to write summary output to %s: %s", path, err)
+		return nil
 	}
+	fmt.Printf("Test summary output to %s successful\n", path)
+	// TODO: uncomment when the metrics collection is proven to be stable
+	// if summary.FiringAlerts() {
+	//	return fmt.Errorf("Alert(s) encountered during test run; see summary for details")
+	//}
```
camilamacedo86: Same here ^

camilamacedo86 (commented):
Hi @dtfranz 

Thank you a lot just remove the commented code and I think we are good to move forward

test/e2e/e2e_suite_test.go
```diff
@@ -39,8 +40,18 @@ func TestMain(m *testing.M) {
 	utilruntime.Must(err)
 
 	res := m.Run()
-	err = utils.PrintSummary(testSummaryOutputEnvVar)
```
camilamacedo86, line 43: I was looking and seem that should be skiped already see:

https://github.com/operator-framework/operator-controller/blob/main/test/utils/summary.go#L175-L198

Then, why we just call it here and not in all e2e tests? 
Could you help me understand how it works?

dtfranz: I wanted to keep the mechanism of selecting output destination distinct in `summary.go` and `e2e_suite_test.go`. For `PrintSummary()` you supply a file path, and if that's empty then we warn and skip, since that's an invalid use of the func. In `e2e_suite_test.go` you supply env, and if that's empty then we skip the summary with a note, because it is totally valid to run e2e without summary. This allowed me to keep the e2e test runner as simple as possible and follow the convention of transparently supplying e2e arguments via env. But I probably overthought it 🤷

camilamacedo86: But each test has one suite right: 

- https://github.com/operator-framework/operator-controller/blob/main/test/experimental-e2e/experimental_e2e_test.go#L45-L56
- https://github.com/operator-framework/operator-controller/blob/main/test/upgrade-e2e/upgrade_e2e_suite_test.go#L30-L60
- https://github.com/operator-framework/operator-controller/blob/main/test/extension-developer-e2e/extension_developer_test.go#L24-L214

So, why we use it only in this one?

dtfranz: The summary is also generated for experimental e2e, you can see it in [here](https://github.com/operator-framework/operator-controller/actions/runs/16859756177/attempts/1#summary-47758222701). The upgrade and extension developer tests are pretty pointless to run prometheus on because they only have about a few seconds of actual runtime.

camilamacedo86 (approved):
Thank you for the answers

@tmshort, I think we will need your help too.
It should allow us to sync downstream and solve the current error.

grokspawn (approved):

test/e2e/e2e_suite_test.go
```diff
@@ -39,8 +40,18 @@ func TestMain(m *testing.M) {
...
+
+	path := os.Getenv(testSummaryOutputEnvVar)
+	if path == "" {
+		fmt.Printf("Note: E2E_SUMMARY_OUTPUT is unset; skipping summary generation")
+	} else {
+		err = utils.PrintSummary(path)
+		if err != nil {
+			// Fail the run if alerts are found
+			fmt.Printf("%s", err)
+			os.Exit(1)
```
tmshort, line 51: /hold
This actually causes an issue downstream. It seems that `PrintSummary` fails there. We need to come up with a solution that works for both upstream and downstream.
Also, did we want to use `%v` rather than `%s`?

tmshort: See #2144 for my workaround.

dtfranz: `PrintSummary` cannot return an error anymore, I disabled that to prevent it from failing the test. I can remove this section though if you want, but the idea was to have `PrintSummary` return an error for alerts once this is stable. 

In addition, downstream does not set `E2E_SUMMARY_OUTPUT`, so this code will never be reached. I don't think your workaround is necessary since this should work better as a permanent solution.

tmshort: So, #2144 merged, and you'll need to rebase onto that (it also had another fix).
If `PrintSummary` can no longer return an error, then it should have no return values.

dtfranz: Removing the error returns was part of fixing the downstream. If that's fine now then I will add the error returns for alerts back in.

dtfranz:
I think we need a way to add exceptions to the `apidiff` check, unless we're fine with overriding it every time we change exported function signatures in `test/`.

dtfranz:
Thank you for your feedback @tmshort , the branch is updated now per your comments; PTAL!

test/e2e/e2e_suite_test.go
tmshort: However, we still want it to exit according the the result of the test. If `PrintSummary` were to return an error, the test will fail.

tmshort: But, the function does not seem to return an error, but if it were to, it would still break because of the `os.Exti(1)`. That line should be removed; but for now, I'll approve.

tmshort: OK, the `E2E_SUMMARY_OUTPUT` variable won't be set downstream...

tmshort (approved):

tmshort:
The go-apidiff is acceptable, it's in test code.

PR #2157, opened 2025-08-21
 📖 Use --strict for mkdocs build/deploy

anik120:
# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/docs/draft/api-reference/catalogd-webserver-metas-endpoint.md
+++ b/docs/draft/api-reference/catalogd-webserver-metas-endpoint.md
@@ -31,7 +31,7 @@ As an example, to access only the [package schema](https://olm.operatorframework
 
 the URL to access the service would be `https://catalogd-service.olmv1-system.svc/catalogs/operatorhubio/api/v1/metas?schema=olm.package`
 
-For more examples of valid queries that can be made to the `api/v1/metas` service endpoint, please see [Catalog Queries](../howto/catalog-queries.md).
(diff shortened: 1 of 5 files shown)


docs/draft/api-reference/catalogd-webserver-metas-endpoint.md
```diff
@@ -31,7 +31,7 @@ As an example, to access only the [package schema](https://olm.operatorframework
 
 the URL to access the service would be `https://catalogd-service.olmv1-system.svc/catalogs/operatorhubio/api/v1/metas?schema=olm.package`
 
-For more examples of valid queries that can be made to the `api/v1/metas` service endpoint, please see [Catalog Queries](../howto/catalog-queries.md).
+For more examples of valid queries that can be made to the `api/v1/metas` service endpoint, please see [Catalog Queries](../../howto/catalog-queries.md).
```
tmshort, line 34: Given that these are in the `docs/draft` folder, and will be moved to the `docs` directory, I don't we need the extra directory level; it will all work out when it's moved out of drafts.

Same with the other two directory level changes.

anik120: Right, but then current `--strict` restrictions will throw up warning which will cause CI to fail. 

If we move forward with the current configuration, when the docs are moved out of draft, there will again be another failed CI event, which will highlight that the links need to be updated.

perdasilva (approved):

PR #2163, opened 2025-08-26
Bundle config api

perdasilva:
# Description

## Reviewer Checklist

- [ ] API Go Documentation


api/v1/clusterextension_types.go
```diff
@@ -92,6 +97,14 @@ type ClusterExtensionSpec struct {
 	//
 	// +optional
 	Install *ClusterExtensionInstallConfig `json:"install,omitempty"`
+
+	// config contains configuration values applied during rendering of the
+	// ClusterExtension's manifests. Values can be specified inline or sourced
+	// from a referenced Secret.
```
joelanford:
```suggestion
	// config contains configuration values applied during rendering of the
	// ClusterExtension's manifests. Values can be specified inline.
```

joelanford: Since the field is optional, we should also say what happens when it is not defined. For example:

>	// config is optional. When not specified, the default configuration
>  // of the resolved bundle will be used.

api/v1/clusterextension_types.go
```diff
@@ -138,6 +150,29 @@ type ClusterExtensionInstallConfig struct {
 	Preflight *PreflightConfig `json:"preflight,omitempty"`
 }
 
+// ClusterExtensionConfig is a discriminated union which selects the source configuration values to be merged into
+// the ClusterExtension's rendered manifests.
+//
+// +union
```
joelanford, line 159: We should add cel expression evaluation here to verify that the config type matches the non-nil struct pointer.

```
// +kubebuilder:validation:XValidation:rule="has(self.configType) && self.configType == 'Inline' ? has(self.inline) : !has(self.inline)",message="inline is required when configType is Inline, and forbidden otherwise"
```

manifests/boxcutter-e2e.yaml
tmshort, line 1: I don't see any reference to this file in this PR. I don't see generation (Makefile) or use.

api/v1/clusterextension_types.go
perdasilva: I had added it originally, but then it was getting rejected so I removed it. Let me check again.

internal/operator-controller/applier/helm.go
```diff
@@ -339,3 +340,10 @@ func ruleDescription(ns string, rule rbacv1.PolicyRule) string {
 	}
 	return sb.String()
 }
+
+// TODO: BundleConfig and its handling will move to the renderer in a follow-up PR. This is here temporarily so we can make
+//
+//	forward progress with the config story and already lay the API foundations for it while having some minimal functionality
+type BundleConfig struct {
```
joelanford: Do we use this anywhere? Perhaps we _could_ use it in `GetWatchNamespace()`?

manifests/boxcutter-e2e.yaml
joelanford, line 1: Accidentally committed?

perdasilva: sorry - strange that I missed cleaning it up...my bad

internal/operator-controller/applier/helm.go
perdasilva: that was vestigial from my original impl before dropping the config feature-gate. It should have been removed with some other old code, but I guess I fudged that one up...

manifests/boxcutter-e2e.yaml
perdasilva: Yup, that was a side think got checked in by mistake. Thanks for catching it =D I've cleaned it up.

api/v1/clusterextension_types.go
```diff
@@ -138,6 +152,30 @@ type ClusterExtensionInstallConfig struct {
 	Preflight *PreflightConfig `json:"preflight,omitempty"`
 }
 
+// ClusterExtensionConfig is a discriminated union which selects the source configuration values to be merged into
+// the ClusterExtension's rendered manifests.
+//
+// <opcon:experimental:validation:XValidation:rule="has(self.configType) && self.configType == 'Inline' ? has(self.inline) : !has(self.inline)",message="inline is required when type is Inline, and forbidden otherwise">
```
joelanford: Does this _need_ `<opcon:experimental:validation:XValidation:...>` instead of the standard `+kubebuilder` one?

I'm guessing the former is generally used for experimentally applying different validation to an existing field in the standard CRD.

In our case, I would have expected that since the entire field is optional and experimental, we could use the standard `+kubebuilder` definition, which would mean that the validation promotes at the same time that the field promotes without us having to remember to change this comment as well?

joelanford: Also, one copy/paste blunder on my end:

Change:
```
message="inline is required when type is Inline, and forbidden otherwise"
```

To:
```
message="inline is required when configType is Inline, and forbidden otherwise"
```

joelanford: Looking at the generated CRDs, I also don't see the CEL expression showing up there.

perdasilva: Got it with the object type, as you suggested in Slack =D

api/v1/clusterextension_types.go
```diff
@@ -92,6 +97,15 @@ type ClusterExtensionSpec struct {
 	//
 	// +optional
 	Install *ClusterExtensionInstallConfig `json:"install,omitempty"`
+
+	// config contains optional configuration values applied during rendering of the
+	// ClusterExtension's manifests. Values can be specified inline.
+	//
+	// config is optional. When not specified, the default configuration of the resolved bundle will be used.
+	//
+	// <opcon:experimental>
```
tmshort, line 106: Look at that _experimental_ tag...

PR #2166, opened 2025-08-26
✨ OPRUN-4090: (rukpak) Add config support for reg+v1 bundle to chart converter

anik120:
# Description

Put watchNamespace in config. 

Work for [RFC​: Config -> registry+v1 bundle config](https://docs.google.com/document/d/1rmcT8tXcezjq0Iy3Cde097Y8ZaCgUXhEgiuyfv1vXUI/edit?usp=sharing)

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/operator-controller/applier/helm.go
+++ b/internal/operator-controller/applier/helm.go
@@ -27,6 +27,7 @@ import (
 	ocv1 "github.com/operator-framework/operator-controller/api/v1"
 	"github.com/operator-framework/operator-controller/internal/operator-controller/authorization"
 	"github.com/operator-framework/operator-controller/internal/operator-controller/features"
+	"github.com/operator-framework/operator-controller/internal/operator-controller/rukpak/bundle"
 	"github.com/operator-framework/operator-controller/internal/operator-controller/rukpak/bundle/source"
 	"github.com/operator-framework/operator-controller/internal/operator-controller/rukpak/preflights/crdupgradesafety"
 	"github.com/operator-framework/operator-controller/internal/operator-controller/rukpak/util"
@@ -57,7 +58,7 @@ type Preflight interface {
 }
 
 type BundleToHelmChartConverter interface {
-	ToHelmChart(bundle source.BundleSource, installNamespace string, watchNamespace string) (*chart.Chart, error)
+	ToHelmChart(bundle source.BundleSource, installNamespace string, config map[string]interface{}) (*chart.Chart, error)
 }
 
 type Helm struct {
@@ -222,7 +223,10 @@ func (h *Helm) buildHelmChart(bundleFS fs.FS, ext *ocv1.ClusterExtension) (*char
 		}
 	}
 
-	return h.BundleToHelmChartConverter.ToHelmChart(source.FromFS(bundleFS), ext.Spec.Namespace, watchNamespace)
+	bundleConfig := map[string]interface{}{
+		bundle.BundleConfigWatchNamespaceKey: watchNamespace,
+	}
+	return h.BundleToHelmChartConverter.ToHelmChart(source.FromFS(bundleFS), ext.Spec.Namespace, bundleConfig)
 }
 
 func (h *Helm) renderClientOnlyRelease(ctx context.Context, ext *ocv1.ClusterExtension, chrt *chart.Chart, values chartutil.Values, post postrender.PostRenderer) (*release.Release, error) {
--- a/internal/operator-controller/applier/helm_test.go
+++ b/internal/operator-controller/applier/helm_test.go
@@ -28,6 +28,7 @@ import (
 	"github.com/operator-framework/operator-controller/internal/operator-controller/applier"
 	"github.com/operator-framework/operator-controller/internal/operator-controller/authorization"
 	"github.com/operator-framework/operator-controller/internal/operator-controller/features"
+	registryv1Bundle "github.com/operator-framework/operator-controller/internal/operator-controller/rukpak/bundle"
 	"github.com/operator-framework/operator-controller/internal/operator-controller/rukpak/bundle/source"
 	"github.com/operator-framework/operator-controller/internal/operator-controller/rukpak/convert"
 )
@@ -559,8 +560,8 @@ func TestApply_InstallationWithSingleOwnNamespaceInstallSupportEnabled(t *testin
 				},
 			},
 			BundleToHelmChartConverter: &fakeBundleToHelmChartConverter{
-				fn: func(bundle source.BundleSource, installNamespace string, watchNamespace string) (*chart.Chart, error) {
-					require.Equal(t, expectedWatchNamespace, watchNamespace)
+				fn: func(bundle source.BundleSource, installNamespace string, config map[string]interface{}) (*chart.Chart, error) {
+					require.Equal(t, expectedWatchNamespace, config[registryv1Bundle.BundleConfigWatchNamespaceKey])
 					return nil, nil
 				},
 			},
@@ -574,7 +575,7 @@ func TestApply_InstallationWithSingleOwnNamespaceInstallSupportEnabled(t *testin
 				Config: &ocv1.ClusterExtensionConfig{
 					ConfigType: ocv1.ClusterExtensionConfigTypeInline,
 					Inline: &apiextensionsv1.JSON{
-						Raw: []byte(fmt.Sprintf(`{"watchNamespace":"%s"}`, expectedWatchNamespace)),
+						Raw: []byte(fmt.Sprintf(`{"%s":"%s"}`, registryv1Bundle.BundleConfigWatchNamespaceKey, expectedWatchNamespace)),
 					},
 				},
 			},
@@ -597,8 +598,8 @@ func TestApply_RegistryV1ToChartConverterIntegration(t *testing.T) {
 				},
 			},
 			BundleToHelmChartConverter: &fakeBundleToHelmChartConverter{
-				fn: func(bundle source.BundleSource, installNamespace string, watchNamespace string) (*chart.Chart, error) {
-					require.Equal(t, expectedWatchNamespace, watchNamespace)
+				fn: func(bundle source.BundleSource, installNamespace string, config map[string]interface{}) (*chart.Chart, error) {
+					require.Equal(t, expectedWatchNamespace, config[registryv1Bundle.BundleConfigWatchNamespaceKey])
 					return nil, nil
 				},
 			},
@@ -617,7 +618,7 @@ func TestApply_RegistryV1ToChartConverterIntegration(t *testing.T) {
 				},
 			},
 			BundleToHelmChartConverter: &fakeBundleToHelmChartConverter{
-				fn: func(bundle source.BundleSource, installNamespace string, watchNamespace string) (*chart.Chart, error) {
+				fn: func(bundle source.BundleSource, installNamespace string, config map[string]interface{}) (*chart.Chart, error) {
 					return nil, errors.New("some error")
 				},
 			},
@@ -629,9 +630,9 @@ func TestApply_RegistryV1ToChartConverterIntegration(t *testing.T) {
 }
 
 type fakeBundleToHelmChartConverter struct {
-	fn func(source.BundleSource, string, string) (*chart.Chart, error)
+	fn func(source.BundleSource, string, map[string]interface{}) (*chart.Chart, error)
 }
 
-func (f fakeBundleToHelmChartConverter) ToHelmChart(bundle source.BundleSource, installNamespace string, watchNamespace string) (*chart.Chart, error) {
-	return f.fn(bundle, installNamespace, watchNamespace)
+func (f fakeBundleToHelmChartConverter) ToHelmChart(bundle source.BundleSource, installNamespace string, config map[string]interface{}) (*chart.Chart, error) {
+	return f.fn(bundle, installNamespace, config)
 }
--- a/internal/operator-controller/rukpak/convert/helm.go
+++ b/internal/operator-controller/rukpak/convert/helm.go
@@ -7,6 +7,7 @@ import (
 
 	"helm.sh/helm/v3/pkg/chart"
 
+	bundlepkg "github.com/operator-framework/operator-controller/internal/operator-controller/rukpak/bundle"
 	"github.com/operator-framework/operator-controller/internal/operator-controller/rukpak/bundle/source"
 	"github.com/operator-framework/operator-controller/internal/operator-controller/rukpak/render"
 )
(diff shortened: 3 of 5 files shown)


internal/operator-controller/rukpak/render/render.go
```diff
@@ -175,3 +202,55 @@ func validateTargetNamespaces(rv1 *bundle.RegistryV1, installNamespace string, t
...
+			errs = append(errs, fmt.Errorf("watchNamespace configuration parameter is required for bundles supporting SingleNamespace and OwnNamespace install modes"))
+		}
+	case !allSupported && !singleSupported && ownSupported:
+		// no All, no Single, Own: watchNamespace should be unknown/non-existent
+		if config.WatchNamespace != "" {
+			errs = append(errs, fmt.Errorf("watchNamespace configuration parameter is not supported for bundles that only support OwnNamespace install mode"))
+		}
+	case !allSupported && !singleSupported && !ownSupported:
+		// no All, no Single, no Own: invalid bundle
+		errs = append(errs, fmt.Errorf("invalid bundle: no supported install modes"))
```
joelanford: This case might be confusing for users if there are operators that support only MultiNamespace install mode.

Maybe we should have a `multiSupported` variable and a new case for "only `multiSupported`", so that we can explicitly say in this function that `"operators supporting only MultiNamespace install mode are not supported by OLMv1"`.

We might somehow preclude that elsewhere, but if we do, it seems like we should consolidate the installModes + targetNamespaces support into this function.

And then finally _this_ case would be augmented to include `!multiSupported`, at which point "invalid bundle: no supported install modes" makes more sense.

internal/operator-controller/rukpak/render/render_test.go
```diff
@@ -267,3 +267,111 @@ func Test_BundleValidatorCallsAllValidationFnsInOrder(t *testing.T) {
...
+			expectedErr: "watchNamespace configuration parameter is required for bundles supporting SingleNamespace and OwnNamespace install modes",
+		},
+		{
+			name:         "OwnNamespace only - rejects watchNamespace",
+			installModes: []v1alpha1.InstallModeType{v1alpha1.InstallModeTypeOwnNamespace},
+			config: &render.BundleConfig{
+				WatchNamespace: "test-namespace",
+			},
+			expectedErr: "watchNamespace configuration parameter is not supported for bundles that only support OwnNamespace install mode",
+		},
```
joelanford: Related to my other comment: add a test case with the following?
```
installModes: []v1alpha1.InstallModeType{v1alpha1.InstallModeTypeMultiNamespace}
```

internal/operator-controller/rukpak/render/render.go
```diff
@@ -175,3 +202,55 @@ func validateTargetNamespaces(rv1 *bundle.RegistryV1, installNamespace string, t
...
+		if config.WatchNamespace != "" {
+			errs = append(errs, fmt.Errorf("watchNamespace configuration parameter is not supported for bundles that only support AllNamespaces install mode"))
+		}
+	case allSupported && singleSupported && !ownSupported:
+		// All, Single, no Own: watchNamespace is optional
+		// No validation needed - any value is acceptable
+	case allSupported && !singleSupported && ownSupported:
+		// All, no Single, Own: watchNamespace is optional (must == install namespace when set)
+		if config.WatchNamespace != "" && config.WatchNamespace != installNamespace {
+			errs = append(errs, fmt.Errorf("watchNamespace must equal install namespace (%s) when set for bundles supporting AllNamespaces and OwnNamespace install modes", installNamespace))
```
joelanford: Nit: Up for debate, but my instinct for clarity is to structure these error messages like this:

"bundle \<bundleName> supports install modes \<supportedInstallModes>: watchNamespace must be \<allowedValues>"


NOTE: If the bundle name is already in some higher level error context as we return this error up the stack, then no need to repeat it here.

```suggestion
			errs = append(errs, fmt.Errorf("bundle %q supports install modes [AllNamespaces, OwnNamespace]; watchNamespace must be unset or equal to the install namespace (%q)", bundle.Name, installNamespace))
```

internal/operator-controller/rukpak/render/render.go
perdasilva: I don't think MultiNamespace support checks should be at this stage of the stack. I think its something that should be caught earlier when determining whether the bundle is valid/support at all. This area of the code should be more for checking that the configuration is sound. I find myself wondering if rather than having BundleConfig be just `WatchNamespace`, maybe we should have a discriminated union and the range of possible values for it it limited by the supported install modes (- MultiNamespace). E.g.

```
type BundleConfig struct {
  InstallMode: {
    InstallModeType: {AllNamespace, SingleNamespace, OwnNamespace},
    WatchNamespace string (only valid for SingleNamespace)
  }
}
```

wdyt?

internal/operator-controller/rukpak/render/render.go
perdasilva: If the mental model is one of configuration schema validation, should we rather have validation schema-ey type errors, e.g. "missing required parameter `watchNamespace`", "invalid value for parameter `watchNamespace` must be one of [...]", etc.?

anik120:
Carrying over a comment from slack to help other reviewers: 

> okay I pushed a change right now with `BundleConfig` converted to a discriminated union, PTAL.
And OpenAPI schema generation sounds cleaner. I kept validation in for now, but it makes sense to introduce OpenAPI schema generation and rely on that validation instead of writing custom logic

internal/operator-controller/rukpak/convert/helm.go
```diff
@@ -77,3 +94,55 @@ func (r *BundleToHelmChartConverter) ToHelmChart(bundle source.BundleSource, ins
 
 	return chrt, nil
 }
+
+// deriveDefaultWatchNamespace determines the default watch namespace based on bundle's supported install modes
+func (r *BundleToHelmChartConverter) deriveDefaultWatchNamespace(rv1 *bundle.RegistryV1, installNamespace string) string {
+	// Check what install modes the bundle supports
+	supportedModes := make(map[string]bool)
```
perdasilva: you can use `map[v1alpha1.InstallModeType]bool` (or use set.New[v1alpha1.InstallModeType]()))

internal/operator-controller/rukpak/convert/helm.go
```diff
@@ -77,3 +94,55 @@ func (r *BundleToHelmChartConverter) ToHelmChart(bundle source.BundleSource, ins
...
+	}
+	if supportedModes["OwnNamespace"] {
+		return installNamespace // OwnNamespace
+	}
+	if supportedModes["SingleNamespace"] {
+		return installNamespace // Treat as OwnNamespace since we don't know the target namespace
+	}
+
+	// Fallback to AllNamespaces
+	return ""
```
perdasilva: return an error if you get here. There should be no default for SingleNamespace or MultiNamespace. Maybe just say something like config parameter `watchNamespace` required

internal/operator-controller/rukpak/convert/helm.go
```diff
@@ -18,11 +19,27 @@ type BundleToHelmChartConverter struct {
...
+		config["watchNamespace"] = watchNamespace
+	}
+
 	rv1, err := bundle.GetBundle()
 	if err != nil {
 		return nil, err
 	}
 
+	if _, present := config["watchNamespace"]; !present {
+		config["watchNamespace"] = r.deriveDefaultWatchNamespace(&rv1, installNamespace)
```
perdasilva: in a comment below I suggest we return an error from `deriveDefaultWatchNamespace`. Check and return that error here...

internal/operator-controller/rukpak/convert/helm.go
```diff
@@ -18,11 +19,27 @@ type BundleToHelmChartConverter struct {
...
+	}
+
+	// Derive target namespaces from watchNamespace
+	targetNamespaces, err := r.deriveTargetNamespaces(config)
+	if err != nil {
+		return nil, fmt.Errorf("error deriving target namespaces: %w", err)
+	}
+
```
perdasilva: Instead of this, create an options slice and add the certificate provider and only add the TargetNamespace option if the configuration is defined.

internal/operator-controller/rukpak/convert/helm.go
```diff
@@ -77,3 +94,55 @@ func (r *BundleToHelmChartConverter) ToHelmChart(bundle source.BundleSource, ins
...
+	if supportedModes["SingleNamespace"] {
+		return installNamespace // Treat as OwnNamespace since we don't know the target namespace
+	}
+
+	// Fallback to AllNamespaces
+	return ""
+}
+
+// deriveTargetNamespaces converts config.watchNamespace to the target namespaces for the renderer
+func (r *BundleToHelmChartConverter) deriveTargetNamespaces(cfg map[string]interface{}) ([]string, error) {
```
perdasilva: I think we can remove this

internal/operator-controller/applier/helm.go
```diff
@@ -222,7 +223,10 @@ func (h *Helm) buildHelmChart(bundleFS fs.FS, ext *ocv1.ClusterExtension) (*char
 		}
 	}
 
-	return h.BundleToHelmChartConverter.ToHelmChart(source.FromFS(bundleFS), ext.Spec.Namespace, watchNamespace)
+	bundleConfig := map[string]interface{}{}
+	bundleConfig[bundle.BundleConfigWatchNamespaceKey] = watchNamespace
+	bundleConfig[bundle.BundleConfigInstallNamespaceKey] = ext.Spec.Namespace
```
perdasilva:
```suggestion
	bundleConfig := map[string]interface{}{
	    bundle.BundleConfigInstallNamespaceKey: ext.Spec.Namespace,
	    bundle.BundleConfigWatchNamespaceKey: watchNamespace,
	}
```

internal/operator-controller/rukpak/convert/helm_test.go
```diff
@@ -24,7 +24,10 @@ func Test_BundleToHelmChartConverter_ToHelmChart_ReturnsBundleSourceFailures(t *
 	var failingBundleSource FakeBundleSource = func() (bundle.RegistryV1, error) {
 		return bundle.RegistryV1{}, errors.New("some error")
 	}
-	_, err := converter.ToHelmChart(failingBundleSource, "install-namespace", "watch-namespace")
+	config := map[string]interface{}{}
+	config[bundle.BundleConfigInstallNamespaceKey] = "install-namespace"
+	config[bundle.BundleConfigWatchNamespaceKey] = "watch-namespace"
```
perdasilva: nit:
```suggestion
	config := map[string]interface{}{
	    bundle.BundleConfigInstallNamespaceKey: "install-namespace",
	    bundle.BundleConfigWatchNamespaceKey: "watch-namespace",
	}
```

same for the others

internal/operator-controller/rukpak/render/render.go
```diff
@@ -111,7 +124,7 @@ type BundleRenderer struct {
 	ResourceGenerators []ResourceGenerator
 }
 
-func (r BundleRenderer) Render(rv1 bundle.RegistryV1, installNamespace string, opts ...Option) ([]client.Object, error) {
+func (r BundleRenderer) Render(rv1 bundle.RegistryV1, opts ...Option) ([]client.Object, error) {
```
perdasilva: Let's not make install namespace optional here. We will _always_ need an install namespace. So, I'd revert the changes to this package.

internal/operator-controller/rukpak/convert/helm.go
```diff
@@ -17,12 +18,24 @@ type BundleToHelmChartConverter struct {
...
 
-func (r *BundleToHelmChartConverter) ToHelmChart(bundle source.BundleSource, installNamespace string, watchNamespace string) (*chart.Chart, error) {
+func (r *BundleToHelmChartConverter) ToHelmChart(bundle source.BundleSource, config map[string]interface{}) (*chart.Chart, error) {
 	rv1, err := bundle.GetBundle()
 	if err != nil {
 		return nil, err
 	}
 
+	// Convert config map to render options
+	var opts []render.Option
```
perdasilva: nit:

```suggestion
	opts := []render.Option{
	    render.WithCertificateProvider(r.CertificateProvider),
	}
```

then remove line 37

internal/operator-controller/rukpak/convert/helm.go
```diff
@@ -17,12 +18,24 @@ type BundleToHelmChartConverter struct {
...
+	if config != nil {
+		if installNs, ok := config[bundlepkg.BundleConfigInstallNamespaceKey].(string); ok {
```
perdasilva: we should return an error here as installNamespace is mandatory.

internal/operator-controller/rukpak/convert/helm.go
```diff
@@ -17,12 +18,24 @@ type BundleToHelmChartConverter struct {
...
+			opts = append(opts, render.WithInstallNamespace(installNs))
+		}
+		if watchNs, ok := config[bundlepkg.BundleConfigWatchNamespaceKey].(string); ok {
```
perdasilva, line 31: let's bring back deriveDefaults and set watchNamespaces to:
- corev1.NamespaceAll iff AllNamespaces is supported
- install-namespace iff the only supported install-mode is OwnNamespace
- an error is its not set and SingleNamespace is the only supported mode
- an error if there are no supported install modes

perdasilva: Axe this. I've moved this behavior into the renderer itself since it makes sense outside olmv1'isms anyway #2178

anik120: Ignoring this as this is being put in the render.

internal/operator-controller/rukpak/render/render.go
anik120: (discussed in slack)

internal/operator-controller/applier/helm_test.go
```diff
@@ -559,8 +559,8 @@ func TestApply_InstallationWithSingleOwnNamespaceInstallSupportEnabled(t *testin
 				},
 			},
 			BundleToHelmChartConverter: &fakeBundleToHelmChartConverter{
-				fn: func(bundle source.BundleSource, installNamespace string, watchNamespace string) (*chart.Chart, error) {
-					require.Equal(t, expectedWatchNamespace, watchNamespace)
+				fn: func(bundle source.BundleSource, installNamespace string, config map[string]interface{}) (*chart.Chart, error) {
+					require.Equal(t, expectedWatchNamespace, config["watchNamespace"])
```
perdasilva: nit: maybe use bundle.BundleConfigWatchNamespaceKey? or split out to a const in the test package?

anik120: Just called in the bundle package since that's what we're doing in the non-test portion too.

perdasilva (approved):
Awesome! Sorry for the churn, and thanks for this!

PR #2169, opened 2025-08-27
🐛 Fix: Truncate large error messages in status conditions (OCPBUGS-59518, OCPBUGS-38567)

camilamacedo86:
When upgrading operators, CRD validation errors can be very large (50KB+). Kubernetes rejects status updates over 32KB with "Too long: may not be more than 32768 bytes". This causes ClusterExtension upgrades to fail and get stuck.

Messages keep important info at the start and add "... [message truncated]" suffix. Now upgrades complete successfully even with large CRD validation errors.

Added unit tests for truncation logic and CRD error scenarios.

## Reviewer Checklist

- [N/A] API Go Documentation
- [X] Tests: Unit Tests (and E2E Tests, if appropriate)
- [X] Comprehensive Commit Messages
- [N/A ] Links to related GitHub Issue(s)

### DIFF
--- a/internal/operator-controller/controllers/clusterextension_controller.go
+++ b/internal/operator-controller/controllers/clusterextension_controller.go
@@ -22,7 +22,6 @@ import (
 	"fmt"
 	"io/fs"
 	"strings"
-	"time"
 
 	"github.com/go-logr/logr"
 	"helm.sh/helm/v3/pkg/release"
@@ -156,15 +155,13 @@ func ensureAllConditionsWithReason(ext *ocv1.ClusterExtension, reason v1alpha1.C
 		cond := apimeta.FindStatusCondition(ext.Status.Conditions, condType)
 		if cond == nil {
 			// Create a new condition with a valid reason and add it
-			newCond := metav1.Condition{
+			SetStatusCondition(&ext.Status.Conditions, metav1.Condition{
 				Type:               condType,
 				Status:             metav1.ConditionFalse,
 				Reason:             string(reason),
 				Message:            message,
 				ObservedGeneration: ext.GetGeneration(),
-				LastTransitionTime: metav1.NewTime(time.Now()),
-			}
-			ext.Status.Conditions = append(ext.Status.Conditions, newCond)
+			})
 		}
 	}
 }
@@ -381,7 +378,7 @@ func SetDeprecationStatus(ext *ocv1.ClusterExtension, bundleName string, depreca
 	if len(deprecationMessages) > 0 {
 		status, reason, message = metav1.ConditionTrue, ocv1.ReasonDeprecated, strings.Join(deprecationMessages, ";")
 	}
-	apimeta.SetStatusCondition(&ext.Status.Conditions, metav1.Condition{
+	SetStatusCondition(&ext.Status.Conditions, metav1.Condition{
 		Type:               ocv1.TypeDeprecated,
 		Reason:             reason,
 		Status:             status,
@@ -403,7 +400,7 @@ func SetDeprecationStatus(ext *ocv1.ClusterExtension, bundleName string, depreca
 				message = fmt.Sprintf("%s\n%s", message, entry.Message)
 			}
 		}
-		apimeta.SetStatusCondition(&ext.Status.Conditions, metav1.Condition{
+		SetStatusCondition(&ext.Status.Conditions, metav1.Condition{
 			Type:               conditionType,
 			Reason:             reason,
 			Status:             status,
--- a/internal/operator-controller/controllers/common_controller.go
+++ b/internal/operator-controller/controllers/common_controller.go
@@ -27,6 +27,30 @@ import (
 	ocv1 "github.com/operator-framework/operator-controller/api/v1"
 )
 
+const (
+	// maxConditionMessageLength set the max message length allowed by Kubernetes.
+	maxConditionMessageLength = 32768
+	// truncationSuffix is the suffix added when a message is cut.
+	truncationSuffix = "\n\n... [message truncated]"
+)
+
+// truncateMessage cuts long messages to fit Kubernetes condition limits
+func truncateMessage(message string) string {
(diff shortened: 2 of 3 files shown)


joelanford:
Can you include some details of the messages that are too long? I feel like arbitrarily truncating the message is sort of papering over the underlying issue, which is that 30k-byte messages in conditions are a poor UX, and the real solution would be to make the message shorter to begin with.

/hold

internal/operator-controller/controllers/common_controller.go
```diff
@@ -27,6 +27,23 @@ import (
 	ocv1 "github.com/operator-framework/operator-controller/api/v1"
 )
 
+const (
+	// maxConditionMessageLength is the Kubernetes limit minus some buffer for safety
+	maxConditionMessageLength = 30000
```
joelanford: If the maximum length is 32768 according to kubernetes validation, we can use all of that. There's no need to have an extra bit of space for safety.

internal/operator-controller/controllers/clusterextension_controller.go
```diff
@@ -160,7 +160,7 @@ func ensureAllConditionsWithReason(ext *ocv1.ClusterExtension, reason v1alpha1.C
 				Type:               condType,
 				Status:             metav1.ConditionFalse,
 				Reason:             string(reason),
-				Message:            message,
+				Message:            truncateMessage(message),
```
joelanford: If there's a limit imposed on all condition messages, it seems like we need to make sure that we truncate all condition messages.

This is one of many places where we set condition messages, right?

We may need to implement a wrapper around the `meta.SetCondition()` that:
1. truncates messages
2. everything throughout our project uses.

camilamacedo86:
Hi @joelanford 

Thank you for your fast review
I just pushed it :-)  

>  I feel like arbitrarily truncating the message is sort of papering over the underlying issue,

I remember when we discussed it in the past, the idea was just truncated
while I agree that we need to make the message better I think we should either truncate
I mean, we should do both

internal/operator-controller/controllers/common_controller.go
camilamacedo86: We need an extra for `truncationSuffix = "\n\n... [message truncated]"` at least

internal/operator-controller/controllers/clusterextension_controller.go
camilamacedo86: Yes, I think wrapper will be better as well +1

internal/operator-controller/controllers/common_controller.go
camilamacedo86: We can use 32.000

camilamacedo86:
Hi @joelanford 

While I agree that we need to fix the message in the first place, I also think we should truncate. So I’d like to propose doing both: truncate for safety (since it doesn’t make sense to fail because of this) and also fix the message itself.

I’ll prepare the message fix in a separate PR 🙂

internal/operator-controller/controllers/common_controller.go
camilamacedo86: Addressed

internal/operator-controller/controllers/clusterextension_controller.go
camilamacedo86: Done

joelanford:
/hold cancel

Thanks Camila! +1 on solving this both ways. Truncate long messages AND try to avoid long messages in the first place!

internal/operator-controller/controllers/common_controller.go
```diff
@@ -27,6 +27,31 @@ import (
 	ocv1 "github.com/operator-framework/operator-controller/api/v1"
 )
 
+const (
+	// maxConditionMessageLength is the Kubernetes API limit (32768) minus truncation suffix length (25 characters)
+	// This ensures final message with suffix never exceeds the API limit: 32768 - 25 = 32743
+	maxConditionMessageLength = 32743
```
perdasilva: If we already account for the suffix here, we will be truncating messages that are still less than the 32768 length.
I.e. we should only truncate if the message is > than the whole size limit (rather than size limit - suffix length).

On line 44 we already take the truncation suffix into account when generating the truncated message and ensuring its at most the max size.

joelanford: +1, we can set this:
```suggestion
	maxConditionMessageLength = 32768
```

joelanford:
> CRD validation errors can be very large (50KB+)

I've noticed this as well. Not only are they large, but the information density is low. IIRC, the CRD upgrade check output can contain a "diff-style" output that show context around the diff, which isn't really useful.

I think we should revisit our message formatting specifically in the case of the CRD upgrade check output to make it easier to understand:
1. Which field(s) is/are problematic?
2. Which changes were considered breaking?
3. (no other information)

internal/operator-controller/controllers/common_controller.go
camilamacedo86: You both are 1000% right
Fixed !

camilamacedo86:
Hi @joelanford @perdasilva 

I think it is OK for us to move forward, now, but in a follow-up, the error that we must solve is 

> "Reconciler error" err=<	validating upgrade for CRD "applicationsets.argoproj.io": v1alpha1: ^.spec.generators[*].matrix.generators: unhandled: unhandled changes found :

I updated : https://issues.redhat.com/browse/OCPBUGS-59518 with the details.

camilamacedo86:
Hi @joelanford @perdasilva 

I think it is good to get merged. I think we should truncate anyway. 
However, the real fix for the issue it seems for me to be: https://github.com/operator-framework/operator-controller/pull/2179

I updated the last comment in : https://issues.redhat.com/browse/OCPBUGS-59518

The problem here is `v1alpha1: ^.status.resources: unhandled: unhandled changes found :`

perdasilva (approved):

PR #2178, opened 2025-09-02
:sparkles: OPRUN-4113: Compute target namespace defaults

perdasilva:
# Description

Updates registry+v1 renderer to compute target namespace defaults based on the bundle's install mode configuration:

- bundle supports AllNamespaces: default is [""] (install in AllNamespaces mode)
- bundle supports OwnNamespace: default is [<install-namespace>]
- bundle supports neither AllNamespaces or OwnNamespace mode: undefined (user must specify TargetNamespaces)

We also update the targetNamespace validation to ensure the own namespace cannot be used unless OwnNamespace install mode is supported

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/operator-controller/rukpak/render/render.go
+++ b/internal/operator-controller/rukpak/render/render.go
@@ -85,9 +85,14 @@ func (o *Options) validate(rv1 *bundle.RegistryV1) (*Options, []error) {
 
 type Option func(*Options)
 
+// WithTargetNamespaces sets the target namespaces to be used when rendering the bundle
+// The value will only be used if len(namespaces) > 0. Otherwise, the default value for the bundle
+// derived from its install mode support will be used (if such a value can be defined).
 func WithTargetNamespaces(namespaces ...string) Option {
 	return func(o *Options) {
-		o.TargetNamespaces = namespaces
+		if len(namespaces) > 0 {
+			o.TargetNamespaces = namespaces
+		}
 	}
 }
 
@@ -149,6 +154,14 @@ func validateTargetNamespaces(rv1 *bundle.RegistryV1, installNamespace string, t
 	set := sets.New[string](targetNamespaces...)
 	switch {
 	case set.Len() == 0:
+		// Note: this function generally expects targetNamespace to contain at least one value set by default
+		// in case the user does not specify the value. The option to set the targetNamespace is a no-op if it is empty.
+		// The only case for which a default targetNamespace is undefined is in the case of a bundle that only
+		// supports SingleNamespace install mode. The if statement here is added to provide a more friendly error
+		// message than just the generic (at least one target namespace must be specified) which would occur
+		// in case only the MultiNamespace install mode is supported by the bundle.
+		// If AllNamespaces mode is supported, the default will be [""] -> watch all namespaces
+		// If only OwnNamespace is supported, the default will be [install-namespace] -> only watch the install/own namespace
 		if supportedInstallModes.Len() == 1 && supportedInstallModes.Has(v1alpha1.InstallModeTypeSingleNamespace) {
 			return errors.New("exactly one target namespace must be specified")
 		}
@@ -169,6 +182,9 @@ func validateTargetNamespaces(rv1 *bundle.RegistryV1, installNamespace string, t
 			return nil
 		}
 	default:
+		if !supportedInstallModes.Has(v1alpha1.InstallModeTypeOwnNamespace) && set.Has(installNamespace) {
+			return fmt.Errorf("supported install modes %v do not support targeting own namespace", sets.List(supportedInstallModes))
+		}
 		if supportedInstallModes.Has(v1alpha1.InstallModeTypeMultiNamespace) && !set.Has("") {
 			return nil
 		}
--- a/internal/operator-controller/rukpak/render/render.go
+++ b/internal/operator-controller/rukpak/render/render.go
@@ -4,7 +4,7 @@ import (
 	"errors"
 	"fmt"
 
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
@@ -74,9 +74,6 @@ func (o *Options) apply(opts ...Option) *Options {
 
 func (o *Options) validate(rv1 *bundle.RegistryV1) (*Options, []error) {
 	var errs []error
-	if len(o.TargetNamespaces) == 0 {
-		errs = append(errs, errors.New("at least one target namespace must be specified"))
-	}
 	if o.UniqueNameGenerator == nil {
 		errs = append(errs, errors.New("unique name generator must be specified"))
 	}
@@ -121,7 +118,7 @@ func (r BundleRenderer) Render(rv1 bundle.RegistryV1, installNamespace string, o
 	genOpts, errs := (&Options{
 		// default options
 		InstallNamespace:    installNamespace,
-		TargetNamespaces:    []string{metav1.NamespaceAll},
+		TargetNamespaces:    defaultTargetNamespacesForBundle(&rv1, installNamespace),
 		UniqueNameGenerator: DefaultUniqueNameGenerator,
 		CertificateProvider: nil,
 	}).apply(opts...).validate(&rv1)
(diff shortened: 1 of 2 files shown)


internal/operator-controller/rukpak/render/render.go
```diff
@@ -147,31 +144,52 @@ func DefaultUniqueNameGenerator(base string, o interface{}) (string, error) {
...
-		if supportedInstallModes.Has(string(v1alpha1.InstallModeTypeMultiNamespace)) && !set.Has("") {
+		if supportedInstallModes.Has(v1alpha1.InstallModeTypeMultiNamespace) && !set.Has("") {
 			return nil
 		}
 	}
-	return fmt.Errorf("supported install modes %v do not support target namespaces %v", sets.List[string](supportedInstallModes), targetNamespaces)
+	return fmt.Errorf("supported install modes %v do not support target namespaces %v", sets.List[v1alpha1.InstallModeType](supportedInstallModes), targetNamespaces)
+}
+
+func getDefaultTargetNamespaces(rv1 *bundle.RegistryV1, installNamespace string) []string {
```
anik120: nit: the usage of `get` is a Java (and languages of that time) paradigm..when I was moving to golang I was told "this is redundant, no need to write "get", coz there are no "set" operations to distinguish from" :)

perdasilva: hahaha - old Java programmer? I know him! He's me!

perdasilva: I've refactored the function name =D

internal/operator-controller/rukpak/render/render.go
```diff
@@ -147,31 +144,52 @@ func DefaultUniqueNameGenerator(base string, o interface{}) (string, error) {
...
+	}
+
+	if supportedInstallModes.Has(v1alpha1.InstallModeTypeOwnNamespace) {
+		return []string{installNamespace}
+	}
+
+	return nil
+}
+
+func getSupportedInstallModesSet(rv1 *bundle.RegistryV1) sets.Set[v1alpha1.InstallModeType] {
```
anik120:
```suggestion
func supportedInstallModes(rv1 *bundle.RegistryV1) sets.Set[v1alpha1.InstallModeType] {
```

also removed the name of the data structure from the function since we can look at the function name and determine, "it's a set"

internal/operator-controller/rukpak/render/render.go
anik120: It's me too! (started out with Java) 😁

internal/operator-controller/rukpak/render/render.go
```diff
@@ -147,31 +144,52 @@ func DefaultUniqueNameGenerator(base string, o interface{}) (string, error) {
...
+	case set.Len() == 1 && set.Has(""):
+		if supportedInstallModes.Has(v1alpha1.InstallModeTypeAllNamespaces) {
 			return nil
 		}
 		return fmt.Errorf("supported install modes %v do not support targeting all namespaces", sets.List(supportedInstallModes))
 	case set.Len() == 1 && !set.Has(""):
-		if supportedInstallModes.Has(string(v1alpha1.InstallModeTypeSingleNamespace)) {
+		if supportedInstallModes.Has(v1alpha1.InstallModeTypeSingleNamespace) {
 			return nil
 		}
```
joelanford: I'm just noticing, but this logic (both before and after) means that an operator could support _only_ `SingleNamespace`, and we'd allow that namespace to be the install namespace.

I don't think that's correct. We should only allow `watchNamespace == installNamespace` if the `OwnNamespace` install mode is supported.

internal/operator-controller/rukpak/render/render_test.go
```diff
@@ -42,23 +42,136 @@ func Test_BundleRenderer_ValidatesBundle(t *testing.T) {
...
+			validate: func(t *testing.T, opts render.Options) {
+				require.Equal(t, []string{expectedInstallNamespace}, opts.TargetNamespaces)
+			},
+		}, {
+			name:                   "returns error if target namespaces is not set bundle supports install modes [SingleNamespace]",
+			csv:                    MakeCSV(WithInstallModeSupportFor(v1alpha1.InstallModeTypeSingleNamespace)),
+			expectedErrMsgFragment: "at least one target namespace must be specified",
+		}, {
+			name:                   "returns error if target namespaces is not set bundle supports install modes [MultiNamespace]",
+			csv:                    MakeCSV(WithInstallModeSupportFor(v1alpha1.InstallModeTypeSingleNamespace)),
```
joelanford:
```suggestion
			csv:                    MakeCSV(WithInstallModeSupportFor(v1alpha1.InstallModeTypeMultiNamespace)),
```

internal/operator-controller/rukpak/render/render_test.go
```diff
@@ -42,23 +42,136 @@ func Test_BundleRenderer_ValidatesBundle(t *testing.T) {
...
+			name: "sets target namespaces to [install-namespace] by default if bundle supports install modes [OwnNamespace, SingleNamespace, MultiNamespace]",
+			csv:  MakeCSV(WithInstallModeSupportFor(v1alpha1.InstallModeTypeOwnNamespace, v1alpha1.InstallModeTypeSingleNamespace, v1alpha1.InstallModeTypeMultiNamespace)),
+			validate: func(t *testing.T, opts render.Options) {
+				require.Equal(t, []string{expectedInstallNamespace}, opts.TargetNamespaces)
+			},
+		}, {
+			name:                   "returns error if target namespaces is not set bundle supports install modes [SingleNamespace]",
+			csv:                    MakeCSV(WithInstallModeSupportFor(v1alpha1.InstallModeTypeSingleNamespace)),
+			expectedErrMsgFragment: "at least one target namespace must be specified",
+		}, {
```
joelanford: If the operator supports only `SingleNamespace` then it's a bit misleading to say "at least one target namespace...", right?
```suggestion
		}, {
			name:                   "returns error if target namespaces is not set bundle supports install modes [SingleNamespace]",
			csv:                    MakeCSV(WithInstallModeSupportFor(v1alpha1.InstallModeTypeSingleNamespace)),
			expectedErrMsgFragment: "exactly one target namespace must be specified",
		}, {
```

internal/operator-controller/rukpak/render/render_test.go
```diff
@@ -42,23 +42,136 @@ func Test_BundleRenderer_ValidatesBundle(t *testing.T) {
...
+			name:                   "returns error if target namespaces is not set bundle supports install modes [MultiNamespace]",
+			csv:                    MakeCSV(WithInstallModeSupportFor(v1alpha1.InstallModeTypeSingleNamespace)),
+			expectedErrMsgFragment: "at least one target namespace must be specified",
+		}, {
+			name:                   "returns error if target namespaces is not set bundle supports install modes [SingleNamespace, MultiNamespace]",
+			csv:                    MakeCSV(WithInstallModeSupportFor(v1alpha1.InstallModeTypeSingleNamespace)),
```
joelanford:
```suggestion
			csv:                    MakeCSV(WithInstallModeSupportFor(v1alpha1.InstallModeTypeSingleNamespace, v1alpha1.InstallModeTypeMultiNamespace)),
```

internal/operator-controller/rukpak/render/render.go
perdasilva: I was trying to look through the v0 codebase to ascertain this anb neither in the enhancement nor in code do I find this a reference to this. But, it's certainly possible (that it is and that I've missed it). Do you have a ref?

joelanford: I don't have a ref, but I do remember conversations about the distinction related to RBAC where some teams wanted to keep operator namespace separate from operand namespace.

I'll try to find a definitive ref. This is non-blocking since this PR doesn't change this logic at all.

perdasilva: Just let me know and I'll create another issue to address this. Should be straightforward enough. And a good one to fix before we go GA with Single/OwnNamespace

joelanford: Found it: https://github.com/operator-framework/api/blob/0433e5e6641217b79b29a90f4f2732ffab72ad74/pkg/operators/v1alpha1/clusterserviceversion.go#L179-L182

anik120 (approved):

internal/operator-controller/rukpak/render/render.go
perdasilva: thank you! I kept looking for references to the specific type. Lemme fix that up.

internal/operator-controller/rukpak/render/render.go
```diff
@@ -147,31 +144,58 @@ func DefaultUniqueNameGenerator(base string, o interface{}) (string, error) {
...
-		}
-	}
+	supportedInstallModes := supportedBundleInstallModes(rv1)
 
 	set := sets.New[string](targetNamespaces...)
 	switch {
-	case set.Len() == 0 || (set.Len() == 1 && set.Has("")):
-		if supportedInstallModes.Has(string(v1alpha1.InstallModeTypeAllNamespaces)) {
+	case set.Len() == 0:
+		if supportedInstallModes.Len() == 1 && supportedInstallModes.Has(v1alpha1.InstallModeTypeSingleNamespace) {
```
joelanford, line 165: Trying to make sense of this. Am I understanding correctly that `set.Len()` will only be `0` if we were unable to set a default target namespace, and _that_ only happens when neither `AllNamespaces` nor `OwnNamespace` are supported.

And therefore, by the time we get to _this_ code, the possibilities are:
- only `SingleNamespace`
- only `MultiNamespace`
- both `SingleNamespace` and `MultiNamespace`

And of those three cases, the only one that requires exactly one target namespace is "only `SingleNamespace`. Hence the `if` conditional you have here.

If that's correct. Can you add a comment that explains this a little bit? I don't think it is obvious why we aren't checking other supported install mode combinations.

joelanford: Also, what if someone explicitly uses option `WithTargetNamespaces(nil)` or `WithTargetNamespaces([]string{}])`?

The result of that would be:
1. The defaulting logic runs (setting a default in the case of All or Own namespace)
2. The functional option runs (unsetting the default)

So we could have `set.Len() == 0` for any combination of supported install modes unless we somehow require at least one namespace to be defined with that option.  What if we did this?

```
func WithTargetNamespaces(namespace string, extraNamespaces ...string) Option {
	return func(o *Options) {
		o.TargetNamespaces = append([]string{namespace}, extraNamespaces...)
	}
}
```

That way, using `WithTargetNamespaces` only accepts `1..N` namespaces, not `0..N`.

perdasilva: While I think this change in option does guarantee that at least one namespace is specified, it's annoying to use for the caller because it will force me to break up my slice of string into components. I won't be able to do targetNamespaces..., I'd need to do something like targetNamespace[0], targetNamespaces[1:]... but that will involve length checks, etc.

I think its ok for you to specify nil or empty...the validator will pick it up.

alternatively, we could just wrap an if around it and only update the value if len(namespaces) > 0? wdyt?

perdasilva: I went with my proposed solution above. I've also added more checks for MultiNamespace (no including install namespace if no OwnNamespace support)

joelanford: Good points! Your solution seems like a good compromise.

PR #2179, opened 2025-09-03
🐛 (fix): unhandle changes for crd upgrade safety ( OCPBUGS-59518 )

camilamacedo86:
## Problem  

We use the **crddiff**   (https://github.com/kubernetes-sigs/crdify) library to check API differences and ensure CRD upgrade safety. See: [CRD upgrade safety docs](https://docs.redhat.com/en/documentation/openshift_container_platform/4.19/html/extensions/cluster-extensions#crd-upgrade-safety)  

When crddiff cannot handle the diff, it reports an **unhandled scenario**.  
In this case, a **full JSON diff** is returned which **causes the problem**

- The JSON is **too large** to update in the CR status.  
- This causes an error:  `Too long: may not be more than 32768 bytes</code>`  

**-As a result:**  
  - **ClusterExtension is not upgraded**.  
  - **Users see no information** about why the upgrade failed.  
  - The only details are hidden in confusing logs in the operator-controller deployment, which are not meaningful or helpful.  

More info: [OCPBUGS-59518](https://issues.redhat.com/browse/OCPBUGS-59518)  

## Solution  

Instead of outputting the **full JSON**, we:  

- Extract only the **meaningful information**.  
- Return a **short, clear message** that explains the unhandled diff.  

This allows:  

- **OLMv1** to update the status conditions.  
- **Users** to understand why the CE upgrade could not happen.  

Example (from unit tests):  

```
require.ErrorContains(t, err, "unhandled changes found")  
require.ErrorContains(t, err, "Format \"\" -> \"email\"")  
```

This way:  
- Users get actionable feedback.  
- Status remains within size limits.
- The CE status will be updated as expected  

## Note  

As discussed, any further improvement beyond this would need to be done directly in the **crddiff** library, to reduce the number of unhandled scenarios.  

## Reviewer Checklist

- [N/A ] API Go Documentation
- [N/A] Links to related GitHub Issue(s)

### DIFF
--- a/internal/operator-controller/rukpak/preflights/crdupgradesafety/crdupgradesafety.go
+++ b/internal/operator-controller/rukpak/preflights/crdupgradesafety/crdupgradesafety.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	apiextensionsv1client "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1"
@@ -165,7 +166,7 @@ func sameVersionErrors(results *runner.Results) []error {
 				for _, err := range result.Errors {
 					msg := err
 					if result.Name == "unhandled" {
-						msg = "unhandled changes found"
+						msg = conciseUnhandledMessage(err)
 					}
 					errs = append(errs, fmt.Errorf("%s: %s: %s: %s", version, property, result.Name, msg))
 				}
@@ -188,7 +189,7 @@ func servedVersionErrors(results *runner.Results) []error {
 				for _, err := range result.Errors {
 					msg := err
 					if result.Name == "unhandled" {
-						msg = "unhandled changes found"
+						msg = conciseUnhandledMessage(err)
 					}
 					errs = append(errs, fmt.Errorf("%s: %s: %s: %s", version, property, result.Name, msg))
 				}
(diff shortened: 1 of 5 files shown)


camilamacedo86:
Hi @joelanford @perdasilva 

It seems to be the real fix for OCPBUGS-59518.
I hope that helps out. 

Moreover, IMHO we should perform additional checks and carefully evaluate the errors in those unhandled scenarios. It might be worth verifying if a newer version of the library addresses this, or if there are other improvements we could apply. In short, we should confirm whether there’s truly nothing else we can do here to make the handling more robust.

internal/operator-controller/rukpak/preflights/crdupgradesafety/crdupgradesafety.go
```diff
@@ -171,7 +172,11 @@ func sameVersionErrors(results *runner.Results) []error {
 		for property, comparisonResults := range propertyResults {
 			for _, result := range comparisonResults {
 				for _, err := range result.Errors {
-					errs = append(errs, fmt.Errorf("%s: %s: %s: %s", version, property, result.Name, err))
+					msg := err
+					if result.Name == "unhandled" {
+						msg = "unhandled changes found"
+					}
+					errs = append(errs, fmt.Errorf("%s: %s: %s: %s", version, property, result.Name, msg))
```
perdasilva, line 171: it feels like we might be occluding the error. Is it possible to format it it in a better way? Then maybe at the end we limit the number of errors we print out?

camilamacedo86: I am not sure if I could understand your suggestion here. 
Could you please clarify?

perdasilva: The main point  is that, the way it is, it's kinda just an HTTP 500 without any explanation. As a user, I don't know what to do with that. Maybe this is an issue with the underlying library we are using and the fix needs to go there. But as a user, I would like to know: what, where, why, and how to fix it. 

The second part of the comment can be ignored as a brain fart. I was in the context of the "we're generating too many errors" problem.

camilamacedo86: I could improve by extracting info from the JSON. 
See wdyt now

camilamacedo86: As we discussed, it looks like an unexpected error. Maybe we could explore whether the library can handle fewer scenarios as unhandled cases. That said, you raise a great point — I was able to extract some information from the JSON, and I think this is the best we can do for now. What do you think?

when the error occurs, we can’t update the CR with the status. This leaves the user without any visibility into why the upgrade didn’t happen. Here’s the bug for reference: https://issues.redhat.com/browse/OCPBUGS-59518 so this PR should fix it.

anik120 (commented):
@camilamacedo86 is it possible to get a 

```
Problem :  __________________________ (eg before this PR, crd upgrade safety feature did not do _____ right)
Solution: _________________________ 
```

statement in the description? 

Helps with reviewing the PR.

camilamacedo86:
@anik120 can you please check the description?

internal/operator-controller/rukpak/preflights/crdupgradesafety/crdupgradesafety.go
anik120: +1 on checking in with the library to see if we can get this handled there. This looks like a lot of maintenance burden for us to take on that'll eventually (hopefully) be tech debt.

anik120:
> @anik120 can you please check the description?

Thank you for that! Really helped with mind mapping what I'm reviewing 🙏🏽

internal/operator-controller/rukpak/preflights/crdupgradesafety/crdupgradesafety.go
camilamacedo86: HI @anik120 

We MUST do this change to deal with unhandled changes. **note that it is a BUG fix**
Any further improvements than is in the lib. That means, if we want the library to handle more scenarios that today are unhandled, then we must propose it there and see if they accept.

PR #2195, opened 2025-09-05
✨migrate containers libs to new mono-repo

anik120:
Ref: https://blog.podman.io/2025/08/upcoming-migration-of-three-containers-repositories-to-monorepo/

Also note, the containers mono-repo has introduced new policy restrictions that were preventing the test containers from running.

The failing tests wre in:
 - internal/shared/util/image/pull_test.go:163
 - Multiple test cases in TestContainersImagePuller_Pull

To fix it, explicit insecure signature policy configuration was added to the test environment by

- Updating `buildSourceContextFunc` to create a `policy.json` file with `{"default":[{"type":"insecureAcceptAnything"}]}`
- Setting `SignaturePolicyPath` in test `SystemContext` to allow test image pulls without signature verification

# Description

## Reviewer Checklist

- [ ] API Go Documentation


internal/shared/util/image/pull_test.go
```diff
@@ -298,8 +298,14 @@ func buildSourceContextFunc(t *testing.T, ref reference.Named) func(context.Cont
...
 		require.NoError(t, f.Close())
 
+		// Create an insecure signature policy for tests
+		policyPath := filepath.Join(configDir, "policy.json")
+		policyContent := `{"default":[{"type":"insecureAcceptAnything"}]}`
+		require.NoError(t, os.WriteFile(policyPath, []byte(policyContent), 0600))
+
 		return &types.SystemContext{
 			SystemRegistriesConfPath: registriesConfPath,
+			SignaturePolicyPath:      policyPath,
```
joelanford: Did something change in 5.37.0 to require this?

anik120: @joelanford unfortunately there are no release notes for the new tag. Looking at the commits, I can see that there's a [new image verification implementation as the verification backend](https://github.com/containers/container-libs/commit/07a9fbc28b7e2ce07aa82c1ab437ff3387743a91). 

The new implementation most likely has stricter signature verification policies, requiring the explicit policy configuration. 

ref: https://sequoia-pgp.org/blog/2023/04/27/rpm-sequoia/

tmshort: Should this have been a build-tag change, as we do for `containers_image_openpgp`?
https://github.com/operator-framework/operator-controller/blob/1ab053e4abf714ebca58d03c69dab8bb47a2ff7f/Makefile#L385

tmshort:
/approve
There's still a question on the `SecurityPolicyPath` change, but this is something we should do.

internal/shared/util/image/pull_test.go
joelanford: I thought I still saw an openpgp implementation in there. I'll take a closer look.

anik120: That build tag controls whether openpgp signature verification is available. This looks like a new backend completely (replacing openpgp), that seems to be have introduced a stricter, "require explicit policy file by default' requirement. 

Regardless of the signature mechanism used though, this is basically just adding an explicit policy configuration, which looks like good practice anyway.

joelanford: I checked out your branch, reverted these lines, ran locally, and it seems like unit tests are passing for me without this change.

joelanford: Can you revert these lines and push up to see what GH CI tells us?

joelanford: More background info:

https://github.com/containers/container-libs/tree/main/image#supported-build-tags
[12:22](https://redhat-internal.slack.com/archives/C06KP34REFJ/p1757521335427249?thread_ts=1757016644.293679&cid=C06KP34REFJ)

That link implies that we are still using openpgp since we did not change our build tag.

Also, https://github.com/containers/container-libs/blob/4c84c257562b37a8a106c439b351118b2313179e/image/signature/policy_config.go#L52
I know we may be hardcoding in other places already, but I'd like to avoid propagating that, at least without having a really good understanding of why it is necessary.

anik120: yea looks like you're right @joelanford, not sure why the tests were failing for me locally before...in any case these lines are out now and the tests are passing

PR #2200, opened 2025-09-09
CER: centralize status updates into big-R Reconcile method

joelanford:
Re-opening here now that poc-boxcutter PR is merged

Original PR: https://github.com/thetechnick/operator-controller/pull/5

# Description

A few small fixups and comment additions to help remember follow-ups:
1. Make `Reconcile` and `reconcile` more consistent between CE and CER reconcilers
   - update status in one place in `Reconcile`
   - only log reconciler start/stop if we actually call `reconcile`

NOTE: There is still one major difference: CER updates finalizers as separate patches and continues reconcile logic. CE updates finalizers and requires re-reconcile to continue beyond. The CER logic may cause issues where the _next_ reconcile sees a stale resource version (from finalizer patch) because the status update event isn't seen by our informer in time. This ultimately leads to reconciling a stale CER that manifests as a conflict error in our logs that is often a red herring for users trying to troubleshoot issues in controllers.

Another noticeable difference is that CE uses controller-runtime Finalizers helpers, and CER handles things manually. 

Changing this broke unit tests, so I decided not to leave that change out to keep the PR scope small. I'll make a separate PR that focuses just on finalizers.

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/operator-controller/controllers/clusterextensionrevision_controller.go
+++ b/internal/operator-controller/controllers/clusterextensionrevision_controller.go
@@ -12,6 +12,7 @@ import (
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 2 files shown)


internal/operator-controller/controllers/clusterextensionrevision_controller.go
```diff
@@ -114,9 +150,19 @@ func (c *ClusterExtensionRevisionReconciler) reconcile(ctx context.Context, rev
...
+			meta.SetStatusCondition(&rev.Status.Conditions, metav1.Condition{
+				Type:               "Available",
+				Status:             metav1.ConditionFalse,
+				Reason:             "ReconcileFailure",
+				Message:            err.Error(),
+				ObservedGeneration: rev.Generation,
+			})
+			return ctrl.Result{}, fmt.Errorf("error removing teardown finalizer: %v", err)
+		}
+		return ctrl.Result{}, nil
```
thetechnick, line 165: It seems semantically wrong to mark a Revision as "Available" == False for any issue with finalizers.

In general I would not report any internal error as a Status Condition.
If we want to show individual errors to users we should publish events on the resource.

joelanford: Totally agree that this is likely the wrong type for this. I would expect to come back and clean this up prior to GA once there is a formal design on status handling.

In the meantime, I do think it is an important UX concern to tell the user _something_ in a place they are likely to see it.

If they've deleted a CER (or an agent has on their behalf), but the finalizer isn't being removed for some reason, it would be a bad to UX for the CER to just... have no update. It would look like we just aren't reconciling it.

PR #2213, opened 2025-09-11
:seedling: OPRUN-4077: Remove the kustomize config

tmshort:
Add a helm/OWNERS file (copied from config/OWNERS) Update hack/tools/update-crds.sh to not reference config directory

# Description

## Reviewer Checklist

- [ ] API Go Documentation


tmshort:
/hold
This _will_ seriously break the downstream bumper, because the manifest source will have changed, and the code to handle the new config is in a late carry commit. This might trigger us to consolidate all the carry commits (something that we acknowledge we have to do occasionally).

Makefile
```diff
@@ -204,8 +204,8 @@ bingo-upgrade: $(BINGO) #EXHELP Upgrade tools
 .PHONY: verify-crd-compatibility
 CRD_DIFF_ORIGINAL_REF := git://main?path=
 CRD_DIFF_UPDATED_REF := file://
-CRD_DIFF_OPCON_SOURCE := config/base/operator-controller/crd/standard/olm.operatorframework.io_clusterextensions.yaml
-CRD_DIFF_CATD_SOURCE := config/base/catalogd/crd/standard/olm.operatorframework.io_clustercatalogs.yaml
+CRD_DIFF_OPCON_SOURCE := helm/olmv1/base/operator-controller/crd/standard/olm.operatorframework.io_clusterextensions.yaml
+CRD_DIFF_CATD_SOURCE := helm/olmv1/base/catalogd/crd/standard/olm.operatorframework.io_clustercatalogs.yaml
```
camilamacedo86, line 203: Why do we actually need the base?
It feels more like a kustomize-style layout.

Since everything is already managed under Helm, why wouldn’t those manifests live under templates instead?
Also, I noticed we have CRDs under templates — shouldn’t this be consistent?

tmshort: The `base` directory is storing the output of the CRDs - these is the "raw" CRDs. This is kept outside of the `templates` directory and is purposely _not_ put into a `crds` directory.

Because the CRDs are be conditionally included into the manifest (i.e. standard vs experimental), and because they are autogenerated, and because we don't want to go through a messy process of post-processing them to include Helm directives, they are put into that `base` directory.

So, the CRDs _are not_ under `templates` directory, but instead those files in the `templates/crds` directory conditionally _include_ CRDs from the `base` directory:

https://github.com/operator-framework/operator-controller/blob/16d1089acce140e309cebe99bc9bd87c4ba7f913/helm/olmv1/templates/crds/customresourcedefinition-clustercatalogs.olm.operatorframework.io.yml#L1-L9

https://github.com/operator-framework/operator-controller/blob/16d1089acce140e309cebe99bc9bd87c4ba7f913/helm/olmv1/templates/crds/customresourcedefinition-clusterextensionrevisions.olm.operatorframework.io.yml#L1-L9

(i.e. We don't want to do any post-processing of the CRDs to include the above.)

This PR, because it's removing the `config` directory, is merely changing the location of where the "raw" CRDs live. It is _NOT_ changing the process of how they are generated nor included into the Helm manifest.

dtfranz:
Should we address [these](https://github.com/operator-framework/operator-controller/blob/47f8d312187f9f6a41118ee1c5c2f11c09afb7ce/Makefile#L74) as well while we're at it?

`Makefile`:
```
KUSTOMIZE_STANDARD_OVERLAY := config/overlays/standard
KUSTOMIZE_STANDARD_E2E_OVERLAY := config/overlays/standard-e2e
KUSTOMIZE_EXPERIMENTAL_OVERLAY := config/overlays/experimental
KUSTOMIZE_EXPERIMENTAL_E2E_OVERLAY := config/overlays/experimental-e2e
```

These [urls](https://github.com/operator-framework/operator-controller/blob/47f8d312187f9f6a41118ee1c5c2f11c09afb7ce/docs/draft/api-reference/network-policies.md?plain=1#L7) will need to be updated too so they don't break:
`docs/draft/api-reference/network-policies.md`:
```
* The catalogd NetworkPolicy is implemented [here](https://github.com/operator-framework/operator-controller/blob/main/config/base/catalogd/manager/network_policy.yaml).
* The operator-controller is implemented [here](https://github.com/operator-framework/operator-controller/blob/main/config/base/operator-controller/manager/network_policy.yaml).
```

PR #2222, opened 2025-09-16
:sparkles: OPRUN-4150: Relax webhook preconditions

perdasilva:
# Description

In OLMv0, only bundles with conversion webhooks must only support AllNamespaces install mode. In the initial implementation of the v1 support, we temporarily extended this to all webhook types. This PR brings v1 inline with v0 webhook support.

## Reviewer Checklist

- [ ] API Go Documentation


internal/operator-controller/rukpak/render/registryv1/validators/validator.go
```diff
@@ -101,21 +101,26 @@ func CheckPackageNameNotEmpty(rv1 *bundle.RegistryV1) []error {
...
-// with OLMv0 behavior for conversion webhooks,
+// CheckConversionWebhookSupport checks that if the bundle cluster service version declares conversion webhook definitions
+// that it only supports AllNamespaces mode. This keeps parity with OLMv0 behavior for conversion webhooks,
 // https://github.com/operator-framework/operator-lifecycle-manager/blob/dfd0b2bea85038d3c0d65348bc812d297f16b8d2/pkg/controller/install/webhook.go#L193
-// Since OLMv1 considers APIs to be cluster-scoped, we initially extend this constraint to validating and mutating webhooks.
-// While this might restrict the number of supported bundles, we can tackle the issue of relaxing this constraint in turn
-// after getting the webhook support working.
-func CheckWebhookSupport(rv1 *bundle.RegistryV1) []error {
-	if len(rv1.CSV.Spec.WebhookDefinitions) > 0 {
+func CheckConversionWebhookSupport(rv1 *bundle.RegistryV1) []error {
```
pedjak, line 107: can we simplify the the return value, and return just single `error`? When we could expect multiple errors, i.e. could we even then use `errors.Joint(...)` to concatenate them?

internal/operator-controller/rukpak/render/registryv1/validators/validator.go
```diff
@@ -101,21 +101,26 @@ func CheckPackageNameNotEmpty(rv1 *bundle.RegistryV1) []error {
...
+	hasConversionWebhooks := false
+	for _, wh := range rv1.CSV.Spec.WebhookDefinitions {
+		if wh.Type == v1alpha1.ConversionWebhook {
+			hasConversionWebhooks = true
+			break
+		}
+	}
+
+	if hasConversionWebhooks {
```
pedjak: * what happens if CSV contains non-conversion webhooks as well?
* what happens if CSV contains more than one conversion webhooks?

internal/operator-controller/rukpak/render/registryv1/validators/validator.go
```diff
@@ -101,21 +101,26 @@ func CheckPackageNameNotEmpty(rv1 *bundle.RegistryV1) []error {
...
 		supportedInstallModes := sets.Set[v1alpha1.InstallModeType]{}
 		for _, mode := range rv1.CSV.Spec.InstallModes {
 			supportedInstallModes.Insert(mode.Type)
 		}
 		if len(supportedInstallModes) != 1 || !supportedInstallModes.Has(v1alpha1.InstallModeTypeAllNamespaces) {
-			return []error{errors.New("bundle contains webhook definitions but supported install modes beyond AllNamespaces")}
+			sortedModes := slices.Sorted(slices.Values(supportedInstallModes.UnsortedList()))
+			return []error{fmt.Errorf("bundle has conversion webhooks with unsupported install modes %v bundle with conversion webhooks must support only AllNamespaces install mode", sortedModes)}
```
pedjak: how about to simplify the error message to something like:
```suggestion
bundle contains conversion webhooks %s with unsupported install modes %v - only AllNamespaces is supported
```

internal/operator-controller/rukpak/render/registryv1/validators/validator.go
perdasilva: - No problem, we support those independent of which install modes are supported by the bundle
- If |conversionWebhooks| >= 1 => no support. So, it's sufficient to bail after finding the first.

internal/operator-controller/rukpak/render/registryv1/validators/validator.go
perdasilva: The current signature for the validation function returns []error. If it's something we'd like to change, I'd say we do it in a separate PR because it would touch every validation function. I think the advantage of having it return an error slice is that it makes it clear to whomever is implementing the next validation function that they should collect as many errors as possible before returning.

internal/operator-controller/rukpak/render/registryv1/validators/validator.go
perdasilva: This one was hard to word. There are two concepts here:
1. The bundle has conversion webhook definitions
2. The bundle's supported install modes != [AllNamespaces]

Conversion webhooks are only supported for bundles that only support the AllNamespaces install mode

perdasilva: gonna try and reword it - sorry for the confusion

internal/operator-controller/rukpak/render/registryv1/validators/validator.go
perdasilva: I've updated the doc comment to hopefully better describe the function

pedjak: > If |conversionWebhooks| >= 1 => no support. So, it's sufficient to bail after finding the first.

Where do we check for the number of conversion webhooks?

internal/operator-controller/rukpak/render/registryv1/validators/validator.go
```diff
@@ -101,21 +101,26 @@ func CheckPackageNameNotEmpty(rv1 *bundle.RegistryV1) []error {
...
+		if wh.Type == v1alpha1.ConversionWebhook {
+			hasConversionWebhooks = true
+			break
+		}
+	}
+
+	if hasConversionWebhooks {
 		supportedInstallModes := sets.Set[v1alpha1.InstallModeType]{}
 		for _, mode := range rv1.CSV.Spec.InstallModes {
 			supportedInstallModes.Insert(mode.Type)
```
pedjak: while tweaking this function, I think this block could be simplified to:

```golang
supportedInstallModes := sets.New(rv1.CSV.Spec.InstallModes...)
```

and then we can query it like:

````golang
!supportedInstallModes.Has(v1alpha1.InstallMode{Supported: true, Type: v1alpha1.InstallModeTypeAllNamespaces}
````

Supported should be set to `true`, right - otherwise, `AllNamespaces` could be listed as install mode, but if `Supported=false` it does not count, if I am not mistaken?

File: internal/operator-controller/rukpak/render/registryv1/validators/validator.go (hunk 2 above)
Comment (perdasilva) [reply internal/operator-controller/rukpak/render/registryv1/validators/validator.go]: we check for existence rather than the exact number, i.e. if any of the webhooks have type `v1alpha1.ConversionWebhook`. Or are you saying you'd prefer multiple errors 1 for each conversion webhook?

File: internal/operator-controller/rukpak/render/registryv1/validators/validator.go (hunk 4 above)
Comment (perdasilva) [reply internal/operator-controller/rukpak/render/registryv1/validators/validator.go]: I've addressed the lack of check for `mode.Supported` but left the overall logic in the same way in order to be able to easily grab the list of supported install modes.

File: internal/operator-controller/rukpak/render/registryv1/validators/validator.go (hunk 3 above)
Comment (perdasilva) [reply internal/operator-controller/rukpak/render/registryv1/validators/validator.go]: I've reworked the logic to include an error for each conversion webhook

File: internal/operator-controller/rukpak/render/registryv1/registryv1.go [hunk 5]
```diff
@@ -22,6 +22,7 @@ var BundleValidator = render.BundleValidator{
 	validators.CheckCRDResourceUniqueness,
 	validators.CheckOwnedCRDExistence,
 	validators.CheckPackageNameNotEmpty,
+	validators.CheckConversionWebhookSupport,
```
pedjak, line 25: shall we add a line in PR description about it?

internal/operator-controller/rukpak/render/registryv1/validators/validator.go
```diff
@@ -101,21 +101,33 @@ func CheckPackageNameNotEmpty(rv1 *bundle.RegistryV1) []error {
...
+
+	if len(conversionWebhookNames) > 0 {
 		supportedInstallModes := sets.Set[v1alpha1.InstallModeType]{}
 		for _, mode := range rv1.CSV.Spec.InstallModes {
-			supportedInstallModes.Insert(mode.Type)
+			if mode.Supported {
+				supportedInstallModes.Insert(mode.Type)
+			}
 		}
+		slices.Sort(conversionWebhookNames)
```
pedjak: we need sort them only if errors need to be returned. Also, not sure if we need to sort them, return errors in order of webhook appearance in the CSV is also fine, IMHO.

perdasilva: fair call - I've remove it

internal/operator-controller/rukpak/render/registryv1/validators/validator_test.go
```diff
@@ -314,25 +287,32 @@ func Test_CheckWebhookSupport(t *testing.T) {
...
 			},
-			expectedErrs: []error{errors.New("bundle contains webhook definitions but supported install modes beyond AllNamespaces")},
 		},
 		{
-			name: "accepts bundles with conversion webhook definitions when they support more modes than AllNamespaces install mode",
+			name: "rejects bundles with conversion webhook definitions when they support more modes than AllNamespaces install mode",
 			bundle: &bundle.RegistryV1{
 				CSV: MakeCSV(
-					WithInstallModeSupportFor(v1alpha1.InstallModeTypeAllNamespaces, v1alpha1.InstallModeTypeSingleNamespace),
+					WithInstallModeSupportFor(v1alpha1.InstallModeTypeSingleNamespace, v1alpha1.InstallModeTypeAllNamespaces),
```
pedjak, line 295: why did we flip the order of install modes in the fixture? Is this important for the test?

internal/operator-controller/rukpak/render/registryv1/validators/validator.go
```diff
@@ -101,21 +101,32 @@ func CheckPackageNameNotEmpty(rv1 *bundle.RegistryV1) []error {
...
+		if wh.Type == v1alpha1.ConversionWebhook {
+			conversionWebhookNames = append(conversionWebhookNames, wh.GenerateName)
+		}
+	}
+
+	if len(conversionWebhookNames) > 0 {
 		supportedInstallModes := sets.Set[v1alpha1.InstallModeType]{}
 		for _, mode := range rv1.CSV.Spec.InstallModes {
-			supportedInstallModes.Insert(mode.Type)
+			if mode.Supported {
```
pedjak, line 118: do we have unit tests for this path?

camilamacedo86 (approved):

internal/operator-controller/rukpak/render/registryv1/validators/validator_test.go
perdasilva: I wanted to exercise the sorting

PR #2224, opened 2025-09-16
⚠ OPRUN-4106: remove support for annotation based config

anik120:
# Description

With https://github.com/operator-framework/operator-controller/pull/2166 and https://github.com/operator-framework/operator-controller/pull/2163 in place, annotation based config support can be removed.

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/hack/demo/resources/own-namespace-demo.yaml
+++ b/hack/demo/resources/own-namespace-demo.yaml
@@ -2,9 +2,6 @@ apiVersion: olm.operatorframework.io/v1
 kind: ClusterExtension
 metadata:
   name: argocd-operator
-  annotations:
-    # watch namespace is the same as intall namespace
-    olm.operatorframework.io/watch-namespace: argocd-system
 spec:
   namespace: argocd-system
   serviceAccount:
--- a/docs/draft/howto/single-ownnamespace-install.md
+++ b/docs/draft/howto/single-ownnamespace-install.md
@@ -91,8 +91,6 @@ apiVersion: olm.operatorframework.io/v1
 kind: ClusterExtension
 metadata:
   name: argocd
-  annotations:
-    olm.operatorframework.io/watch-namespace: argocd
 spec:
   namespace: argocd
   serviceAccount:
--- a/hack/demo/own-namespace-demo-script.sh
+++ b/hack/demo/own-namespace-demo-script.sh
@@ -3,7 +3,14 @@
 #
 # Welcome to the OwnNamespace install mode demo
 #
-trap "trap - SIGTERM && kill -- -$$" SIGINT SIGTERM EXIT
+set -e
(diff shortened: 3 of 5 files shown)


internal/operator-controller/applier/watchnamespace_test.go
```diff
@@ -71,59 +71,54 @@ func TestGetWatchNamespace(t *testing.T) {
...
+							Raw: []byte(`{"watchNamespace":"watch-namespace"}`),
+						},
 					},
 				},
 			},
 			expectError: false,
 		}, {
-			name: "cluster extension configures a watch namespace through annotation with invalid ns",
+			name: "cluster extension configures a watch namespace through spec.config.inline with invalid ns",
 			csv: &ocv1.ClusterExtension{
```
joelanford: Unrelated to this PR, but seems like a typo that we named the CE field `csv`.

So, nit: `s/csv/ce/g`

hack/demo/resources/own-namespace-demo.yaml
```diff
@@ -2,13 +2,13 @@ apiVersion: olm.operatorframework.io/v1
...
 metadata:
   name: argocd-operator
-  annotations:
-    # watch namespace is the same as intall namespace
-    olm.operatorframework.io/watch-namespace: argocd-system
 spec:
   namespace: argocd-system
   serviceAccount:
     name: argocd-installer
+  config:
```
perdasilva: There's some more nuance now with the defaults configuration. Depending on the install modes supported, we'd have different behaviors here. E.g. if the bundle only supports OwnNamespace, you wouldn't specify the watchNamespace. Have you checked the demos still work?

internal/operator-controller/applier/watchnamespace_test.go
```diff
@@ -71,59 +71,54 @@ func TestGetWatchNamespace(t *testing.T) {
 			},
 			expectError: false,
 		}, {
-			name: "cluster extension configures a watch namespace through annotation",
+			name: "cluster extension configures a watch namespace through spec.config.inline",
```
perdasilva: do we already have a unit test under this banner? should we just remove this one?

internal/operator-controller/applier/watchnamespace_test.go
anik120: Went ahead and removed it anyway.

hack/demo/resources/own-namespace-demo.yaml
anik120: Thanks for pointing this out. I've made a few extra changes and put them in a separate commit https://github.com/operator-framework/operator-controller/pull/2224/commits/53b0f1a300099fa2109367d44e4ea92e8b3d7e5b, demos work perfectly now.

internal/operator-controller/applier/watchnamespace_test.go
anik120: Yup, looks like we do. Removed this.

PR #2226, opened 2025-09-17
:sparkles: Add webhook rule validation

perdasilva:
# Description

I recently discovered another boundary for webhook support [[link](https://github.com/operator-framework/operator-lifecycle-manager/blob/ccf0c4c91f1e7673e87f3a18947f9a1f88d48438/pkg/controller/install/webhook.go#L19)] around the api groups and resources defined in the rules.

This PR adds that check to the bundle validator.

## Reviewer Checklist

- [ ] API Go Documentation


perdasilva:
Oh, whoops, I need to sub out the olmv0 api group with the olmv1 one

internal/operator-controller/rukpak/render/registryv1/validators/validator.go
```diff
@@ -264,3 +264,50 @@ func CheckWebhookNameIsDNS1123SubDomain(rv1 *bundle.RegistryV1) []error {
 	}
 	return errs
 }
+
+// unsupportedWebhookRuleAPIGroups contain the API groups that are unsupported for webhook configuration rules in OLMv1
+var unsupportedWebhookRuleAPIGroups = []string{
```
pedjak: how about that we define these two arrays actually as sets, something like: 

```golang
var unsupportedWebhookRuleAPIGroups = sets.New("olm.operatorframework.io", "*")
```

IMHO, it would enable easier reading of `CheckWebhookRules` function later on.

internal/operator-controller/rukpak/render/registryv1/validators/validator.go
```diff
@@ -264,3 +264,50 @@ func CheckWebhookNameIsDNS1123SubDomain(rv1 *bundle.RegistryV1) []error {
...
+	for _, wh := range rv1.CSV.Spec.WebhookDefinitions {
+		// Rules are not used for conversion webhooks
+		if wh.Type == v1alpha1.ConversionWebhook {
+			continue
+		}
+		webhookName := wh.GenerateName
+		for _, rule := range wh.Rules {
+			for _, apiGroup := range rule.APIGroups {
+				if slices.Contains(unsupportedWebhookRuleAPIGroups, apiGroup) {
+					errs = append(errs, fmt.Errorf("webhook %q contains unsupported rule: unsupported API group %q", webhookName, apiGroup))
```
pedjak: is the rule unsupported or forbidden?

internal/operator-controller/rukpak/render/registryv1/validators/validator.go
```diff
@@ -264,3 +264,50 @@ func CheckWebhookNameIsDNS1123SubDomain(rv1 *bundle.RegistryV1) []error {
...
+				}
+				if apiGroup == "admissionregistration.k8s.io" {
+					for _, resource := range rule.Resources {
+						if slices.Contains(unsupportedAdmissionRegistrationResources, strings.ToLower(resource)) {
+							errs = append(errs, fmt.Errorf("webhook %q contains unsupported rule: unsupported resource %q for API group %q", webhookName, resource, apiGroup))
```
pedjak: same question as above?

internal/operator-controller/rukpak/render/registryv1/validators/validator.go
```diff
@@ -264,3 +264,50 @@ func CheckWebhookNameIsDNS1123SubDomain(rv1 *bundle.RegistryV1) []error {
...
+						}
+					}
+				}
+			}
+		}
+	}
+	return slices.SortedFunc(slices.Values(errs), func(err1 error, err2 error) int {
+		return cmp.Compare(err1.Error(), err2.Error())
+	})
```
pedjak: what is the befit or sorting errors before returning them to the caller?

perdasilva: yeah - that's fair. The important thing really is to ensure the errors don't become a source of nondeterminism since that could cause the reconciler to be constantly retriggered. In this case, we don't have any maps, so the order should remain constant. I'll remove!

pedjak: > yeah - that's fair. The important thing really is to ensure the errors don't become a source of nondeterminism since that could cause the reconciler to be constantly retriggered. In this case, we don't have any maps, so the order should remain constant. I'll remove!

:+1: I get it, but the caller (reconciler) should worry about that, not  function.

internal/operator-controller/rukpak/render/registryv1/validators/validator.go
perdasilva: Yup, great call out! I've updated!

internal/operator-controller/rukpak/render/registryv1/validators/validator.go
perdasilva: I've reworded the error - wdyt?

internal/operator-controller/rukpak/render/registryv1/validators/validator.go
perdasilva: In principle that's true. Also, just because we make it deterministic here, does not mean the reconciler shouldn't guard itself from sources of nondeterminism. But I still think its a good idea to make functional output deterministic (if at all possible).

internal/operator-controller/rukpak/render/registryv1/validators/validator.go
```diff
@@ -264,3 +264,45 @@ func CheckWebhookNameIsDNS1123SubDomain(rv1 *bundle.RegistryV1) []error {
...
+// for the admissionregistration.k8s.io api group
+var unsupportedAdmissionRegistrationResources = sets.New(
+	"*",
+	"mutatingwebhookconfiguration",
+	"mutatingwebhookconfigurations",
+	"validatingwebhookconfiguration",
+	"validatingwebhookconfigurations",
+)
+
+// CheckWebhookRules ensures webhook rules do not reference unsupported API groups or resources in line with OLMv0 behavior
```
pedjak: given that you updated the error message, perhaps we should also update the doc as well:

```suggestion
CheckWebhookRules ensures webhook rules do not reference forbidden API groups or resources in line with OLMv0 behavior
```

perdasilva: Great catch! Thank you!!

perdasilva: Fixed

pedjak: Thanks. We should also fix the docs for `unsupportedWebhookRuleAPIGroups` and `unsupportedAdmissionRegistrationResources` and rename them as well.

perdasilva: hahahaha - ahhhh! I'm sorry I forgot that!

perdasilva: Fixed - I've also updated the function's docs to call out the shape of the forbidden rules

PR #2231, opened 2025-09-23
:seedling: OPRUN-4138 `ClusterExtensionRevision` `.spec.revision` must be positive

pedjak:
# Description

Added the validation rules and the unit tests.

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/api/v1/clusterextensionrevision_types.go
+++ b/api/v1/clusterextensionrevision_types.go
@@ -49,9 +49,12 @@ type ClusterExtensionRevisionSpec struct {
 	// +kubebuilder:validation:Enum=Active;Paused;Archived
 	// +kubebuilder:validation:XValidation:rule="oldSelf == 'Active' || oldSelf == 'Paused' || oldSelf == 'Archived' && oldSelf == self", message="can not un-archive"
 	LifecycleState ClusterExtensionRevisionLifecycleState `json:"lifecycleState,omitempty"`
-	// Revision number orders changes over time, must always be previous revision +1.
+	// Revision is a sequence number representing a specific revision of the ClusterExtension instance.
+	// Must be positive. Each ClusterExtensionRevision of the same parent ClusterExtension needs to have
+	// a unique value assigned. It is immutable after creation. The new revision number must always be previous revision +1.
 	//
 	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Minimum:=1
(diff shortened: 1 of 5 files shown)


api/v1/clusterextensionrevision_types.go
```diff
@@ -49,9 +49,12 @@ type ClusterExtensionRevisionSpec struct {
 	// +kubebuilder:validation:Enum=Active;Paused;Archived
 	// +kubebuilder:validation:XValidation:rule="oldSelf == 'Active' || oldSelf == 'Paused' || oldSelf == 'Archived' && oldSelf == self", message="can not un-archive"
 	LifecycleState ClusterExtensionRevisionLifecycleState `json:"lifecycleState,omitempty"`
-	// Revision number orders changes over time, must always be previous revision +1.
+	// Revision is a sequence number representing a specific revision of the ClusterExtension instance.
+	// Must be positive. Each ClusterExtensionRevision of the same parent ClusterExtension needs to have
+	// a unique value assigned. It is immutable after creation. The new revision number must always be previous revision +1.
```
joelanford, line 54: We can validate uniqueness if we require the `metadata.name` to have a pattern that matches `<ceName>-<revNumber>`, right?

But I assume there's no way to guarantee "new rev must be previous rev +1"?

thetechnick: In theory you could add a webhook, but I'd like to avoid webhooks as much as possible.
I don't think you can add a validation rule to .metadata.name, but we could add a CEL transition rule where `.spec.revision` must be equal to `-digit` suffix of the name. :thinking: 

But I'd add that in a new PR.

pedjak: we could probably do that, but for that we need a validation webhook where we could fetch all existing revisions, sort them by creation date and compare their revisions with the one we got in the payload. Not sure if the effort would be justified.

joelanford: It is possible to have a CEL validation rule for `metadata.name`. You just have to declare the validation against the root struct that embeds ObjectMeta.

We _could_ use `ValidationAdmissionPolicy`, which is now preferred over webhooks in general, but there's still the problem that when there are multiple apiserver replicas, they are each acting independently when making validation decisions and (I'm pretty sure) without holding a lock or transaction against etcd for all the data used to make the validation decision.

perdasilva (approved):

PR #2233, opened 2025-09-23
✨ Fix Tilt by adding certManager option to the tilt.yaml chart

trgeiger:
# Description

When the move was made to use helm charts for the Tilt configuration, the TLS components were not added. This means in the current state, both operator-controller and catalogd will exit on launch due to metrics-bind-address requiring the tls-key and tls-cert. This PR adds the certManager option to fix this issue.

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/helm/tilt.yaml
+++ b/helm/tilt.yaml
@@ -2,10 +2,12 @@
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
 
-# Tilt is an exeption to the multi-values case,
-# as the Tilt runner only accepts a single values fle
+# Tilt is an exception to the multi-values case,
+# as the Tilt runner only accepts a single values file
 
 options:
+  certManager:
+    enabled: true
   tilt:
     enabled: true
   featureSet: experimental


dtfranz:
> Looks good when it comes to just Tilt, but I'm confused because cert-manager is set to `enabled:false` [here](https://github.com/operator-framework/operator-controller/blob/main/helm/olmv1/values.yaml#L15-L16)
> 
> Is our deploy target broken currently in that case? How are the e2es passing?

This file you linked is the most basic set of values we use to deploy v1, on top of which we would set values to change deployment environment, for instance upstream and downstream, the latter of which has certmanager disabled. When we install v1 through upstream make targets, certmanager is enabled because all of the [deployment variations](https://github.com/operator-framework/operator-controller/blob/47f8d312187f9f6a41118ee1c5c2f11c09afb7ce/Makefile#L163) use the `helm/cert-manager.yaml` file, which sets that flag to true.

tmshort:
> This file you linked is the most basic set of values we use to deploy v1, on top of which we would set values to change deployment environment, for instance upstream and downstream, the latter of which has certmanager disabled. When we install v1 through upstream make targets, certmanager is enabled because all of the [deployment variations](https://github.com/operator-framework/operator-controller/blob/47f8d312187f9f6a41118ee1c5c2f11c09afb7ce/Makefile#L163) use the `helm/cert-manager.yaml` file, which sets that flag to true.

Exactly, helm allows us to combine multiple files together to create different manifests.

tmshort:
Interesting, we _used_ to have a tilt CI, but it seems to have disappeared?
We do... why isn't it running?
https://github.com/operator-framework/operator-controller/blob/main/.github/workflows/tilt.yaml

tmshort:
Hmmm... it didn't run in this PR, but it does run (and pass) in others (#2213)

dtfranz:
Looks like it ran on PR creation, but not on merge?
https://github.com/operator-framework/operator-controller/actions/runs/17959781940

PR #2237, opened 2025-09-26
:seedling: Refactor NewBundleFS test utility to builder pattern

perdasilva:
# Description

While doing some refactoring work to fix a bug, it became clear I'd need better test tooling to create test bundle filesystems. This PR updates the NewBundleFS utility using the builder pattern to help compose a valid registry+v1 bundle FS for testing

## Reviewer Checklist

- [ ] API Go Documentation


internal/operator-controller/applier/boxcutter_test.go
```diff
@@ -24,13 +24,15 @@ import (
...
+	"github.com/operator-framework/api/pkg/operators/v1alpha1"
+
 	ocv1 "github.com/operator-framework/operator-controller/api/v1"
 	"github.com/operator-framework/operator-controller/internal/operator-controller/applier"
 	"github.com/operator-framework/operator-controller/internal/operator-controller/controllers"
 	"github.com/operator-framework/operator-controller/internal/operator-controller/labels"
 	"github.com/operator-framework/operator-controller/internal/operator-controller/rukpak/bundle"
 	"github.com/operator-framework/operator-controller/internal/operator-controller/rukpak/render"
-	testutils "github.com/operator-framework/operator-controller/internal/operator-controller/rukpak/util/testing"
+	. "github.com/operator-framework/operator-controller/internal/operator-controller/rukpak/util/testing"
```
pedjak: why we could not keep `testutils` import alias?

internal/operator-controller/rukpak/util/testing/bundlefs/bundlefs.go
```diff
@@ -8,57 +8,134 @@ import (
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/yaml"
+
```
pedjak, line 11: general remark: given that functions in this file are used only in testing, how about that we move this under `/test/bundle` folder?

internal/operator-controller/applier/boxcutter_test.go
perdasilva: Feels clearer to read, especially when using options. E.g.

`NewBundleFS(With..(), With...())` vs `testutils.NewBundleFS(testutils.With...(), testutils.With...())`.

Just my opinion, if its poor form, I can put it in.

internal/operator-controller/rukpak/util/testing/bundlefs/bundlefs.go
perdasilva: I think originally these functions were mainly used within the rukpak package itself. Now that they are breaking out, we could move them to `internal/shared/util/testutil/bundle` wdyt? Though, I'd prefer to do it in a separate PR to avoid having all the import renaming changes in this one. I'm happy to fast follow with that after this gets merged.

internal/operator-controller/applier/boxcutter_test.go
pedjak: > Just my opinion, if its poor form, I can put it in.

As per [Effective Go](https://go.dev/doc/effective_go#blank_import), these are typically used for side effects. I would rather move `NewBundleFs` and the options under `/test/bundle` package and improve readability in that way.

internal/operator-controller/rukpak/util/testing/bundlefs/bundlefs.go
pedjak: if only used in testing, I would move them to `test/bundle` or something like that under `test` folder.

internal/operator-controller/applier/boxcutter_test.go
pedjak: Also, if you like to get rid of `testutils` prefix when passing option, then we could create a builder instead:

```golang
bundle := bundle.Builder().WithX(..).WithY(..).Build()
```

internal/operator-controller/rukpak/util/testing/bundlefs/bundlefs.go
perdasilva: I think we already have some testutils until shared/util. Though, I don't have a strong opinion here. Though, I'd say then the test folder would include a mishmash of non-unit tests (e2e, upgrade, regression, etc.) and unit test utility code, which doesn't seem right somehow. What's the problem with having it under some util package? In any case that code won't make it to the binary.

internal/operator-controller/applier/boxcutter_test.go
perdasilva: Side effect imports are _ imports (rather than . imports), right? The code is used in the package. I guess we could use the builder pattern...

internal/operator-controller/rukpak/util/testing/bundlefs/bundlefs.go
pedjak: > I think we already have some testutils until shared/util. Though, I don't have a strong opinion here. Though, I'd say then the test folder would include a mishmash of non-unit tests (e2e, upgrade, regression, etc.) and unit test utility code, which doesn't seem right somehow. What's the problem with having it under some util package? In any case that code won't make it to the binary.

I was unaware that `/test` folder is for e2e tests only. I would put as much as possible of the code used by various tests under `/test` folder, but I am not blocking it, if you would like to keep it under `production` folder hierarchy.

internal/operator-controller/applier/boxcutter_test.go
perdasilva: I've updated it to use the builder pattern. We'll need to do the same with the CSV utility. Let's push moving the package around to a separate PR. I'm not sure the /test directory would be the best place to move it since it would commingle with the e2e/upgrade/etc. tests (which don't make any use of it). What's the issue with carrying it the regular package structure?

pedjak (approved):

internal/operator-controller/applier/boxcutter_test.go
perdasilva: I've also moved it under a bundlefs package in preparation

pedjak: > What's the issue with carrying it the regular package structure?

Nothing big really, IMHO it separates better physically test code from the production one. But just my 2c.

PR #2238, opened 2025-09-26
:bug: OCPBUGS-62021: Fix install mode support gating

perdasilva:
# Description

The RegistryV1HelmChartProvider was letting through bundles that don't support AllNamespaces mode.

RegistryV1HelmChartProvider is updated to handle the SingleOwnNamespace feature in the same way the Webhook feature is handled. In a follow-up PR, GetWatchNamespace() (which checks for the SingleOwnNamespace feature) will be moved into this component. 

Note: This PR used to have an e2e tests to ensure bundles not supported by the standard OLMv1 profile get rejected. But, there's currently no way to execute an e2e test on the standard profile only. So, I've removed it and will add in a follow up PR once we've addressed this the right way.

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/operator-controller/applier/provider.go
+++ b/internal/operator-controller/applier/provider.go
@@ -6,6 +6,9 @@ import (
 	"fmt"
 
 	"helm.sh/helm/v3/pkg/chart"
+	"k8s.io/apimachinery/pkg/util/sets"
+
+	"github.com/operator-framework/api/pkg/operators/v1alpha1"
 
 	ocv1 "github.com/operator-framework/operator-controller/api/v1"
 	"github.com/operator-framework/operator-controller/internal/operator-controller/rukpak/bundle/source"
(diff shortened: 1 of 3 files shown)


tmshort:
The experimental-e2e is not passing...
But I'll run them again (one was a timeout condition)

perdasilva:
> The experimental-e2e is not passing... But I'll run them again (one was a timeout condition)

Yeah - these aren't really meant to be run in experimental mode. It's to check that the things that are enabled in experimental mode aren't enabled out of experimental mode

internal/operator-controller/applier/provider.go
```diff
@@ -48,8 +40,33 @@ func (r *RegistryV1HelmChartProvider) Get(bundle source.BundleSource, ext *ocv1.
...
-	if r.CertificateProvider == nil && len(rv1.CSV.Spec.WebhookDefinitions) > 0 {
-		return nil, fmt.Errorf("unsupported bundle: webhookDefinitions are not supported")
+	installModes := sets.New(rv1.CSV.Spec.InstallModes...)
+	if !r.IsSingleOwnNamespaceEnabled && !installModes.Has(v1alpha1.InstallMode{Type: v1alpha1.InstallModeTypeAllNamespaces, Supported: true}) {
+		return nil, fmt.Errorf("unsupported bundle: bundle does not support AllNamespaces install mode")
+	}
+
+	if !installModes.Has(v1alpha1.InstallMode{Type: v1alpha1.InstallModeTypeAllNamespaces, Supported: true}) &&
+		!installModes.Has(v1alpha1.InstallMode{Type: v1alpha1.InstallModeTypeSingleNamespace, Supported: true}) &&
+		!installModes.Has(v1alpha1.InstallMode{Type: v1alpha1.InstallModeTypeOwnNamespace, Supported: true}) {
```
pedjak: we could use `sets.HasAny(...)` to improve readability of this check.

internal/operator-controller/applier/provider.go
```diff
@@ -48,8 +40,33 @@ func (r *RegistryV1HelmChartProvider) Get(bundle source.BundleSource, ext *ocv1.
...
+	//   1. takes a bundle + cluster extension => manifests
+	//   2. takes a bundle + cluster extension => chart (which will use the component in 1. under the hood)
+	// GetWatchNamespace will move under the component in 1. and also be reused by the component that
+	// takes bundle + cluster extension => revision
+	watchNamespace, err := GetWatchNamespace(ext)
+	if err != nil {
+		return nil, err
+	}
+
+	if watchNamespace != "" {
```
pedjak, line 70: what is happening if `watchNamespace == ""`?

perdasilva: If the return value of GetWatchNamespace is "" => no configuration set.

It's also worth noting that targetNamespaces = [""] tells the renderer to render in AllNamespaces mode (which may or may not be supported). AllNamespaces is the default install mode. So, one would never need to set it explicitly.

internal/operator-controller/applier/provider.go
perdasilva: Nice one! I've updated it =D

thetechnick (approved):

PR #2242, opened 2025-09-29
✨ Deprecate spec.ServiceAccount and remove synthetic permissions feature

trgeiger:
Make spec.ServiceAccount an optional field and note that it's now deprecated and does not perform any function. Make OLM use cluster-admin by default for managing ClusterExtensions.

Remove the synthetic permissions experimental feature flag.

# Description

## Reviewer Checklist

- [ ] API Go Documentation


api/v1/clusterextension_types.go
```diff
@@ -67,14 +66,13 @@ type ClusterExtensionSpec struct {
 	// +kubebuilder:validation:Required
 	Namespace string `json:"namespace"`
 
-	// serviceAccount is a reference to a ServiceAccount used to perform all interactions
+	// Deprecated: ServiceAccount is ignored by OLM and will be removed in a future release.
```
rashmigottipati, line 69: Would it make sense to also log a warning during reconciliation if this field is still set by a user?

hack/tools/crd-generator/testdata/api/v1/clusterextension_types.go
```diff
@@ -93,6 +95,15 @@ type ClusterExtensionSpec struct {
...
 	Install *ClusterExtensionInstallConfig `json:"install,omitempty"`
+
+	// config contains optional configuration values applied during rendering of the
+	// ClusterExtension's manifests. Values can be specified inline.
+	//
+	// config is optional. When not specified, the default configuration of the resolved bundle will be used.
+	//
+	// <opcon:experimental>
+	// +optional
+	Config *ClusterExtensionConfig `json:"config,omitempty"`
```
rashmigottipati: Looks like we are adding ClusterExtensionConfig to the spec. While it looks like a good start for supporting inline config, it seems unrelated to the serviceAccount deprecation work (unless I am missing something here).
Just checking - Is it intentional to include this here, or would it make sense to split it into a separate PR?

grokspawn: I was just coming to note that this looks like it needs to be rebased, since it's including changes from 
> d95f426f :sparkles: ClusterExtensionConfig API (#2163)

trgeiger: I have re-based and just re-based again so I'm not sure why this is here... must be user error by me during some rebase reconciliation the other day. I'll remove this

internal/operator-controller/action/restconfig_test.go
```diff
@@ -1,177 +1,58 @@
...
-		return &http.Response{}, nil
-	})).RoundTrip(&http.Request{})
-}
-
-type fakeRoundTripper func(req *http.Request) (*http.Response, error)
-
-func (f fakeRoundTripper) RoundTrip(request *http.Request) (*http.Response, error) {
-	return f(request)
-}
+// func Test_ServiceAccountRestConfigMapper(t *testing.T) {
```
grokspawn, line 3: still working to get this functioning?

internal/operator-controller/features/features.go
```diff
@@ -11,24 +11,14 @@ import (
...
 	SingleOwnNamespaceInstallSupport  featuregate.Feature = "SingleOwnNamespaceInstallSupport"
-	SyntheticPermissions              featuregate.Feature = "SyntheticPermissions"
 	WebhookProviderCertManager        featuregate.Feature = "WebhookProviderCertManager"
 	WebhookProviderOpenshiftServiceCA featuregate.Feature = "WebhookProviderOpenshiftServiceCA"
 	HelmChartSupport                  featuregate.Feature = "HelmChartSupport"
 	BoxcutterRuntime                  featuregate.Feature = "BoxcutterRuntime"
 )
 
 var operatorControllerFeatureGates = map[featuregate.Feature]featuregate.FeatureSpec{
-	// Add new feature gate definitions
```
grokspawn, line 24: I think these changes are going to require manual merge elsewhere, where the feature flags are used.  Just be aware of it.  Since those are alpha featuregates, I don't think it's a concern.

api/v1/clusterextension_types.go
grokspawn: I have some concerns about generating a log on each reconcile.  It looks like k8s uses a metric to express "thing is using a deprecated API" and that feels like an approach which could scale better.

api/v1/clusterextension_types.go
```diff
@@ -67,14 +66,13 @@ type ClusterExtensionSpec struct {
...
 	// with the cluster that are required to manage the extension.
-	// The ServiceAccount must be configured with the necessary permissions to perform these interactions.
-	// The ServiceAccount must exist in the namespace referenced in the spec.
-	// serviceAccount is required.
+	// serviceAccount is optional.
 	//
-	// +kubebuilder:validation:Required
-	ServiceAccount ServiceAccountReference `json:"serviceAccount"`
+	// +kubebuilder:validation:Optional
+	ServiceAccount ServiceAccountReference `json:"serviceAccount,omitzero"`
```
grokspawn, line 75: I'm not sure what the kubebuilder magic is which would generate the 'deprecated' label for this field, but it would be great to do it, like https://kubernetes.io/blog/2020/09/03/warnings/#custom-resource-definitions.

grokspawn: Rashmi figured it out: https://github.com/openshift/enhancements/pull/1860/files#diff-efdbbf4db1d20789416939675230e4c8f75fcebc51335567d47908a9b1c42c08R72

PR #2244, opened 2025-10-02
:seedling: Refactor MakeCSV utility to builder pattern

perdasilva:
# Description

Since we moved the bundlefs test utility to the builder pattern, this PR moves the MakeCSV test utility to the builder pattern for consistency.

## Reviewer Checklist

- [ ] API Go Documentation


internal/operator-controller/applier/provider_test.go
```diff
@@ -51,7 +52,7 @@ func Test_RegistryV1HelmChartProvider_Get_ReturnsBundleRendererFailures(t *testi
 
 	b := source.FromBundle(
 		bundle.RegistryV1{
-			CSV: MakeCSV(WithInstallModeSupportFor(v1alpha1.InstallModeTypeAllNamespaces)),
+			CSV: csvbuilder.New().WithInstallModeSupportFor(v1alpha1.InstallModeTypeAllNamespaces).Build(),
```
pedjak: to make it consistent with `bundlefs.Builder()` calls, how about that we also do `csv.Builder()` instance of `csvbuilder.New()`?

perdasilva: that was my original approach but csv is such a common variable name I didn't want to overload it with the package. Wdyt of csvbuilder.Builder()? That was my first alternative, but then it felt weird to have csvbuilder.Builder() - but maybe that's ok?

pedjak: IMHO, previously `MakeCSV` sounds also very generic/common, so I would go simply with `csv.Builder()` - adding redundant `builder` suffix does not bring much for clarity. Maybe to call the package `clusterserviceversion` and then we have `clusterserviceversion.Builder()`?

perdasilva: feels so long though T_T

perdasilva: I get that CSV is a particularity of this project. But, within this context, its widely used.

pedjak: > I get that CSV is a particularity of this project. But, within this context, its widely used.

If so, then `csv.Builder()` is pretty clear then.

perdasilva: I've just gone for the original suggestion. I don't like csv because it's a common variable name for the csv.
csv := csv.Builder().Build() feels weird to me.

perdasilva: We're just talking about some small test utility anyway.

internal/operator-controller/rukpak/render/registryv1/registryv1_test.go
```diff
@@ -64,12 +65,11 @@ func Test_ResourceGeneratorsHasAllGenerators(t *testing.T) {
 }
 
 func Test_Renderer_Success(t *testing.T) {
-	bundle := bundle.RegistryV1{
+	someBundle := bundle.RegistryV1{
```
pedjak, line 68: was renaming needed, given that there is single bundle in the test?

internal/operator-controller/rukpak/render/registryv1/registryv1_test.go
```diff
@@ -98,12 +98,11 @@ func Test_Renderer_Success(t *testing.T) {
 }
 
 func Test_Renderer_Failure_UnsupportedKind(t *testing.T) {
-	bundle := bundle.RegistryV1{
```
pedjak, line 101: same as above?

internal/operator-controller/rukpak/render/registryv1/registryv1_test.go
perdasilva: Strictly, probably not. But, again, the variable name shadows the package name. I feel it's easier to read this way.
Is this blocking?

pedjak: > Strictly, probably not. But, again, the variable name shadows the package name. I feel it's easier to read this way. Is this blocking?

Ah, good point! No, no blocker, I have already approved with this comments as well.

PR #2245, opened 2025-10-02
wip remove phase hashing at CER applier

pedjak:
# Description

Drop computing hash in CER boxcutter applier

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/cmd/operator-controller/main.go
+++ b/cmd/operator-controller/main.go
@@ -106,7 +106,10 @@ type config struct {
 	globalPullSecret     string
 }
 
-const authFilePrefix = "operator-controller-global-pull-secrets"
+const (
+	authFilePrefix   = "operator-controller-global-pull-secrets"
+	fieldOwnerPrefix = "olm.operatorframework.io"
+)
 
 // podNamespace checks whether the controller is running in a Pod vs.
 // being run locally by inspecting the namespace file that gets mounted
@@ -542,6 +545,7 @@ func setupBoxcutter(mgr manager.Manager, ceReconciler *controllers.ClusterExtens
 		Scheme:            mgr.GetScheme(),
 		RevisionGenerator: rg,
 		Preflights:        preflights,
+		FieldOwner:        fmt.Sprintf("%s/clusterextension-controller", fieldOwnerPrefix),
 	}
 	ceReconciler.RevisionStatesGetter = &controllers.BoxcutterRevisionStatesGetter{Reader: mgr.GetClient()}
 	ceReconciler.StorageMigrator = &applier.BoxcutterStorageMigrator{
@@ -549,11 +553,6 @@ func setupBoxcutter(mgr manager.Manager, ceReconciler *controllers.ClusterExtens
 		ActionClientGetter: acg,
 		RevisionGenerator:  rg,
 	}
-
-	// Boxcutter
-	const (
-		boxcutterSystemPrefixFieldOwner = "olm.operatorframework.io"
-	)
 
 	discoveryClient, err := discovery.NewDiscoveryClientForConfig(mgr.GetConfig())
 	if err != nil {
@@ -581,8 +580,8 @@ func setupBoxcutter(mgr manager.Manager, ceReconciler *controllers.ClusterExtens
 				machinery.NewObjectEngine(
 					mgr.GetScheme(), trackingCache, mgr.GetClient(),
 					ownerhandling.NewNative(mgr.GetScheme()),
-					machinery.NewComparator(ownerhandling.NewNative(mgr.GetScheme()), discoveryClient, mgr.GetScheme(), boxcutterSystemPrefixFieldOwner),
-					boxcutterSystemPrefixFieldOwner, boxcutterSystemPrefixFieldOwner,
+					machinery.NewComparator(ownerhandling.NewNative(mgr.GetScheme()), discoveryClient, mgr.GetScheme(), fieldOwnerPrefix),
+					fieldOwnerPrefix, fieldOwnerPrefix,
 				),
 				validation.NewClusterPhaseValidator(mgr.GetRESTMapper(), mgr.GetClient()),
 			),
--- a/internal/operator-controller/applier/boxcutter.go
+++ b/internal/operator-controller/applier/boxcutter.go
@@ -29,7 +29,6 @@ import (
 	"github.com/operator-framework/operator-controller/internal/operator-controller/labels"
 	"github.com/operator-framework/operator-controller/internal/operator-controller/rukpak/bundle/source"
 	"github.com/operator-framework/operator-controller/internal/operator-controller/rukpak/render"
-	hashutil "github.com/operator-framework/operator-controller/internal/shared/util/hash"
 )
 
 const (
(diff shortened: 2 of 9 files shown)


internal/operator-controller/applier/boxcutter.go
```diff
@@ -218,84 +218,107 @@ func (bc *Boxcutter) getObjects(rev *ocv1.ClusterExtensionRevision) []client.Obj
...
+	currentRevision := &ocv1.ClusterExtensionRevision{}
 	state := StateNeedsInstall
-	if len(existingRevisions) > 0 {
-		maybeCurrentRevision := existingRevisions[len(existingRevisions)-1]
-		annotations := maybeCurrentRevision.GetAnnotations()
-		if annotations != nil {
-			if revisionHash, ok := annotations[RevisionHashAnnotation]; ok && revisionHash == desiredHash {
-				currentRevision = &maybeCurrentRevision
+	// check if we can update the current revision.
+	if len(existingRevisions) > 0 { // nolint:nestif
```
camilamacedo86: Could we not fix the lint issue instead of ignore?

internal/operator-controller/applier/boxcutter.go
```diff
@@ -218,84 +218,107 @@ func (bc *Boxcutter) getObjects(rev *ocv1.ClusterExtensionRevision) []client.Obj
...
-	for _, preflight := range bc.Preflights {
-		if shouldSkipPreflight(ctx, preflight, ext, state) {
-			continue
-		}
-		switch state {
-		case StateNeedsInstall:
-			err := preflight.Install(ctx, plainObjs)
-			if err != nil {
-				return false, "", err
+	if state != StateUnchanged { // nolint:nestif
```
camilamacedo86: Same.
Also, is not better keep a switch here?

cmd/operator-controller/main.go
```diff
@@ -107,7 +107,10 @@ type config struct {
 	globalPullSecret     string
 }
 
-const authFilePrefix = "operator-controller-global-pull-secrets"
```
camilamacedo86, line 110: Does this change affect end users, OR is that a more internal thing?
If it is an internal thing than the right emoji is 🌱 

The ✨ should only be used for new features (in the POV of end users)

test/experimental-e2e/experimental_e2e_test.go
```diff
@@ -389,6 +390,60 @@ func TestClusterExtensionConfigSupport(t *testing.T) {
...
+		require.NotNil(ct, cond)
+		require.Equal(ct, metav1.ConditionTrue, cond.Status)
+		require.Equal(ct, ocv1.ReasonSucceeded, cond.Reason)
+	}, pollDuration, pollInterval)
+
+	t.Log("It allows to upgrade the ClusterExtension to a non-successor version")
+	t.Log("By updating the ClusterExtension resource to a non-successor version")
+	// 1.2.0 does not replace/skip/skipRange 1.0.0.
+	clusterExtension.Spec.Source.Catalog.Version = "1.2.0"
+	clusterExtension.Spec.Source.Catalog.UpgradeConstraintPolicy = ocv1.UpgradeConstraintPolicySelfCertified
```
perdasilva, line 429:
```suggestion
    t.Log("By setting the upgrade constraint policy to self-certified")
	clusterExtension.Spec.Source.Catalog.UpgradeConstraintPolicy = ocv1.UpgradeConstraintPolicySelfCertified
```

test/experimental-e2e/experimental_e2e_test.go
```diff
@@ -389,6 +390,60 @@ func TestClusterExtensionConfigSupport(t *testing.T) {
 	}, pollDuration, pollInterval)
 }
 
+func TestClusterExtensionForceInstallNonSuccessorVersion(t *testing.T) {
```
perdasilva: Would it make sense to have a negative test as well by either having a separate test where we don't update the upgrade policy, or having this test first try with the standard policy, watch it fail, then update to self-certified and watch it succeed?

testdata/images/bundles/test-operator/v1.2.0/manifests/bundle.configmap.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: test-configmap
+  annotations:
+    shouldNotTemplate: >
+      The namespace is {{ $labels.namespace }}. The templated
+      $labels.namespace is NOT expected to be processed by OLM's
+      rendering engine for registry+v1 bundles.
```
perdasilva, line 9: you could probably remove this

testdata/images/bundles/test-operator/v1.2.0/manifests/testoperator.clusterserviceversion.yaml
```diff
@@ -0,0 +1,151 @@
...
+  - name: simple-bundle
+    url: https://simple-bundle.domain
+  maintainers:
+  - email: main#simple-bundle.domain
+    name: Simple Bundle
+  maturity: beta
+  provider:
+    name: Simple Bundle
+    url: https://simple-bundle.domain
+  version: 1.1.0
```
perdasilva:
```suggestion
  version: 1.2.0
```

internal/operator-controller/applier/boxcutter_test.go
```diff
@@ -528,6 +540,7 @@ func TestBoxcutter_Apply(t *testing.T) {
 					}, nil
 				},
 			},
+			clientIterceptor: allowedRevisionValue(2),
```
perdasilva, line 491: The name for this test (on line 510) is "new revision created when hash differs", maybe just "new revision created" now?

cmd/operator-controller/main.go
pedjak: you are right, it does not change behavior exposed to users, gonna move to :seedling:

test/experimental-e2e/experimental_e2e_test.go
pedjak: I would say it should probably be then `By forcing updating the ClusterExtension resource to a non-successor version` - setting to self-certified policy is just the way to force such updates. wdyt?

test/experimental-e2e/experimental_e2e_test.go
pedjak: The test is not about changing the upgrade policy - is is about asserting that a new revision is created if objects in the new version are different from the previous installed version. Such test did not exist previously. 

The additional tests are not bound to the experimental e2e suite and could be added in a separate PR.

internal/operator-controller/applier/boxcutter_test.go
pedjak: changed to `new revision created when objects in new revision are different`

test/experimental-e2e/experimental_e2e_test.go
perdasilva: Even better ^^

test/experimental-e2e/experimental_e2e_test.go
perdasilva: From the title and behavior it seems to be about ensuring that the user can force install to a version without an upgrade edge from the current one (which of course implies the creation of a new revision). Maybe a new title?

PR #2246, opened 2025-10-02
:sparkles: Add support for TLS profiles

tmshort:
Use Mozilla's profiles to define TLS profiles for operator-controller and catalogd. These are configured via command-line options, and can be customized.

The idea is that downstream, cluster-olm-operator will be able to glean the appropriate configuration, and provide that to the components.

There is a semi-automatic method to update the profiles, if that ever happens (`make update-tls-profiles`).

This adds `gojq` via bingo, which is a golang implementation of jq for the update-tls-profiles target.

# Description

## Reviewer Checklist

- [ ] API Go Documentation


helm/olmv1/templates/deployment-olmv1-system-catalogd-controller-manager.yml
```diff
@@ -53,6 +53,11 @@ spec:
             - --v=${LOG_VERBOSITY}
             - --global-pull-secret=openshift-config/pull-secret
             {{- end }}
+            {{- if not .Values.options.openshift.enabled }}
```
grokspawn: Don't love a distro-specific flag here.  Can we do this in a generic way?

grokspawn:
Is this PR solving an issue or adding a new feature?

helm/olmv1/templates/deployment-olmv1-system-operator-controller-controller-manager.yml
```diff
@@ -54,6 +54,9 @@ spec:
             - --v=${LOG_VERBOSITY}
             - --global-pull-secret=openshift-config/pull-secret
             {{- end }}
+            {{- if not .Values.options.openshift.enabled }}
+            - --tls-profile=modern
+            {{- end }}
```
joelanford: For OCP, I think the expectation is that the TLS settings will always be dynamically configured based on the contents of `apiservers.config.openshift.io.Spec.TLSSecurityProfile` and the logic in `https://github.com/openshift/library-go/blob/f838eb5c601908101db3c84f925c0093b5304611/pkg/operator/configobserver/apiserver/observe_tlssecurityprofile.go#L85-L109`

RH probably needs to export that function (or at least use/implement some other helper that calls it) so that they can use OCP's opinions of what the default should be and what the different profiles actually are when they run OLM in OpenShift.

Makefile
```diff
@@ -189,6 +189,10 @@ fix-lint: $(GOLANGCI_LINT) #EXHELP Fix lint issues
 fmt: #EXHELP Formats code
 	go fmt ./...
 
+.PHONY: update-tls-profiles
+update-tls-profiles: $(GOJQ) #EXHELP Update TLS profiles from the Mozilla wiki
+	env JQ=$(GOJQ) hack/tools/update-tls-profiles.sh
```
joelanford, line 194: Should we include this in `verify` to make sure that no one edits the generated go file outside the expected mechanism?

internal/shared/util/tlsprofiles/flags.go
```diff
@@ -0,0 +1,140 @@
...
+	"maps"
+	"slices"
+	"strings"
+
+	"github.com/spf13/pflag"
+)
+
+func AddFlags(fs *pflag.FlagSet) {
+	fs.Var(&configuredProfile, "tls-profile", "The TLS profile to use. One of "+fmt.Sprintf("%v", slices.Sorted(maps.Keys(profiles))))
+	fs.Var(&customTLSProfile.ciphers, "tls-custom-cipher", "List of ciphers to be used with the custom TLS profile")
```
joelanford:
```suggestion
	fs.Var(&customTLSProfile.ciphers, "tls-custom-ciphers", "List of ciphers to be used with the custom TLS profile")
```

internal/shared/util/tlsprofiles/flags.go
```diff
@@ -0,0 +1,140 @@
...
+	return string(*p)
+}
+
+func (p *tlsProfileName) Type() string {
+	return "string"
+}
+
+func (p *tlsProfileName) Set(value string) error {
+	newValue := tlsProfileName(value)
+	_, err := findTlsProfile(newValue)
```
joelanford:
```suggestion
	_, err := findTLSProfile(newValue)
```

helm/olmv1/templates/deployment-olmv1-system-catalogd-controller-manager.yml
```diff
@@ -53,6 +53,11 @@ spec:
...
+            - --tls-profile=custom
+            - --tls-custom-version=TLSv1.3
+            - --tls-custom-cipher=TLS_AES_128_GCM_SHA256,TLS_AES_256_GCM_SHA384
```
joelanford: To match operator controller's upstream default?
```suggestion
            - --tls-profile=modern
```

helm/olmv1/templates/deployment-olmv1-system-operator-controller-controller-manager.yml
joelanford: I overlooked the `not` part of this, and realized that this templating is setting a default for upstream, so all good! But hopefully the info in my previous comment is helpful for the OCP integration bits of this that would come later.

helm/olmv1/templates/deployment-olmv1-system-catalogd-controller-manager.yml
tmshort: The custom is basically modern missing one cipher. I really meant for it to be e2e.

helm/olmv1/templates/deployment-olmv1-system-catalogd-controller-manager.yml
tmshort: This was really meant to be an e2e flag.

helm/olmv1/templates/deployment-olmv1-system-operator-controller-controller-manager.yml
tmshort: At this point, cluster-olm-operator will watch that API on behalf of op-con and catd, and update the manifests.

tmshort:
> Is this PR solving an issue or adding a new feature?

@grokspawn there's a downstream bug to support TLS profiles, this is adding the functionality upstream without having to be "too" openshift aware.

Makefile
tmshort: Perhaps, but it'd also detect changes to the Mozilla profiles.

grokspawn:
> > Is this PR solving an issue or adding a new feature?
> 
> @grokspawn there's a downstream bug to support TLS profiles, this is adding the functionality upstream without having to be "too" openshift aware.

Given that this is considered a GA component, I would expect to see something like:
1. default behavior which aligned with current behavior (unless there is a defect in that behavior);
2. input options which would allow distros to specify non-default behavior in a non-distro-specific way

tmshort:
> > > Is this PR solving an issue or adding a new feature?
> > 
> > 
> > @grokspawn there's a downstream bug to support TLS profiles, this is adding the functionality upstream without having to be "too" openshift aware.
> 
> Given that this is considered a GA component, I would expect to see something like:
> 
> 1. default behavior which aligned with current behavior (unless there is a defect in that behavior);
> 2. input options which would allow distros to specify non-default behavior in a non-distro-specific way

1. The default behavior is "intermediate", which is a reasonably secure option. Using "old" would be too open, and "modern" might be too limiting. Given that we never specified any of these parameters, we were kinda at the whim of the golang developer's defaults (which could change with every release). "Intermediate" is a reasonable default given that we want to be reasonably secure.

2. The input options use command-line options to the deployment, so are completely non-distro-specific.

grokspawn:
>     2. The input options use command-line options to the deployment, so are completely non-distro-specific.

https://github.com/operator-framework/operator-controller/pull/2246/files#diff-38ad0fe7f1443195aef2d54ca6608694fc2306bcbc738a294655dc8ed4408351R56 ?

tmshort:
> > ```
> > 2. The input options use command-line options to the deployment, so are completely non-distro-specific.
> > ```
> 
> https://github.com/operator-framework/operator-controller/pull/2246/files#diff-38ad0fe7f1443195aef2d54ca6608694fc2306bcbc738a294655dc8ed4408351R56 ?

That is to include the options for the upstream-e2e only, and not on the downstream-e2e; it's for test purposes during the upstream-e2e, and to avoid erroneously adding it to OpenShift. Nothing precludes adding those values during `helm template` via `--set` or `--values`.

helm/olmv1/templates/deployment-olmv1-system-operator-controller-controller-manager.yml
joelanford: Yeah, I'm mainly just saying now that downstream will likely want to follow whatever OCP's opinions are on:
- what the default profile is
- what the definition of each profile is.

grokspawn:
> That is to include the options for the upstream-e2e only, and not on the downstream-e2e; it's for test purposes during the upstream-e2e, and to avoid erroneously adding it to OpenShift. Nothing precludes adding those values during helm template via --set or --values.

IMHO, the upstream should not have to care about any distro.  Responsibility for using it properly in a distro lies with that distro.  
Ideally, this is because the mechanism is general and can be customized as desired by users.

tmshort:
> IMHO, the upstream should not have to care about any distro. Responsibility for using it properly in a distro lies with that distro. 

We already have Openshift as an option in the helm charts for downstream even before this PR. I can remove the check from this PR, but it won't remove the fact that there's an Openshift option in the chart.

> Ideally, this is because the mechanism is general and can be customized as desired by users.

Again, this is possible via `--set` and/or `--values` provided by the user/distro when they generate the manifests.

tmshort:
Removed the `openshift` check.

grokspawn (approved):

PR #2252, opened 2025-10-07
Direct bundle install

perdasilva:
# Description

Adds annotation driven internal bundle install for internal purposes (at least for now) to aid in the testing / validation of extension and revision status development

## Reviewer Checklist

- [ ] API Go Documentation


camilamacedo86 (approved):
👍 For now it seems reasonable 

/lgtm

cmd/operator-controller/main.go
```diff
@@ -410,6 +419,28 @@ func run() error {
...
+			ext.Annotations[directBundleInstallVersionAnnotation] != "" {
+			bundleImageRef := ext.Annotations[directBundleInstallImageAnnotation]
+
+			ref, err := reference.ParseNamed(bundleImageRef)
+			if err != nil {
+				return nil, nil, nil, err
+			}
+
+			packageName := path.Base(ref.Name())
+			bundleVersion := bsemver.MustParse(ext.Annotations[directBundleInstallVersionAnnotation])
```
joelanford: We shouldn't require or trust user input for the bundle version. Maybe fine for internal use, but in a user-facing feature, I think we'll have to dereference the version from the bundle image.

For now, perhaps leave a comment here to that effect?

cmd/operator-controller/main.go
```diff
@@ -410,6 +419,28 @@ func run() error {
 		},
 	}
 
+	var resolver resolve.Func = func(ctx context.Context, ext *ocv1.ClusterExtension, installedBundle *ocv1.BundleMetadata) (*declcfg.Bundle, *bsemver.Version, *declcfg.Deprecation, error) {
```
joelanford: It feels a bit messy for this to live in `main.go`. Maybe we should go ahead and make a composite resolver  in the resolve package that can switch between the implementations?

camilamacedo86: I understand this isn’t the final solution — it’s more like one of the many intermediate steps to achieve the goal. So I’d be fine with handling it in a follow-up.
That said, it’s a really good point. 👍 +1

cmd/operator-controller/main.go
perdasilva: The problem is that the version is embedded all the way down in the CSV. I.e. we'd need to pull the image and unpack it to figure out what it is. There's no guarantee that, e.g., the tag will be a valid semver. But, I agree. The ideal user-facing version of this would be to update the API, add a new source type, and just take the bundle image as input. I'll add the comment though.

cmd/operator-controller/main.go
perdasilva: Yeah - the idea was to be quick and dirty. But fair call. Tomorrow I'll massage it in to something more tenable.

cmd/operator-controller/main.go
perdasilva: Maybe the solution to this is something like: the resolver always returns an image reference, sometimes returns package and version information. After unpack, if we got nothing from the resolver, we just use what the bundle says, if we got something from the resolver we compare notes and error if there's a mismatch.

internal/operator-controller/resolve/bundle.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+	bundleFS, canonicalRef, _, err := r.ImagePuller.Pull(ctx, ext.Name, ext.Annotations[directBundleInstallImageAnnotation], r.ImageCache)
+	if err != nil {
+		return nil, nil, nil, err
+	}
+
+	// TODO: This is a temporary workaround to get the bundle from the filesystem
+	//    until the operator-registry library is updated to support reading from a
+	//    fs.FS. This will be removed once the library is updated.
+	bundlePath, err := getDirFSPath(bundleFS)
```
tmshort, line 40: I really don't like the FS interface, as it's so limiting, and you're already needing to reflect it to get the proper path. I'd almost prefer to get rid of the FS interface, and simply use a path string.

tmshort: (This is really just me whining about the FS interface is all... not asking for a change unless you agree and really want to.)

perdasilva: I don't disagree. But, atm, this FG is just an artifice to help us play around with the boxcutter status without having to deal with catalogs. Ultimately, I'd probably suggest moving away from FS and towards a formalized bundle interface that can surface all this kind of information without too many headaches

internal/operator-controller/resolve/bundle.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/operator-framework/operator-registry/alpha/action"
+	"github.com/operator-framework/operator-registry/alpha/declcfg"
+
+	ocv1 "github.com/operator-framework/operator-controller/api/v1"
+	"github.com/operator-framework/operator-controller/internal/operator-controller/bundleutil"
+	"github.com/operator-framework/operator-controller/internal/shared/util/image"
+)
+
+const (
+	directBundleInstallImageAnnotation = "olm.operatorframework.io/bundle-image"
```
pedjak: do we want to reflect in the annotation name that it is alpha feature, so something like:

```
alpha.olm.operatorframework.io/bundle-image-ref
```

internal/operator-controller/resolve/bundle.go
```diff
@@ -0,0 +1,85 @@
...
+	if err != nil {
+		return nil, nil, nil, err
+	}
+
+	// TODO: This is a temporary workaround to get the bundle from the filesystem
+	//    until the operator-registry library is updated to support reading from a
+	//    fs.FS. This will be removed once the library is updated.
+	bundlePath, err := getDirFSPath(bundleFS)
+	if err != nil {
+		panic(fmt.Errorf("expected to be able to recover bundle path from bundleFS: %v", err))
```
pedjak: why panic instead of

```
return nil, nil, nil, fmt.Errorf("expected to be able to recover bundle path from bundleFS: %v", err)
```

internal/operator-controller/resolve/bundle.go
```diff
@@ -0,0 +1,85 @@
...
+	render := action.Render{
+		Refs:           []string{bundlePath},
+		AllowedRefMask: action.RefBundleDir,
+	}
+	fbc, err := render.Run(ctx)
+	if err != nil {
+		return nil, nil, nil, err
+	}
+	if len(fbc.Bundles) != 1 {
+		return nil, nil, nil, errors.New("expected exactly one bundle")
```
pedjak: perhaps you can provide in the error message the number of bundles we got?

internal/operator-controller/resolve/bundle.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+	bundle := fbc.Bundles[0]
+	bundle.Image = canonicalRef.String()
+	v, err := bundleutil.GetVersion(bundle)
+	if err != nil {
+		return nil, nil, nil, err
```
pedjak: can you provide the error context here?

internal/operator-controller/resolve/bundle.go
```diff
@@ -0,0 +1,85 @@
...
+
+type BundleResolver struct {
+	ImagePuller image.Puller
+	ImageCache  image.Cache
+}
+
+func (r *BundleResolver) Resolve(ctx context.Context, ext *ocv1.ClusterExtension, _ *ocv1.BundleMetadata) (*declcfg.Bundle, *bsemver.Version, *declcfg.Deprecation, error) {
+	if ext.Annotations == nil || ext.Annotations[directBundleInstallImageAnnotation] == "" {
+		return nil, nil, nil, fmt.Errorf("ClusterExtension is missing required annotation %s", directBundleInstallImageAnnotation)
+	}
```
pedjak, line 31: This check is performed even in MultiResolver, before invoking BundleResolver. Hence, we can drop it here.

internal/operator-controller/resolve/bundle.go
```diff
@@ -0,0 +1,85 @@
...
+	ocv1 "github.com/operator-framework/operator-controller/api/v1"
+	"github.com/operator-framework/operator-controller/internal/operator-controller/bundleutil"
+	"github.com/operator-framework/operator-controller/internal/shared/util/image"
+)
+
+const (
+	directBundleInstallImageAnnotation = "olm.operatorframework.io/bundle-image"
+)
+
+type BundleResolver struct {
```
pedjak: How about to name it `BundleImageRefResolver`? IMHO, `BundleResolver` is too generic.

perdasilva:
> it would great to add at least some unit tests and great if we can have an e2e test demonstrating the usage.

100% - the idea here though was to have something quick and dirty to aid us in playing around with the status stuff without having to deal with catalogs. There will definitely need to be tests of all kinds before this ever becomes user facing. Right now it shouldn't be on the critical path of other PRs. If it breaks, that's ok.

internal/operator-controller/resolve/bundle.go
perdasilva: I've gone ahead and added it =D

internal/operator-controller/resolve/bundle.go
perdasilva: I'll leave it just to assume a defensive posture here...

perdasilva:
closing as stale

PR #2261, opened 2025-10-10
:sparkles: Set Availability condition to Unknown on archived revisions

perdasilva:
# Description

This PR updates the ClusterExtensionRevision to set the `Available` condition to `Unknown` with reason `Archived` and message `revision is archived` on archived revisions.

It also adds the `AVAILABLE` print column to the ClusterExtensionRevision for visibility in `kubectl get clusterextensionrevision ...`

```
NAME              AVAILABLE   AGE
test-operator-1   Unknown     17s
test-operator-2   True        3s
```

We also remove the condition setting on finalizer errors due to transience. No need to hit the api server on something that will in all likelihood be fixed in the next reconciliation.

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/operator-controller/controllers/clusterextensionrevision_controller.go
+++ b/internal/operator-controller/controllers/clusterextensionrevision_controller.go
@@ -343,13 +343,6 @@ func (c *ClusterExtensionRevisionReconciler) teardown(ctx context.Context, rev *
 	}
 
 	if err := c.removeFinalizer(ctx, rev, clusterExtensionRevisionTeardownFinalizer); err != nil {
-		meta.SetStatusCondition(&rev.Status.Conditions, metav1.Condition{
-			Type:               ocv1.ClusterExtensionRevisionTypeAvailable,
-			Status:             metav1.ConditionFalse,
-			Reason:             ocv1.ClusterExtensionRevisionReasonReconcileFailure,
-			Message:            err.Error(),
-			ObservedGeneration: rev.Generation,
-		})
 		return ctrl.Result{}, fmt.Errorf("error removing teardown finalizer: %v", err)
 	}
 	return ctrl.Result{}, nil


internal/operator-controller/controllers/clusterextensionrevision_controller.go
```diff
@@ -118,6 +118,7 @@ func (c *ClusterExtensionRevisionReconciler) reconcile(ctx context.Context, rev
 
 	revision, opts, previous := toBoxcutterRevision(rev)
 
+	// nolint:nestif
```
pedjak: let's fix the reporting linting issue if possible?

internal/operator-controller/controllers/clusterextensionrevision_controller.go
```diff
@@ -153,11 +154,25 @@ func (c *ClusterExtensionRevisionReconciler) reconcile(ctx context.Context, rev
...
+			return ctrl.Result{}, nil
+		}
+
 		if err := c.removeFinalizer(ctx, rev, clusterExtensionRevisionTeardownFinalizer); err != nil {
 			meta.SetStatusCondition(&rev.Status.Conditions, metav1.Condition{
-				Type:               "Available",
+				Type:               ocv1.ClusterExtensionRevisionTypeAvailable,
 				Status:             metav1.ConditionFalse,
-				Reason:             "ReconcileFailure",
+				Reason:             ocv1.ClusterExtensionRevisionReasonReconcileFailure,
```
pedjak: IMHO we do not need to report in the status if there is an error in removing finalizer, because such errors are transient. to this point we have archived the revision, and only thing left is to remove the finalizer (btw, why we would like to remove it, if deletion is not requested)?

internal/operator-controller/controllers/clusterextensionrevision_controller.go
perdasilva: I couldn't figure out a way to do it that made sense.

internal/operator-controller/controllers/clusterextensionrevision_controller.go
perdasilva: I don't think this is unreasonable. But could we address it as a follow up to keep this PR as small as possible? Also, since I didn't really write this bit, it would be good to get Nico involved in the discussion (more so on the finalizer side). I don't have strong opinions here. Given that the resource is not managing anything and there's no need for the finalizer when archived, both can make sense - unless there's good reasons to block deletion if the controller isn't up.

perdasilva: I've removed the condition setting - but let's come back to the finalizer on Monday

internal/operator-controller/controllers/clusterextensionrevision_controller.go
perdasilva: I've moved the teardown procedure to its own method - that solved the problem

pedjak:
> /override codecov/patch

shall we investigate this job?

tmshort:
No, it was a matter of the size of the patch coverage. Admittedly, we have a threshold, and the difference (72% vs 76%) exceeded that.

pedjak:
> No, it was a matter of the size of the patch coverage. Admittedly, we have a threshold, and the difference (72% vs 76%) exceeded that.

How strictly we try to keep the coverage on a certain level? cc @perdasilva

PR #2264, opened 2025-10-13
:seedling: OPRUN-4194: Allow openshift catalog versions to be configurable via helm

tmshort:
# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/helm/olmv1/values.yaml
+++ b/helm/olmv1/values.yaml
@@ -20,6 +20,8 @@ options:
     enabled: false
   openshift:
     enabled: false
+    catalogs:
+      version: v4.20
   # This can be one of: standard or experimental
   featureSet: standard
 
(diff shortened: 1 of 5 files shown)


helm/olmv1/values.yaml
```diff
@@ -22,6 +22,7 @@ options:
     enabled: false
   # This can be one of: standard or experimental
   featureSet: standard
+  catalogVersion: v4.20
```
pedjak: how about to make a struct like this:

```yaml
catalog:
  version: v4.20
```

it opens up a room for future extension without having long descriptive variable names?

tmshort: Yeah, I did ponder that...

tmshort:
/unhold
Ok, I _actually_ put it under the openshift option, because it is openshift specific.

helm/olmv1/values.yaml
```diff
@@ -20,6 +20,8 @@ options:
     enabled: false
   openshift:
     enabled: false
+    catalogs:
```
pedjak, line 23: should we have mutiple catalogs with the same version or it should be

```
catalog:
  version: v4.20
```

tmshort: Yeah, I thought of this... and it doesn't matter too much.
I put it under `openshift`.

PR #2266, opened 2025-10-14
:sparkles: Update .spec.config unmarshalling to accept yaml and json inputs

perdasilva:
# Description

#### Changes

1. Makes ClusterExtension `.spec.config.inline.Raw` support both yaml and json input by switching to `k8s.io/yaml` to unmarshall the content
2. Adds additional unit tests to exercise ClusterExtension .spec.config.inline.Raw input validation
3. Updates the ClusterExtension `.spec.config`  and `.spec.config.inline` doc strings to call out that it accepts arbitrary yaml/json objects that are validated at runtime by a schema provided by the underlying bundle

#### Motivation

Ensure ClusterExtension .spec.config.inline can handle both JSON and YAML formats, which are supported by Kubernetes and that validation is carried out whether the input passed through client- and/or server-side validation before reaching the controller or not.

Additionally, to ensure that the api documentation highlights that configuration schema is defined by the bundle and the input is validated at runtime against that schema.

## Reviewer Checklist

- [ ] API Go Documentation


internal/operator-controller/applier/provider_test.go
```diff
@@ -188,6 +188,66 @@ func Test_RegistryV1ManifestProvider_WebhookSupport(t *testing.T) {
...
+		{
+			name:        "accepts json config",
+			configBytes: []byte(`{"watchNamespace": "some-namespace"}`),
+		},
+		{
+			name:        "accepts yaml config",
+			configBytes: []byte(`watchNamespace: some-namespace`),
+		},
+		{
+			name:               "rejects valid json",
```
pedjak: you mean `reject invalid json`?

internal/operator-controller/applier/provider_test.go
```diff
@@ -188,6 +188,66 @@ func Test_RegistryV1ManifestProvider_WebhookSupport(t *testing.T) {
...
+			configBytes:        []byte(`{"hello`),
+			expectedErrMessage: `inline config is not a valid JSON/YAML object`,
+		},
+		{
+			name:               "rejects valid non-object json",
```
pedjak: similar as above?

perdasilva: no - here something like `true` is valid json, but not an object

perdasilva: I've renamed it to hopefully be more clear

internal/operator-controller/applier/provider_test.go
perdasilva: fixed

internal/operator-controller/applier/provider.go
```diff
@@ -92,8 +94,9 @@ func (r *RegistryV1ManifestProvider) getWatchNamespace(ext *ocv1.ClusterExtensio
 		cfg := struct {
 			WatchNamespace string `json:"watchNamespace"`
 		}{}
-		if err := json.Unmarshal(ext.Spec.Config.Inline.Raw, &cfg); err != nil {
-			return "", fmt.Errorf("invalid bundle configuration: %w", err)
+		// Using k8s.io/yaml package as that is able to handle both json and yaml
+		if err := yaml.Unmarshal(ext.Spec.Config.Inline.Raw, &cfg); err != nil {
+			return "", errors.New("inline config is not a valid JSON/YAML object")
```
pedjak: IMHO, we cannot get an error here at parsing because k8s api-server reject invalid payload. Hence, it is enough to do:

```golang
_ = yaml.Unmarshal(ext.Spec.Config.Inline.Raw, &cfg);
```

See similar approach at Flux [HelmRelease](https://github.com/fluxcd/helm-controller/blob/main/api/v2/helmrelease_types.go#L1327-L1332) resource.

pedjak:
> Improves the unmarshalling of .spec.config by ensuring both yaml and json definitions are acceptable by using k8s.io/yaml package.

Are we solving here a bug or there is in user facing experience?

perdasilva:
> Are we solving here a bug or there is in user facing experience?

I had the same question. I originally had :bug: but then because its not released made it :sparkles: - but I don't have a strong opinion. Happy to switch to :bug: if this makes more sense.

camilamacedo86:
It would be nice we ensure good titles and the right emojis since that is the info that generates our release notes. So, whatever we add as PR title will be our release notes.

perdasilva:
> It would be nice we ensure good titles and the right emojis since that is the info that generates our release notes. So, whatever we add as PR title will be our release notes.

No doubt. The question is which is the right emoji?

internal/operator-controller/applier/provider.go
perdasilva: We can still get an error here because we aren't unmarshalling to a `map[string]interface{}`. But this does points to another issue with the error message: the input could still be valid  JSON/YAML just not valid registry+v1 configuration.

camilamacedo86:
If is a bug 🐛 
But it seems to me like an enhancement, so I would keep ✨ as it is.
I am OK with both, I have no strong opnion here

internal/operator-controller/applier/provider.go
perdasilva: Another case where it _could_ produce an error is if you do something like `kubectl create --raw ...` I think...but that's an edge case

perdasilva: I'd also like to follow this PR up with some kind of ingress/validation tests. I'm wondering if those would be too heavy to be in the e2e suite (in the sense that the cost of execution is kinda of high - deploy cluster, etc.). But, I'm not sure if there's an easy way to exercise validation outside of the e2e suite. Any ideas?

pedjak: > We can still get an error here because we aren't unmarshalling to a `map[string]interface{}`. But this does points to another issue with the error message: the input could still be valid JSON/YAML just not valid registry+v1 configuration.

why do we unmarshall to a private type here? We should document this in the API, because it looks that `inline` field can be an arbitrary thing.

pedjak:
> No doubt. The question is which is the right emoji?

@perdasilva Can you perhaps rephrase PR title and description to highlight the concrete improvement?

internal/operator-controller/applier/provider.go
perdasilva: it's arbitrary from the API side but it will be specific to the underlying bundle type:
- Helm -> values files and gated by values.json.schema
- registry+v1 -> its own internal configuration structure
- <future bundle format> -> whatever it uses

perdasilva: I've updated the API documentation to reflect this.

pedjak: > I've updated the API documentation to reflect this.

Can you be more specific here, instead of stating

```
 inline accepts arbitrary JSON/YAML objects whose validation is carried out at runtime against the schema provided by the bundle if one is provided.
```

so that we can help users to understand better what can be put there. Something along your previous comment:

```
Helm -> values files and gated by values.json.schema
registry+v1 -> its own internal configuration structure
```

with documenting how this registry+v1 structure looks like helps more.

api/v1/clusterextension_types.go
```diff
@@ -174,6 +176,8 @@ type ClusterExtensionConfig struct {
 	// ClusterExtension.
 	//
 	// inline must be set if configType is 'Inline'.
+	// inline accepts arbitrary JSON/YAML objects whose validation is carried out at runtime against the schema provided
+	// by the bundle if one is provided.
```
pedjak: Can we be more specific here, instead of stating

```
 inline accepts arbitrary JSON/YAML objects whose validation is carried out at runtime against the schema provided by the bundle if one is provided.
```

so that we can help users to understand better what can be put there. Something along your previous comment:

```
Helm -> values files and gated by values.json.schema
registry+v1 -> its own internal configuration structure
```

with documenting how this registry+v1 structure looks like helps more.

perdasilva: I kept those out of the documentation on purpose because I don't think we want to make the bundle format a user concern and we don't yet support the Helm format anyway.

Maybe there's a better way to communicate that the configuration schema is optionally provided by the bundle and the input is validated against that, but I'm struggling to find it. Maybe you could suggest changes and we could iterate together on that? 

I don't think we should call out formats here though. The mental model is:
- bundles optionally provide a config schema
- users can set the configuration for the bundle through .spec.config
- the user input is validated at runtime against the configuration schema provided by the bundle
- if the bundle does not provide a configuration schema the process will continue on a best-effort basis (enforcing configuration schema is a catalog policy concern at this time)

Would even the list above be an improvement?

pedjak: > Would even the list above be an improvement?

IMHO, yes.

pedjak (approved):

tmshort (approved):
/approve
/lgtm

PR #2267, opened 2025-10-14
✨ Promote Webhook FeatureGates to GA (OPRUN-4098)

camilamacedo86:


docs/draft/howto/enable-webhook-support.md
```diff
@@ -5,7 +5,7 @@ Webhook support is enabled by default. The controller uses the `WebhookProviderC
 feature gate unless you override it. To switch to the OpenShift Service CA provider,
 start the controller with `--feature-gates=WebhookProviderCertManager=false`.
 
-OLMv1 currently does not support the installation of bundles containing webhooks. The webhook support feature enables this capability.
+OLMv1 now supports the installation of bundles containing webhooks out of the box.
```
camilamacedo86: @michaelryanpeter could you please help us out and give a hand on the review?

test/e2e/webhook_support_test.go
tmshort, line 1: This is an example of why multiple files should be used in the experimenta-e2e directory. You _should_ just be able to move the file without significant changes. In this case, we had to create a new file, and delete content from another. :(

docs/draft/howto/enable-webhook-support.md
```diff
@@ -1,10 +1,11 @@
 ## Installation of Bundles containing Webhooks
 
 !!! note
-This feature is still in *alpha*. Either the `WebhookProviderCertManager`, or the `WebhookProviderOpenshiftServiceCA`, feature-gate 
-must be enabled to make use of it. See the instructions below on how to enable the feature-gate.
+Webhook support is enabled by default. The controller uses the `WebhookProviderCertManager`
+feature gate unless you override it. To switch to the OpenShift Service CA provider,
+start the controller with `--feature-gates=WebhookProviderCertManager=false`.
```
tmshort: This is incomplete, because you also would need to enable the `WebhookProviderOpenshiftServiceCA` feature-gate.

docs/draft/howto/enable-webhook-support.md
```diff
@@ -1,10 +1,11 @@
 ## Installation of Bundles containing Webhooks
 
 !!! note
-This feature is still in *alpha*. Either the `WebhookProviderCertManager`, or the `WebhookProviderOpenshiftServiceCA`, feature-gate 
-must be enabled to make use of it. See the instructions below on how to enable the feature-gate.
+Webhook support is enabled by default. The controller uses the `WebhookProviderCertManager`
+feature gate unless you override it. To switch to the OpenShift Service CA provider,
```
tmshort:
```suggestion
feature-gate unless you override it. To switch to the OpenShift Service CA provider,
```
Use hyphens in "feature-gate"

docs/draft/howto/enable-webhook-support.md
tmshort:
```suggestion
OLMv1 supports the installation of bundles containing webhooks by default.
```
We should not be referencing time (i.e. "now").

docs/draft/tutorials/explore-available-content-metas-endpoint.md
```diff
@@ -92,7 +92,7 @@ Then you can query the catalog by using `curl` commands and the `jq` CLI tool to
         ```
 
     !!! important
-        Currently, OLM 1.0 does not support the installation of extensions that use webhooks or that target a single or specified set of namespaces.
+        OLM 1.0 supports installing extensions that define webhooks. Targeting a single or specified set of namespaces still requires enabling the `SingleOwnNamespaceInstallSupport` feature gate.
```
tmshort:
```suggestion
        OLM 1.0 supports installing extensions that define webhooks. Targeting a single or specified set of namespaces requires enabling the `SingleOwnNamespaceInstallSupport` feature-gate.
```
Remove "still" as it implies prior knowledge; use the present tense. Use a hyphen in "feature-gate".

docs/tutorials/explore-available-content.md
```diff
@@ -92,7 +92,7 @@ Then you can query the catalog by using `curl` commands and the `jq` CLI tool to
         ```
 
     !!! important
-        Currently, OLM 1.0 does not support the installation of extensions that use webhooks or that target a single or specified set of namespaces.
+        OLM 1.0 supports installing extensions that define webhooks. Targeting a single or specified set of namespaces still requires enabling the `SingleOwnNamespaceInstallSupport` feature gate.
```
tmshort:
```suggestion
        OLM 1.0 supports installing extensions that define webhooks. Targeting a single or specified set of namespaces requires enabling the `SingleOwnNamespaceInstallSupport` feature-gate.
```
Same reasoning as above.

docs/draft/howto/enable-webhook-support.md
camilamacedo86: WebhookProviderOpenshiftServiceCA is  enabled by default with this PR. 
Users if wish need disable, see: internal/operator-controller/features/features.go

So, wdyt? Do you have you any suggestion here?

tmshort: Actually... _that's_ a problem. **Both of them can not be enabled by default.** Only one of them should be enabled at a time, and those are in different deployment scenarios (cert-manager vs Openshift).

The helm config _does not_ provide a way to disable feature gates, only enable them.
So either:
a) We need to keep them disabled and add the set of feature flags to values.yaml as default on, which also means they would need to be kept in the experimental.yaml file as well.
b) We need to keep them disabled and add the feature-flags with a check for `.Values.options.openshift.enabled` to add the appropriate enabled=True feature-flag.
c) Inverse of (b) keep flags enabled, and then disabled=True based on `.Values.options.openshift.enabled'

Personally, I dislike negative logic, and would prefer that (b) is used, _so it's explicitly clear what's enabled_, rather than "disabling" the one we don't want.

tmshort:
/hold
We can't have both feature gates enabled simultaneously, as one is meant for upstream (`WebhookProviderCertManager`), and the other is meant for downstream (`WebhookProviderOpenshiftServiceCA`).

test/e2e/webhook_support_test.go
perdasilva: I'm in favor of removing the experimental-e2e entirely and introspecting the cluster to see which features are enabled and run/skip accordingly - or even informing it about which environment its running in: standard/experimental. That removes all of this complexity, all e2es live together where they can share packages and libraries easily under the e2e directory, we could even change the flow of existing e2es to keep duplication low, etc.

docs/draft/howto/enable-webhook-support.md
perdasilva: Would introducing another flag to the command like --certificate-provider (that can be one of openshift-serviceca, or cert-manager (default: cert-manager)) help here? Then we can have them both on by default. Then downstream Deployment manifest uses `--certificate-provider=openshift-serviceca` and upstream Deployment manifest uses `--certificate-provider=cert-manager` (or nothing since its the default). Upstream users can then choose which to use to use openshift-serviceca if they want to by altering the Deployment command.

tmshort:
I think my preference would be to leave the feature-gates off by default, and put just the one we want into the `helm/olmv1/values.yaml` file, but this would mean also leaving them in the `experimental.yaml` file as well, which is not desirable (i.e. have it in two places). The alternative is to use an `if/.Values.openshift.enabled/else` sequence to separate the two flags.

BUT, the strangeness comes from the fact that we have cluster-olm-operator getting feature-gates from Openshift. I have a downstream change that would "remove" all upstream gates from `values.yaml`, and that might be needed before this is merged downstream.

BUT, another strangeness occurs because helm does not merge/combine lists; they are overwritten.

I need to think about this some more.

camilamacedo86:
/hold 

Until we decide how to do

tmshort:
That we have two feature-gates that can't be enabled simultaneously is causing us this headache (this should not be an issue with 99%of other features).

Although the CertManager one takes priority: https://github.com/operator-framework/operator-controller/blob/1d685b1856a63df2ae3766c67aad414086dcb6ba/cmd/operator-controller/main.go#L517-L524

Which I guess is a reasonable default. But it means we have to explicitly turn off CertManager to get OpenshiftServiceCA, and we don't really have that capability downstream.

Due to the downstream mapping of feature-gates, we only handle "enabled" flags (i.e. we can't turn things off), and we couldn't turn things off without a way in the helm chart to do so (i.e. additional upstream effort). 

And disabling features just sounds wrong to me. I really think we need to _enable_ features.

Add in the fact that helm doesn't merge lists makes this more challenging.

I'm envisioning a number of possible solutions, none of which I really like, so I'm wondering if this is something to be discussed in a meeting (office hours)?

tmshort:
Options?

1. Keep these feature-gates off, explicitly turn on CertManager upstream through a new `standard.yaml` file, which means that `experimental.yaml` would also need to include it. Downstream, the OpenshiftServiceCA value would be enabled via the Openshift feature-gates.

1a. Add a new "standardFeatures" list that parallels the existing helm feature lists so that the value doesn't have to be duplicated in both `standard.yaml` and `experimental.yaml`. (this is a bunch of effort to avoid duplication of a value in a list).

2. Change the feature-gates to be on, but then add upstream _and_ downstream _and_ cluster-olm-operator mechanisms to be able to disable the feature gates. This would be a lot of additional work.

3. Add a new `--webhook-provider` option to operator-controller to select between CertManager and OpenshiftServiceCA (the default would be no provider). This effectively replaces the feature-gates with a tri-state, and would be enabled in the chart via `.Values.openshift.enabled` or `.Values.certManager.enabled` settings.

test/e2e/webhook_support_test.go
camilamacedo86: We are moving the tests from `experimental_e2e_test.go` to `e2e_suite_test.go`.

At the moment, not all tests live in a single file. 
It might not be fully achievable in every case if we decide to enforce that as a rule, since doing so could require duplicating some code or setups.

However, anyway, IHMO refactoring the e2e tests related to the feature gate is **out of scope** for this PR. We can discuss that, but I think we need to be in a second moment.  
Here, we’re simply promoting the feature and moving its specific tests into the main `e2e` suite.

docs/draft/howto/enable-webhook-support.md
camilamacedo86: I really liked @perdasilva’s idea 💡

The main concern with not keeping it enabled by default is the **user experience** — having too many flags to enable could quickly become cumbersome as this list grows.

We should also think about the long term. At some point, we’ll likely remove the feature flag entirely and make the code part of the main source.  
Maybe it’s worth discussing how we want to handle that transition and what our long-term approach should be? 🤔

camilamacedo86: I pushed a commit with @perdasilva's idea see the latest one. 
@tmshort @perdasilva

test/e2e/webhook_support_test.go
tmshort: Common packages and libraries and functions could still be shared by different directories, those libraries would need to live elsewhere. So, that's a false argument.
The idea was that each feature would have a single file devoted to its tests, and that it _should have_ been a simple file move/package rename to move the tests from experimental to the e2e. I don't see how that's complex...

Adding the code for introspection would be the complexity, IMHO. But, it would work downstream in experimental-e2e better.

cmd/operator-controller/main.go
```diff
@@ -159,6 +162,7 @@ func init() {
 	flags.StringVar(&cfg.cachePath, "cache-path", "/var/cache", "The local directory path used for filesystem based caching")
 	flags.StringVar(&cfg.systemNamespace, "system-namespace", "", "Configures the namespace that gets used to deploy system resources.")
 	flags.StringVar(&cfg.globalPullSecret, "global-pull-secret", "", "The <namespace>/<name> of the global pull secret that is going to be used to pull bundle images.")
+	flags.StringVar(&cfg.certificateProvider, "certificate-provider", certificateProviderCertManager, "The certificate provider to use for webhook support. Options: 'cert-manager' (default) or 'openshift-serviceca'.")
```
tmshort: I really don't think we can have this enabled to cert-manager by default, because we have the option to run _without_ cert-manager.
I do like the idea, however.

helm/olmv1/templates/deployment-olmv1-system-operator-controller-controller-manager.yml
```diff
@@ -44,6 +44,9 @@ spec:
             {{- if not .Values.options.tilt.enabled }}
             - --leader-elect
             {{- end }}
+            {{- if .Values.options.openshift.enabled }}
+            - --certificate-provider=openshift-serviceca
+            {{- end }}
```
tmshort:
```suggestion
            {{- if .Values.options.certManager.enabled }}
            - --certificate-provider=cert-manager
            {{- else if .Values.options.openshift.enabled }}
            - --certificate-provider=openshift-serviceca
            {{- end }}
```

cmd/operator-controller/main.go
tmshort:
```suggestion
	flags.StringVar(&cfg.certificateProvider, "", certificateProviderCertManager, "The certificate provider to use for webhook support. Options: 'cert-manager' or 'openshift-serviceca', disabled by default.")
```

cmd/operator-controller/main.go
```diff
@@ -514,13 +522,14 @@ func run() error {
...
-		return certproviders.CertManagerCertificateProvider{}
-	} else if features.OperatorControllerFeatureGate.Enabled(features.WebhookProviderOpenshiftServiceCA) {
-		return certproviders.OpenshiftServiceCaCertificateProvider{}
+func getCertificateProvider() (render.CertificateProvider, error) {
+	switch strings.ToLower(cfg.certificateProvider) {
+	case "", certificateProviderCertManager:
+		return certproviders.CertManagerCertificateProvider{}, nil
+	case certificateProviderOpenshiftCA:
+		return certproviders.OpenshiftServiceCaCertificateProvider{}, nil
 	}
```
tmshort:
```suggestion
	switch strings.ToLower(cfg.certificateProvider) {
	case "":
	    return nil, nil
	certificateProviderCertManager:
		return certproviders.CertManagerCertificateProvider{}, nil
	case certificateProviderOpenshiftCA:
		return certproviders.OpenshiftServiceCaCertificateProvider{}, nil
	}
```

tmshort:
There is no default certificate provider in OLMv1. We provide explicit options to turn on and use cert-manager and/or Openshift serviceca, and I think we need to be able to reflect that.

Some fairly small changes would be needed to make the `--certificate-provider` a true tri-state.

internal/operator-controller/features/features.go
```diff
@@ -46,26 +44,6 @@ var operatorControllerFeatureGates = map[featuregate.Feature]featuregate.Feature
...
-	// WebhookProviderCertManager enables support for installing
-	// registry+v1 cluster extensions that include validating,
-	// mutating, and/or conversion webhooks with Openshift Service CA
-	// as the certificate provider.
-	WebhookProviderOpenshiftServiceCA: {
-		Default:       false,
-		PreRelease:    featuregate.Alpha,
-		LockToDefault: false,
-	},
-
```
tmshort, line 68: My fear is that this will break downstream... we need to remove downstream feature gate references before we can delete this.
Specifically, the techpreview/devpreview builds will fail. So, this should remain as false, and should not be looked at. As long as we can accept the feature-gate value, we're good. Otherwise, operator-controller will exit early complaining of an unknown feature gate.

internal/operator-controller/features/features.go
```diff
@@ -11,13 +11,11 @@ import (
 const (
 	// Add new feature gates constants (strings)
 	// Ex: SomeFeature featuregate.Feature = "SomeFeature"
-	PreflightPermissions              featuregate.Feature = "PreflightPermissions"
-	SingleOwnNamespaceInstallSupport  featuregate.Feature = "SingleOwnNamespaceInstallSupport"
-	SyntheticPermissions              featuregate.Feature = "SyntheticPermissions"
-	WebhookProviderCertManager        featuregate.Feature = "WebhookProviderCertManager"
-	WebhookProviderOpenshiftServiceCA featuregate.Feature = "WebhookProviderOpenshiftServiceCA"
```
tmshort, line 18: Same here, we can't really delete feature-gates until it's no longer referenced downstream.

cmd/operator-controller/main.go
camilamacedo86: TODO: Change for: --webhook-certificate-provider

test/e2e/webhook_support_test.go
camilamacedo86: Now, I understand the suggestion.
Done. It makes sense. So closing this one.

docs/draft/howto/enable-webhook-support.md
camilamacedo86: We discussed it and decided that it will not be done now to make things simple for the downstream
We will look into that in the future.

helm/olmv1/templates/deployment-olmv1-system-operator-controller-controller-manager.yml
camilamacedo86: Outdated
We will no longer do it for now.

cmd/operator-controller/main.go
camilamacedo86: Outdated
We will no longer do it for now.

internal/operator-controller/features/features.go
camilamacedo86: Outdated
We will no longer do it for now.

internal/operator-controller/features/features.go
camilamacedo86: Outdated
We will no longer do it for now.

cmd/operator-controller/main.go
camilamacedo86: Outdated
We will no longer do it for now.

camilamacedo86:
Hey folks 👋  

@perdasilva @tmshort 

I’ve updated this PR — it now reflects what we discussed.  
It looks good to merge once we get the green light from you and downstream is ready to receive it.  

Please feel free to review when you get a chance 🙏

docs/draft/howto/enable-webhook-support.md
```diff
@@ -1,11 +1,12 @@
...
-must be enabled to make use of it. See the instructions below on how to enable the feature-gate.
+OLMv1 supports the installation of bundles containing webhooks by default.
+The controller uses the `WebhookProviderCertManager`
+feature-gate unless you override it. To switch to the OpenShift Service CA provider,
+start the controller with `--feature-gates=WebhookProviderCertManager=false` and enable `--feature-gates=WebhookProviderOpenshiftServiceCA=true`.
 
-OLMv1 currently does not support the installation of bundles containing webhooks. The webhook support feature enables this capability.
-Webhooks, or more concretely Admission Webhooks, are part of Kuberntes' [Dynamic Admission Control](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/)
+Webhooks, or more concretely Admission Webhooks, are part of Kubernetes' [Dynamic Admission Control](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/)
 feature. Webhooks run as services called by the kube-apiservice in due course of processing a resource related request. They can be used to validate resources, ensure reasonable default values,
```
michaelryanpeter:
```suggestion
Admission webhooks are part of the Kubernetes suite of [Dynamic Admission Control](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/)
plugins. Webhooks run as services called by the kube-apiservice in due course of processing a resource related request. They can be used to validate resources, ensure reasonable default values,
```

docs/draft/howto/enable-webhook-support.md
```diff
@@ -1,11 +1,12 @@
 ## Installation of Bundles containing Webhooks
 
 !!! note
-This feature is still in *alpha*. Either the `WebhookProviderCertManager`, or the `WebhookProviderOpenshiftServiceCA`, feature-gate 
-must be enabled to make use of it. See the instructions below on how to enable the feature-gate.
+OLMv1 supports the installation of bundles containing webhooks by default.
+The controller uses the `WebhookProviderCertManager`
+feature-gate unless you override it. To switch to the OpenShift Service CA provider,
+start the controller with `--feature-gates=WebhookProviderCertManager=false` and enable `--feature-gates=WebhookProviderOpenshiftServiceCA=true`.
```
michaelryanpeter:
```suggestion
OLM v1 supports installing bundles that use admission webhooks.
By default, OLM v1 uses the community Cert Manager package for admission webhook validation. To use the OpenShift Service CA provider, set the `--feature-gates=WebhookProviderOpenshiftServiceCA=true` flag at startup.
```

camilamacedo86: We have more than one kind of webhook that is supported. 
certManager is used only for the admission webhook 
So, I think it would not be very accurate. 

The other suggestion was accepted :-)

michaelryanpeter: I'm confused, isn't that what I wrote?
"By default, OLM v1 uses the community Cert Manager package for admission webhook validation."

(I'm totally fine if you don't take the suggestion, but I thought we are both saying the same thing?)

camilamacedo86: We need to have "OLMv1 supports the installation of bundles containing webhooks by default." 
(because we support more than admission webhooks)
Then, below, we explain that for the specific case, admission, certmanager is used. 

So, I updated it with a mix of your suggestions.

docs/draft/howto/enable-webhook-support.md
```diff
@@ -1,12 +1,13 @@
...
```
camilamacedo86:
```suggestion
By default, OLM v1 uses the community Cert Manager package for admission webhook via the feature-gate flag `WebhookProviderCertManager`. To use the OpenShift Service CA provider, set the `--feature-gates=WebhookProviderOpenshiftServiceCA=true` flag at startup.
```

camilamacedo86:
/hold cancel

@tmshort @perdasilva as we discussed lets merge this one first. 
Waiting for your approvals.

PR #2273, opened 2025-10-16
🐛 Add support for build metadata precedence in bundle version comparison

joelanford:
# Description

This change fixes an issue to ensure that operator-controller properly handles and compares registry+v1 bundle versions that include build metadata as specified in the semver version.

The intention is that we only treat build metadata as a release value for registry+v1 bundles, which already have this precedent set. If/when operator-controller gains support for new bundle types, the intention is to avoid continuing the practice (and semver violation) of treating version build metadata as comparable/orderable.

Key changes:
- Introduce VersionRelease type combining semver version with release metadata
- Update bundle comparison logic to use build metadata for ordering within registry+v1 bundles
- Version field matching ignores build metadata (consistent with semver specification)
- Build metadata is only considered for precedence when comparing otherwise equal versions
- VersionRelease parsing tolerates non-release build metadata while preserving the original version
- Replace GetVersion with GetVersionAndRelease across the codebase

This is particularly important for registry+v1 bundles that encode release information in the build metadata field of their version strings.

The ClusterExtension version field continues to work semantically as before. It ignores build metadata when matching versions (consistent with semver specification), even when build metadata is explicitly included in the version field. Build metadata is only used internally to determine precedence when ordering registry+v1 bundles with otherwise equal versions.

The bug this fixes was originally reported by Red Hat in https://issues.redhat.com/browse/OCPBUGS-60424

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/operator-controller/bundle/versionrelease.go
+++ b/internal/operator-controller/bundle/versionrelease.go
@@ -65,14 +65,22 @@ func (vr *VersionRelease) Compare(other VersionRelease) int {
 	return vr.Release.Compare(other.Release)
 }
 
+// AsLegacyRegistryV1Version converts a VersionRelease into a standard semver version.
+// If the VersionRelease's Release field is set, the returned semver version's build
+// metadata is set to the VersionRelease's Release. Otherwise, the build metadata is
+// set to the VersionRelease's Version field's build metadata.
 func (vr *VersionRelease) AsLegacyRegistryV1Version() bsemver.Version {
-	return bsemver.Version{
+	v := bsemver.Version{
 		Major: vr.Version.Major,
 		Minor: vr.Version.Minor,
 		Patch: vr.Version.Patch,
 		Pre:   vr.Version.Pre,
-		Build: slicesutil.Map(vr.Release, func(i bsemver.PRVersion) string { return i.String() }),
+		Build: vr.Version.Build,
 	}
+	if len(vr.Release) > 0 {
+		v.Build = slicesutil.Map(vr.Release, func(i bsemver.PRVersion) string { return i.String() })
+	}
+	return v
 }
 
 type Release []bsemver.PRVersion
--- a/internal/operator-controller/bundleutil/bundle_test.go
+++ b/internal/operator-controller/bundleutil/bundle_test.go
@@ -4,7 +4,6 @@ import (
 	"encoding/json"
 	"testing"
 
-	"github.com/blang/semver/v4"
 	bsemver "github.com/blang/semver/v4"
 	"github.com/stretchr/testify/require"
 
@@ -29,7 +28,7 @@ func TestGetVersionAndRelease(t *testing.T) {
 				Value: json.RawMessage(`{"version": "1.0.0-pre+1.alpha.2"}`),
 			},
 			wantVersionRelease: &bundle.VersionRelease{
-				Version: semver.MustParse("1.0.0-pre"),
+				Version: bsemver.MustParse("1.0.0-pre"),
 				Release: bundle.Release([]bsemver.PRVersion{
 					{VersionNum: 1, IsNum: true},
 					{VersionStr: "alpha"},
@@ -53,7 +52,7 @@ func TestGetVersionAndRelease(t *testing.T) {
 				Value: json.RawMessage(`{"version": "1.0.0+001"}`),
 			},
 			wantVersionRelease: &bundle.VersionRelease{
-				Version: semver.MustParse("1.0.0+001"),
+				Version: bsemver.MustParse("1.0.0+001"),
 			},
 			wantErr: false,
 		},
--- a/internal/operator-controller/catalogmetadata/compare/compare_test.go
+++ b/internal/operator-controller/catalogmetadata/compare/compare_test.go
@@ -19,40 +19,68 @@ func TestNewVersionRange(t *testing.T) {
 	type testCase struct {
 		name         string
 		versionRange string
-		inputVersion bsemver.Version
-		expect       bool
+		assertFunc   func(t *testing.T, vr bsemver.Range, err error)
 	}
 	for _, tc := range []testCase{
 		{
+			name:         "version without build metadata matches range with build metadata",
 			versionRange: "1.0.0+1",
-			inputVersion: bsemver.MustParse("1.0.0"),
-			expect:       true,
+			assertFunc: func(t *testing.T, vr bsemver.Range, err error) {
+				require.NoError(t, err)
+				assert.True(t, vr(bsemver.MustParse("1.0.0")))
+			},
 		},
 		{
+			name:         "version with different build metadata matches range with build metadata",
 			versionRange: "1.0.0+1",
-			inputVersion: bsemver.MustParse("1.0.0+2"),
-			expect:       true,
+			assertFunc: func(t *testing.T, vr bsemver.Range, err error) {
+				require.NoError(t, err)
+				assert.True(t, vr(bsemver.MustParse("1.0.0+2")))
+			},
 		},
 		{
+			name:         "version with same build metadata matches range with build metadata",
 			versionRange: "1.0.0+1",
-			inputVersion: bsemver.MustParse("1.0.0+1"),
-			expect:       true,
+			assertFunc: func(t *testing.T, vr bsemver.Range, err error) {
+				require.NoError(t, err)
+				assert.True(t, vr(bsemver.MustParse("1.0.0+1")))
+			},
 		},
 		{
+			name:         "exact version match without build metadata",
 			versionRange: "1.0.0",
-			inputVersion: bsemver.MustParse("1.0.0"),
-			expect:       true,
+			assertFunc: func(t *testing.T, vr bsemver.Range, err error) {
+				require.NoError(t, err)
+				assert.True(t, vr(bsemver.MustParse("1.0.0")))
+			},
 		},
 		{
+			name:         "version with build metadata matches range without build metadata",
 			versionRange: "1.0.0",
-			inputVersion: bsemver.MustParse("1.0.0+1"),
-			expect:       true,
+			assertFunc: func(t *testing.T, vr bsemver.Range, err error) {
+				require.NoError(t, err)
+				assert.True(t, vr(bsemver.MustParse("1.0.0+1")))
+			},
+		},
+		{
+			name:         "invalid range returns error",
+			versionRange: "not-a-valid-version",
+			assertFunc: func(t *testing.T, vr bsemver.Range, err error) {
+				require.Error(t, err)
+			},
+		},
+		{
+			name:         "version does not match range",
+			versionRange: ">=2.0.0",
+			assertFunc: func(t *testing.T, vr bsemver.Range, err error) {
+				require.NoError(t, err)
+				assert.False(t, vr(bsemver.MustParse("1.0.0")))
+			},
 		},
 	} {
 		t.Run(tc.name, func(t *testing.T) {
 			versionRange, err := compare.NewVersionRange(tc.versionRange)
-			require.NoError(t, err)
-			assert.Equal(t, tc.expect, versionRange(tc.inputVersion))
+			tc.assertFunc(t, versionRange, err)
 		})
 	}
 }
--- a/internal/operator-controller/bundle/versionrelease_test.go
+++ b/internal/operator-controller/bundle/versionrelease_test.go
@@ -3,11 +3,117 @@ package bundle_test
 import (
 	"testing"
 
+	bsemver "github.com/blang/semver/v4"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
 	"github.com/operator-framework/operator-controller/internal/operator-controller/bundle"
 )
+
+func TestNewLegacyRegistryV1VersionRelease(t *testing.T) {
+	type testCase struct {
+		name   string
+		input  string
+		expect func(*testing.T, *bundle.VersionRelease, error)
+	}
+	for _, tc := range []testCase{
+		{
+			name:  "empty input",
+			input: "",
+			expect: func(t *testing.T, _ *bundle.VersionRelease, err error) {
+				assert.Error(t, err)
+			},
+		},
+		{
+			name:  "v-prefix is invalid",
+			input: "v1.0.0",
+			expect: func(t *testing.T, _ *bundle.VersionRelease, err error) {
+				assert.Error(t, err)
+			},
+		},
+		{
+			name:  "valid semver, valid build metadata, but not a valid release",
+			input: "1.2.3-alpha+4.005",
+			expect: func(t *testing.T, vr *bundle.VersionRelease, err error) {
+				require.NoError(t, err)
+				assert.Equal(t, uint64(1), vr.Version.Major)
+				assert.Equal(t, uint64(2), vr.Version.Minor)
+				assert.Equal(t, uint64(3), vr.Version.Patch)
+				assert.Equal(t, []bsemver.PRVersion{{VersionStr: "alpha"}}, vr.Version.Pre)
+				assert.Equal(t, []string{"4", "005"}, vr.Version.Build)
+				assert.Empty(t, vr.Release)
+			},
+		},
+		{
+			name:  "valid semver, valid build metadata, valid release",
+			input: "1.2.3-alpha+4.5",
+			expect: func(t *testing.T, vr *bundle.VersionRelease, err error) {
+				require.NoError(t, err)
+				assert.Equal(t, uint64(1), vr.Version.Major)
+				assert.Equal(t, uint64(2), vr.Version.Minor)
+				assert.Equal(t, uint64(3), vr.Version.Patch)
+				assert.Equal(t, []bsemver.PRVersion{{VersionStr: "alpha"}}, vr.Version.Pre)
+				assert.Empty(t, vr.Version.Build)
+				assert.Equal(t, bundle.Release([]bsemver.PRVersion{
+					{VersionNum: 4, IsNum: true},
+					{VersionNum: 5, IsNum: true},
+				}), vr.Release)
+			},
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			actual, err := bundle.NewLegacyRegistryV1VersionRelease(tc.input)
+			tc.expect(t, actual, err)
+		})
+	}
+}
+
+func TestVersionRelease_AsLegacyRegistryV1Version(t *testing.T) {
+	type testCase struct {
+		name   string
+		input  bundle.VersionRelease
+		expect bsemver.Version
+	}
+	for _, tc := range []testCase{
+		{
+			name: "Release is set, so release is set as build metadata",
+			input: bundle.VersionRelease{
+				Version: bsemver.Version{
+					Major: 1,
+					Minor: 2,
+					Patch: 3,
+				},
+				Release: bundle.Release([]bsemver.PRVersion{{VersionStr: "release"}}),
+			},
+			expect: bsemver.Version{
+				Major: 1,
+				Minor: 2,
+				Patch: 3,
+				Build: []string{"release"},
+			},
+		},
+		{
+			name: "Release is unset, so version build metadata is set as build metadata",
+			input: bundle.VersionRelease{
+				Version: bsemver.Version{
+					Major: 1,
+					Minor: 2,
+					Patch: 3,
+					Build: []string{"buildmetadata"},
+				},
+			},
+			expect: bsemver.Version{
+				Major: 1,
+				Minor: 2,
+				Patch: 3,
+				Build: []string{"buildmetadata"},
+			}},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			assert.Equal(t, tc.expect, tc.input.AsLegacyRegistryV1Version())
+		})
+	}
+}
 
 func TestLegacyRegistryV1VersionRelease_Compare(t *testing.T) {
 	type testCase struct {
@@ -106,6 +212,12 @@ func TestLegacyRegistryV1VersionRelease_Compare(t *testing.T) {
 			v1:     "0.0.0-0+2",
 			v2:     "0.0.0-0+1",
 			expect: 1,
+		},
+		{
+			name:   "non-release build metadata is less than valid release build metadata",
+			v1:     "0.0.0-0+1.001",
+			v2:     "0.0.0-0+0",
+			expect: -1,
 		},
 	} {
 		t.Run(tc.name, func(t *testing.T) {


internal/operator-controller/bundleutil/bundle.go
```diff
@@ -2,33 +2,118 @@ package bundleutil
...
 				return nil, err
 			}
-			return &vers, nil
+			return vr, nil
 		}
 	}
 	return nil, fmt.Errorf("no package property found in bundle %q", b.Name)
 }
 
+func NewLegacyRegistryV1VersionRelease(vStr string) (*VersionRelease, error) {
```
pedjak: how about to move this and other code added below to a separate file/package, e.g. `internal/bundle` or `internal/bundle/versionrelease?

internal/operator-controller/bundleutil/bundle.go
```diff
@@ -2,33 +2,118 @@ package bundleutil
...
+		Release: rel,
+	}, nil
+}
+
+type VersionRelease struct {
+	Version bsemver.Version
+	Release Release
+}
+
+func (vr VersionRelease) Compare(other VersionRelease) int {
```
pedjak: how about to use a pointer receiver across all `VersionRelease` methods?

internal/operator-controller/catalogmetadata/compare/compare.go
```diff
@@ -1,25 +1,71 @@
...
+// version string or any range that matches the syntax defined in the Masterminds semver library.
+//
+// When the provided version range is a valid semver version that includes build metadata, then the
+// returned function will only match an identical version with the same build metadata.
+//
+// When the provided version range is a valid semver version that does NOT include build metadata,
+// then the returned function will match any version that matches the semver version, ignoring the
+// build metadata of matched versions.
+//
+// Otherwise, Masterminds semver constraints semantics are used to match versions.
```
pedjak: should not both libraries implement semver spec? would it be possible to use just single library?

internal/operator-controller/catalogmetadata/compare/compare.go
```diff
@@ -42,16 +88,16 @@ func ByDeprecationFunc(deprecation declcfg.Deprecation) func(a, b declcfg.Bundle
 	}
 }
 
-// compareErrors returns 0 if both errors are either nil or not nil
-// -1 if err1 is nil and err2 is not nil
-// +1 if err1 is not nil and err2 is nil
+// compareErrors returns 0 if both errors are either nil or not nil,
+// -1 if err1 is not nil and err2 is nil, and
+// +1 if err1 is nil and err2 is not nil
+// The semantic is that errors are "less than" non-errors.
```
pedjak, line 80: is this change really needed?

internal/shared/util/slices/slices.go
```diff
@@ -0,0 +1,9 @@
+package slices
+
+func Map[I any, O any](in []I, f func(I) O) []O {
```
pedjak, line 3: this function is available in https://github.com/samber/lo/blob/master/slice.go#L26 - how about to use it instead?

internal/operator-controller/catalogmetadata/compare/compare.go
joelanford: The main reason we have both is that they have different implementations of version ranges. OLMv0 used blang's version ranges for `skipRange` semantics, so we need to continue using that for back-compat reasons.

However, blang's semver range semantics have some footguns and are not all that ergonomic. So we chose Masterminds range semantics for `spec.source.catalog.version` in OLMv1.

internal/operator-controller/catalogmetadata/compare/compare.go
joelanford: I noticed that polarity was reversed between `compareErrors` and `VersionRelease.Compare`. This change aligns the polarity so that both are the expected ascending semantics. Then callers (e.g. `ByVersion`) use the same polarity when evaluating errors or their corresponding VersionRelease.

(I was tempted to change the polarity of `ByVersion` and `ByDeprecation` to be ascending semantics as well with _their_ callers reversing the arguments to get descending order, but that felt inappropriate for this PR).

internal/shared/util/slices/slices.go
joelanford: Not strongly opinionated. I do notice that that would be a new dependency to pull in to the project, so I could go either way.

Anyone else have opinions?

internal/operator-controller/bundleutil/bundle.go
joelanford: Seems reasonable. I'll take a look.

joelanford: I moved it to `internal/operator-controller/bundle/versionrelease.go`. `operator-controller` is the only component that uses this, so I kept it in that tree.

internal/operator-controller/catalogmetadata/compare/compare.go
pedjak: Thanks, for the explanation, perhaps you could add to the comment in the code?

internal/operator-controller/bundle/versionrelease.go
```diff
@@ -0,0 +1,89 @@
...
+	"errors"
+	"fmt"
+	"strings"
+
+	bsemver "github.com/blang/semver/v4"
+
+	slicesutil "github.com/operator-framework/operator-controller/internal/shared/util/slices"
+)
+
+func NewLegacyRegistryV1VersionRelease(vStr string) (*VersionRelease, error) {
```
pedjak, line 21: please add the docs.

internal/operator-controller/bundle/versionrelease.go
```diff
@@ -0,0 +1,89 @@
...
+		Release: rel,
+	}, nil
+}
+
+type VersionRelease struct {
+	Version bsemver.Version
+	Release Release
+}
+
+func (vr *VersionRelease) Compare(other VersionRelease) int {
```
pedjak, line 61: please document

internal/operator-controller/bundle/versionrelease.go
```diff
@@ -0,0 +1,89 @@
...
+		Minor: vr.Version.Minor,
+		Patch: vr.Version.Patch,
+		Pre:   vr.Version.Pre,
+		Build: slicesutil.Map(vr.Release, func(i bsemver.PRVersion) string { return i.String() }),
+	}
+}
+
+type Release []bsemver.PRVersion
+
+func (r Release) Compare(other Release) int {
```
pedjak, line 98:
```suggestion
func (r *Release) Compare(other Release) int {
```

internal/operator-controller/bundleutil/bundle.go
```diff
@@ -10,20 +10,26 @@ import (
 	"github.com/operator-framework/operator-registry/alpha/property"
 
 	ocv1 "github.com/operator-framework/operator-controller/api/v1"
+	"github.com/operator-framework/operator-controller/internal/operator-controller/bundle"
 )
 
-func GetVersion(b declcfg.Bundle) (*bsemver.Version, error) {
+func GetVersionAndRelease(b declcfg.Bundle) (*bundle.VersionRelease, error) {
```
pedjak, line 16: would it make sense to move this function to the new `bundle/versionsrelease.go` file?

internal/operator-controller/bundleutil/bundle_test.go
```diff
@@ -34,6 +34,14 @@ func TestGetVersion(t *testing.T) {
 			},
 			wantErr: true,
 		},
+		{
+			name: "invalid release",
```
pedjak: can we give here more details why the given release is invalid?

internal/operator-controller/catalogmetadata/compare/compare.go
pedjak: Perhaps instead of leaking that that the function uses Masterminds library under the hood, it would be more helpful to document actually what are all supported version range syntax?

internal/operator-controller/catalogmetadata/compare/compare.go
```diff
@@ -1,25 +1,71 @@
...
+		pre := slicesutil.Map(in.Pre, func(pr bsemver.PRVersion) string { return pr.String() })
+		mmVer := mmsemver.New(in.Major, in.Minor, in.Patch, strings.Join(pre, "."), strings.Join(in.Build, "."))
+		return constraint.Check(mmVer)
+	}, nil
+}
+
+// ByVersionAndRelease is a comparison function that compares bundles by
+// version and release. Bundles with lower versions/releases are
+// considered less than bundles with higher versions/releases.
+func ByVersionAndRelease(b1, b2 declcfg.Bundle) int {
```
pedjak, line 44: nit: how about that we implement `Compare` method on `declcfg.Bundle` instead?

internal/operator-controller/catalogmetadata/filter/bundle_predicates.go
```diff
@@ -1,29 +1,32 @@
...
 
 	"github.com/operator-framework/operator-registry/alpha/declcfg"
 
+	"github.com/operator-framework/operator-controller/internal/operator-controller/bundle"
 	"github.com/operator-framework/operator-controller/internal/operator-controller/bundleutil"
 	"github.com/operator-framework/operator-controller/internal/shared/util/filter"
 )
 
-func InMastermindsSemverRange(semverRange *mmsemver.Constraints) filter.Predicate[declcfg.Bundle] {
+func ExactVersionRelease(expect bundle.VersionRelease) filter.Predicate[declcfg.Bundle] {
```
pedjak, line 16: please document

internal/operator-controller/catalogmetadata/filter/bundle_predicates.go
```diff
@@ -1,29 +1,32 @@
...
-		// No error should occur here because the simple version was successfully parsed by blang
-		// We are unaware of any tests cases that would cause one to fail but not the other
-		// This will cause code coverage to drop for this line. We don't ignore the error because
-		// there might be that one extreme edge case that might cause one to fail but not the other
-		mVersion, err := mmsemver.NewVersion(bVersion.String())
+		return expect.Compare(*actual) == 0
+	}
+}
+
+func InSemverRange(versionRange bsemver.Range) filter.Predicate[declcfg.Bundle] {
```
pedjak, line 29: please document

camilamacedo86:
This one seems related to the bug that @grokspawn has been looking on.
I think it would be nice get his review here.

internal/operator-controller/catalogmetadata/compare/compare.go
grokspawn: Like [this](https://github.com/operator-framework/operator-controller/blob/main/api/v1/clusterextension_types.go#L219-L296)?

joelanford: Given this is an internal function, I'm not super-inclined to re-state the full description of what is supported. Technically, what this function supports is different than what our API supports, because our API rejects hyphen ranges that _are_ supported by Masterminds.

However, the intended use of this function is to parse that specific `Version` field from the API. Perhaps I could put _that_ in the comment so that we are talking about intended usage, but we don't actually document the full syntax, and then need to maintain it in two places.

internal/shared/util/slices/slices.go
joelanford: Hearing no other opinions, my vote is to leave as is. With this commit moving this function into a new package in the `shared/utils` space, hopefully future contributions will find this location to add new generic slice methods, and we can revisit the question again when there is more to be gained from pulling in a new dep.

internal/operator-controller/bundle/versionrelease.go
joelanford: `Release` is a slice, which is already represented as a pointer under the hood. In that case, I'm not sure it makes sense to use a pointer receiver for `Release` methods?

internal/operator-controller/bundleutil/bundle.go
joelanford: Not strongly opinionated here, but I do I kinda like `bundle/versionsrelease.go` as a pure implementation of the type without the baggage of the various kinds of things that we might source or derive a `VersionRelease` from.

internal/operator-controller/catalogmetadata/compare/compare.go
joelanford: The reason we don't have that right now is because `declcfg.Bundle` does not actually embed its own deprecation information. So we can't directly compare `declcfg.Bundle` along deprecation lines.

I suppose we could essentially combine `ByVersionAndRelease` and `ByDeprecationFunc` into a single comparison  function, but:
1. It still wouldn't be a method on `declcfg.Bundle`
2. It would still require extra `declcfg.Deprecation` metadata to do the comparison.

internal/operator-controller/catalogmetadata/compare/compare.go
joelanford: > Thanks, for the explanation, perhaps you could add to the comment in the code?

We already have that documented here: https://github.com/operator-framework/operator-controller/blob/85e8cbfa106594ad25a00ceb597dec1b8eae2e9e/internal/operator-controller/catalogmetadata/filter/successors.go#L54-L58

Do you think that is sufficient?

joelanford:
crd-diff is failing due to the change in the description of the `spec.source.catalog.version` field. Assuming folks agree to the updated API semantics, we either need to override the failure or update the crd-diff check to ignore the description change.

I'm personally inclined to specifically override so that future description changes continue to be flagged as potentially breaking so that those changes get extra scrutiny.

internal/operator-controller/bundle/versionrelease.go
```diff
@@ -0,0 +1,101 @@
...
```
pedjak, line 98: please document

internal/operator-controller/bundle/versionrelease.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+	if len(other) == 0 && len(r) > 0 {
+		return 1
+	}
+	a := bsemver.Version{Pre: r}
+	b := bsemver.Version{Pre: other}
+	return a.Compare(b)
+}
+
+func NewRelease(relStr string) (Release, error) {
```
pedjak, line 115: please document what are the valid inputs here?

pedjak (approved):
/lgtm a few more functions could be documented.

internal/shared/util/slices/slices.go
grokspawn: I'm of two minds on the subject.  The reference is to what looks like a personal library, which normally garners a 'no' from me.  But it has a mess of contributions, releases, and some kind of community, which relaxes that somewhat.  

I'd prefer that this package didn't reuse the same name as a standard library (https://pkg.go.dev/slices). 
But otherwise I'm fine with having a minimal utility library here.

pedjak: > But it has a mess of contributions, releases, and some kind of community, which relaxes that somewhat.

The library has 20.5k stars which is really a lot and is a very good signal.

helm/olmv1/base/operator-controller/crd/experimental/olm.operatorframework.io_clusterextensions.yaml
```diff
@@ -413,6 +413,14 @@ spec:
                           "0.6.0", which means "only install version 0.6.0 and never
                           upgrade from this version".
 
+                          For registry+v1 bundles that include release information in the build
+                          metadata field (e.g., "1.0.0+20230101"), you can pin to an exact version
+                          including the release by specifying the full version string with build
+                          metadata (e.g., "1.0.0+20230101"). This ensures an exact match of both
+                          the semver version and the release. If you specify a version without build
+                          metadata (e.g., "1.0.0"), it will match all bundles with that version
+                          regardless of their release information.
```
grokspawn: We'll need to update this once the bundle release version feature lands, since we will need to mention both approaches, including how one is preferred in future.  Right now the bundle release version approach is to use `[version]-[release]` to express the attribute combo in a string (though the source of truth is always the discrete version and release fields in the `olm.package` property). 

Perhaps this is better resolved in a doc where we can go into more depth?

pedjak (approved):
/lgtm

internal/operator-controller/bundle/versionrelease.go
```diff
@@ -0,0 +1,116 @@
...
+
+// NewLegacyRegistryV1VersionRelease parses a registry+v1 bundle version string and returns
+// a VersionRelease. For registry+v1 bundles, the build metadata field of the semver version
+// is treated as release information (a semver spec violation maintained for backward compatibility).
+// The returned VersionRelease has the build metadata extracted into the Release field, and the
+// Version field has its Build metadata cleared.
+func NewLegacyRegistryV1VersionRelease(vStr string) (*VersionRelease, error) {
+	vers, err := bsemver.Parse(vStr)
+	if err != nil {
+		return nil, err
```
joelanford, line 24: This is not back-compat. We should return a releaseless `VersionRelease` in this case.

internal/operator-controller/catalogmetadata/filter/successors_test.go
```diff
@@ -36,6 +36,7 @@ func TestSuccessorsPredicate(t *testing.T) {
 				{
```
camilamacedo86, line 36: @joelanford, I think we can cover a few more cases

- Parse a valid semver value with build metadata like 3.0.0-pre+2.beta.1.
- Check if we properly deal with invalid builds values (5.0.0+041).

internal/operator-controller/catalogmetadata/compare/compare_test.go
```diff
@@ -5,15 +5,59 @@ import (
...
 
+	bsemver "github.com/blang/semver/v4"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
 	"github.com/operator-framework/operator-registry/alpha/declcfg"
 	"github.com/operator-framework/operator-registry/alpha/property"
 
 	"github.com/operator-framework/operator-controller/internal/operator-controller/catalogmetadata/compare"
 )
```
camilamacedo86, line 17: @joelanford wdy about  use AI here to check if we got tests for all these? 
this logic is kinda complex and I feel maybe some cases missing?
Are we covering things like: 

- order tests (empty, number, text, mix, different length)
- pin == (same version but diff release)

wdyt? just to be safe we cover everything

internal/operator-controller/bundle/versionrelease.go
```diff
@@ -0,0 +1,129 @@
...
+
+func (vr *VersionRelease) AsLegacyRegistryV1Version() bsemver.Version {
+	return bsemver.Version{
+		Major: vr.Version.Major,
+		Minor: vr.Version.Minor,
+		Patch: vr.Version.Patch,
+		Pre:   vr.Version.Pre,
+		Build: slicesutil.Map(vr.Release, func(i bsemver.PRVersion) string { return i.String() }),
+	}
+}
```
camilamacedo86, line 84: @joelanford it might have a small edge case here.
What happens if the build metadata can’t be parsed as a Release?
I think if parsing failed and Release is empty, we’d end up losing the original build metadata. It would be nice have a test to cover this scenario either.

internal/shared/util/slices/slices.go
grokspawn: Agreed.  I'm wondering why it looks like "some guy", but the support/longevity/utility appears to be there.
From a user perspective, I might be more interested in what dependencies it might bring along, but I see no cause for alarm so far.

internal/operator-controller/catalogmetadata/compare/compare.go
grokspawn: Unfortunately, I've been fighting that concern in operator-registry space as well.  I'd _like_ for declcfg.Bundle to have the reusable comparators there, but the historical pattern is that both declcfg and SQLite graph contributions are ultimately converted to `Model` objects, where final type conversions occur, and validation considers the objects' full scope (including deprecations).

Right now the box feels like "all users must convert to Model objects" to me. 😞

internal/operator-controller/catalogmetadata/filter/successors_test.go
grokspawn: A bunch of those cases are already covered in the operator-registry lib which would be where all these fields are introduced into the FBC. 
(for e.g. [alpha/property/property_test.go](https://github.com/operator-framework/operator-registry/pull/1792/files#diff-75453ad868a9281aedc5245336f642f532601ad342330351b34222ea43bfe369R207-R230))
I don't think there's a lot of utility in duplicating those tests here.

internal/operator-controller/bundle/versionrelease.go
grokspawn: Maybe related: https://github.com/operator-framework/operator-controller/pull/2273#discussion_r2455827184

internal/operator-controller/catalogmetadata/filter/successors_test.go
camilamacedo86: But the tests that exist in the other project currently will not cover the same logic implemented here.
Ideally, we would have a shared library that provides this functionality so both projects could consume it directly.
However, since the implementation currently lives here and the checks are performed at runtime in this project, we should still have test coverage here. ( and there ; not centralized )

In my opinion, this is important because any change made in this code could break those runtime scenarios, and the tests in the other project would not be able to catch such regressions.

internal/operator-controller/bundle/versionrelease.go
joelanford: Done. If we can't even parse the semver version, we should (and do) return that error. If below, we failed to parse the build metadata as a release, then we return a releaseless `VersionRelease` where the version's build metadata is left intact.

internal/operator-controller/catalogmetadata/compare/compare_test.go
joelanford: I found two more cases to cover:
1. An invalid range returns an error
2. A range that does not include the version returns false.

Adding these.

internal/operator-controller/catalogmetadata/filter/successors_test.go
joelanford: I have all of these cases implemented in `internal/operator-controller/bundle/versionrelease_test.go`.

internal/operator-controller/bundle/versionrelease.go
joelanford: Yeah, I caught that as well. That code has been updated to return a BundleVersion where `Version` contains `version+buildmetadata` and no `Release`

I've also updated `AsLegacyRegistryV1Version` to make sure it correctly handles a "releaseless" BundleVersion.

internal/operator-controller/catalogmetadata/compare/compare.go
```diff
@@ -42,16 +74,16 @@ func ByDeprecationFunc(deprecation declcfg.Deprecation) func(a, b declcfg.Bundle
...
-// -1 if err1 is nil and err2 is not nil
-// +1 if err1 is not nil and err2 is nil
+// compareErrors returns 0 if both errors are either nil or not nil,
+// -1 if err1 is not nil and err2 is nil, and
+// +1 if err1 is nil and err2 is not nil
+// The semantic is that errors are "less than" non-errors.
 func compareErrors(err1 error, err2 error) int {
 	if err1 != nil && err2 == nil {
-		return 1
+		return -1
```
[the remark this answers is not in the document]
joelanford: The overall polarity of this function was flipped in this PR to align with other changes in this PR and with general expectations of what comparison functions produce (-1 means the first parameter is "less than" the second parameter).

Everything in the comment aligns with the updated implementation.

PR #2274, opened 2025-10-16
🌱 Add AGENTS.md to assist AI agents interacting with this repository

joelanford:
# Description

In order for maintainers and external developers to have a tailored experience using AI development tools, it is helpful to include an `AGENTS.md` file in the repo to guide AI agents in their tasks.

I used Claude Code to generate the `AGENTS.md` file in this PR.

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,346 @@
+# AGENTS.md - AI Coding Assistant Briefing
+
+This document serves as a comprehensive briefing for AI coding assistants working with the operator-controller
+repository. It covers the WHAT, WHY, and HOW of contributing to this codebase.
+
+---
+
+## Architecture Overview
+
+operator-controller is the central component of Operator Lifecycle Manager (OLM) v1, extending Kubernetes with APIs to
+install and manage cluster extensions. The project follows a microservices architecture with two main binaries:
+
+**operator-controller** 
+ - manages `ClusterExtension` and `ClusterExtensionRevision` CRDs
+ - resolves bundles from configured source
+ - unpacks bundles and renders manifests from them
+ - applies manifests with phase-based rollouts
+ - monitors extension lifecycle
+
+**catalogd**
+ - manages user-defined `ClusterCatalog` resources to make references catalog metadata available to the cluster.
+ - unpacks and serves operator catalog content via HTTP.
+ - serves catalog metadata to clients in the cluster that need to use or present information about the contents of the
+   catalog. For example, operator-controller queries catalogd for available bundles.
+
+---
+
+## Tech Stack
+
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -0,0 +1,251 @@
+# AGENTS.md
```
camilamacedo86: Hi @joelanford 

Seems that Copilot accept AGENTS.md -> https://code.visualstudio.com/docs/copilot/customization/custom-instructions

But will that be understood by CLAUD and ANY other possible ones as well?

OH seems like they agreed upon: https://agents.md/

That is amazing !!!! 

I researched that 2 months ago, and I could not find a generic/standard one. Things are changing too fast 🚀

AGENTS.md
```diff
@@ -0,0 +1,251 @@
...
+
+This file provides instructions for autonomous agents working in the operator-controller repository.
+
+## Repository Context
```
camilamacedo86: But before we add it, what info here is valid only for upstream?
If we add this, will all be valid for downstream? I do not think so,
So, how will we have information for upstream and another for downstream?

@joelanford @tmshort have you ideas?
What about we ignore those files in the sync/bumper?
Do we want to have the same for downstream? Could we have a version there that is not removed by the bumper as well?

pedjak:
> I used Claude Code to generate the AGENTS.md file in this PR.

@joelanford out of curiosity, how did you instruct Cloude to generate this file?

AGENTS.md
tmshort: We can delete it downstream if we want to via the bumper

AGENTS.md
```diff
@@ -0,0 +1,251 @@
...
+
+This is **operator-controller**, part of OLM v1. Two main Go binaries: `operator-controller` and `catalogd`. Kubernetes controllers using controller-runtime framework. Focus on simplicity, security, predictability.
```
trgeiger: Does it want these weird fragmentary sentences? I would think we should make this more comprehensible, like "This repository contains operator-controller, a core component of the Operator Lifecycle Manager (OLM) version 1. It is composed of 2 Go-language binaries, operator-controller and catalogd. These binaries run Kubernetes controllers using the controller-runtime framework. Their operation is focused on simplicity, security, and predictability."

AGENTS.md
tmshort: But, it's still appropriate for downstream, the content doesn't really change.

tmshort:
We should try to revive this PR...

rashmigottipati (approved):
/lgtm

PR #2278, opened 2025-10-20
:seedling: Add registry+v1 bundle config unmarshal and validation layer

perdasilva:
# Description

### Changes
- Adds an `UnmarshalConfig` function to the (registry+v1) bundle package that unmarshals a json/yaml byte slice into a registry+v1 configuration struct and subsequently validates it
- Refactors the RegistryV1ManifestProvider to use the new function and removes the `getWatchNamespace` function

For this PR we do manual registry+v1 bundle config schema validation as this is sufficient. Once the configuration surface expands (e.g. by adding SubscriptionConfig support, or others), and/or we add Helm bundle support, we'd probably want to move to a JSONSchema-based approach.

#### Motivation

As things were, we'd accept bundle configuration containing `watchNamespace` for any bundle. In keeping with the mental model that each bundle would provide its own configuration schema, this shouldn't be the case. E.g. There are bundles that won't have any configuration like bundles that only support AllNamespaces mode. Therefore, we needed an additional layer of validation that checked the provided configuration against what is provided by the bundle.

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/operator-controller/rukpak/bundle/config.go
+++ b/internal/operator-controller/rukpak/bundle/config.go
@@ -14,7 +14,7 @@ import (
 )
 
 type Config struct {
-	WatchNamespace string `json:"watchNamespace"`
+	WatchNamespace *string `json:"watchNamespace"`
 }
 
 // UnmarshallConfig returns a deserialized and validated *bundle.Config based on bytes and validated
@@ -23,12 +23,9 @@ type Config struct {
 // - bytes is not a valid YAML/JSON object
 // - bytes is a valid YAML/JSON object but does not follow the registry+v1 schema
 // if bytes is nil a nil bundle.Config is returned
-func UnmarshallConfig(bytes []byte, rv1 *RegistryV1, installNamespace string) (*Config, error) {
+func UnmarshallConfig(bytes []byte, rv1 RegistryV1, installNamespace string) (*Config, error) {
 	if bytes == nil {
 		return nil, nil
-	}
-	if rv1 == nil {
-		return nil, errors.New("bundle is nil")
 	}
 
 	bundleConfig := &Config{}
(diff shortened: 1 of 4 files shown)


internal/operator-controller/rukpak/bundle/config.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+func validateConfig(config *Config, rv1 *RegistryV1, installNamespace string) error {
+	// no config, no problem
+	if config == nil {
+		return nil
+	}
+
+	// collect bundle install modes
+	installModeSet := sets.New(rv1.CSV.Spec.InstallModes...)
```
pedjak: `rv1` is passed into this function only to create this set of install modes. Hence, we can pass to the function these install modes directly, e.g.

```golang
func validateConfig(config *Config, rv1 *RegistryV1, installNamespace string, installModes sets.Set[v1alpha1.InstallMode]) error
```

nit: also, perhaps would be more readable if we implement `validateConfig` as receiver function on `Config`:

```golang
function (c *Config) ValidateFor(installNamespaces string, installModes sets.Set[v1alpha1.InstallMode]) error
```

internal/operator-controller/rukpak/bundle/config.go
```diff
@@ -0,0 +1,108 @@
...
+
+	// only accept a non-empty value for watchNamespace if the bundle configuration accepts the watchNamespace config
+	if config.WatchNamespace != "" && !hasWatchNamespaceAsConfig(installModeSet) {
+		return errors.New(`unknown field "watchNamespace"`)
```
pedjak, line 55: given that `WatchNamespace` is not empty, I believe that the error should not be `unknown field...`?

internal/operator-controller/rukpak/bundle/config.go
```diff
@@ -0,0 +1,108 @@
...
+
+	if config.WatchNamespace != installNamespace &&
+		!installModeSet.Has(v1alpha1.InstallMode{Type: v1alpha1.InstallModeTypeSingleNamespace, Supported: true}) {
+		return fmt.Errorf("invalid 'watchNamespace' %q: must be install namespace (%s)", config.WatchNamespace, installNamespace)
+	}
+
+	return nil
+}
+
+func hasWatchNamespaceAsConfig(bundleInstallModeSet sets.Set[v1alpha1.InstallMode]) bool {
```
pedjak: the function is used once only - thus for the readability it would be ok to put it inline in `validate()` function. BTW, the function name mentions watchnamespace, but it does not appear in the arguments - perhaps the function name should be rephrased to reflect what it does?

internal/operator-controller/rukpak/bundle/config.go
```diff
@@ -0,0 +1,108 @@
...
+// against rv1 and the desired install namespaces. It will error if:
+// - rv is nil
+// - bytes is not a valid YAML/JSON object
+// - bytes is a valid YAML/JSON object but does not follow the registry+v1 schema
+// if bytes is nil a nil bundle.Config is returned
+func UnmarshallConfig(bytes []byte, rv1 *RegistryV1, installNamespace string) (*Config, error) {
+	if bytes == nil {
+		return nil, nil
+	}
+	if rv1 == nil {
```
pedjak: if `nil` is not allowed, why do we not just `RegistryV1` instead a pointer to it?

internal/operator-controller/rukpak/bundle/config.go
```diff
@@ -0,0 +1,108 @@
...
+		return nil, errors.New("bundle is nil")
```
pedjak: should not be `registry must not be nil`?

internal/operator-controller/rukpak/bundle/config.go
perdasilva: This is the case now. But, not necessarily in the future. We want to validate based on what's in the bundle. At the moment the only pertinent thing is the install mode. I think as an API, taking in the bundle makes more sense.

pedjak: > I think as an API, taking in the bundle makes more sense.

This is a private function, so API can change whenever we see a need.

internal/operator-controller/rukpak/bundle/config.go
perdasilva: If a bundle does not support the watchNamespace configuration, we should treat it as an unknown field. There's a bug here because we cannot distinguish between not set and set to "". I'm reworking everything s.t. watchNamespace is a *string to add that distinction. We also need the case for when watchNamespace is *required* but its not set.

internal/operator-controller/rukpak/bundle/config.go
perdasilva: The question the function answers is whether the bundle supports the watchNamespace configuration - this is driven off the install mode support. If I don't need it again for the "required" case I'll inline it and add the appropriate comments

internal/operator-controller/rukpak/bundle/config.go
perdasilva: oh sorry - I misread I thought this comment was under the Unmarshal function. I'll fix that up!

perdasilva: But then I wouldn't make it package public though...for the reasons I sited in the first response. I think as an API taking the bundle as a parameter is the right thing

perdasilva: Done

internal/operator-controller/rukpak/bundle/config.go
perdasilva: I've refactored to `isWatchNamespaceConfigSupported`. I prefer to have these more complex clauses aliased under a more understandable function. So, I've left it in.

internal/operator-controller/rukpak/bundle/config.go
perdasilva: I think its just a tick from my side to avoid passing structs by value. The if was just there for defensive purposes (I wouldn't expect a nil to get there). I've made rv1 a concrete parameter now.

internal/operator-controller/rukpak/bundle/config.go
perdasilva: I've removed this clause

thetechnick (approved):

PR #2281, opened 2025-10-21
:sparkles: Propagate ClusterExtensionRevision conditions to ClusterExtension status

perdasilva:
# Description

* `ClusterExtension`'s `Available` and `Progressing` conditions are equal to the counterparts on the latest installed active revision
* When a new revision is rolling out, mirror its `Progressing` condition to its counterpart on `ClusterExtension` to improve visibility and to signal that a transition is in progress
* Adds `.status.activeRevisions` field to the `ClusterExtension` to track the active revisions and mirrors `Available` and `Progressing` conditions for those not marked yet as installed

This changes opens up a possibility for an observer to get insights on `ClusterExtension` update process:
* When Reason on `Progressing` condition flips to `RollingOut`, an update is in progress
* More then one revision could be listed under `.status.activeRevisions` during upgrade (installed + those in progress)
* `Available` and `Progressing` conditions of an active, but not yet available revision, are listed under `.status.activeRevisions.conditions` so that can be easier to understand update issues

The new behavior is marked as experimental behind Boxcutter feature gate.

## Reviewer Checklist

- [x] API Go Documentation


internal/operator-controller/controllers/clusterextensionrevision_controller.go
```diff
@@ -126,36 +126,51 @@ func (c *ClusterExtensionRevisionReconciler) reconcile(ctx context.Context, rev
...
-			Status:             metav1.ConditionFalse,
-			Reason:             ocv1.ClusterExtensionRevisionReasonReconcileFailure,
-			Message:            err.Error(),
-			ObservedGeneration: rev.Generation,
-		})
 		return ctrl.Result{}, fmt.Errorf("error ensuring teardown finalizer: %v", err)
 	}
 
-	if err := c.establishWatch(ctx, rev, revision); err != nil {
+	inRollout := meta.FindStatusCondition(rev.Status.Conditions, ocv1.ClusterExtensionRevisionTypeAvailable) == nil
```
perdasilva: I worry about looking at the status to drive the reconciliation logic here. We should probably look at the state of the objects themselves to make this determination, no? Or, is this just temporary and in lieu of the new boxcutter PR that should surface this state more easily?

internal/operator-controller/controllers/clusterextensionrevision_controller.go
```diff
@@ -126,36 +126,51 @@ func (c *ClusterExtensionRevisionReconciler) reconcile(ctx context.Context, rev
...
+	if inRollout {
+		if err := c.establishWatch(ctx, rev, revision); err != nil {
```
perdasilva: should establishWatch just be idempotent and we call it whether were rolling out or not?

api/v1/clusterextensionrevision_types.go
```diff
@@ -164,6 +171,46 @@ type ClusterExtensionRevision struct {
 	Status ClusterExtensionRevisionStatus `json:"status,omitempty"`
 }
 
+func (cer *ClusterExtensionRevision) MarkAsProgressing(reason, message string) {
```
perdasilva: should we export these?

internal/operator-controller/controllers/clusterextensionrevision_controller.go
```diff
@@ -165,27 +165,41 @@ func (c *ClusterExtensionRevisionReconciler) reconcile(ctx context.Context, rev
...
-			meta.SetStatusCondition(&rev.Status.Conditions, metav1.Condition{
-				Type:               ocv1.ClusterExtensionRevisionTypeAvailable,
-				Status:             metav1.ConditionFalse,
-				Reason:             ocv1.ClusterExtensionRevisionReasonPhaseValidationError,
-				Message:            fmt.Sprintf("phase %d validation error: %s", i, verr),
-				ObservedGeneration: rev.Generation,
-			})
+			if inRollout {
+				// given that we retry, we are going to keep Progressing condition True
+				rev.MarkAsProgressing(ocv1.ClusterExtensionRevisionReasonPhaseValidationError, fmt.Sprintf("phase %d validation error: %s", i, verr))
```
perdasilva: nit: I wonder if we should just use a "Retrying" reason rather than PhaseValidationError/RevisionValidationFailure/etc.".

api/v1/clusterextensionrevision_types.go
perdasilva: nevermind - I misread it

internal/operator-controller/controllers/clusterextensionrevision_controller.go
```diff
@@ -140,7 +140,7 @@ func (c *ClusterExtensionRevisionReconciler) reconcile(ctx context.Context, rev
 			rev.MarkAsProgressing(ocv1.ClusterExtensionRevisionReasonReconcileFailure, werr.Error())
 			return ctrl.Result{}, werr
 		}
-		rev.MarkAsProgressing(ocv1.ClusterExtensionRevisionReasonRolloutInProgress, fmt.Sprintf("Revision %s is being rolled out.", revVersion))
+		// rev.MarkAsProgressing(ocv1.ClusterExtensionRevisionReasonRolloutInProgress, fmt.Sprintf("Revision %s is being rolled out.", revVersion))
```
pedjak: @perdasilva IMHO, this should stay - progression condition is not necessarily bound  to boxcutter notion of progression. Reconciliation is more than just delegating things to the boxcutter and looking what it has to say about the progress. 

If we are able to set watches on object types present in the revision, we should mark that we are in progress. 
This condition is not persisted immediately - it can be ofcourse overwritten further down, if reported by boxcutter differently. However, if some transient error happens, we are going to mark our revision as progressing, which is a good UX.

internal/operator-controller/controllers/clusterextensionrevision_controller.go
```diff
@@ -207,8 +208,11 @@ func (c *ClusterExtensionRevisionReconciler) reconcile(ctx context.Context, rev
 
 		var collidingObjs []string
 		for _, ores := range pres.GetObjects() {
-			if ores.Action() == machinery.ActionCollision {
+			switch ores.Action() {
+			case machinery.ActionCollision:
 				collidingObjs = append(collidingObjs, ores.String())
+			case machinery.ActionCreated, machinery.ActionUpdated, machinery.ActionProgressed, machinery.ActionRecovered:
+				mutations++
```
pedjak: what is the purpose of this change? We count mutations, but not use that info later on.

internal/operator-controller/controllers/clusterextensionrevision_controller.go
```diff
@@ -225,13 +229,14 @@ func (c *ClusterExtensionRevisionReconciler) reconcile(ctx context.Context, rev
...
 	}
 
-	if !rres.InTransistion() {
-		// we have rolled out all objects in all phases, not interested in probes here
-		rev.MarkAsNotProgressing(ocv1.ClusterExtensionRevisionReasonRolledOut, fmt.Sprintf("Revision %s is rolled out.", revVersion))
-	}
+	//if !rres.InTransistion() {
+	//	// we have rolled out all objects in all phases, not interested in probes here
+	//	rev.MarkAsNotProgressing(ocv1.ClusterExtensionRevisionReasonRolledOut, fmt.Sprintf("Revision %s is rolled out.", revVersion))
+	//}
```
pedjak: Flipping `Progressing` from `True` to `False` at rolling out should not be bound to the state of availability. Simply, we should signal that all objects are applied to the cluster. Availability has its own lifecycle.

internal/operator-controller/controllers/clusterextensionrevision_controller.go
```diff
@@ -258,6 +263,7 @@ func (c *ClusterExtensionRevisionReconciler) reconcile(ctx context.Context, rev
 			ObservedGeneration: rev.Generation,
 		})
 	} else {
+		rev.MarkAsProgressing(ocv1.ClusterExtensionRevisionReasonRolloutInProgress, fmt.Sprintf("Revision %s is being rolled out.", revVersion))
```
pedjak: This is going to cause that everytime when `Availability` flips to `False` - we are going to set `Progressing` to `True` and we do not want that. Once all object are applied to cluster and we set `Progressing` to `False`, we should not return to `True`.

internal/operator-controller/controllers/clusterextensionrevision_controller.go
perdasilva: sorry - that was left over from some experimentation. I'll remove it.

internal/operator-controller/controllers/clusterextensionrevision_controller.go
perdasilva: Yup - all good. I've dropped my commit.

perdasilva: I think we need to establish a difference between progression probes and availability probes though.

internal/operator-controller/controllers/clusterextensionrevision_controller.go
perdasilva: I agree. It's just that right now we are mixing up availability and progression probes.

internal/operator-controller/controllers/clusterextensionrevision_controller.go
perdasilva: Yeah, that's fine. Sorry about this, it was left over from experimentation. I was doing this late afternoon and my brain was already pretty dead.

internal/operator-controller/controllers/boxcutter_reconcile_steps.go
```diff
@@ -54,12 +55,14 @@ func (d *BoxcutterRevisionStatesGetter) GetRevisionStates(ctx context.Context, e
...
 		}
 
-		// TODO: the setting of these annotations (happens in boxcutter applier when we pass in "revisionAnnotations")
+		// TODO: the setting of these annotations (happens in boxcutter applier when we pass in "storageLabels")
 		//   is fairly decoupled from this code where we get the annotations back out. We may want to co-locate
 		//   the set/get logic a bit better to make it more maintainable and less likely to get out of sync.
 		rm := &RevisionMetadata{
-			Package: rev.Annotations[labels.PackageNameKey],
-			Image:   rev.Annotations[labels.BundleReferenceKey],
+			RevName:    rev.Name,
```
perdasilva: Maybe the complete name would be better?

```suggestion
			RevisionName:    rev.Name,
```

internal/operator-controller/controllers/boxcutter_reconcile_steps.go
```diff
@@ -54,12 +55,14 @@ func (d *BoxcutterRevisionStatesGetter) GetRevisionStates(ctx context.Context, e
 			continue
 		}
 
-		// TODO: the setting of these annotations (happens in boxcutter applier when we pass in "revisionAnnotations")
+		// TODO: the setting of these annotations (happens in boxcutter applier when we pass in "storageLabels")
 		//   is fairly decoupled from this code where we get the annotations back out. We may want to co-locate
 		//   the set/get logic a bit better to make it more maintainable and less likely to get out of sync.
 		rm := &RevisionMetadata{
-			Package: rev.Annotations[labels.PackageNameKey],
```
perdasilva, line 61: From yesterday's discussion, should we stick to annotations?

internal/operator-controller/controllers/boxcutter_reconcile_steps.go
```diff
@@ -89,3 +92,55 @@ func MigrateStorage(m StorageMigrator) ReconcileStepFunc {
...
+			// report the error via the Progressing condition.
+			setStatusProgressing(ext, wrapErrorWithResolutionInfo(state.resolvedRevisionMetadata.BundleMetadata, err))
+			return nil, err
+		}
+
+		// Mirror Available/Progressing conditions from the installed revision
+		if i := state.revisionStates.Installed; i != nil {
+			for _, cndType := range []string{ocv1.ClusterExtensionRevisionTypeAvailable, ocv1.ClusterExtensionRevisionTypeProgressing} {
+				if cnd := apimeta.FindStatusCondition(i.Conditions, cndType); cnd != nil {
+					cnd.ObservedGeneration = ext.GetGeneration()
```
perdasilva, line 122: Not for this PR, but I wonder if we should carry the CE generation as an annotation on the CER...

api/v1/clusterextension_types.go
```diff
@@ -523,6 +541,7 @@ type ClusterExtensionInstallStatus struct {
 // +kubebuilder:printcolumn:name="Installed Bundle",type=string,JSONPath=`.status.install.bundle.name`
 // +kubebuilder:printcolumn:name=Version,type=string,JSONPath=`.status.install.bundle.version`
 // +kubebuilder:printcolumn:name="Installed",type=string,JSONPath=`.status.conditions[?(@.type=='Installed')].status`
+// +kubebuilder:printcolumn:name="Available",type=string,JSONPath=`.status.conditions[?(@.type=='Available')].status`
```
perdasilva: might be worth calling this out in the description as well

perdasilva: Actually,  this is affecting the standard definition as well. I wonder if we can wrap it with some opcon:experimental tags to confine it to the experimental set...? If not, maybe we should remove it until we can...=/

test/e2e/cluster_extension_revision_test.go
```diff
@@ -130,6 +142,14 @@ func TestClusterExtensionRevision(t *testing.T) {
 		require.Equal(ct, ocv1.ClusterExtensionRevisionReasonProbesSucceeded, cond.Reason)
 	}, pollDuration, pollInterval)
 
+	t.Log("By propagating Available:True to ClusterExtension")
+	require.EventuallyWithT(t, func(ct *assert.CollectT) {
+		require.NoError(ct, c.Get(context.Background(), types.NamespacedName{Name: clusterExtension.Name}, clusterExtension))
+		cond := apimeta.FindStatusCondition(clusterExtensionRevision.Status.Conditions, ocv1.ClusterExtensionRevisionTypeAvailable)
```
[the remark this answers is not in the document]
perdasilva: I agree with the robot here - I think we want to check the CE

test/e2e/cluster_extension_revision_test.go
```diff
@@ -112,6 +116,14 @@ func TestClusterExtensionRevision(t *testing.T) {
 		require.Equal(ct, ocv1.ClusterExtensionRevisionReasonProbeFailure, cond.Reason)
 	}, pollDuration, pollInterval)
 
+	t.Log("By propagating Available:False to ClusterExtension")
+	require.EventuallyWithT(t, func(ct *assert.CollectT) {
+		require.NoError(ct, c.Get(context.Background(), types.NamespacedName{Name: clusterExtension.Name}, clusterExtension))
+		cond := apimeta.FindStatusCondition(clusterExtensionRevision.Status.Conditions, ocv1.ClusterExtensionRevisionTypeAvailable)
```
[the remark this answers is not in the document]
perdasilva: I think the robot is correct here

test/e2e/cluster_extension_revision_test.go
```diff
@@ -37,8 +37,9 @@ func TestClusterExtensionRevision(t *testing.T) {
 		Source: ocv1.SourceConfig{
 			SourceType: "Catalog",
 			Catalog: &ocv1.CatalogFilter{
-				PackageName: "test",
-				Version:     "1.0.1",
+				PackageName:             "test",
+				Version:                 "1.0.1",
+				UpgradeConstraintPolicy: ocv1.UpgradeConstraintPolicySelfCertified,
```
perdasilva, line 43: How come we need this?

perdasilva: Because there's no upgrade edge between 1.0.1 and 1.0.2? should we call it out in a comment?

api/v1/clusterextension_types.go
pedjak: printer columns are not not a part of schema declared: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#additional-printer-columns so we would need to improve `crd-generator` to support that.

Given that the focus on the PR is not adding this column at I am going to remove it.

internal/operator-controller/controllers/boxcutter_reconcile_steps.go
```diff
@@ -89,3 +92,55 @@ func MigrateStorage(m StorageMigrator) ReconcileStepFunc {
...
+					apimeta.SetStatusCondition(&ext.Status.Conditions, *cnd)
```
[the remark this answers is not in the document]
pedjak: no need for that because we are just mirroring stuff.

internal/operator-controller/controllers/boxcutter_reconcile_steps.go
```diff
@@ -89,3 +92,55 @@ func MigrateStorage(m StorageMigrator) ReconcileStepFunc {
...
+			}
+			ext.Status.ActiveRevisions = []ocv1.RevisionStatus{{Name: i.RevName}}
+		}
+		for idx, r := range state.revisionStates.RollingOut {
+			rs := ocv1.RevisionStatus{Name: r.RevName, Conditions: r.Conditions}
+			// Mirror Progressing condition from the latest active revision
+			if idx == len(state.revisionStates.RollingOut)-1 {
+				if pcnd := apimeta.FindStatusCondition(r.Conditions, ocv1.ClusterExtensionRevisionTypeProgressing); pcnd != nil {
+					pcnd.ObservedGeneration = ext.GetGeneration()
+					apimeta.SetStatusCondition(&ext.Status.Conditions, *pcnd)
```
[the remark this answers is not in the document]
pedjak: same as above.

michaelryanpeter:
LGTM

PR #2283, opened 2025-10-22
🐛 OPRUN-4217: OwnNamespace default handling

perdasilva:
# Description

Currently, if a bundle only supports OwnNamespace the `watchNamespace` configuration will no be present for that bundle and the bundle will be automatically installed in OwnNamespace mode (by setting the targetNamespaces to the install namespace). While this is intuitive and makes life easier for the user, it can can lead to a break in default behavior if the package decided to add AllNamespaces support. For instance:

v1 only supports OwnNamespace install mode. The ClusterExtension has no configuration so the bundle is installed in OwnNamespace mode.

v1.1 adds AllNamespaces support and the ClusterExtension detects the upgrade and initiates it. The default is now AllNamespaces mode. So, the operator is now installed in AllNamespace mode without knowledge or acceptance by the user.

To address this issue, we modify the defaulting and configuration behavior for this case by requiring `watchNamespace` to be set and removing the defaulting behavior, i.e. making it undefined. I.e. only bundles with AllNamespaces mode can be installed without `watchNamespace` configuration. This means that when v2 comes out with AllNamespaces support, the configuration will already be present in the ClusterExtension and v2 will be installed in the same mode. The user can then remove the configuration if they wish to change install modes.

So, in this PR, we:
- Add some additional unit tests around watchNamespaces being null/empty (this is not exactly related but was identified while writing this PR)
- Modify the configuration validation to ensure that watchNamespaces is required when OwnNamespace is the only supported install mode (we include MultiNamespaces in the tests for completeness but this install mode will never be supported)
- Modify the defaulting behavior in the reg+v1 renderer to ensure no defaults are set for bundles that don't support AllNamespaces install mode
- Add `ExtensionConfigBytes()` method to `ClusterExtension` to easily grab user configuration
- Update `UnmarshalConfig` to treat `nil` as `{}` to force schema validation (catch missing required values, etc.)
- Update `RegistryV1ManifestProvider` tests to ensure bundle configuration is being properly validated
- Add single-namespace- and own-namespace-operator fixtures and update e2e tests to check for bad configuration as well as good

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/testdata/images/bundles/own-namespace-operator/v1.0.0/manifests/ownnamespaceoperator.clusterserviceversion.yaml
+++ b/testdata/images/bundles/own-namespace-operator/v1.0.0/manifests/ownnamespaceoperator.clusterserviceversion.yaml
@@ -27,123 +27,123 @@ spec:
   apiservicedefinitions: {}
   customresourcedefinitions:
   owned:
-  - description: A dummy resource for an operator that only supports own namespace install mode
-    displayName: OwnNamespace
-    kind: OwnNamespace
-    name: ownnamespaces.olm.operatorframework.io
-    version: v1
+    - description: A dummy resource for an operator that only supports own namespace install mode
+      displayName: OwnNamespace
+      kind: OwnNamespace
+      name: ownnamespaces.olm.operatorframework.io
+      version: v1
   description: OLM OwnNamespace Testing Operator
   displayName: test-operator
   icon:
-  - base64data: ""
-    mediatype: ""
+    - base64data: ""
+      mediatype: ""
   install:
     spec:
       deployments:
-      - label:
-          app.kubernetes.io/component: controller
-          app.kubernetes.io/name: own-namespace-operator
-          app.kubernetes.io/version: 1.0.0
-        name: own-namespace-operator
-        spec:
-          replicas: 1
-          selector:
-            matchLabels:
-              app: ownnamespacetest
-          template:
-            metadata:
-              labels:
+        - label:
+            app.kubernetes.io/component: controller
+            app.kubernetes.io/name: own-namespace-operator
+            app.kubernetes.io/version: 1.0.0
+          name: own-namespace-operator
+          spec:
+            replicas: 1
+            selector:
+              matchLabels:
                 app: ownnamespacetest
-            spec:
-              terminationGracePeriodSeconds: 0
-              containers:
-              - name: busybox
-                image: busybox:1.36
-                command:
-                - 'sleep'
-                - '1000'
-                securityContext:
-                  runAsUser: 1000
-                  runAsNonRoot: true
-              serviceAccountName: simple-bundle-manager
+            template:
+              metadata:
+                labels:
+                  app: ownnamespacetest
+              spec:
+                terminationGracePeriodSeconds: 0
+                containers:
+                  - name: busybox
+                    image: busybox:1.36
+                    command:
+                      - 'sleep'
+                      - '1000'
+                    securityContext:
+                      runAsUser: 1000
+                      runAsNonRoot: true
+                serviceAccountName: simple-bundle-manager
       clusterPermissions:
-      - rules:
-        - apiGroups:
-          - authentication.k8s.io
-          resources:
-          - tokenreviews
-          verbs:
-          - create
-        - apiGroups:
-          - authorization.k8s.io
-          resources:
-          - subjectaccessreviews
-          verbs:
-          - create
-        serviceAccountName: simple-bundle-manager
+        - rules:
+            - apiGroups:
+                - authentication.k8s.io
+              resources:
+                - tokenreviews
+              verbs:
+                - create
+            - apiGroups:
+                - authorization.k8s.io
+              resources:
+                - subjectaccessreviews
+              verbs:
+                - create
+          serviceAccountName: simple-bundle-manager
       permissions:
-      - rules:
-        - apiGroups:
-          - ""
-          resources:
-          - configmaps
-          - serviceaccounts
-          verbs:
-          - get
-          - list
-          - watch
-          - create
-          - update
-          - patch
-          - delete
-        - apiGroups:
-          - networking.k8s.io
-          resources:
-          - networkpolicies
-          verbs:
-          - get
-          - list
-          - create
-          - update
-          - delete
-        - apiGroups:
-          - coordination.k8s.io
-          resources:
-          - leases
-          verbs:
-          - get
-          - list
-          - watch
-          - create
-          - update
-          - patch
-          - delete
-        - apiGroups:
-          - ""
-          resources:
-          - events
-          verbs:
-          - create
-          - patch
-        serviceAccountName: simple-bundle-manager
+        - rules:
+            - apiGroups:
+                - ""
+              resources:
+                - configmaps
+                - serviceaccounts
+              verbs:
+                - get
+                - list
+                - watch
+                - create
+                - update
+                - patch
+                - delete
+            - apiGroups:
+                - networking.k8s.io
+              resources:
+                - networkpolicies
+              verbs:
+                - get
+                - list
+                - create
+                - update
+                - delete
+            - apiGroups:
+                - coordination.k8s.io
+              resources:
+                - leases
+              verbs:
+                - get
+                - list
+                - watch
+                - create
+                - update
+                - patch
+                - delete
+            - apiGroups:
+                - ""
+              resources:
+                - events
+              verbs:
+                - create
+                - patch
+          serviceAccountName: simple-bundle-manager
     strategy: deployment
   installModes:
-  - supported: true
-    type: OwnNamespace
-  - supported: false
-    type: SingleNamespace
-  - supported: false
-    type: MultiNamespace
-  - supported: false
-    type: AllNamespaces
+    - supported: true
+      type: OwnNamespace
+    - supported: false
+      type: SingleNamespace
+    - supported: false
+      type: MultiNamespace
+    - supported: false
+      type: AllNamespaces
   keywords:
-  - registry
+    - registry
   links:
-  - name: simple-bundle
-    url: https://simple-bundle.domain
+    - name: simple-bundle
+      url: https://simple-bundle.domain
   maintainers:
-  - email: main#simple-bundle.domain
-    name: Simple Bundle
+    - email: main#simple-bundle.domain
+      name: Simple Bundle
   maturity: beta
   provider:
     name: Simple Bundle
--- a/testdata/images/bundles/own-namespace-operator/v1.0.0/manifests/olm.operatorframework.com_ownnamespaces.yaml
+++ b/testdata/images/bundles/own-namespace-operator/v1.0.0/manifests/olm.operatorframework.com_ownnamespaces.yaml
@@ -1,4 +1,3 @@
----
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
--- a/testdata/images/bundles/single-namespace-operator/v1.0.0/manifests/olm.operatorframework.com_singlenamespaces.yaml
+++ b/testdata/images/bundles/single-namespace-operator/v1.0.0/manifests/olm.operatorframework.com_singlenamespaces.yaml
@@ -1,4 +1,3 @@
----
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:


testdata/images/bundles/single-namespace-operator/v1.0.0/manifests/singlenamespaceoperator.clusterserviceversion.yaml
```diff
@@ -0,0 +1,151 @@
...
+    createdAt: "2024-10-24T19:21:40Z"
+    operators.operatorframework.io/builder: operator-sdk-v1.34.1
+    operators.operatorframework.io/project_layout: go.kubebuilder.io/v4
+  name: singlenamespaceoperator.v1.0.0
+  namespace: placeholder
+spec:
+  apiservicedefinitions: {}
+  customresourcedefinitions:
+  owned:
+    - description: Configures subsections of Alertmanager configuration specific to each namespace
```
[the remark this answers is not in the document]
perdasilva: I've updated it to something else

testdata/images/bundles/own-namespace-operator/v1.0.0/manifests/ownnamespaceoperator.clusterserviceversion.yaml
```diff
@@ -0,0 +1,151 @@
...
+              containers:
+              - name: busybox
+                image: busybox:1.36
+                command:
+                - 'sleep'
+                - '1000'
+                securityContext:
+                  runAsUser: 1000
+                  runAsNonRoot: true
+                serviceAccountName: simple-bundle-manager
```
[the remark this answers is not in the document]
perdasilva: fixed

internal/operator-controller/rukpak/bundle/config.go
```diff
@@ -22,10 +22,10 @@ type Config struct {
 // - rv is nil
 // - bytes is not a valid YAML/JSON object
 // - bytes is a valid YAML/JSON object but does not follow the registry+v1 schema
-// if bytes is nil a nil bundle.Config is returned
+// if bytes is nil it will be treated as an empty json object ({})
 func UnmarshallConfig(bytes []byte, rv1 RegistryV1, installNamespace string) (*Config, error) {
```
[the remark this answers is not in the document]
perdasilva: Fixed

internal/operator-controller/rukpak/bundle/config.go
```diff
@@ -83,20 +83,19 @@ func validateConfig(config *Config, installNamespace string, bundleInstallModeSe
...
-// AllNamespaces install mode is not supported and SingleNamespace is supported
+// AllNamespaces install mode is not supported and SingleNamespace and/or OwnNamespace is supported
 func isWatchNamespaceConfigRequired(bundleInstallModeSet sets.Set[v1alpha1.InstallMode]) bool {
-	return !bundleInstallModeSet.Has(v1alpha1.InstallMode{Type: v1alpha1.InstallModeTypeAllNamespaces, Supported: true}) &&
-		bundleInstallModeSet.Has(v1alpha1.InstallMode{Type: v1alpha1.InstallModeTypeSingleNamespace, Supported: true})
+	return isWatchNamespaceConfigSupported(bundleInstallModeSet) &&
+		!bundleInstallModeSet.Has(v1alpha1.InstallMode{Type: v1alpha1.InstallModeTypeAllNamespaces, Supported: true})
 }
 
 // formatUnmarshallError format JSON unmarshal errors to be more readable
```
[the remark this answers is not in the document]
perdasilva: Fixed

internal/operator-controller/rukpak/render/render.go
```diff
@@ -160,10 +160,10 @@ func validateTargetNamespaces(rv1 *bundle.RegistryV1, installNamespace string, t
 		// in case only the MultiNamespace install mode is supported by the bundle.
 		// If AllNamespaces mode is supported, the default will be [""] -> watch all namespaces
 		// If only OwnNamespace is supported, the default will be [install-namespace] -> only watch the install/own namespace
```
camilamacedo86, line 162:
```suggestion
		// If AllNamespaces is not supported, no default is applied and configuration must provide the target namespace(s)
```

api/v1/clusterextension_types.go
```diff
@@ -535,6 +535,23 @@ type ClusterExtension struct {
 	Status ClusterExtensionStatus `json:"status,omitempty"`
```
camilamacedo86, line 535: I think we need to change here: https://github.com/operator-framework/operator-controller/blob/926d57e1e6559f7ac0ef69a47b3d50d96975ec38/docs/draft/howto/single-ownnamespace-install.md?plain=1#L61

With somthing like

`- *AllNamespaces*: `watchNamespace` is empty, or not set **and** the bundle supports the AllNamespaces install mode. Bundles that lack AllNamespaces support now require explicit configuration.`

internal/operator-controller/applier/provider_test.go
```diff
@@ -330,11 +357,53 @@ func Test_RegistryV1ManifestProvider_SingleOwnNamespaceSupport(t *testing.T) {
...
+				Config: &ocv1.ClusterExtensionConfig{
+					ConfigType: ocv1.ClusterExtensionConfigTypeInline,
+					Inline: &apiextensionsv1.JSON{
+						Raw: []byte(`{"watchNamespace": "install-namespace"}`),
+					},
+				},
 			},
 		})
 		require.NoError(t, err)
 	})
```
camilamacedo86, line 361:
```suggestion
	// Checks that OwnNamespace bundles use the install namespace as targetNamespaces
	// when watchNamespace is the same. (targetNamespaces == [install-namespace] and no error)
	t.Run("sets targetNamespaces to the install namespace for {OwnNamespace} bundles", func(t *testing.T) {
		expectedTargetNamespace := "install-namespace"
		provider := applier.RegistryV1ManifestProvider{
			BundleRenderer: render.BundleRenderer{
				ResourceGenerators: []render.ResourceGenerator{
					func(rv1 *bundle.RegistryV1, opts render.Options) ([]client.Object, error) {
						require.Equal(t, []string{expectedTargetNamespace}, opts.TargetNamespaces)
						return nil, nil
					},
				},
			},
			IsSingleOwnNamespaceEnabled: true,
		}
		bundleFS := bundlefs.Builder().WithPackageName("test").
			WithCSV(clusterserviceversion.Builder().WithInstallModeSupportFor(v1alpha1.InstallModeTypeOwnNamespace).Build()).Build()

		_, err := provider.Get(bundleFS, &ocv1.ClusterExtension{
			Spec: ocv1.ClusterExtensionSpec{
				Namespace: expectedTargetNamespace,
				Config: &ocv1.ClusterExtensionConfig{
					ConfigType: ocv1.ClusterExtensionConfigTypeInline,
					Inline: &apiextensionsv1.JSON{
						Raw: []byte(`{"watchNamespace": "install-namespace"}`),
					},
				},
			},
		})
		require.NoError(t, err)
	})
```

WDYT?

api/v1/clusterextension_types.go
```diff
@@ -535,6 +535,23 @@ type ClusterExtension struct {
...
+	case ClusterExtensionConfigTypeInline:
+		if e.Spec.Config.Inline != nil {
+			return e.Spec.Config.Inline.Raw
+		}
+		return nil
+	default:
+		return nil
+	}
+}
+
```
tmshort: Is this the right place for this? A function within a `_types.go` file?

perdasilva: I'm not sure. I think the method makes sense. But I have no strong opinions on whether it should be factored out into its own file or not, or what that file should be called.

internal/operator-controller/applier/provider_test.go
perdasilva: If I got it right, what was missing was checking that the appropriate option was being set (i.e. `require.Equal(t, []string{expectedTargetNamespace}, opts.TargetNamespaces)`), right?

I've updated the `accepts bundles with {OwnNamespace} install modes when the appropriate configuration is given` to cover that

internal/operator-controller/applier/provider_test.go
```diff
@@ -321,20 +312,89 @@ func Test_RegistryV1ManifestProvider_SingleOwnNamespaceSupport(t *testing.T) {
...
-			WithCSV(clusterserviceversion.Builder().WithInstallModeSupportFor(v1alpha1.InstallModeTypeOwnNamespace).Build()).Build()
+			WithCSV(clusterserviceversion.Builder().WithInstallModeSupportFor(v1alpha1.InstallModeTypeSingleNamespace).Build()).Build()
+
 		_, err := provider.Get(bundleFS, &ocv1.ClusterExtension{
 			Spec: ocv1.ClusterExtensionSpec{
 				Namespace: "install-namespace",
 			},
 		})
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "required field \"watchNamespace\" is missing")
```
camilamacedo86, line 329:
``` suggestion
		require.Contains(t, err.Error(), "watchNamespace is required for SingleNamespace install mode")
```

What about adding an install mode and having unique errors, not only to make it easier to understand but also to make troubleshooting easier?

internal/operator-controller/applier/provider_test.go
```diff
@@ -321,20 +312,89 @@ func Test_RegistryV1ManifestProvider_SingleOwnNamespaceSupport(t *testing.T) {
...
+		}
+		bundleFS := bundlefs.Builder().WithPackageName("test").
+			WithCSV(clusterserviceversion.Builder().WithInstallModeSupportFor(v1alpha1.InstallModeTypeOwnNamespace).Build()).Build()
+		_, err := provider.Get(bundleFS, &ocv1.ClusterExtension{
+			Spec: ocv1.ClusterExtensionSpec{
+				Namespace: "install-namespace",
+			},
+		})
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "required field \"watchNamespace\" is missing")
```
camilamacedo86, line 374:
```suggestion
		require.Contains(t, err.Error(), "watchNamespace is required for OwnNamespace install mode")
```
What about adding an install mode and having unique errors, not only to make it easier to understand but also to make troubleshooting easier?

internal/operator-controller/rukpak/bundle/config_test.go
```diff
@@ -91,16 +91,28 @@ func Test_UnmarshallConfig(t *testing.T) {
...
-			rawConfig:             []byte(`{"watchNamespace": "some-namespace"}`),
-			expectedErrMessage:    "unknown field \"watchNamespace\"",
+			rawConfig:             []byte(`{"watchNamespace": null}`),
+			expectedErrMessage:    "required field \"watchNamespace\" is missing",
+		},
+		{
+			name:                  "reject with required field when install modes {OwnNamespace} and watchNamespace is missing",
+			supportedInstallModes: []v1alpha1.InstallModeType{v1alpha1.InstallModeTypeOwnNamespace},
+			rawConfig:             []byte(`{}`),
+			expectedErrMessage:    "required field \"watchNamespace\" is missing",
```
camilamacedo86, line 103:
```suggestion
			expectedErrMessage:    "watchNamespace is required for OwnNamespace install mode",
```

What about adding an install mode and having unique errors, not only to make it easier to understand but also to make troubleshooting easier?

internal/operator-controller/rukpak/bundle/config_test.go
```diff
@@ -91,16 +91,28 @@ func Test_UnmarshallConfig(t *testing.T) {
...
 		},
 		{
-			name:                  "reject with unknown field when install modes {MultiNamespace, OwnNamespace}",
+			name:                  "reject with required field when install modes {MultiNamespace, OwnNamespace} and watchNamespace is null",
 			supportedInstallModes: []v1alpha1.InstallModeType{v1alpha1.InstallModeTypeMultiNamespace, v1alpha1.InstallModeTypeOwnNamespace},
-			rawConfig:             []byte(`{"watchNamespace": "some-namespace"}`),
-			expectedErrMessage:    "unknown field \"watchNamespace\"",
+			rawConfig:             []byte(`{"watchNamespace": null}`),
+			expectedErrMessage:    "required field \"watchNamespace\" is missing",
```
camilamacedo86:
```suggestion
			expectedErrMessage:    "watchNamespace is required for OwnNamespace install mode",
```

api/v1/clusterextension_types.go
perdasilva: I've done a more comprehensive doc update ^^ thanks for pointing this out - I totally forgot about it

internal/operator-controller/applier/provider_test.go
perdasilva: We don't want to surface install mode concerns through the errors. Treat it as configuration.

internal/operator-controller/applier/provider_test.go
perdasilva: I don't think we want to surface these concerns up. It's not that bundle supports this or that. It's configuration is required/optional, the value is valid/invalid, etc. Install modes are a v0 thing. We don't want to leak it here I don't think.

api/v1/clusterextension_types.go
camilamacedo86: If we do an analogy with the OO (object-oriented) way, it would live inside the class itself;  
i.e. something like `MyObject.GetBytes()`. So yeah, thinking in DDD/OO style, it seems to fit here.

But I have one concern. Since this is in the API package, it becomes public,  
and once it’s public, we can’t easily change or remove it later.

So unless we’re 100% sure we want to keep and support this method as part of the public API,  
it might be better to move it to an internal package — or avoid exporting it for now.

camilamacedo86 (commented):
Great work and amazing catch 🎉  

I have only one real concern → https://github.com/operator-framework/operator-controller/pull/2283/files#r2456773553  
I think we should move that part out of the public API.  

The other things are just small nits — it would be nice if we could include the install modes in the error messages, so each case has a more unique and clearer message. That usually helps a lot when troubleshooting.  

Beyond that, it looks like a very solid solution to me 🚀

api/v1/clusterextension_types.go
joelanford: I'm trying to remember the specifics, but I remember it being inadvisable to attach functionality directly to the API types in the API package.

I _think_ it was mostly related to introducing coupling with shared dependencies that makes it hard/impossible for other projects to import the API.

Should we stick to type definitions in the API package and define a separate package that has helpers like this?

joelanford: Also, this function doesn't seem particularly well-suited to handle potential future extensions of sourcing configs from non-inline sources. Sourcing configs from secrets, for example, would require a kubernetes secret client, which we cannot attach to the type as a field and which we are not able to pass into the current function as a parameter.

camilamacedo86: >Should we stick to type definitions in the API package and define a separate package that has helpers like this?

@joelanford  
Will those helpers become public as well?  

- If yes, then maybe it makes sense to keep them in the API file.  
- If not, then that actually addresses my concern above. 🙂

joelanford: fwiw, this is what I had come up with as a helper in my prototype: https://github.com/joelanford/operator-controller/blob/5f255eef69b424e09d2630366059944ba5875411/internal/values/values.go

perdasilva: I'll remove it

perdasilva:
> Just because of the new method added to the public API (exported). I think we should move it out of the API package. → https://github.com/operator-framework/operator-controller/pull/2283/files#r2456565041

I've removed that and made it a private utility closer to where it is used

PR #2284, opened 2025-10-23
:seedling: Add yamlfmt to ensure consistent testdata yaml file formatting

perdasilva:
# Description

- Adds the `yamlfmt` tool [ref](https://github.com/google/yamlfmt) to bingo 
- Updates the `make fmt` target to run yamlfmt against the testdata directory
- Applies consistent formatting to the testdata yaml files

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/testdata/images/bundles/test-operator/v1.0.0/manifests/bundle.configmap.yaml
+++ b/testdata/images/bundles/test-operator/v1.0.0/manifests/bundle.configmap.yaml
@@ -4,9 +4,8 @@ metadata:
   name: test-configmap
   annotations:
     shouldNotTemplate: >
-      The namespace is {{ $labels.namespace }}. The templated
-      $labels.namespace is NOT expected to be processed by OLM's
-      rendering engine for registry+v1 bundles.
+      The namespace is {{ $labels.namespace }}. The templated $labels.namespace is NOT expected to be processed by OLM's rendering engine for registry+v1 bundles.
+
 data:
   version: "v1.0.0"
   name: "test-configmap"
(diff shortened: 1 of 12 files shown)


Makefile
```diff
@@ -186,8 +186,9 @@ fix-lint: $(GOLANGCI_LINT) #EXHELP Fix lint issues
 	$(GOLANGCI_LINT) run --fix --build-tags $(GO_BUILD_TAGS) $(GOLANGCI_LINT_ARGS)
 
 .PHONY: fmt
-fmt: #EXHELP Formats code
+fmt: $(YAMLFMT) #EXHELP Formats code
 	go fmt ./...
+	$(YAMLFMT) testdata
```
camilamacedo86, line 191: But I think it won’t be caught in the CI.
I mean, will it actually fail after this PR if the code isn’t formatted?
I think we should make sure that it’s being checked as well.

Just to share — in Kubebuilder we use this: ( to lint yaml )
👉 https://github.com/kubernetes-sigs/kubebuilder/blob/master/Makefile#L128-L131

perdasilva: it should be caught by the verify target since it calls the fmt target as well. So, changes ought to get caught by CI.

perdasilva: right?

perdasilva: Dropping the formatting commit to test

perdasilva: It does fail: https://github.com/operator-framework/operator-controller/actions/runs/18756733765/job/53510512776?pr=2284

PR #2291, opened 2025-10-28
:seedling: Refactor BundleValidator to a concrete class

perdasilva:
# Description

The current BundleValidator is implemented almost as a framework, its a singleton that always exists in memory, and it feels a bit awkward to me. I was thinking it might be better to have it as a concrete thing that can be instantiated independently and at the point of use, if necessary.

In this PR we move BundleValidator to a concrete struct with the validation functions hanging off that.

I'm wondering whether:
- people agree that a concrete struct might be better in general
- whether I should make the Check* methods private, and
- whether we should just have a single big tabular test that exercises all of the Check* functions, or if its better to test them individually and have a test that checks whether all check functions are being exercised by the `Validate` method.

Note: the changes are currently implemented under a `v2` directory. The final form of this PR would replace the current implementation in `validators` with this. Maybe even moving it to the top level of the `render` package.

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/operator-controller/rukpak/render/registryv1/validators/v2/validator.go
+++ b/internal/operator-controller/rukpak/render/registryv1/validators/v2/validator.go
@@ -1,0 +1,356 @@
+package validator
+
+import (
+	"cmp"
+	"errors"
+	"fmt"
+	"maps"
+	"slices"
+	"strings"
+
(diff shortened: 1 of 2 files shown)


perdasilva:
closing as stale

Issue #2008, opened 2025-06-04
Install error message getting propagated to most conditions


perdasilva:
In cases where the extension cannot be installed, the error message is getting added to most conditions. We should probably not propagate the error message to the deprecation conditions. For example:

```
{
    "lastTransitionTime": "2025-06-03T07:51:05Z",
    "message": "validating bundle \"devspacesoperator.v3.21.0\": bundle \"devspacesoperator.v3.21.0\" has a dependency declared via property \"olm.package.required\" which is currently not supported",
    "observedGeneration": 1,
    "reason": "Retrying",
    "status": "True",
    "type": "Progressing"
},
{
    "lastTransitionTime": "2025-06-03T07:51:05Z",
    "message": "No bundle installed",
    "observedGeneration": 1,
    "reason": "Failed",
    "status": "False",
    "type": "Installed"
},
{
    "lastTransitionTime": "2025-06-03T07:51:05Z",
    "message": "validating bundle \"devspacesoperator.v3.21.0\": bundle \"devspacesoperator.v3.21.0\" has a dependency declared via property \"olm.package.required\" which is currently not supported",
    "observedGeneration": 1,
    "reason": "Failed",
    "status": "False",
    "type": "Deprecated"
},
{
    "lastTransitionTime": "2025-06-03T07:51:05Z",
    "message": "validating bundle \"devspacesoperator.v3.21.0\": bundle \"devspacesoperator.v3.21.0\" has a dependency declared via property \"olm.package.required\" which is currently not supported",
    "observedGeneration": 1,
    "reason": "Failed",
    "status": "False",
    "type": "PackageDeprecated"
},
{
    "lastTransitionTime": "2025-06-03T07:51:05Z",
    "message": "validating bundle \"devspacesoperator.v3.21.0\": bundle \"devspacesoperator.v3.21.0\" has a dependency declared via property \"olm.package.required\" which is currently not supported",
    "observedGeneration": 1,
    "reason": "Failed",
    "status": "False",
    "type": "ChannelDeprecated"
},
{
    "lastTransitionTime": "2025-06-03T07:51:05Z",
    "message": "validating bundle \"devspacesoperator.v3.21.0\": bundle \"devspacesoperator.v3.21.0\" has a dependency declared via property \"olm.package.required\" which is currently not supported",
    "observedGeneration": 1,
    "reason": "Failed",
    "status": "False",
    "type": "BundleDeprecated"
}
```

joelanford:
[...]

PR #2296, opened 2025-11-01
🐛  Fix deprecation conditions showing install errors and improve condition semantics

camilamacedo86:
Fixes deprecation conditions showing install errors instead of actual deprecation status.

### 1. Install errors leaked into deprecation conditions

When installation failed, the error appeared in ALL conditions including deprecation ones:

```yaml
- type: PackageDeprecated
  status: "False"
  reason: Failed
  message: "validating bundle: dependency not supported"  # ❌ Wrong place!
```

Users couldn't tell if something was actually deprecated.

### 2. Wrong bundle checked during upgrades

When upgrading v1.0.0 → v1.0.1, deprecation showed v1.0.1's status even when the upgrade failed.

This violates Kubernetes convention: **status must show actual state, not desired state**.

---

### What we do now

| Situation | What we show |
|-----------|--------------|
| **IS deprecated** | Condition with `status: True` |
| **NOT deprecated** | Condition absent (no clutter!) |
| **Can't check catalog** | Condition with `status: Unknown` |
| **No bundle installed** | BundleDeprecated shows `status: Unknown, reason: Absent` |

### Key improvements

1. **Install errors stay separate**
   - Errors → `Progressing` and `Installed` conditions only
   - Deprecation → catalog data only (never errors)

2. **Use installed bundle** (not resolved)
   - Checks what's actually running
   - Follows Kubernetes "status = actual state" convention

3. **Correct reason values**
   - `Deprecated` → when `True`
   - `DeprecationStatusUnknown` → when `Unknown`
   - `Absent` → when no bundle installed (BundleDeprecated only)

---

## Before/After Comparison

| What Changed | Main (Before) | This PR (After) |
|--------------|---------------|-----------------|
| **Install errors** | Leak into deprecation conditions ❌ | Stay in Progressing/Installed only ✅ |
| **Catalog unavailable** | Shows `False` (misleading) ❌ | Shows `Unknown` (honest) ✅ |
| **Bundle reference** | Uses resolved bundle ❌ | Uses installed bundle ✅ |
| **Reason values** | Always `Deprecated` ❌ | Context-specific ✅ |

---

### Example 1: Install Error (Bug #2008)

**Before:**
```yaml
- type: Progressing
  status: "True"
  message: "validation failed: dependency not supported"
- type: PackageDeprecated
  status: "False"
  reason: Failed
  message: "validation failed: dependency not supported"  # ❌ Error leaked!
```

**After:**
```yaml
- type: Progressing
  status: "True"
  message: "validation failed: dependency not supported"
# ✅ PackageDeprecated absent (error didn't leak!)
```

### Example 2: Catalog Removed

**Before:**
```yaml
- type: BundleDeprecated
  status: "False"              # ❌ Misleading: claims "not deprecated"
  reason: Failed
```

**After:**
```yaml
- type: BundleDeprecated
  status: "Unknown"            # ✅ Honest: we can't check
  reason: DeprecationStatusUnknown
  message: "deprecation status unknown: catalog data unavailable"
```


internal/operator-controller/controllers/clusterextension_controller.go
```diff
@@ -326,83 +343,143 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1.Cl
...
+
+	return info
+}
+
+func setDeprecatedConditions(ext *ocv1.ClusterExtension, info DeprecationInfo) {
+	packageMessages := collectDeprecationMessages(info.PackageEntries)
+	channelMessages := collectDeprecationMessages(info.ChannelEntries)
+	bundleMessages := collectDeprecationMessages(info.BundleEntries)
+
+	messages := make([]string, 0, len(packageMessages)+len(channelMessages)+len(bundleMessages))
```
perdasilva: suggestion: you could probably use `slices.Concat` to simplify a little

camilamacedo86: Done

internal/operator-controller/controllers/clusterextension_controller.go
```diff
@@ -326,83 +343,140 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1.Cl
...
+//   - BundleDeprecated: reflects the installed bundle (Unknown/Absent when nothing installed)
+//   - Deprecated (rollup): true if any of the above signals a deprecation
+//
+// Install or validation errors never appear here because they belong on the
+// Progressing/Installed conditions instead. Callers should invoke this after reconcile
+// finishes (for example via a defer) so catalog data replaces any transient error messages.
 func SetDeprecationStatus(ext *ocv1.ClusterExtension, bundleName string, deprecation *declcfg.Deprecation) {
-	deprecations := map[string][]declcfg.DeprecationEntry{}
-	channelSet := sets.New[string]()
+	info := buildDeprecationInfo(ext, bundleName, deprecation)
```
perdasilva: suggestion: since this method is quite small, it might be worth merging it with `setDeprecatedConditions` (or rather just do something like:

```
info := buildDeprecationInfo(ext, bundleName, deprecation)
// setDeprecatedConditions implementation
```

and removing `setDeprecatedConditions`

internal/operator-controller/controllers/clusterextension_controller.go
```diff
@@ -326,83 +343,140 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1.Cl
...
+	return info
+}
+
+func setDeprecatedConditions(ext *ocv1.ClusterExtension, info DeprecationInfo) {
+	packageMessages := collectDeprecationMessages(info.PackageEntries)
+	channelMessages := collectDeprecationMessages(info.ChannelEntries)
+	bundleMessages := collectDeprecationMessages(info.BundleEntries)
+
+	messages := slices.Concat(packageMessages, channelMessages)
+	if info.BundleStatus == metav1.ConditionTrue {
```
perdasilva: we already skip adding bundle entries if the bundleName is "" so we can probably just skip this check

internal/operator-controller/controllers/clusterextension_controller.go
```diff
@@ -229,37 +230,53 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1.Cl
...
+		// even if installation later fails or never begins.
 		if err != nil {
 			// Note: We don't distinguish between resolution-specific errors and generic errors
 			setStatusProgressing(ext, err)
 			setInstalledStatusFromRevisionStates(ext, revisionStates)
+			// ensureAllConditionsWithReason writes the failure message to every condition.
+			// The deferred SetDeprecationStatus call overwrites the deprecation conditions right
+			// away so they only reflect catalog data. This also supports the case where resolution
+			// returned deprecation info even though it could not find an installable bundle.
 			ensureAllConditionsWithReason(ext, ocv1.ReasonFailed, err.Error())
```
perdasilva: Maybe we should refactor this method as well? Seems weird to set everything to failed, then to overwrite that. Maybe let the defer handle deprecations and this method (with a new name) handle everything else?

perdasilva: It seems this is the only place this method is called anyway - so it could be just for setting failures?

camilamacedo86: I tried to address so I am closing to facilitate
Let me know wdyt?

internal/operator-controller/controllers/clusterextension_controller.go
camilamacedo86: Marking as solved 
I think it is addressed as well

internal/operator-controller/controllers/clusterextension_controller.go
```diff
@@ -157,6 +161,17 @@ func ensureAllConditionsWithReason(ext *ocv1.ClusterExtension, reason v1alpha1.C
 	}
 }
 
+// isDeprecationCondition reports whether the given type is one of the deprecation
+// conditions we manage separately.
+func isDeprecationCondition(condType string) bool {
```
pedjak: nit: I would rather add this helper function to `api/v1/clusterextension_types.go`

internal/operator-controller/controllers/clusterextension_controller.go
```diff
@@ -139,13 +140,16 @@ func (r *ClusterExtensionReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
+// If one is missing, we add it with the given reason and message so users see why
+// reconcile failed. Deprecation conditions are handled later by SetDeprecationStatus.
+func ensureFailureConditionsWithReason(ext *ocv1.ClusterExtension, reason v1alpha1.ConditionReason, message string) {
 	for _, condType := range conditionsets.ConditionTypes {
+		if isDeprecationCondition(condType) {
+			continue
+		}
 		cond := apimeta.FindStatusCondition(ext.Status.Conditions, condType)
 		if cond == nil {
-			// Create a new condition with a valid reason and add it
```
pedjak, line 177: nit: while doing changes here, we can simplify the code and drop the above `if` statement if we use something like this:

```golang
SetStatusCondition(&ext.Status.Conditions, metav1.Condition{
				Type:               condType,
				Reason:             string(reason),
				Message:            message,
			})
```

because `SetStatusCondtion` add the new condition if it does not exist, but update the existing one if message/reason/status changes.

internal/operator-controller/controllers/clusterextension_controller.go
```diff
@@ -326,83 +345,141 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1.Cl
...
+// This keeps the deprecation conditions focused on catalog information:
+//   - PackageDeprecated: true if the catalog marks the package deprecated
+//   - ChannelDeprecated: true if any requested channel is marked deprecated
+//   - BundleDeprecated: reflects the installed bundle (Unknown/Absent when nothing installed)
+//   - Deprecated (rollup): true if any of the above signals a deprecation
+//
+// Install or validation errors never appear here because they belong on the
+// Progressing/Installed conditions instead. Callers should invoke this after reconcile
+// finishes (for example via a defer) so catalog data replaces any transient error messages.
+func SetDeprecationStatus(ext *ocv1.ClusterExtension, installedBundleName string, deprecation *declcfg.Deprecation) {
```
pedjak: would it make sens to move this function to api and restructure it like:

```golang
func (ext *ocv1.ClusterExtension) SetDeprecationStatus  (installedBundleName string, deprecation *declcfg.Deprecation)
```

internal/operator-controller/controllers/clusterextension_controller.go
camilamacedo86: > nit: I would rather add this helper function to api/v1/clusterextension_types.go

I don’t think we should add this helper to api/v1/clusterextension_types.go.
Anything we put in the public API becomes part of what users can rely on, which means we have to support it for a very long time and can’t easily change or remove it.

Since this helper is meant for internal use only and not something we want people using directly, it shouldn’t be part of the public API. We should keep it internal instead.

internal/operator-controller/controllers/clusterextension_controller.go
camilamacedo86: I understand from where you came from, but in our case we should really avoid as much as possible expose anything in the public API. 

Same: https://github.com/operator-framework/operator-controller/pull/2296#discussion_r2490216197 reason for we do not add there.

camilamacedo86:
Hi @pedjak 

> I think we should also update the API docs:
> https://github.com/operator-framework/operator-controller/blob/main/api/v1/clusterextension_types.go#L486-L491
> given that we currently say that that we MAY set them if something is deprecated. This PR changes the behavior and sets the deprecated conditions always.


Thanks for calling that out! 

I agree the block could be clearer overall, but this PR doesn’t change the behavior the comment describes—we still only surface deprecation conditions when the catalog tells us something is deprecated. The new logic just makes sure the controller actually honors what’s written there (and adds the Unknown case until a bundle lands). I’d rather keep doc polish out of this PR so we can tackle a broader wording update separately.

WDYT in a fallow up make / enhance doc api explaination for .. with this purpose?

internal/operator-controller/controllers/clusterextension_controller.go
camilamacedo86: Great callout, thanks! For this helper I think we want to only want to add the condition when it’s missing. Other code already sets richer status and messages (i.e  `setInstalledStatusCondition` or `setStatusProgressing`), so I think would be better we ensure that we don’t overwritten that out. So it seems for me that is better we keep the guard in place, but I appreciate the idea.

pedjak:
> I agree the block could be clearer overall, but this PR doesn’t change the behavior the comment describes—we still only surface deprecation conditions when the catalog tells us something is deprecated. The new logic just makes sure the controller actually honors what’s written there (and adds the Unknown case until a bundle lands). I’d rather keep doc polish out of this PR so we can tackle a broader wording update separately.


Maybe it is just how I read it, here is the actual snippet:

```
When the ClusterExtension is sourced from a catalog, if may also communicate a deprecation condition.
These are indications from a package owner to guide users away from a particular package, channel, or bundle.
BundleDeprecated is set if the requested bundle version is marked deprecated in the catalog.
ChannelDeprecated is set if the requested channel is marked deprecated in the catalog.
PackageDeprecated is set if the requested package is marked deprecated in the catalog.
Deprecated is a rollup condition that is present when any of the deprecated conditions are present.
```

`May` sound to me that this condition is optional, but it is not anymore, right? Your PR changes the logic so that it is exposed always, in case when there is no deprecation the condition status if `False`, correct? If so, the description saying that `BundleDeprecated is set if the requested bundle version is marked deprecated in the catalog.` does not hold anymore, i.e. it should be changed to `BundleDeprecated is set to True if the requested bundle version is marked deprecated in the catalog.`

> WDYT in a fallow up make / enhance doc api explaination for .. with this purpose?

What blocks us from doing it this PR?

internal/operator-controller/controllers/clusterextension_controller.go
pedjak: > I don’t think we should add this helper to api/v1/clusterextension_types.go. Anything we put in the public API becomes part of what users can rely on, which means we have to support it for a very long time and can’t easily change or remove it.

I though our API is our CRD, not golang bindings?

internal/operator-controller/controllers/clusterextension_controller.go
pedjak: > Great callout, thanks! For this helper I think we want to only want to add the condition when it’s missing. Other code already sets richer status and messages (i.e `setInstalledStatusCondition` or `setStatusProgressing`), so I think would be better we ensure that we don’t overwritten that out. So it seems for me that is better we keep the guard in place, but I appreciate the idea.

In case when a reason/status/message change, do we update somewhere the existing condition to reflect that?

camilamacedo86:
Hi @pedjak 

- On main we already set these conditions every reconcile. If the catalog had nothing, they still showed up as False. This PR keeps that behavior; the only tweak is that BundleDeprecated now stays Unknown/Absent until a bundle actually installs. See the issue: https://github.com/operator-framework/operator-controller/issues/2008
- Once something installs, we look at that bundle: if the catalog marks it deprecated we set BundleDeprecated=True; if not, False; otherwise Unknown. The doc line could be clearer about that.

I’d rather keep this PR focused on the bug fix **(stopping the install error leaks and handling Unknown/Absent correctly)**.
I agree the wording could be improved, but I’m a bit concerned that updating it here might give the wrong impression that we’re changing behavior, which could make it harder for this PR to land. That’s was only my concern.

**However, following your suggestion, I updated the descriotion to be more compreenshive.**

api/v1/clusterextension_types.go
```diff
@@ -483,12 +483,12 @@ type ClusterExtensionStatus struct {
...
+	// When the ClusterExtension is sourced from a catalog, it surfaces deprecation conditions based on catalog metadata.
 	// These are indications from a package owner to guide users away from a particular package, channel, or bundle.
-	// BundleDeprecated is set if the requested bundle version is marked deprecated in the catalog.
-	// ChannelDeprecated is set if the requested channel is marked deprecated in the catalog.
-	// PackageDeprecated is set if the requested package is marked deprecated in the catalog.
-	// Deprecated is a rollup condition that is present when any of the deprecated conditions are present.
+	// PackageDeprecated becomes True when the catalog marks the requested package deprecated; otherwise it stays False.
+	// ChannelDeprecated becomes True when any requested channel is marked deprecated; otherwise it stays False.
+	// BundleDeprecated reports the catalog status of the installed bundle: it remains Unknown until a bundle installs, then becomes True or False depending on whether the catalog marks that bundle deprecated.
+	// Deprecated is a rollup that mirrors True whenever any of the specific deprecation conditions are True.
```
camilamacedo86: @pedjak ^ see

internal/operator-controller/controllers/clusterextension_controller.go
pedjak: Found https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md#no-functions

internal/operator-controller/controllers/clusterextension_controller.go
```diff
@@ -229,37 +248,45 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1.Cl
...
+		installedBundleName := ""
+		if revisionStatesPtr != nil && *revisionStatesPtr != nil && (*revisionStatesPtr).Installed != nil {
+			installedBundleName = (*revisionStatesPtr).Installed.Name
+		}
+		var resolvedDeprecationVal *declcfg.Deprecation
+		if resolvedDeprecationPtr != nil {
+			resolvedDeprecationVal = *resolvedDeprecationPtr
+		}
+		SetDeprecationStatus(ext, installedBundleName, resolvedDeprecationVal)
+	}(&resolvedDeprecation, &revisionStates)
```
pedjak: Could we actually simplify the deferring function, i.e. avoid passing args, and declare it as a closure?
```golang
defer func() {
		installedBundleName := ""
		if revisionStates != nil && revisionStates.Installed != nil {
			installedBundleName = revisionStates.Installed.Name
		}
		SetDeprecationStatus(ext, installedBundleName, resolvedDeprecation)
	}()
```

internal/operator-controller/controllers/clusterextension_controller.go
```diff
@@ -326,83 +353,141 @@ func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1.Cl
...
+func buildDeprecationInfo(ext *ocv1.ClusterExtension, installedBundleName string, deprecation *declcfg.Deprecation) DeprecationInfo {
+	info := DeprecationInfo{BundleStatus: metav1.ConditionUnknown}
+	var channelSet sets.Set[string]
 	if ext.Spec.Source.Catalog != nil {
-		for _, channel := range ext.Spec.Source.Catalog.Channels {
-			channelSet.Insert(channel)
-		}
+		channelSet = sets.New(ext.Spec.Source.Catalog.Channels...)
+	} else {
+		channelSet = sets.New[string]()
```
pedjak: to avoid `else` statement, we could simplify to:

```golang
channelSet := sets.New[string]()
if ext.Spec.Source.Catalog != nil {
  channelSet.Insert(ext.Spec.Source.Catalog.Channels...)
}
```

internal/operator-controller/controllers/clusterextension_controller.go
camilamacedo86: Add comment to address this one. Thx
So marking as solved

internal/operator-controller/controllers/clusterextension_controller.go
camilamacedo86: Added a comment to explain. Thx 👍

camilamacedo86:
>[crd-diff / crd-diff (pull_request)](https://github.com/operator-framework/operator-controller/actions/runs/19178753838/job/54829995498?pr=2296)F

That fails due change in the description to properly supplement it.
Not due breaking changes.
We should be able to ovewrite 

/overwrite crd-diff

api/v1/clusterextension_types.go
```diff
@@ -483,12 +483,15 @@ type ClusterExtensionStatus struct {
...
 	// When Progressing is False and the Reason is Blocked, the ClusterExtension has encountered an error that requires manual intervention for recovery.
 	//
-	// When the ClusterExtension is sourced from a catalog, if may also communicate a deprecation condition.
+	// When the ClusterExtension is sourced from a catalog, it may surface deprecation conditions based on catalog metadata.
 	// These are indications from a package owner to guide users away from a particular package, channel, or bundle.
-	// BundleDeprecated is set if the requested bundle version is marked deprecated in the catalog.
-	// ChannelDeprecated is set if the requested channel is marked deprecated in the catalog.
-	// PackageDeprecated is set if the requested package is marked deprecated in the catalog.
-	// Deprecated is a rollup condition that is present when any of the deprecated conditions are present.
+	// PackageDeprecated reports the requested package as deprecated when the catalog says so. When no catalog talks
```
perdasilva:
```suggestion
	// PackageDeprecated is set to True if the requested package is marked as deprecated in the catalog, False if it is not, or Unknown if the package is not found in the catalog
```

api/v1/clusterextension_types.go
```diff
@@ -483,12 +483,15 @@ type ClusterExtensionStatus struct {
...
+	// about the package, the condition stays Unknown instead of claiming False.
+	// ChannelDeprecated follows the same rules for each requested channel.
```
perdasilva:
```suggestion
	// ChannelDeprecated is set to True if the requested channel is marked as deprecated in the catalog, False if it is not, or Unknown if the channel is not found in the catalog
```

api/v1/clusterextension_types.go
```diff
@@ -483,12 +483,15 @@ type ClusterExtensionStatus struct {
...
+	// BundleDeprecated describes the installed bundle once one exists. Until a bundle installs, or when no catalog
```
perdasilva:
```suggestion
	// BundleDeprecated is set to True if the installed bundle is marked as deprecated in the catalog, False if it is not, or Unknown if the bundle is not found in the catalog, or could not be installed
```

api/v1/clusterextension_types.go
```diff
@@ -483,12 +483,15 @@ type ClusterExtensionStatus struct {
...
+	// data is available, it remains Unknown.
+	// Deprecated is still the rollup: it goes True when any individual deprecation condition is True, and Unknown
+	// when we have no catalog information to report.
```
perdasilva:
```suggestion
	// Deprecated is a rollup condition that is True when any individual deprecation condition is True, False when all deprecation conditions are False, and Unknown
	// when we have no catalog information to report.
```

camilamacedo86:
@perdasilva
Thank you so much for your help with the wording :-) and for taking the time to review and share all your suggestions.
I think everything is good now — all comments have been addressed!
Please feel free to take another look whenever you have a moment.

joelanford:
Not sure if this has been addressed in other comments (if so, please update the PR description!).

I noticed that we have `Reason: Deprecated` even when that status is Unknown or False. That should be fixed too.
- False should have `Reason: NotDeprecated` (or `Current`, `Active`, etc.)
- Unknown should perhaps have something about why it is unknown? (e.g. `DeprecationMetadataUnavailable` or something?)

joelanford:
One thing I didn't see answered in the description is the open question in the original TODO comment about what happens when the package and/or channel is present in multiple catalogs, but we didn't resolve a bundle:

> 		//       - Open question though: what if different catalogs have different opinions of what's deprecated.
> 		//         If we can't resolve a bundle, how do we know which catalog to trust for deprecation information?
> 		//         Perhaps if the package shows up in multiple catalogs and deprecations don't match, we can set
> 		//         the deprecation status to unknown? Or perhaps we somehow combine the deprecation information from
> 		//         all catalogs?

joelanford:
Is there an RFC for this? It kinda feels like we should have something written up in that form? I also wonder if we should tackle the question of whether to explicitly include the Deprecation conditions when we _would_ set them as False. They add a lot of noise to the YAML/JSON and `describe` output. And it feels like absence of these conditions would be reasonable interpretted as "not deprecated".

camilamacedo86:
Hi @joelanford 

Thanks for the feedback! You're absolutely right - **we should keep the TODO and open question for item 2**. **I've done that.** See the updated TODO with scenarios here: internal/operator-controller/controllers/clusterextension_controller.go

**The core logic for handling multiple catalogs hasn't changed** - the resolver still picks one. What changed is we now correctly handle the failure cases and don't lose deprecation information when things go wrong. For that yes, I think we need a follow up PR/discussion. **( out of scope )** 

Just to clarify the scope: this PR solves **item 1** (use installed bundle instead of resolved) and **#2008** (install errors leaking into deprecation conditions). The open question about conflicting deprecations from multiple catalogs is a separate edge case that deserves its own focused discussion and PR (IHMO, lets go step by step).

### Key Differences (What This PR Fixes)

| Aspect | Main | This PR |
|--------|------|---------|
| **Channel filtering** | ✅ Same | ✅ Same |
| **Catalog selection** | ✅ Same (resolver picks one) | ✅ Same |
| **When deprecation set** | Only after successful resolution | After resolution, after install, during rollout |
| **Catalog unavailable** | Shows `False` ❌ (wrong!) | Shows `Unknown` ✅ |
| **Resolution fails** | Deprecation lost ❌ | Deprecation shown ✅ |
| **Bundle reference** | Uses resolved bundle ❌ | Uses installed bundle ✅ |
| **Install errors** | Leak into deprecation conditions ❌ | Stay in Progressing/Installed ✅ |


Hope this addresses your concern! Let me know if you'd like me to clarify anything else.
I think we need to solve per parts, based on the above you still think that we need a RFC to move forward with the bug fix here, those changes or can we move forward and track the need to create a RFC to discuss item 2.

camilamacedo86:
Hey @joelanford — all your comments have been addressed.

We also have an RFC here: https://docs.google.com/document/d/1_XtISZ74rsd9-Zh9oQje1mEHjAfx0CHlf_50UJkXHTQ
.
In the doc description you’ll find the list of issues fixed, plus demos with examples.

I think it’s ready to 🚀 now.

c/c @pedjak @perdasilva @grokspawn

internal/operator-controller/controllers/clusterextension_controller.go
```diff
@@ -184,83 +192,154 @@ func ensureAllConditionsWithReason(ext *ocv1.ClusterExtension, reason v1alpha1.C
...
+	apimeta.RemoveStatusCondition(&ext.Status.Conditions, ocv1.TypeBundleDeprecated)
+
+	if !hasCatalogData {
+		// When catalog is unavailable, set all to Unknown.
+		// BundleDeprecated uses Absent only when no bundle installed.
+		bundleReason := ocv1.ReasonAbsent
+		if installedBundleName != "" {
+			bundleReason = ocv1.ReasonDeprecationStatusUnknown
+		}
+		setDeprecationCondition(ext, ocv1.TypeDeprecated, metav1.ConditionUnknown, ocv1.ReasonDeprecationStatusUnknown, "")
```
pedjak: should not we put in Reason some explanation why the status is equal to `Unknown`? Something like `NoCatalogData`? And also add the appropriate human readable message, instead of the empty string?

internal/operator-controller/controllers/clusterextension_controller.go
```diff
@@ -184,83 +192,154 @@ func ensureAllConditionsWithReason(ext *ocv1.ClusterExtension, reason v1alpha1.C
...
 
-	// next, set the Deprecated condition
-	status, reason, message := metav1.ConditionFalse, ocv1.ReasonDeprecated, ""
-	if len(deprecationMessages) > 0 {
-		status, reason, message = metav1.ConditionTrue, ocv1.ReasonDeprecated, strings.Join(deprecationMessages, ";")
-	}
+	return info
+}
+
+// setDeprecationCondition sets a single deprecation condition with less boilerplate.
```
pedjak: where do we save some boilerplate? `setDeprecationCondtion` is literary a wrapper around `SetStatusCondition`. IMHO, I see no added value,

internal/operator-controller/controllers/clusterextension_reconcile_steps.go
```diff
@@ -86,50 +86,63 @@ func RetrieveRevisionStates(r RevisionStatesGetter) ReconcileStepFunc {
...
-				// TODO: Right now, operator-controller only supports registry+v1 bundles and has no concept
-				//   of a "release" field. If/when we add a release field concept or a new bundle format
-				//   we need to re-evaluate use of `AsLegacyRegistryV1Version` so that we avoid propagating
-				//   registry+v1's semver spec violations of treating build metadata as orderable.
-				BundleMetadata: bundleutil.MetadataFor(resolvedBundle.Name, resolvedBundleVersion.AsLegacyRegistryV1Version()),
-			}
-		} else {
-			resolvedRevisionMetadata = state.revisionStates.RollingOut[0]
+		// Resolve a new bundle from the catalog
+		l.Info("resolving bundle")
```
pedjak: can we have this on debug level?

internal/operator-controller/controllers/common_controller_test.go
```diff
@@ -146,7 +146,7 @@ func TestClusterExtensionDeprecationMessageTruncation(t *testing.T) {
 		deprecationMessages = append(deprecationMessages, fmt.Sprintf("API version 'v1beta1' of resource 'customresources%d.example.com' is deprecated, use 'v1' instead", i))
 	}
 
-	longDeprecationMsg := strings.Join(deprecationMessages, "; ")
+	longDeprecationMsg := strings.Join(deprecationMessages, "\n")
```
pedjak, line 149: why is this change necessary?

internal/operator-controller/controllers/clusterextension_reconcile_steps.go
```diff
@@ -86,50 +86,63 @@ func RetrieveRevisionStates(r RevisionStatesGetter) ReconcileStepFunc {
...
+			setInstalledStatusFromRevisionStates(ext, state.revisionStates)
+			ensureFailureConditionsWithReason(ext, ocv1.ReasonFailed, err.Error())
+			return nil, err
+		}
+
+		state.resolvedRevisionMetadata = &RevisionMetadata{
+			Package: resolvedBundle.Package,
+			Image:   resolvedBundle.Image,
+			// TODO: Right now, operator-controller only supports registry+v1 bundles and has no concept
+			//   of a "release" field. If/when we add a release field concept or a new bundle format
```
pedjak, line 149: what tests are covering the added logic?

internal/operator-controller/controllers/clusterextension_controller.go
camilamacedo86: It seems a good idea. Added 

```
# When catalog is unavailable:
- type: PackageDeprecated
  status: Unknown
  reason: DeprecationStatusUnknown
  message: "deprecation status unknown: catalog data unavailable"

# When no bundle installed:
- type: BundleDeprecated
  status: Unknown
  reason: Absent
  message: "no bundle installed yet"
```

internal/operator-controller/controllers/clusterextension_controller.go
camilamacedo86: The diff is that we do not need add in all places 

```
    SetStatusCondition(&ext.Status.Conditions, metav1.Condition{
        Type:               condType,
        Status:             status,
        Reason:             reason,
        Message:            message,
    })
```

Example

```
SetStatusCondition(&ext.Status.Conditions, metav1.Condition{
			Type:               ocv1.TypeDeprecated,
			Status:             metav1.ConditionUnknown,
			Reason:             ocv1.ReasonDeprecationStatusUnknown,
			Message:            "deprecation status unknown: catalog data unavailable",
		})
		SetStatusCondition(&ext.Status.Conditions, metav1.Condition{
			Type:               ocv1.TypePackageDeprecated,
			Status:             metav1.ConditionUnknown,
			Reason:             ocv1.ReasonDeprecationStatusUnknown,
			Message:            "deprecation status unknown: catalog data unavailable",
		})
		SetStatusCondition(&ext.Status.Conditions, metav1.Condition{
			Type:               ocv1.TypeChannelDeprecated,
			Status:             metav1.ConditionUnknown,
			Reason:             ocv1.ReasonDeprecationStatusUnknown,
			Message:            "deprecation status unknown: catalog data unavailable",
		})
		SetStatusCondition(&ext.Status.Conditions, metav1.Condition{
			Type:               ocv1.TypeBundleDeprecated,
			Status:             metav1.ConditionUnknown,
			Reason:             bundleReason,
			Message:            bundleMessage,
		})
```


instead  we have 

```
setDeprecationCondition(ext, ocv1.TypeDeprecated, metav1.ConditionUnknown, ocv1.ReasonDeprecationStatusUnknown, "deprecation status unknown: catalog data unavailable")
        setDeprecationCondition(ext, ocv1.TypePackageDeprecated, metav1.ConditionUnknown, ocv1.ReasonDeprecationStatusUnknown, "deprecation status unknown: catalog data unavailable")
        setDeprecationCondition(ext, ocv1.TypeChannelDeprecated, metav1.ConditionUnknown, ocv1.ReasonDeprecationStatusUnknown, "deprecation status unknown: catalog data unavailable")
        setDeprecationCondition(ext, ocv1.TypeBundleDeprecated, metav1.ConditionUnknown, bundleReason, bundleMessage)
```
   
But if you do not see value in the helper I have to keep the directly call.
DONE 👍

internal/operator-controller/controllers/common_controller_test.go
camilamacedo86: **On main (before):**
- **Controller code:** `strings.Join(deprecationMessages, ";")`  ← no space
- **Test code:** `strings.Join(deprecationMessages, "; ")`  ← **has space** (already inconsistent!)

**In this PR (after):**
- **Controller code:** `strings.Join(...Messages, "\n")`
- **Test code:** `strings.Join(deprecationMessages, "\n")`
- Both now use newlines ✅

Note that, this test builds one **long** deprecation message to check **truncation**. The test needs to use the **same separator** as the controller code to accurately validate truncation behavior.

On main, they were already inconsistent (`";"` vs `"; "`). This PR fixes that **and** improves readability by using newlines.

**Why newlines are better**

- **With semicolons (hard to read):**
```yaml
- type: PackageDeprecated
  status: "True"
  message: "package 'foo' is deprecated;channel 'alpha' is deprecated;bundle 'bar' is deprecated"
```

- **With newlines (clearer):**
```yaml
- type: PackageDeprecated
  status: "True"
  message: |
    package 'foo' is deprecated
    channel 'alpha' is deprecated
    bundle 'bar' is deprecated
```

Each warning on its own line - much easier to read in `kubectl describe`. And [k8s API conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties) recommend human-readable condition messages. Multiple issues should be clearly distinguishable. Also, as far I understand other k8s resources (Pods, Events, Jobs) all use newlines for multi-line messages, not semicolons.

**PS:**If I properly remember, the newline change was suggested by Copilot during development.

internal/operator-controller/controllers/clusterextension_reconcile_steps.go
camilamacedo86: Asked for CLAUDE answered that ;-P 
Following its answer
# Test Coverage for Added Deprecation Logic

## Changes: Main vs This PR

### What changed in `clusterextension_reconcile_steps.go`

**On main branch:**
```go
// main branch
l.Info("resolving bundle")
var bm *ocv1.BundleMetadata
if state.revisionStates.Installed != nil {
    bm = &state.revisionStates.Installed.BundleMetadata
}
resolvedBundle, resolvedBundleVersion, resolvedDeprecation, err := r.Resolve(ctx, ext, bm)
if err != nil {
    // return error immediately - no deprecation set
    setInstalledStatusFromRevisionStates(ext, state.revisionStates)
    ensureAllConditionsWithReason(ext, ocv1.ReasonFailed, err.Error())
    return nil, err
}

// Only set deprecation after successful resolution
// TODO: deprecation status should reflect currently installed bundle, not resolved bundle
SetDeprecationStatus(ext, resolvedBundle.Name, resolvedDeprecation)  // ← uses RESOLVED bundle name
resolvedRevisionMetadata = &RevisionMetadata{
    Package: resolvedBundle.Package,
    Image:   resolvedBundle.Image,
    BundleMetadata: bundleutil.MetadataFor(resolvedBundle.Name, resolvedBundleVersion.AsLegacyRegistryV1Version()),
}
```

**In this PR:**
```go
// This PR
l.V(1).Info("resolving bundle")  // ← changed to debug level
var bm *ocv1.BundleMetadata
if state.revisionStates.Installed != nil {
    bm = &state.revisionStates.Installed.BundleMetadata
}
resolvedBundle, resolvedBundleVersion, resolvedDeprecation, err := r.Resolve(ctx, ext, bm)

// Get the installed bundle name for deprecation status.
// BundleDeprecated should reflect what's currently running, not what we're trying to install.
installedBundleName := ""
if state.revisionStates.Installed != nil {
    installedBundleName = state.revisionStates.Installed.Name  // ← uses INSTALLED bundle name
}

// Set deprecation status based on resolution results:
//  - If resolution succeeds: hasCatalogData=true, deprecation shows catalog data (nil=not deprecated)
//  - If resolution fails but returns deprecation: hasCatalogData=true, show package/channel deprecation warnings
//  - If resolution fails with nil deprecation: hasCatalogData=false, all conditions go Unknown
hasCatalogData := err == nil || resolvedDeprecation != nil  // ← NEW logic
SetDeprecationStatus(ext, installedBundleName, resolvedDeprecation, hasCatalogData)  // ← Set BEFORE error check

if err != nil {
    // deprecation already set above
    setInstalledStatusFromRevisionStates(ext, state.revisionStates)
    ensureFailureConditionsWithReason(ext, ocv1.ReasonFailed, err.Error())
    return nil, err
}

state.resolvedRevisionMetadata = &RevisionMetadata{
    Package: resolvedBundle.Package,
    Image:   resolvedBundle.Image,
    BundleMetadata: bundleutil.MetadataFor(resolvedBundle.Name, resolvedBundleVersion.AsLegacyRegistryV1Version()),
}
```

### Key Changes

| Change | Main | This PR |
|--------|------|---------|
| **Bundle name used** | `resolvedBundle.Name` (what we're trying to install) | `installedBundleName` (what's currently running) |
| **When set** | After successful resolution only | Before error check (always) |
| **Catalog unavailable** | No deprecation set (conditions stay False) | Deprecation set to Unknown |
| **hasCatalogData logic** | Not present | `err == nil \|\| resolvedDeprecation != nil` |
| **Log level** | `l.Info("resolving bundle")` | `l.V(1).Info("resolving bundle")` |

---

## Test Coverage

### New Tests Added in This PR

#### File: `clusterextension_controller_test.go`

**1. TestClusterExtensionResolutionFailsWithDeprecationData**
- **Line:** 191
- **Covers:** 
  - `hasCatalogData = true` when `err != nil && resolvedDeprecation != nil`
  - Deprecation shown even when resolution fails
  - Package/channel deprecation visible despite installation failure

**2. TestClusterExtensionUpgradeShowsInstalledBundleDeprecation**
- **Line:** 254
- **Covers:**
  - Uses **installed** bundle name (not resolved)
  - v1.0.0 installed (deprecated), v2.0.0 resolved (not deprecated)
  - BundleDeprecated shows True for v1.0.0 (actual state, not desired)
  - **This is the key behavioral change**

**3. TestClusterExtensionResolutionFailsWithoutCatalogDeprecationData**
- **Line:** 352
- **Covers:**
  - `hasCatalogData = false` when `err != nil && resolvedDeprecation == nil`
  - All deprecation conditions → Unknown (not False)
  - Uses installed bundle name when bundle exists

**4. TestClusterExtensionBoxcutterApplierFailsDoesNotLeakDeprecationErrors**
- **Line:** 641
- **Covers:**
  - Early return for rolling out bundles
  - Deprecation set to Unknown during rollout
  - Apply errors stay in Progressing condition only

**5. TestSetDeprecationStatus**
- **Line:** 1264
- **Covers:** (11 test cases)
  - Direct unit testing of `SetDeprecationStatus` function
  - All combinations of `hasCatalogData` and deprecation states
  - Installed vs no bundle scenarios
  - Reason values (Deprecated, DeprecationStatusUnknown, Absent)

---

### Modified Tests in This PR

#### File: `common_controller_test.go`

**TestClusterExtensionDeprecationMessageTruncation**
- **Changed:** Separator from `"; "` → `"\n"` to match production code
- **Covers:** Message truncation with newline-separated deprecation warnings

---

## Coverage Summary

| Logic Change | Test Coverage |
|--------------|---------------|
| **Use installed bundle (not resolved)** | ✅ TestClusterExtensionUpgradeShowsInstalledBundleDeprecation<br>✅ TestSetDeprecationStatus<br>✅ TestClusterExtensionResolutionFailsWithoutCatalogDeprecationData |
| **`hasCatalogData = err == nil \|\| resolvedDeprecation != nil`** | ✅ TestClusterExtensionResolutionFailsWithDeprecationData<br>✅ TestClusterExtensionResolutionFailsWithoutCatalogDeprecationData |
| **Set deprecation before error check** | ✅ TestClusterExtensionResolutionFailsWithDeprecationData<br>✅ TestClusterExtensionResolutionFailsWithoutCatalogDeprecationData |
| **Early return for rolling out bundles** | ✅ TestClusterExtensionBoxcutterApplierFailsDoesNotLeakDeprecationErrors |
| **Deprecation conditions absent when not deprecated** | ✅ TestSetDeprecationStatus (11 cases) |
| **Unknown when catalog unavailable** | ✅ TestClusterExtensionResolutionFailsWithoutCatalogDeprecationData<br>✅ TestSetDeprecationStatus |
| **Newline separator for messages** | ✅ TestClusterExtensionDeprecationMessageTruncation |

---

## Total Test Coverage

- **5 new integration/reconciliation tests**
- **11 unit test cases** in TestSetDeprecationStatus
- **1 modified test** for message formatting

**Total: 17 test cases** covering all code paths and behavioral changes.

camilamacedo86:
/hold 

We will need to change here to set False always we should always set something

camilamacedo86:
/hold cancel

Changed to keep Deprecated False as before.

perdasilva:
crd-diff failed due to changes in field description - the changes don't have no semantic impact and further clarify the deprecated condition behavior - setting override

PR #2298, opened 2025-11-03
:sparkles: OPRUN-4240: Add test profiling toolchain for heap and CPU analysis

tmshort:
✨ Add e2e profiling toolchain for heap and CPU analysis
Introduces automated memory and CPU profiling for e2e tests with:
- Automatic port-forwarding to Kubernetes deployment pprof endpoints
- Configurable periodic heap and CPU profile collection with differential timing
- Analysis report generation with growth metrics and top allocators
- Makefile targets: start-profiling, stop-profiling, analyze-profiles

## Reviewer Checklist

- [ ] API Go Documentation


hack/tools/e2e-profiling/collect-profiles.sh
```diff
@@ -0,0 +1,304 @@
...
+            exit 1
+        fi
+        sleep 2
+        ELAPSED=$((ELAPSED + 2))
+    done
+    log_success "Deployment ${DEPLOYMENT} exists"
+
+    # Now wait for deployment to become available
+    log_info "Waiting for deployment ${DEPLOYMENT} to become available..."
+    if ! kubectl wait --for=condition=Available -n "${NAMESPACE}" deployment "${DEPLOYMENT}" --timeout=300s; then
```
pedjak: we are going to wait until condition `Available` appears, but we are not checking its status, thus we might continue even if the status is set to `False`. We should actually continue only when it is set to `True`

hack/tools/e2e-profiling/collect-profiles.sh
```diff
@@ -0,0 +1,304 @@
...
+    PODS[$component]="${POD}"
+
+    # Create component output directory
+    mkdir -p "${OUTPUT_DIR}/${component}"
+
+    # Set up port forwarding
+    log_info "Setting up port forwarding to ${POD}:${PPROF_PORT} -> localhost:${LOCAL_PORT}..."
+    kubectl port-forward -n "${NAMESPACE}" "${POD}" "${LOCAL_PORT}:${PPROF_PORT}" > "${OUTPUT_DIR}/${component}/port-forward.log" 2>&1 &
+    PF_PIDS[$component]=$!
+    log_info "Port-forward started (PID: ${PF_PIDS[$component]})"
```
pedjak: we could set port-forwarding on the deployment directly, and that would simplify the script, because we do not need to search pods - deployment name is fixed.

hack/tools/e2e-profiling/collect-profiles.sh
```diff
@@ -0,0 +1,304 @@
...
+done
+
+# Wait for port forwards to be ready
+log_info "Waiting for port forwards to initialize..."
+sleep 5
```
pedjak: instead of wait here, I would try to connect to the endpoint and retry several times if I get the error.

hack/tools/e2e-profiling/collect-profiles.sh
```diff
@@ -0,0 +1,304 @@
...
+        sleep "${sleep_time}"
+    else
+        log_warn "Iteration took ${elapsed}s, longer than interval ${INTERVAL}s. Skipping sleep."
+    fi
+done
+
+log_success "Collection complete. Collected ${n} profiles from each component."
+
+# Clean up empty profiles (created during cluster teardown)
+log_info "Cleaning up empty profile files..."
```
pedjak: perhaps this should be executed also when user presses ctrl-c?

hack/tools/e2e-profiling/e2e-profile.sh
```diff
@@ -0,0 +1,129 @@
...
+    echo -e "${BLUE}[INFO]${NC} $*"
+}
+
+log_error() {
+    echo -e "${RED}[ERROR]${NC} $*"
+}
+
+log_success() {
+    echo -e "${GREEN}[SUCCESS]${NC} $*"
+}
```
pedjak: All added scripts have a share of common code. Perhaps we should extract it into something like `common.sh` and source it other scripts?

helm/olmv1/templates/deployment-olmv1-system-operator-controller-controller-manager.yml
```diff
@@ -41,6 +41,9 @@ spec:
         - args:
             - --health-probe-bind-address=:8081
             - --metrics-bind-address=:8443
+            {{- if .Values.options.e2e.enabled }}
```
pedjak: Profiling could we turned on also for other reasons, not just for e2e tests. Hence, I would suggest to name this option something like:

```
.Values.options.profiling.enabled
```

helm/olmv1/templates/deployment-olmv1-system-catalogd-controller-manager.yml
```diff
@@ -44,6 +44,9 @@ spec:
             - --leader-elect
             {{- end }}
             - --metrics-bind-address=:7443
+            {{- if .Values.options.e2e.enabled }}
```
pedjak: Profiling could we turned on also for other reasons, not just for e2e tests. Hence, I would suggest to name this option something like:

```
.Values.options.profiling.enabled
```

hack/tools/e2e-profiling/PLUGIN_SUMMARY.md
```diff
@@ -0,0 +1,321 @@
...
+Same as operator-controller project.
+
+## 🙏 Acknowledgments
+
+Based on the OpenAPI caching optimization workflow that successfully reduced memory usage by 16.9%.
+
+---
+
+**Created:** 2025-10-28
+**Status:** Production Ready
```
pedjak: Is it?

hack/tools/e2e-profiling/PLUGIN_SUMMARY.md
```diff
@@ -0,0 +1,321 @@
...
+**Version:** 1.0.0
```
pedjak: Are we gonna track the version of this separately from the released artifacts?

hack/tools/e2e-profiling/USAGE_EXAMPLES.md
```diff
@@ -0,0 +1,315 @@
...
+## Example 2: Run a New Test
+
+```bash
+# Run a new baseline test
+./hack/tools/e2e-profiling/e2e-profile.sh run new-baseline
+
+# This will:
+# 1. Start make test-experimental-e2e
+# 2. Wait for pod to be ready
+# 3. Collect profiles every 15s
```
pedjak: the code and PR description are actually speaking about 10s intervals.

hack/tools/e2e-profiling/USAGE_EXAMPLES.md
```diff
@@ -0,0 +1,315 @@
+# Usage Examples
+
+## Example 1: Analyze Existing Test Data
+
+The repository already has test data from the OpenAPI caching optimization:
+
+```bash
+# Set up symlinks to existing data
+ln -s /home/tshort/experimental-e2e-testing/test1 e2e-profiles/baseline
```
pedjak: perhaps we should make some more abstract paths instead of publishing `/home/tshort` ?

hack/tools/e2e-profiling/PLUGIN_SUMMARY.md
tmshort: Claude thinks it is. :)

hack/tools/e2e-profiling/USAGE_EXAMPLES.md
tmshort: Yup, I changed the default, and Claude didn't pick up on this.

hack/tools/e2e-profiling/USAGE_EXAMPLES.md
tmshort: Yeah, this is probably a bad example... and Claude didn't clean it up.

tmshort:
> Given that all other interactions with the code are happening through Makefile targets, it would be nice if we can continue doing that for the profiling as well, something like:
> 
> ```shell
> make baseline-profile test-experiment-e2e
> ```
> 
> or maybe
> 
> ```shell
> make test-experiment-e2e PROFILING=baseline
> ```
> 
> Similar goes for other added scripts.

I'm going to disagree here. We have a number of other utilities that can be run outside the Makefile. Also, passing arguments to Makefiles is onerous (as given by your example of having to use `PROFILING=` to specify the name of the profile).

hack/tools/e2e-profiling/collect-profiles.sh
tmshort: The `--for=condition=Available` explicitly means when condition Available is true, as `true` is the default. I would have to have appended `=false` for it to occur.

See: https://kubernetes.io/docs/reference/kubectl/generated/kubectl_wait/
```
# The default value of status condition is true; you can wait for other targets after an equal delimiter (compared after Unicode simple case folding, which is a more general form of case-insensitivity)
  kubectl wait --for=condition=Ready=false pod/busybox1
```

hack/tools/e2e-profiling/PLUGIN_SUMMARY.md
tmshort: There's no need to version this, so we won't bother trying to keep it in sync.

hack/tools/e2e-profiling/collect-profiles.sh
pedjak: Thanks, my bad.

pedjak:
> I'm going to disagree here. We have a number of other utilities that can be run outside the Makefile. Also, passing arguments to Makefiles is onerous (as given by your example of having to use `PROFILING=` to specify the name of the profile).

We should not prevent people to call the scripts directly, but having no target on Makefile level makes them harder to discover.

I would also say that profiling itself is orthogonal to running e2e tests. We could profile controllers even if do not run tests.  If we make target for starting and stopping profiling, then we could do something like:

```sh
make baseline-profiling e2e stop-profiling
```

If we want to profile without running tests, we can then do

```sh
make baseline-profiling
# play with OLM
# ...
# ...
make stop-profiling
```

hack/tools/e2e-profiling/PLUGIN_SUMMARY.md
```diff
@@ -0,0 +1,319 @@
+# E2E Profiling Plugin - Summary
+
+## 📊 What Was Created
```
anik120: Same comment as the other PR, we probably shouldn't commit this to main. The Readme.md and Usage.md seems to be doing the job.

tmshort:
> We should not prevent people to call the scripts directly, but having no target on Makefile level makes them harder to discover.
> 
> I would also say that profiling itself is orthogonal to running e2e tests. We could profile controllers even if do not run tests. If we make target for starting and stopping profiling, then we could do something like:
> 
> ```shell
> make baseline-profiling e2e stop-profiling
> ```
> 
> If we want to profile without running tests, we can then do
> 
> ```shell
> make baseline-profiling
> # play with OLM
> # ...
> # ...
> make stop-profiling
> ```

I hear your concerns. Because the scripts are expecting to run a specific e2e test, that would require a bit of refactoring. The functionality works now with the existing e2e tests. I would consider doing that as a follow up.

hack/tools/e2e-profiling/PLUGIN_SUMMARY.md
tmshort: This file is useful for Claude to know how to use the tool as a plug-in.

tmshort:
> Consider this separation:
> 
> 1. e2e profiling toolchain for heap and CPU analysis in introduced (independent of any Claude/AI stuff)
> 2. A helpful Claude command is introduced to help with the AI-ness (in a different PR)
> 
> Isn't a that more cleaner?
> 
> 1 is something we'd do traditionally (but now it's being done with assistance from AI). 2 is the new world stuff.

That's starting to sound like busy-work. I can do it, but I don't really see the benefit.
EDIT: I will also point out that having two separate PRs make it harder to test together... but that's a _me_ problem.

tmshort:
/hold
I'm updating this to work with `make`... so holding

Makefile
```diff
@@ -334,6 +334,14 @@ test-upgrade-experimental-e2e: $(TEST_UPGRADE_E2E_TASKS) #HELP Run upgrade e2e t
 e2e-coverage:
 	COVERAGE_NAME=$(COVERAGE_NAME) ./hack/test/e2e-coverage.sh
 
+.PHONY: start-profiling
+start-profiling: #HELP Start profiling in background. Run your tests, then use 'make stop-profiling'. Use PROFILE_NAME=<name> to specify output name.
+	./hack/tools/e2e-profiling/start-profiling.sh $(PROFILE_NAME)
```
pedjak: if we craft the target like:

```
start-profiling/%:
  ./hack/tools/e2e-profiling/start-profiling.sh $*
```

then we could simply the invocation with:

```sh
make start-profiling/foo
```

hack/tools/e2e-profiling/analyze-profiles.sh
```diff
@@ -0,0 +1,391 @@
...
+EOF
+
+    # OpenAPI-specific analysis
+    log_info "  Analyzing OpenAPI allocations..." >&2
+    cat >> "${REPORT_FILE}" << 'EOF'
+
+### OpenAPI-Related Allocations
+
+```
+EOF
```
pedjak: I think these three `cat` invocation (and perhaps others too) could be merged into single one:

```
cat >> "${REPORT_FILE}" << EOF
### Top Memory Allocators (Peak Profile)
` ``
$(cd "${component_dir}" && go tool pprof -top "${peak_name}" 2>/dev/null | head -20)
` ``
### OpenAPI-Related Allocations
` ``
$(cd "${component_dir}" && go tool pprof -text "${peak_name}" 2>/dev/null | grep -i openapi | head -20 >> "${REPORT_FILE}" || echo "No OpenAPI allocations found")
` ``
EOF
```
 It would mprove readability if we wrap the calls to `go tool pprof` into bash functions.

pedjak:
I did a lint pass with [ShellCheck](https://www.shellcheck.net):

```
In ./common.sh line 44:
        local dir="$(cd -P "$(dirname "$source")" && pwd)"
              ^-^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In ./common.sh line 59:
        local dir=$(dirname "$path")
              ^-^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In ./common.sh line 60:
        local base=$(basename "$path")
              ^--^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In ./common.sh line 64:
        local parent=$(dirname "$path")
              ^----^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In ./common.sh line 65:
        local base=$(basename "$path")
              ^--^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In analyze-profiles.sh line 34:
    local profile_count=$(find -L "${component_dir}" -maxdepth 1 -name "heap*.pprof" -type f 2>/dev/null | wc -l)
          ^-----------^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In analyze-profiles.sh line 43:
    local peak_profile=$(ls -lS "${component_dir}"/heap*.pprof 2>/dev/null | head -1 | awk '{print $NF}')
          ^----------^ SC2155 (warning): Declare and assign separately to avoid masking return values.
                         ^-- SC2012 (info): Use find instead of ls to better handle non-alphanumeric filenames.


In analyze-profiles.sh line 44:
    local peak_name=$(basename "${peak_profile}")
          ^-------^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In analyze-profiles.sh line 45:
    local peak_size=$(du -h "${peak_profile}" | cut -f1)
          ^-------^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In analyze-profiles.sh line 54:
    local peak_stats=$(cd "${component_dir}" && go tool pprof -top "${peak_name}" 2>/dev/null | head -6)
          ^--------^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In analyze-profiles.sh line 55:
    local peak_total=$(echo "${peak_stats}" | grep "^Showing" | awk '{print $7, $8}')
          ^--------^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In analyze-profiles.sh line 74:
    for f in $(ls "${component_dir}"/heap*.pprof 2>/dev/null | sort -V); do
               ^-- SC2012 (info): Use find instead of ls to better handle non-alphanumeric filenames.


In analyze-profiles.sh line 75:
        local name=$(basename "$f")
              ^--^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In analyze-profiles.sh line 76:
        local size=$(stat -c%s "$f")
              ^--^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In analyze-profiles.sh line 192:
    local cpu_profile_count=$(find -L "${component_dir}" -maxdepth 1 -name "cpu*.pprof" -type f 2>/dev/null | wc -l)
          ^---------------^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In analyze-profiles.sh line 201:
    local peak_cpu_profile=$(ls -lS "${component_dir}"/cpu*.pprof 2>/dev/null | head -1 | awk '{print $NF}')
          ^--------------^ SC2155 (warning): Declare and assign separately to avoid masking return values.
                             ^-- SC2012 (info): Use find instead of ls to better handle non-alphanumeric filenames.


In analyze-profiles.sh line 202:
    local peak_cpu_name=$(basename "${peak_cpu_profile}")
          ^-----------^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In analyze-profiles.sh line 203:
    local peak_cpu_size=$(du -h "${peak_cpu_profile}" | cut -f1)
          ^-----------^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In analyze-profiles.sh line 209:
    local cpu_stats=$(cd "${component_dir}" && go tool pprof -top "${peak_cpu_name}" 2>/dev/null | head -6)
          ^-------^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In analyze-profiles.sh line 210:
    local cpu_total=$(echo "${cpu_stats}" | grep "^Showing" | awk '{print $7, $8}')
          ^-------^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In analyze-profiles.sh line 385:
    peak_profile=$(ls -lS "${component_dir}"/heap*.pprof 2>/dev/null | head -1 | awk '{print $NF}')
                   ^-- SC2012 (info): Use find instead of ls to better handle non-alphanumeric filenames.


In ./common.sh line 44:
        local dir="$(cd -P "$(dirname "$source")" && pwd)"
              ^-^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In ./common.sh line 59:
        local dir=$(dirname "$path")
              ^-^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In ./common.sh line 60:
        local base=$(basename "$path")
              ^--^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In ./common.sh line 64:
        local parent=$(dirname "$path")
              ^----^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In ./common.sh line 65:
        local base=$(basename "$path")
              ^--^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In collect-profiles.sh line 51:
    source "${STATE_FILE}"
           ^-------------^ SC1090 (warning): ShellCheck can't follow non-constant source. Use a directive to specify location.


In common.sh line 44:
        local dir="$(cd -P "$(dirname "$source")" && pwd)"
              ^-^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In common.sh line 59:
        local dir=$(dirname "$path")
              ^-^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In common.sh line 60:
        local base=$(basename "$path")
              ^--^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In common.sh line 64:
        local parent=$(dirname "$path")
              ^----^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In common.sh line 65:
        local base=$(basename "$path")
              ^--^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In ./common.sh line 44:
        local dir="$(cd -P "$(dirname "$source")" && pwd)"
              ^-^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In ./common.sh line 59:
        local dir=$(dirname "$path")
              ^-^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In ./common.sh line 60:
        local base=$(basename "$path")
              ^--^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In ./common.sh line 64:
        local parent=$(dirname "$path")
              ^----^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In ./common.sh line 65:
        local base=$(basename "$path")
              ^--^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In compare-profiles.sh line 47:
PEAK1=$(ls -lS "${TEST1_DIR}"/operator-controller/heap*.pprof 2>/dev/null | head -1 | awk '{print $NF}')
        ^-- SC2012 (info): Use find instead of ls to better handle non-alphanumeric filenames.


In compare-profiles.sh line 48:
PEAK2=$(ls -lS "${TEST2_DIR}"/operator-controller/heap*.pprof 2>/dev/null | head -1 | awk '{print $NF}')
        ^-- SC2012 (info): Use find instead of ls to better handle non-alphanumeric filenames.


In compare-profiles.sh line 51:
PEAK1_CATALOGD=$(ls -lS "${TEST1_DIR}"/catalogd/heap*.pprof 2>/dev/null | head -1 | awk '{print $NF}')
                 ^-- SC2012 (info): Use find instead of ls to better handle non-alphanumeric filenames.


In compare-profiles.sh line 52:
PEAK2_CATALOGD=$(ls -lS "${TEST2_DIR}"/catalogd/heap*.pprof 2>/dev/null | head -1 | awk '{print $NF}')
                 ^-- SC2012 (info): Use find instead of ls to better handle non-alphanumeric filenames.


In compare-profiles.sh line 137:
    (ls "${TEST1_DIR}"/operator-controller/heap*.pprof 2>/dev/null | sed 's/.*heap\([0-9]*\)\.pprof/\1/';
     ^-- SC2012 (info): Use find instead of ls to better handle non-alphanumeric filenames.


In compare-profiles.sh line 138:
     ls "${TEST2_DIR}"/operator-controller/heap*.pprof 2>/dev/null | sed 's/.*heap\([0-9]*\)\.pprof/\1/') \
     ^-- SC2012 (info): Use find instead of ls to better handle non-alphanumeric filenames.


In compare-profiles.sh line 192:
    (ls "${TEST1_DIR}"/catalogd/heap*.pprof 2>/dev/null | sed 's/.*heap\([0-9]*\)\.pprof/\1/';
     ^-- SC2012 (info): Use find instead of ls to better handle non-alphanumeric filenames.


In compare-profiles.sh line 193:
     ls "${TEST2_DIR}"/catalogd/heap*.pprof 2>/dev/null | sed 's/.*heap\([0-9]*\)\.pprof/\1/') \
     ^-- SC2012 (info): Use find instead of ls to better handle non-alphanumeric filenames.


In ./common.sh line 44:
        local dir="$(cd -P "$(dirname "$source")" && pwd)"
              ^-^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In ./common.sh line 59:
        local dir=$(dirname "$path")
              ^-^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In ./common.sh line 60:
        local base=$(basename "$path")
              ^--^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In ./common.sh line 64:
        local parent=$(dirname "$path")
              ^----^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In ./common.sh line 65:
        local base=$(basename "$path")
              ^--^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In ./common.sh line 44:
        local dir="$(cd -P "$(dirname "$source")" && pwd)"
              ^-^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In ./common.sh line 59:
        local dir=$(dirname "$path")
              ^-^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In ./common.sh line 60:
        local base=$(basename "$path")
              ^--^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In ./common.sh line 64:
        local parent=$(dirname "$path")
              ^----^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In ./common.sh line 65:
        local base=$(basename "$path")
              ^--^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In run-profiled-test.sh line 138:
        TEST_EXIT=$?
        ^-------^ SC2034 (warning): TEST_EXIT appears unused. Verify use (or export if used externally).


In run-profiled-test.sh line 152:
            LATEST=$(ls -t "${OUTPUT_DIR}"/heap*.pprof 2>/dev/null | head -1)
                     ^-- SC2012 (info): Use find instead of ls to better handle non-alphanumeric filenames.


In ./common.sh line 44:
        local dir="$(cd -P "$(dirname "$source")" && pwd)"
              ^-^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In ./common.sh line 59:
        local dir=$(dirname "$path")
              ^-^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In ./common.sh line 60:
        local base=$(basename "$path")
              ^--^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In ./common.sh line 64:
        local parent=$(dirname "$path")
              ^----^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In ./common.sh line 65:
        local base=$(basename "$path")
              ^--^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In start-profiling.sh line 70:
nohup bash -c '
              ^-- SC2016 (info): Expressions don't expand in single quotes, use double quotes for that.


In ./common.sh line 44:
        local dir="$(cd -P "$(dirname "$source")" && pwd)"
              ^-^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In ./common.sh line 59:
        local dir=$(dirname "$path")
              ^-^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In ./common.sh line 60:
        local base=$(basename "$path")
              ^--^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In ./common.sh line 64:
        local parent=$(dirname "$path")
              ^----^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In ./common.sh line 65:
        local base=$(basename "$path")
              ^--^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In stop-profiling.sh line 52:
source "${STATE_FILE}"
       ^-------------^ SC1090 (warning): ShellCheck can't follow non-constant source. Use a directive to specify location.

For more information:
  https://www.shellcheck.net/wiki/SC1090 -- ShellCheck can't follow non-const...
  https://www.shellcheck.net/wiki/SC2034 -- TEST_EXIT appears unused. Verify ...
  https://www.shellcheck.net/wiki/SC2155 -- Declare and assign separately to ...
````

File: hack/tools/e2e-profiling/start-profiling.sh [hunk 15]
```diff
@@ -0,0 +1,347 @@
...
+export PROFILE_NAMESPACE="${NAMESPACE}"
+export PROFILE_INTERVAL="${INTERVAL}"
+export PROFILE_OUTPUT_DIR="${OUTPUT_DIR}"
+export PROFILE_CPU_DURATION="${CPU_PROFILE_DURATION}"
+export PROFILE_MODE_VAR="${PROFILE_MODE}"
+export PROFILE_STATE_FILE="${STATE_FILE}"
+
+# Start everything in background
+nohup bash -c '
+set -euo pipefail
```
pedjak: could we move this into a separare script instead keeping it in this large string?

tmshort:
> I did a lint pass with [ShellCheck](https://www.shellcheck.net/):

We don't use ShellCheck for anything... so that is a requirement that has not been placed on anything else. I don't think it's appropriate to apply it on this and not anything else. In other words, if we want to use ShellCheck, we should create a verify target for it. It does pass a `bash -n` check.

However, I'll look at some of those issues.

hack/tools/e2e-profiling/analyze-profiles.sh
tmshort: Can you please confirm your requsted change? I don't think it would work as given.

helm/olmv1/templates/deployment-olmv1-system-operator-controller-controller-manager.yml
```diff
@@ -41,6 +41,9 @@ spec:
         - args:
             - --health-probe-bind-address=:8081
             - --metrics-bind-address=:8443
+            {{- if .Values.options.profiling.enabled }}
+            - --pprof-bind-address=:6060
+            {{- end }}
```
camilamacedo86, line 46: 👍 Liked we do not do that by default
Perfect

hack/tools/e2e-profiling/README.md
```diff
@@ -0,0 +1,759 @@
...
+- [x] Shared common library for code reuse
+- [x] Automatic cleanup of empty profile files
+
+**Future Enhancements:**
+- [ ] Add goroutine profiling
+- [ ] Support multiple pods (replicas)
+- [ ] Add real-time dashboard
+- [ ] Support different output formats (JSON, CSV)
+- [ ] Add mutex profiling
+- [ ] Support custom component configurations
```
camilamacedo86: I think it all is generated by AI
Could we just shape it and have only what is required here, making sense?

hack/tools/e2e-profiling/analyze-profiles.sh
pedjak: > Can you please confirm your requsted change? I don't think it would work as given.

It is a nit, but IMHO is more readable. See this little example:

```bash
function bar() {
  echo "hello world"
}
cat << EOF
first line
$(bar)
second line
EOF
```

pedjak:
general comment/question: the introduced logic is not so trivial, and I wonder if we would get on maintainability if it is written in some other more suited language than bash? (not pushing for anything particular here)

tmshort:
> general comment/question: the introduced logic is not so trivial, and I wonder if we would get on maintainability if it is written in some other more suited language than bash? (not pushing for anything particular here)

If it's non-trivial in bash, it's likely non-trivial in any other language. Given how far this has progressed, I'm reluctant to switch to another language and take even longer to get this merged. I've also tried to minimize the number of required tools, and doing it in anything other than Bash or Go will add new build environment requirements. As of now, nothing new is required to run this.

That being said... I can look to see what would happen if it were written in golang, but I don't want that to hold up this PR.

hack/tools/e2e-profiling/README.md
tmshort: I've reduced it.

hack/tools/e2e-profiling/analyze-profiles.sh
tmshort: I've cleaned this up a bit, hopefully more readable.

pedjak (approved):
/lgtm thanks!

helm/olmv1/values.yaml
```diff
@@ -24,6 +24,8 @@ options:
     enabled: false
   e2e:
     enabled: false
+  profiling:
+    enabled: false
```
camilamacedo86, line 28: 👍 for stable disabled by default great

hack/tools/e2e-profiling/README.md
```diff
@@ -0,0 +1,234 @@
...
+- `run-profiled-test.sh` - Orchestrate test + profiling
+- `analyze-profiles.sh` - Generate analysis reports
+- `compare-profiles.sh` - Create comparison reports
+- `common.sh` - Shared utilities and logging
+
+**Key Features:**
+- Deployment-based port-forwarding (survives pod restarts)
+- Automatic retry with 30s timeout
+- Graceful cleanup on exit/interrupt
+- Multi-component support (operator-controller + catalogd)
```
camilamacedo86:
```suggestion
```

I think we can remove this section — what do you think?
It looks like it might have been AI-generated and doesn’t seem necessary for our use case.
It also feels a bit verbose and not quite aligned with the writing style we usually use for these READMEs.

.gitignore
```diff
@@ -38,8 +38,13 @@ vendor/
...
 
-# AI temp files files
-.claude/
+# AI temp/local files
+.claude/settings.local.json
+.claude/history/
+.claude/cache/
+.claude/logs/
+.claude/.session*
+.claude/*.log
```
camilamacedo86: Could not be 
```
# Ignore Claude local settings, history, and logs
.claude/
```

Also, it seems out of scope here
Are we adding any claude implementation?

camilamacedo86 (commented):
Hey @tmshort 

I added just a few nits.
I think the README we need to shape 
AI will add info not really required and or not helpful. 
Often too verbose, so I think it requires some human adjustments.
PS.: I did not test it out locally. 

Beyond that I am ok with we add it.
@dtfranz worked in the performance checks so I think it would be nice get his review.

hack/tools/e2e-profiling/README.md
tmshort: I think the level is fine. It was definitely reduced, and can always be edited back later. Generally, for _documentation_, more detail is better.

.gitignore
tmshort: This is what Claude said to include in the .gitignore; it knows what should be there! :)

hack/tools/e2e-profiling/README.md
anik120: This is actually a great example of how it's helpful to generate docs like this to provide additional context. 

>and not quite aligned with the writing style we usually use for these READMEs.

Writing styles before were plagued with **not-enough context** syndrome, which made contributors reliant on person/people who wrote the code to understand context (tribal knowledge). 

I personally saw this file while doing my first "eye over" of the PR, and was grateful to get more context from this file about what to expect from this PR (just like how reading the summary of a research paper helps understand the rest of the paper)

.gitignore
anik120: @tmshort this does feel a little out of place tbh. Everything else is looking great.

tmshort:
OK, I've converted this into golang (vs. bash) and generalized it a bit more:
* different namespaces for catd and opcon
* renamed from `e2e-profiles` to make it more generic
* Reduced the verbiage of the README.md (sorry, @anik120)
* Removed AI-specific stuff (specifically around .gitignore)

hack/tools/test-profiling/README.md
```diff
@@ -0,0 +1,87 @@
...
+```
+
+## Requirements
+
+**Runtime:**
+- Kubernetes cluster access (via kubeconfig)
+
+**Build:**
+- go 1.24+
+- make
```
camilamacedo86: Will be hard we keep it maintained 
when we bump do we really need this section?

tmshort: I'm looking to remove it, and make other improvements... this is becoming an HnH project today,...

camilamacedo86 (approved):

PR #2309, opened 2025-11-05
:sparkles: OPRUN-4240: Add Claude Code integration for test profiling

tmshort:
Add `/test-profile` slash command to enable interactive profiling workflow through Claude Code interface.

Follow-on to #2298 

# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/.claude/commands/test-profile.md
+++ b/.claude/commands/test-profile.md
@@ -1,0 +1,98 @@
+---
+description: Profile memory and CPU usage during e2e tests and analyze results
+---
+
+# Test Profiling
+
+Profile memory and CPU usage during e2e tests using the Go-based `bin/test-profile` tool.
+
+## Available Commands
+
+### /test-profile start [name]
+Start background profiling daemon. Collects heap/CPU profiles from operator-controller and catalogd every 10s.
+- Auto-stops after 3 consecutive collection failures (e.g., cluster teardown)
+- Use with any test command: `make test-e2e`, `make test-experimental-e2e`, etc.
+- Follow with `/test-profile stop` to generate analysis
+
+**Example:**
+```bash
+/test-profile start baseline
+make test-e2e
+/test-profile stop
+```
+
+### /test-profile stop
+Stop profiling daemon and generate analysis report. Cleans up port-forwards and empty profiles.
+
+### /test-profile run [name] [test-target]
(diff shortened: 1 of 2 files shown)


camilamacedo86 (changes requested):
**I think this is something we should discuss before merging.**
Personally, I’d prefer that we keep our instructions AI-agnostic, rather than tied to a specific vendor or model like Claude.

For example, instead of adding a CLAUDE.md, we could introduce something like an AGENTS.md file that describes generic AI/agent usage and workflows. Then, if needed, we could create aliases or vendor-specific adapters for Claude, ChatGPT, etc. — without locking ourselves to one ecosystem.

Also, just to confirm: can every contributor actually use Claude?
Do all contributors have access to it without a paid plan?
If not, that might make a vendor-specific setup harder to maintain or exclude contributors unintentionally.

camilamacedo86 (approved):
Based on the discussion: https://kubernetes.slack.com/archives/C04066HP28K/p1762846782356659?thread_ts=1762534450.554899&cid=C04066HP28K

I think we can move forward and after we see if we can make it work with others AI toolings as well.

tmshort:
/unhold
Based on the slack thread.
I have also updated this to be more precise. @camilamacedo86

.claude/commands/test-profile.md
```diff
@@ -0,0 +1,98 @@
...
+
+- Source: `hack/tools/test-profiling/` (Go-based CLI)
+- Binary: `bin/test-profile`
+- Make targets: `make build-test-profiler`, `make start-profiling/[name]`, `make stop-profiling`
+
+## Requirements
+
+- kubectl access to cluster
+- go tool pprof (for analysis)
+- Go 1.22+ (for building)
```
camilamacedo86: @tmshort ^ is that true
How will we keep it updated? 
Should not minimal go requirement same of go.mod ?

tmshort: I changed it to be a reference to the appropriate `go.mod`

camilamacedo86 (approved):

camilamacedo86:
Great 👍 
I think it is fine we can improve in the long run as well.

It is amazing see our first command instruction 🎉 
Well done.

PR #2311, opened 2025-11-06
:seedling: Add health probes to test-operator bundles

pedjak:
# Description

Enhance test-operator v1.0.0 and v1.2.0 bundles with HTTP-based health probes
to enable testing of startup, liveness, and readiness probe behavior. Replaces
the simple sleep container with a busybox httpd server that exposes probe
endpoints via a ConfigMap-mounted script.

The probes can be flipped from true to false by executing `rm /var/www{stared,ready,live}`
and back to true by performin `touch /var/www{stared,ready,live}`

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/testdata/images/bundles/test-operator/v1.0.0/manifests/testoperator.clusterserviceversion.yaml
+++ b/testdata/images/bundles/test-operator/v1.0.0/manifests/testoperator.clusterserviceversion.yaml
@@ -89,7 +89,7 @@ spec:
                         port: 80
                       initialDelaySeconds: 1
                       periodSeconds: 1
-                    serviceAccountName: simple-bundle-manager
+                serviceAccountName: simple-bundle-manager
       clusterPermissions:
         - rules:
             - apiGroups:
--- a/testdata/images/bundles/test-operator/v1.2.0/manifests/testoperator.clusterserviceversion.yaml
+++ b/testdata/images/bundles/test-operator/v1.2.0/manifests/testoperator.clusterserviceversion.yaml
@@ -149,7 +149,7 @@ spec:
               verbs:
                 - create
                 - patch
-          serviceAccountName: simple-bundle-manager
+      serviceAccountName: simple-bundle-manager
     strategy: deployment
   installModes:
     - supported: false


testdata/images/bundles/test-operator/v1.0.0/manifests/testoperator.clusterserviceversion.yaml
```diff
@@ -56,15 +56,39 @@ spec:
...
+                        port: 80
+                      failureThreshold: 1
+                      periodSeconds: 2
+                    readinessProbe:
+                      httpGet:
+                        path: /ready
+                        port: 80
+                      initialDelaySeconds: 1
+                      periodSeconds: 1
                     serviceAccountName: simple-bundle-manager
```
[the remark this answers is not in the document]
pedjak: Incorrect, all e2e tests are passing.

camilamacedo86 (changes requested):
Hi @pedjak 👋  

Thank you for the contribution!  

I’m not sure what problem this change solves — could you please clarify?  

- How does it help us test OLM features?  

Otherwise, I’d prefer to keep the samples simple and focused only on what’s needed to test OLM.

testdata/images/bundles/test-operator/v1.0.0/manifests/script.configmap.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: ConfigMap
+metadata:
+  name: httpd-script
+data:
+  httpd.sh: |
+    #!/bin/sh
+    echo true > /var/www/started
+    echo true > /var/www/ready
+    echo true > /var/www/live
+    exec httpd -f -h /var/www -p 80
```
camilamacedo86, line 11: **TL;DR (just to share):**

- Common case of content managed by OLM is scaffolded with **Operator SDK/Kubebuilder**, which use **controller-runtime**.  
- We already have a project used in downstream tests that we maintain — [test-operator](https://github.com/operator-framework/test-operator) — which includes health and ready probes. See the default implementation: https://github.com/operator-framework/test-operator/blob/main/cmd/main.go#L238-L245 ( controller runtime )

If there’s a strong reason to add probe implementations here (which doesn’t seem to be the case right now), I’d prefer to rely on the existing setup — meaning we can create a bundle using [test-operator](https://github.com/operator-framework/test-operator).  

Otherwise, adding probes here feels unnecessary. I’d suggest using the existing **test-operator** instead.

testdata/images/bundles/test-operator/v1.0.0/manifests/testoperator.clusterserviceversion.yaml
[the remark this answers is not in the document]
camilamacedo86: The fact that the tests are passing doesn’t necessarily mean the implementation is correct.
It depends on what the end-to-end (E2E) tests are actually verifying with respect to this test data.

Also, note that the serviceAccountName should not be defined inside the container specification.
You can refer to this example for the correct structure:
https://github.com/operator-framework/operator-sdk/blob/master/testdata/go/v4/memcached-operator/bundle/manifests/memcached-operator.clusterserviceversion.yaml#L198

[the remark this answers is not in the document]
perdasilva: I'll check this out and poke at it. It seems weird that the e2es would pass. But, I guess they are only checking for the installed status and maybe the underlying helm operation succeeded while the operator is itself it bung?

testdata/images/bundles/test-operator/v1.0.0/manifests/script.configmap.yaml
perdasilva: I don't think we ever reached a steady state on the test-operator or use it within our e2es. I'm actually wondering if we should nuke it for now.

The nice thing about having the health and readiness probes defined in this way is that we can very easily change their state at runtime by creating/deleting the files in /var/www, which is great for e2e testing. This is the real value added here by this PR.

camilamacedo86: That might be a true point. However, unless we add tests that actually take advantage of it, I don’t see a strong reason to include it.

Could we add some unit tests or use case scenarios that demonstrate where this feature would be helpful? If so, I’d be completely on board with it — that would show not only that this change is useful, but also that it works as expected.

perdasilva: These will come when we start to lay down more boxcutter tests. This was added to help with manual testing of the conditions changes we've been introducing. I think it's ok to add this now, to reduce the size of PRs later.

testdata/images/bundles/test-operator/v1.0.0/manifests/testoperator.clusterserviceversion.yaml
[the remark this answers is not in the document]
perdasilva: yeah - the pod is getting pushed onto the default service account:

```
$ kubectl get pods -n test-operator -o yaml | grep -i serviceaccount                                                                                                                                                                                                                                                      
          - mountPath: /var/run/secrets/kubernetes.io/serviceaccount
    serviceAccount: default
    serviceAccountName: default
        - serviceAccountToken:
      - mountPath: /var/run/secrets/kubernetes.io/serviceaccount
```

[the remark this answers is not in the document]
perdasilva: I've updated the resources:

```
$  kubectl get pods -n test-operator -o yaml | grep -i serviceaccount                                                                                                                                                                                                                                                          
    serviceAccount: simple-bundle-manager
    serviceAccountName: simple-bundle-manager
   ...
```

testdata/images/bundles/test-operator/v1.0.0/manifests/script.configmap.yaml
camilamacedo86: If you want to move forward, I’m okay with that.
But I think we should properly validate that:

- a) the configuration here is correct,
- b) it provides the flexibility we want, and
- c) it’s actually useful in practice (and not just theoretical).

I’d prefer to have at least one test using it, but I won’t block this one! 👍

perdasilva: :thinking: lemme see what I can do

PR #2314, opened 2025-11-08
🐛 fix: boxcutter logging levels

camilamacedo86:
- Use Error level for all error conditions (validation, collision, engine failures)
- Move verbose success reports to V(1) debug mode
- Include full diagnostic reports when errors occur

c/c @jianzhangbjz @Xia-Zhao-rh

### DIFF
--- a/internal/operator-controller/controllers/clusterextensionrevision_controller.go
+++ b/internal/operator-controller/controllers/clusterextensionrevision_controller.go
@@ -149,6 +149,11 @@ func (c *ClusterExtensionRevisionReconciler) reconcile(ctx context.Context, rev 
 
 	rres, err := c.RevisionEngine.Reconcile(ctx, *revision, opts...)
 	if err != nil {
+		if rres != nil {
+			l.Error(err, "revision reconcile failed", "report", rres.String())
+		} else {
+			l.Error(err, "revision reconcile failed")
+		}
 		meta.SetStatusCondition(&rev.Status.Conditions, metav1.Condition{
 			Type:               ocv1.ClusterExtensionRevisionTypeAvailable,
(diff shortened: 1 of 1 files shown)


internal/operator-controller/controllers/clusterextensionrevision_controller.go
```diff
@@ -149,6 +149,11 @@ func (c *ClusterExtensionRevisionReconciler) reconcile(ctx context.Context, rev
 
 	rres, err := c.RevisionEngine.Reconcile(ctx, *revision, opts...)
 	if err != nil {
+		if rres != nil {
+			l.Error(err, "revision reconcile failed", "report", rres.String())
+		} else {
+			l.Error(err, "revision reconcile failed")
+		}
```
perdasilva, line 156: I don't know if it's much of an improvement, but if we want to eliminate the duplication here we could do something like

```suggestion
        var kvs []string
		if rres != nil {
		    kvs = append(kvs, "report", rres.String())
		}
		l.Error(err, "revision reconcile failed", kvs...)
```
or something like that

camilamacedo86: Yep, we can done.

camilamacedo86:
@perdasilva all done.
So, I think this small one can fly :-)

internal/operator-controller/controllers/clusterextensionrevision_controller.go
camilamacedo86: We need keep the explicity approach otehrwise will fail in the lint; https://github.com/operator-framework/operator-controller/actions/runs/19230041339/job/54966203269?pr=2314

PR #2315, opened 2025-11-09
⚠️ (Boxcutter Runtime) Revision discovery now uses label-based cache lookups, removing Spec.Previous and preventing orphaned revisions on delete.

camilamacedo86:
Revisions are now discovered using **label-based cache lookups** instead of explicit `Spec.Previous` chains.

This makes lookups **faster** and ensures **proper cleanup** when a ClusterExtension is deleted.

## What Changed

**API (Breaking - OK in experimental mode):**
- Removed `Spec.Previous` field from ClusterExtensionRevision
- Removed `ClusterExtensionRevisionPrevious` type

**Discovery:**
- Revisions queried by label `olm.operatorframework.io/owner` via TrackingCache
- Filtered in-memory (skip current, archived, deleting, higher revision numbers)

**Migration:**
- Sets `ownerReferences` so revisions are garbage collected with their ClusterExtension
- Uses natural reconciliation: sets `Available=Unknown`, lets revision controller verify cluster state
- Re-fetches after Create() to get accurate Generation for ObservedGeneration

**Garbage Collection:**
- Renamed function and constant for clarity (`garbageCollectOldRevisions`, `ClusterExtensionRevisionRetentionLimit`)
- Still enforces 5 revision limit for archived revisions

## Why It's Better

**Performance:**
- Uses TrackingCache for label-based lookups instead of API calls
- Single query replaces chain walking

**Reliability:**
- Deleting ClusterExtension auto-deletes all its revisions via ownerReferences
- No orphaned revisions left behind

**Simplicity:**
- Removed explicit chain tracking from API
- Revision numbers provide self-evident ordering

## Migration Behavior

When upgrading from Helm storage:
1. Migration creates revision with `Available=Unknown` status
2. ClusterExtension briefly shows as not installed
3. Revision controller reconciles, verifies cluster state via probes
4. Once healthy, sets `Available=True` and `Succeeded=True`
5. ClusterExtension converges to installed state

## Demo

[asciicast](https://asciinema.org/a/qupccwlIdeprj8VvjF7OOko9t)


internal/operator-controller/applier/boxcutter.go
```diff
@@ -140,30 +143,57 @@ func (r *SimpleRevisionGenerator) buildClusterExtensionRevision(
 	ext *ocv1.ClusterExtension,
 	annotations map[string]string,
 ) *ocv1.ClusterExtensionRevision {
+	revisionLabels := map[string]string{
+		controllers.ClusterExtensionRevisionOwnerLabel: ext.Name,
+	}
+	// Package name must be in labels for BoxcutterRevisionStatesGetter to find it
+	if pkg, ok := annotations[labels.PackageNameKey]; ok {
+		revisionLabels[labels.PackageNameKey] = pkg
+	}
```
camilamacedo86: **Why:**
- Main already reads `Package` from `rev.Labels["package-name"]` in BoxcutterRevisionStatesGetter
- But main doesn't set it during migration - Package field was always empty
- This fixes the bug by copying package name from annotations to labels

internal/operator-controller/controllers/clusterextensionrevision_controller.go
```diff
@@ -428,14 +441,53 @@ func (c *ClusterExtensionRevisionReconciler) removeFinalizer(ctx context.Context
...
+			continue
+		}
+		// Skip archived or deleting revisions
+		if r.Spec.LifecycleState == ocv1.ClusterExtensionRevisionLifecycleStateArchived ||
+			!r.DeletionTimestamp.IsZero() {
+			continue
+		}
+		// Only archive revisions with lower revision numbers (actual previous revisions)
+		if r.Spec.Revision >= rev.Spec.Revision {
+			continue
```
camilamacedo86, line 485: We still populate `Spec.Previous` even though we now use labels for discovery because:

- 1. **API Compatibility** - It's a published field in the CRD and marked immutable. Removing it would be a breaking change.
- 2. **Garbage Collection** - `setPreviousRevisions()` uses it to enforce the 5 revision limit and delete older archived revisions.
- 3. **Audit Trail** - Provides a historical record of the revision chain, useful for debugging and understanding upgrade history.

internal/operator-controller/applier/boxcutter.go
```diff
@@ -270,6 +330,12 @@ func (bc *Boxcutter) apply(ctx context.Context, contentFS fs.FS, ext *ocv1.Clust
 		case err == nil:
 			// inplace patch was successful, no changes in phases
 			state = StateUnchanged
+			// Re-fetch the revision to get updated metadata and status after the patch
+			updated := &ocv1.ClusterExtensionRevision{}
+			if err := bc.Client.Get(ctx, client.ObjectKey{Name: currentRevision.Name}, updated); err != nil {
+				return false, "", fmt.Errorf("fetching updated revision: %w", err)
+			}
+			currentRevision = updated
```
camilamacedo86: After in-place patch, `currentRevision` pointed to stale object with old annotations.
It is a bug fixes required to allow we move forward here and pass on the tests. 
The scenario where it would fail is: v1.0.0 → v1.0.1 with identical bundle (only metadata changes)

internal/operator-controller/controllers/clusterextensionrevision_controller.go
```diff
@@ -428,14 +441,53 @@ func (c *ClusterExtensionRevisionReconciler) removeFinalizer(ctx context.Context
...
-		previous = append(previous, prev)
+// ListPreviousRevisions returns active revisions belonging to the same ClusterExtension with lower revision numbers.
+//
+// Scenario: Given a ClusterExtensionRevision, find all related active previous revisions
+//   - When the revision has no owner label, return empty list (revision not properly labeled)
+//   - When looking up by owner label, filter out the current revision itself
+//   - When a revision is archived or being deleted, exclude it from results
+//   - When a revision has a higher or equal revision number, exclude it (not a previous revision)
+//   - Otherwise include active and paused revisions with matching owner
+func (c *ClusterExtensionRevisionReconciler) ListPreviousRevisions(ctx context.Context, rev *ocv1.ClusterExtensionRevision) ([]client.Object, error) {
```
camilamacedo86: Added `ListPreviousRevisions(ctx, rev)` that queries revisions by owner label and filters:
- Same ClusterExtension owner
- Revision number < current (only actual previous revisions)
- Active or paused (not archived)

**Why the revision number filter matters:**  
Without it, Revision-1 completing would archive Revision-2, causing "revision is archived" errors.

internal/operator-controller/applier/boxcutter.go
```diff
@@ -195,9 +225,39 @@ func (m *BoxcutterStorageMigrator) Migrate(ctx context.Context, ext *ocv1.Cluste
 		return err
 	}
 
+	// Set ownerReference to ensure proper lifecycle management and garbage collection.
+	// This makes the migrated revision consistent with revisions created by normal Boxcutter flow.
+	if err := controllerutil.SetControllerReference(ext, rev, m.Scheme); err != nil {
+		return fmt.Errorf("set ownerref: %w", err)
+	}
```
camilamacedo86, line 259: @tmshort That should address the JIRA you added: https://issues.redhat.com/browse/OPRUN-4038


Once this PR is merged, I plan to extend the test coverage to validate the orphan scenario, ensuring we’re not leaving any orphans behind. If we identify any, we’ll handle them properly in a follow-up PR.

internal/operator-controller/applier/boxcutter.go
```diff
@@ -140,30 +143,57 @@ func (r *SimpleRevisionGenerator) buildClusterExtensionRevision(
...
 			Phases: PhaseSort(objects),
 		},
 	}
 }
 
+// BoxcutterStorageMigrator migrates ClusterExtensions from Helm-based storage to Boxcutter's
+// ClusterExtensionRevision storage.
+//
+// Scenario: Given a ClusterExtension, determine if migration from Helm storage is needed
+//   - When ClusterExtensionRevisions already exist, no migration needed (return early)
```
perdasilva:
```suggestion
//   - When ClusterExtensionRevisions already exists, no migration needed (return early)
```

internal/operator-controller/applier/boxcutter.go
```diff
@@ -195,9 +225,39 @@ func (m *BoxcutterStorageMigrator) Migrate(ctx context.Context, ext *ocv1.Cluste
...
 	}
+
+	// Mark the migrated revision as succeeded since the Helm release was already deployed.
+	// Without Succeeded=True, BoxcutterRevisionStatesGetter treats it as RollingOut instead of Installed,
+	// which breaks resolution and prevents upgrades.
+	// Note: After Create(), rev.Generation is populated by the API server (typically 1 for new objects)
+	meta.SetStatusCondition(&rev.Status.Conditions, metav1.Condition{
+		Type:               ocv1.ClusterExtensionRevisionTypeAvailable,
+		Status:             metav1.ConditionTrue,
+		Reason:             ocv1.ClusterExtensionRevisionReasonAvailable,
```
perdasilva: I wonder if we should set this to `Unknown` since we haven't observed the probes yet

internal/operator-controller/controllers/clusterextensionrevision_controller.go
perdasilva: Do we need to export this method?

internal/operator-controller/controllers/clusterextensionrevision_controller.go
```diff
@@ -428,14 +441,53 @@ func (c *ClusterExtensionRevisionReconciler) removeFinalizer(ctx context.Context
...
+		r := &revList.Items[i]
+		if r.Name == rev.Name {
+			continue
+		}
+		// Skip archived or deleting revisions
+		if r.Spec.LifecycleState == ocv1.ClusterExtensionRevisionLifecycleStateArchived ||
+			!r.DeletionTimestamp.IsZero() {
+			continue
+		}
+		// Only archive revisions with lower revision numbers (actual previous revisions)
```
perdasilva: This comment might be confusing as it mentions archival but this is a List function. Just a thought...

internal/operator-controller/controllers/clusterextensionrevision_controller.go
perdasilva: I'm not sure about this one. I think we can violate `i` since we're still in experimental mode and breaking changes are allowed, we should probably update `ii` to use the List function anyway, and lastly, revisions are numbered, so the chain should be self-evident, no?

internal/operator-controller/applier/boxcutter.go
```diff
@@ -195,9 +225,39 @@ func (m *BoxcutterStorageMigrator) Migrate(ctx context.Context, ext *ocv1.Cluste
...
+	// This makes the migrated revision consistent with revisions created by normal Boxcutter flow.
+	if err := controllerutil.SetControllerReference(ext, rev, m.Scheme); err != nil {
+		return fmt.Errorf("set ownerref: %w", err)
+	}
+
 	if err := m.Client.Create(ctx, rev); err != nil {
 		return err
 	}
+
+	// Mark the migrated revision as succeeded since the Helm release was already deployed.
```
perdasilva: Maybe it's ok if it treats it as rolling out and sets its own status? It would just be another round of reconciliation until the ClusterExtension converges back to installed, no?

internal/operator-controller/controllers/clusterextensionrevision_controller.go
camilamacedo86: Perfect! I fixed it. 🚀

internal/operator-controller/applier/boxcutter.go
camilamacedo86: I think it is a good point. We can enhance this one and add a new reason to make it more properly accurate and honest. Done 👍 Feel free to give a look.

internal/operator-controller/applier/boxcutter.go
camilamacedo86: Yep, it think we might can simplify.
Done. Let me know if is that you were thinking about.

internal/operator-controller/controllers/clusterextensionrevision_controller.go
camilamacedo86: No, we do not need. 
Changed it and refac the code.

api/v1/clusterextensionrevision_types.go
```diff
@@ -66,10 +66,6 @@ type ClusterExtensionRevisionSpec struct {
 	// +listMapKey=name
 	// +optional
 	Phases []ClusterExtensionRevisionPhase `json:"phases,omitempty"`
-	// Previous references previous revisions that objects can be adopted from.
-	//
-	// +kubebuilder:validation:XValidation:rule="self == oldSelf", message="previous is immutable"
-	Previous []ClusterExtensionRevisionPrevious `json:"previous,omitempty"`
```
camilamacedo86, line 72: Removed we are using labels now.

api/v1/clusterextensionrevision_types.go
```diff
@@ -129,13 +125,6 @@ const (
 	CollisionProtectionNone CollisionProtection = "None"
 )
 
-type ClusterExtensionRevisionPrevious struct {
-	// +kubebuilder:validation:Required
-	Name string `json:"name"`
-	// +kubebuilder:validation:Required
-	UID types.UID `json:"uid"`
-}
```
camilamacedo86, line 138: Removed we are using labels now.

internal/operator-controller/controllers/clusterextensionrevision_controller.go
camilamacedo86: I think you are 100% right and we can keep things simple here by removing it.

camilamacedo86:
@perdasilva all addressed here now 
Please feel free to give a look.

internal/operator-controller/applier/boxcutter.go
```diff
@@ -87,6 +87,9 @@ func (r *SimpleRevisionGenerator) GenerateRevisionFromHelmRelease(
 	})
 	rev.Name = fmt.Sprintf("%s-1", ext.Name)
 	rev.Spec.Revision = 1
+	// Explicitly set LifecycleState to Active to ensure migrated revision is recognized as active
+	// even before CRD default is applied by the API server
+	rev.Spec.LifecycleState = ocv1.ClusterExtensionRevisionLifecycleStateActive
```
pedjak: if we set it explicitly now, can we remove then defaulting?

internal/operator-controller/applier/boxcutter.go
pedjak: should not we fix BoxcutterRevisionStatesGetter instead?

internal/operator-controller/applier/boxcutter.go
pedjak: IMO this change could extracted into a separate PR.

internal/operator-controller/applier/boxcutter.go
```diff
@@ -242,9 +258,35 @@ func (m *BoxcutterStorageMigrator) Migrate(ctx context.Context, ext *ocv1.Cluste
...
+	// Set ownerReference for proper garbage collection when the ClusterExtension is deleted.
+	if err := controllerutil.SetControllerReference(ext, rev, m.Scheme); err != nil {
+		return fmt.Errorf("set ownerref: %w", err)
+	}
+
 	if err := m.Client.Create(ctx, rev); err != nil {
 		return err
 	}
+
+	// Re-fetch the created revision to ensure we have the latest Generation from the API server.
```
pedjak: does this mean that we use non-cache based client here? If yes, let's try to avoid it - it is going to make more pressure on api server.

pedjak (commented):
It seems that this PR is more than what PR title says: it is not about introducing just labels for tracking revisions belonging to a cluster extension - it also adds additional fixes. Could we extract those into separate PRs for easier review and handling?

camilamacedo86:
>It seems that this PR is more than what PR title says: it is not about introducing just labels for tracking revisions belonging to a cluster extension - it also adds additional fixes. Could we extract those into separate PRs for easier review and handling?


After add the tests we found bugs
So, to move forward it was mandory solve nit things

internal/operator-controller/applier/boxcutter.go
camilamacedo86: We should keep the re-fetch. This isn’t about cache performance — it's to ensure we’re updating the latest version of the resource. Between reading the object and updating it, things may change, and without re-fetching the update can fail with a conflict error.

This is a common pattern in Kubernetes controllers (see the Kubebuilder example: 
https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v4-with-plugins/internal/controller/busybox_controller.go#L109-L118). It might clarifies things up.  

**Re-fetching before an update avoids unnecessary reconcile retries and noisy logs, so it's simply good practice.**
PS. Indeed [Copilot](https://github.com/operator-framework/operator-controller/pull/2315#discussion_r2533621772) suggested it up.

camilamacedo86:
HI @pedjak 

I do not know why your comments is not done in a way that allows reply as a thread.
Anyway, 

Regards:
 
```
	// This makes the migrated revision consistent with revisions created by normal Boxcutter flow.
	if err := controllerutil.SetControllerReference(ext, rev, m.Scheme); err != nil {
		return fmt.Errorf("set ownerref: %w", err)
	}

```

> IMO this change could extracted into a separate PR.

No we cannot. We MUST to do that.
Otherwise we cannot se the Owner Ref and then we cannot achieve the goal of the PR.

internal/operator-controller/applier/boxcutter.go
camilamacedo86: That was a good point. 
I re-checked this one and I think the best option is  added explicit LifecycleState in buildClusterExtensionRevision and remove here. Done 👍

internal/operator-controller/applier/boxcutter.go
camilamacedo86: We should always use SetControllerReference when the controller creates or manages a resource, since that’s the standard Kubernetes way to ensure correct ownership and GC.

The new model relies on ownerReferences now that Spec.Previous is gone. If we don’t set the controller reference here, the API and the implementation don’t align—we say revisions are owned and GC’d, but new ones wouldn’t be. Splitting this into another PR would create a moment where discovery is label-based but GC is incomplete for new revisions, which isn’t ideal. (IHMO)

I hope that makes sense — but if you strongly disagree, and it is break deal for we move forward here. I am happy to revisit it.

internal/operator-controller/applier/boxcutter.go
pedjak: ClusterExtensionRevision `.spec` is immutable after creation, so the generation cannot be changed. Even if the generation is changed somehow, we would save here nothing because that change would trigger another reconciliation round.

Hence, let's remove this from PR.

internal/operator-controller/controllers/clusterextension_controller.go
```diff
@@ -553,7 +553,7 @@ func (d *BoxcutterRevisionStatesGetter) GetRevisionStates(ctx context.Context, e
 		//   is fairly decoupled from this code where we get the annotations back out. We may want to co-locate
 		//   the set/get logic a bit better to make it more maintainable and less likely to get out of sync.
 		rm := &RevisionMetadata{
-			Package: rev.Labels[labels.PackageNameKey],
+			Package: rev.Annotations[labels.PackageNameKey],
```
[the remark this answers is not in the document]
camilamacedo86: We have an inconsistency, but I will let it be fixed in a follow-up.

internal/operator-controller/applier/boxcutter.go
camilamacedo86: You're absolutely right - the re-fetch is unnecessary here
Create() already gives us the correct ResourceVersion
Conflict errors won't happen in this migration scenario
Even if they did, the re-fetch wouldn't prevent them
The code works perfectly without the re-fetch!

pedjak:
Given there is no user-facing changes, I would change PR type to :seedling:

camilamacedo86:
> Given there is no user-facing changes, I would change PR type to 🌱

We’re changing the API (tech-preview) and removed Spec.Previous in favor of labels. Because of that, this IMO shouldn’t be a 🌱 change (it is more like tests/infra things that are not docs and would not be in the release notes), and we also haven’t been using that emoji consistently or for release generation (which would be the real motivation).

Originally, the emojis were meant to work with the kubebuilder release-tools, but that tooling is no longer maintained (see; https://github.com/kubernetes-sigs/kubebuilder-release-tools/?tab=readme-ov-file#release-notes-generation).

The main purpose behind the emojis was automated release-note categorisation:
✨ Features / 🐛 Bugs / ⚠️ Breaking Changes.
Example: https://github.com/kubernetes-sigs/kubebuilder/releases/tag/v4.9.0

pedjak:
> We’re changing the API (tech-preview) and removed Spec.Previous in favor of labels. Because of that, this IMO shouldn’t be a 🌱 change (it is more like tests/infra things that are not docs and would not be in the release notes), and we also haven’t been using that emoji consistently or for release generation (which would be the real motivation).

It aligns well with suggestion in https://github.com/operator-framework/operator-controller/pull/2245#discussion_r2410179970

camilamacedo86:
Hi @pedjak,

We are changing the public API:
https://github.com/operator-framework/operator-controller/pull/2315/files#diff-35832abf0db7539022428aa7858395b2f994892cb14e0f2a41c3d05137af599d

If this had already been promoted to GA, we wouldn’t be able to do that. It currently fails the [go-apidiff check](https://github.com/operator-framework/operator-controller/actions/runs/19439941147/job/55620508906?pr=2315), **which confirms it’s a user-facing change.** Anything that affects what users see or experience is considered a user-facing change. In contrast, changes that only modify how we do things internally—without altering how information is presented, stored, or accessed on the cluster—are not user-facing.

So it should actually have a ⚠️ in case someone is still using it.
I’ve updated the title to make that clearer.

Additionally, this change resolves a user-facing issue by ensuring that orphaned revisions are no longer left behind on the cluster.

Thank you 🙏

internal/operator-controller/controllers/clusterextensionrevision_controller.go
```diff
@@ -440,14 +453,47 @@ func (c *ClusterExtensionRevisionReconciler) removeFinalizer(ctx context.Context
...
+// Filters out the current revision, archived revisions, deleting revisions, and revisions with equal or higher numbers.
+func (c *ClusterExtensionRevisionReconciler) listPreviousRevisions(ctx context.Context, rev *ocv1.ClusterExtensionRevision) ([]client.Object, error) {
+	ownerLabel, ok := rev.Labels[ClusterExtensionRevisionOwnerLabel]
+	if !ok {
+		// No owner label means this revision isn't properly labeled - return empty list
+		return nil, nil
+	}
+
+	revList := &ocv1.ClusterExtensionRevisionList{}
+	if err := c.TrackingCache.List(ctx, revList, client.MatchingLabels{
```
pedjak, line 466: what is the advantage of using `c.TrackingCache.List()` vs. `c.Client.List()`?

internal/operator-controller/controllers/clusterextensionrevision_controller.go
```diff
@@ -440,14 +453,47 @@ func (c *ClusterExtensionRevisionReconciler) removeFinalizer(ctx context.Context
...
-	previous := make([]client.Object, 0, len(rev.Spec.Previous))
-	for _, specPrevious := range rev.Spec.Previous {
-		prev := &unstructured.Unstructured{}
-		prev.SetName(specPrevious.Name)
-		prev.SetUID(specPrevious.UID)
-		prev.SetGroupVersionKind(ocv1.GroupVersion.WithKind(ocv1.ClusterExtensionRevisionKind))
-		previous = append(previous, prev)
+// listPreviousRevisions returns active revisions belonging to the same ClusterExtension with lower revision numbers.
+// Filters out the current revision, archived revisions, deleting revisions, and revisions with equal or higher numbers.
+func (c *ClusterExtensionRevisionReconciler) listPreviousRevisions(ctx context.Context, rev *ocv1.ClusterExtensionRevision) ([]client.Object, error) {
```
pedjak: any reason why we do not return `[]*ocv1.ClusterExtensionRevision`?

internal/operator-controller/controllers/clusterextensionrevision_controller.go
camilamacedo86: Good question. TrackingCache.List() comes from Boxcutter’s managedcache (https://pkg.go.dev/pkg.package-operator.run/boxcutter@v0.7.1/managedcache
). It tracks object-to-revision relationships, which the standard c.Client doesn’t. This is mostly experimental — we could use the default controller-runtime client instead.

Since this is in tech preview, I’m was looking for us to use it for now, but we should keep an eye on the usual trade-offs: double caching, some memory overhead, possible some maintenance complexity. If the tracking features don’t end up paying off, we can switch to c.Client.List() later.

My idea was try out it now, and see if we found bugs, allow we epxlore more in follow-ups and add some E2E coverage around it. Let me know if this makes sense.

@thetechnick have you any reason for we do not use it?

internal/operator-controller/controllers/clusterextensionrevision_controller.go
camilamacedo86: Changed ! I think is better your proposal

camilamacedo86:
/hold 

Doing some tests locally

PR #2316, opened 2025-11-09
✨ Add Bundle-Agnostic Configuration Validation Using JSON Schema

camilamacedo86:
When we install an operator using ClusterExtension, we can now configure things like which namespace it should watch. Previously, if you made a typo or configured it incorrectly, the error would only show up later during deployment. Now, configuration is validated immediately using JSON Schema, and we get clear error messages right away.

## Why We Did This

Issue: https://issues.redhat.com/browse/OPRUN-4112

Users were getting confusing errors when they misconfigured their operators. By validating configuration upfront, we can give much better error messages that tell you exactly what's wrong and how to fix it.

## How It Works

We introduced a `ConfigSchemaProvider` interface that lets different bundle types describe their own configuration rules and all packages format types (registry/v1 or helm) use the same validation process - only the source of the rules changes.

## Do I Need to Configure `watchNamespace`?

It depends on what install modes your operator supports:

| What Your Operator Supports | Do You Need watchNamespace? |
|------------------------------|------------------------------|
| AllNamespaces only | **No** - you can omit it or set it to null |
| OwnNamespace only | **Yes** - must be the same namespace where the operator is installed |
| SingleNamespace only | **Yes** - must be a different namespace than where the operator is installed |
| AllNamespaces + OwnNamespace | **No** - optional, defaults to AllNamespaces if you don't specify |
| OwnNamespace + SingleNamespace | **Yes** - you must pick a namespace |

### Design Diagram

```
┌─────────────────────────────────────────────────────────────┐
│                    ConfigSchemaProvider                     │
│                         (Interface)                         │
└────────────┬────────────────────────────────┬───────────────┘
             │                                │
             ▼                                ▼
    ┌────────────────┐              ┌─────────────────┐
    │  RegistryV1    │              │  Future: Helm,  │
    │                │              │  Registry+v2    │
    │ GetConfigSchema│              │                 │
    │ (generates     │              │ GetConfigSchema │
    │  from install  │              │ (reads from     │
    │  modes)        │              │  chart/bundle)  │
    └────────┬───────┘              └────────┬────────┘
             │                               │
             └───────────┬───────────────────┘
                         ▼
              ┌──────────────────────┐
              │  UnmarshalConfig()   │
              │                      │
              │  1. Get schema       │
              │  2. Validate         │──→ Custom Validators:
              │  3. Parse            │    • ownNamespaceInstallMode
              └──────────┬───────────┘    • singleNamespaceInstallMode
                         │
                         ▼
                 ┌──────────────┐
                 │ bundle.Config│
                 │              │
                 │ - Opaque data│
                 │ - Accessors  │
                 └──────────────┘
```

**TL"DR:**

<details>
<summary>Examples - Use cases </summary>

### No configuration (AllNamespaces mode)

```yaml
spec:
  packageName: my-operator
```

### Watch specific namespace (SingleNamespace mode)

```yaml
metadata:
  namespace: operator-install-ns
spec:
  packageName: my-operator
  config:
    watchNamespace: target-namespace  # Different from install namespace
```

### Watch install namespace (OwnNamespace mode)

```yaml
metadata:
  namespace: my-namespace
spec:
  packageName: my-operator
  config:
    watchNamespace: my-namespace  # Same as install namespace
```

## Error Messages

All errors start with `invalid ClusterExtension configuration: invalid configuration:` followed by the specific problem.

### Typo in Field Name

```yaml
spec:
  config:
    watchNamespce: my-namespace  # Typo
```

**Error:**
```
invalid ClusterExtension configuration: invalid configuration: unknown field "watchNamespce"
```

### Missing Required Field

```yaml
spec:
  packageName: my-operator
  # No config - but operator requires watchNamespace
```

**Error:**
```
invalid ClusterExtension configuration: invalid configuration: required field "watchNamespace" is missing
```

### Wrong Type

```yaml
spec:
  config:
    watchNamespace: 123  # Should be a string
```

**Error:**
```
invalid ClusterExtension configuration: invalid configuration: invalid value type for field "watchNamespace": expected "string" but got "number"
```

### OwnNamespace Mode - Wrong Namespace

```yaml
metadata:
  namespace: correct-namespace
spec:
  config:
    watchNamespace: wrong-namespace  # Must match install namespace
```

**Error:**
```
invalid ClusterExtension configuration: invalid configuration: configuration validation failed: 
- at '/watchNamespace': 'wrong-namespace' is not valid ownNamespaceInstallMode: 
  invalid value "wrong-namespace": watchNamespace must be "correct-namespace" 
  (the namespace where the operator is installed) because this operator only 
  supports OwnNamespace install mode
```

### SingleNamespace Mode - Can't Use Install Namespace

```yaml
metadata:
  namespace: install-ns
spec:
  config:
    watchNamespace: install-ns  # Must be different from install namespace
```

**Error:**
```
invalid ClusterExtension configuration: invalid configuration: configuration validation failed: 
- at '/watchNamespace': 'install-ns' is not valid singleNamespaceInstallMode: 
  invalid value "install-ns": watchNamespace must be different from "install-ns" 
  (the install namespace) because this operator uses SingleNamespace install mode 
  to watch a different namespace
```

### Invalid JSON/YAML

```yaml
spec:
  config: {"incomplete
```

**Error:**
```
invalid ClusterExtension configuration: invalid configuration: found unexpected end of stream
```

</details>

<details>
<summary>Why we are formatting the messages:</summary>

The minimal option here would be like

```go
// formatSchemaError removes technical schema file paths from validation errors
// while preserving the library's clear error messages.
func formatSchemaError(err error) error {
	msg := err.Error()
	// Strip the "jsonschema validation failed with 'file:///.../schema.json#'" prefix
	// Keep everything after the newline which has the actual error details
	if idx := strings.Index(msg, "\n"); idx != -1 {
		return fmt.Errorf("invalid configuration:%s", msg[idx:])
	}
	return fmt.Errorf("invalid configuration: %s", msg)
}
```

However, I followed your suggestion to use DetailedOutput() / BasicOutput(), and I think we've achieved a better result with improved error messages that are neither overly fragile nor difficult to maintain.

## Output Comparison (Minimal vs Current Approach)

| Scenario | Example Input | **Minimal Approach** | **Current Approach** | Winner |
|----------|---------------|---------------------|---------------------|--------|
| **Missing required field** | `{}` when `watchNamespace` required | `invalid configuration:`<br/>`- at '': missing property 'watchNamespace'` | `required field "watchNamespace" is missing` | **Current** - Clearer terminology |
| **Null instead of required** | `{"watchNamespace": null}` | `invalid configuration:`<br/>`- at '/watchNamespace': got null, want string` | `required field "watchNamespace" is missing` | **Current** - Recognizes intent |
| **Unknown field** | `{"unknownField": "value"}` | `invalid configuration:`<br/>`- at '': additional properties 'unknownField' not allowed` | `unknown field "unknownField"` | **Current** - Concise |
| **Type mismatch** | `{"watchNamespace": 123}` | `invalid configuration:`<br/>`- at '/watchNamespace': got number, want string` | `invalid type for field "watchNamespace": got number, want string` | **Current** - Adds field context |
| **Nested field error** | `{"resources": {"memory": 512}}` | `invalid configuration:`<br/>`- at '/resources/memory': got number, want string` | `invalid type for field "resources.memory": got number, want string` | **Current** - Dot notation for paths |
| **Root type error** | `true` (not an object) | `invalid configuration:`<br/>`- at '': got boolean, want object` | `invalid type: got boolean, want object` | **Similar** |
| **Constraint violation** | `{"replicaCount": 0}` with min=1 | `invalid configuration:`<br/>`- at '/replicaCount': value should be >= 1` | `value should be >= 1` | **Similar** - Pass through |
| **Enum violation** | `{"type": "Invalid"}` | `invalid configuration:`<br/>`- at '/type': value should be one of [...]` | `value should be one of [...]` | **Similar** - Pass through |
| **Custom format validator** | `{"watchNamespace": "wrong-ns"}` | `invalid configuration:`<br/>`- at '/watchNamespace': 'wrong-ns' is not valid ownNamespaceInstallMode: ...` | `invalid value "wrong-ns": watchNamespace must be "install-ns" (the namespace where the operator is installed) ...` | **Current** - No bullet prefix |
| **Multiple errors** | `{}` requires 2 fields | `invalid configuration:`<br/>`- at '': missing property 'replicaCount'`<br/>*(stops at first error)* | `multiple errors found:`<br/>`  - required field "replicaCount" is missing`<br/>`  - required field "image" is missing` | **Current** - Shows ALL errors! |

</details>


internal/operator-controller/rukpak/bundle/config.go
```diff
@@ -6,119 +6,498 @@ import (
...
-	if config == nil {
-		return nil
+// buildSchemaForInstallModes creates validation rules for registry+v1 bundles.
+//
+//   - AllNamespaces only -> watchNamespace not allowed (watches all namespaces)
+//   - OwnNamespace only -> watchNamespace required (must equal install namespace)
+//   - SingleNamespace only -> watchNamespace required (must NOT equal install namespace)
+//   - AllNamespaces + OwnNamespace -> watchNamespace optional
+//   - Empty install modes -> no configuration allowed (bundle is invalid)
+func buildSchemaForInstallModes(installModes sets.Set[v1alpha1.InstallMode]) (string, error) {
```
perdasilva: In the not-so-distant future we'll introduce subscription config fields, so maybe a more generic name like:

```suggestion
func buildBundleConfigSchema(installModes sets.Set[v1alpha1.InstallMode]) (string, error) {
```

internal/operator-controller/rukpak/bundle/config.go
```diff
@@ -6,119 +6,498 @@ import (
...
+//   - When bundle supports AllNamespaces + OwnNamespace -> OPTIONAL (can default to AllNamespaces)
+//   - When bundle supports only AllNamespaces -> NOT APPLICABLE (no watchNamespace field)
+func isWatchNamespaceConfigRequired(installModes sets.Set[v1alpha1.InstallMode]) bool {
+	return isWatchNamespaceConfigurable(installModes) &&
+		!installModes.Has(v1alpha1.InstallMode{Type: v1alpha1.InstallModeTypeAllNamespaces, Supported: true})
 }
 
-// formatUnmarshalError format JSON unmarshal errors to be more readable
+// formatWatchNamespaceError formats watchNamespace-specific validation errors.
+func formatWatchNamespaceError(msg string, hasWatchNamespace bool, watchNamespaceValue string) error {
```
perdasilva: I don't think we should have special handling for a specific field. Bundle configuration is meant to be opaque. E.g. another bundle can come along in the future with a watchNamespace configuration that has a completely different semantic.

internal/operator-controller/rukpak/bundle/config_test.go
```diff
@@ -46,73 +59,73 @@ func Test_UnmarshalConfig(t *testing.T) {
 			name:                  "rejects invalid json",
 			supportedInstallModes: []v1alpha1.InstallModeType{v1alpha1.InstallModeTypeSingleNamespace},
 			rawConfig:             []byte(`{"hello`),
-			expectedErrMessage:    `error unmarshalling registry+v1 configuration: found unexpected end of stream`,
+			expectedErrMessage:    `error validating configuration: found unexpected end of stream`,
```
perdasilva: We should probably not use `error validating configuration` as there could be validation errors and there could be errors while validating the configuration. Maybe something like `invalid bundle configuration` instead?

internal/operator-controller/rukpak/bundle/config.go
```diff
@@ -6,119 +6,498 @@ import (
...
 	if bytes == nil {
-		return nil, nil
+		bytes = []byte("{}")
 	}
 
-	bundleConfig := &Config{}
-	if err := yaml.UnmarshalStrict(bytes, bundleConfig); err != nil {
-		return nil, fmt.Errorf("error unmarshalling registry+v1 configuration: %w", formatUnmarshalError(err))
+	// Step 1: Get the schema from the bundle
+	schema, err := schemaProvider.GetConfigSchema()
```
perdasilva: should we handle the case where schema is empty or nil (i.e. the bundle doesn't provide configuration or doesn't provide a configuration schema)?

internal/operator-controller/rukpak/bundle/config.go
```diff
@@ -6,119 +6,498 @@ import (
...
+}
+
+// selectNamespaceFormat chooses the appropriate format validator based on install modes.
+//
+//   - When OwnNamespace only -> "installNamespace" (must equal install namespace)
+//   - When SingleNamespace only -> "notInstallNamespace" (must NOT equal install namespace)
+//   - Otherwise -> "k8sNamespace" (any valid namespace)
+func selectNamespaceFormat(hasOwnNamespace, hasSingleNamespace bool) string {
+	if hasOwnNamespace && !hasSingleNamespace {
+		return "installNamespace"
```
perdasilva: maybe const these out as `formatInstallNamespace`, `formatNotInstallNamespace`, `formatK8sNamespace` since they are also used when setting up the schema validation?

internal/operator-controller/rukpak/bundle/config.go
```diff
@@ -6,119 +6,498 @@ import (
...
+	// Register custom formats for validation
+	compiler.RegisterFormat(&jsonschema.Format{
+		Name:     "k8sNamespace",
+		Validate: validateK8sNamespace,
+	})
+	compiler.RegisterFormat(&jsonschema.Format{
+		Name: "installNamespace",
+		Validate: func(value interface{}) error {
+			// First validate it's a valid namespace name
+			if err := validateK8sNamespace(value); err != nil {
```
perdasilva: I wonder if this is necessary here since the install namespace is provided by the system and expected to be a valid namespace name. It could also mean the user goes through two rounds of invalid configuration: first the namespace name isn't valid, then finds out it cannot be the install namespace. I don't think it's the end of the world, just a thought...wdyt?

internal/operator-controller/config/config.go
```diff
@@ -6,119 +6,498 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/santhosh-tekuri/jsonschema/v6"
```
perdasilva, line 33: maybe we should move config.go, config_test.go, and error_formatting_test.go out of this package and into a top level `bundle` package? wdyt? Maybe just leave the registry+v1 bundle building code here or move it to registryv1.go?

internal/operator-controller/rukpak/bundle/config.go
camilamacedo86: Done 🎉

internal/operator-controller/rukpak/bundle/config.go
camilamacedo86: Good cacher 🥇 
Add a simple condition to ignore in this case and the tests to cover the scenarios.

internal/operator-controller/rukpak/bundle/config_test.go
camilamacedo86: Thank you for raising it. Motivated by your comment I decided to look into it deeper. 

IMHO we should stick with the term that we agreed on: **ClusterExtension**, and avoid using **bundle**. For me, *bundle* is specific for the OLM packaging formats (registry-v1, registry-v2, etc.). I know we could say that a Helm chart is “a bundle of files”, but for me bundle is aka a term for OLM specific format. That is why I would like to avoid that.  

So, WDYT about we keep **ClusterExtension** as the consistent and format-agnostic term?**

```go
  invalid ClusterExtension configuration:
    invalid configuration:
      required field "watchNamespace" is missing
```
So, WDYT? Let me know if this works for you or if you want to adjust anything!

internal/operator-controller/rukpak/bundle/config.go
camilamacedo86: Done. Good idea. 🚀

internal/operator-controller/config/config.go
```diff
@@ -6,119 +6,498 @@ import (
...
+		Name: "installNamespace",
+		Validate: func(value interface{}) error {
+			// First validate it's a valid namespace name
+			if err := validateK8sNamespace(value); err != nil {
+				return err
+			}
+			// Then check it equals install namespace (if installNamespace is set)
+			if installNamespace == "" {
+				return nil
+			}
```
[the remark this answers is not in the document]
camilamacedo86: Add a comment to clarify
it is valid

internal/operator-controller/config/config.go
```diff
@@ -6,119 +6,498 @@ import (
...
+		Name: "notInstallNamespace",
+		Validate: func(value interface{}) error {
+			// First validate it's a valid namespace name
+			if err := validateK8sNamespace(value); err != nil {
+				return err
+			}
+			// Then check it does NOT equal install namespace (if installNamespace is set)
+			if installNamespace == "" {
+				return nil
+			}
```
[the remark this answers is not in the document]
camilamacedo86: valid

internal/operator-controller/rukpak/bundle/config.go
camilamacedo86: I think I agree with you.
We can remove validateK8sNamespace since it will be returned by k8s validations anyway.

internal/operator-controller/controllers/clusterextension_controller.go
```diff
@@ -446,7 +446,7 @@ func (r *ClusterExtensionReconciler) SetupWithManager(mgr ctrl.Manager, opts ...
 }
 
 func wrapErrorWithResolutionInfo(resolved ocv1.BundleMetadata, err error) error {
-	return fmt.Errorf("error for resolved bundle %q with version %q: %w", resolved.Name, resolved.Version, err)
+	return fmt.Errorf("error for resolved package %q with version %q: %w", resolved.Name, resolved.Version, err)
```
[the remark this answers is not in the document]
camilamacedo86: It might out of scope but I suggested a change here

internal/operator-controller/applier/provider.go
```diff
@@ -70,18 +70,13 @@ func (r *RegistryV1ManifestProvider) Get(bundleFS fs.FS, ext *ocv1.ClusterExtens
...
 		bundleConfigBytes := extensionConfigBytes(ext)
-		// treat no config as empty to properly validate the configuration
-		// e.g. ensure that validation catches missing required fields
-		if bundleConfigBytes == nil {
-			bundleConfigBytes = []byte(`{}`)
-		}
 		bundleConfig, err := bundle.UnmarshalConfig(bundleConfigBytes, rv1, ext.Spec.Namespace)
 		if err != nil {
-			return nil, fmt.Errorf("invalid bundle configuration: %w", err)
+			return nil, fmt.Errorf("invalid ClusterExtension configuration: %w", err)
```
pedjak, line 80: at line 73 we unmarshall  bundle config, but then in the case of error we report now that ClusterExtension config is invalid? Should it stay `invalid bundle configuration`?

internal/operator-controller/rukpak/bundle/config.go
```diff
@@ -6,119 +6,489 @@ import (
...
+	//
+	// Example schema:
+	//   {
+	//     "type": "object",
+	//     "required": ["watchNamespace"],
+	//     "properties": {
+	//       "watchNamespace": {"type": "string"}
+	//     }
+	//   }
+	GetConfigSchema() (string, error)
```
pedjak: can we shorten the name to `Get()`? given that the interface is already called `ConfigSchemaProvider`? Or `GetJSONSchema()`? Also, can we return a more structured type, e.g. `map[string]any` instead of just string?

internal/operator-controller/config/config.go
```diff
@@ -6,119 +6,489 @@ import (
...
+	val, exists := c.data["watchNamespace"]
+	if !exists {
+		return nil
+	}
+	// Handle explicit null
+	if val == nil {
+		return nil
+	}
+	// Extract string value
+	if str, ok := val.(string); ok {
```
pedjak: what if value is not a string? We should not return `nil`, IMO. Instead how about this:

```golang
return fmt.Sprintf("%v", val)
```

internal/operator-controller/rukpak/bundle/config.go
```diff
@@ -6,119 +6,489 @@ import (
...
 	if bytes == nil {
-		return nil, nil
+		bytes = []byte("{}")
 	}
 
-	bundleConfig := &Config{}
-	if err := yaml.UnmarshalStrict(bytes, bundleConfig); err != nil {
-		return nil, fmt.Errorf("error unmarshalling registry+v1 configuration: %w", formatUnmarshalError(err))
+	// Step 1: Get the schema from the bundle
+	schema, err := schemaProvider.GetConfigSchema()
```
pedjak: given that we pass `schemaProvider` in this function just to call `GetConfigSchema`, how about that we pass instead schema as the function arg?

internal/operator-controller/rukpak/bundle/config.go
```diff
@@ -6,119 +6,489 @@ import (
...
-			return fmt.Errorf("invalid 'watchNamespace' %q: namespace must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character", *config.WatchNamespace)
+	hasOwnNamespace := installModes.Has(v1alpha1.InstallMode{Type: v1alpha1.InstallModeTypeOwnNamespace, Supported: true})
+	hasSingleNamespace := installModes.Has(v1alpha1.InstallMode{Type: v1alpha1.InstallModeTypeSingleNamespace, Supported: true})
+
+	format := selectNamespaceFormat(hasOwnNamespace, hasSingleNamespace)
+
+	if isWatchNamespaceConfigRequired(installModes) {
+		watchNSProperty["type"] = "string"
+		watchNSProperty["format"] = format
+		*requiredFields = append(*requiredFields, "watchNamespace")
```
pedjak: it looks like that we have a side-effect in the function by appending fields to passed `requiredFields` arg, correct?

internal/operator-controller/rukpak/bundle/config.go
camilamacedo86: Yep, it think I missed the boat here. 
Removed.

internal/operator-controller/rukpak/bundle/config.go
camilamacedo86: Marking as resolved to facilitate the review

internal/operator-controller/rukpak/bundle/config.go
camilamacedo86: Marking as resolved to facilitate de review

internal/operator-controller/rukpak/bundle/config.go
camilamacedo86: Marking as resolved to facilitate de review

internal/operator-controller/rukpak/bundle/config.go
camilamacedo86: Marking as resolved to facilitate de review

internal/operator-controller/rukpak/bundle/config.go
```diff
@@ -6,119 +6,489 @@ import (
...
+	return "" // No format constraint needed for generic case
+}
+
+// validateConfigWithSchema validates the config bytes against the generated JSON schema.
+//
+// Note: We create a new compiler on each call because the custom formats (installNamespace,
+// notInstallNamespace) are closures that capture the installNamespace value, which varies
+// per ClusterExtension. If this becomes a performance bottleneck, we could cache compiled
+// schemas keyed by (schema JSON + installNamespace), but for now the overhead is negligible.
+func validateConfigWithSchema(configBytes []byte, schemaJSON string, installNamespace string) error {
```
pedjak: could we perhaps add receiver function to `Config` struct:

```golang
func (c *Config) Validate(installNamespace string) error
```

perhaps Config struct could contain reference to `ConfigSchemaProvider`?

internal/operator-controller/applier/provider.go
camilamacedo86: I don't think we should use the term **bundle**.

**Why?**  
"Bundle" is tightly tied to the `registry/v1` OLM bundle format.  
As we begin supporting other types—like Helm—I worry that continuing to use this term could cause confusion.

We had already aligned on moving away from the term **Operators**, since our goal is to support any kind of content (Helm and beyond). We also agreed to use **ClusterExtension** as our unified term. So I feel it would be clearer and more consistent if we continue using `ClusterExtension`.

In this context, it's basically a Config that is given a ClusterExtension.

pedjak: thanks for explaining it. :+1:

internal/operator-controller/rukpak/bundle/config.go
camilamacedo86: I think seems a good idea 👍

internal/operator-controller/applier/provider.go
camilamacedo86: @pedjak — I think others either can considers Helm as a “bundle.” any set of files. For me means specific OLM package format ( due docs and etc )
If we end up debating too much over the terminology, I’m fine reverting to “bundle” so we can move forward.
But, I’d still prefer something more agnostic, especially where it can apply broadly.

internal/operator-controller/rukpak/bundle/config.go
camilamacedo86: Closing as solved to facilitate the review.

internal/operator-controller/config/config.go
camilamacedo86: I think it is solved now. 
Feel free to resolve if you agree.

internal/operator-controller/rukpak/bundle/config.go
camilamacedo86: I think it is solved now. 
Feel free to resolve if you agree.

internal/operator-controller/rukpak/bundle/config.go
camilamacedo86: Thank you for the proposal. 

I think you’re suggesting this because of the idea of adding something like
a Validate(installNamespace string). The thing is, we already validate before
creating a Config, and we recreate and re-validate a fresh one on every
reconcile. So we never have an invalid Config in memory, and adding a
Validate method it seems for me we would likely just make us run the same checks twice.

I added a few comments in the code to explain this flow and the design choice
more clearly.  Happy to revisit it if I miss understanding something.

internal/operator-controller/config/config.go
```diff
@@ -6,119 +30,317 @@ import (
...
+// the data in a flexible format and provide accessor methods to get values out.
+//
+// Why there's no Validate() method:
+// We validate configuration when creating a Config. If you have a Config object,
+// it's already been validated - you don't need to check it again. The unexported
+// 'data' field means you can't create a Config yourself; you have to use
+// UnmarshalConfig, which does the validation.
 type Config struct {
-	WatchNamespace *string `json:"watchNamespace"`
+	data map[string]any
```
camilamacedo86: @perdasilva we are removing to make more Generic 
So, I think is important we move with it , if we want change it, before we promote Config API ( consequentely single/own ns )

internal/operator-controller/config/config.go
pedjak: you can drop now lines 100-102, because it is covered by line 103.

internal/operator-controller/config/config.go
```diff
@@ -6,119 +30,317 @@ import (
...
+// Parameters:
+//   - bytes: the user's configuration in YAML or JSON. If nil, we treat it as empty ({})
+//   - schema: describes what configuration is valid. If nil, we skip validation
+//   - installNamespace: the namespace where the operator will be installed. We use this
+//     to validate namespace constraints (e.g., OwnNamespace mode requires watchNamespace
+//     to equal installNamespace)
+//
+// If the user doesn't provide any configuration but the bundle requires some fields
+// (like watchNamespace), validation will fail with a helpful error.
+func UnmarshalConfig(bytes []byte, schema map[string]any, installNamespace string) (*Config, error) {
```
pedjak, line 118: the function unmarshal + validatates optionally the config. I would split these into two functions at least so that we can achieve something like:

```golang
c := NewConfigBuilder.FromBytes(bytes).WithSchema(schema).WithInstallNamespace(installNamespace).Build()
err := v.Validate()
```

internal/operator-controller/rukpak/bundle/config.go
```diff
@@ -6,119 +30,317 @@ import (
...
+			if endIdx := strings.Index(remaining, "'"); endIdx != -1 {
+				fieldName := remaining[:endIdx]
+				return fmt.Errorf("required field %q is missing", fieldName)
+			}
+		}
+		return errors.New("required field is missing")
+	}
+
+	// Required field set to null
+	if strings.Contains(msg, "got null, want string") {
```
pedjak: why we cannot use `schemaErr.DetailedOutput()` to parse and adapt messages if needed? Expecting and matching substrings is very fragile IMO.

internal/operator-controller/config/config.go
```diff
@@ -6,119 +30,317 @@ import (
...
+// formatSchemaError converts technical JSON schema errors into messages users can understand.
+//
+// The JSON schema library gives us error objects, but they don't have all the details
+// we need to make helpful messages (like which field name or what value was wrong).
+// So we parse the error strings to extract that information.
+//
+// The Test_ErrorFormatting_* tests make sure this parsing works. If we upgrade the
+// JSON schema library and those tests start failing, we'll know we need to update
+// this parsing logic.
+func formatSchemaError(err error) error {
```
pedjak, line 214: `formatSchemaError` is unprecise because it converts `ValueError` into another error instance. Also, why a user could not understand the original ValueError message?

internal/operator-controller/config/config.go
camilamacedo86: Thank you for raising it. But if we drop we will break, see:

```
val = nil

- if str, ok := val.(string); ok  // false, nil is not a string
- str := fmt.Sprintf("%v", nil)   // str = "<nil>" (the STRING)
- return &str                     // Returns pointer to "<nil>" 
```

Expected: config.GetWatchNamespace() returns nil (meaning "no specific namespace")
If we drop the check: config.GetWatchNamespace() returns &"<nil>" (a pointer to the literal string "<nil>")
The operator would then try to watch a namespace called "<nil>" instead of watching all namespaces.

We have unit test to cover those as well.

internal/operator-controller/rukpak/bundle/config.go
camilamacedo86: Closing as solved to facilitate the review.

internal/operator-controller/rukpak/bundle/config.go
camilamacedo86: I think that is a good idea

internal/operator-controller/config/config.go
camilamacedo86: The minimal option here would be like

```go
// formatSchemaError removes technical schema file paths from validation errors
// while preserving the library's clear error messages.
func formatSchemaError(err error) error {
	msg := err.Error()
	// Strip the "jsonschema validation failed with 'file:///.../schema.json#'" prefix
	// Keep everything after the newline which has the actual error details
	if idx := strings.Index(msg, "\n"); idx != -1 {
		return fmt.Errorf("invalid configuration:%s", msg[idx:])
	}
	return fmt.Errorf("invalid configuration: %s", msg)
}
```

However, I followed your suggestion to use DetailedOutput() / BasicOutput(), and I think we've achieved a better result with improved error messages that are neither overly fragile nor difficult to maintain.

## Output Comparison (Minimal vs Current Approach)

| Scenario | Example Input | **Minimal Approach** | **Current Approach** | Winner |
|----------|---------------|---------------------|---------------------|--------|
| **Missing required field** | `{}` when `watchNamespace` required | `invalid configuration:`<br/>`- at '': missing property 'watchNamespace'` | `required field "watchNamespace" is missing` | **Current** - Clearer terminology |
| **Null instead of required** | `{"watchNamespace": null}` | `invalid configuration:`<br/>`- at '/watchNamespace': got null, want string` | `required field "watchNamespace" is missing` | **Current** - Recognizes intent |
| **Unknown field** | `{"unknownField": "value"}` | `invalid configuration:`<br/>`- at '': additional properties 'unknownField' not allowed` | `unknown field "unknownField"` | **Current** - Concise |
| **Type mismatch** | `{"watchNamespace": 123}` | `invalid configuration:`<br/>`- at '/watchNamespace': got number, want string` | `invalid type for field "watchNamespace": got number, want string` | **Current** - Adds field context |
| **Nested field error** | `{"resources": {"memory": 512}}` | `invalid configuration:`<br/>`- at '/resources/memory': got number, want string` | `invalid type for field "resources.memory": got number, want string` | **Current** - Dot notation for paths |
| **Root type error** | `true` (not an object) | `invalid configuration:`<br/>`- at '': got boolean, want object` | `invalid type: got boolean, want object` | **Similar** |
| **Constraint violation** | `{"replicaCount": 0}` with min=1 | `invalid configuration:`<br/>`- at '/replicaCount': value should be >= 1` | `value should be >= 1` | **Similar** - Pass through |
| **Enum violation** | `{"type": "Invalid"}` | `invalid configuration:`<br/>`- at '/type': value should be one of [...]` | `value should be one of [...]` | **Similar** - Pass through |
| **Custom format validator** | `{"watchNamespace": "wrong-ns"}` | `invalid configuration:`<br/>`- at '/watchNamespace': 'wrong-ns' is not valid ownNamespaceInstallMode: ...` | `invalid value "wrong-ns": watchNamespace must be "install-ns" (the namespace where the operator is installed) ...` | **Current** - No bullet prefix |
| **Multiple errors** | `{}` requires 2 fields | `invalid configuration:`<br/>`- at '': missing property 'replicaCount'`<br/>*(stops at first error)* | `multiple errors found:`<br/>`  - required field "replicaCount" is missing`<br/>`  - required field "image" is missing` | **Current** - Shows ALL errors! |


**I think we should keep the custom as it is now. 
WDYT?**

internal/operator-controller/rukpak/bundle/config.go
camilamacedo86: Closing since it is done. Just to facilitate the review.

internal/operator-controller/config/config.go
camilamacedo86: I like the Builder approach, but I think we should **add it only IF/when we start to have many args**. 
I asked Claude AI to evaluate the current approach vs Builder Pattern. Here's the summary:

---

## Claude AI Analysis Summary

### Current Approach
```go
config, err := bundle.UnmarshalConfig(bytes, schema, installNamespace)
```

**Design principle:** Validation happens at construction. If you have a Config object, it's already valid.

### Builder Pattern Option 1: Validation in Build()
```go
config, err := bundle.NewConfigBuilder().
    FromBytes(bytes).
    WithSchema(schema).
    WithInstallNamespace(ns).
    Build()  // validation happens here
```

**Result:** Functionally identical to current approach, just more verbose (1 function call → 5 method calls).

### Builder Pattern Option 2: Separate Validate()
```go
config := builder.Build()  // unvalidated
err := config.Validate()   // validate separately
```

**Problems:**
- Allows invalid Config objects to exist
- Developers might forget to call Validate()
- Would need to run validation twice (inefficient)
- Breaks safety guarantee

### When Builder Pattern Makes Sense

✅ Good for:
- 7-10+ parameters
- Many optional parameters
- Parameters can be set independently
- Multiple valid construction paths

❌ Our situation:
- 3 required parameters
- All interdependent
- One construction path

### Maintainability Comparison

**Current:** 1 function to maintain
- Add parameter → update 1 function + call sites
- Bug → debug 1 function
- Test → 1 test function

**Builder:** 5+ methods to maintain  
- Add parameter → update struct + new method + Build() + call sites
- Bug → debug 5 methods to find which one
- Test → 6+ test functions

### Comparison Table

| Aspect | Current | Builder Option 1 | Builder Option 2 |
|--------|---------|------------------|------------------|
| Lines of code | 1 | 5 | 6 |
| API methods | 1 | 4 | 5 |
| Validation timing | At construction | At construction | After construction |
| Can have invalid Config? | ❌ No | ❌ No | ✅ Yes (bad!) |
| Double validation risk | ❌ No | ❌ No | ✅ Yes |
| Maintainability | ✅ High | ⚠️ Medium | ❌ Low |
| Safety guarantees | ✅ Strong | ✅ Strong | ❌ Weak |

### Problems with Each Option

**Builder Option 1 Problems:**
- More code to write and maintain (5 methods vs 1 function)
- When adding new parameter: must update struct + new method + Build() + all callers
- Bug fixing: must search through 5 methods instead of 1
- Testing: need 6+ test functions instead of 1
- More opportunities for breaking changes (5 methods can break vs 1)

**Builder Option 2 Problems:**
- **Critical:** Allows invalid Config objects to exist
- Developers can forget to call Validate()
- Must track validation state internally
- **Double validation problem:** If Build() validates → Validate() runs same checks again (waste). If Build() doesn't validate → invalid Config objects exist until Validate() is called (unsafe). Either way, it's problematic.
- Errors can occur in multiple places (Build() and Validate())
- Violates "safe by construction" principle

### Conclusion from Analysis (Recommendation)

**Keep current approach.** Builder pattern would add complexity without benefit for 3 required interdependent parameters. Refactor to builder only if we grow to 7+ parameters.

The current design already follows best practices:
- ✅ Validation at construction time
- ✅ Immutable validated objects  
- ✅ Simple, clear API
- ✅ Strong safety guarantees

### Why wait?

1. **YAGNI principle** - Don't add complexity for hypothetical future needs
2. **Current code is clean** - Simple, safe, well-tested
3. **Easy to refactor later** - If we add more parameters, we can migrate then
4. **No breaking change risk** - Adding builder later doesn't break existing code

camilamacedo86: So, @pedjak could we agree in keep it as it is until we need to change it ? Can we go [KISS](https://en.wikipedia.org/wiki/KISS_principle)?

internal/operator-controller/config/config.go
camilamacedo86: It is definitely a great call. Done 🎉

internal/operator-controller/config/config.go
pedjak: If I understand the code correctly, if `val == nil`, we return already on the line 99, so therefore we can drop lines where we check if val is the string because we maybe convert it to string anyway on the line 106.

internal/operator-controller/config/config.go
pedjak: > So, @pedjak could we agree in keep it as it is until we need to change it ? Can we go [KISS](https://en.wikipedia.org/wiki/KISS_principle)?

Sure, my only concern is `schema` arg is optional and if passed we can then trigger validation. So, similar to arguments above - if user forget to pass schema (i.e. passes `nil` or an empty map) it is going to get possibly invalid config instances.

camilamacedo86: Thanks for raising this — I get the concern. In short, the schema being optional is **intentional**.

Each package format decides whether it needs validation:

- registry+v1 → always provides a schema  
- Helm → may or may not  
- Future formats → choose their own rules  

When `schema == nil`, it just means that format doesn’t support validation. The config is still parsed, fields are still accessible, and JSON/YAML errors are still caught.  
When a schema *is* provided, we enforce full validation.

For registry+v1 specifically, the schema is always generated from the CSV, so there’s no risk of “forgetting” it.

Here’s the core logic, simplified:

```go
// Treat nil config as {} so required fields can still be validated
if bytes == nil {
    bytes = []byte("{}")
}

// Validate only if a schema is provided
if schema != nil {
    if err := validateConfigWithSchema(bytes, schema, installNamespace); err != nil {
        return nil, fmt.Errorf("invalid configuration: %w", err)
    }
}
```

I hope that clarifies

internal/operator-controller/config/config.go
camilamacedo86: OH I think you are right I changed that. 
Yep we can dropped.

internal/operator-controller/config/config.go
camilamacedo86: @pedjak but anyway that is something that tomorrow we can change
If we have more args and etc we can always change.
I do not think that should be a blocker for the PR change get merged. 
WDYT?

camilamacedo86:
Hi @perdasilva @pedjak,

Thank you both so much for all your suggestions and for the thorough review.
I’ve updated the description to clarify the points you raised, and addressed all the feedback so far.

WDYT?
Are we good to get this merged, or is there anything else you'd like to check or suggest?

Thanks again!

internal/operator-controller/config/config.go
```diff
@@ -0,0 +1,377 @@
...
+// SchemaProvider lets each package format type describe what configuration it accepts.
+//
+// Different package format types provide schemas in different ways:
+//   - registry+v1: builds schema from the operator's install modes
+//   - Helm: reads schema from values.schema.json in the chart
+//   - registry+v2: (future) will have its own way
+//
+// Note: We don't store this in the Config struct. We only need it when validating
+// the user's input. After that, the Config has the validated data and doesn't need
+// the schema anymore.
```
pedjak: this is an implementation detail that does not need be documented here.

internal/operator-controller/config/config.go
pedjak: if Config has just single field, why do not declare it as

```golang
type Config ma[string]any
```

internal/operator-controller/config/config.go
```diff
@@ -0,0 +1,377 @@
...
+		return nil
+	}
+	val, exists := c.data["watchNamespace"]
+	if !exists {
+		return nil
+	}
+	// User set watchNamespace: null - treat as "not configured"
+	if val == nil {
+		return nil
+	}
```
pedjak, line 94: nit: we could maybe collapse this into single `if`:

```golang
if !exists || val == nil {
  return nil
}
```

internal/operator-controller/config/config.go
```diff
@@ -0,0 +1,377 @@
...
+		return err
+	}
+
+	// Use BasicOutput() to get structured error information
+	// This is more robust than parsing error strings
+	output := ve.BasicOutput()
+	if output == nil || len(output.Errors) == 0 {
+		// No structured errors available, fallback to error message
+		// Note: Using %s not %w since ve.Error() is already a formatted string
+		return fmt.Errorf("%s", ve.Error())
```
pedjak: why we do not return `ve` directly:

```golang
return ve
```

internal/operator-controller/config/config.go
pedjak: I would suggesting choosing a better name for this function, because we are not formatting error  - we convert it.

internal/operator-controller/config/config.go
```diff
@@ -0,0 +1,377 @@
...
+func formatSingleError(errUnit jsonschema.OutputUnit) string {
+	// Check the keyword location to identify the error type
+	switch {
+	case strings.Contains(errUnit.KeywordLocation, "/required"):
+		// Missing required field
+		fieldName := extractFieldNameFromMessage(errUnit.Error)
+		if fieldName != "" {
+			return fmt.Sprintf("required field %q is missing", fieldName)
+		}
+		return "required field is missing"
```
pedjak, line 264: how a user would know which field is missing?

internal/operator-controller/config/config.go
```diff
@@ -0,0 +1,377 @@
...
+
+	case strings.Contains(errUnit.KeywordLocation, "/additionalProperties"):
+		// Unknown/additional field
+		fieldName := extractFieldNameFromMessage(errUnit.Error)
+		if fieldName != "" {
+			return fmt.Sprintf("unknown field %q", fieldName)
+		}
+		return "unknown field"
```
pedjak, line 272: same question as above

internal/operator-controller/rukpak/bundle/registryv1.go
```diff
@@ -17,3 +20,134 @@ type RegistryV1 struct {
 	CRDs        []apiextensionsv1.CustomResourceDefinition
 	Others      []unstructured.Unstructured
 }
+
+// Get builds a validation schema based on what install modes the operator supports.
+//
+// For registry+v1 bundles, we look at the CSV's install modes and generate a schema
+// that matches. For example, if the operator only supports OwnNamespace mode, we'll
+// require the user to provide a watchNamespace that equals the install namespace.
+func (rv1 RegistryV1) Get() (map[string]any, error) {
```
pedjak: perhaps to use pointer for the receiver?

```suggestion
func (rv1 *RegistryV1) Get() (map[string]any, error) {
```

camilamacedo86:
Hi @Predrag 

Thank you. Regards:

```
if Config has just single field, why do not declare it as

type Config ma[string]any
```

Using a type alias for Config looks simple now, but it blocks future changes. Once Config is a map alias, you can never turn it into a struct without breaking every caller that does config["key"] = value. If we think we may add fields later (metadata, versioning, validation helpers, etc.), a struct is the only forward-compatible option. Starting with a struct avoids a guaranteed breaking change later.

**We cannot change it after we promote that is way we cannot use alias here.**

internal/operator-controller/config/config.go
camilamacedo86: OK. I can remove

internal/operator-controller/config/config.go
camilamacedo86: I’d prefer to keep it as it is for readability. I’d rather have code that’s easier to understand than reduce the number of lines.

internal/operator-controller/config/config.go
camilamacedo86: If the field name isn’t provided, we simply can’t report it — we can only surface what exists. 
Ideally the field name should always be present, but if it’s missing, there’s nothing for us to expose. 

This check ensures we handle all cases safely. And if one day a missing name shows up unexpectedly, then we’ll need to fix the code to guarantee the field actually has a valid Name.

internal/operator-controller/config/config.go
camilamacedo86: Same above. We Cannot inform a field Name when we have not it.
So, we just handle the text.

internal/operator-controller/rukpak/bundle/registryv1.go
camilamacedo86: Yes, I think it make sense done.

internal/operator-controller/config/config.go
camilamacedo86: Thanks for the question — it’s a good one. 👍 

The short version is: **we wrap the validation error so we don’t leak the concrete `jsonschema` type to callers**. If we return the raw `ValidationError`, people can start type-asserting it, using its fields, and depending on that library. Once they do that, we can’t ever change validators without breaking their code.

#### Example of what we avoid (returning the raw error)

```go
func formatSchemaError(err error) error {
    ve := &jsonschema.ValidationError{}
    errors.As(err, &ve)
    return ve // exposes jsonschema.ValidationError
}
```

A caller could now do:

```go
if v, ok := err.(*jsonschema.ValidationError); ok {
    // Now they depend on jsonschema internals
    use(v.InstanceLocation)
}
```

**This creates a tight coupling between our API and the library.**

#### Example of what we do instead (wrap the error)

```
func formatSchemaError(err error) error {
    ve := &jsonschema.ValidationError{}
    errors.As(err, &ve)
    return fmt.Errorf("%s", ve.Error()) // just the message
}
```

Now callers only get:

```
log.Error("validation failed: %v", err)
```

They can’t depend on any internal types.

---

#### Why `%s` instead of `%w`?

- `%w` keeps the original type in the error chain → callers can still run `errors.As()`
- `%s` gives only the message → no type leakage

Since the whole point is to prevent exposing `jsonschema.ValidationError`, `%s` is intentional.

---

#### Pros & Cons

### Returning the raw `ValidationError`

**Pros**
- Simpler code  
- Gives callers full validation details  

**Cons**
- Leaks our validator library  
- Encourages users to depend on its types  
- Makes switching libraries a breaking change  
- Locks us into this error shape forever  

---

##### Wrapping the error (`fmt.Errorf("%s", ve.Error())`)

**Pros**
- Hides internal validation library  
- We can change validators later without breaking anything  
- Gives us freedom to refactor or improve messages  
- Keeps API surface clean and stable  

**Cons**
- One extra line of code  
- Callers don’t get low-level library details (which is intentional)

---

In short, this is defensive API design: **we only expose what we want to support long-term**, and we keep library details private so we can evolve things later without breaking users.

camilamacedo86:
Hi @pedjak,

Regarding the function name:

> I would suggest choosing a better name for this function, because we are not formatting the error — we convert it.

I am not the best to pick names but IMHO, `formatSchemaError` is still the right name. The purpose of the function is to
format the error message into something user-friendly. The conversion step is just
the internal mechanism behind that formatting ("how" and not "what"). Naming should reflect the goal,
not the implementation details.

When someone sees `formatSchemaError`, they immediately understand: “this takes a
schema error and formats it for users.” That’s exactly what it does.

Would keeping this name be a blocker for moving forward?

internal/operator-controller/config/config.go
```diff
@@ -0,0 +1,373 @@
...
+// SchemaProvider lets each package format type describe what configuration it accepts.
+//
+// Different package format types provide schemas in different ways:
+//   - registry+v1: builds schema from the operator's install modes
+//   - Helm: reads schema from values.schema.json in the chart
+//   - registry+v2: (future) will have its own way
+type SchemaProvider interface {
+	// Get returns a JSON Schema describing what configuration is valid.
+	// Returns nil if this package format type doesn't need configuration validation.
+	Get() (map[string]any, error)
```
perdasilva: I get that at this level it becomes redundant, but maybe we should say what we're getting? something like?
```suggestion
	GetConfigSchema() (map[string]any, error)
```

camilamacedo86: OK. Done.

pedjak (approved):
/lgtm

PR #2319, opened 2025-11-11
🌱 Revision manifest sanitization

dtfranz:
# Description

Removes metadata fields used by the kube api to avoid situations like #2295 

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/operator-controller/applier/boxcutter.go
+++ b/internal/operator-controller/applier/boxcutter.go
@@ -20,6 +20,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/yaml"
 
 	helmclient "github.com/operator-framework/helm-operator-plugins/pkg/client"
@@ -35,8 +36,9 @@ const (
 )
 
 type ClusterExtensionRevisionGenerator interface {
-	GenerateRevision(bundleFS fs.FS, ext *ocv1.ClusterExtension, objectLabels, revisionAnnotations map[string]string) (*ocv1.ClusterExtensionRevision, error)
+	GenerateRevision(ctx context.Context, bundleFS fs.FS, ext *ocv1.ClusterExtension, objectLabels, revisionAnnotations map[string]string) (*ocv1.ClusterExtensionRevision, error)
 	GenerateRevisionFromHelmRelease(
+		ctx context.Context,
 		helmRelease *release.Release, ext *ocv1.ClusterExtension,
 		objectLabels map[string]string,
 	) (*ocv1.ClusterExtensionRevision, error)
@@ -48,6 +50,7 @@ type SimpleRevisionGenerator struct {
 }
 
 func (r *SimpleRevisionGenerator) GenerateRevisionFromHelmRelease(
+	ctx context.Context,
 	helmRelease *release.Release, ext *ocv1.ClusterExtension,
 	objectLabels map[string]string,
 ) (*ocv1.ClusterExtensionRevision, error) {
@@ -64,11 +67,11 @@ func (r *SimpleRevisionGenerator) GenerateRevisionFromHelmRelease(
 		maps.Copy(labels, existingLabels)
 		maps.Copy(labels, objectLabels)
 		obj.SetLabels(labels)
-		sanitize(&obj)
 
 		// Memory optimization: strip large annotations
 		// Note: ApplyStripTransform never returns an error in practice
 		_ = cache.ApplyStripAnnotationsTransform(&obj)
+		sanitizedUnstructured(ctx, &obj)
 
 		objs = append(objs, ocv1.ClusterExtensionRevisionObject{
 			Object:              obj,
@@ -88,6 +91,7 @@ func (r *SimpleRevisionGenerator) GenerateRevisionFromHelmRelease(
 }
 
 func (r *SimpleRevisionGenerator) GenerateRevision(
+	ctx context.Context,
 	bundleFS fs.FS, ext *ocv1.ClusterExtension,
 	objectLabels, revisionAnnotations map[string]string,
 ) (*ocv1.ClusterExtensionRevision, error) {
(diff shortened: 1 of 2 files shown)


internal/operator-controller/applier/boxcutter.go
```diff
@@ -135,6 +140,54 @@ func (r *SimpleRevisionGenerator) GenerateRevision(
...
+	if status, ok := m["status"].(map[string]any); ok && len(status) > 0 {
+		strippedFields = append(strippedFields, "status")
+		delete(m, "status")
+	}
+
+	var metadata map[string]any
+	if metaRaw, ok := m["metadata"]; ok {
+		metadata, ok = metaRaw.(map[string]any)
+		if !ok {
+			return
```
rashmigottipati, line 166:
```suggestion
            l.Info("warning: unexpected metadata type in manifest, skipping sanitization",
             "kind", obj.GetKind(),
             "name", obj.GetName(),
             "type", fmt.Sprintf("%T", metaRaw))
			return
```

dtfranz: I like the idea but drilling down into `obj.GetKind()` and `obj.GetName()` it looks like eventually they perform the exact same type-cast, and return `""` when it fails. Therefore if we fail the type-cast the `kind` and `name` in this log will always be empty strings - assuming I've interpreted it correctly ([reference](https://github.com/kubernetes/apimachinery/blob/b72d93d174332f952a8d431419fece5e6f044bcb/pkg/apis/meta/v1/unstructured/helpers.go#L61)). We could log regardless, but if we can't pull any identifiers will it still be helpful?

internal/operator-controller/applier/boxcutter.go
```diff
@@ -135,6 +140,54 @@ func (r *SimpleRevisionGenerator) GenerateRevision(
 	return r.buildClusterExtensionRevision(objs, ext, revisionAnnotations), nil
 }
 
+var restrictedMetadata = []string{
```
perdasilva: There was a suggestion by Predrag on the ticket to maybe do it as an allow list instead, i.e. just allow:
.spec, .metadata.name, .metadata.namespace, .metadata.annotations, and .metadata.labels. Wdyt?

camilamacedo86 (changes requested):
Thanks for working on this! I really appreciate the effort to make things cleaner.

Just a small thought — **I don’t think we should get in the middle of how the Kubernetes API manages its own fields.**  
We should let the API do its job and handle those values as it’s designed to.  
If we start stripping or modifying them ourselves, we might cause unexpected behavior, ownership issues, or race conditions later on.

- 🆗 It’s totally fine to clean up our **testdata** for readability — that’s helpful and safe.  
- 🚫 But in the **controller or reconciler**, it’s best to let k8s manage those fields naturally.

Letting k8s handle its own fields keeps our controller simple, predictable, and safe. 🌿

dtfranz:
> Thanks for working on this! I really appreciate the effort to make things cleaner.
> 
> Just a small thought — **I don’t think we should get in the middle of how the Kubernetes API manages its own fields.** We should let the API do its job and handle those values as it’s designed to. If we start stripping or modifying them ourselves, we might cause unexpected behavior, ownership issues, or race conditions later on.
> 
>     * 🆗 It’s totally fine to clean up our **testdata** for readability — that’s helpful and safe.
> 
>     * 🚫 But in the **controller or reconciler**, it’s best to let k8s manage those fields naturally.
> 
> 
> Letting k8s handle its own fields keeps our controller simple, predictable, and safe. 🌿

Thanks Camila! But TBH, I'm not sure what you're referring to here WRT "getting in the middle of things". What we're doing here is removing fields controlled/modified/etc. by the kube API server from the bundle manifests before we add them to the CER phases. Bundle creators should not be setting these fields. If they do, and we don't clear them, the reconciler will see that a field was modified from the original revision and may infinitely create new revisions - see [here](https://github.com/operator-framework/operator-controller/pull/2295#issue-3575876632).

internal/operator-controller/applier/boxcutter.go
dtfranz: Ooooh I didn't understand that message at first, I didn't realize he meant a "white" list 😆 But yes, I can do that and that would be simpler.

dtfranz: On second thought this might make it more difficult to give high-granularity warnings when we find a field to remove, but I'll think about it some more.

tmshort:
> Thanks Camila! But TBH, I'm not sure what you're referring to here WRT "getting in the middle of things". What we're doing here is removing fields controlled/modified/etc. by the kube API server from the bundle manifests before we add them to the CER phases. Bundle creators should not be setting these fields. If they do, and we don't clear them, the reconciler will see that a field was modified from the original revision and may infinitely create new revisions - see [here](https://github.com/operator-framework/operator-controller/pull/2295#issue-3575876632).

Which is something (infinitely create new revisions) I'm seeing in another bug.

So, this is likely a case where we _do_ want to intervene in the processing of these fields.

internal/operator-controller/applier/boxcutter.go
tmshort: > Ooooh I didn't understand that message at first, I didn't realize he meant a "white" list 😆 But yes, I can do that and that would be simpler.

The terms "white" and "black" lists are problematic; the preferred terms are "allow" and "block" lists.

camilamacedo86:
Hi @tmshort @dtfranz 

So, I think now I understand it, the goal here is to make sure that when we create new revisions (e.g., v2, v3, v4), they’re based on the latest revision object — which makes total sense. I agree that we shouldn’t carry over all the existing fields as-is into the new revision.

However, instead of stripping those fields out, what do you think about defining a new struct (or helper) that explicitly copies only the fields that make sense for the revision? This way, if Kubernetes adds new fields in the future, our code will still behave correctly without needing manual updates.

Maybe something like a NewRevisionFrom() function that builds a new revision object and only fills in the fields we need — that might make the intent clearer and the logic more robust.

internal/operator-controller/applier/boxcutter.go
dtfranz: Thanks Todd - I avoid them myself but only mentioned it here because the original suggestion used that terminology.

internal/operator-controller/applier/boxcutter.go
rashmigottipati: Yup that makes sense, I don't think the log message will be helpful if we aren't pulling any identifiers. 
Thanks for clarifying @dtfranz!

dtfranz:
Branch updated; as suggested, we now filter the `metadata` through an allow-list rather than a block-list. However, for the root-level fields, I've set the filtering to remove only `status`, as there are too many root-level "speccy" fields aside from just `spec` on various k8s resources to reasonably create a comprehensive allow-list - see k8s `NetworkPolicy` API for instance.

rashmigottipati (approved):
/lgtm

PR #2328, opened 2025-11-12
:seedling: Use Patch instead of Update for finalizer operations

tmshort:
Refactor all controllers to use Patch() instead of Update()
when adding or removing finalizers to improve performance, and to avoid
removing non-cached fields erroneously.

This is necesary because we no longer cache the last-applied-configuration
annotation, so when we add/remove the finalizers, we are removing that field
from the metadata. This causes issues with clients when they don't see that
annotation (e.g. apply the same ClusterExtension twice).

Update ClusterCatalog and ClusterExtension controllers to use Patch-based
finalizer management (ClusterExtensionRevision already uses Patch())

This also restores the non-caching of "last-applied-config".

# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/cmd/operator-controller/main.go
+++ b/cmd/operator-controller/main.go
@@ -78,6 +78,7 @@ import (
 	"github.com/operator-framework/operator-controller/internal/operator-controller/rukpak/render/registryv1"
 	"github.com/operator-framework/operator-controller/internal/operator-controller/scheme"
 	sharedcontrollers "github.com/operator-framework/operator-controller/internal/shared/controllers"
+	cacheutil "github.com/operator-framework/operator-controller/internal/shared/util/cache"
 	fsutil "github.com/operator-framework/operator-controller/internal/shared/util/fs"
 	httputil "github.com/operator-framework/operator-controller/internal/shared/util/http"
 	imageutil "github.com/operator-framework/operator-controller/internal/shared/util/image"
@@ -257,6 +258,8 @@ func run() error {
 			cfg.systemNamespace: {LabelSelector: k8slabels.Everything()},
 		},
 		DefaultLabelSelector: k8slabels.Nothing(),
+		// Memory optimization: strip managed fields and large annotations from cached objects
+		DefaultTransform: cacheutil.StripAnnotations(),
 	}
 
 	if features.OperatorControllerFeatureGate.Enabled(features.BoxcutterRuntime) {
--- a/cmd/catalogd/main.go
+++ b/cmd/catalogd/main.go
@@ -59,6 +59,7 @@ import (
 	"github.com/operator-framework/operator-controller/internal/catalogd/storage"
 	"github.com/operator-framework/operator-controller/internal/catalogd/webhook"
 	sharedcontrollers "github.com/operator-framework/operator-controller/internal/shared/controllers"
+	cacheutil "github.com/operator-framework/operator-controller/internal/shared/util/cache"
 	fsutil "github.com/operator-framework/operator-controller/internal/shared/util/fs"
 	httputil "github.com/operator-framework/operator-controller/internal/shared/util/http"
 	imageutil "github.com/operator-framework/operator-controller/internal/shared/util/image"
@@ -254,6 +255,8 @@ func run(ctx context.Context) error {
 
 	cacheOptions := crcache.Options{
 		ByObject: map[client.Object]crcache.ByObject{},
+		// Memory optimization: strip managed fields and large annotations from cached objects
+		DefaultTransform: cacheutil.StripManagedFieldsAndAnnotations(),
 	}
 
 	saKey, err := sautil.GetServiceAccount()
--- a/internal/operator-controller/applier/boxcutter.go
+++ b/internal/operator-controller/applier/boxcutter.go
@@ -27,6 +27,7 @@ import (
 
 	ocv1 "github.com/operator-framework/operator-controller/api/v1"
 	"github.com/operator-framework/operator-controller/internal/operator-controller/labels"
+	"github.com/operator-framework/operator-controller/internal/shared/util/cache"
 )
 
 const (
@@ -66,6 +67,9 @@ func (r *SimpleRevisionGenerator) GenerateRevisionFromHelmRelease(
 		maps.Copy(labels, objectLabels)
 		obj.SetLabels(labels)
 
+		// Memory optimization: strip large annotations
+		// Note: ApplyStripTransform never returns an error in practice
+		_ = cache.ApplyStripAnnotationsTransform(&obj)
 		sanitizedUnstructured(ctx, &obj)
 
 		objs = append(objs, ocv1.ClusterExtensionRevisionObject{
@@ -117,6 +121,10 @@ func (r *SimpleRevisionGenerator) GenerateRevision(
 		unstr := unstructured.Unstructured{Object: unstrObj}
 		unstr.SetGroupVersionKind(gvk)
 
+		// Memory optimization: strip large annotations
+		if err := cache.ApplyStripAnnotationsTransform(&unstr); err != nil {
+			return nil, err
+		}
 		sanitizedUnstructured(ctx, &unstr)
 
 		objs = append(objs, ocv1.ClusterExtensionRevisionObject{
--- a/internal/shared/util/cache/transform.go
+++ b/internal/shared/util/cache/transform.go
@@ -1,0 +1,91 @@
+package cache
+
+import (
+	"maps"
+
+	toolscache "k8s.io/client-go/tools/cache"
+	crcache "sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// stripAnnotations removes memory-heavy annotations that aren't needed for controller operations.
+func stripAnnotations(obj interface{}) (interface{}, error) {
+	if metaObj, ok := obj.(client.Object); ok {
+		// Remove the last-applied-configuration annotation which can be very large
+		// Clone the annotations map to avoid modifying shared references
+		annotations := metaObj.GetAnnotations()
+		if annotations != nil {
+			annotations = maps.Clone(annotations)
+			delete(annotations, "kubectl.kubernetes.io/last-applied-configuration")
+			if len(annotations) == 0 {
+				metaObj.SetAnnotations(nil)
+			} else {
+				metaObj.SetAnnotations(annotations)
+			}
+		}
+	}
+	return obj, nil
+}
+
+// StripManagedFieldsAndAnnotations returns a cache transform function that removes
+// memory-heavy fields that aren't needed for controller operations.
+// This significantly reduces memory usage in informer caches by removing:
+// - Managed fields (can be several KB per object)
+// - kubectl.kubernetes.io/last-applied-configuration annotation (can be very large)
+//
+// Use this function as a DefaultTransform in controller-runtime cache.Options
+// to reduce memory overhead across all cached objects.
+//
+// Example:
+//
+//	cacheOptions := cache.Options{
+//	    DefaultTransform: cacheutil.StripManagedFieldsAndAnnotations(),
+//	}
+func StripManagedFieldsAndAnnotations() toolscache.TransformFunc {
+	// Use controller-runtime's built-in TransformStripManagedFields and compose it
+	// with our custom annotation stripping transform
+	managedFieldsTransform := crcache.TransformStripManagedFields()
+
+	return func(obj interface{}) (interface{}, error) {
+		// First strip managed fields using controller-runtime's transform
+		obj, err := managedFieldsTransform(obj)
+		if err != nil {
+			return obj, err
+		}
+
+		// Then strip the large annotations
+		return stripAnnotations(obj)
+	}
+}
+
+// StripAnnotations returns a cache transform function that removes
+// memory-heavy annotation fields that aren't needed for controller operations.
+// This significantly reduces memory usage in informer caches by removing:
+// - kubectl.kubernetes.io/last-applied-configuration annotation (can be very large)
+//
+// Use this function as a DefaultTransform in controller-runtime cache.Options
+// to reduce memory overhead across all cached objects.
+//
+// Example:
+//
+//	cacheOptions := cache.Options{
+//	    DefaultTransform: cacheutil.StripAnnotations(),
+//	}
+func StripAnnotations() toolscache.TransformFunc {
+	return func(obj interface{}) (interface{}, error) {
+		// Strip the large annotations
+		return stripAnnotations(obj)
+	}
+}
+
+// ApplyStripAnnotationsTransform applies the strip transform directly to an object.
+// This is a convenience function for cases where you need to strip fields
+// from an object outside of the cache transform context.
+//
+// Note: This function never returns an error in practice, but returns error
+// to satisfy the TransformFunc interface.
+func ApplyStripAnnotationsTransform(obj client.Object) error {
+	transform := StripAnnotations()
+	_, err := transform(obj)
+	return err
+}


tmshort:
The ClusterExtension and ClusterCatalog finalizer code used Update, the ClusterExtensionRevision used Apply, so the CE and CE finalizer code was rewritten to be like CER. There is common finalizer code between them now.

tmshort:
Ping @joelanford

tmshort:
catalogd pod is panicing...  it only fails due to the summary generated, it's not "noticeable" during a regular run!

pedjak:
> This is necesary because we no longer cache thelast-applied-configuration annotation, so when we add/remove the finalizers, we are removing that field from the metadata. This causes issues with clients when they don't see that annotation (e.g. apply the same ClusterExtension twice).

If we want to remove conflicts, we should use patch, but adding/removing finalizer we do not touch removed fields (including the annotation). Controller-runtime cache is write-through cache, so an update or patch operation does not perform any direct modification on the cache - all changes are made by the informer.

In what situations did we see conflicts, given that only single controller is responsible for a resource type?

tmshort:
> If we want to remove conflicts, we should use patch, but adding/removing finalizer we do not touch removed fields (including the annotation). Controller-runtime cache is write-through cache, so an update or patch operation does not perform any direct modification on the cache - all changes are made by the informer.
> 
> In what situations did we see conflicts, given that only single controller is responsible for a resource type?

This is not a conflict issue. The problem was that we are using Update(), which uses the cached version of the resource to make the changes. The cached version no longer includes the `last-applied-configuration` annotation, so that annotation is removed on the Update().

This causes a problem with e.g. kubectl applying a ClusterExtension a second time. If you apply a CE twice without this code (i.e. current `main` branch), you will see the following:
```
tshort@cube:~/.../config/samples (main %=)$ kubectl apply -f ce.yaml
clusterextension.olm.operatorframework.io/argocd created
tshort@cube:~/.../config/samples (main %=)$ oc get clusterextension -o yaml
apiVersion: v1
items:
- apiVersion: olm.operatorframework.io/v1
  kind: ClusterExtension
  metadata:
    creationTimestamp: "2025-11-12T15:22:25Z"
    finalizers:
    - olm.operatorframework.io/cleanup-unpack-cache
    - olm.operatorframework.io/cleanup-contentmanager-cache
    generation: 1
    name: argocd
    resourceVersion: "1296"
    uid: b096377d-577e-4d25-8f3d-b0e0b12fe894
...
tshort@cube:~/.../config/samples (main %=)$ kubectl apply -f ce.yaml
Warning: resource clusterextensions/argocd is missing the kubectl.kubernetes.io/last-applied-configuration annotation which is required by kubectl apply. kubectl apply should only be used on resources created declaratively by either kubectl create --save-config or kubectl apply. The missing annotation will be patched automatically.
clusterextension.olm.operatorframework.io/argocd configured
tshort@cube:~/.../config/samples (main %=)$
```
The problem is that the `last-applied-configuration` annotation is not present. Thus, causing the second `kubectl apply` command to complain. The expected result is that the second `kubectl apply` states `clusterextension.olm.operatorframework.io/argocd unchanged`

internal/shared/util/finalizer/finalizer.go
```diff
@@ -0,0 +1,91 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+// EnsureFinalizer adds a finalizer to the object using Patch.
+// If the finalizer already exists, this is a no-op and returns (false, nil).
+// Returns (true, nil) if the finalizer was added.
+// Note: This function will update the passed object with the server response.
+func EnsureFinalizer(ctx context.Context, c client.Client, obj client.Object, finalizer string) (bool, error) {
```
joelanford: If we want to add multiple finalizers, it will mean multiple Patch calls to the apiserver as this is written. Perhaps we should have this function accept multiple finalizers so that we can limit the number of patch calls (and watch events produced/received by apiserver/informers)?

internal/shared/util/finalizer/finalizer.go
```diff
@@ -0,0 +1,91 @@
...
+	if controllerutil.ContainsFinalizer(obj, finalizer) {
+		return false, nil
+	}
+
+	// Create a copy of the finalizers list to avoid mutating the object
+	currentFinalizers := obj.GetFinalizers()
+	newFinalizers := make([]string, len(currentFinalizers), len(currentFinalizers)+1)
+	copy(newFinalizers, currentFinalizers)
+	newFinalizers = append(newFinalizers, finalizer)
```
joelanford: Sort the new finalizers to ensure deterministic ordering?

internal/shared/util/finalizer/finalizer.go
```diff
@@ -0,0 +1,91 @@
...
+			"resourceVersion": obj.GetResourceVersion(),
+			"finalizers":      newFinalizers,
+		},
+	}
+	patchJSON, err := json.Marshal(patch)
+	if err != nil {
+		return false, fmt.Errorf("marshalling patch to add finalizer: %w", err)
+	}
+	// Note: Patch will update obj with the server response, including ResourceVersion
+	if err := c.Patch(ctx, obj, client.RawPatch(types.MergePatchType, patchJSON)); err != nil {
```
joelanford: Should we use SSA here?

internal/shared/util/finalizer/finalizer.go
```diff
@@ -0,0 +1,91 @@
...
+		return false, fmt.Errorf("adding finalizer: %w", err)
+	}
+	return true, nil
+}
+
+// RemoveFinalizer removes a finalizer from the object using Patch.
+// If the finalizer doesn't exist, this is a no-op.
+func RemoveFinalizer(ctx context.Context, c client.Client, obj client.Object, finalizer string) error {
```
joelanford: Same comment here: re accept multiple finalizers to avoid multiple patch calls?

internal/shared/util/finalizer/finalizer.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	// Create a copy of the finalizers list to avoid mutating the object
+	currentFinalizers := obj.GetFinalizers()
+	newFinalizers := make([]string, 0, len(currentFinalizers))
+	for _, f := range currentFinalizers {
+		if f != finalizer {
+			newFinalizers = append(newFinalizers, f)
+		}
+	}
```
joelanford: `newFinalizers := slices.Delete(slices.Clone(currentFinalizers), finalizer)`?

joelanford: Or if we reorganize to accept multiple finalizers, then `slices.DeleteFunc` could be more appropriate (e.g. by building a set of finalizers to remove and then checking set presence in the predicate function).

pedjak:
> This is not a conflict issue. The problem was that we are using Update(), which uses the cached version of the resource to make the changes. The cached version no longer includes the `last-applied-configuration` annotation, so that annotation is removed on the Update().

Thanks for explaining it, make sense. Given that I think we should update PR description to reflect that, currently we have:

> Refactor all controllers to use client.Patch() instead of Update() when adding or removing finalizers to reduce conflicts and improve performance. 

Given the explanation above, we are not reducing conflict or improving performances - we are ensuring correctness, i.e. not removing the annotation.

internal/shared/util/finalizer/finalizer.go
pedjak: that could be a good improvement,  that would simplify the reconcile logic a lot, i.e. we do not need to care about appending/sorting finalizers at all.

cmd/operator-controller/main.go
```diff
@@ -391,12 +389,11 @@ func run() error {
...
-	if err := clusterExtensionFinalizers.Register(controllers.ClusterExtensionCleanupUnpackCacheFinalizer, finalizers.FinalizerFunc(func(ctx context.Context, obj client.Object) (crfinalizer.Result, error) {
-		return crfinalizer.Result{}, imageCache.Delete(ctx, obj.GetName())
-	})); err != nil {
-		setupLog.Error(err, "unable to register finalizer", "finalizerKey", controllers.ClusterExtensionCleanupUnpackCacheFinalizer)
-		return err
+	// Set up finalizer handlers for ClusterExtension
+	clusterExtensionFinalizerHandlers := map[string]controllers.FinalizerHandler{
+		controllers.ClusterExtensionCleanupUnpackCacheFinalizer: func(ctx context.Context, ext *ocv1.ClusterExtension) error {
+			return imageCache.Delete(ctx, ext.GetName())
+		},
```
pedjak: given that we use single finalizer, was it needed to move to a map of single entry?

internal/catalogd/controllers/core/clustercatalog_controller.go
```diff
@@ -171,32 +150,47 @@ func (r *ClusterCatalogReconciler) reconcile(ctx context.Context, catalog *ocv1.
...
+			return ctrl.Result{}, fmt.Errorf("error removing finalizer: %v", err)
+		}
+
+		// Set status.conditions[type=Progressing] to True as we are done with
+		// all that needs to be done with the catalog
+		updateStatusProgressingUserSpecifiedUnavailable(&catalog.Status, catalog.GetGeneration())
+		// Clear URLs, ResolvedSource, and LastUnpacked since catalog is unavailable
+		catalog.Status.ResolvedSource = nil
+		catalog.Status.URLs = nil
+		catalog.Status.LastUnpacked = nil
```
pedjak: not sure if this change is in the focus of PR theme, perhaps we should extract it into a separate PR?

internal/catalogd/controllers/core/clustercatalog_controller.go
```diff
@@ -419,30 +413,16 @@ func (r *ClusterCatalogReconciler) needsPoll(lastSuccessfulPoll time.Time, catal
...
-		err := r.deleteCatalogCache(ctx, catalog)
-		return crfinalizer.Result{StatusUpdated: true}, err
-	}))
-	if err != nil {
-		return err
+	a.ManagedFields, b.ManagedFields = nil, nil
+	a.ResourceVersion, b.ResourceVersion = "", ""
+	// Remove kubectl's last-applied-configuration annotation which may be added by the API server
+	if a.Annotations != nil {
+		delete(a.Annotations, "kubectl.kubernetes.io/last-applied-configuration")
```
pedjak: declare the annotation key as the constant

internal/catalogd/controllers/core/clustercatalog_controller.go
```diff
@@ -419,30 +413,16 @@ func (r *ClusterCatalogReconciler) needsPoll(lastSuccessfulPoll time.Time, catal
...
-		if !ok {
-			panic("could not convert object to clusterCatalog")
-		}
-		err := r.deleteCatalogCache(ctx, catalog)
-		return crfinalizer.Result{StatusUpdated: true}, err
-	}))
-	if err != nil {
-		return err
+	a.ManagedFields, b.ManagedFields = nil, nil
+	a.ResourceVersion, b.ResourceVersion = "", ""
```
pedjak: instead of nulling fields, I would actually just compare resource parts where we do not expect changes, in this case is probably `a.Spec` and `b.Spec`? Metadata and statuses can be updated by many actors, but we would like to check if `equality.Semantic.DeepQual(a.Spec, b.Spec)` is false.

internal/operator-controller/controllers/clusterextension_controller.go
```diff
@@ -73,9 +74,12 @@ type ClusterExtensionReconciler struct {
 	StorageMigrator      StorageMigrator
 	Applier              Applier
 	RevisionStatesGetter RevisionStatesGetter
-	Finalizers           crfinalizer.Finalizers
+	FinalizerHandlers    map[string]FinalizerHandler
```
pedjak: a controller should set single finalizer, more than one is strange. I would suggest reverting the change.

internal/shared/util/finalizer/finalizer.go
```diff
@@ -0,0 +1,137 @@
...
+
+	// Create an unstructured object for server-side apply
+	u := &unstructured.Unstructured{}
+	u.SetGroupVersionKind(gvk)
+	u.SetName(obj.GetName())
+	u.SetNamespace(obj.GetNamespace())
+	u.SetFinalizers(newFinalizers)
+
+	// Use server-side apply to update finalizers
+	if err := c.Patch(ctx, u, client.Apply, client.ForceOwnership, client.FieldOwner("finalizer-controller")); err != nil {
```
pedjak: each controller (ClusterExtension, Catalog, ClusterExtensionRevisio/n) should has its own field owner, following this convention:

```
olm.operatorframework.io/<CONTROLLER-NAME>
```

for example: 
* `olm.operatorframework.io/clusterextension-controller`
* `olm.operatorframework.io/clustercatalog-controller`
* `olm.operatorframework.io/clusterextensionrevision-controller`

internal/shared/util/finalizer/finalizer.go
```diff
@@ -0,0 +1,137 @@
...
+	if err != nil {
+		return false, fmt.Errorf("getting object kind: %w", err)
+	}
+
+	// Create an unstructured object for server-side apply
+	u := &unstructured.Unstructured{}
+	u.SetGroupVersionKind(gvk)
+	u.SetName(obj.GetName())
+	u.SetNamespace(obj.GetNamespace())
+	u.SetFinalizers(newFinalizers)
```
pedjak: it has to be `u.SetFinalizers(finalizers)`

internal/shared/util/finalizer/finalizer.go
```diff
@@ -0,0 +1,137 @@
...
+
+	// Use server-side apply to update finalizers
+	if err := c.Patch(ctx, u, client.Apply, client.ForceOwnership, client.FieldOwner("finalizer-controller")); err != nil {
+		return false, fmt.Errorf("adding finalizer: %w", err)
+	}
+
+	// Update the passed object with the new finalizers
+	obj.SetFinalizers(newFinalizers)
```
pedjak: it has to get `obj.SetFinalizers(u.GetFinalizers())

internal/shared/util/finalizer/finalizer.go
```diff
@@ -0,0 +1,137 @@
...
+	obj.SetResourceVersion(u.GetResourceVersion())
+
+	return true, nil
+}
+
+// RemoveFinalizer removes one or more finalizers from the object using server-side apply.
+// If none of the finalizers exist, this is a no-op.
+func RemoveFinalizer(ctx context.Context, c client.Client, obj client.Object, finalizers ...string) error {
```
pedjak: no need to have this function because `EnsureFinalizer(ctx, c, obj)` is going to remove previously added finalizers that we owned.

internal/catalogd/controllers/core/clustercatalog_controller.go
joelanford: I'd probably also include metadata labels and annotations in the "don't expect to our reconciler to change them" list. Those are in domain of actors other than the reconciler of the spec.

internal/operator-controller/controllers/clusterextension_controller.go
joelanford: I think there is more than one for CEs right now because there are multiple finalizer concerns:
1. Cleaning up the unpacked bundle image directory
2. Shutting down the content manager informers (helm-only)

I'm not sure how the current helm-to-boxcutter migration deals with the content manager informer shutdown. Do those informers get cleaned up when boxcutter takes over or when the CE is deleted (potentially much later). Ideally we'd shutdown the helm-only content manager informers as soon as boxcutter takes over.

joelanford: (but that's getting a bit outside the scope of this PR)

internal/catalogd/controllers/core/clustercatalog_controller.go
tmshort: Yeah, this section partially duplicates code above, and is also not necessarily needed.

internal/operator-controller/controllers/clusterextension_controller.go
tmshort: In this case, there ARE two pre-existing finalizers on the ClusterExtension

internal/shared/util/finalizer/finalizer.go
tmshort: I would I would suggest that because the finalizers are being treated "separately" that they have a different owner referencing "finalizer".
That being said, this is generic code shared among all three... I'll have to add another argument.

cmd/operator-controller/main.go
tmshort: The ClusterExtension has two finalizers.
See L559 and L672 of cmd/operator-controller/main.go

internal/shared/util/finalizer/finalizer.go
tmshort: No. this is based on the assumption that the arguments to `EnsureFinalizers` represent the _only_ finalizers we want in the list. This is not the intent.

The EnsureFinalizers API ensures that the given finalizers are present. The intent is _not_ to set the finalizers to _only_ be those given, and is not intended to _remove_ finalizers.

Because this is a list, we can't patch individual items into the list, we need to specify the whole list (at least that's how the SSA API is working).  I verified this in the code by calling `EnsureFinalizers()` separately for each finalizer string, rather than in bulk. Only the last finalizer showed up when it was `u.SetFinalizers(finalizers)`. Both finalizers showed up when the code was as above.

Yes, I realize the name `EnsureFinalizers` is a bit confusing, and should probably be `AddFinalizers`, but I was following the naming scheme (and behavior) of the removed code, which was inconsistent.

internal/shared/util/finalizer/finalizer.go
tmshort: The intent of `EnsureFinalizers` is to _only_ add finalizers to the list of finalizers, not to remove them. The `RemoveFinalizers` removes finalizers individually.

internal/operator-controller/controllers/clusterextension_controller.go
tmshort: @joelanford 
>I'm not sure how the current helm-to-boxcutter migration deals with the content manager informer shutdown. Do those informers get cleaned up when boxcutter takes over or when the CE is deleted (potentially much later). Ideally we'd shutdown the helm-only content manager informers as soon as boxcutter takes over.

In order to switch from the Helm applier to the Boxcutter applier, the pod has to restart, so everything is effectively shutdown. The content-manager finalizer is a no-op under Boxcutter, but the finalizer still needs to be removed.

internal/catalogd/controllers/core/clustercatalog_controller.go
tmshort: OK, I misinterpreted this comment... Are we saying to just compare:
* .spec
* .metadata.labels
* .metadata.annotations
??

internal/shared/util/finalizer/finalizer.go
pedjak: > Because this is a list, we can't patch individual items into the list, we need to specify the whole list (at least that's how the SSA API is working). 

I need to disagree here, because Finalizer is defined as a set with merge patch stratefy:

https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/types.go#L256-L259

Therefore, we do not patch item by item, we set all finalizers we need to own and make SSA requests.

BTW, my understanding here is that `finalizers` argument contains all finalizers we would like to set, not just one.

internal/catalogd/controllers/core/clustercatalog_controller.go
tmshort: This makes more sense, since we're not copying around data structures, and then deleting elements, we're only doing comparisons... 
Note: those changes were added to make the cluster catalog check similar to the other controllers, because I was experiencing panics due to the API server changing the `ResourceVersion`.

internal/shared/util/finalizer/finalizer.go
pedjak: in SSA approach, you are not remove things by omitting them in request, if you owned them previously. Hence, if we had somewhere in the code:

```golang
EnsureFinalizers(ctx, client, ce, "foo", "bar")
```

to remove `bar`, we can perform:

```golang
EnsureFinalizers(ctx, client, ce, "foo")
```

tmshort: But that is not the intent of the API; it's to add items individually; not to remove. Just because the underlying mechanism is SSA, doesn't mean the layered API has to behave that way.

IMHO, It's confusing to have "Ensure" remove items, "Ensure" means to check that something is true or correct. In this case "Ensure" is confusing, so I'm renaming the function.

internal/shared/util/finalizer/finalizer.go
tmshort: >Therefore, we do not patch item by item, we set all finalizers we need to own and make SSA requests.

Agreed, that's how SSA works, but that's not how the API layer above it has to work.

The function as defined _adds_ finalizers, that's the intent. The intent is not to make the finalizer list equal to the arguments. 

I'm renaming the function to indicate the desired (existing) behavior. It is confusing to have the mechanism to remove finalizers as "ensuring nothing".

If we wanted the behavior as you suggest, then `SetFinalizers()` would be more appropriate. But it would be at the cost of flexibility, and potential issues if _other_ controllers add their own finalizers.

tmshort:
~~After digging into the solutions (Add/Ensure/Remove/Update/Set), none of them are optimal. Specifically, if another controller adds a finalizer to our resources, we see it, and that messes with our finalizer management. This means we hit the API server with a SSA more than we should.~~

~~If a single-function mechanism (e.g. Set/Ensure) is used, then on _every_ reconcile we'll see the extra finalizer and attempt to set the finalizers to the input list (which may be empty) which attempts to remove the extra finalizer, and because we don't own it, it will never be removed. So any checks to ensure finalizers are set as we expect will fail, forcing an API server call.~~

~~I prefer the explicit add and remove mechanism, as in this current PR. It explicitly checks the current list of finalizers and only cares about the ones that are input. It will use the existing list of finalizers to add/remove, and will only issue a SSA if changes are needed; it will ignore the extra finalizer. However, in the unlikely case where the cache is out-of-sync, then we will hit the API server with a SSA until the cache is synced. It's better than every reconcile as in the single-function case.~~

EDIT: All our finalizers use the same prefix, other controllers' finalizers _should_ use a different prefix. So, now that I thought about it some more, I think a single `Ensure` function will work. However, sorting is still required to check if the finalizers are in the right state, such that we reduce the number of Server API calls.

pedjak:
@tmshort apologize I did not mention it earlier: moving to SSA requires also to migrate the information stored under `.metadata.managedField` - otherwise field removal would not be possible, see for example:

https://github.com/kubernetes/kubernetes/issues/99003

In our case, it is easy to reproduce:
* deploy OLM from `main` branch, create a `ClusterExtension`. Under `.metadata.managedField` we have now:

```yaml
- apiVersion: olm.operatorframework.io/v1                                                                                                                                                                                         
  fieldsType: FieldsV1                                                                                                                                                                                                            
  fieldsV1:                                                                                                                                                                                                                       
       f:metadata:
         f:finalizers:                                                                                                                                                                                                               
           .: {}                                                                                                                                                                                                                     
           v:"olm.operatorframework.io/cleanup-contentmanager-cache": {}                                                                                                                                                             
           v:"olm.operatorframework.io/cleanup-unpack-cache": {}                                                                                                                                                                     
  manager: operator-controller                                                                                                                                                                                                    
  operation: Update                                 
```

This is expected, because we have added finalizers via `Update` call.

* Now, build and deploy OLM using this PR and try to remove the existing `ClusterExtension`. The operation is not going to be successful, even if there is no errors to be see in the logs. The controller is not able to remove finalizers, because we now use operation `Apply` and the combination of `manager` and `operation` determines the field owner. Clearly, given that we do not provide `finalizers`, we are not able to remove it because we are not the owners.

To fix that we would need to migrate the ownership under `.metadata.managedFields`. However, it looks like a way more work for the cause that needs to be fixed: removal of `last-applied-configuration` annotation to save a bit of memory.

Instead of going to through SSA migration, I would suggest that we revert removal of that annotation as the quick fix, and then introduce SSA if needed + detailed testing.

tmshort:
/hold
Yes, this does not take into account migration of pre-existing resources. I could switch it back to using Patch(), as moving to SAA with existing fields will give us some trouble.

That being said, there is currently a failure in the test-experimental-e2e, which seems to be triggered by this change; since I'm not seeing the issue on main. I would need to fix that before this could be merged.

I will point out that managed fields is not a "bit of memory", but can double the size of the cache. The memory savings was significant.

pedjak:
> I will point out that managed fields is not a "bit of memory", but can double the size of the cache. The memory savings was significant.

I agree, but managed fields are readded in the cache with https://github.com/operator-framework/operator-controller/pull/2318

tmshort:
> > I will point out that managed fields is not a "bit of memory", but can double the size of the cache. The memory savings was significant.
> 
> I agree, but managed fields are readded in the cache with #2318

Whoops, I meant last-applied-config.

tmshort:
Reworked (i.e. rebased to `main`) to use Patch only for the CE and CC controllers, since CER already used Patch. This is a _much_ smaller change than before, as I didn't attempt to create a common finalizer infrastructure.

internal/catalogd/controllers/core/clustercatalog_controller.go
```diff
@@ -125,10 +127,17 @@ func (r *ClusterCatalogReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+		// Use patch to update finalizers on the server
+		patch := map[string]any{
+			"metadata": map[string]any{
+				"finalizers": finalizers,
+			},
+		}
+		patchJSON, err := json.Marshal(patch)
+		if err != nil {
+			reconcileErr = errors.Join(reconcileErr, fmt.Errorf("marshalling patch to ensure finalizers: %w", err))
+		} else if err := r.Patch(ctx, reconciledCatsrc, client.RawPatch(types.MergePatchType, patchJSON)); err != nil {
```
pedjak: How about that we use already available controller-runtime's [`CreateOrPatch`](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/controller/controllerutil/controllerutil.go#L382)?

internal/operator-controller/controllers/clusterextension_controller.go
```diff
@@ -142,18 +143,26 @@ func (r *ClusterExtensionReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
+		// Use patch to update finalizers on the server
+		patch := map[string]any{
+			"metadata": map[string]any{
+				"finalizers": finalizers,
+			},
+		}
+		patchJSON, err := json.Marshal(patch)
+		if err != nil {
+			reconcileErr = errors.Join(reconcileErr, fmt.Errorf("marshalling patch to ensure finalizers: %w", err))
+		} else if err := r.Patch(ctx, reconciledExt, client.RawPatch(types.MergePatchType, patchJSON)); err != nil {
```
pedjak: same question as above.

internal/operator-controller/controllers/clusterextension_controller.go
```diff
@@ -179,11 +188,15 @@ func ensureAllConditionsWithReason(ext *ocv1.ClusterExtension, reason v1alpha1.C
 	}
 }
 
-// Compare resources - ignoring status & metadata.finalizers
+// Compare resources - annotations, labels and spec
 func checkForUnexpectedClusterExtensionFieldChange(a, b ocv1.ClusterExtension) bool {
```
pedjak: [nit] perhaps we could come up with a single `checkForUnexpectedFieldChange` that works with all types, to avoid code duplication.

tmshort: I had thought about it, they are different types, although that could be handled, I just think it's a bit out-of-scope here.

internal/catalogd/controllers/core/clustercatalog_controller.go
tmshort: Possibly overkill (looking at the size of the function), but I can look.

pedjak: > Possibly overkill (looking at the size of the function), but I can look.

I would create `RawPatch`, only if really needed. `CreateOrPatch()` covers us both in updating finalizers and status and could simplify the code a lot, i.e. we do not need to check if statuses/finalizers are changed during the reconcile, that function does it all. The whole `updateFinalizers { ... }` block could be replaced with something like:

```golang
if _, err := controllerutil.CreateOrPatch(ctx, r.Client, &existingCatsrc, func() error {
		existingCatsrc.Finalizers = reconciledCatsrc.Finalizers
		return nil
	}); err != nil {
		return ctrl.Result{}, err
	}
```

and we could also go beyond that (slightly out of the PR scope, could be done separately) and also to drop `if updateStatus {...}` block with:

```golang
if _, err := controllerutil.CreateOrPatch(ctx, r.Client, &existingCatsrc, func() error {
		existingCatsrc.Finalizers = reconciledCatsrc.Finalizers
		existingCatsrc.Status = reconciledCatsrc.Status
		return nil
	}); err != nil {
		return ctrl.Result{}, err
	}
```

tmshort: >Possibly overkill (looking at the size of the function), but I can look.

As in the size of `CreateOrPatch()`, that is...

Dammit, I forgot to push my changes last night,...

internal/shared/util/k8s/k8s.go
tmshort, line 1: Not a fan of this package name, but I couldn't think of anything better? Suggestions welcome.

api/v1/helpers.go
```diff
@@ -0,0 +1,19 @@
...
+// This method is provided to satisfy generic interfaces that require a GetSpec() method.
+func (c *ClusterCatalog) GetSpec() ClusterCatalogSpec {
+	return c.Spec
+}
+
+// GetSpec returns the Spec field of the ClusterExtensionRevision.
+// This method is provided to satisfy generic interfaces that require a GetSpec() method.
+func (c *ClusterExtensionRevision) GetSpec() ClusterExtensionRevisionSpec {
+	return c.Spec
+}
```
pedjak: It is recommended to avoid adding helper functions:

https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md#no-functions

internal/shared/util/k8s/k8s.go
```diff
@@ -0,0 +1,41 @@
...
+	"k8s.io/apimachinery/pkg/api/equality"
+)
+
+// ObjectWithSpec represents any Kubernetes custom resource that embeds metav1.ObjectMeta
+// and has a Spec field. The type parameter T should be the Spec type.
+type ObjectWithSpec[T any] interface {
+	GetAnnotations() map[string]string
+	GetLabels() map[string]string
+	GetSpec() T
+}
```
pedjak: How about that we do not introduce this interface and get access to `.Spec` via reflection api? with that `CheckForUnexpectedFieldChange` signature could become just:

```golang
func CheckForUnexpectedFieldChange[T metav1.Object](a, b T) bool
```
and we do not need to implement any additional interface on our API.

tmshort: I hear you. I will note that most references to the reflect API indicate a performance penalty...

pedjak: > I hear you. I will note that most references to the reflect API indicate a performance penalty...

ack, but it really depends on applied domain. Control plane side (resource  reconciliation) is not a high-throughput/low-latency environment. Reflection is used already on many places in [controller-runtime](https://github.com/search?q=repo%3Akubernetes-sigs%2Fcontroller-runtime+reflect.&type=code) and we use it in `CheckForUnexpectedFieldChange()` as well calling `reflect.DeepEqual()`. 

IMO, before the reflection usage becomes a bottleneck, there are other places in the current code where we can achieve more significant savings.

internal/shared/util/k8s/k8s.go
```diff
@@ -0,0 +1,57 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// CheckForUnexpectedFieldChange compares two Kubernetes objects and returns true
+// if their annotations, labels, or spec have changed. This is useful for detecting
+// unexpected modifications during reconciliation.
+//
+// The function compares:
+//   - Annotations (via GetAnnotations)
+//   - Labels (via GetLabels)
```
pedjak, line 16: for the completeness we use reflection for comparing annotations and labels as well.

internal/shared/util/k8s/k8s.go
```diff
@@ -0,0 +1,57 @@
...
+//   - Spec (using reflection to access the Spec field, with semantic equality)
+//
+// Status and finalizers are intentionally not compared, as these are expected
+// to change during reconciliation.
+//
+// Type parameter:
+//   - O: The object type that implements metav1.Object and has a Spec field
```
pedjak: Given that O is interface, no need to declare a generic function, we can do it like:

```golang
func CheckForUnexpectedFieldChange(a, b metav1.Object) bool {
```

internal/shared/util/k8s/k8s.go
tmshort: Since it's an `metav1.Object type`, it's using accessors for those values for the comparison; it's not using reflection. There's no need to use reflection there. Are you asking me to switch, because it doesn't seem necessary.

pedjak: > Since it's an `metav1.Object type`, it's using accessors for those values for the comparison; it's not using reflection. There's no need to use reflection there. 

Once the labels and annotations are retrieved, they are compared using `equality.Semantic.DeepEqual()`  which uses the reflection under the hood.

> Are you asking me to switch, because it doesn't seem necessary.

[nit]: I am not sure if we should document how to we access the fields that are compared (i.e. what underlying accessor).

tmshort:
/approve
There are two LGTMs on it.

PR #2329, opened 2025-11-12
:bug: Fix Boxcutter manifest ordering inconsistency

tmshort:
Issue:
Manifest ordering inconsistency: CRDs from Helm release manifest and
bundle manifest appeared in different orders, causing PhaseSort to
produce different phase structures even though they contained the same
objects.

Solution:
Added deterministic sorting in PhaseSort (phase.go):
 - Sort objects within each phase by Group, Version, Kind, Namespace, Name
 - Ensures consistent phase structure regardless of input order
 - Critical for comparing revisions from different sources

# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/operator-controller/applier/phase.go
+++ b/internal/operator-controller/applier/phase.go
@@ -1,6 +1,9 @@
 package applier
 
 import (
+	"cmp"
+	"slices"
+
 	"k8s.io/apimachinery/pkg/runtime/schema"
 
 	ocv1 "github.com/operator-framework/operator-controller/api/v1"
@@ -111,6 +114,23 @@ func init() {
 	}
 }
 
+// Sort objects within the phase deterministically by Group, Version, Kind, Namespace, Name
+// to ensure consistent ordering regardless of input order. This is critical for
+// Helm-to-Boxcutter migration where the same resources may come from different sources
+// (Helm release manifest vs bundle manifest) and need to produce identical phases.
+func compareClusterExtensionRevisionObjects(a, b ocv1.ClusterExtensionRevisionObject) int {
+	aGVK := a.Object.GroupVersionKind()
+	bGVK := b.Object.GroupVersionKind()
+
+	return cmp.Or(
+		cmp.Compare(aGVK.Group, bGVK.Group),
+		cmp.Compare(aGVK.Version, bGVK.Version),
+		cmp.Compare(aGVK.Kind, bGVK.Kind),
+		cmp.Compare(a.Object.GetNamespace(), b.Object.GetNamespace()),
+		cmp.Compare(a.Object.GetName(), b.Object.GetName()),
+	)
+}
+
 // PhaseSort takes an unsorted list of objects and organizes them into sorted phases.
 // Each phase will be applied in order according to DefaultPhaseOrder. Objects
 // within a single phase are applied simultaneously.
@@ -125,6 +145,9 @@ func PhaseSort(unsortedObjs []ocv1.ClusterExtensionRevisionObject) []ocv1.Cluste
 
 	for _, phaseName := range defaultPhaseOrder {
 		if objs, ok := phaseMap[phaseName]; ok {
+			// Sort objects within the phase deterministically
+			slices.SortFunc(objs, compareClusterExtensionRevisionObjects)
+
 			phasesSorted = append(phasesSorted, ocv1.ClusterExtensionRevisionPhase{
 				Name:    string(phaseName),
 				Objects: objs,
--- a/internal/operator-controller/applier/phase_test.go
+++ b/internal/operator-controller/applier/phase_test.go
@@ -262,16 +262,16 @@ func Test_PhaseSort(t *testing.T) {
 						{
 							Object: unstructured.Unstructured{
 								Object: map[string]interface{}{
-									"apiVersion": "apps/v1",
-									"kind":       "Deployment",
+									"apiVersion": "v1",
+									"kind":       "ConfigMap",
 								},
 							},
 						},
 						{
 							Object: unstructured.Unstructured{
 								Object: map[string]interface{}{
-									"apiVersion": "v1",
-									"kind":       "ConfigMap",
+									"apiVersion": "apps/v1",
+									"kind":       "Deployment",
 								},
 							},
 						},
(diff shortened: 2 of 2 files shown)


joelanford:
> CollisionProtection mismatch: The migrated revision had collisionProtection=None (needed to adopt Helm-managed resources) while the bundle-generated revision had collisionProtection=Prevent (the default value).

I think this is intentional. We need `None` in order to adopt the existing objects. But then we want `Prevent` to ensure that we _stop_ automatically adopting those objects in the future. I'm pretty sure that @thetechnick intended the adoption mechanics of `None` to be a one-time thing.

perdasilva:
> > CollisionProtection mismatch: The migrated revision had collisionProtection=None (needed to adopt Helm-managed resources) while the bundle-generated revision had collisionProtection=Prevent (the default value).
> 
> I think this is intentional. We need `None` in order to adopt the existing objects. But then we want `Prevent` to ensure that we _stop_ automatically adopting those objects in the future. I'm pretty sure that @thetechnick intended the adoption mechanics of `None` to be a one-time thing.

I think the original behavior makes sense. We shouldn't keep `None` beyond the initial migration to assume a conservative posture by default going forward. Leaving it as `None` opens the cluster to possible unforeseen pain in the future. I'd also say it's fine to generate a new revision with the new collisionProtection setting to keep an audit trail.

internal/operator-controller/applier/phase.go
```diff
@@ -111,6 +113,57 @@ func init() {
 	}
 }
 
+// Sort objects within the phase deterministically by Group, Version, Kind, Namespace, Name
+// to ensure consistent ordering regardless of input order. This is critical for
+// Helm-to-Boxcutter migration where the same resources may come from different sources
+// (Helm release manifest vs bundle manifest) and need to produce identical phases.
+func compareClusterExtensionRevision(a, b ocv1.ClusterExtensionRevisionObject) int {
```
perdasilva:
```suggestion
func compareClusterExtensionRevisionObject(a, b ocv1.ClusterExtensionRevisionObject) int {
```

perdasilva: or even just compareRevisionObject ?

tmshort:
> > > CollisionProtection mismatch: The migrated revision had collisionProtection=None (needed to adopt Helm-managed resources) while the bundle-generated revision had collisionProtection=Prevent (the default value).
> > 
> > 
> > I think this is intentional. We need `None` in order to adopt the existing objects. But then we want `Prevent` to ensure that we _stop_ automatically adopting those objects in the future. I'm pretty sure that @thetechnick intended the adoption mechanics of `None` to be a one-time thing.
> 
> I think the original behavior makes sense. We shouldn't keep `None` beyond the initial migration to assume a conservative posture by default going forward. Leaving it as `None` opens the cluster to possible unforeseen pain in the future. I'd also say it's fine to generate a new revision with the new collisionProtection setting to keep an audit trail.

So, removing either of the changes will cause the bug (two CERs created) to assert itself. If we are OK with this, then we don't necessarily need this fix, although sorting the objects during the phase is probably still a good idea. @dtfranz and I found the same solution to the problem:

>On its own, the manifest sanitization does not completely fix the other issue, but it does make it quite a bit better. Following the repro steps in OCPBUGS-62943 on main, I get infinite additional CERs. When I add my fix for OPRUN-4238, I only get one additional CER. This extra CER is being created due to the diff between revisions created by GenerateRevisionFromHelmRelease and GenerateRevision; the former sets the CollisionProtection flag, and the latter does not. The objects within the phases may also need to be sorted. If you resolve those differences, the extra CER does not get generated. I'm not 100% sure of the fix method (or, as @perdasilva mentioned, if it's even an issue that needs fixing), but I'm pretty sure we need to disregard the CollisionProtection flag when comparing observed vs desired CERs. If we do that then the controller should understand that no new revision is required.

If we're good with having two initial CERs upon transition from the helm applier to the boxcutter applier, then I can reduce this to be just the sorting changes, and remove the CollisionProtection change. I would then close the bug as intentional.

internal/operator-controller/applier/phase.go
tmshort: I have a rename for this, but I'm waiting on it, in case the decision is that we don't want the CollisionProtection change.

tmshort:
@joelanford @dtfranz giving me 👍 doesn't really tell me what you want! Do we want to fix the two CER issue or just fix the sorting?

tmshort:
@joelanford @dtfranz now only the sorting remains.

camilamacedo86 (approved):

internal/operator-controller/applier/phase.go
```diff
@@ -111,6 +113,57 @@ func init() {
 	}
 }
 
+// Sort objects within the phase deterministically by Group, Version, Kind, Namespace, Name
+// to ensure consistent ordering regardless of input order. This is critical for
+// Helm-to-Boxcutter migration where the same resources may come from different sources
+// (Helm release manifest vs bundle manifest) and need to produce identical phases.
+func compareClusterExtensionRevisionObjects(a, b ocv1.ClusterExtensionRevisionObject) int {
```
perdasilva, line 121: would using cmp.Compare be a nice alternative here? e.g.

`cmp.Compare(cmp.Or(aGVK.Group, bGVK.Group), cmp.Or(...), ...)`

perdasilva:
```
return cmp.Compare(
  cmp.Or(a.Object.GroupVersionKind().Group, b.Object.GroupVersionKind().Group),
  cmp.Or(a.Object.GroupVersionKind().Version, b.Object.GroupVersionKind().Version),
  cmp.Or(a.Object.GroupVersionKind().Kind, b.Object.GroupVersionKind().Kind),
  cmp.Or(a.Object.GetNamespace(), b.Object.GetNamespace()),
  cmp.Or(a.Object.GetName(), b.Object.GetName()),
)
```

perdasilva: only a suggestion - if it doesn't spark joy, I'm happy to approve. Let me know.

tmshort: I'm not familiar with that API, right now, it's not sparking joy (😉), but what's here works... so let's leave as-is.

perdasilva: np - it's the cool kid in town: https://pkg.go.dev/cmp basically obviates a lot the branching here

perdasilva (approved):

internal/operator-controller/applier/phase.go
```diff
@@ -111,6 +113,57 @@ func init() {
...
+
+	// Compare Group
+	if aGVK.Group < bGVK.Group {
+		return -1
+	} else if aGVK.Group > bGVK.Group {
+		return 1
+	}
+
+	// Compare Version
+	if aGVK.Version < bGVK.Version {
```
perdasilva: Add another couple of resources to the test that exercise the version comparison (or just change one of the current resources to use the apps/v2 api group or something) - I think that will clear the codecov problem

tmshort: If I do this, then I might as well look at `cmp`

tmshort: And it's still less than threshold... 😢

PR #2332, opened 2025-11-13
:seedling: Refactor `ClusterExtension` reconciler to use composable step-based pipeline

pedjak:
# Description

Replaces the monolithic 170-line reconcile() method with a flexible step-based architecture that executes discrete reconciliation phases in sequence. Each phase (`HandleFinalizers`, `RetrieveRevisionStates`, `RetrieveRevisionMetadata`, `UnpackBundle`, `ApplyBundle`) is now a standalone function that can be composed differently for Helm vs Boxcutter workflows.

The motivation for this PR came out the work in https://github.com/operator-framework/operator-controller/pull/2281 where we have realized that handling conditions with Boxcutter differently would require introducing conditional executions depending on the backend type in one part of the reconcile loop, although the rest of the reconciliation is the same. In order to keep things easy to understand and test, I have divided the reconciliation into steps so that some of them could be swapped out/replaced depending on the used backend. Moreover, we are opening a possibility to add and write simpler unit tests.

Changes:
  - Introduce `ReconcileStepFunc` type and `ReconcileSteps` executor
  - Extract reconcile logic into individual step functions in new file  `clusterextension_reconcile_steps.go`
  - Move `BoxcutterRevisionStatesGetter` to `boxcutter_reconcile_steps.go`  alongside `MigrateStorage` step
  - Configure step pipelines in `main.go` for each applier type
  - Refactor tests to use functional options pattern for reconciler setup

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/cmd/operator-controller/main.go
+++ b/cmd/operator-controller/main.go
@@ -50,7 +50,6 @@ import (
 	crcache "sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/certwatcher"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	crcontroller "sigs.k8s.io/controller-runtime/pkg/controller"
 	crfinalizer "sigs.k8s.io/controller-runtime/pkg/finalizer"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/log"
@@ -68,6 +67,7 @@ import (
 	"github.com/operator-framework/operator-controller/internal/operator-controller/catalogmetadata/cache"
 	catalogclient "github.com/operator-framework/operator-controller/internal/operator-controller/catalogmetadata/client"
 	"github.com/operator-framework/operator-controller/internal/operator-controller/contentmanager"
+	cmcache "github.com/operator-framework/operator-controller/internal/operator-controller/contentmanager/cache"
 	"github.com/operator-framework/operator-controller/internal/operator-controller/controllers"
 	"github.com/operator-framework/operator-controller/internal/operator-controller/features"
 	"github.com/operator-framework/operator-controller/internal/operator-controller/finalizers"
@@ -106,6 +106,31 @@ type config struct {
 	catalogdCasDir       string
 	pullCasDir           string
 	globalPullSecret     string
+}
+
+type reconcilerConfigurator interface {
+	Configure(cer *controllers.ClusterExtensionReconciler) error
+}
+
+type boxcutterReconcilerConfigurator struct {
+	mgr                   manager.Manager
+	preflights            []applier.Preflight
+	regv1ManifestProvider applier.ManifestProvider
+	resolver              resolve.Resolver
+	imageCache            imageutil.Cache
+	imagePuller           imageutil.Puller
+	finalizers            crfinalizer.Finalizers
+}
+
+type helmReconcilerConfigurator struct {
+	mgr                   manager.Manager
+	preflights            []applier.Preflight
+	regv1ManifestProvider applier.ManifestProvider
+	resolver              resolve.Resolver
+	imageCache            imageutil.Cache
+	imagePuller           imageutil.Puller
+	finalizers            crfinalizer.Finalizers
+	watcher               cmcache.Watcher
 }
 
 const (
@@ -440,11 +465,7 @@ func run() error {
 	}
 
 	ceReconciler := &controllers.ClusterExtensionReconciler{
-		Client:      cl,
-		Resolver:    resolver,
-		ImageCache:  imageCache,
-		ImagePuller: imagePuller,
-		Finalizers:  clusterExtensionFinalizers,
+		Client: cl,
 	}
 	ceController, err := ceReconciler.SetupWithManager(mgr, ctrlBuilderOpts...)
 	if err != nil {
@@ -459,13 +480,30 @@ func run() error {
 		IsWebhookSupportEnabled:     certProvider != nil,
 		IsSingleOwnNamespaceEnabled: features.OperatorControllerFeatureGate.Enabled(features.SingleOwnNamespaceInstallSupport),
 	}
-
+	var cerCfg reconcilerConfigurator
 	if features.OperatorControllerFeatureGate.Enabled(features.BoxcutterRuntime) {
-		err = setupBoxcutter(mgr, ceReconciler, preflights, clusterExtensionFinalizers, regv1ManifestProvider)
+		cerCfg = &boxcutterReconcilerConfigurator{
+			mgr:                   mgr,
+			preflights:            preflights,
+			regv1ManifestProvider: regv1ManifestProvider,
+			resolver:              resolver,
+			imageCache:            imageCache,
+			imagePuller:           imagePuller,
+			finalizers:            clusterExtensionFinalizers,
+		}
 	} else {
-		err = setupHelm(mgr, ceReconciler, preflights, ceController, clusterExtensionFinalizers, regv1ManifestProvider)
+		cerCfg = &helmReconcilerConfigurator{
+			mgr:                   mgr,
+			preflights:            preflights,
+			regv1ManifestProvider: regv1ManifestProvider,
+			resolver:              resolver,
+			imageCache:            imageCache,
+			imagePuller:           imagePuller,
+			finalizers:            clusterExtensionFinalizers,
+			watcher:               ceController,
+		}
 	}
-	if err != nil {
+	if err := cerCfg.Configure(ceReconciler); err != nil {
 		setupLog.Error(err, "unable to setup lifecycler")
 		return err
 	}
@@ -524,19 +562,13 @@ func getCertificateProvider() render.CertificateProvider {
 	return nil
 }
 
-func setupBoxcutter(
-	mgr manager.Manager,
-	ceReconciler *controllers.ClusterExtensionReconciler,
-	preflights []applier.Preflight,
-	clusterExtensionFinalizers crfinalizer.Registerer,
-	regv1ManifestProvider applier.ManifestProvider,
-) error {
-	coreClient, err := corev1client.NewForConfig(mgr.GetConfig())
+func (c *boxcutterReconcilerConfigurator) Configure(ceReconciler *controllers.ClusterExtensionReconciler) error {
+	coreClient, err := corev1client.NewForConfig(c.mgr.GetConfig())
 	if err != nil {
 		return fmt.Errorf("unable to create core client: %w", err)
 	}
-	cfgGetter, err := helmclient.NewActionConfigGetter(mgr.GetConfig(), mgr.GetRESTMapper(),
-		helmclient.StorageDriverMapper(action.ChunkedStorageDriverMapper(coreClient, mgr.GetAPIReader(), cfg.systemNamespace)),
+	cfgGetter, err := helmclient.NewActionConfigGetter(c.mgr.GetConfig(), c.mgr.GetRESTMapper(),
+		helmclient.StorageDriverMapper(action.ChunkedStorageDriverMapper(coreClient, c.mgr.GetAPIReader(), cfg.systemNamespace)),
 		helmclient.ClientNamespaceMapper(func(obj client.Object) (string, error) {
 			ext := obj.(*ocv1.ClusterExtension)
 			return ext.Spec.Namespace, nil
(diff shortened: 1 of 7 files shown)


internal/operator-controller/controllers/clusterextension_reconcile_steps.go
```diff
@@ -0,0 +1,211 @@
...
+			}
+			setInstalledStatusConditionUnknown(ext, err.Error())
+			setStatusProgressing(ext, errors.New("retrying to get installed bundle"))
+			return ctx, nil, err
+		}
+		return context.WithValue(ctx, revisionStatesKey{}, revisionStates), nil, nil
+	}
+}
+
+func RetrieveRevisionMetadata(r resolve.Resolver) ReconcileStepFunc {
```
perdasilva: Maybe we should name this in a way that makes it clear that this is the resolution step? E.g. 
```suggestion
func ResolveBundle(r resolve.Resolver) ReconcileStepFunc {
```

or something like that

pedjak:
> Although, I don't feel like a well structured monolith is necessarily a problem when the number of steps is relatively small, and the pipeline is pretty linear since that makes handling the code easier imo.

The motivation for this PR came out the work in https://github.com/operator-framework/operator-controller/pull/2281 where we have realized that handling conditions with Boxcutter differently would require introducing conditional executions depending on the backend type in one part of the reconcile loop, although the rest of the reconciliation is the same.  In order to keep things easy to understand and test, I have divided the reconciliation into steps so that some of them could be swapped out/replaced depending on the used backend. Moreover, we are opening a possibility to add write simpler unit tests.

> Another concern is passing values between the steps using context. I think there's a few drawbacks like bypassing compile type checks, and opaque/implicit dependencies. Would having a state object instead be a better alternative? Such an object might also provide a good data source for determining the underlying state and setting the conditions appropriately.

Golang context is used for sharing request-scoped cross-cutting concerns by many frameworks. The added `valueFromContext()` perform necessary checks at the runtime.

> E.g., what if we decompose the steps into methods to help with isolation testing, adapt the component interfaces to be usable in both helm and boxcutter world (I think we might already be there to a large extent?), and, pass state object between the steps and drive status setting from what falls out of that.

I am not sure what would be the difference to what is the proposed here? Can you provide some sketches?

cmd/operator-controller/main.go
```diff
@@ -700,6 +708,14 @@ func setupHelm(
 		Manager:                       cm,
 	}
 	ceReconciler.RevisionStatesGetter = &controllers.HelmRevisionStatesGetter{ActionClientGetter: acg}
+	ceReconciler.ReconcileSteps = []controllers.ReconcileStepFunc{
+		controllers.HandleFinalizers(ceReconciler.Finalizers),
+		controllers.RetrieveRevisionStates(ceReconciler.RevisionStatesGetter),
+		controllers.RetrieveRevisionMetadata(ceReconciler.Resolver),
+		controllers.UnpackBundle(ceReconciler.ImagePuller, ceReconciler.ImageCache),
+		controllers.ApplyBundle(ceReconciler.Applier),
```
joelanford: Do we need `Finalizers`, `RevisionStatesGetter`, `Resolver`, `ImagePuller`, etc. to still be fields on the `ceReconciler`? It seems like those details are no longer a direct concern of the reconciler?

perdasilva:
> I am not sure what would be the difference to what is the proposed here? Can you provide some sketches?

Maybe we could use a state/result type struct to collect the the results of each step and use that instead of the context to pass results between subsequent steps, e.g.:

```
type Result struct {
	RevisionStates            *RevisionStates
	BundleResolutionResult    *BundleResolutionResult // contains bundle resolution state: bundle/deprecation metadata, etc.
	BundleUnpackResult.       *BundleUnpackResult // contains bundle unpack state: fs, time unpacked, still unpacking, etc.
	ApplierResult             *ApplierResult // contains applier results
}

type ReconcileStepFunc func(context.Context, *ocv1.ClusterExtension, *Result) (context.Context, *ctrl.Result, error)
```

Errors could either be embedded in the step results or just be raised by the step function itself through its return values like we have it now.

If the only point of contention is condition setting, we could strip the steps of setting any conditions and just have one step at the end that does it based on the results. Then have one impl for helm and one for boxcutter and just configure the reconciler accordingly. 

If the folks feel a framework for a linear 5-6 step pipeline is overkill (I don't have super strong opinions here - I can live with either). We could still get the testing benefits by decomposing the reconcile function similarly to what you have done here but just call them in turn to eliminate the indirection. 

> Golang context is used for sharing request-scoped cross-cutting concerns by many frameworks. The added valueFromContext() perform necessary checks at the runtime.

I get that. I'm just wondering if we're better off using a state object to carry results between steps. The impression I got from searching around and talking to Gemini is that context values are typically used for more lightweight information such as request metadata (ids, users, headers, etc.) rather than intermediate computation results. 

I think we should call out the motivation in the description. As it is now, we've refactored things to make it explicit that storage migration isn't part of the helm reconciler.

I hope all of this is cogent. I'm pretty beat. But we can take it up again tomorrow ^^

cmd/operator-controller/main.go
pedjak: > Do we need `Finalizers`, `RevisionStatesGetter`, `Resolver`, `ImagePuller`, etc. to still be fields on the `ceReconciler`? It seems like those details are no longer a direct concern of the reconciler?

Correct, I would remove them if we agree about PR direction.

pedjak:
> I get that. I'm just wondering if we're better off using a state object to carry results between steps. The impression I got from searching around and talking to Gemini is that context values are typically used for more lightweight information such as request metadata (ids, users, headers, etc.) rather than intermediate computation results.

Having a dedicated state instead of the golang context:
* would not save us from runtime errors because any of fields in the state could be `nil`, so we still need to check for that
* for each reconciler we would need to develop a dedicated little step framework that would pass the proper state (yes, we could apply the same approach to other controllers in this repo).

>  If the folks feel a framework for a linear 5-6 step pipeline is overkill (I don't have super strong opinions here - I can live with either).

I would not consider as a framework something that consists of 15 lines of code:

```golang
func (steps *ReconcileSteps) Reconcile(ctx context.Context, ext *ocv1.ClusterExtension) (ctrl.Result, error) {
	var res *ctrl.Result
	var err error
	nextStepCtx := ctx
	for _, step := range *steps {
		nextStepCtx, res, err = step(nextStepCtx, ext)
		if err != nil {
			return ctrl.Result{}, err
		}
		if res != nil {
			return *res, nil
		}
		if nextStepCtx == nil {
			return ctrl.Result{}, nil
		}
	}
	return ctrl.Result{}, nil
}
```

>  We could still get the testing benefits by decomposing the reconcile function similarly to what you have done here but just call them in turn to eliminate the indirection.

Not really, because the wiring is fixed - by the introduced indirection, we are now able to add for example migration to boxcutter, and apply bundles and reports conditions  in a different way in cases of Helm and Boxcutter backends.

> I think we should call out the motivation in the description. As it is now, we've refactored things to make it explicit that storage migration isn't part of the helm reconciler.

Done.

joelanford:
To help overcome the concern about the context value type safety, perhaps we could encapsulate the context value handling into a sub-internal package with 100% test coverage, where only typed accessor/getter functions are exported and available to our reconcile functions?

The only downside at that point would be that we'd need `nil` checks for those values just not being present in the context, and we'd have to handle those errors (or panic).

joelanford:
In general, I see this design as an improvement over what we currently have because it tidies things up a bit, although I also think @perdasilva's comments have merit.

Decomposing and separating out the set of functions and their order makes things easier to physically see. But I think it does potentially introduce surface area for unintended divergence between helm and boxcutter setups (but maybe our test suite has us covered there?).

And I do agree with @perdasilva that context values are worse, in isolation, than our current monolithic structure where the right values are always available in the right places. We _could_ define a single interface that has methods for each of the steps with the right signatures for data exchange. That would be brittle, but maybe the brittleness would be a benefit because it would force us to confront and account for the changes needed in one reconcile flavor vs another?

camilamacedo86:
Thank you, @pedjak, for this thoughtful refactoring! I really appreciate the effort to improve structure and testability.

After reviewing the changes, I’m not fully sure this pattern is the best fit for our situation. I might be wrong, but since **we may end up with a single runtime long-term (likely only Boxcutter)**, the extra flexibility could turn into unnecessary complexity. In that case, the benefits of the step framework might be short-lived, while the cost of maintaining and later unwinding it remains.

The current Helm/Boxcutter reconcile is already quite simple and explicit. This PR adds several new types, files, and ~100 lines of framework logic. That feels like more machinery than we need right now. If/when Helm gets removed, the existing code path is very easy to clean up. With the framework pattern, we’d have more scaffolding to unwind, even though the composability is no longer needed.

I also have some technical concerns around how `context` is used. Using `context.WithValue` to pass internal state between steps is less type-safe and makes dependencies harder to see in controller code. The Go team recommends using context values only for request-scoped data that crosses API boundaries (see the [official Go blog](https://go.dev/blog/context)), which is different from the use here.

### A Simpler Alternative  ( Common Approach for Reconciles )

If the main goal is improving testability and separation of concerns, we could break the reconciliation flow into smaller helper methods with explicit parameters. This keeps things type-safe, idiomatic, and easier to follow, while still giving us good structure:

```go
func (r *ClusterExtensionReconciler) reconcile(ctx context.Context, ext *ocv1.ClusterExtension) (ctrl.Result, error) {
    if res, err := r.handleFinalizers(ctx, ext); res != nil || err != nil {
        return *res, err
    }

    if res, err := r.migrateStorageIfNeeded(ctx, ext); res != nil || err != nil {
        return *res, err
    }

    states, err := r.getRevisionStates(ctx, ext)
    if err != nil {
        return r.handleRevisionStatesError(ctx, ext, err)
    }

    resolved, err := r.resolveBundle(ctx, ext, states)
    if err != nil {
        return r.handleResolutionError(ctx, ext, states, err)
    }

    imageFS, err := r.unpackBundle(ctx, ext, resolved)
    if err != nil {
        return r.handleUnpackError(ctx, ext, states, resolved, err)
    }

    return r.applyBundle(ctx, ext, states, resolved, imageFS)
}

**Comment by camilamacedo86** [comment]
Hi @pedjak,  

Just to supplement the discussion — and please take this as a gentle suggestion:

Another important point I'd like to raise is that we should be using controller-runtime's EnvTest (https://book.kubebuilder.io/reference/envtest) to validate our reconcilers, rather than relying on unit tests. Without EnvTest, we're not really exercising the reconciler in a realistic control-plane environment, so these changes won't fully address the underlying issue. Unit tests also require mocks that EnvTest makes unnecessary.

**If we follow the recommended EnvTest approach, would we really need the testing benefits this PR is trying to introduce?** I'm honestly not sure we would, because unit-testing reconcilers isn't considered a good practice — EnvTest already provides the proper, realistic way to test them without the need for mocks or additional framework layers.

I would gently suggest we **hold this change for now**, until we have:

- Boxcutter in a more mature state,  
- clarity on how much the reconciliation flows will actually diverge (or not),  
- clarity on whether we will support more than one runtime,  
- a better understanding of the next steps for OLM v1.  

Once we have that information, it will be much easier to decide whether this framework approach truly fits our future needs.

**Comment by pedjak** [comment]
> And I do agree with @perdasilva that context values are worse, in isolation, than our current monolithic structure where the right values are always available in the right places. We _could_ define a single interface that has methods for each of the steps with the right signatures for data exchange. That would be brittle, but maybe the brittleness would be a benefit because it would force us to confront and account for the changes needed in one reconcile flavor vs another?

@joelanford @perdasilva I have added a dedicated reconcile state holder for exchanging data between the steps, ptal. IMO we do not need to perform `nil` checks, given that the potential panics are going to bubble up as the reconciliation error, given that controller-runtime `RecoverPanic` config param is enabled by the default.

**Comment by pedjak** [comment]
> The current Helm/Boxcutter reconcile is already quite simple and explicit. This PR adds several new types, files, and ~100 lines of framework logic.

@camilamacedo86 

Not really, it adds just a step function definition and ~15 lines of code that iterate over the provided steps. The rest of PR is just splitting the long existing reconcile function into a number of logically separated steps, boundaries were already visible by dedicated log messages like:
* `"handling finalizers"`
* `"getting installed bundle"`
* `"resolving bundle"`
* etc.

I mentioned already the motivation for this move in PR description - we have currently two backends and some apply and condition logic is going to be different. Instead of trying to add more conditional logic, splitting it into a chain if steps that could replaced/extended is good way to combat this problem and obtain additional benefits long-term.

> The Go team recommends using context values only for request-scoped data that crosses API boundaries (see the [official Go blog](https://go.dev/blog/context)), which is different from the use here.

It depends how you look at: an argument for the reconcile function is `ctrl.Request`, a sort of request that needs to be processed. Hence, carrying request-scoped data in the context might be perfectly fine.

**Comment by perdasilva** [comment]
I think the addition of the state object is a good compromise here and resolves the runtime type check issue, which was my biggest concern. If we can get agreement amongst the reviewers here, I'm happy to move forward with this.

File: cmd/operator-controller/main.go (hunk 2 above)
Comment (pedjak) [reply cmd/operator-controller/main.go]: Removed, ptal.

File: cmd/operator-controller/main.go [hunk 3]
```diff
@@ -108,6 +108,31 @@ type config struct {
 	globalPullSecret     string
 }
 
+type clusterExtensionReconcilerConfigurator interface {
+	Configure(cer *controllers.ClusterExtensionReconciler) error
+}
+
+type boxcutterCERConfigurator struct {
```
joelanford: Naming nit: my first thought here was that CER stands for `ClusterExtensionRevision`, but it is actually ClusterExtension reconciler? Can we disambiguate? Same for helmCERConfigurator.

joelanford:
This is shaping up nicely IMO! Thanks for this.

/approve

I'll leave lgtm to another maintainer to make sure there is consensus from at least one other person working regularly in this codebase.

PR #2333, opened 2025-11-13
✨ OPRUN-4221: Use operator-controller SA by default, make SA field optional

trgeiger:
Changes the ClusterExtension API field spec.ServiceAccount to be optional. Operator-controller will use its own service account by default unless the spec.ServiceAccount field is set. RBAC PreAuthorization only happens if the optional SA field is set, as well.

Give operator-controller's SA cluster-admin by default.

Addresses [OPRUN-4221](https://issues.redhat.com/browse/OPRUN-4221)

Wasn't sure if I should mark this major or minor change.

# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/api/v1/clusterextension_types.go
+++ b/api/v1/clusterextension_types.go
@@ -67,14 +67,14 @@ type ClusterExtensionSpec struct {
 	// +kubebuilder:validation:Required
 	Namespace string `json:"namespace"`
 
-	// serviceAccount is a reference to a ServiceAccount used to perform all interactions
-	// with the cluster that are required to manage the extension.
+	// serviceAccount is an optional field that references a ServiceAccount used to
+	// perform all interactions with the cluster that are required to manage the extension.
+	// If not set, operator-controller will use its own ServiceAccount for extension management.
 	// The ServiceAccount must be configured with the necessary permissions to perform these interactions.
 	// The ServiceAccount must exist in the namespace referenced in the spec.
-	// serviceAccount is required.
 	//
-	// +kubebuilder:validation:Required
-	ServiceAccount ServiceAccountReference `json:"serviceAccount"`
+	// +optional
+	ServiceAccount ServiceAccountReference `json:"serviceAccount,omitzero"`
 
 	// source is a required field which selects the installation source of content
 	// for this ClusterExtension. Selection is performed by setting the sourceType.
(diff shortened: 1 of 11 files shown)


trgeiger:
Due to the changes to the clusterrole, I don't think this is ever going to pass upgrade-e2e since it can't change the clusterrole in place. What's the cleanest way of handling this, versioning our clusterrole moving forward? Just make a new one like the existing boxcutter experimental config does?

api/v1/clusterextension_types.go
```diff
@@ -403,7 +403,7 @@ type ServiceAccountReference struct {
 	// +kubebuilder:validation:MaxLength:=253
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="name is immutable"
```
joelanford, line 404: Should we allow a CE to unset the SA now that it is optional? Open question. I think we need to discuss.

trgeiger:
That's a great question. Does it make sense to hash that out in the RFC?

joelanford:
We need to add an e2e that checks that we can correctly install a ClusterExtension that does not have a service account configured.

The JSON serialization also seems a bit finicky, so I would also suggest adding a unit test that serializes a ClusterExtension lacking a service account into JSON and checks to ensure that there is no `spec.serviceAccount` field in the output at all.

api/v1/clusterextension_types.go
```diff
@@ -403,7 +403,7 @@ type ServiceAccountReference struct {
...
 	// +kubebuilder:validation:XValidation:rule="self.matches(\"^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$\")",message="name must be a valid DNS1123 subdomain. It must contain only lowercase alphanumeric characters, hyphens (-) or periods (.), start and end with an alphanumeric character, and be no longer than 253 characters"
-	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Optional
```
joelanford, line 406:
```suggestion
	// +optional
```

api/v1/clusterextension_types.go
```diff
@@ -403,7 +403,7 @@ type ServiceAccountReference struct {
...
 	Name string `json:"name"`
```
joelanford, line 407:
```suggestion
	Name string `json:"name,omitempty"`
```

PR #2334, opened 2025-11-14
:seedling: Add enum validation annotation to CollisionProtection

perdasilva:
# Description

Adds missing enum validation on CollisionProtection `+kubebuilder:validation:Enum=Prevent;IfNoController;None`

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/api/v1/clusterextensionrevision_types.go
+++ b/api/v1/clusterextensionrevision_types.go
@@ -108,6 +108,7 @@ type ClusterExtensionRevisionObject struct {
 	// already existing on the cluster or even owned by another controller.
 	//
 	// +kubebuilder:default="Prevent"
+	// +kubebuilder:validation:Enum=Prevent;IfNoController;None
 	// +optional
 	CollisionProtection CollisionProtection `json:"collisionProtection,omitempty"`
 }
--- a/helm/olmv1/base/operator-controller/crd/experimental/olm.operatorframework.io_clusterextensionrevisions.yaml
+++ b/helm/olmv1/base/operator-controller/crd/experimental/olm.operatorframework.io_clusterextensionrevisions.yaml
@@ -87,6 +87,10 @@ spec:
                             description: |-
                               CollisionProtection controls whether OLM can adopt and modify objects
                               already existing on the cluster or even owned by another controller.
+                            enum:
+                            - Prevent
+                            - IfNoController
+                            - None
                             type: string
                           object:
                             type: object
--- a/manifests/experimental-e2e.yaml
+++ b/manifests/experimental-e2e.yaml
@@ -678,6 +678,10 @@ spec:
                             description: |-
                               CollisionProtection controls whether OLM can adopt and modify objects
                               already existing on the cluster or even owned by another controller.
+                            enum:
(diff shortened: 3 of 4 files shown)


api/v1/clusterextensionrevision_types.go
```diff
@@ -55,9 +55,9 @@ type ClusterExtensionRevisionSpec struct {
 	// a unique value assigned. It is immutable after creation. The new revision number must always be previous revision +1.
 	//
 	// +kubebuilder:validation:Required
-	// +kubebuilder:validation:Minimum:=1
+	// +kubebuilder:validation:Minimum=1
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf", message="revision is immutable"
-	Revision int64 `json:"revision"`
+	Revision int32 `json:"revision"`
```
pedjak: what we are improving/fixing with changing the type?

api/v1/clusterextensionrevision_types.go
```diff
@@ -55,9 +55,9 @@ type ClusterExtensionRevisionSpec struct {
 	// a unique value assigned. It is immutable after creation. The new revision number must always be previous revision +1.
 	//
 	// +kubebuilder:validation:Required
-	// +kubebuilder:validation:Minimum:=1
+	// +kubebuilder:validation:Minimum=1
```
pedjak: The version with `:=` is documented: https://book.kubebuilder.io/reference/markers/crd-validation

You can also see that this is a no-op change - check the generated CRD.

api/v1/clusterextensionrevision_types.go
rashmigottipati: It also looks like this change is causing type mismatch errors as boxcutter still expects `int64` in some places.

pedjak: In that case we should revert it back to `int64`.

tmshort: Isn't this technically an API change (due to the size change)?

rashmigottipati: And possibly a breaking change too from an API perspective.

api/v1/clusterextensionrevision_types.go
perdasilva: Ah, I didn't know that. Is there value in using a single format in the CRDs we produce?

api/v1/clusterextensionrevision_types.go
perdasilva: It seems to be best practice to use int32 if the extra precision isn't needed as it avoid issues with JSON serialization on some clients.

I don't have a big dog in this hunt since I don't think we'll experience the json serialization issues (since we'll never reach a number large enough for it to show up). I guess we could also cap the value at the JSON max int to avoid this? I'd also like to understand whether we think we need the extra precision and what do we lose by not having it...

It would be a breaking change. But, is my understanding correct that we are allowed to have breaking API changes while in experimental mode? Or is the expectation that we change the version? e.g. v1alpha2?

rashmigottipati (approved):
/lgtm

PR #2339, opened 2025-11-17
🐛 override local pull policy with permissive policy

grokspawn:
# Description
The image pull test requires that the user running it has a permissive image pull policy in place to be successful.  However, I run a restrictive policy on my machine, so I get test failures. 
Ideally, the unit test should not be dependent on the user's environment, so here's a change which specifies a test-specific, permissive pull policy which will be used instead of the user's environment. 

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/shared/util/image/pull_test.go
+++ b/internal/shared/util/image/pull_test.go
@@ -298,8 +298,15 @@ func buildSourceContextFunc(t *testing.T, ref reference.Named) func(context.Cont
 		require.NoError(t, enc.Encode(registriesConf))
 		require.NoError(t, f.Close())
 
+		// Create an insecure policy for testing to override any system-level policy
+		// that might reject unsigned images
+		policyPath := filepath.Join(configDir, "policy.json")
+		insecurePolicy := `{"default":[{"type":"insecureAcceptAnything"}]}`
+		require.NoError(t, os.WriteFile(policyPath, []byte(insecurePolicy), 0600))
+
 		return &types.SystemContext{
 			SystemRegistriesConfPath: registriesConfPath,
+			SignaturePolicyPath:      policyPath,
 		}, nil
 	}
 }


internal/shared/util/image/pull_test.go
```diff
@@ -298,8 +298,15 @@ func buildSourceContextFunc(t *testing.T, ref reference.Named) func(context.Cont
...
 
+		// Create an insecure policy for testing to override any system-level policy
+		// that might reject unsigned images
+		policyPath := filepath.Join(configDir, "policy.json")
+		insecurePolicy := `{"default":[{"type":"insecureAcceptAnything"}]}`
+		require.NoError(t, os.WriteFile(policyPath, []byte(insecurePolicy), 0644))
+
 		return &types.SystemContext{
 			SystemRegistriesConfPath: registriesConfPath,
+			SignaturePolicyPath:      policyPath,
```
camilamacedo86, line 309: I think we need to here like we here: https://github.com/openshift/operator-framework-operator-controller/blob/main/openshift/default-catalog-consistency/pkg/extract/extract.go#L231-L251

grokspawn: Thanks for the review @camilamacedo86!
What's the concern here?  Is it that we're dropping a file in t.TempDir() for use in this test scenario? 
That's the only difference I see between the approaches.

grokspawn: While I would love to avoid the temp file generation, this is just a different API than the app uses, and an attempt to avoid writing the overriding signature policy file results in uglier shenanigans.

camilamacedo86: The delta is that we are trying to solve it with the signature if not be possible then we let it go instead of fixed se `	"default": [{"type": "insecureAcceptAnything"}]`.

grokspawn: I'm not following you.  
Right now this test fails on users' machines with a restrictive signing policy, but the test should not be sensitive to users' environments. 
[Here](https://github.com/operator-framework/operator-controller/pull/2339/files#diff-10967a567bc639d23ffbfaed1334bfe5627ffd23ad3d980b94a289a0cfd493cdR304) creates an insecure signing policy for the scoped tests to be used instead of detecting the users' policies.

camilamacedo86: I think it is fine 

/lgtm

PR #2340, opened 2025-11-18
:warning: Refine ClusterExtentionRevision conditions

perdasilva:
# Description

## Summary

  This PR refactors the ClusterExtensionRevision status conditions to provide clearer, more actionable feedback about revision lifecycle states. The changes introduce a new Progressing condition, simplify condition reasons, and ensure that revision-level errors (retrying states) are properly surfaced to the parent ClusterExtension.

This PR does introduce API breaking changes. But, only to experimental APIs. It should be ok to override.

##  Key Changes

  1. Refactored ClusterExtensionRevision Status Conditions

  Simplified condition types:
  - Retained Available and Succeeded conditions
  - Added new Progressing condition for better visibility into rollout state
  - Removed obsolete condition reasons and consolidated error handling

  Updated condition reasons to be more semantic and actionable:
  - Before: ReconcileFailure, RevisionValidationFailure, PhaseValidationError, ObjectCollisions, RolloutSuccess, Incomplete, Progressing
  - After: Retrying, Blocked, RollingOut, Archived, Migrated

  API Changes:
  - Added Progressing column to kubectl get clusterextensionrevisions output

  2. Surfaced Revision Retrying Condition to ClusterExtension

  Modified the Boxcutter applier to propagate retrying errors from ClusterExtensionRevision to the parent ClusterExtension. When a revision is in a Progressing=True state with Reason=Retrying, the error is now surfaced to the ClusterExtension, providing better visibility into failed reconciliation attempts. When the revision is Progressing=True/Succeeded the applier returns success.

  3. Added Comprehensive E2E Tests

  Added end-to-end tests for ClusterExtensionRevision lifecycle scenarios to validate the new condition states and transitions.

 4. Added ClusterExtension documentation to the RolloutInProgress `Progression` reason 

### Progressing Condition (experimental only)

  | Status  | Reason          | Description                                                         |
  |---------|-----------------|---------------------------------------------------------------------|
  | True    | RollingOut | The latest active revision is rolling out                           |

###  Progressing Condition

 Indicates whether the revision is progressing to its next state. It follows the same semantic as the ClusterExtension and Deployment `Progressing` condition.

  | Status | Reason     | Description                                                                                                |
  |--------|------------|------------------------------------------------------------------------------------------------------------|
  | True   | Retrying   | Reconciliation failed due to validation errors, object collisions, or other recoverable errors; will retry |
  | True   | RollingOut | Revision is actively rolling out objects across phases                                                     |
  | True  | Succeeded  | Revision has completed its rollout successfully                                                            |
  | False  | Blocked  | Revision reconciliation failed with unrecoverable error - manual intervention might be required                                                           |
  | False  | Archived   | Revision has been archived and is no longer progressing                                                    |

Not a part of this PR yet, but `Progressing` can also be `Blocked` for non-retryable errors.

###  Available Condition

  Indicates whether the revision's objects are available and passing the object status probes.

  | Status  | Reason          | Description                                                         |
  |---------|-----------------|---------------------------------------------------------------------|
  | True    | ProbesSuceeded | Objects are available and pass all probes                           |
  | False   |ProbeFailed | One or more objects are failing their availability probes           |
  | Unknown | Reconciling  | Probes could not be observed / intermittent reconciliation error |
  | Unknown | Archived  | Revision is archived |
  | Unknown | Migrated  | Revision was migrated from Helm release |

###  Succeeded Condition

  Indicates whether the revision has successfully completed its rollout.

  | Status | Reason    | Description                                                      |
  |--------|-----------|------------------------------------------------------------------|
  | True   |Succeeded| Revision succeeded rolling out all objects and passed all probes |

  Migration Notes

  - The Progressing condition replaces the previous pattern of setting Available=False with various failure reasons
  - All retrying scenarios (validation failures, collisions, reconcile errors) now consistently set Progressing=True, Reason=Retrying and Available=Unknown
  - The new condition structure provides clearer separation between "something is wrong" (Progressing=Retrying) vs "rollout is in progress" (Progressing=RollingOut)

## Reviewer Checklist

- [ ] API Go Documentation


api/v1/clusterextensionrevision_types.go
```diff
@@ -25,21 +25,22 @@ const (
...
-	ClusterExtensionRevisionReasonRolloutSuccess            = "RolloutSuccess"
-	ClusterExtensionRevisionReasonProbeFailure              = "ProbeFailure"
-	ClusterExtensionRevisionReasonIncomplete                = "Incomplete"
-	ClusterExtensionRevisionReasonProgressing               = "Progressing"
-	ClusterExtensionRevisionReasonArchived                  = "Archived"
-	ClusterExtensionRevisionReasonMigrated                  = "Migrated"
+	ClusterExtensionRevisionReasonRetrying   = "Retrying"
+	ClusterExtensionRevisionReasonBlocked    = "Blocked"
+	ClusterExtensionRevisionReasonRollingOut = "RollingOut"
+	ClusterExtensionRevisionReasonRolledOut  = "RolledOut"
```
pedjak: given the discussion we had yesterday, I think we should call it `Succeeded`.

api/v1/clusterextensionrevision_types.go
```diff
@@ -25,21 +25,22 @@ const (
...
+
+	ClusterExtensionRevisionReasonProbeFailure    = "ProbeFailure"
+	ClusterExtensionRevisionReasonProbesSucceeded = "ProbesSucceeded"
+	ClusterExtensionRevisionReasonReconciling     = "Reconciling"
```
pedjak, line 38: Also, given the discussion yesterday, perhaps we do not need them at all?

api/v1/clusterextensionrevision_types.go
```diff
@@ -104,7 +105,6 @@ type ClusterExtensionRevisionObject struct {
 	// already existing on the cluster or even owned by another controller.
 	//
 	// +kubebuilder:default="Prevent"
-	// +kubebuilder:validation:Enum=Prevent;IfNoController;None
```
pedjak, line 165: why is the validation not needed anymore?

internal/operator-controller/controllers/clusterextensionrevision_controller.go
```diff
@@ -605,3 +526,64 @@ var (
 		FieldB: ".status.replicas",
 	}
 )
+
+func setRetryingConditions(cer *ocv1.ClusterExtensionRevision, message string) {
+	markAsProgressing(cer, ocv1.ClusterExtensionRevisionReasonRetrying, message)
+	markAsAvailableUnknown(cer, ocv1.ClusterExtensionRevisionReasonReconciling, message)
```
pedjak: I see no added value if we publish `Available(status=Unknown)` before we have rolled out the resources successfully. You can see that by duplicating the message on two places - such redundancy does not bring anything to users. I would suggest that we keep the logic introduced in https://github.com/operator-framework/operator-controller/pull/2281

internal/operator-controller/controllers/clusterextensionrevision_controller.go
```diff
@@ -117,60 +118,37 @@ func (c *ClusterExtensionRevisionReconciler) reconcile(ctx context.Context, rev
...
 	//
 	if err := c.ensureFinalizer(ctx, rev, clusterExtensionRevisionTeardownFinalizer); err != nil {
-		meta.SetStatusCondition(&rev.Status.Conditions, metav1.Condition{
-			Type:               ocv1.ClusterExtensionRevisionTypeAvailable,
-			Status:             metav1.ConditionFalse,
-			Reason:             ocv1.ClusterExtensionRevisionReasonReconcileFailure,
-			Message:            err.Error(),
-			ObservedGeneration: rev.Generation,
-		})
+		l.Error(err, "failed to ensure finalizer")
```
pedjak: no need for that, because this error will be logged by controller-runtime anyway after returning from the reconcile loop.

internal/operator-controller/controllers/clusterextensionrevision_controller.go
```diff
@@ -239,41 +202,42 @@ func (c *ClusterExtensionRevisionReconciler) reconcile(ctx context.Context, rev
 				// TODO: It feels like an error here needs to propagate to a status _somewhere_.
 				//   Not sure the current CER makes sense? But it also feels off to set the CE
 				//   status from outside the CE reconciler.
+				l.Error(err, "failed to patch previous revision", "revision", a.GetName())
+				setRetryingConditions(rev, fmt.Sprintf("error archiving previous revision: %s", err))
```
pedjak: we should not change `Progressing` condition on the actual revision if we are not able to patch previous revisions. All conditions on a revision describe just the state of that revision.

internal/operator-controller/controllers/clusterextensionrevision_controller.go
```diff
@@ -239,41 +202,42 @@ func (c *ClusterExtensionRevisionReconciler) reconcile(ctx context.Context, rev
 				// TODO: It feels like an error here needs to propagate to a status _somewhere_.
 				//   Not sure the current CER makes sense? But it also feels off to set the CE
 				//   status from outside the CE reconciler.
+				l.Error(err, "failed to patch previous revision", "revision", a.GetName())
```
pedjak: no need to log error here - it is going to be logged by controller-runtime after finishing the reconcile loop.

internal/operator-controller/controllers/clusterextensionrevision_controller.go
```diff
@@ -321,18 +263,11 @@ func (c *ClusterExtensionRevisionReconciler) teardown(ctx context.Context, rev *
...
 			l.Error(err, "revision teardown failed")
 		}
-		meta.SetStatusCondition(&rev.Status.Conditions, metav1.Condition{
-			Type:               ocv1.ClusterExtensionRevisionTypeAvailable,
-			Status:             metav1.ConditionFalse,
-			Reason:             ocv1.ClusterExtensionRevisionReasonReconcileFailure,
-			Message:            err.Error(),
-			ObservedGeneration: rev.Generation,
-		})
+		setRetryingConditions(rev, err.Error())
```
pedjak: we should not touch `Progressing` here because we are not rolling out anything. Let's keep `markAsUnavailable(...)` or even set it unknown.

internal/operator-controller/controllers/clusterextensionrevision_controller.go
```diff
@@ -346,26 +281,12 @@ func (c *ClusterExtensionRevisionReconciler) teardown(ctx context.Context, rev *
...
 
 	if err := c.TrackingCache.Free(ctx, rev); err != nil {
-		meta.SetStatusCondition(&rev.Status.Conditions, metav1.Condition{
-			Type:               ocv1.ClusterExtensionRevisionTypeAvailable,
-			Status:             metav1.ConditionFalse,
-			Reason:             ocv1.ClusterExtensionRevisionReasonReconcileFailure,
-			Message:            err.Error(),
-			ObservedGeneration: rev.Generation,
-		})
+		setRetryingConditions(rev, err.Error())
```
pedjak: same as above.

internal/operator-controller/controllers/clusterextensionrevision_controller.go
```diff
@@ -465,7 +386,7 @@ func (c *ClusterExtensionRevisionReconciler) listPreviousRevisions(ctx context.C
 	}
 
 	revList := &ocv1.ClusterExtensionRevisionList{}
-	if err := c.TrackingCache.List(ctx, revList, client.MatchingLabels{
+	if err := c.Client.List(ctx, revList, client.MatchingLabels{
```
pedjak: is this change required/in focus of this PR?

api/v1/clusterextensionrevision_types.go
perdasilva: merge conflict error - thanks for calling it out!

api/v1/clusterextensionrevision_types.go
perdasilva: Fixed that one up

api/v1/clusterextensionrevision_types.go
perdasilva: Need to think a little more about this one. It's possible.

internal/operator-controller/controllers/clusterextensionrevision_controller.go
perdasilva: But we're progressing to the next state for this revision, right?

internal/operator-controller/controllers/clusterextensionrevision_controller.go
perdasilva: There are stages in the reconciliation that can fail and put us in a retry state even before we get to probe the cluster to see if we're rolling something out or not. I think the reason I put these together (at least for now) is you can be carrying a stale availability. E.g.

reconcile 1: everything is good rolled out and available we post up Available=True Progressing=True/Succeeded
reconcile 2: something goes wrong before we look at the probes or know whether we are rolling out or have rolled out. Then we post up Available=True Progressing=True/Retrying/some error

internal/operator-controller/controllers/clusterextensionrevision_controller.go
perdasilva: We'd see something like:
Progressing=True/Succeeded
Available=False/Reconciling/_some tear down error_

vs

Progressing=True/Retrying/_some tear down error_
Available=True/Reconciling/_same some tear down error_

I don't love either :/

api/v1/clusterextensionrevision_types.go
perdasilva: I.e. just use True/False?

internal/operator-controller/controllers/clusterextensionrevision_controller.go
perdasilva: The down side here is that there can be flapping. I think for a first pass that could be ok. But we should definitely try to glob flapping and post up errors when we're sure it's a persistent problem and not some transient thing that's fixed in the next reconcile.

api/v1/clusterextensionrevision_types.go
perdasilva: I think that makes sense =D

perdasilva: I've removed them

internal/operator-controller/controllers/clusterextensionrevision_controller.go
```diff
@@ -605,3 +532,61 @@ var (
...
+		Message:            message,
+		ObservedGeneration: cer.Generation,
+	})
+}
+
+func setArchivedConditions(cer *ocv1.ClusterExtensionRevision) bool {
+	const msg = "revision is archived"
+	updated := markAsNotProgressing(cer, ocv1.ClusterExtensionRevisionReasonArchived, msg)
+	return markAsAvailableUnknown(cer, msg) || updated
+}
```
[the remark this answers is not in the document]
perdasilva: Due to linting complaining about return values not being used...

api/v1/clusterextensionrevision_types.go
perdasilva: That was bad. What do you think we should roll these into?

internal/operator-controller/controllers/clusterextensionrevision_controller.go
pedjak: > reconcile 2: something goes wrong before we look at the probes or know whether we are rolling out or have rolled out. Then we post up Available=True Progressing=True/Retrying/some error

Let's try to extract what we have in https://github.com/operator-framework/operator-controller/pull/2281 without adding more - fine tuning can happen in follow-ups.

In case of the quoted scenario: if we have `Progressing(status=True,reason=Retrying,message=some error)` we cannot have `Available(status=True)` and also there is no benefit of posting `Available(status=Unknown)`

BTW, it would be super useful for the review, if you could create a state diagram for `Progressing` and `Available` condition and add it to PR description.

internal/operator-controller/controllers/clusterextensionrevision_controller.go
```diff
@@ -605,3 +532,61 @@ var (
 		FieldB: ".status.replicas",
 	}
 )
+
+func setRetryingConditions(cer *ocv1.ClusterExtensionRevision, message string) {
+	markAsProgressing(cer, ocv1.ClusterExtensionRevisionReasonRetrying, message)
+	markAsAvailableUnknown(cer, message)
```
pedjak: if `Available` is not already present on status we should not introduce it with unknown status.

api/v1/clusterextensionrevision_types.go
pedjak: Ah sorry, my bad :facepalm:  - they are used with `Available` condition.

api/v1/clusterextensionrevision_types.go
```diff
@@ -25,21 +25,16 @@ const (
...
-	ClusterExtensionRevisionReasonProbeFailure              = "ProbeFailure"
-	ClusterExtensionRevisionReasonIncomplete                = "Incomplete"
-	ClusterExtensionRevisionReasonProgressing               = "Progressing"
-	ClusterExtensionRevisionReasonArchived                  = "Archived"
-	ClusterExtensionRevisionReasonMigrated                  = "Migrated"
+	ClusterExtensionRevisionReasonRetrying   = "Retrying"
+	ClusterExtensionRevisionReasonBlocked    = "Blocked"
+	ClusterExtensionRevisionReasonRollingOut = "RollingOut"
+
+	ClusterExtensionRevisionReasonArchived = "Archived"
```
tmshort: Does it make sense to sort these alphabetically

test/e2e/e2e_suite_test.go
```diff
@@ -35,6 +37,9 @@ func TestMain(m *testing.M) {
 	c, err = client.New(cfg, client.Options{Scheme: scheme.Scheme})
 	utilruntime.Must(err)
 
+	cs, err = kubernetes.NewForConfig(cfg)
```
tmshort, line 40: Is `cs` used anywhere?

perdasilva: https://github.com/operator-framework/operator-controller/pull/2340/files#diff-e3fea391931353952582f955139f0665c6c03b6ea92395fa283d7c22676371a1R200

internal/operator-controller/controllers/clusterextensionrevision_controller.go
```diff
@@ -214,18 +172,17 @@ func (c *ClusterExtensionRevisionReconciler) reconcile(ctx context.Context, rev
...
-				Reason:             ocv1.ClusterExtensionRevisionReasonObjectCollisions,
-				Message:            fmt.Sprintf("revision object collisions in phase %d\n%s", i, strings.Join(collidingObjs, "\n\n")),
-				ObservedGeneration: rev.Generation,
-			})
+			setRetryingConditions(rev, fmt.Sprintf("revision object collisions in phase %d\n%s", i, strings.Join(collidingObjs, "\n\n")))
 			return ctrl.Result{RequeueAfter: 10 * time.Second}, nil
 		}
 	}
 
+	if !rres.InTransistion() {
```
anik120, line 181: nit: 

```suggestion
	if !rres.InTransition() {
```

perdasilva: that comes from the boxcutter library - it will be fixed once we update it

api/v1/clusterextensionrevision_types.go
perdasilva: done

internal/operator-controller/controllers/clusterextensionrevision_controller.go
```diff
@@ -117,60 +118,31 @@ func (c *ClusterExtensionRevisionReconciler) reconcile(ctx context.Context, rev
...
+		werr := fmt.Errorf("establish watch: %v", err)
+		setRetryingConditions(rev, werr.Error())
+		return ctrl.Result{}, werr
 	}
 
 	rres, err := c.RevisionEngine.Reconcile(ctx, *revision, opts...)
 	if err != nil {
-		if rres != nil {
-			l.Error(err, "revision reconcile failed")
-			l.V(1).Info("reconcile failure report", "report", rres.String())
```
camilamacedo86, line 163: @perdasilva here we need keep debug logs.
I will clarify that in the code and create a test
OR we must remove those reports they are too verbose. 

https://issues.redhat.com/browse/OCPBUGS-62964

camilamacedo86: @pedjak @perdasilva You both have been working with this more recently.
I also don’t find the report very valuable in the logs — it’s confusing and hard to read.

What do you think about removing the report entirely?
And instead, formatting the message in a clearer way, similar to what we do in other places, so it becomes more meaningful? It would either solve the bug above

perdasilva: I've re-added it (but only in debug mode). Let's leave it like that for this PR and address this separately.

camilamacedo86: @perdasilva 

I got up and was looking at this today.
I’m thinking the best solution might be similar to what we do for crddiff.

We need a format to extract the report information and normalize it into both machine-readable messages and human-readable logs. Then we can use it for any situation—info or error—while normalizing and stripping out only the data that is actually validatable.

This definitely needs to be done in a separate PR, and we’ll need to discuss a better long-term approach for handling this.

Thank you for the attention !!!

c/c @pedjak

api/v1/clusterextension_types.go
```diff
@@ -482,8 +482,11 @@ type ClusterExtensionStatus struct {
 	// When Progressing is True and the Reason is Succeeded, the ClusterExtension is making progress towards a new state.
 	// When Progressing is True and the Reason is Retrying, the ClusterExtension has encountered an error that could be resolved on subsequent reconciliation attempts.
 	// When Progressing is False and the Reason is Blocked, the ClusterExtension has encountered an error that requires manual intervention for recovery.
+	// <opcon:experimental:description>
+	// When Progressing is True and Reason is RolloutInProgress, the ClusterExtension has one or more ClusterExtensionRevisions in active roll out.
```
pedjak: why not unifying here to have the same reason on the extension and revision? Right now we have `RolloutInProgress` on the extension and on the revision we have `RollingOut`,

api/v1/clusterextensionrevision_types.go
```diff
@@ -140,6 +138,7 @@ type ClusterExtensionRevisionStatus struct {
 
 // ClusterExtensionRevision is the Schema for the clusterextensionrevisions API
 // +kubebuilder:printcolumn:name="Available",type=string,JSONPath=`.status.conditions[?(@.type=='Available')].status`
+// +kubebuilder:printcolumn:name="Progressing",type=string,JSONPath=`.status.conditions[?(@.type=='Progressing')].status`
```
pedjak, line 212: without exposing the reason as well, status alone does not tell us enough.

internal/operator-controller/applier/boxcutter.go
```diff
@@ -397,6 +397,12 @@ func (bc *Boxcutter) apply(ctx context.Context, contentFS fs.FS, ext *ocv1.Clust
 	if progressingCondition == nil && availableCondition == nil && succeededCondition == nil {
 		return false, "New revision created", nil
 	} else if progressingCondition != nil && progressingCondition.Status == metav1.ConditionTrue {
+		switch progressingCondition.Reason {
+		case ocv1.ReasonSucceeded:
+			return true, "", nil
+		case ocv1.ClusterExtensionRevisionReasonRetrying:
+			return false, "", errors.New(progressingCondition.Message)
+		}
 		return false, progressingCondition.Message, nil
```
pedjak: please move this line into the switch above as `detault` option.

internal/operator-controller/controllers/clusterextensionrevision_controller.go
perdasilva: I don't necessarily disagree. But this is something we'll need to bake into the Boxcutter library itself. So, it's out of scope for this PR. I'll bring it up with Nico.

api/v1/clusterextensionrevision_types.go
perdasilva: I agree. Though, I'd say it at least tells us there's nothing blocking progression so I'd say it's better to have it as it is as opposed to not having it at all. Though, I agree we should revisit this as a compound value or additional column in a follow up.

api/v1/clusterextension_types.go
```diff
@@ -482,6 +482,9 @@ type ClusterExtensionStatus struct {
 	// When Progressing is True and the Reason is Succeeded, the ClusterExtension is making progress towards a new state.
 	// When Progressing is True and the Reason is Retrying, the ClusterExtension has encountered an error that could be resolved on subsequent reconciliation attempts.
 	// When Progressing is False and the Reason is Blocked, the ClusterExtension has encountered an error that requires manual intervention for recovery.
+	// <opcon:experimental:description>
+	// When Progressing is True and Reason is RollingOut, the ClusterExtension has one or more ClusterExtensionRevisions in active roll out.
```
pedjak, line 486: It would be nice to report in the PR description the changes to `ClusterExtension` conditions, similar to what you did for `ClusterExtensionRevision`

api/v1/clusterextensionrevision_types.go
```diff
@@ -190,17 +183,16 @@ type ClusterExtensionRevisionStatus struct {
 	// ClusterExtensionRevision.
 	//
 	// The Progressing condition represents whether the revision is actively rolling out:
-	//   - When status is True and reason is Progressing, the revision rollout is actively making progress and is in transition.
-	//   - When Progressing is not present, the revision is not currently in transition.
+	//   - When status is True and reason is RollingOut, the ClusterExtensionRevision rollout is actively making progress and is in transition.
+	//   - When status is True and reason is Retrying, the ClusterExtensionRevision has encountered an error that could be resolved on subsequent reconciliation attempts.
+	//   - When status is True and reason is Succeeded, the ClusterExtensionRevision has reached the desired state.
+	//   - When status is False and reason is Blocked, the ClusterExtensionRevision has encountered an error that requires manual intervention for recovery.
+	//   - When status is False and reason is Archived, the ClusterExtensionRevision is archived and not being actively reconciled.
```
pedjak, line 190: please update the table in the PR description, it looks it is out-of-date.

api/v1/clusterextensionrevision_types.go
```diff
@@ -190,17 +183,16 @@ type ClusterExtensionRevisionStatus struct {
...
-	//   - When status is False and reason is Incomplete, the revision rollout has not yet completed but no specific failures have been detected.
+	//   - When status is True and reason is ProbesSucceeded, the revision has been successfully rolled out and all objects pass their readiness probes.
 	//   - When status is False and reason is ProbeFailure, one or more objects are failing their readiness probes during rollout.
-	//   - When status is False and reason is ReconcileFailure, the revision has encountered a general reconciliation failure.
-	//   - When status is False and reason is RevisionValidationFailure, the revision failed preflight validation checks.
-	//   - When status is False and reason is PhaseValidationError, a phase within the revision failed preflight validation checks.
-	//   - When status is False and reason is ObjectCollisions, objects in the revision collide with existing cluster objects that cannot be adopted.
+	//   - When status is Unknown and reason is Reconciling, the revision has encountered an error that prevented it from observing the probes.
```
pedjak: the docs here and the table reported in the PR description are not in sync, the table has more rows.

internal/operator-controller/applier/boxcutter.go
```diff
@@ -397,7 +397,14 @@ func (bc *Boxcutter) apply(ctx context.Context, contentFS fs.FS, ext *ocv1.Clust
...
+		case ocv1.ReasonSucceeded:
+			return true, "", nil
+		case ocv1.ClusterExtensionRevisionReasonRetrying:
+			return false, "", errors.New(progressingCondition.Message)
+		default:
+			return false, progressingCondition.Message, nil
+		}
 	} else if availableCondition != nil && availableCondition.Status != metav1.ConditionTrue {
 		return false, "", errors.New(availableCondition.Message)
 	} else if succeededCondition != nil && succeededCondition.Status != metav1.ConditionTrue {
```
pedjak, line 410: [nit] would it possible to simplify/make more readable this big `if-else-swtich-else-else` block?

internal/operator-controller/controllers/clusterextensionrevision_controller.go
```diff
@@ -117,91 +117,50 @@ func (c *ClusterExtensionRevisionReconciler) reconcile(ctx context.Context, rev
...
+		werr := fmt.Errorf("establish watch: %v", err)
+		setRetryingConditions(rev, werr.Error())
+		return ctrl.Result{}, werr
 	}
 
 	rres, err := c.RevisionEngine.Reconcile(ctx, *revision, opts...)
+	if rres != nil {
+		// Log detailed reconcile reports only in debug mode (V(1)) to reduce verbosity.
+		l.V(1).Info("reconcile report", "report", rres.String())
+	}
```
pedjak: We should move this after we checked for the errors, because in the case of errors we should still log with `l.Error(..)`

internal/operator-controller/controllers/clusterextensionrevision_controller.go
pedjak: In case of error I think it would still very useful to log et error level, assuming that errors are happening occasionally.

internal/operator-controller/controllers/clusterextensionrevision_controller.go
```diff
@@ -605,3 +517,66 @@ var (
...
+	return meta.SetStatusCondition(&cer.Status.Conditions, metav1.Condition{
+		Type:               ocv1.ClusterExtensionRevisionTypeAvailable,
+		Status:             metav1.ConditionUnknown,
+		Reason:             reason,
+		Message:            message,
+		ObservedGeneration: cer.Generation,
+	})
+}
+
+func setArchivedConditions(cer *ocv1.ClusterExtensionRevision) bool {
```
pedjak: [nit] perhaps to rename to `markAsArchived` to match the `markAs*` functions?

internal/operator-controller/controllers/clusterextensionrevision_controller_test.go
```diff
@@ -440,7 +626,7 @@ func Test_ClusterExtensionRevisionReconciler_Reconcile_ValidationError_Retries(t
 	}
 }
 
-func Test_ClusterExtensionRevisionReconciler_Reconcile_Deletion(t *testing.T) {
+func Test_ClusterExtensionRevisionReconciler_Reconcile_Teardown(t *testing.T) {
```
pedjak: [nit] was rename needed?

test/e2e/cluster_extension_revision_test.go
```diff
@@ -0,0 +1,209 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	ocv1 "github.com/operator-framework/operator-controller/api/v1"
+	"github.com/operator-framework/operator-controller/internal/operator-controller/features"
+	. "github.com/operator-framework/operator-controller/internal/shared/util/testutils"
+	. "github.com/operator-framework/operator-controller/test/helpers"
+)
+
+func TestClusterExtensionRevision(t *testing.T) {
+	SkipIfFeatureGateDisabled(t, string(features.BoxcutterRuntime))
```
pedjak, line 28: it would be nice to split this large test into several smaller, focused on particular aspects:

* clusterextension install
* clusterextension update
* workload availability change
* etc.

Also, should we keep already existing test `TestClusterExtensionVersionUpdate` in `test/e2e/single_namespace_support_test.go`?

pedjak (commented):
please update/extend PR title to summarize what this PR is about.

api/v1/clusterextensionrevision_types.go
perdasilva: I think it's right: 

ProbesSucceeded/True
ProbeFailure/False
Reconciling/Unknown
Migrated/Unknown
Archived/Unknown

or is it the specific text?

I'm going to push an update anyway because in one text we say revision in another ClusterExtensionRevision. So, I'll make that consistent.

internal/operator-controller/applier/boxcutter.go
perdasilva: linting complains about nested-block complexity. I'll leave it as anyway this should go away with the ClusterExtension conditions PR, right?

test/e2e/cluster_extension_revision_test.go
perdasilva: Let's circle back to this in a follow-up. I think we need to think a bit more deeply how we want to handle this. Many of the regular lifecycle e2es are covered by the standard set. But, they lack checks for the `Availability` condition. 

Maybe through the cucumber exercise we could add additional conditions checks to the install/update test if boxcutter is on then strip this one down to just the availability check?

Let's keep the `TestClusterExtensionVersionUpdate` as is for this PR and either remove it in a follow-up, or maybe better, move it to the standard set? I think it's anyway more meaningful there, no?

internal/operator-controller/controllers/clusterextensionrevision_controller.go
perdasilva: if there's a collision it will perma-occur. I think this was the original problem.

api/v1/clusterextension_types.go
perdasilva: Done

api/v1/clusterextensionrevision_types.go
perdasilva: Done

internal/operator-controller/controllers/clusterextensionrevision_controller.go
perdasilva: I'll leave it still as debug for now because the report can be VERY verbose. If we still want it, let's address it it in a follow-up.

internal/operator-controller/controllers/clusterextensionrevision_controller.go
perdasilva: Done

internal/operator-controller/controllers/clusterextensionrevision_controller_test.go
perdasilva: I've reverted. But just curious about the concern. Are the adverse effects I should be considering with such changes? or are they generally benign?

internal/operator-controller/applier/boxcutter.go
pedjak: :shrug:

internal/operator-controller/controllers/clusterextensionrevision_controller_test.go
pedjak: My comment was marked as [nit], I was more curious to understand the motivation and the improved achieved.

internal/operator-controller/controllers/clusterextensionrevision_controller.go
pedjak: > I'll leave it still as debug for now because the report can be VERY verbose. If we still want it, let's address it it in a follow-up.

Leaving as debug is fine for regular reconcile path (i.e. no errors). I was asking to print it to the debug level, only if there are no errors. In case of errors, I would still log it on the error level, so that could be caught  in production (debug level is typically not set there).

api/v1/clusterextensionrevision_types.go
```diff
@@ -190,22 +183,21 @@ type ClusterExtensionRevisionStatus struct {
...
-	//   - When status is Unknown and reason is Migrated, the revision was migrated from an existing release and object status probe results have not yet been observed.
+	//   - When status is Unknown and reason is Reconciling, the ClusterExtensionRevision has encountered an error that prevented it from observing the probes.
+	//   - When status is Unknown and reason is Archived, the ClusterExtensionRevision has been archived and its objects have been torn down.
+	//   - When status is Unknown and reason is Migrated, the ClusterExtensionRevision was migrated from an existing release and object status probe results have not yet been observed.
 	//
 	// The Succeeded condition represents whether the revision has successfully completed its rollout:
-	//   - When status is True and reason is RolloutSuccess, the revision has successfully completed its rollout. This condition is set once and persists even if the revision later becomes unavailable.
+	//   - When status is True and reason is RolloutSuccess, the ClusterExtensionRevision has successfully completed its rollout. This condition is set once and persists even if the revision later becomes unavailable.
```
tmshort: I'm not seeing `RolloutSuccess` as a valid reason (it was removed)

api/v1/clusterextensionrevision_types.go
```diff
@@ -190,22 +183,21 @@ type ClusterExtensionRevisionStatus struct {
 	// ClusterExtensionRevision.
 	//
 	// The Progressing condition represents whether the revision is actively rolling out:
-	//   - When status is True and reason is Progressing, the revision rollout is actively making progress and is in transition.
-	//   - When Progressing is not present, the revision is not currently in transition.
+	//   - When status is True and reason is RollingOut, the ClusterExtensionRevision rollout is actively making progress and is in transition.
+	//   - When status is True and reason is Retrying, the ClusterExtensionRevision has encountered an error that could be resolved on subsequent reconciliation attempts.
+	//   - When status is True and reason is Succeeded, the ClusterExtensionRevision has reached the desired state.
```
tmshort, line 188: I don't see `Succeeded` as a reason (There's `ProbesSucceeded`, but it's not the same)

api/v1/clusterextensionrevision_types.go
```diff
@@ -190,22 +183,21 @@ type ClusterExtensionRevisionStatus struct {
 	// ClusterExtensionRevision.
 	//
 	// The Progressing condition represents whether the revision is actively rolling out:
-	//   - When status is True and reason is Progressing, the revision rollout is actively making progress and is in transition.
-	//   - When Progressing is not present, the revision is not currently in transition.
+	//   - When status is True and reason is RollingOut, the ClusterExtensionRevision rollout is actively making progress and is in transition.
```
tmshort, line 186: I don't see `RollingOut` as a reason.

internal/operator-controller/controllers/clusterextensionrevision_controller.go
```diff
@@ -117,91 +117,50 @@ func (c *ClusterExtensionRevisionReconciler) reconcile(ctx context.Context, rev
...
 
 	if !rev.DeletionTimestamp.IsZero() || rev.Spec.LifecycleState == ocv1.ClusterExtensionRevisionLifecycleStateArchived {
 		return c.teardown(ctx, rev, revision)
 	}
 
+	revVersion := rev.GetAnnotations()[labels.BundleVersionKey]
 	//
 	// Reconcile
 	//
 	if err := c.ensureFinalizer(ctx, rev, clusterExtensionRevisionTeardownFinalizer); err != nil {
```
tmshort, line 132: Status update was removed here? Is that on purpose?

internal/operator-controller/controllers/clusterextensionrevision_controller.go
```diff
@@ -605,3 +517,66 @@ var (
...
+		Reason:             reason,
+		Message:            message,
+		ObservedGeneration: cer.Generation,
+	})
+}
+
+func markAsArchived(cer *ocv1.ClusterExtensionRevision) bool {
+	const msg = "revision is archived"
+	updated := markAsNotProgressing(cer, ocv1.ClusterExtensionRevisionReasonArchived, msg)
+	return markAsAvailableUnknown(cer, ocv1.ClusterExtensionRevisionReasonArchived, msg) || updated
```
tmshort, line 581: This feels like it ought to be "Unavailable" rather than "Unknown"? Is there a reason for the distinction here?

internal/operator-controller/controllers/clusterextensionrevision_controller.go
```diff
@@ -605,3 +517,66 @@ var (
 		FieldB: ".status.replicas",
 	}
 )
+
+func setRetryingConditions(cer *ocv1.ClusterExtensionRevision, message string) {
+	markAsProgressing(cer, ocv1.ClusterExtensionRevisionReasonRetrying, message)
+	if meta.FindStatusCondition(cer.Status.Conditions, ocv1.ClusterExtensionRevisionTypeAvailable) != nil {
+		markAsAvailableUnknown(cer, ocv1.ClusterExtensionRevisionReasonReconciling, message)
```
rashmigottipati, line 524: Using `ReasonReconciling` here does not seem apt to me as this is separate from probe observation failures. I think we should use `ReasonRetrying` here.

```suggestion
		markAsAvailableUnknown(cer, ocv1.ClusterExtensionRevisionReasonRetrying, message)
```

internal/operator-controller/controllers/clusterextensionrevision_controller.go
```diff
@@ -605,3 +517,66 @@ var (
 		FieldB: ".status.replicas",
 	}
 )
+
+func setRetryingConditions(cer *ocv1.ClusterExtensionRevision, message string) {
+	markAsProgressing(cer, ocv1.ClusterExtensionRevisionReasonRetrying, message)
```
rashmigottipati, line 522: setting `Progressing=True` here can conflict with `Available=True`, should we consider setting `Available=Unknown`?

api/v1/clusterextensionrevision_types.go
perdasilva: Good catch

api/v1/clusterextensionrevision_types.go
perdasilva: It's in common_types: https://github.com/operator-framework/operator-controller/pull/2340/files#diff-7ad21c885e2b008e93dfa228f394825987e36cb8e539fe80be5661d692dbc47aR27

api/v1/clusterextensionrevision_types.go
perdasilva: It's an existing reason in common_types: https://github.com/operator-framework/operator-controller/pull/2340/files#diff-7ad21c885e2b008e93dfa228f394825987e36cb8e539fe80be5661d692dbc47aL35

internal/operator-controller/controllers/clusterextensionrevision_controller.go
perdasilva: If the Available status conditions exists, we set it to unknown below. If it hasn't been set yet, we leave it unset.

internal/operator-controller/controllers/clusterextensionrevision_controller.go
perdasilva: I'm wondering if we'll remove it all together in a follow-up. The reason for it being unknown is that there are situation in which it could be considered to be available. If rev-2 is rev-1 + some random resource, its possible that rev-1 could be available.

internal/operator-controller/controllers/clusterextensionrevision_controller.go
perdasilva: We felt that it was unnecessary as the errors should be rare and short lived...

pedjak: > We felt that it was unnecessary as the errors should be rare and short lived...

Yes, if there is an error at adding finalizer (sending API request), there is very high chance that updating the status is also going to fail. Hoping that the issue is short-lived/transient, we just retry without bugging user about it.

api/v1/clusterextensionrevision_types.go
perdasilva: Fixed

rashmigottipati (approved):
/lgtm

PR #2344, opened 2025-11-18
🌱 (fix) ClusterExtensionRevision had inconsistent owner labeling and confusing code

camilamacedo86:
## Problem

ClusterExtensionRevision had inconsistent owner labeling and confusing code:

1. **Inconsistent Labels**: Managed objects use `owner-kind` + `owner-name`, but ClusterExtensionRevision used only `owner`
2. **Confusing Code**: Variable named `storeLbls` actually contained annotations
3. **Unclear API**: Hard to understand what parameters were for

## Solution

**Fixed owner label consistency** and **improved code clarity** while keeping bundle metadata in annotations (no breaking changes).

### Changes

✅ **Owner Label Unification**
- All objects now use `owner-kind` + `owner-name` consistently

✅ **Code Clarity**
- Renamed `storeLbls` → separate `revisionLabels` + `revisionAnnotations`
- Clear function signatures showing labels vs annotations

✅ **Bundle Metadata Location**
- Kept in annotations (informational, no query use case, no length limits)

### Why This Matters

- **Consistency**: Same labeling pattern across all OLM resources
- **Maintainability**: Clear, self-documenting code
- **Safety**: No breaking changes, no length restrictions
- **Simplicity**: Easier to review and merge

### DIFF
--- a/internal/operator-controller/applier/boxcutter.go
+++ b/internal/operator-controller/applier/boxcutter.go
@@ -26,7 +26,6 @@ import (
 	helmclient "github.com/operator-framework/helm-operator-plugins/pkg/client"
 
 	ocv1 "github.com/operator-framework/operator-controller/api/v1"
-	"github.com/operator-framework/operator-controller/internal/operator-controller/controllers"
 	"github.com/operator-framework/operator-controller/internal/operator-controller/labels"
 )
 
@@ -183,7 +182,7 @@ func (r *SimpleRevisionGenerator) buildClusterExtensionRevision(
 		ObjectMeta: metav1.ObjectMeta{
 			Annotations: annotations,
 			Labels: map[string]string{
-				controllers.ClusterExtensionRevisionOwnerLabel: ext.Name,
+				labels.OwnerNameKey: ext.Name,
 			},
 		},
 		Spec: ocv1.ClusterExtensionRevisionSpec{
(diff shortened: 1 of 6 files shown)


test/upgrade-e2e/post_upgrade_test.go
```diff
@@ -161,6 +161,30 @@ func TestClusterExtensionAfterOLMUpgrade(t *testing.T) {
...
+
+	// Wait for new revision to be created with correct label structure
+	t.Log("Waiting for new ClusterExtensionRevision to be created")
+	require.EventuallyWithT(t, func(ct *assert.CollectT) {
+		var newRevList ocv1.ClusterExtensionRevisionList
+		err := c.List(ctx, &newRevList)
+		require.NoError(ct, err)
+		require.NotEmpty(ct, newRevList.Items, "New revision should be created")
+	}, 2*time.Minute, time.Second)
+	// -----------------------------------------------------------------------------------------------------------
```
camilamacedo86: The e2e upgrade experimental/Test can not pass. 
- OLD code installed (before our change in this PR): Creates ClusterExtensionRevision with metadata in annotations:
-  Upgrade happens - NEW code deployed (our change)
- NEW code reads from labels so clusterExtension.Status.Install.Bundle.Version is empty

That is why we have the above code with TODO.
We can remove it after it get merge

camilamacedo86: See: https://github.com/operator-framework/operator-controller/actions/runs/19478013715/job/55742664496?pr=2344

joelanford:
Is there a use case for querying by any of the values that are currently in annotations? As far as I know, those annotations are there to serve as hints to the resolver (?), not to be queried.

If there aren't any specific query use cases yet, maybe we should hold off on changing this right now.

Annotations have fewer restrictions on their value, so I think we should only put values in labels when we know we need to query by them?

Maybe @pedjak would have a better rubric than I do on the label vs annotation debate in this case.

camilamacedo86:
Hi @joelanford 

Thank you for your input.

> Is there a use case for querying by any of the values that are currently in annotations? As far as I know, those annotations are there to serve as hints to the resolver (?), not to be queried.
> If there aren't any specific query use cases yet, maybe we should hold off on changing this right now.
> Annotations have fewer restrictions on their value, so I think we should only put values in labels when we know we need to query by them?
>Maybe @pedjak would have a better rubric than I do on the label vs annotation debate in this case.

This PR is doing more than just moving the annotations to labels. I reverted the changes and kept the other fix. **BUT**I will open a new PR only for that so the smaller change doesn’t get blocked by this discussion.**
I won’t push strongly for this if the team disagrees.

PR #2346, opened 2025-11-18
:seedling: Use local cache to check kind version

tmshort:
Use the local cached source of kind to determine the k8s version it supports. It avoids hitting github if possible.

# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/hack/tools/validate_kindest_node.sh
+++ b/hack/tools/validate_kindest_node.sh
@@ -4,8 +4,24 @@
 # Extract the version of kind, by removing the "${GOBIN}/kind-" prefix
 KIND=${KIND#${GOBIN}/kind-}
 
-# Get the version of the image
-KIND_VER=$(curl -L -s https://github.com/kubernetes-sigs/kind/raw/refs/tags/${KIND}/pkg/apis/config/defaults/image.go | grep -Eo 'v[0-9]+\.[0-9]+')
+GOMODCACHE=$(go env GOMODCACHE)
+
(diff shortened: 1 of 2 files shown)


tmshort:
github had issues yesterday; this avoids git hub when running kind locally.

pedjak (approved):
/lgtm

hack/tools/validate_kindest_node.sh
```diff
@@ -4,8 +4,24 @@
...
+
+# Get the version of the image from the local kind build
+if [ -d "${GOMODCACHE}" ]; then
+    KIND_VER=$(grep -Eo "${REGEX}" ${GOMODCACHE}/sigs.k8s.io/kind@${KIND}/pkg/apis/config/defaults/image.go)
+fi
+
+# Get the version of the image from github
+if [ -z "${KIND_VER}" ]; then
+    KIND_VER=$(curl -L -s https://github.com/kubernetes-sigs/kind/raw/refs/tags/${KIND}/pkg/apis/config/defaults/image.go | grep -Eo "${REGEX}")
+fi
```
pedjak, line 19: is there a situation when we cannot resolve `KIND_VER` from the local cache? Can we remove this actually?

tmshort: This is the original check that broke when github was down yesterday. It's possible someone may have cleaned up their local cache, so I'd rather just leave it as a "just-in-case".

PR #2347, opened 2025-11-19
:bug: Fix an unclear validation error when configuring watchNamespace on an operator restricted to AllNamespaces mode.

camilamacedo86:
## Summary of Changes

When a ClusterExtension with `watchNamespace` config is applied to an AllNamespaces-only, the error message has been improved to be more descriptive.

### Scenario: User tries to configure watchNamespace on AllNamespaces-only operator

**BEFORE:**
```
invalid ClusterExtension configuration: invalid configuration: unknown field "watchNamespace"
```

**NOW:**
```
invalid ClusterExtension configuration: invalid configuration: watchNamespace configuration is not supported when the content only supports "AllNamespaces" install mode
```

## Motivation

https://issues.redhat.com//browse/OCPBUGS-63611

### DIFF
--- a/internal/operator-controller/config/config.go
+++ b/internal/operator-controller/config/config.go
@@ -45,6 +45,9 @@ const (
 	// FormatSingleNamespaceInstallMode defines the format check to ensure that
 	// the watchNamespace must differ from install namespace
 	FormatSingleNamespaceInstallMode = "singleNamespaceInstallMode"
+	// FormatAllNamespacesOnlyInstallMode defines the format check to reject
+	// watchNamespace when only AllNamespaces mode is supported (registry+v1 specific)
+	FormatAllNamespacesOnlyInstallMode = "allNamespacesOnlyInstallMode"
 )
 
 // SchemaProvider lets each package format type describe what configuration it accepts.
@@ -190,6 +193,14 @@ func validateConfigWithSchema(configBytes []byte, schema map[string]any, install
 				return fmt.Errorf("invalid value %q: watchNamespace must be different from %q (the install namespace) because this operator uses SingleNamespace install mode to watch a different namespace", str, installNamespace)
 			}
 			return nil
+		},
+	})
+	compiler.RegisterFormat(&jsonschema.Format{
+		Name: FormatAllNamespacesOnlyInstallMode,
+		Validate: func(value interface{}) error {
+			// Always reject - this format is used when AllNamespaces is the only supported mode
+			// and watchNamespace configuration doesn't make sense
+			return fmt.Errorf("watchNamespace configuration is not supported when the content only supports AllNamespaces install mode")
 		},
 	})
 
--- a/internal/operator-controller/config/config_test.go
+++ b/internal/operator-controller/config/config_test.go
@@ -80,22 +80,10 @@ func Test_UnmarshalConfig(t *testing.T) {
 			expectedErrMessage:    `got object, want string`,
 		},
 		{
-			name:                  "rejects with unknown field when install modes {AllNamespaces}",
+			name:                  "rejects with descriptive message when install modes {AllNamespaces}",
 			supportedInstallModes: []v1alpha1.InstallModeType{v1alpha1.InstallModeTypeAllNamespaces},
 			rawConfig:             []byte(`{"watchNamespace": "some-namespace"}`),
-			expectedErrMessage:    `unknown field "watchNamespace"`,
-		},
-		{
-			name:                  "rejects with unknown field when install modes {MultiNamespace}",
-			supportedInstallModes: []v1alpha1.InstallModeType{v1alpha1.InstallModeTypeMultiNamespace},
-			rawConfig:             []byte(`{"watchNamespace": "some-namespace"}`),
-			expectedErrMessage:    `unknown field "watchNamespace"`,
-		},
-		{
-			name:                  "reject with unknown field when install modes {AllNamespaces, MultiNamespace}",
-			supportedInstallModes: []v1alpha1.InstallModeType{v1alpha1.InstallModeTypeAllNamespaces, v1alpha1.InstallModeTypeMultiNamespace},
-			rawConfig:             []byte(`{"watchNamespace": "some-namespace"}`),
-			expectedErrMessage:    `unknown field "watchNamespace"`,
+			expectedErrMessage:    `watchNamespace configuration is not supported when the content only supports AllNamespaces install mode`,
 		},
 		{
 			name:                  "reject with required field when install modes {OwnNamespace} and watchNamespace is null",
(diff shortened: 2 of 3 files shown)


internal/operator-controller/config/config.go
```diff
@@ -267,6 +267,11 @@ func formatSingleError(errUnit jsonschema.OutputUnit) string {
 		// Unknown/additional field
 		fieldName := extractFieldNameFromMessage(errUnit.Error)
 		if fieldName != "" {
+			// Special case: if watchNamespace is rejected as unknown, it means only
+			// AllNamespaces install mode is supported
+			if fieldName == "watchNamespace" {
+				return fmt.Sprintf("field %q not allowed: watchNamespace configuration is not supported when the content only supports \"AllNamespaces\" install mode", fieldName)
```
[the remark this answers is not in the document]
camilamacedo86: Not valid we do not support MultiNS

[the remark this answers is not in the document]
camilamacedo86: Not valid

internal/operator-controller/config/config_test.go
```diff
@@ -80,22 +80,10 @@ func Test_UnmarshalConfig(t *testing.T) {
...
-			supportedInstallModes: []v1alpha1.InstallModeType{v1alpha1.InstallModeTypeMultiNamespace},
-			rawConfig:             []byte(`{"watchNamespace": "some-namespace"}`),
-			expectedErrMessage:    `unknown field "watchNamespace"`,
-		},
-		{
-			name:                  "reject with unknown field when install modes {AllNamespaces, MultiNamespace}",
-			supportedInstallModes: []v1alpha1.InstallModeType{v1alpha1.InstallModeTypeAllNamespaces, v1alpha1.InstallModeTypeMultiNamespace},
-			rawConfig:             []byte(`{"watchNamespace": "some-namespace"}`),
-			expectedErrMessage:    `unknown field "watchNamespace"`,
+			expectedErrMessage:    `field "watchNamespace" not allowed: watchNamespace configuration is not supported when the content only supports "AllNamespaces" install mode`,
```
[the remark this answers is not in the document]
camilamacedo86: Not valid . Do not address the issue

pedjak:
not sure if this qualifies as a bug fix, perhaps is just a :seedling: ?

camilamacedo86:
Hi @pedjak,

Thank you for review 🥇 

> not sure if this qualifies as a bug fix, perhaps is just a 🌱 ?

This is a **bug**.

**Before**: The message shown to the end user was unclear and unhelpful.
**Now**: The fix updates the message the end user sees, improving clarity.
Because it **changes end-user behavior/output**, it should be marked as a **bug fix** 
That is why it should fix a OCPBUG: https://issues.redhat.com/browse/OCPBUGS-63611

**When to use 🌱** 

Use the 🌱 label ONLY when the change **has zero impact on end users**. This includes:

- No new or modified behavior
- No new or changed error messages
- No new logs
- No new labels, fields, or resources
- No visible changes in how the system behaves or is used

In other words, 🌱 **is reserved strictly for internal, non-functional updates such as
refactoring, cleanup, infra changes or code organisation. If an end user could notice the change
in any way, the 🌱 label should NOT be used.**

internal/operator-controller/config/config_test.go
perdasilva: We shouldn't merge this. This type of error message treats install modes as a first class thing that concerns all bundles. While this is implicitly true while we only support registry+v1 bundles we shouldn't make it explicit.

Adding a `watchNamespace` field to the configuration against a bundle that doesn't support it should be treated no differently than if the user adds any arbitrary non-existent field.

rashmigottipati: That makes sense. 
Currently, configuring `watchNamespace` in `AllNamespaces` mode prevents the ClusterExtension from progressing. In fact, any unknown field in the configuration will block rollout, as the extension treats all unrecognized fields the same way. 

I agree that we shouldn’t make this behavior explicit.

tmshort: /hold 
@perdasilva are you asking for this error message to be changed?

perdasilva: I think we can close this PR entirely. I don't think it's a bug.

rashmigottipati: IIUC, @perdasilva was saying that we don’t need a special error message for watchNamespace. Any unknown field in the configuration will block the rollout anyways, and that behavior is already sufficient. 

oops didn't see Per's comment before writing mine :sweat_smile:

tmshort: OK, I agree, now that I'm looking at the diffs, this is specific to one error message, and when a bundle is AllNamespaces, it should ignore `watchNamespaces`. I understand why it was done, though

camilamacedo86: I agree — we can’t solve it that way.
I was without coffee this morning. ☕🙂

We need to fix this specifically for CSV / bundle registry v1 only.

camilamacedo86:
/hold cancel

@perdasilva @rashmigottipati @tmshort 

Sorry about my slip-up earlier. Everything is all set now! Thanks for the support

internal/operator-controller/config/config_test.go
camilamacedo86: Fixed

tmshort:
The headline and PR/commit summary do not patch what the PR now does, which is to add a new install mode.

joelanford:
@perdasilva can correct me if I am wrong, but my understanding of the overall intended progression of the registry+v1 configuration schema is that we will:
1. Implement support for subscription.spec.config
2. Shift to generating (and validating with) JSON Schema for the combined watchNamespace and subscription.spec.config configuration based on the shape of the registry+v1 bundle.
3. Be positioned to handle other formats (e.g. helm charts) that already ship with JSON schemas of their configuration.

With that in mind (and if that is still true), I don't think we should be spending cycles implementing and reviewing the current validation implementation. The focus (I think) is on the order of operations above.

joelanford:
Dropped a comment in the bug as well to try to explain our stance: https://issues.redhat.com/browse/OCPBUGS-63611?focusedId=28504231&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-28504231

Design decision doc also mentions this descoping aspect of OLMv1 near the very top: https://operator-framework.github.io/operator-controller/project/olmv1_design_decisions/#watched-namespaces-cannot-be-configured-in-a-first-class-api

camilamacedo86:
Hi @joelanford 

Regards your comment: https://github.com/operator-framework/operator-controller/pull/2347#issuecomment-3554643834
> @perdasilva can correct me if I am wrong, but my understanding of the overall intended progression of the registry+v1 configuration schema is that we will:

Yes, that is ALL true but we either have the option to create a customize validations and messages **if we need to do so.**  You can check the design in: https://github.com/operator-framework/operator-controller/pull/2316 

But I completely agree that we can’t create a custom message or behavior for every single flag and option, as that would lead to infinite maintenance cycles. I appreciate your comment in the JIRA, and I’m closing this one as won’t do.

PR #2348, opened 2025-11-19
🐛ClusterExtension configurations using watchNamespace: "" (empty string) no longer fail with a confusing validation error.

camilamacedo86:
## Motivation
https://issues.redhat.com//browse/OCPBUGS-63612

### DIFF
--- a/internal/operator-controller/config/config.go
+++ b/internal/operator-controller/config/config.go
@@ -79,7 +79,8 @@ func newConfig(data map[string]any) *Config {
 }
 
 // GetWatchNamespace returns the watchNamespace value if present in the configuration.
-// Returns nil if watchNamespace is not set or is explicitly set to null.
+// Returns nil if watchNamespace is not set, is explicitly set to null, or is set to an empty string.
+// An empty string is treated as equivalent to "not configured" (AllNamespaces mode).
 func (c *Config) GetWatchNamespace() *string {
 	if c == nil || *c == nil {
 		return nil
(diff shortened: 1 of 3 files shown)


perdasilva:
This is the desired behavior. OLM should not know about the semantics of a bundle configuration. A bundle that only support AllNamespaces does not provide a `watchNamespace` configuration. Therefore, `watchNamespace` should be treated as any other non-existing configuration key.

tmshort:
> This is the desired behavior. OLM should not know about the semantics of a bundle configuration. A bundle that only support AllNamespaces does not provide a `watchNamespace` configuration. Therefore, `watchNamespace` should be treated as any other non-existing configuration key.

@perdasilva do you not approve (although you gave it LGTM), or are you asking for someone else to tag it approved?

perdasilva:
> > This is the desired behavior. OLM should not know about the semantics of a bundle configuration. A bundle that only support AllNamespaces does not provide a `watchNamespace` configuration. Therefore, `watchNamespace` should be treated as any other non-existing configuration key.
> 
> @perdasilva do you not approve (although you gave it LGTM), or are you asking for someone else to tag it approved?

I did originally approve. But then it occurred to me it's not right and I took the label away. I've put it on hold now until I can clarify everything with Camila

camilamacedo86:
I’m closing this one.
I agree with @perdasilva.

I get up today, and start to look on it and I don’t think it’s maintainable for us to support that. It would require normalizing every option and flag, which leads to a worse UX overall. Some options treat an empty string as nil, others don’t, and we don’t have a reliable or consistent way to handle that across the board.

Close it. We should not do it.

PR #2349, opened 2025-11-19
:warning: fix: rename owner label and clarify variable naming

camilamacedo86:
Rename ClusterExtensionRevision owner label from `owner` to `owner-name` for consistency. Also rename misleading `storeLbls` variable to `revisionAnnotations` to clarify it contains annotations, not labels.

This removes the duplicate ClusterExtensionRevisionOwnerLabel constant and uses labels.OwnerNameKey directly throughout the codebase.

⚠️ Breaks the upgrade. However, we did not promoted yet so we can move forward with it: https://github.com/operator-framework/operator-controller/actions/runs/19496963917/job/55801395894?pr=2349#step:4:493 . If we re-install and install would pass.

### DIFF
--- a/internal/operator-controller/applier/boxcutter_test.go
+++ b/internal/operator-controller/applier/boxcutter_test.go
@@ -28,7 +28,6 @@ import (
 
 	ocv1 "github.com/operator-framework/operator-controller/api/v1"
 	"github.com/operator-framework/operator-controller/internal/operator-controller/applier"
-	"github.com/operator-framework/operator-controller/internal/operator-controller/controllers"
 	"github.com/operator-framework/operator-controller/internal/operator-controller/labels"
 )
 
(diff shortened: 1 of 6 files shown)


camilamacedo86:
@tmshort 

⚠️ Breaks the upgrade. However, we did not promoted yet so we can move forward with it: https://github.com/operator-framework/operator-controller/actions/runs/19496963917/job/55801395894?pr=2349#step:4:493 . If we re-install and install would pass.

Could you help us by overwriting it?

internal/operator-controller/applier/boxcutter_test.go
```diff
@@ -86,7 +85,7 @@ func Test_SimpleRevisionGenerator_GenerateRevisionFromHelmRelease(t *testing.T)
 				"olm.operatorframework.io/package-name":     "my-package",
 			},
 			Labels: map[string]string{
-				"olm.operatorframework.io/owner": "test-123",
+				"olm.operatorframework.io/owner-name": "test-123",
```
tmshort: Should `labels.OwnerNameKey` be used here, and in other places?
(Which might just be the `t.Log()` below.)

camilamacedo86: I think it is the only place that is not using the const so make sense besides be a test

camilamacedo86: @tmshort done

Could you pelase lGTM and allow it get merged?

PR #2350, opened 2025-11-19
:book: Update ClusterExtensionRevision API docs

perdasilva:
# Description

Update the ClusterExtensionRevision API docstrings to provide more thorough API documentation for the user.

Note: changes were generated using AI tools and manually updated

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/api/v1/clusterextensionrevision_types.go
+++ b/api/v1/clusterextensionrevision_types.go
@@ -24,11 +24,15 @@ import (
 const (
 	ClusterExtensionRevisionKind = "ClusterExtensionRevision"
 
-	// Condition Types
+	// ClusterExtensionRevisionTypeAvailable is the condition type that represents whether the
+	// ClusterExtensionRevision is available and has been successfully rolled out.
 	ClusterExtensionRevisionTypeAvailable = "Available"
+
+	// ClusterExtensionRevisionTypeSucceeded is the condition type that represents whether the
+	// ClusterExtensionRevision rollout has succeeded.
 	ClusterExtensionRevisionTypeSucceeded = "Succeeded"
 
-	// Condition Reasons
+	// Condition reasons
 	ClusterExtensionRevisionReasonAvailable                 = "Available"
 	ClusterExtensionRevisionReasonReconcileFailure          = "ReconcileFailure"
 	ClusterExtensionRevisionReasonRevisionValidationFailure = "RevisionValidationFailure"
@@ -44,22 +48,47 @@ const (
 
 // ClusterExtensionRevisionSpec defines the desired state of ClusterExtensionRevision.
 type ClusterExtensionRevisionSpec struct {
-	// Specifies the lifecycle state of the ClusterExtensionRevision.
+	// lifecycleState specifies the lifecycle state of the ClusterExtensionRevision.
+	//
+	// When set to "Active" (the default), the revision is actively managed and reconciled.
+	// When set to "Archived", the revision is inactive and any resources not managed by a subsequent revision are deleted.
+	// The revision is removed from the owner list of all objects previously under management.
+	// All objects that did not transition to a succeeding revision are deleted.
+	//
+	// Once a revision is set to "Archived", it cannot be un-archived.
 	//
 	// +kubebuilder:default="Active"
-	// +kubebuilder:validation:Enum=Active;Paused;Archived
-	// +kubebuilder:validation:XValidation:rule="oldSelf == 'Active' || oldSelf == 'Paused' || oldSelf == 'Archived' && oldSelf == self", message="can not un-archive"
+	// +kubebuilder:validation:Enum=Active;Archived
+	// +kubebuilder:validation:XValidation:rule="oldSelf == 'Active' || oldSelf == 'Archived' && oldSelf == self", message="cannot un-archive"
 	LifecycleState ClusterExtensionRevisionLifecycleState `json:"lifecycleState,omitempty"`
-	// Revision is a sequence number representing a specific revision of the ClusterExtension instance.
-	// Must be positive. Each ClusterExtensionRevision of the same parent ClusterExtension needs to have
-	// a unique value assigned. It is immutable after creation. The new revision number must always be previous revision +1.
+
+	// revision is a required, immutable sequence number representing a specific revision
+	// of the parent ClusterExtension.
+	//
+	// The revision field must be a positive integer.
+	// Each ClusterExtensionRevision belonging to the same parent ClusterExtension must have a unique revision number.
+	// The revision number must always be the previous revision number plus one, or 1 for the first revision.
 	//
 	// +kubebuilder:validation:Required
 	// +kubebuilder:validation:Minimum:=1
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf", message="revision is immutable"
 	Revision int64 `json:"revision"`
-	// Phases are groups of objects that will be applied at the same time.
-	// All objects in the phase will have to pass their probes in order to progress to the next phase.
+
+	// phases is an optional, immutable list of phases that group objects to be applied together.
+	//
+	// Objects are organized into phases based on their Group-Kind. Common phases include:
+	//   - namespaces: Namespace objects
+	//   - policies: ResourceQuota, LimitRange, NetworkPolicy objects
+	//   - rbac: ServiceAccount, Role, RoleBinding, ClusterRole, ClusterRoleBinding objects
+	//   - crds: CustomResourceDefinition objects
+	//   - storage: PersistentVolume, PersistentVolumeClaim, StorageClass objects
+	//   - deploy: Deployment, StatefulSet, DaemonSet, Service, ConfigMap, Secret objects
+	//   - publish: Ingress, APIService, Route, Webhook objects
+	//
+	// All objects in a phase are applied in no particular order.
+	// The revision progresses to the next phase only after all objects in the current phase pass their readiness probes.
+	//
+	// Once set, even if empty, the phases field is immutable.
 	//
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf || oldSelf.size() == 0", message="phases is immutable"
 	// +listType=map
@@ -75,33 +104,62 @@ const (
 	// ClusterExtensionRevisionLifecycleStateActive / "Active" is the default lifecycle state.
 	ClusterExtensionRevisionLifecycleStateActive ClusterExtensionRevisionLifecycleState = "Active"
 	// ClusterExtensionRevisionLifecycleStatePaused / "Paused" disables reconciliation of the ClusterExtensionRevision.
-	// Only Status updates will still propagated, but object changes will not be reconciled.
+	// Object changes will not be reconciled. However, status updates will be propagated.
 	ClusterExtensionRevisionLifecycleStatePaused ClusterExtensionRevisionLifecycleState = "Paused"
-	// ClusterExtensionRevisionLifecycleStateArchived / "Archived" disables reconciliation while also "scaling to zero",
-	// which deletes all objects that are not excluded via the pausedFor property and
-	// removes itself from the owner list of all other objects previously under management.
+	// ClusterExtensionRevisionLifecycleStateArchived / "Archived" archives the revision for historical or auditing purposes.
+	// The revision is removed from the owner list of all other objects previously under management and all objects
+	// that did not transition to a succeeding revision are deleted.
 	ClusterExtensionRevisionLifecycleStateArchived ClusterExtensionRevisionLifecycleState = "Archived"
 )
 
-// ClusterExtensionRevisionPhase are groups of objects that will be applied at the same time.
-// All objects in the a phase will have to pass their probes in order to progress to the next phase.
+// ClusterExtensionRevisionPhase represents a group of objects that are applied together. The phase is considered
+// complete only after all objects pass their status probes.
 type ClusterExtensionRevisionPhase struct {
-	// Name identifies this phase.
+	// name is a required identifier for this phase.
+	//
+	// phase names must follow the DNS label standard as defined in [RFC 1123].
+	// They must contain only lowercase alphanumeric characters or hyphens (-),
+	// start and end with an alphanumeric character, and be no longer than 63 characters.
+	//
+	// Common phase names include: namespaces, policies, rbac, crds, storage, deploy, publish.
+	//
+	// [RFC 1123]: https://tools.ietf.org/html/rfc1123
 	//
 	// +kubebuilder:validation:MaxLength=63
 	// +kubebuilder:validation:Pattern=`^[a-z]([-a-z0-9]*[a-z0-9])?$`
 	Name string `json:"name"`
-	// Objects are a list of all the objects within this phase.
+
+	// objects is a required list of all Kubernetes objects that belong to this phase.
+	//
+	// All objects in this list are applied to the cluster in no particular order.
 	Objects []ClusterExtensionRevisionObject `json:"objects"`
 }
 
-// ClusterExtensionRevisionObject contains an object and settings for it.
+// ClusterExtensionRevisionObject represents a Kubernetes object to be applied as part
+// of a phase, along with its collision protection settings.
 type ClusterExtensionRevisionObject struct {
+	// object is a required embedded Kubernetes object to be applied.
+	//
+	// This object must be a valid Kubernetes resource with apiVersion, kind, and metadata fields.
+	//
 	// +kubebuilder:validation:EmbeddedResource
 	// +kubebuilder:pruning:PreserveUnknownFields
 	Object unstructured.Unstructured `json:"object"`
-	// CollisionProtection controls whether OLM can adopt and modify objects
-	// already existing on the cluster or even owned by another controller.
+
+	// collisionProtection controls whether the operator can adopt and modify objects
+	// that already exist on the cluster.
+	//
+	// When set to "Prevent" (the default), the operator only manages objects it created itself.
+	// This prevents ownership collisions.
+	//
+	// When set to "IfNoController", the operator can adopt and modify pre-existing objects
+	// that are not owned by another controller.
+	// This is useful for taking over management of manually-created resources.
+	//
+	// When set to "None", the operator can adopt and modify any pre-existing object, even if
+	// owned by another controller.
+	// Use this setting with extreme caution as it may cause multiple controllers to fight over
+	// the same resource, resulting in increased load on the API server and etcd.
 	//
 	// +kubebuilder:default="Prevent"
 	// +kubebuilder:validation:Enum=Prevent;IfNoController;None
@@ -128,6 +186,27 @@ const (
 
 // ClusterExtensionRevisionStatus defines the observed state of a ClusterExtensionRevision.
 type ClusterExtensionRevisionStatus struct {
+	// conditions is an optional list of status conditions describing the state of the
+	// ClusterExtensionRevision.
+	//
+	// The Progressing condition represents whether the revision is actively rolling out:
+	//   - When status is True and reason is Progressing, the revision rollout is actively making progress and is in transition.
+	//   - When Progressing is not present, the revision is not currently in transition.
+	//
+	// The Available condition represents whether the revision has been successfully rolled out and is available:
+	//   - When status is True and reason is Available, the revision has been successfully rolled out and all objects pass their readiness probes.
+	//   - When status is False and reason is Incomplete, the revision rollout has not yet completed but no specific failures have been detected.
+	//   - When status is False and reason is ProbeFailure, one or more objects are failing their readiness probes during rollout.
+	//   - When status is False and reason is ReconcileFailure, the revision has encountered a general reconciliation failure.
+	//   - When status is False and reason is RevisionValidationFailure, the revision failed preflight validation checks.
+	//   - When status is False and reason is PhaseValidationError, a phase within the revision failed preflight validation checks.
+	//   - When status is False and reason is ObjectCollisions, objects in the revision collide with existing cluster objects that cannot be adopted.
+	//   - When status is Unknown and reason is Archived, the revision has been archived and its objects have been torn down.
+	//   - When status is Unknown and reason is Migrated, the revision was migrated from an existing release and object status probe results have not yet been observed.
+	//
+	// The Succeeded condition represents whether the revision has successfully completed its rollout:
+	//   - When status is True and reason is RolloutSuccess, the revision has successfully completed its rollout. This condition is set once and persists even if the revision later becomes unavailable.
+	//
 	// +listType=map
 	// +listMapKey=type
 	// +optional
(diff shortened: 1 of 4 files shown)


camilamacedo86 (approved):
/lgtm

camilamacedo86:
Hi @perdasilva 

Can we fix the emoj for 📖 ( docs )?

camilamacedo86:
@perdasilva I think it is also valid to address the Copilot reviews. WDYT?

perdasilva:
> @perdasilva I think it is also valid to address the Copilot reviews. WDYT?

Done =D

api/v1/clusterextensionrevision_types.go
```diff
@@ -24,42 +24,121 @@ import (
...
+	// ClusterExtensionRevision rollout has succeeded at least once.
 	ClusterExtensionRevisionTypeSucceeded = "Succeeded"
 
-	// Condition Reasons
-	ClusterExtensionRevisionReasonAvailable                 = "Available"
-	ClusterExtensionRevisionReasonReconcileFailure          = "ReconcileFailure"
+	// ClusterExtensionRevisionReasonAvailable is set when the revision is available
+	// and passes all probes after a successful rollout.
+	// Condition: Available, Status: True
+	ClusterExtensionRevisionReasonAvailable = "Available"
```
rashmigottipati: Changes look good overall!

One small suggestion: the Available condition currently uses Reason: "Available", which duplicates the condition type. To make the status more informative, it might help to use a more descriptive reason (for e.g AllObjectsReady maybe?).

The same applies to any other reasons that currently match their condition types.

perdasilva: This is being resolved in a follow-up PR. I just wanted to document what is in there right now in case that PR doesn't make the cut.

perdasilva: Great call out tho!

rashmigottipati: Ah I see. that makes sense. 

Thanks for clarifying :)

tmshort:
/hold
@perdasilva: To allow other people to give it a look while I throw on an approval. Please remove when you think we've had enough eyes on it.

perdasilva:
> /hold @perdasilva: To allow other people to give it a look while I throw on an approval. Please remove when you think we've had enough eyes on it.

good thinking! thank you!!

helm/olmv1/base/operator-controller/crd/experimental/olm.operatorframework.io_clusterextensionrevisions.yaml
```diff
@@ -63,36 +92,84 @@ spec:
...
+                  Once set (even if empty), this field is immutable.
                 items:
                   description: |-
-                    ClusterExtensionRevisionPhase are groups of objects that will be applied at the same time.
-                    All objects in the a phase will have to pass their probes in order to progress to the next phase.
+                    ClusterExtensionRevisionPhase represents a group of objects that are applied together.
+
+                    Objects within a phase are applied simultaneously, and all objects must pass their
+                    readiness probes before the revision progresses to the next phase. Phases are applied
+                    in a well-defined order based on the types of objects they contain.
```
michaelryanpeter: Doesn't this contradict line 116?

michaelryanpeter (commented):
So I've looked it over, and I think it is pretty good overall, but I find the way that the LLM described the fields to be really confusing. For example, "phases is an optional, immutable list of phases..." or "objects is...". I think that something like "the phases property/field specifies an optional, immutaable list of phases. A phase is a ..." would be easier to grok. But I don't think this is a blocker and I know time is tight. Since it is tech preview, we can update these comments right?

api/v1/clusterextensionrevision_types.go
```diff
@@ -24,11 +24,15 @@ import (
 const (
 	ClusterExtensionRevisionKind = "ClusterExtensionRevision"
 
-	// Condition Types
+	// ClusterExtensionRevisionTypeAvailable is the condition type that represents whether the
+	// ClusterExtensionRevision is available and has been successfully rolled out.
 	ClusterExtensionRevisionTypeAvailable = "Available"
+
+	// ClusterExtensionRevisionTypeSucceeded is the condition type that represents whether the
+	// ClusterExtensionRevision rollout has succeeded at least once.
```
pedjak: given that revisions are immutable, it cannot be updated so that  a rollout can happen. Hence,  it can succeed max once.  

```suggestion
// ClusterExtensionRevision rollout has succeeded
```

api/v1/clusterextensionrevision_types.go
```diff
@@ -43,23 +47,54 @@ const (
 )
 
 // ClusterExtensionRevisionSpec defines the desired state of ClusterExtensionRevision.
+//
+// A ClusterExtensionRevision represents a specific immutable snapshot of the objects
+// to be installed for a ClusterExtension. Each revision is rolled out in phases,
+// with objects organized by their Group-Kind into well-known phases such as namespaces,
+// rbac, crds, and deploy.
```
pedjak: This doc refers `ClusterExtensionRevision`, thus it should be placed on `type ClusterExtensionRevision struct`

api/v1/clusterextensionrevision_types.go
```diff
@@ -43,23 +47,54 @@ const (
...
 type ClusterExtensionRevisionSpec struct {
-	// Specifies the lifecycle state of the ClusterExtensionRevision.
+	// The lifecycleState field specifies the lifecycle state of the ClusterExtensionRevision.
```
pedjak: We should follow convention: https://go.dev/blog/godoc :

````
Notice this comment is a complete sentence that begins with the name of the element it describes.
````

```suggestion
//  LifecycleState specify the state of the ClusterExtensionRevision
```

api/v1/clusterextensionrevision_types.go
```diff
@@ -43,23 +47,54 @@ const (
...
+	//
+	// When set to "Active" (the default), the revision is actively managed and reconciled.
+	// When set to "Paused", reconciliation is disabled but status updates continue.
```
pedjak: Do you want to drop it, given that it is not implemented and reintroduce it when needed?

api/v1/clusterextensionrevision_types.go
```diff
@@ -43,23 +47,54 @@ const (
...
+	// When set to "Archived", the revision is torn down and scaled to zero.
```
pedjak: what does it mean "scaled to zero" in this context, given that we can install an arbitrary workload? Does "Archived" means that the revision does not manage resources?

api/v1/clusterextensionrevision_types.go
```diff
@@ -43,23 +47,54 @@ const (
...
+	// Objects are organized into phases based on their Group-Kind. Common phases include:
+	//   - namespaces: Namespace objects
+	//   - policies: ResourceQuota, LimitRange, NetworkPolicy objects
+	//   - rbac: ServiceAccount, Role, RoleBinding, ClusterRole, ClusterRoleBinding objects
+	//   - crds: CustomResourceDefinition objects
+	//   - storage: PersistentVolume, PersistentVolumeClaim, StorageClass objects
+	//   - deploy: Deployment, StatefulSet, DaemonSet, Service, ConfigMap, Secret objects
+	//   - publish: Ingress, APIService, Route, Webhook objects
+	//
+	// All objects in a phase are applied simultaneously.
```
pedjak: this is an implementation detail and not sure if this holds even in the current implementation. Perhaps we should drop this line and replace it with something like:

"Objects in a phase are applied without deterministic order"

api/v1/clusterextensionrevision_types.go
```diff
@@ -75,33 +110,68 @@ const (
...
+	// The revision is removed from the owner list of all other objects previously under management and all objects
+	// that did not transition to a succeeding revision are deleted.
 	ClusterExtensionRevisionLifecycleStateArchived ClusterExtensionRevisionLifecycleState = "Archived"
 )
 
-// ClusterExtensionRevisionPhase are groups of objects that will be applied at the same time.
-// All objects in the a phase will have to pass their probes in order to progress to the next phase.
+// ClusterExtensionRevisionPhase represents a group of objects that are applied together.
+//
+// Objects in a phase are applied simultaneously.
```
pedjak: same comment as above.

api/v1/clusterextensionrevision_types.go
```diff
@@ -75,33 +110,68 @@ const (
...
+// All objects must pass their readiness probes before the revision progresses to the next phase.
+// Phases are applied in a defined order based on the types of objects they contain.
```
pedjak: not sure if this should be here, perhaps we should move it `phases` field of `ClusterExtensionRevisionSpec`?

api/v1/clusterextensionrevision_types.go
```diff
@@ -75,33 +110,68 @@ const (
...
 type ClusterExtensionRevisionPhase struct {
-	// Name identifies this phase.
+	// The name field is a required identifier for this phase.
```
pedjak: We should follow convention https://go.dev/blog/godoc

api/v1/clusterextensionrevision_types.go
```diff
@@ -75,33 +110,68 @@ const (
...
+	// Common phase names include: namespaces, policies, rbac, crds, storage, deploy, publish.
+	//
+	// [RFC 1123]: https://tools.ietf.org/html/rfc1123
 	//
 	// +kubebuilder:validation:MaxLength=63
 	// +kubebuilder:validation:Pattern=`^[a-z]([-a-z0-9]*[a-z0-9])?$`
 	Name string `json:"name"`
-	// Objects are a list of all the objects within this phase.
+
+	// The objects field is a required list of all Kubernetes objects in this phase.
```
pedjak: We should follow convention https://go.dev/blog/godoc

api/v1/clusterextensionrevision_types.go
```diff
@@ -75,33 +110,68 @@ const (
...
+	//
+	// All objects in this list are applied to the cluster simultaneously.
```
pedjak: See earlier comments.

api/v1/clusterextensionrevision_types.go
```diff
@@ -75,33 +110,68 @@ const (
...
+	// The phase is considered complete only after all objects pass their readiness probes.
```
pedjak: not sure if we need this line here, given it talks about phases.

api/v1/clusterextensionrevision_types.go
```diff
@@ -75,33 +110,68 @@ const (
...
 	Objects []ClusterExtensionRevisionObject `json:"objects"`
 }
 
-// ClusterExtensionRevisionObject contains an object and settings for it.
+// ClusterExtensionRevisionObject represents a Kubernetes object to be applied as part
+// of a ClusterExtensionRevision, along with its collision protection settings.
```
pedjak:
```suggestion
of a phase, along with its collision protection settings.
```

api/v1/clusterextensionrevision_types.go
```diff
@@ -75,33 +110,68 @@ const (
...
 type ClusterExtensionRevisionObject struct {
+	// The object field is a required embedded Kubernetes object to be applied.
```
pedjak: We should follow convention https://go.dev/blog/godoc

api/v1/clusterextensionrevision_types.go
```diff
@@ -75,33 +110,68 @@ const (
...
+	//
+	// This object must be a valid Kubernetes resource with apiVersion, kind, and metadata fields.
+	// Status fields are not permitted and are removed if present.
+	// Only specific metadata fields are preserved: name, namespace, labels, and annotations.
```
pedjak: We are not actively removing any fields, but we are not going to apply them if they are present.

api/v1/clusterextensionrevision_types.go
```diff
@@ -75,33 +110,68 @@ const (
...
+	//
 	// +kubebuilder:validation:EmbeddedResource
 	// +kubebuilder:pruning:PreserveUnknownFields
 	Object unstructured.Unstructured `json:"object"`
-	// CollisionProtection controls whether OLM can adopt and modify objects
-	// already existing on the cluster or even owned by another controller.
+
+	// The collisionProtection field controls whether the operator can adopt and modify objects
```
pedjak: We should follow convention https://go.dev/blog/godoc

api/v1/clusterextensionrevision_types.go
```diff
@@ -128,6 +198,27 @@ const (
 
 // ClusterExtensionRevisionStatus defines the observed state of a ClusterExtensionRevision.
 type ClusterExtensionRevisionStatus struct {
+	// conditions is an optional list of status conditions describing the state of the
```
pedjak, line 189:
```suggestion
Conditions is an optional list of status conditions describing the state of the
```

api/v1/clusterextensionrevision_types.go
```diff
@@ -128,6 +198,27 @@ const (
...
+	// ClusterExtensionRevision.
+	//
+	// The Progressing condition represents whether the revision is actively rolling out:
+	//   - When status is True and reason is Progressing, the revision rollout is actively making progress and is in transition.
+	//   - When Progressing is not present, the revision is not currently in transition.
```
pedjak, line 194: On the main we do not have `Progression` condition - let's postpone its documentation once we have it.

api/v1/clusterextensionrevision_types.go
```diff
@@ -128,6 +198,27 @@ const (
...
+	//   - When status is False and reason is ProbeFailure, one or more objects are failing their readiness probes during rollout.
+	//   - When status is False and reason is ReconcileFailure, the revision has encountered a general reconciliation failure.
+	//   - When status is False and reason is RevisionValidationFailure, the revision failed preflight validation checks.
+	//   - When status is False and reason is PhaseValidationError, a phase within the revision failed preflight validation checks.
+	//   - When status is False and reason is ObjectCollisions, objects in the revision collide with existing cluster objects that cannot be adopted.
+	//   - When status is Unknown and reason is Archived, the revision has been archived and its objects have been torn down.
+	//   - When status is Unknown and reason is Migrated, the revision was migrated from an existing release and object status probe results have not yet been observed.
+	//
+	// The Succeeded condition represents whether the revision has successfully completed its rollout at least once:
```
pedjak: same comment as earlier, we should drop `at least once`, IMO.

api/v1/clusterextensionrevision_types.go
```diff
@@ -137,19 +228,38 @@ type ClusterExtensionRevisionStatus struct {
 // +kubebuilder:object:root=true
 // +kubebuilder:resource:scope=Cluster
 // +kubebuilder:subresource:status
-
-// ClusterExtensionRevision is the Schema for the clusterextensionrevisions API
 // +kubebuilder:printcolumn:name="Available",type=string,JSONPath=`.status.conditions[?(@.type=='Available')].status`
 // +kubebuilder:printcolumn:name=Age,type=date,JSONPath=`.metadata.creationTimestamp`
+
+// ClusterExtensionRevision is the schema for the ClusterExtensionRevisions API.
+//
```
pedjak: I would drop this line, because it is redundant. The next line clearly describe the semantic.

api/v1/clusterextensionrevision_types.go
```diff
@@ -137,19 +228,38 @@ type ClusterExtensionRevisionStatus struct {
...
+// A ClusterExtensionRevision represents an immutable snapshot of Kubernetes objects
+// for a specific version of a ClusterExtension. Each revision contains objects
+// organized into phases that roll out sequentially.
+//
+// The ClusterExtension controller creates and manages ClusterExtensionRevisions automatically.
```
pedjak: Implementation detail, not sure if needs to be mentioned.

api/v1/clusterextensionrevision_types.go
```diff
@@ -137,19 +228,38 @@ type ClusterExtensionRevisionStatus struct {
...
+// Do not create them directly. Each ClusterExtension can have multiple revisions as it
+// upgrades or reconfigures.
+//
+// The revision rollout uses a phased approach where objects apply in a defined order based
+// on type (for example: namespaces, then RBAC, then CRDs, then deployments). All objects
+// in a phase apply simultaneously and must pass readiness probes before rollout continues
```
pedjak: see earlier comments on "simultaneously"

api/v1/clusterextensionrevision_types.go
```diff
@@ -137,19 +228,38 @@ type ClusterExtensionRevisionStatus struct {
...
+// to the next phase.
+//
+// Revisions have three lifecycle states:
+//   - Active: The revision is actively managed and reconciled (default)
+//   - Paused: Reconciliation is disabled but status updates continue
+//   - Archived: The revision is torn down, objects are deleted, and the revision removes
+//     itself from the owner list of managed objects
```
pedjak: Not sure if we need to repeat it here, given that we documented on the `lifecycleState` field.

api/v1/clusterextensionrevision_types.go
```diff
@@ -137,19 +228,38 @@ type ClusterExtensionRevisionStatus struct {
...
 type ClusterExtensionRevision struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
-	// spec is an optional field that defines the desired state of the ClusterExtension.
+	// The spec field is optional and defines the desired state of the ClusterExtensionRevision.
```
pedjak: Please follow convention: https://go.dev/blog/godoc

BTW, how spec could be optional, i.e. what is the semantic of a revision without spec?

api/v1/clusterextensionrevision_types.go
```diff
@@ -137,19 +228,38 @@ type ClusterExtensionRevisionStatus struct {
...
 	// +optional
 	Spec ClusterExtensionRevisionSpec `json:"spec,omitempty"`
 
-	// status is an optional field that defines the observed state of the ClusterExtension.
+	// The status field is optional and defines the observed state of the ClusterExtensionRevision.
```
pedjak: Please follow convention: https://go.dev/blog/godoc

api/v1/clusterextensionrevision_types.go
perdasilva: As it is right now [[ref](https://github.com/operator-framework/operator-controller/blob/main/internal/operator-controller/controllers/clusterextensionrevision_controller.go#L254)] we never set it to False and only set it once the revision has completed (all objects crated, at spec, and all progression probes have succeeded. Progression and Availability conditions can flap in the meantime but Succeeded will always be true.

api/v1/clusterextensionrevision_types.go
perdasilva: @michaelryanpeter something for you to keep in mind

api/v1/clusterextensionrevision_types.go
perdasilva: I'm going to double-check with Nico if there was any (technical) point to the spec being optional.

api/v1/clusterextensionrevision_types.go
perdasilva: meh - I'll updated it. It might be that we'll remove this condition eventually. I think we only kept it as part of the interim glue with the ClusterExtension

api/v1/clusterextensionrevision_types.go
perdasilva: I'm having trouble finding the early comment. But, I've gone ahead and removed this for now as it speaks to implementation. Having said that, I think we're still a bit fuzzy on how we want to present this API to the user and how much that phased approach with progression checks are implementation detail vs a first class mental model that the user should have in their heads.

api/v1/clusterextensionrevision_types.go
perdasilva: Removed

api/v1/clusterextensionrevision_types.go
perdasilva: The reason its optional is to facilitate the sharding story because we'll need to set the owner ref on the shards to ensure things are cleaned up. And for that, the revision will need to exist first. In principle we could also create the shards then create the revision object, which then takes over ownership of the shards, but I guess this leaves open the possibility of dangling resources if the revision gets deleted before that happens.

For now, I've updated the docs (removing the optional description since rn it doesn't add much) but left the // +optional tag so that we don't break the API later on if we decide to go with the create revision first approach described above.

api/v1/clusterextensionrevision_types.go
perdasilva: I've cleaned up the language here. Great call out.

api/v1/clusterextensionrevision_types.go
```diff
@@ -44,22 +48,47 @@ const (
 
 // ClusterExtensionRevisionSpec defines the desired state of ClusterExtensionRevision.
 type ClusterExtensionRevisionSpec struct {
-	// Specifies the lifecycle state of the ClusterExtensionRevision.
+	// lifecycleState specifies the lifecycle state of the ClusterExtensionRevision.
```
pedjak, line 51: It should `LifecycleState` not `lifecycleState`, please update across all changes.

api/v1/clusterextensionrevision_types.go
perdasilva: We currently do [e.g.](https://github.com/operator-framework/operator-controller/blob/main/internal/operator-controller/controllers/clusterextensionrevision_controller.go#L305). We just have heaps of reasons under it.

api/v1/clusterextensionrevision_types.go
perdasilva: I'm keeping the format consistent with the other APIs. I'll bring this up with the team and address it at once over all the APIs in a follow-up.

api/v1/clusterextensionrevision_types.go
pedjak: Ah, sorry my bad. Thanks.

api/v1/clusterextensionrevision_types.go
perdasilva: I've changed it to "in not particular order". Let's leave it at that for now. I think there's a wider discussion here around how much of the implementation should be documented and part of the mental model. Especially because we are describing a deployment strategy - which can be understood as a specific implementation.

api/v1/clusterextensionrevision_types.go
perdasilva: I've removed it to keep the top-level doc succinct and just left it in `phases`

api/v1/clusterextensionrevision_types.go
perdasilva: I moved it up to the `ClusterExtensionRevisionPhase`.

api/v1/clusterextensionrevision_types.go
perdasilva: I'm going to remove these two lines. We currently sanitize the bundle manifests at the applier level [ref](https://github.com/operator-framework/operator-controller/blob/main/internal/operator-controller/applier/boxcutter.go#L69). I think boxcutter just creates them as is...? ([e.g.](https://github.com/package-operator/boxcutter/blob/main/machinery/objects.go#L235))

api/v1/clusterextensionrevision_types.go
```diff
@@ -75,33 +104,68 @@ const (
 	// ClusterExtensionRevisionLifecycleStateActive / "Active" is the default lifecycle state.
 	ClusterExtensionRevisionLifecycleStateActive ClusterExtensionRevisionLifecycleState = "Active"
 	// ClusterExtensionRevisionLifecycleStatePaused / "Paused" disables reconciliation of the ClusterExtensionRevision.
-	// Only Status updates will still propagated, but object changes will not be reconciled.
+	// Object changes will not be reconciled. However, status updates will be propagated.
 	ClusterExtensionRevisionLifecycleStatePaused ClusterExtensionRevisionLifecycleState = "Paused"
```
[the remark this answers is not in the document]
perdasilva: I'll remove the code references for Paused in a follow up PR.

PR #2353, opened 2025-11-20
🌱 (fix): Experimental E2E Upgrade does breaks up we do a release when breaking changes on experimental features are merged

camilamacedo86:
This PR updates the experimental upgrade path to avoid failures caused by expected breaking changes in experimental features by moving experimental upgrade tests to main → PR to prevent experimental breakages from failing and forcing us to do a release. 

### Current issue
The current flow is: latest release → PR, with all experimental features enabled.  
If an experimental feature breaks, the upgrade job fails, blocking merges until the next release.

### Proposed change
Switch the experimental upgrade flow to: main → PR.  
This keeps the stable upgrade path clean while still validating experimental changes against `main`.
**So, we can safe merge a breaking change for experimental and then the follow PRs will pass and we still raising those scenarios to be aware of.**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -305,6 +305,10 @@ test-extension-developer-e2e: run-internal image-registry extension-developer-e2
 run-latest-release:
 	curl -L -s https://github.com/operator-framework/operator-controller/releases/latest/download/$(notdir $(RELEASE_INSTALL)) | bash -s
 
+.PHONY: run-main-install
+run-main-install:
+	KIND_CLUSTER_NAME=$(KIND_CLUSTER_NAME) ./hack/test/install-from-main.sh $(notdir $(SOURCE_MANIFEST))
+
 .PHONY: pre-upgrade-setup
 pre-upgrade-setup:
 	./hack/test/pre-upgrade-setup.sh $(CATALOG_IMG) $(TEST_CLUSTER_CATALOG_NAME) $(TEST_CLUSTER_EXTENSION_NAME)
@@ -315,6 +319,7 @@ post-upgrade-checks:
 
 
 TEST_UPGRADE_E2E_TASKS := kind-cluster run-latest-release image-registry pre-upgrade-setup docker-build kind-load kind-deploy post-upgrade-checks kind-clean
+TEST_UPGRADE_EXPERIMENTAL_E2E_TASKS := kind-cluster run-main-install image-registry pre-upgrade-setup docker-build kind-load kind-deploy post-upgrade-checks kind-clean
 
 .PHONY: test-upgrade-e2e
 test-upgrade-e2e: SOURCE_MANIFEST := $(STANDARD_MANIFEST)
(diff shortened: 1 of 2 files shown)


hack/test/install-from-main.sh
```diff
@@ -0,0 +1,68 @@
...
+echo "Building images from main branch source code..."
+make docker-build
+
+echo "Loading images into kind cluster ${KIND_CLUSTER_NAME}..."
+make kind-load KIND_CLUSTER_NAME="${KIND_CLUSTER_NAME}"
+
+echo "Deploying manifests from main branch..."
+# Extract CERT_MGR_VERSION from main branch Makefile
+# Expected Makefile line format: export CERT_MGR_VERSION := <value>
+CERT_MGR_VERSION_FROM_MAIN=$(grep "^export CERT_MGR_VERSION" Makefile | awk -F'[ :=]+' '{print $NF}')
```
perdasilva, line 58: if we're anyway in the checked out main branch, why not use some of the make targets like `make kind-deploy`?

camilamacedo86: We need to run the install script (same generated when we do a release)
That’s different from running make kind-deploy.

camilamacedo86:
> My only concern is that we're re-determining the cert manager version. Maybe we could move some (or all?) of the scripts logic into the Makefile in the future.

The CertManager is defined in the Makefile
The script is just getting the value used in the Makefile
We are not redefining it here.

Makefile
```diff
@@ -305,6 +305,10 @@ test-extension-developer-e2e: run-internal image-registry extension-developer-e2
 run-latest-release:
 	curl -L -s https://github.com/operator-framework/operator-controller/releases/latest/download/$(notdir $(RELEASE_INSTALL)) | bash -s
 
+.PHONY: run-main-install
```
tmshort, line 308: This duplicates the `run-internal` target, but with a script.

tmshort:
I'm actually going to disagree with this approach.

It’s not just main->experimental, there’s two axis here: standard or experimental, and prior-release or main (not-yet-released). With those we can see that we have 4 possibilities… my initial look at the Makefile indicates that for test-upgrade-experimental-e2e, it’s always using the experimental manifest, so it’s doing an experimental upgrade. We seem to be missing a number of interesting possibilities here (note “main” could represent the PR as well).

`test-upgrade-e2e`: standard[prior-release] -> standard[main/PR] (REQUIRED)
`test-upgrade-experimental-e2e`: experimental[prior-release] -> experimental[main/PR] (REQUIRED -> INFORMATIVE?)

What we don’t have is the situation where we cross standard and experimental:

standard[main] -> experimental[mainPR] (REQUIRED?)
standard[prior] -> experimental[main/PR] (INFORMATIVE?)

This PR just modifies the existing `test-upgrade-experimental-e2e`, and I think that's the wrong thing to do. I think that one should be made informative, and new tests added for these standard-> experimental cases.

See https://github.com/operator-framework/operator-controller/pull/2360 as an alternative

hack/test/install-from-main.sh
tmshort: Exactly, this is duplicating a lot of existing Makefile behavior.

tmshort:
Closing in favor of #2360

PR #2358, opened 2025-11-21
🌱  new tag symmetry and required validations

grokspawn:
# Description

Adds new standard:description tag support to allow inclusion of standard-manifest-only field descriptions. 
Adds new optional/required validation tag support to allow field tags to impact the required list of their parent field.
Adds unit tests for all new capabilities, since there are no existing uses of the new tags (yet!).

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/hack/tools/crd-generator/main.go
+++ b/hack/tools/crd-generator/main.go
@@ -137,7 +137,7 @@ func runGenerator(args ...string) {
 				if channel == StandardChannel && strings.Contains(version.Name, "alpha") {
 					channelCrd.Spec.Versions[i].Served = false
 				}
-				version.Schema.OpenAPIV3Schema.Properties, version.Schema.OpenAPIV3Schema.Required = opconTweaksMap(channel, version.Schema.OpenAPIV3Schema.Properties, version.Schema.OpenAPIV3Schema.Required)
+				channelCrd.Spec.Versions[i].Schema.OpenAPIV3Schema.Properties = opconTweaksMap(channel, channelCrd.Spec.Versions[i].Schema.OpenAPIV3Schema)
 			}
 
 			conv, err := crd.AsVersion(*channelCrd, apiextensionsv1.SchemeGroupVersion)
@@ -180,10 +180,11 @@ func runGenerator(args ...string) {
 	}
 }
 
-// Apply Opcon specific tweaks to all properties in a map, and return a list of required fields according to opcon tags.
-// For opcon validation optional/required tags, the required list is updated accordingly.
-func opconTweaksMap(channel string, props map[string]apiextensionsv1.JSONSchemaProps, existingRequired []string) (map[string]apiextensionsv1.JSONSchemaProps, []string) {
-	newRequired := slices.Clone(existingRequired)
+// Apply Opcon specific tweaks to all properties in a map, and update the parent schema's required list according to opcon tags.
+// For opcon validation optional/required tags, the parent schema's required list is mutated directly.
+// TODO: if we need to support other conditions from opconTweaks, it will likely be preferable to convey the parent schema to facilitate direct alteration.
+func opconTweaksMap(channel string, parentSchema *apiextensionsv1.JSONSchemaProps) map[string]apiextensionsv1.JSONSchemaProps {
+	props := parentSchema.Properties
 
 	for name := range props {
 		jsonProps := props[name]
@@ -195,23 +196,18 @@ func opconTweaksMap(channel string, props map[string]apiextensionsv1.JSONSchemaP
 			// Update required list based on tag
 			switch reqStatus {
 			case statusRequired:
-				if !slices.Contains(newRequired, name) {
-					newRequired = append(newRequired, name)
+				if !slices.Contains(parentSchema.Required, name) {
+					parentSchema.Required = append(parentSchema.Required, name)
 				}
 			case statusOptional:
-				newRequired = slices.DeleteFunc(newRequired, func(s string) bool { return s == name })
+				parentSchema.Required = slices.DeleteFunc(parentSchema.Required, func(s string) bool { return s == name })
 			default:
 				// "" (unspecified) means keep existing status
 			}
 		}
 	}
-	return props, newRequired
+	return props
 }
-
-// Custom Opcon API Tweaks for tags prefixed with `<opcon:` that get past
-// the limitations of Kubebuilder annotations.
-// Returns the modified schema and a string indicating required status where indicated by opcon tags:
-// "required", "optional", or "" (no decision -- preserve any non-opcon required status).
 
 const (
 	statusRequired  = "required"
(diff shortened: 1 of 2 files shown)


hack/tools/crd-generator/main.go
```diff
@@ -179,25 +180,51 @@ func runGenerator(args ...string) {
 	}
 }
 
-func opconTweaksMap(channel string, props map[string]apiextensionsv1.JSONSchemaProps) map[string]apiextensionsv1.JSONSchemaProps {
+// Apply Opcon specific tweaks to all properties in a map, and return a list of required fields according to opcon tags.
+// For opcon validation optional/required tags, the required list is updated accordingly.
+func opconTweaksMap(channel string, props map[string]apiextensionsv1.JSONSchemaProps, existingRequired []string) (map[string]apiextensionsv1.JSONSchemaProps, []string) {
```
joelanford: It would likely be more future proof to pass a pointer of the parent into this function and apply mutations directly. The current style in the PR would mean that any other tweaks that live at the parent level would mean more fields would have to be passed back and forth. And that _could_ get pretty cumbersome.

hack/tools/crd-generator/main.go
```diff
@@ -246,34 +288,42 @@ func opconTweaks(channel string, name string, jsonProps apiextensionsv1.JSONSche
...
-		re := regexp.MustCompile(regexPattern)
-		match := re.FindStringSubmatch(description)
-		if len(match) != 2 {
-			log.Fatalf("Invalid <opcon:experimental:description> tag for %s", name)
+	tagset := []struct {
+		channel string
+		start   string
+		end     string
+	}{
+		{channel: ExperimentalChannel, start: "<opcon:standard:description>", end: "</opcon:standard:description>"},
```
joelanford: Maybe simplify even more like this:
```suggestion
		{channel: ExperimentalChannel, tag: "opcon:standard:description"},
```
And then have code that constructs the start/end tags so that we avoid duplicating the tag text itself.

hack/tools/crd-generator/main.go
grokspawn: Conceded, and I deliberately wasn't opening the gate too wide at this juncture, as its pretty simple to iterate to that if we _have to_, but my preference is to limit scope on what is an unexported, internal API.
However, that kinda represents saved state (as in "right now we know we limited things to avoid just passing a mutable parent... but we have to remember").  
That could get problematic in future as it is then incumbent on future reviewers to hold the line on parameter growth.  
Is there a way to make the desires more clear here?

joelanford: I'm guessing that if we would have designed this from the start to handle optional vs. required we would have ended up with a field manipulator that has access to the parent object for this exact reason. It turns out that applying mutations about a single field requires touching multiple places in the parent.

So IMO, a mutable parent pointer is probably the correct design anyway?

grokspawn: Agreed.  Can't fault me for trying the smaller gate first. 😄

pedjak:
> Adds unit tests for all new capabilities, since there are no existing uses of the new tags (yet!).

What is the motivation for adding this logic, if there are not need for it? Should we postpone it and add within a PR where such logic is needed?

joelanford:
@pedjak motivation is making the `clusterextension.spec.serviceAccount` field optional in the experimental channel. To me, it seems like a reasonable thing to split off from that.

hack/tools/crd-generator/main.go
```diff
@@ -179,25 +180,51 @@ func runGenerator(args ...string) {
...
+			switch reqStatus {
+			case statusRequired:
+				if !slices.Contains(parentSchema.Required, name) {
+					parentSchema.Required = append(parentSchema.Required, name)
+				}
+			case statusOptional:
+				parentSchema.Required = slices.DeleteFunc(parentSchema.Required, func(s string) bool { return s == name })
+			default:
+				// "" (unspecified) means keep existing status
+			}
```
joelanford, line 206: I was originally thinking that we'd pass `parentSchema` all the way down into `opconTweaks` (which I ultimately still think is the correct architectural decision).

However, I see that this function already had some pre-existing code for handling `p == nil` (or not) and making changes in the parent at this level.

With that in mind, I think your approach makes sense to align with the current architecture.

Maybe leave a comment that it might be worth refactoring in the future to allow `opconTweaks` to make any/all necessary changes to the parent to account for the markers that are on the `name`-ed field?

pedjak:
> @pedjak motivation is making the `clusterextension.spec.serviceAccount` field optional in the experimental channel. To me, it seems like a reasonable thing to split off from that.

Thanks @joelanford for the context. I see now that is extracted from https://github.com/operator-framework/operator-controller/pull/2355

hack/tools/crd-generator/main.go
```diff
@@ -179,25 +180,51 @@ func runGenerator(args ...string) {
...
 		}
 	}
 	return props
 }
 
 // Custom Opcon API Tweaks for tags prefixed with `<opcon:` that get past
 // the limitations of Kubebuilder annotations.
-func opconTweaks(channel string, name string, jsonProps apiextensionsv1.JSONSchemaProps) *apiextensionsv1.JSONSchemaProps {
+// Returns the modified schema and a string indicating required status where indicated by opcon tags:
+// "required", "optional", or "" (no decision -- preserve any non-opcon required status).
```
pedjak: instead of strings, would it make sense to use a `*bool` that can represent these three states as well?

pedjak: this is the documentatino for `opconTweaks` function? Should we move it next to the function signature?

tmshort: `bool` has a limited range of values (even if you include `nil` with `*bool`). Using `string` allows for additional flexibility.

hack/tools/crd-generator/main.go
```diff
@@ -237,6 +264,21 @@ func opconTweaks(channel string, name string, jsonProps apiextensionsv1.JSONSche
...
+			}
+
+			numValid++
+			switch optReqMatch[1] {
+			case "Optional":
+				requiredStatus = statusOptional
+			case "Required":
+				requiredStatus = statusRequired
+			}
+		}
```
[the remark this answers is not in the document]
tmshort: Yeah, we should error if we try to set this a second time.

hack/tools/crd-generator/main.go
grokspawn: I'm generally aligning with [CRD guidance](https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md#do-not-use-boolean-fields) that bools end up becoming enums/strings/complex types, and just trying to plumb through a large enough opening to allow expansion, if other decisions need to be added.

grokspawn: I thought to put it before the types which were relevant to that context, but I've now moved it immediately before the function definition.

hack/tools/crd-generator/main.go
grokspawn: Done.

PR #2359, opened 2025-11-24
:book: Updating docs strings for API go definitions

michaelryanpeter:
# Description
Helping @perdasilva by updating the docs strings for the API definitions.

:crossed_fingers: I did this right :)
## Reviewer Checklist

- [ ] API Go Documentation


tmshort:
The CRD diff is only the text.

api/v1/clustercatalog_types.go
```diff
@@ -51,7 +51,7 @@ const (
 //+kubebuilder:printcolumn:name="Serving",type=string,JSONPath=`.status.conditions[?(@.type=="Serving")].status`
 //+kubebuilder:printcolumn:name=Age,type=date,JSONPath=`.metadata.creationTimestamp`
 
-// ClusterCatalog enables users to make File-Based Catalog (FBC) catalog data available to the cluster.
+// The ClusterCatalog resource makes File-Based Catalog (FBC) data available to your cluster.
```
perdasilva: We should follow the godoc format where the comment starts with name of the property/field as it was =/
I learned this recently in the ClusterExtensionRevision API doc

api/v1/clustercatalog_types.go
```diff
@@ -60,16 +60,15 @@ type ClusterCatalog struct {
 	// More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#metadata
 	metav1.ObjectMeta `json:"metadata"`
 
-	// spec is the desired state of the ClusterCatalog.
-	// spec is required.
-	// The controller will work to ensure that the desired
-	// catalog is unpacked and served over the catalog content HTTP server.
+	// The spec field defines the desired state of the ClusterCatalog.
+	// The spec field is required.
+	// The controller ensures that the catalog is unpacked and served over the catalog content HTTP server.
```
perdasilva: Since these docs will be converted into docs for the users with kubectl explain, we use the user face facing field name, i.e. lowercase (i.e. instead of // Spec we do // spec) 

```suggestion
	// spec is a required field that defines the desired state of the ClusterCatalog.
	// The controller ensures that the catalog is unpacked and served over the catalog content HTTP server.
```

perdasilva (changes requested):
It seems we should follow the godoc standard with the yaml field names (lowercase) by convention. I've left a couple of illustrative suggestions. If we make those slight adjustments, everything else should be ok, I think.

api/v1/clustercatalog_types.go
pedjak: Actually https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md#use-json-field-names-in-godoc recommends lowercase variant.

tmshort:
This needs to be rebased, and that will also fix the e2e issues.

michaelryanpeter:
Heard. I am going to update this PR and refactor my downstream slash command for godocs formatting. Hopefully, I can get this completed before shutdown. This was a one-off bit of untracked work for me, and a handful of other things popped up that are more urgent.

PR #2360, opened 2025-11-24
:sparkles: OPRUN-4288: Add upgrade e2es for standard to experimental

tmshort:
The current set of upgrade-e2e tests upgrade from the last release to the current PR/repo state. Standard is tested against standard, experimental is tested against experimental.

We should have upgrade e2e tests that upgrade from standard to experimental, including from the current PR/repo state, and also from the prior release.

This also renames the targets to make it clear what is being upgraded.

# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -9,105 +9,67 @@ on:
       - main
 
 jobs:
-  extension-developer-e2e:
+  e2e:
     runs-on: ubuntu-latest
-    steps:
-    - uses: actions/checkout@v6
-
-    - uses: actions/setup-go@v6
-      with:
-        go-version-file: go.mod
-
-    - name: Run the extension developer e2e test
-      run: make test-extension-developer-e2e
-
-  e2e-kind:
-    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        test:
+          - name: extension-developer-e2e
+            make-target: test-extension-developer-e2e
+            use-artifacts: false
+            use-codecov: false
+          - name: e2e
+            make-target: test-e2e
+            use-artifacts: true
+            use-codecov: true
+          - name: experimental-e2e
+            make-target: test-experimental-e2e
+            use-artifacts: true
+            use-codecov: true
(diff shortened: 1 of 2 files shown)


tmshort:
/override upgrade-e2e
The upgrade-e2e has been renamed.

tmshort:
upgrade-ex2ex-e2e will continue to fail until we issue a release.

tmshort:
When this merges, the `upgrade-e2e` will no longer be required, but we will want to make the other upgrade tests possibly required.

.github/workflows/e2e.yaml
```diff
@@ -103,11 +103,47 @@ jobs:
...
+    - uses: actions/checkout@v6
+
+    - uses: actions/setup-go@v6
+      with:
+        go-version-file: go.mod
+
+    - name: Run the upgrade e2e test
+      run: ARTIFACT_PATH=/tmp/artifacts make test-st2ex-e2e
+
+    - uses: actions/upload-artifact@v5
```
pedjak: Given that we are adding two more test jobs in this PR (in total now 4), would it make sense to declare a [GitHub matrix job](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/run-job-variations#about-matrix-strategies) to execute them in parallel?

tmshort: They are similar enough...

pedjak: > They are similar enough...

Right, hence I though we could pack them in matrix - they differ just by the makefile target it gets invoked.

tmshort: It's a bit more than that, but it's been updated to be a matrix

tmshort:
Due to the renaming of some tests, the `e2e-kind` and `upgrade-e2e` are no longer found, but they do pass (as `e2e/e2e` and `e2e/upgrade-st2st-e2e`)

pedjak:
[nit] Given that this PR is about test infra work, and no user-facing changes, does in qualify to be marked as :sparkles: ?

PR #2362, opened 2025-11-26
:bug: add PDB to make sure at least 1 pod is always available during upgrade

jianzhangbjz:
# Description

To address [OCPBUGS-62517](https://issues.redhat.com/browse/OCPBUGS-62517). Currently, the operator-controller lacks `PodDisruptionBudget` configuration. During node drain operations or cluster upgrades, all controller pods can be evicted
Simultaneously, causing the operator to report `Available=False`, which violates the OpenShift API contract:

  > "A component must not report Available=False during the course of a normal upgrade."
  > — [OpenShift API Contract](https://github.com/openshift/api/blob/master/config/v1/types_cluster_operator.go#L156)

  Add PodDisruptionBudget resources with `minAvailable: 1` for both controllers to ensure at least one pod remains available during:
  - Rolling updates
  - Node drain operations
  - Cluster upgrades

## Reviewer Checklist

- [ ] API Go Documentation

Assisted-by: Claude code

### DIFF
--- a/helm/olmv1/templates/poddisruptionbudget-olmv1-system-catalogd.yml
+++ b/helm/olmv1/templates/poddisruptionbudget-olmv1-system-catalogd.yml
@@ -1,0 +1,22 @@
+{{- if and .Values.options.catalogd.enabled .Values.options.catalogd.podDisruptionBudget.enabled }}
+apiVersion: policy/v1
+kind: PodDisruptionBudget
+metadata:
+  name: catalogd-controller-manager
+  namespace: {{ .Values.namespaces.olmv1.name }}
+  labels:
+    app.kubernetes.io/name: catalogd
+    {{- include "olmv1.labels" . | nindent 4 }}
+  annotations:
+    {{- include "olmv1.annotations" . | nindent 4 }}
+spec:
+  {{- if .Values.options.catalogd.podDisruptionBudget.minAvailable }}
+  minAvailable: {{ .Values.options.catalogd.podDisruptionBudget.minAvailable }}
+  {{- end }}
+  {{- if .Values.options.catalogd.podDisruptionBudget.maxUnavailable }}
+  maxUnavailable: {{ .Values.options.catalogd.podDisruptionBudget.maxUnavailable }}
+  {{- end }}
+  selector:
+    matchLabels:
+      control-plane: catalogd-controller-manager
+{{- end }}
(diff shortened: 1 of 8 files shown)


helm/olmv1/values.yaml
```diff
@@ -12,6 +12,9 @@ options:
     features:
       enabled: []
       disabled: []
+    podDisruptionBudget:
+      enabled: true
```
pedjak, line 16: [nit]: perhaps would be better to name it `create:`?

helm/olmv1/templates/poddisruptionbudget-olmv1-system-catalogd.yml
```diff
@@ -0,0 +1,24 @@
+{{- if .Values.options.catalogd.enabled }}
+{{- if .Values.options.catalogd.podDisruptionBudget.enabled }}
```
pedjak: [nit]: perhaps we can combine these two into single `if`

helm/olmv1/templates/poddisruptionbudget-olmv1-system-operator-controller.yml
```diff
@@ -0,0 +1,24 @@
+{{- if .Values.options.operatorController.enabled }}
```
pedjak: same comments as for the other pdb.

tmshort:
/approve
It's needed, just may need a few tweeks.

helm/olmv1/values.yaml
tmshort: However, all the other options are named `enabled`

tmshort: Actually, is there a reason for this option, and not just create it based on `catalog` or `operatorController` state? (rhetorical question)

helm/olmv1/templates/poddisruptionbudget-olmv1-system-catalogd.yml
jianzhangbjz: Yes, updated it.

helm/olmv1/templates/poddisruptionbudget-olmv1-system-operator-controller.yml
jianzhangbjz: Updated it.

helm/olmv1/values.yaml
jianzhangbjz: I think it gives users explicit control over a resource and follows the existing pattern in the codebase where features are individually toggleable.

jianzhangbjz:
Updated the `hack/test/install-prometheus.sh` to add `timeout` to address the below error:
```console
Waiting for Prometheus Operator pod to become ready...
error: no matching resources found
Cleaning up /tmp/tmp.Gv71SsiUrG
make: *** [Makefile:295: prometheus] Error 1
Error: Process completed with exit code 2.
```

PR #2365, opened 2025-11-28
:seedling: OPRUN-4261 Migrate e2e tests to Godog BDD framework

pedjak:
# Description

Replace traditional Go e2e tests with Godog (Cucumber for Go) to improve test readability and maintainability through behavior-driven development.

Benefits:

- **Living Documentation**: Test scenarios serve as up-to-date documentation of system behavior
- **Better Collaboration**: Product owners can read and validate test scenarios
- **Reduced Duplication**: Reusable step definitions eliminate code repetition
- **Improved Maintainability**: Changes to common patterns happen in one place
- **Clearer Intent**: Gherkin syntax makes test purpose immediately obvious
- **Easier Debugging**: Clear separation between what is tested (features) and how (steps)
- **Concurrent Execution**: Set the ground work for running tests in parallel

Changes:
- Convert existing test scenarios to Gherkin feature files
- Implement reusable step definitions in steps/steps.go
- Add scenario hooks for setup/teardown and feature gate detection
- Provide comprehensive documentation in test/e2e/README.md
- Remove legacy test files (cluster_extension_install_test.go, etc.)
- Added detailed [README](test/e2e/README.md) covering:
   - Architecture and design patterns
   - How to write new tests
   - Running tests with various options
   - Best practices and troubleshooting
- Go test driving code reduced by ~1000 lines

## Migration Notes

   - No changes to test behavior or coverage - this is purely a refactoring
   - All existing test scenarios are preserved with equivalent Gherkin implementations
   - Test execution remains the same via `make test-e2e`

Assisted-By: Claude <noreply@anthropic.com>

## Reviewer Checklist

- [ ] API Go Documentation


test/e2e/features/install.feature
```diff
@@ -0,0 +1,272 @@
...
+
+  As an OLM user I would like to install a cluster extension from catalog
+  or get an appropriate information in case of an error.
+
+  Background:
+    Given OLM is available
+    And "test" catalog serves bundles
+    And Service account "olm-sa" with needed permissions is available in test namespace
+  
+  Scenario Outline: Install latest available version from the default channel
```
joelanford: Default channel is not a concept in OLMv1. This scenario would install highest version in any channel.

test/e2e/features/recover.feature
```diff
@@ -0,0 +1,117 @@
...
+          metadata:
+            labels:
+              app: test-operator
+          spec:
+            containers:
+              - command:
+                - "sleep"
+                args:
+                - "1000"
+                image: busybox:1.36
```
joelanford, line 81: Is this the same image we've always used?

I recall there being issues in the past with rate limiting from Docker Hub.

test/e2e/features/update.feature
```diff
@@ -0,0 +1,247 @@
...
+          catalog:
+            packageName: test
+            selector:
+              matchLabels:
+                "olm.operatorframework.io/metadata.name": test-catalog
+            version: 1.2.0
+      """
+    Then ClusterExtension reports Progressing as True with Reason Retrying:
+      """
+      error upgrading from currently installed version "1.0.0": no bundles found for package "test" matching version "1.2.0"
```
joelanford, line 81: Unrelated to this PR, but something we need to fix separately. This message makes it sound like 1.2.0 just doesn't exist. But it does. It just isn't a successor of the currently installed version.

test/e2e/features_test.go
```diff
@@ -0,0 +1,81 @@
...
+	"log"
+	"os"
+	"testing"
+
+	"github.com/cucumber/godog"
+	"github.com/cucumber/godog/colors"
+	"github.com/spf13/pflag"
+	ctrl "sigs.k8s.io/controller-runtime"
+	//ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
```
joelanford: Avoid new zap dependency? I think we're using klog in our main.go's. Can we use that here too?

test/e2e/steps/hooks.go
```diff
@@ -0,0 +1,122 @@
...
+	return context.WithValue(ctx, scenarioContextKey, scCtx), nil
+}
+
+func scenarioCtx(ctx context.Context) *scenarioContext {
+	return ctx.Value(scenarioContextKey).(*scenarioContext)
+}
+
+func ScenarioCleanup(ctx context.Context, _ *godog.Scenario, err error) (context.Context, error) {
+	sc := scenarioCtx(ctx)
+	for _, p := range sc.backGroundCmds {
```
joelanford: Kill and wait processes concurrently? Or does order matter?

test/e2e/steps/hooks.go
```diff
@@ -0,0 +1,122 @@
...
+		return ctx, err
+	}
+
+	forDeletion := []resource{}
+	if sc.clusterExtensionName != "" {
+		forDeletion = append(forDeletion, resource{name: sc.clusterExtensionName, kind: "clusterextension"})
+	}
+	forDeletion = append(forDeletion, sc.addedResources...)
+	forDeletion = append(forDeletion, resource{name: sc.namespace, kind: "namespace"})
+	for _, r := range forDeletion {
```
joelanford: Same here: Can we delete objects concurrently?

test/e2e/features/install.feature
```diff
@@ -0,0 +1,272 @@
...
+            selector:
+              matchLabels:
+                "olm.operatorframework.io/metadata.name": test-catalog
+      """
+    And ClusterExtension reports Progressing as True with Reason Retrying:
+      """
+      error for resolved bundle "single-namespace-operator.1.0.0" with version "1.0.0":
+      invalid ClusterExtension configuration: invalid configuration: required field "watchNamespace" is missing
+      """
+    When ClusterExtension is updated
```
joelanford: I wonder if we could strategic merge patch instead and have a smaller yaml to make it more clear what's changing?

joelanford (commented):
Looks like a really nice improvement overall. Just some minor comments/questions.

test/e2e/features/install.feature
```diff
@@ -0,0 +1,272 @@
...
+  Scenario: watchNamespace config is required for extension supporting single namespace
+    Given Service account "olm-admin" in test namespace is cluster admin
+    And resource is applied
+      """
+      apiVersion: v1
+      kind: Namespace
+      metadata:
+        name: single-namespace-operator-target
+      """
+    And ClusterExtension is applied
```
tmshort, line 102: Just nothing the inconsistency here:

* resource is applied
vs
* ClusterExtension is applied

The first one is generic, the second one is indicating a very specific resource. What's the reasoning behind this?

test/e2e/features/install.feature
tmshort: That would be an improvement.

test/e2e/features_test.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+func TestMain(m *testing.M) {
+	// parse CLI arguments
+	pflag.Parse()
+	opts.Paths = pflag.Args()
+
+	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&logOpts)))
+
+	//opts.DefaultContext = ctrl.LoggerInto(context.Background(), ctrllog.Log)
```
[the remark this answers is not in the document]
tmshort: Not if it's new code...

test/e2e/features_test.go
```diff
@@ -0,0 +1,79 @@
...
+				// Fail the run if alerts are found
+				fmt.Printf("%v", err)
+				os.Exit(1)
+			}
+		}
+		return
+
+	//	1 - failed
+	//	2 - command line usage error
+	// 128 - or higher, os signal related error exit codes
```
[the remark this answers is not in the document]
tmshort: Why are `case`s for these values not implemented?

test/e2e/steps/hooks.go
```diff
@@ -0,0 +1,122 @@
...
+		if _, err := kubectl("delete", r.kind, r.name, "-n", sc.namespace); err != nil {
```
tmshort: There's no distinction here between namespace-scoped resources (e.g. SAs), and cluster-scoped resources (e.g. CE). 'kubectl' may not complain about the `-n` argument for cluster-scoped resources, but it's basically wrong.

test/e2e/steps/hooks.go
```diff
@@ -0,0 +1,117 @@
...
+
+func scenarioCtx(ctx context.Context) *scenarioContext {
+	return ctx.Value(scenarioContextKey).(*scenarioContext)
+}
+
+func ScenarioCleanup(ctx context.Context, _ *godog.Scenario, err error) (context.Context, error) {
+	sc := scenarioCtx(ctx)
+	for _, p := range sc.backGroundCmds {
+		p.Process.Kill() // nolint: errcheck // we don't care about the error here, we just want to kill the process
+		p.Process.Wait() // nolint: errcheck // same as above, we just want to wait for the process to exit, and do not want to fail the test if it does not
```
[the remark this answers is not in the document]
tmshort: Will `errcheck` still complain even if `_ =  p.Process...()` is used?

test/e2e/network_policy_test.go
tmshort, line 1: This test was not rewritten for Cucumber, are there plans?

test/e2e/steps/steps.go
```diff
@@ -0,0 +1,492 @@
...
+	olmNamespace      = "olmv1-system"
+	olmDeploymentName = "operator-controller-controller-manager"
+	timeout           = 300 * time.Second
+	tick              = 1 * time.Second
+)
+
+var kubeconfigPath string
+
+func init() {
+	kubeconfigPath = os.Getenv("HOME") + "/.kube/config"
```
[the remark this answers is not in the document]
tmshort: This. The KUBECONFIG variable should be honored. Although admittedly, the tests depend on `kind` updating the default context in the kube config.

test/e2e/steps/steps.go
```diff
@@ -0,0 +1,531 @@
...
+}
+
+var (
+	logger = log.Log
+)
+
+func kubectl(args ...string) (string, error) {
+	cmd := exec.Command("kubectl", args...)
+	logger.V(1).Info(strings.Join(cmd.Args, " "))
```
tmshort: A bit more context here, e.g. "Running command" would be helpful. Right now, this just dumps a command line as a log line.

test/e2e/steps/steps.go
```diff
@@ -0,0 +1,531 @@
...
+	cmd.Env = append(os.Environ(), fmt.Sprintf("KUBECONFIG=%s", kubeconfigPath))
```
tmshort, line 110: Here, KUBECONFIG is set? But uses the default kube config file, rather than one set by KUBECONFIG? Same in other places.

test/e2e/steps/steps.go
```diff
@@ -0,0 +1,531 @@
...
+	var u map[string]any
+	if err := yaml.Unmarshal([]byte(yamlContent), &u); err != nil {
+		return nil, err
+	}
+	return &unstructured.Unstructured{Object: u}, nil
+}
+
+func substituteScenarioVars(content string, sc *scenarioContext) string {
+	result := strings.ReplaceAll(content, "$TEST_NAMESPACE", sc.namespace)
+	result = strings.ReplaceAll(result, "$NAME", sc.clusterExtensionName)
```
tmshort: I could see us wanting to expand this list, or even to make this a bit more dynamic. But probably not now.

test/e2e/steps/steps.go
```diff
@@ -0,0 +1,531 @@
...
+func init() {
+	kubeconfigPath = os.Getenv("HOME") + "/.kube/config"
+}
+
+var (
+	logger = log.Log
+)
+
+func kubectl(args ...string) (string, error) {
+	cmd := exec.Command("kubectl", args...)
```
tmshort: Since we run these tests in other environments, we might need to consider the use of `oc` as a substitute for `kubectl`. Here and elsewhere.

test/e2e/steps/steps.go
tmshort: Also noting an inconsistency in substitution mechanisms. Here it's `$VAR`, where as at: https://github.com/operator-framework/operator-controller/pull/2365/files#diff-37528e433a53ab946ef66fda327001b3a125c05c7ac9dfd2b49529fbfdc50cd3R378 it's `{var}`

test/e2e/steps/steps.go
```diff
@@ -0,0 +1,531 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to read RBAC template yaml: %v", err)
+	}
+
+	// Replace template variables
+	yaml := string(yamlContent)
+	yaml = strings.ReplaceAll(yaml, "{namespace}", sc.namespace)
+	yaml = strings.ReplaceAll(yaml, "{serviceaccount_name}", serviceAccount)
+	yaml = strings.ReplaceAll(yaml, "{clusterextension_name}", sc.clusterExtensionName)
+	if len(keyValue) > 0 {
```
tmshort, line 474: Should this be checked for an even value? Having an odd number of arguments is a programming error.

test/e2e/steps/steps.go
```diff
@@ -0,0 +1,492 @@
...
+		"spec": map[string]any{
+			"source": map[string]any{
+				"image": map[string]any{
+					"ref": fmt.Sprintf("%s:%s", base, version),
+				},
+			},
+		},
+	}
+	pb, err := json.Marshal(patch)
+	if err != nil {
```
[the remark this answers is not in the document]
tmshort: Copilot missed `return err`??

jianzhangbjz:
Hi team, what’s the reason for choosing `Cucumber` instead of `Ginkgo`? What’s the background behind this decision? Thanks! I’m asking because the OpenShift QE team used the `Cucumber` framework for E2E testing before 2020, but it took the team almost a year to fully migrate to `Ginkgo`. The main reasons for choosing Ginkgo include:

1. Go-Native design, which fits better with Go-based projects.
2. Expressive BDD-Style Syntax that is cleaner and easier to maintain.
3. More powerful and stable test organization compared to `Cucumber`.
4. Rich Assertion Library (Gomega) that simplifies writing test expectations.
5. Built-in Parallelization for faster test execution.
6. Better for Unit and Integration Tests
7. Widely adopted in the Kubernetes ecosystem, making it a de-facto standard.
8. Progressive enhancement, with active development and community support.

pedjak:
> Hi team, what’s the reason for choosing `Cucumber` instead of `Ginkgo`? 

@jianzhangbjz  thanks for reaching out. I think I have summarized the motivation in the PR description, let me know if some of the points you raised are unanswered. IMO Ginko for e2e is still less readable for non-devs (and for devs as well)

test/e2e/steps/steps.go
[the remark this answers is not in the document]
pedjak: fixed.

test/e2e/features_test.go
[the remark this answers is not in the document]
pedjak: This is just to document exit codes, we do not need to handle separately.

test/e2e/steps/hooks.go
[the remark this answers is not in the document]
pedjak: > Will `errcheck` still complain even if `_ = p.Process...()` is used?

Actually no. Changed to `_ = p.Process.Kill()`

test/e2e/steps/hooks.go
```diff
@@ -0,0 +1,117 @@
...
+			}
+		}
+	}
+	logger.Info(fmt.Sprintf("Enabled feature gates: %v", featureGates))
+}
+
+func CheckFeatureTags(ctx context.Context, sc *godog.Scenario) (context.Context, error) {
+	for _, tag := range sc.Tags {
+		if enabled, found := featureGates[tag.Name[1:]]; !found || (found && !enabled) {
+			logger.V(1).Info(fmt.Sprintf("Skipping scenario %q because feature gate %q is disabled", sc.Name, tag.Name[1:]))
```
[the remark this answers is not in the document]
pedjak: not applicable.

test/e2e/features/recover.feature
pedjak: Yes, that is the one.

test/e2e/steps/hooks.go
pedjak: Actually I think we do not need to wait at all.

test/e2e/steps/hooks.go
pedjak: We could try, but not sure what we are gonna gain with it?

test/e2e/features/install.feature
pedjak: > I wonder if we could strategic merge patch instead and have a smaller yaml to make it more clear what's changing?

We could ofcourse, we craft the grammar and the semantic: how would like to look like? If we would writing user docs, how this should be read by users?

test/e2e/network_policy_test.go
pedjak: I spoke with @joelanford earlier about that and we do not see these as real e2e tests - we would probably like to replace them with an optional CI job that would detect changes in networkpolicies provided inside helm chart and fail the job, similar to crd-diff job.

test/e2e/steps/steps.go
pedjak: fixed.

test/e2e/features/install.feature
pedjak: > The first one is generic, the second one is indicating a very specific resource. What's the reasoning behind this?

Improved readability/focus on what matters. The fact that we use the same go code under the hood for both step is not important here - the reader should immediately understand what a step is about. Hence, we could even replace the generic step "resource is applied" with something like `([[:alnum:]]+) is applied` or even `([[:alnum:]]+) is available` so that we better document what is going on. Also, in this particular case, we could even create very concrete step `namespace ([[:alnum:]]+) is available` that is going to assure that the given namespace is created if not exists already.

test/e2e/steps/hooks.go
joelanford: Deletion can sometimes take a little while if finalizers need to be processed. I assume we want foreground deletion so that we can be sure cleanup is complete before we move on. Seems like it could speed up cleanup considerably in those cases.

test/e2e/steps/steps.go
pedjak: > Also noting an inconsistency in substitution mechanisms. Here it's `$VAR`, where as at: https://github.com/operator-framework/operator-controller/pull/2365/files#diff-37528e433a53ab946ef66fda327001b3a125c05c7ac9dfd2b49529fbfdc50cd3R378 it's `{var}`

IMO, bash-style like variables are understandable/known for a wider audience. We could use those in testdata templates as well.

test/e2e/steps/steps.go
pedjak: added `--k8s.cli` command arg that can tweak that.

test/e2e/steps/hooks.go
pedjak: Sure, we could do it.

test/e2e/features/recover.feature
dtfranz: I think we were hitting rate-limiting because we had the image untagged/set to latest, so it would pull every time no matter what. This is tagged so it should be fine.

test/e2e/features_test.go
```diff
@@ -0,0 +1,71 @@
...
+	default:
+		log.Fatalf("non-zero status returned (%d), failed to run feature tests", sc)
```
tmshort, line 64: Should we capture these different returns, as in:
```
switch {
case sc == 0:
    ... success
case sc == 1:
    ... failure
case sc == 2:
    ... command line usage error
case sc >= 128:
    ... os signal related error
default:
    ... unexpected error
}

File: test/e2e/steps/hooks.go [hunk 22]
```diff
@@ -0,0 +1,142 @@
...
+	if devMode {
+		logger = textlogger.NewLogger(textlogger.NewConfig(textlogger.Verbosity(1)))
+	} else {
+		logger = textlogger.NewLogger(textlogger.NewConfig())
+	}
+
+	raw, err := k8scli("get", "deployment", "-n", olmNamespace, olmDeploymentName, "-o", "json")
+	if err != nil {
+		return
+	}
```
tmshort: We should try to capture and display errors in cases like these, and not fail silently

test/e2e/steps/steps.go
```diff
@@ -0,0 +1,676 @@
...
+	sc.Step(`^resource "deployment/test-operator" reports as (not ready|ready)$`, MarkTestOperatorNotReady)
+
+	sc.Step(`^resource apply fails with error msg containing "([^"]+)"$`, ResourceApplyFails)
+	sc.Step(`^resource "([^"]+)" is eventually restored$`, ResourceRestored)
+	sc.Step(`^resource "([^"]+)" matches$`, ResourceMatches)
+
+	sc.Step(`^Service account "([^"]*)" with needed permissions is available in test namespace$`, ServiceAccountWithNeededPermissionsIsAvailableInNamespace)
+	sc.Step(`^Service account "([^"]*)" in test namespace is cluster admin$`, ServiceAccountWithClusterAdminPermissionsIsAvailableInNamespace)
+	sc.Step(`^Service account "([^"]+)" in test namespace has permissions to fetch "([^"]+)" metrics$`, ServiceAccountWithFetchMetricsPermissions)
+	sc.Step(`^Service account "([^"]+)" sends request to "([^"]+)" endpoint of "([^"]+)" service$`, SendMetricsRequest)
```
tmshort: How consistent do we want to be with resource names. Should these all start with `^serviceaccount`

test/e2e/steps/steps.go
```diff
@@ -0,0 +1,676 @@
...
+
+	sc.Step(`^"([^"]+)" catalog is updated to version "([^"]+)"$`, CatalogIsUpdatedToVersion)
+	sc.Step(`^"([^"]+)" catalog serves bundles$`, CatalogServesBundles)
+	sc.Step(`^"([^"]+)" catalog image version "([^"]+)" is also tagged as "([^"]+)"$`, TagCatalogImage)
```
tmshort, line 86: Same here, do we want these to be referred to as `ClusterCatalogs`?

test/e2e/steps/steps.go
```diff
@@ -0,0 +1,676 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/yaml"
+
+	ocv1 "github.com/operator-framework/operator-controller/api/v1"
+)
+
+const (
+	olmNamespace      = "olmv1-system"
```
tmshort: Since the upstream tests are run downstream, this namespace will change, and the catalogd namespace may differ from the operator-controller namespace. At this time, I'm not sure how this is determined, but it's something we need to handle.
Also, some tests are skipped when running the upstream tests downstream, we need to be able to keep that mechanism.

test/e2e/steps/steps.go
```diff
@@ -0,0 +1,676 @@
...
+	olmDeploymentName = "operator-controller-controller-manager"
+	timeout           = 300 * time.Second
```
tmshort:
```suggestion
	timeout           = 5 * time.Minute
```

test/e2e/steps/steps.go
tmshort: I think both are known, and in BASH, the recommendation is to use `${VAR}` vs `$VAR`, so going with `{VAR}` is not that much different.

test/e2e/features_test.go
pedjak: I would leave it as for now, we can always improve it in the future. I do not any direct benefit for now.

test/e2e/steps/hooks.go
pedjak: Good catch, we cannot return error here due because `BeforeSuite` does not allow returning any value, but we can panic.

test/e2e/steps/steps.go
pedjak: I have updated the reg exps to accept both.

test/e2e/steps/steps.go
pedjak: > Same here, do we want these to be referred to as `ClusterCatalogs`?

Added also `ClusterCatalog "foo" is updated...` alternatives. Not particularly biased here, what reads better in your opinion?

test/e2e/features/install.feature
```diff
@@ -0,0 +1,272 @@
...
+    And "test" catalog serves bundles
+    And service account "olm-sa" with needed permissions is available in test namespace
+  
+  Scenario Outline: Install latest available version
+    When ClusterExtension is applied
+      """
+      apiVersion: olm.operatorframework.io/v1
+      kind: ClusterExtension
+      metadata:
+        name: $NAME
```
perdasilva: could it be worth being more specific with the placeholder names, e.g. `$EXTENSION_NAME` and `$INSTALL_NAMESPACE` to disambiguate from placeholders in other resources? I don't know exactly the scope of the placeholders.

test/e2e/features/install.feature
```diff
@@ -0,0 +1,272 @@
...
+    And resource "configmap/test-configmap" is installed
+    And resource "deployment/test-operator" is installed
+
+    Examples:
+      | package-name |
+      | test         |
+      | test-mirrored |
+
+
+  Scenario: Report that bundle cannot be installed when exists in multiple catalogs with same priority
```
perdasilva:
```suggestion
  Scenario: Report that bundle cannot be installed when it exists in multiple catalogs with same priority
```

test/e2e/features/install.feature
```diff
@@ -0,0 +1,272 @@
...
+      spec:
+        namespace: $TEST_NAMESPACE
+        serviceAccount:
+          name: olm-sa
+        source:
+          sourceType: Catalog
+          catalog:
+            packageName: test
+      """
+    Then ClusterExtension reports Progressing as True with Reason Retrying:
```
perdasilva: nit/suggestion:

```suggestion
    Then ClusterExtension reports Progressing as True with Reason Retrying and Message:
```

test/e2e/features/install.feature
pedjak: `$NAME` represents a unique name among all tests, so we could use it for various purposed within a test and ensure that we do not overlap with other tests, so that parallel execution could be possible. Once it is `metadata.name` of a `ClusterExtension`, it becomes the extension name. Similar goes for `$TEST_NAMESPACE` - a separate namespace is created as a part of each scenario fixture. It becomes the install namespace once we refer it in a `ClusterExtension`

It is a short-cut for test writer, so she/he does not need to worry about name uniqueness. 

Having said that, we can still polish all this is followup PR, nothing is set in stone.

test/e2e/features/install.feature
perdasilva: I guess the underlying user action would be something like `kubectl patch` - so maybe `And user patches ClusterExtension with kubectl` or something like that?

test/e2e/features/install.feature
```diff
@@ -0,0 +1,272 @@
...
+                "olm.operatorframework.io/metadata.name": test-catalog
+      """
+    Then ClusterExtension reports Installed as True
+    And bundle "single-namespace-operator.1.0.0" is installed in version "1.0.0"
+    And operator "single-namespace-operator" target namespace is "single-namespace-operator-target"
+
+  @SingleOwnNamespaceInstallSupport
+  Scenario: watchNamespace config is required for extension supporting own namespace
+    Given service account "olm-admin" in test namespace is cluster admin
+    And ClusterExtension is applied
```
perdasilva: nit: would be cool if we could call out that the ClusterExtension is applied without the watchNamespace configuration, or something like that

test/e2e/features/install.feature
```diff
@@ -0,0 +1,272 @@
...
+              matchLabels:
+                "olm.operatorframework.io/metadata.name": test-catalog
+      """
+    And ClusterExtension reports Progressing as True with Reason Retrying:
+      """
+      error for resolved bundle "own-namespace-operator.1.0.0" with version
+      "1.0.0": invalid ClusterExtension configuration: invalid configuration: required
+      field "watchNamespace" is missing
+      """
+    And ClusterExtension is updated
```
perdasilva: nit: similar as above, "...update to include the watchNamespace configuration"

test/e2e/features/install.feature
```diff
@@ -0,0 +1,272 @@
...
+    """
+      apiVersion: webhook.operators.coreos.io/v1
+      kind: WebhookTest
+      metadata:
+        name: $NAME
+        namespace: $TEST_NAMESPACE
+      spec:
+        valid: true
+        mutate: true
+      """
```
perdasilva: missing new line at the end of the file =D

test/e2e/features/metrics.feature
```diff
@@ -0,0 +1,14 @@
...
+
+  Scenario Outline: component exposes metrics
+    Given service account "metrics-reader" in test namespace has permissions to fetch "<component>" metrics
+    When service account "metrics-reader" sends request to "/metrics" endpoint of "<component>" service
+    Then Prometheus metrics are returned in the response
+
+    Examples:
+      | component           |
+      | operator-controller |
+      | catalogd            |
```
perdasilva: missing new line at the end

test/e2e/features/recover.feature
```diff
@@ -0,0 +1,117 @@
...
+        source:
+          sourceType: Catalog
+          catalog:
+            packageName: test
+            selector:
+              matchLabels:
+                "olm.operatorframework.io/metadata.name": test-catalog
+      """
+    And ClusterExtension is available
+    And resource "configmap/test-configmap" is available
```
perdasilva: nit: 
```suggestion
    And resource "configmap/test-configmap" is exists
```

test/e2e/features/recover.feature
```diff
@@ -0,0 +1,117 @@
...
+        source:
+          sourceType: Catalog
+          catalog:
+            packageName: test
+            selector:
+              matchLabels:
+                "olm.operatorframework.io/metadata.name": test-catalog
+      """
+    And ClusterExtension reports Progressing as True with Reason Retrying
+    When service account "olm-sa" with needed permissions is available in test namespace
```
perdasilva: I wonder if it should be:

```suggestion
    When service account "olm-sa" with needed permissions is available in $TEST_NAMESPACE
```

To make it (extra) clear those two things are the same..?

test/e2e/features/recover.feature
perdasilva: true - we may want to copy it over to quay.io/operator-framework or something (assuming there's no licensing issue with that) to avoid those issues (not for this PR tho...)

test/e2e/features/status.feature
```diff
@@ -0,0 +1,45 @@
+Feature: Report status of the managed ClusterExtension workload
+
+  As an OLM user, I would like to see reported on ClusterExtension the availability
+  change of the managed workload.
+
+  Background:
+    Given OLM is available
+    And "test" catalog serves bundles
+    And service account "olm-sa" with needed permissions is available in test namespace
```
perdasilva: Similar to a previous comment. If we can't do "...is available in the $TEST_NAMESPACE", we might want to use "install namespace" instead - since that might be closer to the user's mental model and the type of verbiage that would normally be used.

test/e2e/features/status.feature
```diff
@@ -0,0 +1,45 @@
...
+            version: 1.0.0
+      """
+    And ClusterExtension is rolled out
+    And ClusterExtension is available
+
+  @BoxcutterRuntime
+  Scenario: Report availability change when managed workload is not ready
+    When resource "deployment/test-operator" reports as not ready
+    Then ClusterExtension reports Available as False with Reason ProbeFailure
+    And ClusterExtensionRevision "$NAME-1" reports Available as False with Reason ProbeFailure
```
perdasilva: Where did $NAME-1 come from? Why not just $NAME?

test/e2e/features/update.feature
```diff
@@ -0,0 +1,298 @@
...
+          catalog:
+            packageName: test
+            selector:
+              matchLabels:
+                "olm.operatorframework.io/metadata.name": test-catalog
+            version: 1.0.0
+      """
+    And ClusterExtension is rolled out
+    And ClusterExtension is available
+    When ClusterExtension is applied
```
perdasilva: nit: would it read better if we had a step specifically for updating the version? e.g.

"And ClusterExtension $NAME is update to select package verion 1.0.1" or something like that...

test/e2e/features/install.feature
```diff
@@ -0,0 +1,272 @@
...
+              matchLabels:
+                "olm.operatorframework.io/metadata.name": test-catalog
+      """
+    And ClusterExtension reports Progressing as True with Reason Retrying:
+      """
+      error for resolved bundle "own-namespace-operator.1.0.0" with version
+      "1.0.0": invalid ClusterExtension configuration: invalid configuration: 'some-ns'
+      is not valid ownNamespaceInstallMode: invalid value "some-ns": watchNamespace
+      must be "$TEST_NAMESPACE" (the namespace where the operator is installed) because this
+      operator only supports OwnNamespace install mode
```
perdasilva, line 206: Not for this PR, but we should eliminate notions of install modes from the errors. This is registry+v1 terminology.

perdasilva: #2386

test/e2e/features/update.feature
perdasilva: Created #2385

test/e2e/features/status.feature
perdasilva: Ah, nevermind, it's for the ClusterExtensionRevision

test/e2e/features/update.feature
```diff
@@ -0,0 +1,298 @@
...
+    Then bundle "test-operator.1.2.0" is installed in version "1.2.0"
+    And ClusterExtension is rolled out
+    And ClusterExtension is available
+    And ClusterExtension reports "$NAME-2" as active revision
+    And ClusterExtensionRevision "$NAME-2" reports Progressing as True with Reason Succeeded
+    And ClusterExtensionRevision "$NAME-2" reports Available as True with Reason ProbesSucceeded
+    And ClusterExtensionRevision "$NAME-1" is archived
+
+  @BoxcutterRuntime
+    @foo
```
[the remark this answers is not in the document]
perdasilva: I think the robot might be right here

test/e2e/steps/hooks.go
```diff
@@ -0,0 +1,150 @@
...
+
+type contextKey string
+
+const (
+	scenarioContextKey contextKey = "scenario-context"
+)
+
+var (
+	devMode      = false
+	featureGates = map[featuregate.Feature]bool{
```
perdasilva, line 45: could we drive this from the configuration already in the features package?

perdasilva: or maybe solely the configuration in the --feature-gates flag?

test/e2e/features/install.feature
pedjak: I would keep it declarative as much as possible, hence mentioning `kubectl` would not be good.

test/e2e/steps/hooks.go
pedjak: we need to prepopulate it here, because not all `--feature-gates` flag are present on the command line, some of them are actually set to true by default. 

we can improve this in followup PRs.

test/e2e/steps/hooks.go
tmshort: Or log, either way.

test/e2e/steps/steps.go
tmshort: Yeah, we should be specific in terms of which resources we are referring to.

test/e2e/features/install.feature
tmshort: I would be concerned with `$NAME` erroneously replacing `$NAME_EXTENSION` (as a bad example), which is why bash also supports braces in variable references (e.g. `${NAME}`), and shellcheck warns on it.

pedjak: I have added braces around variables and I am using `os.Expand(..)` function to resolving.

test/e2e/steps/steps.go
pedjak: added `@mirrored-registry` and `@catalog-updates` that could be used for excluding those tests.

test/e2e/steps/hooks.go
tmshort: We have both disabled and enabled value.yaml fields for the flags, so we _could_ depend on just those `--feature-gates` flags if all feature gates were properly placed in values.yaml.

test/e2e/README.md
```diff
@@ -0,0 +1,346 @@
...
+
+Resources are managed declaratively using YAML templates embedded in feature files as docstrings:
+
+```gherkin
+When ClusterExtension is applied
+"""
+  apiVersion: olm.operatorframework.io/v1
+  kind: ClusterExtension
+  metadata:
+    name: $NAME
```
tmshort: Nit, here and elsewhere. Since we now support bracket variables, and that's what's used in the `.feature` files.
```suggestion
    name: ${NAME}
```
Although I realize both formats are accepted.

test/e2e/features/install.feature
```diff
@@ -0,0 +1,299 @@
...
+    And ClusterExtension is available
+    And bundle "<package-name>-operator.1.2.0" is installed in version "1.2.0"
+    And resource "networkpolicy/test-operator-network-policy" is installed
+    And resource "configmap/test-configmap" is installed
+    And resource "deployment/test-operator" is installed
+
+    Examples:
+      | package-name  |
+      | test-mirrored |
+      | dynamic       |
```
tmshort, line 67: What is this for/do? Does it impact the test?

test/e2e/features/install.feature
```diff
@@ -0,0 +1,299 @@
...
+            selector:
+              matchLabels:
+                "olm.operatorframework.io/metadata.name": test-catalog
+      """
+    And ClusterExtension reports Progressing as True with Reason Retrying and Message:
+      """
+      error for resolved bundle "single-namespace-operator.1.0.0" with version "1.0.0":
+      invalid ClusterExtension configuration: invalid configuration: required field "watchNamespace" is missing
+      """
+    When ClusterExtension is updated to set config.watchNamespace field
```
tmshort, line 125: Future enhancement, patching of just the modified fields?

test/e2e/steps/hooks.go
pedjak: > We have both disabled and enabled value.yaml fields for the flags, so we _could_ depend on just those `--feature-gates` flags if all feature gates were properly placed in values.yaml.

`WebhookProviderCertManager` is by default set to `true`, and not passed via `--future-gates` flag. Ofcourse, we can fix that, but I would rather keep this PR as focused as possible. We could improve that in followup PR, wdyt?

test/e2e/features/recover.feature
tmshort: We use `registry.k8s.io/e2e-test-images/busybox:1.36.1-1` for downstream testing, it ought to be compatible.

test/e2e/features/install.feature
```diff
@@ -0,0 +1,299 @@
...
+      """
+      apiVersion: webhook.operators.coreos.io/v1
+      kind: WebhookTest
+      metadata:
+        name: ${NAME}
+        namespace: ${TEST_NAMESPACE}
+      spec:
+        valid: false # webhook rejects it as invalid value
+      """
+    And resource is applied
```
tmshort, line 267: This feels like it should be a `When` followed by a `Then`

test/e2e/features/status.feature
```diff
@@ -0,0 +1,45 @@
...
+    Then ClusterExtension reports Available as False with Reason ProbeFailure
+    And ClusterExtensionRevision "${NAME}-1" reports Available as False with Reason ProbeFailure
+
+  @BoxcutterRuntime
+  Scenario: Report availability change when managed workload restores its readiness
+    Given resource "deployment/test-operator" reports as not ready
+    And ClusterExtension reports Available as False with Reason ProbeFailure
+    And ClusterExtensionRevision "${NAME}-1" reports Available as False with Reason ProbeFailure
+    When resource "deployment/test-operator" reports as ready
+    Then ClusterExtension is available
```
tmshort, line 44: These two feel like they should be `And`s.

test/e2e/features/install.feature
pedjak: > What is this for/do? Does it impact the test?

A sort of table tests: https://cucumber.io/docs/gherkin/reference/#scenario-outline

test/e2e/features/install.feature
pedjak: I would do in followup PR, if you do not mind.

tmshort:
Regarding the `.feature` files, it seems to me that:

`When` corresponds to an action that the test does, e.g. update a resource

Subsequently:

`Then` and `And` corresponds to conditional checks.

There were some inconsistencies in the test definitions, and I wanted to make sure I understood what was happening

test/e2e/features/recover.feature
pedjak: > We use `registry.k8s.io/e2e-test-images/busybox:1.36.1-1` for downstream testing, it ought to be compatible.

I did not modify this part at all in this PR, see https://github.com/operator-framework/operator-controller/blob/main/test/e2e/cluster_extension_install_test.go#L722

test/e2e/steps/hooks.go
tmshort: Yeah, I'm ok with that. It was something we _could_ do.

test/e2e/network_policy_test.go
```diff
@@ -377,3 +389,25 @@ func validateSingleIngressRule(t *testing.T, policyName string, clusterIngressRu
...
+}
+
+func init() {
+	cfg = ctrl.GetConfigOrDie()
+
+	var err error
+	utilruntime.Must(apiextensionsv1.AddToScheme(scheme.Scheme))
+	c, err = client.New(cfg, client.Options{Scheme: scheme.Scheme})
+	utilruntime.Must(err)
+}
```
tmshort, line 413: You retrieve the deployments for operator-controller and catalogd in test/e2e/steps/hooks.go:BeforeSuite(). Could that be used to get the namespaces instead?
(I realize this is just moving the original definition from somewhere else)

test/e2e/features/install.feature
pedjak: > This feels like it should be a `When` followed by a `Then`

`When` describes the behavior under test, Then's are are assertions, i.e. consequences. In this case,  when we apply the extension, then 

* we assert it is rolled out
* available
* trying to apply problematic resource fails due to implemented webhook logic
* but applying other conformant resource is possible

Therefore, `And` is ok here.

Perhaps we could rephrase/extend `resource is applied` step to indicated it?

test/e2e/features/status.feature
pedjak: > These two feel like they should be `And`s.

We assert here propagation of workload status. Hence, when deployment is reported to be ready, then cluster extension also reports under its status that is available.

pedjak:
> Regarding the `.feature` files, it seems to me that:
> `When` corresponds to an action that the test does, e.g. update a resource
> Subsequently:
> `Then` and `And` corresponds to conditional checks.

Correct.

> There were some inconsistencies in the test definitions, and I wanted to make sure I understood what was happening

:+1:

test/e2e/network_policy_test.go
pedjak: > You retrieve the deployments for operator-controller and catalogd in test/e2e/steps/hooks.go:BeforeSuite(). Could that be used to get the namespaces instead? (I realize this is just moving the original definition from somewhere else)

These tests were not touched, these functions are just copied from the removed test files. I would suggest to rework these tests in the followup PR, wdyt?

test/e2e/steps/testdata/extra-catalog-template.yaml
tmshort, line 1: Should this be named `extra-catalog-template.yaml`?

test/e2e/steps/testdata/test-catalog-template.yaml
tmshort, line 1: Should this be called `test-catalog-template.yaml`?

test/e2e/steps/steps.go
```diff
@@ -0,0 +1,707 @@
...
+
+	sc.Step(`^resource apply fails with error msg containing "([^"]+)"$`, ResourceApplyFails)
+	sc.Step(`^resource "([^"]+)" is eventually restored$`, ResourceRestored)
+	sc.Step(`^resource "([^"]+)" matches$`, ResourceMatches)
+
+	sc.Step(`^[Ss]ervice account "([^"]*)" with needed permissions is available in test namespace$`, ServiceAccountWithNeededPermissionsIsAvailableInNamespace)
+	sc.Step(`^[Ss]ervice account "([^"]*)" with needed permissions is available in \${TEST_NAMESPACE}$`, ServiceAccountWithNeededPermissionsIsAvailableInNamespace)
+	sc.Step(`^[Ss]ervice account "([^"]*)" in test namespace is cluster admin$`, ServiceAccountWithClusterAdminPermissionsIsAvailableInNamespace)
+	sc.Step(`^[Ss]ervice account "([^"]+)" in test namespace has permissions to fetch "([^"]+)" metrics$`, ServiceAccountWithFetchMetricsPermissions)
+	sc.Step(`^[Ss]ervice account "([^"]+)" sends request to "([^"]+)" endpoint of "([^"]+)" service$`, SendMetricsRequest)
```
tmshort: "serviceaccount" should be one word, as that's the name of the resource.

test/e2e/steps/steps.go
```diff
@@ -0,0 +1,707 @@
...
+	req.Header.Set("Authorization", "Bearer "+token)
+
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	return resp, nil
+}
+
+func SendMetricsRequest(ctx context.Context, serviceAccount string, endpoint string, controllerName string) error {
```
tmshort, line 538: Note that his is using port-forwarding to scrape metrics, rather than a pod in the cluster (as the original test does).

test/e2e/network_policy_test.go
tmshort: Since it's just moving the function from a deleted file, that should be fine.

test/e2e/features/status.feature
tmshort: But `When` seems to be more of an action the test takes, rather than the checking of a condition.

pedjak:
> This wasn't always consistent, as some actions seems to be under Then/Ands.

If there are multiple actions then scenario are structured like:

```cucumber
When action1
And action2
And action3
```

Similar for where there are multiple checks:

```cucumber
Then check1
And check2
And check3
```

test/e2e/features/status.feature
tmshort: What is what this appears to be (reporting status).

tmshort:
> > This wasn't always consistent, as some actions seems to be under Then/Ands.
> 
> If there are multiple actions then scenario are structured like:
> 
> ```
> When action1
> And action2
> And action3
> ```
> 
> Similar for where there are multiple checks:
> 
> ```
> Then check1
> And check2
> And check3
> ```

There still seem to be inconsistencies, however, but I will have to go through them since the latest changes,

test/e2e/steps/steps.go
pedjak: Correct, the advantage is that we do not need to pull an additional image + deploy another container.

tmshort: Yup, it's simply noting the change to the test process.

tmshort:
The code looks ok, but when I run this locally on Fedora 42 x86_64 via `make test-e2e`, I'm not seeing any output. Seems to be hanging. I'm also seeing two zombied `kubectl` processes.
```
3366938 pts/13   S+     0:00  |   \_ make test-e2e
3393333 pts/13   Sl+    0:00  |       \_ go test -count=1 -v ./test/e2e/...
3393342 pts/13   Sl+    0:05  |           \_ /home/tshort/sdk/go1.25.3/bin/go test -count=1 -v ./test/e2e/...
3406425 pts/13   Sl+    0:02  |               \_ /tmp/go-build554489889/b001/e2e.test -test.paniconexit0 -test.timeout=10m0s -test.count=1 -test.v=true
3409322 pts/13   Z+     0:00  |                   \_ [kubectl] <defunct>
3411033 pts/13   Z+     0:00  |                   \_ [kubectl] <defunct>
```
...and it finally finished, but the output comes out all at once, which isn't acceptable. We should really be able to see the progress of the test as it runs.

Ah, it was terminated via SIGQUIT for taking too long...
```
SIGQUIT: quit
PC=0x48b6c1 m=0 sigcode=0

...

*** Test killed with quit: ran too long (11m0s).
FAIL    github.com/operator-framework/operator-controller/test/e2e      660.022s
?       github.com/operator-framework/operator-controller/test/e2e/steps        [no test files]
FAIL
make: *** [Makefile:218: e2e] Error 1
```
I rand this twice... I'd like someone else to run this locally to ensure it's _not just me_.

pedjak:
> The code looks ok, but when I run this locally on Fedora 42 x86_64 via `make test-e2e`, I'm not seeing any output. Seems to be hanging. I'm also seeing two zombied `kubectl` processes

I can confirm you the same - I though first it was maybe because `make` buffers the output, but I fixed by slightly changing how we invoke tests to:

```
go test -count=1 -v ./test/e2e/features_test.go
```

pushed the change.

tmshort:
It's working in my local repo now.
I think we want to create a Jira tickets for follow-on work:
* restoring color output
* getting component namespaces from BeforeSuite()
* anything else we want to clean up
* any updates needed for downstreaming

PR #2366, opened 2025-11-28
:seedling: Update inline config docs and annotations

perdasilva:
# Description

Addresses [enhancement PR ](https://github.com/openshift/enhancements/pull/1849) reviewer comments by:

 - Updates inline configuration documentation
 - Adds missing `// +unionMember` annotation to `Inline`
 - Adds `// +kubebuilder:validation:MinProperties=1` validation to `Inline` to ensure empty configurations `{}` cannot be specified
  - Adds additional admission unit test inc. to test the added validation annotation

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/api/v1/clusterextension_types.go
+++ b/api/v1/clusterextension_types.go
@@ -176,12 +176,14 @@ type ClusterExtensionConfig struct {
 	// inline contains JSON or YAML values specified directly in the
 	// ClusterExtension.
 	//
-	// inline must be set if configType is 'Inline'.
-	// inline accepts arbitrary JSON/YAML objects.
-	// inline is validation at runtime against the schema provided by the bundle if a schema is provided.
+	// inline is used to specify arbitrary configuration values for the ClusterExtension.
+	// It must be set if configType is 'Inline' and must be a valid JSON/YAML object containing at least one property.
+	// The configuration values are validated at runtime against a JSON schema provided by the bundle.
 	//
 	// +kubebuilder:validation:Type=object
+	// +kubebuilder:validation:MinProperties=1
 	// +optional
+	// +unionMember
 	Inline *apiextensionsv1.JSON `json:"inline,omitempty"`
 }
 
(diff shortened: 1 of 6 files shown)


api/v1/clusterextension_types.go
```diff
@@ -176,12 +176,14 @@ type ClusterExtensionConfig struct {
...
-	// inline accepts arbitrary JSON/YAML objects.
-	// inline is validation at runtime against the schema provided by the bundle if a schema is provided.
+	// inline is used to specify arbitrary configuration values for the ClusterExtension.
+	// It must be set if configType is 'Inline' and must be a valid JSON/YAML object containing at least one property.
+	// The configuration values are validated at runtime against a JSON schema provided by the bundle.
 	//
 	// +kubebuilder:validation:Type=object
+	// +kubebuilder:validation:MinProperties=1
 	// +optional
+	// +unionMember
```
pedjak, line 186: Adding this did not produce any change on CRD, should it?

perdasilva: Great call out! I went poking about and here's what I found out (with the help of Gemini):

|Annotation	|Effect on CRD Schema	|Notes|
|-------------|-------------------------|-------|
// +union	| None (Visible) |	Treated as documentation only. |
// +unionMember|	None	| Fields remain standard optional pointers. |
// +unionDiscriminator|	None (usually)	| Does not automatically generate an Enum or validation. |

This means that the CEL expressions must be manually created to enforce

While `controller-gen` generates no visible effect, in `client-gen` they seem to be used.

```
If you are generating a full-blown Kubernetes Clientset (using k8s.io/code-generator rather than just controller-runtime), these annotations become functional.

Specifically, they affect the generation of ApplyConfigurations.

    When using Server-Side Apply (SSA) in Go code, the generated ApplyConfiguration structs need to know if a field is a union.

    If tagged correctly, the generated Go client code for "applying" the object will know to clear the other fields in the union when you set a new one.
```

It also suggests to keep them for other developers to make the intent clear to other developers (not a huge argument for me since it's pretty clear anyway, but ok...)

pedjak: ah, the CEL rules is already defined on `ClusterExtensionConfig`.

PR #2371, opened 2025-12-04
:bug: Make deployment replicas configurable via Helm values

jianzhangbjz:
<!--
Please prefix the title of this PR with one of the following icons:

    * 🐛 (:bug:, patch/bug fix)
    * 📖 (:book:, docs)
    * 🌱 (:seedling:, other)

-->

# Description

<!--
Thank you for your contribution!

Please provide a summary of the changes and the motivation behind the change.
-->

## Make deployments HA-ready with configurable replica count

### Problem                                                                                                                                                             
                                                                                                                                                                        
The current deployment configuration hardcodes `replicas: 1` for both `operator-controller-controller-manager` and `catalogd-controller-manager`. This prevents users from deploying OLM in HA mode without modifying templates directly.                                                                                 
                                                                                                                                                                          
### Solution                                                                                                                                                            
                                                                                                                                                                          
Makes the replica count **configurable** via Helm values (default remains **1**):  
  1. Added `replicas` to Helm values:                                                                                                                                     
     - `options.operatorController.deployment.replicas: 1`
     - `options.catalogd.deployment.replicas: 1`                                                                                                                          
  2. Updated Helm templates to use `{{ .Values }}` instead of hardcoded `1`                                                                                               
  3. Regenerated all manifests                                                                                                                                            
                                                                                                                                                                          
### Usage                                                                                                                                                                                                                                                                                                                                 
```yaml
  options:
    operatorController:
      deployment:
        replicas: 2
    catalogd:                                                                                                                                                             
      deployment:
        replicas: 2                                                                                                                                                       
```                                                                                                                                                          
 ### Note  
HA-specific tests (node drain, multi-replica assertions) are intentionally excluded — they require multi-node clusters and belong downstream. This PR provides the necessary configuration knobs. 

## Reviewer Checklist

- [ ] API Go Documentation

Assisted-By: Claude Code

### DIFF
--- a/helm/olmv1/templates/deployment-olmv1-system-catalogd-controller-manager.yml
+++ b/helm/olmv1/templates/deployment-olmv1-system-catalogd-controller-manager.yml
@@ -12,11 +12,11 @@ metadata:
   namespace: {{ .Values.namespaces.olmv1.name }}
 spec:
   minReadySeconds: 5
-  replicas: 1
+  replicas: {{ .Values.options.catalogd.deployment.replicas }}
   strategy:
     type: RollingUpdate
     rollingUpdate:
-      maxSurge: 1          # Allow temporary 2 pods (1 + 1) for zero-downtime updates
+      maxSurge: 1          # Allow temporary extra pod for zero-downtime updates
       maxUnavailable: 0    # Never allow pods to be unavailable during updates
   selector:
     matchLabels:
--- a/helm/olmv1/templates/deployment-olmv1-system-operator-controller-controller-manager.yml
+++ b/helm/olmv1/templates/deployment-olmv1-system-operator-controller-controller-manager.yml
@@ -11,11 +11,11 @@ metadata:
   name: operator-controller-controller-manager
   namespace: {{ .Values.namespaces.olmv1.name }}
 spec:
-  replicas: 1
+  replicas: {{ .Values.options.operatorController.deployment.replicas }}
   strategy:
     type: RollingUpdate
     rollingUpdate:
-      maxSurge: 1          # Allow temporary 2 pods (1 + 1) for zero-downtime updates
+      maxSurge: 1          # Allow temporary extra pod for zero-downtime updates
       maxUnavailable: 0    # Never allow pods to be unavailable during updates
   selector:
     matchLabels:
--- a/helm/olmv1/values.yaml
+++ b/helm/olmv1/values.yaml
@@ -8,6 +8,7 @@ options:
     enabled: true
     deployment:
       image: quay.io/operator-framework/operator-controller:devel
+      replicas: 1
       extraArguments: []
     features:
       enabled: []
@@ -19,6 +20,7 @@ options:
     enabled: true
     deployment:
       image: quay.io/operator-framework/catalogd:devel
+      replicas: 1
       extraArguments: []
     features:
       enabled: []
--- a/manifests/experimental-e2e.yaml
+++ b/manifests/experimental-e2e.yaml
@@ -2582,7 +2582,7 @@ spec:
   strategy:
     type: RollingUpdate
     rollingUpdate:
-      maxSurge: 1          # Allow temporary 2 pods (1 + 1) for zero-downtime updates
+      maxSurge: 1          # Allow temporary extra pod for zero-downtime updates
       maxUnavailable: 0    # Never allow pods to be unavailable during updates
   selector:
     matchLabels:
@@ -2733,7 +2733,7 @@ spec:
   strategy:
     type: RollingUpdate
     rollingUpdate:
-      maxSurge: 1          # Allow temporary 2 pods (1 + 1) for zero-downtime updates
+      maxSurge: 1          # Allow temporary extra pod for zero-downtime updates
       maxUnavailable: 0    # Never allow pods to be unavailable during updates
   selector:
     matchLabels:
--- a/manifests/experimental.yaml
+++ b/manifests/experimental.yaml
@@ -2502,7 +2502,7 @@ spec:
   strategy:
     type: RollingUpdate
     rollingUpdate:
-      maxSurge: 1          # Allow temporary 2 pods (1 + 1) for zero-downtime updates
+      maxSurge: 1          # Allow temporary extra pod for zero-downtime updates
       maxUnavailable: 0    # Never allow pods to be unavailable during updates
   selector:
     matchLabels:
@@ -2640,7 +2640,7 @@ spec:
   strategy:
     type: RollingUpdate
     rollingUpdate:
-      maxSurge: 1          # Allow temporary 2 pods (1 + 1) for zero-downtime updates
+      maxSurge: 1          # Allow temporary extra pod for zero-downtime updates
       maxUnavailable: 0    # Never allow pods to be unavailable during updates
   selector:
     matchLabels:
--- a/manifests/standard-e2e.yaml
+++ b/manifests/standard-e2e.yaml
@@ -1782,7 +1782,7 @@ spec:
   strategy:
     type: RollingUpdate
     rollingUpdate:
-      maxSurge: 1          # Allow temporary 2 pods (1 + 1) for zero-downtime updates
+      maxSurge: 1          # Allow temporary extra pod for zero-downtime updates
       maxUnavailable: 0    # Never allow pods to be unavailable during updates
   selector:
     matchLabels:
@@ -1932,7 +1932,7 @@ spec:
   strategy:
     type: RollingUpdate
     rollingUpdate:
-      maxSurge: 1          # Allow temporary 2 pods (1 + 1) for zero-downtime updates
+      maxSurge: 1          # Allow temporary extra pod for zero-downtime updates
       maxUnavailable: 0    # Never allow pods to be unavailable during updates
   selector:
     matchLabels:
--- a/manifests/standard.yaml
+++ b/manifests/standard.yaml
@@ -1702,7 +1702,7 @@ spec:
   strategy:
     type: RollingUpdate
     rollingUpdate:
-      maxSurge: 1          # Allow temporary 2 pods (1 + 1) for zero-downtime updates
+      maxSurge: 1          # Allow temporary extra pod for zero-downtime updates
       maxUnavailable: 0    # Never allow pods to be unavailable during updates
   selector:
     matchLabels:
@@ -1839,7 +1839,7 @@ spec:
   strategy:
     type: RollingUpdate
     rollingUpdate:
-      maxSurge: 1          # Allow temporary 2 pods (1 + 1) for zero-downtime updates
+      maxSurge: 1          # Allow temporary extra pod for zero-downtime updates
       maxUnavailable: 0    # Never allow pods to be unavailable during updates
   selector:
     matchLabels:


jianzhangbjz:
Encounter this error again.
```go
=== RUN   TestClusterExtensionInstallRegistry/no_registry_configuration_necessary
    cluster_extension_install_test.go:57: When a cluster extension is installed from a catalog
    cluster_extension_install_test.go:58: When the extension bundle format is registry+v1
    helpers.go:263: Ensuring ClusterCatalog has Status.Condition of Progressing with a status == True and reason == Succeeded
    helpers.go:273: Checking that catalog has the expected metadata label
    helpers.go:278: Ensuring ClusterCatalog has Status.Condition of Type = Serving with status == True
    cluster_extension_install_test.go:79: It resolves the specified package with correct bundle path
    cluster_extension_install_test.go:80: By creating the ClusterExtension resource
    cluster_extension_install_test.go:83: By eventually reporting a successful resolution and bundle path
    cluster_extension_install_test.go:88: By eventually reporting progressing as True
    cluster_extension_install_test.go:89: 
        	Error Trace:	/home/runner/work/operator-controller/operator-controller/test/e2e/cluster_extension_install_test.go:94
        	            				/opt/hostedtoolcache/go/1.24.6/x64/src/runtime/asm_amd64.s:1700
        	Error:      	Not equal: 
        	            	expected: "Succeeded"
        	            	actual  : "Retrying"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-Succeeded
        	            	+Retrying
    cluster_extension_install_test.go:89: 
        	Error Trace:	/home/runner/work/operator-controller/operator-controller/test/e2e/cluster_extension_install_test.go:89
        	Error:      	Condition never satisfied
        	Test:       	TestClusterExtensionInstallRegistry/no_registry_configuration_necessary
Failed to list cluster extensions: no matches for kind "ClusterExtensionRevision" in version "olm.operatorframework.io/v1"    helpers.go:321: By deleting ClusterCatalog "test-catalog-jnwbqjlw"
    helpers.go:330: By deleting ClusterExtension "clusterextension-fqz4fls8"
    helpers.go:294: By waiting for CustomResourceDefinitions of "clusterextension-fqz4fls8" to be deleted
    helpers.go:302: By waiting for ClusterRoleBindings of "clusterextension-fqz4fls8" to be deleted
    helpers.go:310: By waiting for ClusterRoles of "clusterextension-fqz4fls8" to be deleted
    helpers.go:340: By deleting ServiceAccount "clusterextension-fqz4fls8"
    helpers.go:349: By deleting Namespace "clusterextension-fqz4fls8"
```

tmshort:
Changing the number of replicas directly impacts the expectations of our e2es. Perhaps default it to 1 or update the tests with some additional logic.

Allowing us to configure the number of replicas and other values will give us better downstream control.

jianzhangbjz:
It failed at, more https://github.com/operator-framework/operator-controller/actions/runs/19926553979/job/57127694168?pr=2371
```go
./testdata/build-test-registry.sh operator-controller-e2e docker-registry localhost/e2e-test-registry:devel
namespace/operator-controller-e2e created
certificate.cert-manager.io/operator-controller-e2e-registry created
deployment.apps/docker-registry created
service/docker-registry created
error: timed out waiting for the condition on deployments/docker-registry
make: *** [Makefile:267: image-registry] Error 1
```

jianzhangbjz:
Hi @tmshort , I have updated all related e2e test cases. Could you help have a review? Thanks!

hack/test/pre-upgrade-setup.sh
```diff
@@ -155,5 +155,5 @@ spec:
       version: 1.0.0
 EOF
 
-kubectl wait --for=condition=Serving --timeout=60s ClusterCatalog $TEST_CLUSTER_CATALOG_NAME
-kubectl wait --for=condition=Installed --timeout=60s ClusterExtension $TEST_CLUSTER_EXTENSION_NAME
+kubectl wait --for=condition=Serving --timeout=5m ClusterCatalog $TEST_CLUSTER_CATALOG_NAME
+kubectl wait --for=condition=Installed --timeout=5m ClusterExtension $TEST_CLUSTER_EXTENSION_NAME
```
pedjak: are these changes really needed in this PR, i.e. can we do them separately?

helm/olmv1/templates/deployment-olmv1-system-catalogd-controller-manager.yml
```diff
@@ -12,11 +12,11 @@ metadata:
   namespace: {{ .Values.namespaces.olmv1.name }}
 spec:
   minReadySeconds: 5
-  replicas: 1
+  replicas: {{ .Values.options.catalogd.deployment.replicas }}
   strategy:
     type: RollingUpdate
     rollingUpdate:
-      maxSurge: 1          # Allow temporary 2 pods (1 + 1) for zero-downtime updates
+      maxSurge: 1          # Allow temporary extra pod (e.g., 3 pods with 2 replicas) for zero-downtime updates
```
pedjak: from operational point of view this is a no-op change - please revert it.

helm/olmv1/templates/deployment-olmv1-system-operator-controller-controller-manager.yml
```diff
@@ -11,11 +11,11 @@ metadata:
   name: operator-controller-controller-manager
   namespace: {{ .Values.namespaces.olmv1.name }}
 spec:
-  replicas: 1
+  replicas: {{ .Values.options.operatorController.deployment.replicas }}
   strategy:
     type: RollingUpdate
     rollingUpdate:
-      maxSurge: 1          # Allow temporary 2 pods (1 + 1) for zero-downtime updates
+      maxSurge: 1          # Allow temporary extra pod (e.g., 3 pods with 2 replicas) for zero-downtime updates
```
pedjak: same as above.

helm/olmv1/values.yaml
```diff
@@ -8,6 +8,7 @@ options:
     enabled: true
     deployment:
       image: quay.io/operator-framework/operator-controller:devel
+      replicas: 2
```
pedjak: I think the default should remain `1`, and set `podDisruptionBudget.enabled` to `false` as default as well. ` >1` is required for HA setups. We should not implicitly enforce after an upgrade that OLM suddenly consumes more cluster resources (i.e. pods). Such things should not be hidden from users.

helm/olmv1/values.yaml
```diff
@@ -19,6 +20,7 @@ options:
     enabled: true
     deployment:
       image: quay.io/operator-framework/catalogd:devel
+      replicas: 2
```
pedjak: same as above.

manifests/experimental-e2e.yaml
```diff
@@ -2111,11 +2107,11 @@ metadata:
   namespace: olmv1-system
 spec:
   minReadySeconds: 5
-  replicas: 1
+  replicas: 2
   strategy:
     type: RollingUpdate
     rollingUpdate:
-      maxSurge: 1          # Allow temporary 2 pods (1 + 1) for zero-downtime updates
+      maxSurge: 1          # Allow temporary extra pod (e.g., 3 pods with 2 replicas) for zero-downtime updates
```
pedjak: same comment as earlier.

manifests/experimental-e2e.yaml
```diff
@@ -2111,11 +2107,11 @@ metadata:
   namespace: olmv1-system
 spec:
   minReadySeconds: 5
-  replicas: 1
+  replicas: 2
```
pedjak: same as above.

manifests/experimental-e2e.yaml
```diff
@@ -2262,11 +2258,11 @@ metadata:
   name: operator-controller-controller-manager
   namespace: olmv1-system
 spec:
-  replicas: 1
+  replicas: 2
```
pedjak: same comment as earlier.

test/e2e/cluster_extension_install_test.go
```diff
@@ -27,8 +27,11 @@ import (
 )
 
 const (
-	artifactName         = "operator-controller-e2e"
-	pollDuration         = time.Minute
+	artifactName = "operator-controller-e2e"
+	// pollDuration is set to 3 minutes to account for leader election time in multi-replica deployments.
```
pedjak: our e2e test do not need to run now in HA setup, having just single replica should be fine. If some additional HA tests are needed, they should be provided downstream.

tmshort: However, the upstream e2e's do run downstream as well, and we need to ensure they can work there as well.

helm/olmv1/templates/deployment-olmv1-system-catalogd-controller-manager.yml
tmshort: Updating the comment makes sense, especially since the number of pods is based on deployment configuration. But I'd suggest something else.

tmshort:
```suggestion
      maxSurge: 1          # Allow temporary extra pod for zero-downtime updates
```

manifests/experimental-e2e.yaml
tmshort: This is a generated file, the change comes from `values.yaml`.

helm/olmv1/values.yaml
tmshort: I generally agree with @pedjak here, we should be defaulting to 1 upstream, but allow the configuration to updated downstream. However, because we are allowing changes here, we will need to ensure that the e2e tests can still be run upstream and downstream.

tmshort:
It would be good to run this downstream as well.

pedjak:
> It would be good to run this downstream as well.

Sure, and so far were were running e2e tests with single replicas downstream, is that not possible meanwhile?

tmshort:
> > It would be good to run this downstream as well.
> 
> Sure, and so far were were running e2e tests with single replicas downstream, is that not possible meanwhile?

If we have to run two replicas downstream to handle the upgrade bug, then we'll be running the tests against two replicas.

I'm trying this out now, we'll see how it goes.

tmshort:
I just noticed we have `{{ .Values.options.operatorController.podDisruptionBudget.maxUnavailable }}` in the template, but we don't set a default in values.yaml.

tmshort:
Can you also update the commit description?

pedjak:
> If we have to run two replicas downstream to handle the upgrade bug, then we'll be running the tests against two replicas.

If I understand correctly, the test reported as failing in OCPBUGS-62517 is not a part of upstream e2e suite.

tmshort:
> > If we have to run two replicas downstream to handle the upgrade bug, then we'll be running the tests against two replicas.
> 
> If I understand correctly, the test reported as failing in OCPBUGS-62517 is not a part of upstream e2e suite.

Correct. But to fix OCPBUGS-62517, we will need to have two replicas on OCP. Thus, when we run our upstream e2e's on OCP, they will encounter two replicas,and so the upstream e2e's will need to handle more than one replica, _even if we default upstream to 1_, which I also recommend.

My downstream testing seems to be passing, once I updated the downstream helm config to 2 replicas and re-enabling the PDB.

helm/olmv1/values.yaml
tmshort: Now thinking, _maybe_ we want to use 1 for the default, and put 2 in experimental? That way, we can test both values.

hack/test/pre-upgrade-setup.sh
jianzhangbjz: I guess no. I updated it since it failed during the test. You can check the test history.

helm/olmv1/templates/deployment-olmv1-system-catalogd-controller-manager.yml
jianzhangbjz: The comment is necessary so that others can understand the reason.

helm/olmv1/values.yaml
jianzhangbjz: The community user will also encounter a similar upgrade error([OCPBUGS-62517](https://issues.redhat.com/browse/OCPBUGS-62517)) if we disable the PDB as the default.

jianzhangbjz:
To address:
```go
=== RUN   TestClusterExtensionVersionUpdate
    single_namespace_support_test.go:361: When a cluster extension is installed from a catalog
    single_namespace_support_test.go:362: When resolving upgrade edges
    helpers.go:266: Ensuring ClusterCatalog has Status.Condition of Progressing with a status == True and reason == Succeeded
    helpers.go:276: Checking that catalog has the expected metadata label
    helpers.go:281: Ensuring ClusterCatalog has Status.Condition of Type = Serving with status == True
    single_namespace_support_test.go:368: By creating an ClusterExtension at a specified version
    single_namespace_support_test.go:383: By eventually reporting a successful resolution
    single_namespace_support_test.go:392: It allows to upgrade the ClusterExtension to a non-successor version
    single_namespace_support_test.go:393: By forcing update of ClusterExtension resource to a non-successor version
    single_namespace_support_test.go:398: By eventually reporting a satisfiable resolution
    single_namespace_support_test.go:406: We should have two ClusterExtensionRevision resources
panic: test timed out after 10m0s
	running tests:
		TestClusterExtensionVersionUpdate (10s)

goroutine 2466 [running]:
testing.(*M).startAlarm.func1()
```

helm/olmv1/values.yaml
jianzhangbjz: In my opinion, we should maintain consistency between the upstream and downstream projects. This is also **one of the core principles of open source:** fostering alignment, collaboration, and transparency across the entire ecosystem. When upstream and downstream diverge without clear justification, it creates unnecessary complexity, increases maintenance overhead, and makes it harder for users and contributors to understand or adopt the project.

By keeping the two sides consistent, we ensure that new features, bug fixes, and improvements flow smoothly in both directions. It also strengthens long-term sustainability, because the community can rely on a unified architecture and shared expectations. Ultimately, this consistency reinforces the value of open source itself—working together toward a common, coherent, and maintainable solution.

test/e2e/cluster_extension_install_test.go
pedjak: Sure, but if I am not mistaken, [OCPBUGS-62517](https://issues.redhat.com/browse/OCPBUGS-62517) reports a downstream test to fail, not an upstream one?

helm/olmv1/values.yaml
pedjak: > The community user will also encounter a similar upgrade error

As usual, it depends. IMHO, this is the downstream bug because the downstream has certain expectations about components availability during upgrade. If OLM deployment has just single pod, it is perfectly normal that the deployment becomes shortly unavailable during upgrade, and many upstream users might be ok with that, especially because OLM is a controller that performs reconciliations asynchronously (i.e. eventually). All ClusterExtensions that get created/modified during OLM upgrade are not lost and they are going to be reconciled - hence to service interruption does not happens at all.

Having said that, it is also perfectly fine that we add knobs on the upstream, so that the number of pods and pdbs can be configured/added for users that have stronger requirements, but we should not enforce them, especially if that means that suddenly running OLM requires twice as much resources.

test/e2e/cluster_extension_install_test.go
tmshort: Yes, but it's all the same downstream (OCP) configuration. We _can't_ change the OCP configuration _just_ to run our upstream e2e's. Our upstream e2e's are expected to be able to run mostly unmodified on OCP.

The upstream e2e will encounter 2 replicas when run in OCP (because that's the OCP configuration), but will encounter 1 replica upstream (because that's the upstream configuration).

And because we'd also allow the user to support multiple replicas upstream, our upstream e2e should accommodate that.

helm/olmv1/values.yaml
tmshort: I would still want this to be 1 for the standard setup (and tests).

We can set this to 2 in `experimental.yaml` to also test the multi-replica e2e without forcing the default user to have 2 replicas.

test/e2e/cluster_extension_install_test.go
pedjak: > our upstream e2e should accommodate that.

Agree that our e2e tests should be agnostic to underlying setup, but our upstream e2e CI test jobs should run against an OLM deployment with single replicas (as they do until now).

tmshort:
I'm going to walk back my comment on putting `replicas: 2` into experimental. I agree with @pedjak that we should have 1, but support 2. What I _really_ want is to have our e2e tests be able to test both the 1 replica and 2 replica scenarios.

EDIT: This can easily be done by making `experimental-e2e.yaml` have two replicas - which won't impact the released YAML files. Create a new values.yaml file that _only_ specifies 2 replicas, and add it to list in the Makefile here: https://github.com/operator-framework/operator-controller/blob/d8fcdd97157131c2efd798e27008bf4f3ed4ce62/Makefile#L161

jianzhangbjz:
Updated this PR based on the above comments. Please have a review, thanks!

helm/replicas-2.yaml
```diff
@@ -0,0 +1,9 @@
+# Override values to set replicas to 2 for both operator-controller and catalogd
```
pedjak: can you come up with some better name for this file, give that these are helm values for HA setup?

test/e2e/cluster_extension_install_test.go
```diff
@@ -663,7 +669,7 @@ func TestClusterExtensionRecoversFromNoNamespaceWhenFailureFixed(t *testing.T) {
 		require.Equal(ct, ocv1.ReasonSucceeded, cond.Reason)
 		require.Contains(ct, cond.Message, "Installed bundle")
 		require.NotEmpty(ct, clusterExtension.Status.Install)
-	}, pollDuration, pollInterval)
+	}, 5*time.Minute, pollInterval)
```
pedjak: why replacing a variable with an value here, we could update `pollDuration` varibale to 5m?

test/e2e/cluster_extension_install_test.go
```diff
@@ -785,7 +793,7 @@ func TestClusterExtensionRecoversFromExistingDeploymentWhenFailureFixed(t *testi
...
```
pedjak: same as above.

test/e2e/single_namespace_support_test.go
```diff
@@ -404,9 +405,12 @@ func TestClusterExtensionVersionUpdate(t *testing.T) {
...
 	t.Log("We should have two ClusterExtensionRevision resources")
+	// Use 5 minutes for checking ClusterExtensionRevision creation after upgrade.
+	// In multi-replica deployments, revision creation happens after the upgrade completes,
+	// which includes leader election time (up to 163s) plus reconciliation overhead.
 	require.EventuallyWithT(t, func(ct *assert.CollectT) {
 		cerList := &ocv1.ClusterExtensionRevisionList{}
 		require.NoError(ct, c.List(context.Background(), cerList))
 		require.Len(ct, cerList.Items, 2)
-	}, pollDuration, pollInterval)
+	}, 5*time.Minute, pollInterval)
```
pedjak: same as above.

test/e2e/webhook_support_test.go
```diff
@@ -147,7 +150,7 @@ func TestWebhookSupport(t *testing.T) {
 		require.Contains(ct, cond.Message, "Installed bundle")
 		require.NotNil(ct, clusterExtension.Status.Install)
 		require.NotEmpty(ct, clusterExtension.Status.Install.Bundle)
-	}, pollDuration, pollInterval)
+	}, 5*time.Minute, pollInterval)
```
pedjak: same as above.

test/helpers/helpers.go
```diff
@@ -289,31 +292,34 @@ func ValidateCatalogUnpackWithName(t *testing.T, catalogName string) {
 func EnsureNoExtensionResources(t *testing.T, clusterExtensionName string) {
 	ls := labels.Set{"olm.operatorframework.io/owner-name": clusterExtensionName}
 
-	// CRDs may take an extra long time to be deleted, and may run into the following error:
-	// Condition=Terminating Status=True Reason=InstanceDeletionFailed Message="could not list instances: storage is (re)initializing"
+	// CRDs may take extra time to be deleted in multi-replica deployments due to leader election
```
pedjak: CRD deletion is not anyhow impacted by having a controller in HA setup or not. Please correct/revert the comment.

test/upgrade-e2e/post_upgrade_test.go
```diff
@@ -217,13 +235,101 @@ func waitForDeployment(t *testing.T, ctx context.Context, controlPlaneLabel stri
...
 		require.NoError(ct, c.List(ctx, &managerPods, client.MatchingLabelsSelector{Selector: deploymentLabelSelector}))
-		require.Len(ct, managerPods.Items, 1)
+		require.Len(ct, managerPods.Items, int(desiredNumReplicas))
 	}, time.Minute, time.Second)
-	return &managerPods.Items[0]
+}
+
+// findLeaderPod finds the pod that has acquired the leader lease by checking logs of all pods
+func findLeaderPod(ctx context.Context, controlPlaneLabel string) (*corev1.Pod, error) {
+	deploymentLabelSelector := labels.Set{"app.kubernetes.io/name": controlPlaneLabel}.AsSelector()
```
pedjak: Checking logs is for a message to understand if a pod is leader is fragile - instead we should inspect directly k8s lease resource to understand who is the lease holder. See https://kubernetes.io/docs/concepts/architecture/leases/

manifests/experimental-e2e.yaml
tmshort: This is the experimental-e2e file, so we should expect to see 2 here with the new helm values file.

test/e2e/cluster_extension_install_test.go
tmshort: Based on the latest update, the standard `test-e2e` will use a single replica. Running `test-experimental-e2e` will use two replicas. Installing either standard or experimental (outside of the e2e) will use a single replica.

test/helpers/helpers.go
```diff
@@ -289,31 +292,34 @@ func ValidateCatalogUnpackWithName(t *testing.T, catalogName string) {
...
+	// and finalizer processing. Use 3 minutes which accounts for leader election (up to 163s)
+	// plus buffer for reconciliation overhead.
 	t.Logf("By waiting for CustomResourceDefinitions of %q to be deleted", clusterExtensionName)
 	require.EventuallyWithT(t, func(ct *assert.CollectT) {
 		list := &apiextensionsv1.CustomResourceDefinitionList{}
 		err := c.List(context.Background(), list, client.MatchingLabelsSelector{Selector: ls.AsSelector()})
 		require.NoError(ct, err)
 		require.Empty(ct, list.Items)
-	}, 5*pollDuration, pollInterval)
+	}, pollDuration, pollInterval)
```
tmshort: Previously this was 5m, you've reduced it down to 3m?

test/e2e/cluster_extension_install_test.go
tmshort: Looks as though `pollDuration` was changed from 1m to 3m, but in this instance 5m is desired? I agree that we should probably just use `pollDuration` set to 5m.

test/helpers/helpers.go
jianzhangbjz: Yes, to address this test failure. I understood the root cause is not the `timeout`, but the `namespace deletion`.
```go
> === RUN   TestClusterExtensionVersionUpdate
    single_namespace_support_test.go:362: When a cluster extension is installed from a catalog
    single_namespace_support_test.go:363: When resolving upgrade edges
    helpers.go:266: Ensuring ClusterCatalog has Status.Condition of Progressing with a status == True and reason == Succeeded
    helpers.go:276: Checking that catalog has the expected metadata label
    helpers.go:281: Ensuring ClusterCatalog has Status.Condition of Type = Serving with status == True
    single_namespace_support_test.go:369: By creating an ClusterExtension at a specified version
    single_namespace_support_test.go:384: By eventually reporting a successful resolution
    single_namespace_support_test.go:393: It allows to upgrade the ClusterExtension to a non-successor version
    single_namespace_support_test.go:394: By forcing update of ClusterExtension resource to a non-successor version
    single_namespace_support_test.go:399: By eventually reporting a satisfiable resolution
    single_namespace_support_test.go:407: We should have two ClusterExtensionRevision resources
    helpers.go:324: By deleting ClusterCatalog "test-catalog-mc67gxcb"
    helpers.go:333: By deleting ClusterExtension "clusterextension-k8dcdzfb"
    helpers.go:297: By waiting for CustomResourceDefinitions of "clusterextension-k8dcdzfb" to be deleted
    helpers.go:305: By waiting for ClusterRoleBindings of "clusterextension-k8dcdzfb" to be deleted
    helpers.go:313: By waiting for ClusterRoles of "clusterextension-k8dcdzfb" to be deleted
    helpers.go:343: By deleting ServiceAccount "clusterextension-k8dcdzfb"
    helpers.go:352: By deleting Namespace "clusterextension-k8dcdzfb"
panic: test timed out after 10m0s
    running tests:
        TestClusterExtensionVersionUpdate (1m22s)
```
We were running `go test` without specifying `-timeout`, so it used the **default 10-minute timeout**, which is sufficient for a single test. I then optimized the resource cleanup phase and reduced the cleanup duration from `15/6/6` minutes to `3/3/3` minutes.

jianzhangbjz:
Encounter this `deleting Namespace` timeout error again. To address it, I set `cleanupTimeout = 2 * time.Minute` and use `extendedPollDuration(5mins)` for namespace deletion.
```go
=== RUN   TestClusterExtensionRevision
    feature_gates.go:73: Feature-gate "APIV1MetasHandler" enabled
    feature_gates.go:73: Feature-gate "SingleOwnNamespaceInstallSupport" enabled
    feature_gates.go:73: Feature-gate "PreflightPermissions" enabled
    feature_gates.go:73: Feature-gate "HelmChartSupport" enabled
    feature_gates.go:73: Feature-gate "BoxcutterRuntime" enabled
    feature_gates.go:76: Feature-gate "WebhookProviderOpenshiftServiceCA" disabled
    cluster_extension_revision_test.go:29: When a cluster extension is installed from a catalog
    cluster_extension_revision_test.go:30: When the extension bundle format is registry+v1
    helpers.go:269: Ensuring ClusterCatalog has Status.Condition of Progressing with a status == True and reason == Succeeded
    helpers.go:279: Checking that catalog has the expected metadata label
    helpers.go:284: Ensuring ClusterCatalog has Status.Condition of Type = Serving with status == True
    cluster_extension_revision_test.go:52: It resolves the specified package with correct bundle path
    cluster_extension_revision_test.go:53: By creating the ClusterExtension resource
    cluster_extension_revision_test.go:56: By eventually reporting a successful resolution and bundle path
    cluster_extension_revision_test.go:61: By revision-1 eventually reporting Progressing:True:Succeeded and Available:True:ProbesSucceeded conditions
    cluster_extension_revision_test.go:76: By eventually reporting progressing as True
    cluster_extension_revision_test.go:85: By eventually installing the package successfully
    cluster_extension_revision_test.go:96: Check Deployment Availability Probe
    cluster_extension_revision_test.go:97: By making the operator pod not ready
    cluster_extension_revision_test.go:101: By revision-1 eventually reporting Progressing:True:Succeeded and Available:False:ProbeFailure conditions
    cluster_extension_revision_test.go:115: By making the operator pod ready
    cluster_extension_revision_test.go:119: By revision-1 eventually reporting Progressing:True:Succeeded and Available:True:ProbesSucceeded conditions
    cluster_extension_revision_test.go:133: Check archiving
    cluster_extension_revision_test.go:134: By upgrading the cluster extension to v1.2.0
    cluster_extension_revision_test.go:141: By revision-2 eventually reporting Progressing:True:Succeeded and Available:True:ProbesSucceeded conditions
    cluster_extension_revision_test.go:155: By eventually reporting progressing, available, and installed as True
    cluster_extension_revision_test.go:171: By revision-1 eventually reporting Progressing:False:Archived and Available:Unknown:Archived conditions
    helpers.go:327: By deleting ClusterCatalog "test-catalog-dgjm7x6k"
    helpers.go:336: By deleting ClusterExtension "clusterextension-7tzfwk6l"
    helpers.go:300: By waiting for CustomResourceDefinitions of "clusterextension-7tzfwk6l" to be deleted
    helpers.go:308: By waiting for ClusterRoleBindings of "clusterextension-7tzfwk6l" to be deleted
    helpers.go:316: By waiting for ClusterRoles of "clusterextension-7tzfwk6l" to be deleted
    helpers.go:346: By deleting ServiceAccount "clusterextension-7tzfwk6l"
    helpers.go:355: By deleting Namespace "clusterextension-7tzfwk6l"
panic: test timed out after 10m0s
	running tests:
		TestClusterExtensionRevision (34s)
```

helm/high-availability.yaml
```diff
@@ -0,0 +1,10 @@
+# High Availability (HA) configuration for OLMv1
```
pedjak: I think this file should be moved to `helm/olmv1` and name it `values-high-availability.yaml` so it is clear that it contains chart values.

test/e2e/cluster_extension_install_test.go
```diff
@@ -169,18 +175,19 @@ location = "docker-registry.operator-controller-e2e.svc.cluster.local:5000"`,
...
+	// Total: 2 min (ConfigMap) + 2.7 min (leader election) + buffer = 5 minutes
 	t.Log("By eventually reporting progressing as True")
 	require.EventuallyWithT(t, func(ct *assert.CollectT) {
 		require.NoError(ct, c.Get(context.Background(), types.NamespacedName{Name: clusterExtension.Name}, clusterExtension))
 		cond := apimeta.FindStatusCondition(clusterExtension.Status.Conditions, ocv1.TypeProgressing)
 		require.NotNil(ct, cond)
 		require.Equal(ct, metav1.ConditionTrue, cond.Status)
 		require.Equal(ct, ocv1.ReasonSucceeded, cond.Reason)
-	}, 2*time.Minute, pollInterval)
+	}, 5*time.Minute, pollInterval)
```
pedjak: we could use here `extendedPollDuration` variable.

pedjak (commented):
There are still unaddressed comment like this https://github.com/operator-framework/operator-controller/pull/2371/files#r2598622299 from the previous PR review round.

Also, please update PR title to be set to `Make deployments HA-ready with configurable replica count` existing in PR description.

helm/high-availability.yaml
tmshort: We've been placing values files directly into the `helm` directory, so I think the locations is fine for now.
We can look at moving values files "into the charts" in another PR.

test/e2e/cluster_extension_install_test.go
joelanford: I always perk up when there's a suggestion to modify a global timeout to move higher. If we change it globally, that means everything that is passing quickly with the current interval will be able to regress to the new interval without anyone noticing.

My question: what's wrong with the way `pollDuration` is set now, and why _can't_ we make sure the tests continue running within that time budget?

test/e2e/cluster_extension_install_test.go
```diff
@@ -27,8 +27,14 @@ import (
 )
 
 const (
-	artifactName         = "operator-controller-e2e"
-	pollDuration         = time.Minute
+	artifactName = "operator-controller-e2e"
+	// pollDuration is set to 3 minutes to account for leader election time in multi-replica deployments.
+	// In the worst case (previous leader crashed), leader election can take up to 163 seconds
+	// (LeaseDuration: 137s + RetryPeriod: 26s). Adding buffer for reconciliation time.
+	pollDuration = 3 * time.Minute
```
joelanford: We originally has `pollDuration = 1 * time.Minute`. I think we should stick to that. And any tests that require a pod restart (and thus a leader election) can use the extended interval. If we do that, what breaks (and why)?

test/e2e/cluster_extension_install_test.go
```diff
@@ -27,8 +27,14 @@ import (
 )
 
 const (
-	artifactName         = "operator-controller-e2e"
-	pollDuration         = time.Minute
+	artifactName = "operator-controller-e2e"
+	// pollDuration is set to 3 minutes to account for leader election time in multi-replica deployments.
+	// In the worst case (previous leader crashed), leader election can take up to 163 seconds
+	// (LeaseDuration: 137s + RetryPeriod: 26s). Adding buffer for reconciliation time.
```
joelanford: We are using `LeaderReleaseOnCancel: true` for both catalogd and operator-controller, which means that as long as we are not crashing, the lease should be released immediately, and the election should take no longer than the RetryPeriod (26s), right?

test/e2e/cluster_extension_install_test.go
joelanford: Leader election happens right now, even in the non-HA configuration right? If so, it seems like the addition of another replica should not affect the timing. What am I missing?

test/e2e/cluster_extension_install_test.go
```diff
@@ -655,6 +662,8 @@ func TestClusterExtensionRecoversFromNoNamespaceWhenFailureFixed(t *testing.T) {
 	// backoff of this eventually check we MUST ensure we do not touch the ClusterExtension
 	// after creating int the Namespace and ServiceAccount.
 	t.Log("By eventually installing the package successfully")
+	// Use 5 minutes for recovery tests to account for exponential backoff after repeated failures
+	// plus leader election time (up to 163s in worst case)
```
joelanford: Why do we need to account for leader election time here? This test should not cause a restart of our controllers. The original 1m duration seems like it has been sufficient for expotential backoff concerns?

test/e2e/cluster_extension_install_test.go
```diff
@@ -785,7 +796,7 @@ func TestClusterExtensionRecoversFromExistingDeploymentWhenFailureFixed(t *testi
 		require.Equal(ct, ocv1.ReasonSucceeded, cond.Reason)
 		require.Contains(ct, cond.Message, "Installed bundle")
 		require.NotEmpty(ct, clusterExtension.Status.Install)
-	}, pollDuration, pollInterval)
+	}, extendedPollDuration, pollInterval)
```
joelanford: Same as above:
> Why do we need to account for leader election time here? This test should not cause a restart of our controllers. The original 1m duration seems like it has been sufficient for expotential backoff concerns?

test/e2e/cluster_extension_install_test.go
joelanford: In addition, the underlying assumption of this check requires that the same controller process returns the first error and then backs off until the issue is resolved. If there is a restart of the pod during this test, that would result in a false positive because we are longer sure that the error self-resolves _without_ requiring a pod restart.

NOTE: OLMv0 has quite a few cases where we say "have you tried restarted the catalog-operator and olm-operator pods?" This test and other tests related to automatic recovery and self-healing are here to make sure we do not require pod restarts to get things working again.

jianzhangbjz:
I often encounter this kind of failure, which was not introduced by this PR.  It shows that reducing `pollDuration` from 3 minutes back to 1 minute caused a timeout. The test is expecting Reason: "Succeeded" but it's still "Retrying" after 1 minute. 
**The previous timeout adjustment was made to address various timeout failures.** Anyway, to preserve `pollDuration = time.Minute`, I added `catalogPollDuration = 3 * time.Minute`.
```go
=== RUN   TestClusterExtensionInstallReResolvesWhenNewCatalog
    cluster_extension_install_test.go:473: When a cluster extension is installed from a catalog
    cluster_extension_install_test.go:474: It resolves again when a new catalog is available
    cluster_extension_install_test.go:515: It resolves the specified package with correct bundle path
    cluster_extension_install_test.go:516: By creating the ClusterExtension resource
    cluster_extension_install_test.go:519: By reporting a successful resolution and bundle path
    cluster_extension_install_test.go:520: 
        	Error Trace:	/home/runner/work/operator-controller/operator-controller/test/e2e/cluster_extension_install_test.go:525
        	            				/opt/hostedtoolcache/go/1.24.6/x64/src/runtime/asm_amd64.s:1700
        	Error:      	Not equal: 
        	            	expected: "Succeeded"
        	            	actual  : "Retrying"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-Succeeded
        	            	+Retrying
    cluster_extension_install_test.go:520: 
        	Error Trace:	/home/runner/work/operator-controller/operator-controller/test/e2e/cluster_extension_install_test.go:520
        	Error:      	Condition never satisfied
        	Test:       	TestClusterExtensionInstallReResolvesWhenNewCatalog
    helpers.go:330: By deleting ClusterCatalog "test-catalog-v8j2d2kn"
    helpers.go:339: By deleting ClusterExtension "clusterextension-bqjfp4lc"
    helpers.go:303: By waiting for CustomResourceDefinitions of "clusterextension-bqjfp4lc" to be deleted
    helpers.go:311: By waiting for ClusterRoleBindings of "clusterextension-bqjfp4lc" to be deleted
    helpers.go:319: By waiting for ClusterRoles of "clusterextension-bqjfp4lc" to be deleted
    helpers.go:349: By deleting ServiceAccount "clusterextension-bqjfp4lc"
    helpers.go:358: By deleting Namespace "clusterextension-bqjfp4lc"
--- FAIL: TestClusterExtensionInstallReResolvesWhenNewCatalog (68.41s)
```

jianzhangbjz:
We’re seeing this same issue again…
```go
=== RUN   TestWebhookSupport
    feature_gates.go:56: Feature-gate "WebhookProviderCertManager" not found in arguments, defaulting to true
    webhook_support_test.go:31: Test support for bundles with webhooks
    webhook_support_test.go:40: By creating install namespace, and necessary rbac resources
    webhook_support_test.go:85: By creating the webhook-operator ClusterCatalog
    webhook_support_test.go:105: By waiting for the catalog to serve its metadata
    webhook_support_test.go:114: By installing the webhook-operator ClusterExtension
    webhook_support_test.go:140: By waiting for webhook-operator extension to be installed successfully
panic: test timed out after 10m0s
	running tests:
		TestWebhookSupport (3s)

goroutine 2508 [running]:
testing.(*M).startAlarm.func1()
	/opt/hostedtoolcache/go/1.24.6/x64/src/testing/testing.go:2484 +0x394
created by time.goFunc
	/opt/hostedtoolcache/go/1.24.6/x64/src/time/sleep.go:215 +0x2d
```

jianzhangbjz:
Again...
```go
=== RUN   TestClusterExtensionForceInstallNonSuccessorVersion
    cluster_extension_install_test.go:314: When a cluster extension is installed from a catalog
    cluster_extension_install_test.go:315: When resolving upgrade edges
    helpers.go:271: Ensuring ClusterCatalog has Status.Condition of Progressing with a status == True and reason == Succeeded
    helpers.go:281: Checking that catalog has the expected metadata label
    helpers.go:286: Ensuring ClusterCatalog has Status.Condition of Type = Serving with status == True
    cluster_extension_install_test.go:321: By creating an ClusterExtension at a specified version
    cluster_extension_install_test.go:336: By eventually reporting a successful resolution
    cluster_extension_install_test.go:337: 
        	Error Trace:	/home/runner/work/operator-controller/operator-controller/test/e2e/cluster_extension_install_test.go:342
        	            				/opt/hostedtoolcache/go/1.24.6/x64/src/runtime/asm_amd64.s:1700
        	Error:      	Not equal: 
        	            	expected: "Succeeded"
        	            	actual  : "Retrying"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-Succeeded
        	            	+Retrying
    cluster_extension_install_test.go:337: 
        	Error Trace:	/home/runner/work/operator-controller/operator-controller/test/e2e/cluster_extension_install_test.go:337
        	Error:      	Condition never satisfied
        	Test:       	TestClusterExtensionForceInstallNonSuccessorVersion
    helpers.go:333: By deleting ClusterCatalog "test-catalog-lt5hffsm"
    helpers.go:342: By deleting ClusterExtension "clusterextension-clhcbr2s"
    helpers.go:306: By waiting for CustomResourceDefinitions of "clusterextension-clhcbr2s" to be deleted
    helpers.go:314: By waiting for ClusterRoleBindings of "clusterextension-clhcbr2s" to be deleted
    helpers.go:322: By waiting for ClusterRoles of "clusterextension-clhcbr2s" to be deleted
    helpers.go:352: By deleting ServiceAccount "clusterextension-clhcbr2s"
    helpers.go:361: By deleting Namespace "clusterextension-clhcbr2s"
--- FAIL: TestClusterExtensionForceInstallNonSuccessorVersion (69.35s)
```

helm/high-availability.yaml
pedjak: > We've been placing values files directly into the `helm` directory, so I think the locations is fine for now. We can look at moving values files "into the charts" in another PR.

Ah, ok I missed that.

helm/olmv1/templates/deployment-olmv1-system-catalogd-controller-manager.yml
```diff
@@ -12,11 +12,11 @@ metadata:
   namespace: {{ .Values.namespaces.olmv1.name }}
 spec:
   minReadySeconds: 5
-  replicas: 1
+  replicas: {{ .Values.options.catalogd.deployment.replicas }}
```
joelanford, line 15: Do we already have node anti affinity configured to make sure these replicas do not end up on the same node? If not, we need that as well (but only when replicas > 1).

test/upgrade-e2e/post_upgrade_test.go
```diff
@@ -217,13 +236,84 @@ func waitForDeployment(t *testing.T, ctx context.Context, controlPlaneLabel stri
...
+				continue
+			}
+
+			// If there's only one pod, it must be the leader
+			if len(managerPods.Items) == 1 {
+				return &managerPods.Items[0], nil
+			}
+
+			// For multi-replica deployments, check the Lease resource to find the leader
+			var lease coordinationv1.Lease
```
pedjak: Given that we use `testify` library in this project, we could use `require.Eventually()` for polling instead of implementing our custom solution and improve readability.

test/upgrade-e2e/post_upgrade_test.go
```diff
@@ -217,13 +236,84 @@ func waitForDeployment(t *testing.T, ctx context.Context, controlPlaneLabel stri
...
+			if err := c.List(ctx, &managerPods, client.MatchingLabelsSelector{Selector: deploymentLabelSelector}); err != nil {
+				continue
+			}
+
+			if len(managerPods.Items) == 0 {
+				continue
+			}
+
+			// If there's only one pod, it must be the leader
+			if len(managerPods.Items) == 1 {
```
pedjak: we could be wrong here, if we get response from API server right before second pod is created.

Instead of first listing pods, I would suggest to fetch lease first, locate pod via lease holder and then check if that pod runs the appropriate OLM version.

tmshort:
> Given that PR is about fixing upgrade tests. I would suggest that we drop changes on `test/e2e` files, because they should not be affected and if needed we can always address them in a followup PR.

However, we will still need changes to `test/e2e` to validate that we can handle multiple replicas.

joelanford:
I think we need to investigate further to understand:
1. Whether using 2 replicas actually requires different poll durations, and if so what specifically that reason is
2. Whether the e2e failures you are seeing are existing flakes on `main`, and if so what we should separately do to resolve those.

My instinct is that using multiple replicas should not affect timing, and if it does, it is a bug in the code, not in the test.

joelanford:
> However, we will still need changes to test/e2e to validate that we can handle multiple replicas.

If there are specific assumptions in the e2e tests that there is one replica and tests fail when there are two replicas, yes we should resolve that and remove those assumptions. But I would reiterate that I don't think anything about timing or functional test expectations themselves should be changing in this PR.

helm/olmv1/templates/deployment-olmv1-system-catalogd-controller-manager.yml
tmshort: However, I will point out that this may cause an issue on our  single-node `kind` experimental-e2e tests where we have two replicas (such that we are validating that two replicas does not cause issues with the e2e tests).

test/upgrade-e2e/post_upgrade_test.go
```diff
@@ -190,14 +210,13 @@ func TestClusterExtensionAfterOLMUpgrade(t *testing.T) {
...
+	}, 10*time.Minute, pollInterval)
 }
 
 // waitForDeployment checks that the updated deployment with the given app.kubernetes.io/name label
 // has reached the desired number of replicas and that the number pods matches that number
-// i.e. no old pods remain. It will return a pointer to the first pod. This is only necessary
-// to facilitate the mitigation put in place for https://github.com/operator-framework/operator-controller/issues/1626
-func waitForDeployment(t *testing.T, ctx context.Context, controlPlaneLabel string) *corev1.Pod {
+// i.e. no old pods remain.
+func waitForDeployment(t *testing.T, ctx context.Context, controlPlaneLabel string) {
```
tmshort: This PR: https://github.com/operator-framework/operator-controller/pull/2379
Handles the changes for waitForDeployment, which are independent of these changes, and can be merged quicker because it only deals with multiple replicas in the

jianzhangbjz:
> My instinct is that using multiple replicas should not affect timing, and if it does, it is a bug in the code, not in the test.

Those `timeout` failures not only happened in this PR, but others. If you rerun it for times, it will pass. However, there is no `retest` mechanism as the downstream does. I tried increase the timeout to fix it and make this PR test pass.

test/e2e/cluster_extension_install_test.go
jianzhangbjz: > why can't we make sure the tests continue running within that time budget?

The image pulling costing time is not the same. You will find the `timeout` failure is a general issue if you check the test history of this PR or other PRs.

jianzhangbjz:
Hi @joelanford , could you help point out how to solve this general failure(https://github.com/operator-framework/operator-controller/pull/2371#issuecomment-3635372643) if we don't increase the timeout? Just `/retest`? Thanks!

helm/olmv1/templates/deployment-olmv1-system-catalogd-controller-manager.yml
jianzhangbjz: Good point! I added `podAntiAffinity` and used the `preferred` rule.  Besides, I created https://github.com/openshift/release/pull/72395 to add `SNO` upgrade test for the downstream OLMv1 and OLMv0, please take a look, thanks!
```yaml
  podAntiAffinity:
    preferredDuringSchedulingIgnoredDuringExecution:
      - weight: 100
        podAffinityTerm:
          labelSelector:
            matchExpressions:
              - key: control-plane
                operator: In
                values:
                  - operator-controller-controller-manager
                  - catalogd-controller-manager
          topologyKey: kubernetes.io/hostname
```

jianzhangbjz:
Encounter this kind of failure again. I'd prefer to set `-timeout=20m` for the current `go test -count=1 -v  ./test/e2e/...`. However, given your sensitivity to the timeout duration, I’ll leave this issue here for you to decide.
```go
=== RUN   TestClusterExtensionRecoversFromExistingDeploymentWhenFailureFixed
    cluster_extension_install_test.go:686: When a cluster extension is installed from a catalog
    cluster_extension_install_test.go:687: When the extension bundle format is registry+v1
    helpers.go:271: Ensuring ClusterCatalog has Status.Condition of Progressing with a status == True and reason == Succeeded
    helpers.go:281: Checking that catalog has the expected metadata label
    helpers.go:286: Ensuring ClusterCatalog has Status.Condition of Type = Serving with status == True
    cluster_extension_install_test.go:710: By creating a new Deployment that can not be adopted
    cluster_extension_install_test.go:753: It resolves the specified package with correct bundle path
    cluster_extension_install_test.go:754: By creating the ClusterExtension resource
    cluster_extension_install_test.go:757: By eventually reporting Progressing == True with Reason Retrying
    cluster_extension_install_test.go:766: By eventually failing to install the package successfully due to no adoption support
    cluster_extension_install_test.go:779: By deleting the new Deployment
    cluster_extension_install_test.go:786: By eventually installing the package successfully
    cluster_extension_install_test.go:798: By eventually reporting Progressing == True with Reason Success
    helpers.go:333: By deleting ClusterCatalog "test-catalog-jmcpp5dt"
    helpers.go:342: By deleting ClusterExtension "clusterextension-2fx4dcfk"
    helpers.go:306: By waiting for CustomResourceDefinitions of "clusterextension-2fx4dcfk" to be deleted
    helpers.go:314: By waiting for ClusterRoleBindings of "clusterextension-2fx4dcfk" to be deleted
    helpers.go:322: By waiting for ClusterRoles of "clusterextension-2fx4dcfk" to be deleted
    helpers.go:352: By deleting ServiceAccount "clusterextension-2fx4dcfk"
    helpers.go:361: By deleting Namespace "clusterextension-2fx4dcfk"
panic: test timed out after 10m0s
	running tests:
		TestClusterExtensionRecoversFromExistingDeploymentWhenFailureFixed (20s)

goroutine 2094 [running]:
testing.(*M).startAlarm.func1()
	/opt/hostedtoolcache/go/1.24.6/x64/src/testing/testing.go:2484 +0x394
```

test/upgrade-e2e/post_upgrade_test.go
```diff
@@ -217,13 +237,82 @@ func waitForDeployment(t *testing.T, ctx context.Context, controlPlaneLabel stri
...
+	}
+
+	leaseName, ok := leaseNameMap[controlPlaneLabel]
+	if !ok {
+		return nil, fmt.Errorf("unknown control plane label: %s", controlPlaneLabel)
+	}
+
+	var leaderPod *corev1.Pod
+
+	// Use require.Eventually for polling instead of custom ticker implementation
```
pedjak: the comment states that we use it on the next line, but actually don't. Also, not sure why are mentioning custom ticker implementation, given that it is not a part of the existing code base.

test/upgrade-e2e/post_upgrade_test.go
```diff
@@ -217,13 +237,82 @@ func waitForDeployment(t *testing.T, ctx context.Context, controlPlaneLabel stri
...
+			// No leader elected yet, retry
+			return false, nil
+		}
+
+		// The HolderIdentity is typically in the format "pod-name_hash"
+		holderIdentity := *lease.Spec.HolderIdentity
+
+		// Now fetch pods and find the one matching the lease holder
+		var managerPods corev1.PodList
+		if err := c.List(ctx, &managerPods, client.MatchingLabelsSelector{Selector: deploymentLabelSelector}); err != nil {
```
pedjak: no need to list pods. After finding out the holder identity, we can directly fetch the particular pod with the given name.

tmshort:
> Hi @joelanford , could you help point out how to solve this general failure([#2371 (comment)](https://github.com/operator-framework/operator-controller/pull/2371#issuecomment-3635372643)) if we don't increase the timeout? Just `/retest`? Thanks!

You need to use standard GitHub mechanisms to rerun tests: clock on the failed test, and then there is a "Re-run Jobs" button in the upper right.

But... we shouldn't really fail that often.

tmshort:
This should be rebased given the basic changes to support replicas in the tests.

jianzhangbjz:
> You need to use standard GitHub mechanisms to rerun tests: clock on the failed test, and then there is a "Re-run Jobs" button in the upper right.

I don't find that "Re-run Jobs" button. The permission issue?
image

jianzhangbjz:
According to [Re-running workflows and jobs](https://docs.github.com/en/actions/how-tos/manage-workflow-runs/re-run-workflows-and-jobs?tool=webui): People with **write permissions to a repository** can re-run workflows in the repository. I believe this is the reason why I cannot find the `re-run` button.

test/helpers/helpers.go
```diff
@@ -283,37 +291,41 @@ func ValidateCatalogUnpackWithName(t *testing.T, catalogName string) {
...
+	}, catalogPollDuration, pollInterval)
 }
 
 func EnsureNoExtensionResources(t *testing.T, clusterExtensionName string) {
 	ls := labels.Set{"olm.operatorframework.io/owner-name": clusterExtensionName}
 
-	// CRDs may take an extra long time to be deleted, and may run into the following error:
+	// Use 2 minute timeout for cleanup operations to ensure they complete within the test timeout.
+	// This is shorter than pollDuration (3 min) to leave buffer for the overall test suite.
+	cleanupTimeout := 2 * time.Minute
```
tmshort: This is actually a _shorter_ timeout. It was originally `5*pollDuration` which is equal to `5 * time.Minute`.
Do we need this change?

tmshort:
@joelanford are you still looking for minimization of the diffs for polling timeouts?

tmshort:
Can this be rebased onto the tip of main. The base for this PR is ~2 months old.

jianzhangbjz:
It was rebased 3 days ago.
image

pedjak (approved):

PR #2374, opened 2025-12-04
:warning: Remove Migrated reason from ClusterExtensionRevision Available condition

perdasilva:
# Description

Removes the Migrated reason from the ClusterExtensionRevision Available condition. It would be short-lived and it makes more sense to let the revision controller set the conditions. If there is need to call out migration, we could set an annotation on the revision resource.

## Reviewer Checklist

- [ ] API Go Documentation


pedjak (approved):
/lgtm

PR #2379, opened 2025-12-10
:seedling: Don't assume there is only one replica in e2e tests

tmshort:
For the upgrade e2e tests, don't assume there is onle one replica. Get the number of replicas from the deployment and wait for the deployment to have that many available. Use the lease to determine the leader pod and reference that.

Note that the name format of leases for operator-controller and catalogd are quite different; this doesn't change that, as it may have an impact on the upgrade test itself.

Assisted-by: Claude Code

# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/test/e2e/metrics_test.go
+++ b/test/e2e/metrics_test.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"io"
 	"os/exec"
+	"strings"
 	"testing"
 	"time"
 
@@ -33,16 +34,17 @@ func TestOperatorControllerMetricsExportedEndpoint(t *testing.T) {
 	client := utils.FindK8sClient(t)
 	curlNamespace := createRandomNamespace(t, client)
 	componentNamespace := getComponentNamespace(t, client, "control-plane=operator-controller-controller-manager")
-	metricsURL := fmt.Sprintf("https://operator-controller-service.%s.svc.cluster.local:8443/metrics", componentNamespace)
 
 	config := NewMetricsTestConfig(
 		client,
 		curlNamespace,
+		componentNamespace,
 		"operator-controller-metrics-reader",
 		"operator-controller-metrics-binding",
 		"operator-controller-metrics-reader",
 		"oper-curl-metrics",
-		metricsURL,
+		"app.kubernetes.io/name=operator-controller",
+		operatorControllerMetricsPort,
 	)
 
 	config.run(t)
@@ -53,16 +55,17 @@ func TestCatalogdMetricsExportedEndpoint(t *testing.T) {
 	client := utils.FindK8sClient(t)
 	curlNamespace := createRandomNamespace(t, client)
 	componentNamespace := getComponentNamespace(t, client, "control-plane=catalogd-controller-manager")
-	metricsURL := fmt.Sprintf("https://catalogd-service.%s.svc.cluster.local:7443/metrics", componentNamespace)
 
 	config := NewMetricsTestConfig(
 		client,
 		curlNamespace,
+		componentNamespace,
 		"catalogd-metrics-reader",
 		"catalogd-metrics-binding",
 		"catalogd-metrics-reader",
 		"catalogd-curl-metrics",
-		metricsURL,
+		"app.kubernetes.io/name=catalogd",
+		catalogdMetricsPort,
 	)
 
 	config.run(t)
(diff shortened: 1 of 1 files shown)


test/e2e/metrics_test.go
```diff
@@ -33,16 +33,17 @@ func TestOperatorControllerMetricsExportedEndpoint(t *testing.T) {
...
 		client,
 		curlNamespace,
+		componentNamespace,
 		"operator-controller-metrics-reader",
 		"operator-controller-metrics-binding",
 		"operator-controller-metrics-reader",
 		"oper-curl-metrics",
-		metricsURL,
+		"app.kubernetes.io/name=operator-controller",
+		8443,
```
perdasilva: should we move the ports to consts to improve readability? e.g.

```
const (
  operatorControllerMetricsPort = 8443
  catalogdMetricsPort = 7443
)
```

or something like that?

test/upgrade-e2e/post_upgrade_test.go
```diff
@@ -215,15 +217,51 @@ func waitForDeployment(t *testing.T, ctx context.Context, controlPlaneLabel stri
...
 	t.Logf("Ensure the number of remaining pods equal the desired number of replicas (%d)", desiredNumReplicas)
 	require.EventuallyWithT(t, func(ct *assert.CollectT) {
 		require.NoError(ct, c.List(ctx, &managerPods, client.MatchingLabelsSelector{Selector: deploymentLabelSelector}))
-		require.Len(ct, managerPods.Items, 1)
+		require.Len(ct, managerPods.Items, int(desiredNumReplicas))
 	}, time.Minute, time.Second)
-	return &managerPods.Items[0]
+
+	// Find the leader pod by checking the lease
+	t.Log("Finding the leader pod")
```
perdasilva, line 243: I wonder if we're carrying too much complexity here. It seems we're returning the pod for two reasons:

1. To at some point check the logs for some specific log lines to do with leader election, e.g.

```
t.Log("Waiting for acquired leader election")
	leaderCtx, leaderCancel := context.WithTimeout(ctx, 3*time.Minute)
	defer leaderCancel()
	leaderSubstrings := []string{"successfully acquired lease"}
	leaderElected, err := watchPodLogsForSubstring(leaderCtx, &managerPod, leaderSubstrings...)
	require.NoError(t, err)
	require.True(t, leaderElected)

t.Log("Reading logs to make sure that ClusterCatalog was reconciled by catalogdv1")
	logCtx, cancel := context.WithTimeout(ctx, time.Minute)
	defer cancel()
	substrings := []string{
		"reconcile ending",
		fmt.Sprintf(`ClusterCatalog=%q`, testClusterCatalogName),
	}
	found, err := watchPodLogsForSubstring(logCtx, &managerPod, substrings...)
	require.NoError(t, err)
	require.True(t, found)
```

2. Because there's a possible bug in catalogd where after a pod restart, the ClusterCatalog still reports that it is serving the catalog. This causes some flakes by making the e2e test progress quickly to the Eventually waiting for installation, which times out because its also implicitly waiting for the catalog to unpack.

I wonder if we could refactor this helper to wait for the Available condition to be True and .status.replicas == .status.updatedReplicas, or something like that, and drop the return value. 

I question whether we need 1 at all and we could try to tackle 2. in a different way, e.g. by hitting the catalog service endpoint until it doesn't return a 404. wdyt?

test/e2e/metrics_test.go
```diff
@@ -154,19 +160,30 @@ func (c *MetricsTestConfig) createCurlMetricsPod(t *testing.T) {
...
 	require.NoError(t, waitErr, "Error waiting for curl pod to be ready: %s", string(waitOutput))
 
-	t.Log("Validating the metrics endpoint")
-	curlCmd := exec.Command(c.client, "exec", c.curlPodName, "--namespace", c.namespace, "--",
-		"curl", "-v", "-k", "-H", "Authorization: Bearer "+token, c.metricsURL)
-	output, err := curlCmd.CombinedOutput()
-	require.NoError(t, err, "Error calling metrics endpoint: %s", string(output))
-	require.Contains(t, string(output), "200 OK", "Metrics endpoint did not return 200 OK")
+	// Get all pod IPs for the component
+	podIPs := c.getComponentPodIPs(t)
```
pedjak, line 172: we could send requests to pod using their FQDN, no need to find out their IPs.

test/e2e/metrics_test.go
tmshort: I kinda wish they were the same value!

test/upgrade-e2e/post_upgrade_test.go
tmshort: I wonder if changing the test _that_ much is in scope here. This is preserving the test as-is, but making sure the returned pod is correct.

test/e2e/metrics_test.go
tmshort: But yeah, it looks as though they are already defined.

test/e2e/metrics_test.go
tmshort: But it appears that the FQDN uses dashed IP addresses, the pods don't have a hostname/subdomain set.
EDIT: https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/

test/upgrade-e2e/post_upgrade_test.go
tmshort: >I wonder if we could refactor this helper to wait for the Available condition to be True and .status.replicas == .status.updatedReplicas, or something like that, and drop the return value.

We are already checking updatedReplicas, replicas, availableReplicas and readyReplicas:
https://github.com/operator-framework/operator-controller/blob/0240528e154444fa8962530292108d25be94b7d7/test/upgrade-e2e/post_upgrade_test.go#L211-L217

Do we need to do more than that?

Which Available condition are you referring to? Deployment? ClusterCatalog?

perdasilva: On the Deployment

perdasilva: > I wonder if changing the test that much is in scope here. This is preserving the test as-is, but making sure the returned pod is correct.

That's fair. I just don't know that we want to carry so much complexity and it might be worth questioning the need for the return value and .zip things. I think it's ok to move it over to another PR (or maybe the cucumber tests might already cover this). But, I thought I'd call it out.

tmshort: The cucumber stuff will likely rewrite some of that, so I'd say defer it to that, but I _can_ add a quick check for the deployment Available status, since we're already checking things on the deployment.

test/e2e/metrics_test.go
pedjak: > But it appears that the FQDN uses dashed IP addresses, the pods don't have a hostname/subdomain set. EDIT: https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/

Ah, you are absolutely right.

nit: you can read all IP addresses for a service (i.e. all pod IPs) by fetching [Endpoint resource](https://kubernetes.io/docs/reference/kubernetes-api/service-resources/endpoints-v1/)  (the name is the same as of service).

pedjak (approved):
/lgtm

tmshort:
It appears there are Github API issues causing some of our test failures.

PR #2382, opened 2025-12-11
:seedling: Change the experimental-e2e to use a two-node kind cluster

tmshort:
In order to fully support multiple replicas, we need to be able to run our tests in a multi-node environment, but we don't _always_ want to run in a multi-node test environment.

Update the experimental-e2e to run a 2-node kind cluster (both control- plane).

Update the `kind-load` recipe to directly load a docker image via `load docker-image`, since `load image-archive` has issues with multi-node clusters.

# Description

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/kind-config-experimental.yaml
+++ b/kind-config-experimental.yaml
@@ -1,0 +1,45 @@
+apiVersion: kind.x-k8s.io/v1alpha4
+kind: Cluster
+nodes:
+  - role: control-plane
+    extraPortMappings:
+    # e2e image registry service's NodePort
+      - containerPort: 30000
+        hostPort: 30000
+        listenAddress: "127.0.0.1"
+        protocol: tcp
+    # prometheus metrics service's NodePort
+      - containerPort: 30900
+        hostPort: 30900
+        listenAddress: "127.0.0.1"
+        protocol: tcp
+    kubeadmConfigPatches:
+      - |
+        kind: ClusterConfiguration
+        apiServer:
+            extraArgs:
+              enable-admission-plugins: OwnerReferencesPermissionEnforcement
+      - |
+        kind: InitConfiguration
+        nodeRegistration:
+          kubeletExtraArgs:
+            node-labels: "ingress-ready=true"
+          taints: []
+    extraMounts:
+      - hostPath: ./hack/kind-config/containerd/certs.d
+        containerPath: /etc/containerd/certs.d
+  - role: control-plane
+    kubeadmConfigPatches:
+      - |
+        kind: JoinConfiguration
+        nodeRegistration:
+          kubeletExtraArgs:
+            node-labels: "ingress-ready=true"
+          taints: []
+    extraMounts:
+      - hostPath: ./hack/kind-config/containerd/certs.d
+        containerPath: /etc/containerd/certs.d
+containerdConfigPatches:
+  - |-
+    [plugins."io.containerd.grpc.v1.cri".registry]
+      config_path = "/etc/containerd/certs.d"
--- a/Makefile
+++ b/Makefile
@@ -62,6 +62,10 @@ ifeq ($(origin KIND_CLUSTER_NAME), undefined)
 KIND_CLUSTER_NAME := operator-controller
 endif
 
+ifeq ($(origin KIND_CONFIG), undefined)
+KIND_CONFIG := ./kind-config.yaml
+endif
+
 
 ifneq (, $(shell command -v docker 2>/dev/null))
 CONTAINER_RUNTIME := docker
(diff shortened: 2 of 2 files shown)


kind-config-experimental.yaml
```diff
@@ -0,0 +1,45 @@
+apiVersion: kind.x-k8s.io/v1alpha4
```
pedjak, line 1: nit: `kind-config-experimental.yaml` sounds to me like some sort of experimental kind config, but actually it is just a two-node config. Perhaps we could reflect that in the filename?

BTW, does this file need to be in the project root?

kind-config-experimental.yaml
```diff
@@ -0,0 +1,45 @@
...
+      - |
+        kind: InitConfiguration
+        nodeRegistration:
+          kubeletExtraArgs:
+            node-labels: "ingress-ready=true"
+          taints: []
+    extraMounts:
+      - hostPath: ./hack/kind-config/containerd/certs.d
+        containerPath: /etc/containerd/certs.d
+  - role: control-plane
```
pedjak, line 31: we do not need another control-plane node, a worker node could be just fine. Furthermore,  each control plane node runs a copy of control plane pods (etcd, apiserver, scheduler, controller-manager) doubling the resources consumed by kind. Additional worker node keeps the resource usage on previous level.

kind-config-experimental.yaml
tmshort: It's the same location as the other kind config; keep similar files together.

joelanford:
> Update the kind-load recipe to directly load a docker image via load docker-image, since load image-archive has issues with multi-node clusters.

The reason we use the archive variant for image loading is to support contributors that use podman instead of docker. Will this change break their workflows?

kind-config-experimental.yaml
joelanford: > does this file need to be in the project root

Now that there are multiple kind config files at play, it makes sense to me to bury this in a subdirectory somewhere.

kind-config-experimental.yaml
tmshort: This is on purpose,. operator-controller and catalogd are supposed to run only on control-plane nodes; not worker nodes. It's part of what we want to test. This is also only for the experimental-e2e, so it's not used for normal run-time.

tmshort:
> The reason we use the archive variant for image loading is to support contributors that use podman instead of docker. Will this change break their workflows?

I know, but the `load image-archive` command doesn't work with multi-node **kind** (at least it wasn't in my setup).
I have **docker** and **podman** installed, and it works. But you're right, it fails with just **podman**, despite "enabling experimental podman provider":
```
/home/tshort/go/bin/kind-v0.30.0 load docker-image quay.io/operator-framework/operator-controller:devel --name operator-controller-e2e
enabling experimental podman provider
ERROR: image: "quay.io/operator-framework/operator-controller:devel" not present locally
make: *** [Makefile:390: kind-load] Error 1
tshort@cube:~/git/operator-framework/operator-controller (multi-node-kind >)$ podman images | grep operator-controller
quay.io/operator-framework/operator-controller  devel                                 18e72de2e0e1  7 minutes ago  83.4 MB
```

Issues:  https://github.com/kubernetes-sigs/kind/issues/3105 (closed and dup'd to: https://github.com/kubernetes-sigs/kind/issues/2038)

kind-config-experimental.yaml
tmshort: That's fair, and easy.

tmshort:
> The reason we use the archive variant for image loading is to support contributors that use podman instead of docker. Will this change break their workflows?

And now I'm trying without my changes, and it's failing with just **podman**... So, it looks as though we're already broken...
```
/home/tshort/go/bin/kind-v0.30.0 export kubeconfig --name operator-controller-e2e
enabling experimental podman provider
Set kubectl context to "kind-operator-controller-e2e"
podman save quay.io/operator-framework/operator-controller:devel | /home/tshort/go/bin/kind-v0.30.0 load image-archive /dev/stdin --name operator-controller-e2e
enabling experimental podman provider
ERROR: failed to load image: command "podman exec --privileged -i operator-controller-e2e-control-plane ctr --namespace=k8s.io images import --all-platforms --digests --snapshotter=fuse-overlayfs -" failed with error: exit status 1
Command Output: ctr: unable to initialize unpacker: no unpack platforms defined: invalid argument
make: *** [Makefile:385: kind-load] Error 1

kind-config-experimental.yaml
pedjak: Do we assert it during e2e tests on what nodes we run?

pedjak (approved):
/lgtm

kind-config-experimental.yaml
tmshort: Not yet, that was something that #2371 will be trying to do when they add 2 replicas.

PR #2543, opened 2026-03-06
✨  Add support for parsing and persisting explicit pkg.Release field

rashmigottipati:
# Description

## Summary

Adds support for parsing the explicit `pkg.Release` field from bundle metadata and persisting it through cluster storage (roundtripping).  This PR is gated behind the `BundleReleaseSupport` feature gate, which is disabled by default. The feature gate is explicitly enabled in experimental manifests to facilitate testing of experimental features.

## Changes

- Added `BundleReleaseSupport` alpha feature gate (disabled by default)
 - Added `Release` field to `BundleMetadata` CRD type
 - Added `BundleReleaseKey` label constant for annotation storage
 - Updated `bundleutil.parseVersionRelease()` to: Parse explicit `pkg.Release` when feature gate enabled and field present. Fall back to legacy registry+v1 behavior (release in build metadata) otherwise
 - Updated `bundleutil.MetadataFor()` to serialize release value from `VersionRelease`
-  Added roundtripping support: store release in CRD field and annotations, read it back in filters and controllers

 #### Feature Gate Behavior

  **When enabled:**
  - Bundles with explicit `pkg.Release` field: release parsed separately, build metadata preserved for proper semver purpose
  - Bundles without explicit release: falls back to registry+v1 parsing (release from build metadata)

  **When feature gate disabled:**
  - All bundles use registry+v1 parsing (release from build metadata)
  - Explicit `pkg.Release` field ignored
  
## Reviewer Checklist

- [x] API Go Documentation
- [x] Tests: Unit Tests 

Link to Github Issue: https://github.com/operator-framework/operator-controller/issues/2495
Epic: https://github.com/operator-framework/operator-controller/issues/2479


internal/operator-controller/catalogmetadata/filter/successors_test.go
```diff
@@ -240,3 +240,36 @@ func TestLegacySuccessor(t *testing.T) {
 	assert.True(t, f(b5))
 	assert.False(t, f(emptyBundle))
 }
+
+// TestSuccessorsOf_WithReleaseVersionSupport_FeatureGateDisabled verifies higher releases
+// are NOT successors when ReleaseVersionPriority gate is disabled (default).
+// TODO: Feature gate enabled behavior must be tested in E2E tests.
+func TestSuccessorsOf_WithReleaseVersionSupport_FeatureGateDisabled(t *testing.T) {
```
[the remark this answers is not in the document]
rashmigottipati: Fixed

internal/operator-controller/catalogmetadata/filter/successors.go
```diff
@@ -28,11 +29,22 @@ func SuccessorsOf(installedBundle ocv1.BundleMetadata, channels ...declcfg.Chann
...
 		ExactVersionRelease(*installedVersionRelease),
-	), nil
+	}
+
+	// If release version priority is enabled, also consider bundles
+	// with the same semantic version but higher release as valid successors.
+	// This allows upgrades to re-released bundles (e.g., 2.0.0+1 -> 2.0.0+2).
+	if features.OperatorControllerFeatureGate.Enabled(features.ReleaseVersionPriority) {
+		predicates = append(predicates, SameVersionHigherRelease(*installedVersionRelease))
+	}
```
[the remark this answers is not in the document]
rashmigottipati: Done

joelanford:
If I understand this correctly, it looks like this introduces the new behavior that an explicit upgrade edge doesn't actually have to exist in the catalog to upgrade to a bundle that has the same version and a higher release. Is that the intent? (If so let's make that clear in the PR description).

Is it _also_ the intent that we'll inherit the upgrade edges of the first release of the version?

internal/operator-controller/catalogmetadata/filter/bundle_predicates_test.go
```diff
@@ -68,3 +69,88 @@ func TestInAnyChannel(t *testing.T) {
 	assert.False(t, fStable(b2))
 	assert.False(t, fStable(b3))
 }
+
+func TestSameVersionHigherRelease(t *testing.T) {
+	const testPackageName = "test-package"
+
+	// Expected bundle version 2.0.0+1
+	expect, err := bundle.NewLegacyRegistryV1VersionRelease("2.0.0+1")
```
joelanford: I'd suggest putting the `expect` value in the test case, and adding cases that start from different values. For example: from a version without a release.

Also related: do we need to add code that knows how to parse the new `Release` field of the `olm.package` property (around here: https://github.com/operator-framework/operator-controller/blob/dee8ed5ec2d15cd19b152315e347fda38036fd94/internal/operator-controller/bundleutil/bundle.go#L28)?

joelanford: iirc, @grokspawn's design called for a specific precedence when handling both `release` and build metadata that we will need to account for.

internal/operator-controller/catalogmetadata/filter/successors.go
```diff
@@ -28,11 +29,22 @@ func SuccessorsOf(installedBundle ocv1.BundleMetadata, channels ...declcfg.Chann
...
+	// Build the final predicate based on feature gate status
+	predicates := []filter.Predicate[declcfg.Bundle]{
 		successorsPredicate,
 		ExactVersionRelease(*installedVersionRelease),
-	), nil
+	}
+
+	// If release version priority is enabled, also consider bundles
+	// with the same semantic version but higher release as valid successors.
+	// This allows upgrades to re-released bundles (e.g., 2.0.0+1 -> 2.0.0+2).
```
joelanford: What if the re-release version is already in the explicit upgrade graph? In that case, should we ignore it and let its explicit channel entry take precedence? That would preserve existing behavior.

internal/operator-controller/catalogmetadata/filter/bundle_predicates.go
```diff
@@ -48,3 +48,21 @@ func InAnyChannel(channels ...declcfg.Channel) filter.Predicate[declcfg.Bundle]
...
 	}
 }
+
+// SameVersionHigherRelease returns a predicate that matches bundles with the same
+// semantic version as the provided version-release, but with a higher release value.
+// This is used to identify re-released bundles (e.g., 2.0.0+2 when 2.0.0+1 is installed).
+func SameVersionHigherRelease(expect bundle.VersionRelease) filter.Predicate[declcfg.Bundle] {
+	return func(b declcfg.Bundle) bool {
+		actual, err := bundleutil.GetVersionAndRelease(b)
+		if err != nil {
```
perdasilva: could failing silently here lead to some hard to spot bugs?

grokspawn:
~~We discussed earlier today.  I think the first step is to essentially revert https://github.com/operator-framework/operator-controller/pull/2273 so that we're not attempting to interpret build metadata as release info anymore.  Any catalogs built with tooling that supports release versioning will automagically propagate the build metadata into the CSV's `spec.release` field anyway, so regardless of whether the operator author used the old build metadata approach _or_ the new spec.release approach, they would still have spec.release here.~~

~~The next step is to ensure that we're sorting successors using the [CompositeVersion](https://github.com/operator-framework/operator-registry/blob/41e3619ea065e1b84a78a17385c87f20df67eb2e/alpha/declcfg/declcfg.go#L211), which is easily performed if using the [Bundle comparator](https://github.com/operator-framework/operator-registry/blob/41e3619ea065e1b84a78a17385c87f20df67eb2e/alpha/declcfg/declcfg.go#L233) during the sort.~~  

~~This will prefer version+release over version.~~


#2273 makes the claim that _future bundle types_ will drop this approach, not that registry+v1 bundles should.  So that's my bad.  
From the perspective of fulfilling handling precedents for registry+v1 bundles, we need to retain the ability to differentiate/order them by build metadata. :(

grokspawn:
> If I understand this correctly, it looks like this introduces the new behavior that an explicit upgrade edge doesn't actually have to exist in the catalog to upgrade to a bundle that has the same version and a higher release. Is that the intent? (If so let's make that clear in the PR description).
> 
> Is it _also_ the intent that we'll inherit the upgrade edges of the first release of the version?

Sadly, we have to rely on the presence of existing graph edges or we break assumptions users have with the registry+v1 bundle format, coming from v0.  

So even though we have the ability to prefer version+release over version, since replaces/skips use named bundles instead of bundle versions we have to support the older behavior.

joelanford:
> I think the first step is to essentially revert https://github.com/operator-framework/operator-controller/pull/2273 so that we're not attempting to interpret build metadata as release info anymore.

That would be a regression of a bug fix that #2273 made though, right?

> Sadly, we have to rely on the presence of existing graph edges or we break assumptions users have with the registry+v1 bundle format, coming from v0.
>
> So even though we have the ability to prefer version+release over version, since replaces/skips use named bundles instead of bundle versions we have to support the older behavior.

Yeah, agreed. I reached the same conclusion after thinking about this more.

So if I understand correctly now:
1. The existing code on `main` already honors the upgrade graph and orders "same version, different release" correctly when parsing from build metadata
2. The existing code on `main` knows nothing about the new release field.

So the change we need now is "look for `release` in the `olm.package` property, and if present, prefer it over build metadata"?

grokspawn:
Adjusted my comments.  I missed the part of the #2273 description when it said that we wouldn't divert from semver ordering for any new bundle formats.  
From the perspective of registry+v1 bundles, this is precedent and we should support the new case with .spec.release is available, otherwise falling back on the 'is it bundle metadata' ordering logic.

rashmigottipati:
@grokspawn @joelanford  Updated based on your feedback.

PR now implements the hybrid approach: uses `Bundle.Compare()` to read explicit `pkg.Release` from olm.package, falling back to build metadata when comparison returns equal. This supports both registry+v1 bundles and new bundle formats.

Added feature gate `CompositeVersionComparison` (alpha, disabled by default) for backward compatibility.

Ready for re-review when you have a chance. Thanks!

rashmigottipati:
The failing sanity check is due to an external dependency issue with Mozilla's TLS profiles API, not related to the code changes in this PR.

grokspawn:
We would also want to enable the new featuregate in the experimental [release](https://github.com/operator-framework/operator-controller/blob/main/helm/experimental.yaml#L8-L17), right?

internal/operator-controller/catalogmetadata/filter/bundle_predicates_test.go
rashmigottipati: @joelanford addressed the test structure based on your suggestion. `expect` is now parameterized in test cases with coverage for different starting values (with/without release, different base versions).

Also updated the PR that matches @grokspawn's design which explicitly checks for `pkg.Release` first, with fallback to build metadata only if empty.

rashmigottipati:
@grokspawn  Done. Updated the PR to enable the feature gate in experimental release.

grokspawn:
@rashmigottipati would you please resolve the copilot comments in this review?  I'm trying to not duplicate guidance with them, and it's challenging to know what is still current.

What I have been doing is:
- thumbs-up and resolve, for good feedback that I updated
- thumbs-down and resolve, for clear hallucinations or obvious craziness
- thumbs-down and DON'T resolve, for a disagreement I cannot dismiss (and we can talk more about). 

WDYT?

internal/operator-controller/features/features.go
```diff
@@ -19,6 +19,7 @@ const (
 	HelmChartSupport                  featuregate.Feature = "HelmChartSupport"
 	BoxcutterRuntime                  featuregate.Feature = "BoxcutterRuntime"
 	DeploymentConfig                  featuregate.Feature = "DeploymentConfig"
+	BundleReleaseSupport              featuregate.Feature = "BundleReleaseSupport"
 )
```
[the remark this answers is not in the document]
rashmigottipati: Aligned the PR description that matches the changes in the PR including renaming of the feature gate

internal/operator-controller/catalogmetadata/filter/bundle_predicates_test.go
```diff
@@ -68,3 +70,126 @@ func TestInAnyChannel(t *testing.T) {
...
+			Name:    "test-package.v2.0.0-r2",
+			Package: testPackageName,
+			Properties: []property.Property{
+				{Type: property.TypePackage, Value: json.RawMessage(`{"packageName": "test-package", "version": "2.0.0", "release": "2"}`)},
+			},
+		}
+
+		f := filter.SameVersionHigherRelease(*expect)
+		assert.True(t, f(bundleWithExplicitRelease), "explicit release=2 should be higher than 2.0.0+1")
+	})
```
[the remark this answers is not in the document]
rashmigottipati: Fixed. Test now explicitly enables BundleReleaseSupport=true with t.Cleanup restoration instead of skipping.

internal/operator-controller/bundleutil/bundle_test.go
```diff
@@ -83,12 +84,113 @@ func TestGetVersionAndRelease(t *testing.T) {
...
+
+		// Should parse build metadata (+2), not explicit release field (999)
+		expected := &bundle.VersionRelease{
+			Version: bsemver.MustParse("1.0.0"),
+			Release: bundle.Release([]bsemver.PRVersion{
+				{VersionNum: 2, IsNum: true},
+			}),
+		}
+		require.Equal(t, expected, actual)
+	}
```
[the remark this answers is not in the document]
rashmigottipati: This is resolved. Split into two subtests that explicitly set BundleReleaseSupport=true/false with t.Cleanup restoration. Both code paths now execute deterministically in CI.

internal/operator-controller/bundleutil/bundle.go
```diff
@@ -16,25 +16,42 @@ import (
...
+		}
+		rel, err := bundle.NewRelease(pkg.Release)
+		if err != nil {
+			return nil, fmt.Errorf("error parsing release %q: %w", pkg.Release, err)
+		}
+		return &bundle.VersionRelease{
+			Version: vers,
+			Release: rel,
+		}, nil
+	}
```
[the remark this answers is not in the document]
rashmigottipati: I've addressed this. Added feature gate check at line 33: explicit pkg.Release is only parsed when BundleReleaseSupport feature gate is enabled. When disabled, falls back to legacy build metadata parsing.

internal/operator-controller/catalogmetadata/compare/compare.go
```diff
@@ -39,9 +40,22 @@ func newMastermindsRange(versionRange string) (bsemver.Range, error) {
 }
 
 // ByVersionAndRelease is a comparison function that compares bundles by
-// version and release. Bundles with lower versions/releases are
-// considered less than bundles with higher versions/releases.
+// version and release. When the CompositeVersionComparison feature gate is
+// enabled, it uses Bundle.Compare() (which reads pkg.Release from olm.package)
+// and falls back to build metadata comparison if equal. When disabled, it uses
+// version+release from build metadata (backward compatible).
```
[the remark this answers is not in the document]
rashmigottipati: This comment is resolved. Updated parseVersionRelease() to gate explicit pkg.Release parsing behind BundleReleaseSupport check. When disabled, only build metadata is parsed (legacy behavior). Doc comment now matches implementation.

internal/operator-controller/catalogmetadata/compare/compare_test.go
```diff
@@ -168,3 +169,50 @@ func TestByDeprecationFunc(t *testing.T) {
...
+		Properties: []property.Property{
+			{Type: property.TypePackage, Value: json.RawMessage(`{"packageName": "package1", "version": "1.0.0+2"}`)},
+		},
+	}
+
+	// Detect current feature gate state and test the appropriate path
+	if features.OperatorControllerFeatureGate.Enabled(features.CompositeVersionComparison) {
+		t.Log("CompositeVersionComparison enabled - testing hybrid comparison with build metadata fallback")
+		result := compare.ByVersionAndRelease(b1, b2)
+		assert.Positive(t, result, "Bundle.Compare() returns 0 for registry+v1, fallback to build metadata: 1.0.0+2 > 1.0.0+1")
```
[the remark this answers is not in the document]
rashmigottipati: Fixed. Split into two subtests with explicit gate control (true/false) and t.Cleanup restoration. Both comparison paths now execute in CI.

internal/operator-controller/catalogmetadata/filter/successors_test.go
```diff
@@ -240,3 +241,44 @@ func TestLegacySuccessor(t *testing.T) {
...
+	}
+
+	predicate, err := SuccessorsOf(installedBundle, channel)
+	require.NoError(t, err)
+
+	// Detect current feature gate state and test the appropriate behavior
+	if features.OperatorControllerFeatureGate.Enabled(features.CompositeVersionComparison) {
+		t.Log("CompositeVersionComparison enabled - same-version higher-release bundles are valid successors")
+		assert.True(t, predicate(higherRelease), "1.0.0+2 should be a successor of 1.0.0+1 when feature gate enabled")
+	} else {
```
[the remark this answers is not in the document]
rashmigottipati: Resolved this. Split into two subtests with explicit BundleReleaseSupport=true/false and t.Cleanup restoration.
SameVersionHigherRelease successor path now executes in CI

internal/operator-controller/catalogmetadata/compare/compare.go
```diff
@@ -39,9 +40,22 @@ func newMastermindsRange(versionRange string) (bsemver.Range, error) {
...
 func ByVersionAndRelease(b1, b2 declcfg.Bundle) int {
+	if features.OperatorControllerFeatureGate.Enabled(features.CompositeVersionComparison) {
+		// Use CompositeVersion comparison (reads pkg.Release from olm.package)
+		result := b2.Compare(&b1)
+		if result != 0 {
+			return result
+		}
+		// If CompositeVersion comparison is equal, fall back to build metadata
+		return compareByBuildMetadata(b1, b2)
+	}
```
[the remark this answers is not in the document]
rashmigottipati: if profiling shows this is actually a problem, we can optimize later. I see it more like a "nice to have" optimization but not worth the added complexity right now..

internal/operator-controller/catalogmetadata/compare/compare_test.go
```diff
@@ -161,10 +163,47 @@ func TestByDeprecationFunc(t *testing.T) {
...
+		require.NoError(t, features.OperatorControllerFeatureGate.Set(fmt.Sprintf("%s=false", features.CompositeVersionComparison)))
+		result := compare.ByVersionAndRelease(b1, b2)
+		assert.Positive(t, result, "should sort by build metadata: 1.0.0+2 > 1.0.0+1")
+	})
+
+	t.Run("feature gate enabled - registry+v1 bundles still work via build metadata fallback", func(t *testing.T) {
+		require.NoError(t, features.OperatorControllerFeatureGate.Set(fmt.Sprintf("%s=true", features.CompositeVersionComparison)))
+		result := compare.ByVersionAndRelease(b1, b2)
+		assert.Positive(t, result, "Bundle.Compare() returns 0, fallback to build metadata: 1.0.0+2 > 1.0.0+1")
+	})
```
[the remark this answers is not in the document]
rashmigottipati: Fixed. Added test case for explicit pkg.Release ordering `TestByVersionAndRelease_WithBundleReleaseSupport` (lines 202-217).

internal/operator-controller/catalogmetadata/filter/bundle_predicates.go
```diff
@@ -48,3 +48,21 @@ func InAnyChannel(channels ...declcfg.Channel) filter.Predicate[declcfg.Bundle]
...
+			return false
+		}
+
+		if expect.Version.Compare(actual.Version) != 0 {
+			return false
+		}
+
+		return expect.Release.Compare(actual.Release) < 0
+	}
```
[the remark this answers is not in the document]
rashmigottipati: Resolving this as `bundleutil.GetVersionAndRelease()` now parses explicit pkg.Release field when BundleReleaseSupport is enabled

internal/operator-controller/catalogmetadata/compare/compare.go
```diff
@@ -54,6 +68,18 @@ func ByVersionAndRelease(b1, b2 declcfg.Bundle) int {
 	return vr2.Compare(*vr1)
 }
 
+// compareByBuildMetadata compares bundles using build metadata parsing.
+// This is used as a fallback when CompositeVersion comparison returns equal.
+func compareByBuildMetadata(b1, b2 declcfg.Bundle) int {
+	vr1, err1 := bundleutil.GetVersionAndRelease(b1)
+	vr2, err2 := bundleutil.GetVersionAndRelease(b2)
+
```
pedjak: **Misleading name**: When the `BundleReleaseSupport` gate is enabled, this function calls `GetVersionAndRelease()`, which itself prefers the explicit `pkg.Release` field over build metadata (via `parseVersionRelease`). So this is not actually a "build metadata" fallback for bundles that have an explicit release — it will use the explicit release again, producing the same result as `Bundle.Compare()`.

It only truly falls back to build metadata for bundles with **no** explicit `pkg.Release` (registry+v1 bundles). Consider renaming to something like `compareByVersionRelease` to reflect what it actually does, or add a comment clarifying the actual fallback semantics.

internal/operator-controller/catalogmetadata/compare/compare.go
```diff
@@ -39,9 +40,22 @@ func newMastermindsRange(versionRange string) (bsemver.Range, error) {
...
+// enabled, it uses Bundle.Compare() (which reads pkg.Release from olm.package)
+// and falls back to build metadata comparison if equal. When disabled, it uses
+// version+release from build metadata (backward compatible).
 func ByVersionAndRelease(b1, b2 declcfg.Bundle) int {
+	if features.OperatorControllerFeatureGate.Enabled(features.BundleReleaseSupport) {
+		// Use CompositeVersion comparison (reads pkg.Release from olm.package)
+		result := b2.Compare(&b1)
+		if result != 0 {
+			return result
+		}
```
pedjak: **Silent error swallowing**: Upstream `Bundle.Compare()` returns `0` in three cases:
1. Bundle names are identical (correct short-circuit)
2. Versions and releases are equal (correct)
3. **Any** error parsing `CompositeVersion` (silent swallow)

Case 3 means a malformed version/release will silently fall through to `compareByBuildMetadata`, which may also fail (or succeed with different data). Consider logging when `Bundle.Compare()` returns 0 but neither bundle has a parse error, or at least adding a comment acknowledging this behavior.

internal/operator-controller/bundleutil/bundle.go
```diff
@@ -11,30 +11,48 @@ import (
...
+	if err := json.Unmarshal(pkgData, &pkg); err != nil {
+		return nil, fmt.Errorf("error unmarshalling package property: %w", err)
+	}
+
+	// When BundleReleaseSupport is enabled and bundle has explicit release field, use it
+	if features.OperatorControllerFeatureGate.Enabled(features.BundleReleaseSupport) && pkg.Release != "" {
+		vers, err := bsemver.Parse(pkg.Version)
+		if err != nil {
+			return nil, fmt.Errorf("error parsing version %q: %w", pkg.Version, err)
+		}
```
pedjak: **Inconsistent build metadata handling**: The explicit release path here preserves build metadata in the `Version` (e.g., `1.0.0+ignored`), but the legacy path in `NewLegacyRegistryV1VersionRelease` (versionrelease.go:44) **clears** it with `vr.Version.Build = nil`.

This means `VersionRelease` structs from the two paths have different shapes. While `Compare()` ignores build metadata per semver spec, any code that checks structural equality or serializes versions could see differences. Consider stripping build metadata here when the release is read from the dedicated field, since the build metadata is semantically meaningless at that point.

pedjak (commented):
Thanks for the PR! Here are some review comments on the implementation.

internal/operator-controller/catalogmetadata/filter/successors.go
```diff
@@ -28,11 +29,22 @@ func SuccessorsOf(installedBundle ocv1.BundleMetadata, channels ...declcfg.Chann
...
+	}
+
+	// If BundleReleaseSupport is enabled, also consider bundles
+	// with the same semantic version but higher release as valid successors.
+	// This allows upgrades to re-released bundles (e.g., 2.0.0+1 -> 2.0.0+2).
+	if features.OperatorControllerFeatureGate.Enabled(features.BundleReleaseSupport) {
+		predicates = append(predicates, SameVersionHigherRelease(*installedVersionRelease))
+	}
+
+	// Bundle matches if it's a channel successor, current version (no upgrade),
```
pedjak: **Bypasses channel upgrade graph**: When the gate is enabled, `SameVersionHigherRelease` is OR'd with channel successors, meaning a re-released bundle is always a valid upgrade target even if the channel graph doesn't include it. Is this intentional? If so, it's worth a comment explaining that same-version re-releases are deliberately exempt from channel graph constraints, since users may expect the channel to be the sole authority on upgrade paths.

internal/operator-controller/catalogmetadata/filter/bundle_predicates.go
pedjak: **Mixed-source comparison**: The `expect` argument in `SuccessorsOf` comes from `NewLegacyRegistryV1VersionRelease(installedBundle.Version)`, which always uses the legacy path (build metadata to release, build metadata cleared from version). But `GetVersionAndRelease()` called inside this predicate may return a `VersionRelease` parsed from the explicit `pkg.Release` path (build metadata preserved in version). The comparison still works correctly because `Compare()` ignores build metadata, but a brief comment explaining the mixed-source comparison is intentional and safe would help future readers.

internal/operator-controller/features/features.go
```diff
@@ -2,7 +2,7 @@ package features
 
 import (
 	"github.com/go-logr/logr"
-	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/component-base/featuregate"
```
pedjak, line 7: Nit: This `utilruntime` to `runtime` import alias rename is unrelated to the feature. Consider keeping it as a separate commit or reverting to reduce noise in the diff.

internal/operator-controller/catalogmetadata/compare/compare_test.go
```diff
@@ -168,3 +170,59 @@ func TestByDeprecationFunc(t *testing.T) {
...
+
+	prevEnabled := features.OperatorControllerFeatureGate.Enabled(features.CompositeVersionComparison)
+	t.Cleanup(func() {
+		require.NoError(t, features.OperatorControllerFeatureGate.Set(fmt.Sprintf("%s=%t", features.CompositeVersionComparison, prevEnabled)))
+	})
+
+	t.Run("feature gate disabled - uses build metadata", func(t *testing.T) {
+		require.NoError(t, features.OperatorControllerFeatureGate.Set(fmt.Sprintf("%s=false", features.CompositeVersionComparison)))
+		result := compare.ByVersionAndRelease(registryV1_b1, registryV1_b2)
+		assert.Greater(t, result, 0, "should sort by build metadata: 1.0.0+2 > 1.0.0+1")
```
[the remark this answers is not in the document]
rashmigottipati: Valid suggestion. 
Currently, tests save and restore gate state with t.Cleanup(), so each test cleans up after itself. There's still a theoretical race if two tests run at the same time, but a proper fix would require bigger changes (dependency injection or serializing tests). This is alpha-gated experimental code and we haven't seen test failures, so this approach seems reasonable for now. Can revisit in a follow-up if CI shows issues. So won't be doing this for now.

internal/operator-controller/catalogmetadata/filter/successors_test.go
```diff
@@ -240,3 +242,81 @@ func TestLegacySuccessor(t *testing.T) {
...
+
+// TestSuccessorsOf_WithCompositeVersionComparison_FeatureGateDisabled verifies higher releases
+// are NOT successors when CompositeVersionComparison gate is disabled (testing the default behavior).
+func TestSuccessorsOf_WithCompositeVersionComparison_FeatureGateDisabled(t *testing.T) {
+	// Explicitly disable the feature gate for this test
+	prevEnabled := features.OperatorControllerFeatureGate.Enabled(features.CompositeVersionComparison)
+	require.NoError(t, features.OperatorControllerFeatureGate.Set(fmt.Sprintf("%s=false", features.CompositeVersionComparison)))
+	t.Cleanup(func() {
+		require.NoError(t, features.OperatorControllerFeatureGate.Set(fmt.Sprintf("%s=%t", features.CompositeVersionComparison, prevEnabled)))
+	})
```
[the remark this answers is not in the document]
rashmigottipati: Valid suggestion. 
Same comment as https://github.com/operator-framework/operator-controller/pull/2543#discussion_r3052318296

internal/operator-controller/features/features.go
```diff
@@ -17,8 +17,9 @@ const (
 	WebhookProviderCertManager        featuregate.Feature = "WebhookProviderCertManager"
 	WebhookProviderOpenshiftServiceCA featuregate.Feature = "WebhookProviderOpenshiftServiceCA"
 	HelmChartSupport                  featuregate.Feature = "HelmChartSupport"
-	BoxcutterRuntime                  featuregate.Feature = "BoxcutterRuntime"
-	DeploymentConfig                  featuregate.Feature = "DeploymentConfig"
+	BoxcutterRuntime           featuregate.Feature = "BoxcutterRuntime"
+	DeploymentConfig           featuregate.Feature = "DeploymentConfig"
+	CompositeVersionComparison featuregate.Feature = "CompositeVersionComparison"
 )
```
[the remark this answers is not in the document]
rashmigottipati: Fixed. Renamed from CompositeVersionComparison to BundleReleaseSupport per grokspawn's feedback. Updated PR description to match.

internal/operator-controller/catalogmetadata/filter/successors.go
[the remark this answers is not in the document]
rashmigottipati: Resolved. Added feature gate check at line 41 in successors.go - when BundleReleaseSupport is enabled, SameVersionHigherRelease predicate is appended to allow same version higher release upgrades.

internal/operator-controller/catalogmetadata/filter/successors.go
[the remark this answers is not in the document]
rashmigottipati: Good catch, will fix this

internal/operator-controller/catalogmetadata/filter/successors.go
rashmigottipati: Yes,  explicit channel entries work exactly as before. If a re-release is in the upgrade graph via replaces/skips/skipRange, it's a valid successor (this remains unchanged). New behavior is (which is gate enabled): Re-releases that are not in the explicit graph are now also valid successors via SameVersionHigherRelease

internal/operator-controller/catalogmetadata/filter/bundle_predicates.go
rashmigottipati: This is a valid concern. silently returning false on parsing errors could make debugging harder. However, this approach is consistent with other predicates in this file. Both `ExactVersionRelease` and `InSemverRange` use the same pattern of returning false on errors rather than propagating them. But I agree that this is a limitation.

Since this is a broader issue with the predicate pattern (signature doesn't allow error propagation), it's outside of this PR's scope and I prefer to address it in a followup that revisits all the predicates together. Sounds reasonable?

Will leave this as unresolved for now to see if other reviewers have an opinion on this

Edit: resolving this as all of the SameVersionHigherRelease logic is not needed in this PR

internal/operator-controller/catalogmetadata/filter/successors.go
rashmigottipati: Thinking about this a bit more... My understanding is that when BundleReleaseSupport is enabled, SameVersionHigherRelease is OR'd with channel successors, meaning re-releases (e.g., 2.0.0+1 -> 2.0.0+2) bypass the channel graph and qualify as valid successors even without explicit channel entries (replaces/skips/skipRange).

Is this the intended behavior? Should catalog authors be required to add explicit channel entries for re-releases, or should same version higher release upgrades be deliberately exempt from channel constraints?

 I've added a comment documenting the current behavior, but want to confirm this is the design we want before we finalize this PR.
 cc @joelanford @grokspawn

internal/operator-controller/catalogmetadata/compare/compare.go
rashmigottipati: Renamed to `compareByVersionRelease`

internal/operator-controller/catalogmetadata/compare/compare.go
rashmigottipati: Valid suggestion. Added a comment documenting this behavior

internal/operator-controller/bundleutil/bundle.go
rashmigottipati: Added logic to strip build metadata in explicit release path

internal/operator-controller/catalogmetadata/filter/successors.go
rashmigottipati: In my opinion, this is intentional. I am seeking further clarification on the intended design through https://github.com/operator-framework/operator-controller/pull/2543#discussion_r3053400582

For now, I've added a comment explaining that same version re-releases are deliberately exempt from channel graph constraints

internal/operator-controller/bundleutil/bundle.go
```diff
@@ -11,30 +11,51 @@ import (
...
+		// Strip build metadata when using explicit release field to maintain consistency
+		// with NewLegacyRegistryV1VersionRelease(), which also clears build metadata.
+		vers.Build = nil
```
joelanford: If we have an explicit release, that means that build metadata can actually be treated as build metadata. So we can leave it alone, I think.

The reason we clear it in `NewLegacyRegistryV1VersionRelease` is because we interpret the build metadata as the release, so in that case it is the release, not the build metadata.

rashmigottipati: Thanks for clarifying. I reverted striping of build metadata when parsing explicit pkg.Release field. Added a comment explaining the difference between this and legacy registry+v1 behavior, which interprets build metadata as the release value.

internal/operator-controller/catalogmetadata/compare/compare.go
```diff
@@ -39,9 +40,27 @@ func newMastermindsRange(versionRange string) (bsemver.Range, error) {
...
 
 // ByVersionAndRelease is a comparison function that compares bundles by
-// version and release. Bundles with lower versions/releases are
-// considered less than bundles with higher versions/releases.
+// version and release. When the BundleReleaseSupport feature gate is
+// enabled, it uses Bundle.Compare() (which reads pkg.Release from olm.package)
+// and falls back to build metadata comparison if equal. When disabled, it uses
+// version+release from build metadata (backward compatible).
 func ByVersionAndRelease(b1, b2 declcfg.Bundle) int {
+	if features.OperatorControllerFeatureGate.Enabled(features.BundleReleaseSupport) {
```
joelanford: This change feels off to me.

It seems like _this_ function shouldn't change at all. The change, as I understand it, boils down to how we build a version/release.

Without the feature gate: Only do it the legacy way by inspecting/parsing build metadata

With the feature gate: If release field is set, parse it directly. Only if unset do we fall back to legacy registry+v1 build metadata parsing.

But then once we have a VersionRelease, the comparison is the same.

Am I missing something else?

joelanford: Maybe the thing I'm missing is:
- operator-registry has a CompositeVersion type
- operator-controller has a VersionRelease type.

These represent the same exact concept, and we're trying to move in the direction of using the operator-registry variant?

IMO, we shouldn't mix them. We should pick one and delete the other.

internal/operator-controller/catalogmetadata/compare/compare.go
```diff
@@ -39,9 +40,27 @@ func newMastermindsRange(versionRange string) (bsemver.Range, error) {
...
+		// Use CompositeVersion comparison (reads pkg.Release from olm.package).
+		// Note: Bundle.Compare() returns 0 in three cases:
+		//   1. Bundle names are identical
+		//   2. Versions and releases are equal
+		//   3. Parsing errors occur (silently swallowed)
+		// Catalog validation should catch malformed versions before they reach this code.
+		result := b2.Compare(&b1)
```
joelanford: Comparing at the bundle level seems incorrect to me based on the name of this function. What if the bundle comparison method considers more than the version and release?

Seems like we'd want `b2.CompositeVersion.Compare(&b1.CompositeVersion)` or something like that?

internal/operator-controller/catalogmetadata/filter/successors.go
rashmigottipati: After discussing with @grokspawn , I'm working on removing the `SameVersionHigherRelease` logic entirely. Re-releases now require explicit channel entries to be valid successors.

internal/operator-controller/catalogmetadata/filter/successors.go
rashmigottipati: after discussing with @grokspawn, I'm working on removing all `SameVersionHigherRelease` predicate logic from the PR. The channel graph is now the sole authority on upgrade paths. Based on my new understanding, dynamic reordering will be an option for future bundle formats, but not for registry+v1. 
Marking this comment as resolved.

internal/operator-controller/catalogmetadata/compare/compare.go
grokspawn: I completely agree that we need to consolidate to a single representative type. 
This PR is not intended to force the type consistency yet, since if it adopted the CompositeVersion type right now it would lack a formal Release type which we determined was essential to encourage broad adoption. 
It just provides the featuregate and the sensitivity to an explicit release version.
I'd intended to merge https://github.com/operator-framework/operator-registry/pull/1938 and then we would have follow-ups to op-reg & op-con to adopt the base type consistently throughout. 
I was just hoping to be able to merge the two PRs independently.

internal/operator-controller/catalogmetadata/filter/successors.go
```diff
@@ -20,19 +20,22 @@ func SuccessorsOf(installedBundle ocv1.BundleMetadata, channels ...declcfg.Chann
...
-	return filter.Or(
+	// Bundle matches if it's a channel successor or current version (no upgrade).
+	// Re-releases must have explicit channel entries to be valid successors.
+	predicates := []filter.Predicate[declcfg.Bundle]{
 		successorsPredicate,
 		ExactVersionRelease(*installedVersionRelease),
-	), nil
+	}
+
+	return filter.Or(predicates...), nil
```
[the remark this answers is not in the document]
grokspawn: Fail to question the design choice in introduction, but questioning it when we remove it as erroneous.  Nice work, slop-bot.

internal/operator-controller/catalogmetadata/compare/compare.go
joelanford: So WDYT about this PR focusing on "if release explicitly set, load it into `VersionRelease` value". And then a later PR comes in a moves everything over to the operator-registry types/comparisons?

internal/operator-controller/catalogmetadata/filter/successors.go
```diff
@@ -20,19 +20,22 @@ func SuccessorsOf(installedBundle ocv1.BundleMetadata, channels ...declcfg.Chann
 	//    registry+v1 and embed release in build metadata.
 	installedVersionRelease, err := bundle.NewLegacyRegistryV1VersionRelease(installedBundle.Version)
 	if err != nil {
-		return nil, fmt.Errorf("failed to get version and release of installed bundle: %v", err)
+		return nil, fmt.Errorf("failed to get version and release of installed bundle: %w", err)
 	}
```
joelanford, line 60: Does this section of code also need to change similar to `parseVersionRelease`?

If we resolve a bundle in the catalog that has an explicit release field, we'll install it, but we don't persist it's release value anywhere.

Later (here) it'll show up as `installedBundle`, but it'll lack a `release` value (there's not even a field for it). And worse, we'll try to parse the `Version` field as a legacy registry+v1 version. And since `parseVersionRelease` preserves the original build metadata, we'll actually try to parse it as the release value here.

TL;DR: I think this PR needs to also add a new field to the `installedBundle` struct (which I think ends up in the CRD?) so that the release can roundtrip properly.

internal/operator-controller/catalogmetadata/compare/compare.go
grokspawn: Seems reasonable to me.

rashmigottipati: Thanks for the thorough review Joe. Your suggestions sound good to me. I've updated the PR to focus on:
  - If `pkg.Release` is explicitly set (and feature gate enabled), parse it and load into `VersionRelease` value
  - Roundtrip it through CRD field and annotations
  - Removed all the changes to comparison logic 
  
 The type consolidation with operator-registry's `CompositeVersion` can happen in a follow-up PR once operator-framework/operator-registry#1938 is merged.

internal/operator-controller/catalogmetadata/compare/compare.go
rashmigottipati: Marking this as resolved as this comment is now obsolete. Per your earlier feedback, I've removed all comparison changes to focus solely on parsing the explicit `pkg.Release` field and roundtripping it.

internal/operator-controller/catalogmetadata/filter/successors.go
rashmigottipati: Addressed this. Updated the PR to add a new `Release` field to `BundleMetadata` struct so the release value roundtrips properly. 

When a bundle with an explicit release is resolved, `bundleutil.MetadataFor()` serializes the release into `BundleMetadata.Release` and controllers persist it in annotations. Later, when that bundle shows up as `installedBundle`, this code in `successors.go` reads the release field back and uses it to determine the correct parsing path (explicit vs registry+v1).

api/v1/clusterextension_types.go
```diff
@@ -466,6 +466,17 @@ type BundleMetadata struct {
 	// +required
 	// +kubebuilder:validation:XValidation:rule="self.matches(\"^([0-9]+)(\\\\.[0-9]+)?(\\\\.[0-9]+)?(-([-0-9A-Za-z]+(\\\\.[-0-9A-Za-z]+)*))?(\\\\+([-0-9A-Za-z]+(-\\\\.[-0-9A-Za-z]+)*))?\")",message="version must be well-formed semver"
 	Version string `json:"version"`
+
+	// release is an optional field that references the release value for this bundle.
+	// The release follows pre-release/build metadata syntax as defined in https://semver.org/,
+	// consisting of dot-separated identifiers where numeric identifiers must not have leading zeros.
+	// For bundles with explicit pkg.Release metadata, this field contains that release value.
+	// For registry+v1 bundles, this field contains the release extracted from version's build metadata (e.g., '2' from '1.0.0+2').
```
joelanford:
```suggestion
	// For registry+v1 bundles lacking an explicit release value, this field contains the release extracted from version's build metadata (e.g., '2' from '1.0.0+2').
```

api/v1/clusterextension_types.go
```diff
@@ -466,6 +466,17 @@ type BundleMetadata struct {
 	// +required
 	// +kubebuilder:validation:XValidation:rule="self.matches(\"^([0-9]+)(\\\\.[0-9]+)?(\\\\.[0-9]+)?(-([-0-9A-Za-z]+(\\\\.[-0-9A-Za-z]+)*))?(\\\\+([-0-9A-Za-z]+(-\\\\.[-0-9A-Za-z]+)*))?\")",message="version must be well-formed semver"
 	Version string `json:"version"`
+
+	// release is an optional field that references the release value for this bundle.
+	// The release follows pre-release/build metadata syntax as defined in https://semver.org/,
+	// consisting of dot-separated identifiers where numeric identifiers must not have leading zeros.
```
joelanford: I think the GoDoc for Release needs two things:

1. Explain what the release field is and what it conveys. 
2. Describe the syntax and semantics on their own terms, without referencing semver pre-release syntax. Mentioning semver pre-release in the API docs could mislead users into thinking this field has semver pre-release semantics (e.g., that it indicates a pre-release version, or that it affects version precedence the way semver pre-releases do).

Perhaps something more along these lines?
```go
	// release is an optional field that identifies a specific release of this bundle's version.
	// A release represents a re-publication of the same version, typically used to deliver
	// packaging or metadata changes without changing the version number. When multiple
	// releases exist for the same version, higher releases are preferred.
	//
	// The value consists of dot-separated identifiers, where each identifier is either a
	// numeric value (without leading zeros) or an alphanumeric string (e.g., "2", "1.el9",
	// "3.alpha.1"). Releases are compared identifier by identifier: numeric identifiers are
	// compared as integers, alphanumeric identifiers are compared lexically, and numeric
	// identifiers always sort before alphanumeric identifiers.
```

joelanford: Oh, we should also say that an unset release is less preferred than all other release values.

api/v1/clusterextension_types.go
```diff
@@ -466,6 +466,17 @@ type BundleMetadata struct {
...
+	// For bundles with explicit pkg.Release metadata, this field contains that release value.
+	// For registry+v1 bundles, this field contains the release extracted from version's build metadata (e.g., '2' from '1.0.0+2').
+	// This field may be omitted if there is no explicit release and the version contains no parseable build metadata.
```
joelanford: I wouldn't mention "no parsable build metadata" here since that is specific to registry+v1 bundles, and we cover it above. I think I'd just say that we omit it when the bundle does not specific a release.

```suggestion
	// This field is omitted when the bundle's release value is unset.
```

api/v1/clusterextension_types.go
```diff
@@ -466,6 +466,17 @@ type BundleMetadata struct {
...
+	//
+	// +optional
+	// +kubebuilder:validation:XValidation:rule="self.matches(\"^$|^(0|[1-9][0-9]*|[0-9]*[A-Za-z-][0-9A-Za-z-]*)(\\\\.(0|[1-9][0-9]*|[0-9]*[A-Za-z-][0-9A-Za-z-]*))*$\")",message="release must be empty or well-formed pre-release/build metadata (dot-separated identifiers, numeric parts without leading zeros)"
```
joelanford: Since this field is feature-gated, we should add the experimental marker to it to make sure it only shows up in the experimental CRD for now.

internal/operator-controller/bundleutil/bundle.go
```diff
@@ -11,34 +12,62 @@ import (
...
+		if err != nil {
+			return nil, fmt.Errorf("error parsing release %q: %w", pkg.Release, err)
+		}
+		return &bundle.VersionRelease{
+			Version: vers,
+			Release: rel,
+		}, nil
+	}
+
+	// Fall back to legacy registry+v1 behavior (release in build metadata)
```
joelanford, line 47: Please preserve a modified TODO comment here:

```suggestion
	// Fall back to legacy registry+v1 behavior (release in build metadata)
	//
	// TODO: For now, we assume that all bundles are registry+v1 bundles.
	//   In the future, for supporting other bundle formats, we should not
	//   use the legacy registry+v1 mechanism (i.e. using build metadata in
	//   the version) to determine the bundle's release.
```

internal/operator-controller/bundleutil/bundle.go
```diff
@@ -11,34 +12,62 @@ import (
...
+func MetadataFor(bundleName string, vr bundle.VersionRelease) ocv1.BundleMetadata {
+	bm := ocv1.BundleMetadata{
 		Name:    bundleName,
-		Version: bundleVersion.String(),
+		Version: vr.Version.String(),
+	}
+	if len(vr.Release) > 0 {
+		parts := slicesutil.Map(vr.Release, func(pr bsemver.PRVersion) string { return pr.String() })
+		bm.Release = strings.Join(parts, ".")
 	}
```
joelanford, line 108: If we move this logic into a `String()` method defined on the Release type, we could just directly set:

```go
	bm := ocv1.BundleMetadata{
		Name:    bundleName,
		Version: vr.Version.String(),
		Release: vr.Release.String(),
	}
```

internal/operator-controller/catalogmetadata/filter/successors.go
```diff
@@ -13,14 +13,36 @@ import (
...
+		installedVersionRelease = &bundle.VersionRelease{
+			Version: version,
+			Release: release,
+		}
+	} else {
+		// Legacy registry+v1: release embedded in version's build metadata
+		installedVersionRelease, err = bundle.NewLegacyRegistryV1VersionRelease(installedBundle.Version)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get version and release of installed bundle: %w", err)
+		}
```
joelanford: Since we will be parsing/writing the `BundleMetadata` with a Release value into the API from bundles that we resolve, this function may no longer need to worry about falling back to the legacy parsing.

However there is the edge case of:
- Old OLMv1 version installed a bundle that had a release value in its build metadata prior to this feature being in place
- New OLMv1 is not reconciling a CE whose bundle metadata still stores the Release value as build metadata.

We can't tell if that means "release was explicitly empty and build metadata is actually build metadata" or it if means "build metadata should be interpreted as the release value"

To solve that, I think we may need to add `Release` to BundleMetadata as a `*string` so that we can tell the difference between "no release" and an explicit empty release, and that means code that knows about the release field would actually store `""` as the release value so that we can tell the difference.

joelanford: That _also_ means that there is a difference between these two `olm.package` property values:
1. `{"version": "1.2.0+1"}`
2. `{"version": "1.2.0+1", "release": ""}`

For (1), we would parse a release from legacy registry+v1 build metadata
- `{Version: "1.2.0", Release: "1"}`

For (2), we would say "you've provided an explicit release value, so build metadata will be treated as build metadata
- `{Version: "1.2.0+1", Release: ""}`

rashmigottipati: @joelanford Good point. 
I was thinking about keeping Release as a non-pointer and relying on omitzero, but I don’t think that works for this case.

This seems to be one of the rare cases where we'd want to tell the difference between the field being omitted vs field explicitly empty. With omitzero, both of these end up with the same representation once serialized/deserialized. 

Based on your suggestion, using `*string` would let us preserve that distinction and handle the upgrade edge case correctly. Thanks for calling this out.

I'll update the PR

api/v1/clusterextension_types.go
rashmigottipati: Done. Updated in [latest commit](https://github.com/operator-framework/operator-controller/pull/2543/changes/a8330020dd788646844bf66dba86ac242fa2d62f)

api/v1/clusterextension_types.go
rashmigottipati: Rewrote GoDoc to explain what release is, describe syntax/semantics without referencing semver pre-release, and added note that unset release is less preferred than all other release values.
Updated in [latest commit](https://github.com/operator-framework/operator-controller/pull/2543/changes/a8330020dd788646844bf66dba86ac242fa2d62f)

api/v1/clusterextension_types.go
rashmigottipati: Done, updated in [latest commit](https://github.com/operator-framework/operator-controller/pull/2543/changes/a8330020dd788646844bf66dba86ac242fa2d62f)

api/v1/clusterextension_types.go
rashmigottipati: Done. Added `<opcon:experimental>` marker. Verified release field only appears in experimental CRD.

internal/operator-controller/bundleutil/bundle.go
rashmigottipati: Done. Added TODO comment about assuming all bundles are registry+v1.

internal/operator-controller/bundleutil/bundle.go
rashmigottipati: Done. Added `String()` method to Release type and simplified `MetadataFor` accordingly.

internal/operator-controller/catalogmetadata/filter/successors.go
rashmigottipati: Done. Changed `Release` to `*string`.

internal/operator-controller/controllers/clusterextension_reconcile_steps.go
```diff
@@ -191,11 +191,11 @@ func ResolveBundle(r resolve.Resolver, c client.Client) ReconcileStepFunc {
...
-			// TODO: Right now, operator-controller only supports registry+v1 bundles and has no concept
-			//   of a "release" field. If/when we add a release field concept or a new bundle format
-			//   we need to re-evaluate use of `AsLegacyRegistryV1Version` so that we avoid propagating
-			//   registry+v1's semver spec violations of treating build metadata as orderable.
-			BundleMetadata: bundleutil.MetadataFor(resolvedBundle.Name, resolvedBundleVersion.AsLegacyRegistryV1Version()),
+			// MetadataFor accepts VersionRelease and populates both Version and Release fields.
+			// - With BundleReleaseSupport enabled + explicit pkg.Release: Release from pkg.Release field
+			// - Registry+v1 bundles (e.g., "1.0.0+2"): Release extracted from build metadata (e.g., "2")
+			// - Both result in separate Version and Release fields in BundleMetadata for roundtripping
+			BundleMetadata: bundleutil.MetadataFor(resolvedBundle.Name, *resolvedBundleVersion),
```
pedjak: **Ungated behavioral change to the `Version` field.**

This changes `MetadataFor` from using `AsLegacyRegistryV1Version()` (which reconstituted build metadata, e.g. `"1.0.0+2"`) to using `vr.Version.String()` (which has build metadata already stripped by `NewLegacyRegistryV1VersionRelease` at `versionrelease.go:44`). This runs for **all** users regardless of the `BundleReleaseSupport` gate.

For **standard CRD users** (where the `release` field is pruned by the API server):
- Before: `status.install.bundle.version: "1.0.0+2"`
- After: `status.install.bundle.version: "1.0.0"` (release information completely lost)

This also triggers a one-time re-apply for every installed bundle with build metadata on controller upgrade, because `bundleUnchanged` at line 377 compares `"1.0.0" != "1.0.0+2"`.

**Suggestion:** When `BundleReleaseSupport` is disabled, preserve the old behavior:
```go
if features.OperatorControllerFeatureGate.Enabled(features.BundleReleaseSupport) {
    BundleMetadata: bundleutil.MetadataFor(resolvedBundle.Name, *resolvedBundleVersion),
} else {
    BundleMetadata: bundleutil.MetadataFor(resolvedBundle.Name, resolvedBundleVersion.AsLegacyRegistryV1Version()),
}
```
Or keep the old `MetadataFor(string, bsemver.Version)` signature for the ungated path.

pedjak (commented):
## Deep Review: Roundtrip Correctness and Ungated API Change

The core design (separating release from version, feature-gating the new parsing path) is sound. The CRD field documentation, validation regex, and deep-copy generation all look correct.

However, I traced the full data lifecycle through all paths (**resolve → MetadataFor → annotations → GetRevisionStates → SuccessorsOf → ExactVersionRelease**) and found two issues that need attention before merge:

1. **Ungated Version field change** — The `MetadataFor` signature change strips build metadata from the `Version` field for *all* users, not just those with `BundleReleaseSupport` enabled. Standard CRD users lose release information entirely (the `release` field is pruned by the API server).

2. **Roundtrip correctness bug during controller upgrade** — Writing `BundleReleaseKey` unconditionally creates a nil-vs-empty-string Release ambiguity that causes `ExactVersionRelease` to fail matching the installed bundle against itself in the catalog.

Both issues and suggested fixes are detailed in the inline comments below.

internal/operator-controller/controllers/clusterextension_reconcile_steps.go
```diff
@@ -408,11 +408,16 @@ func UnpackBundle(i imageutil.Puller, cache imageutil.Cache) ReconcileStepFunc {
...
+		releaseValue := ""
+		if state.resolvedRevisionMetadata.Release != nil {
+			releaseValue = *state.resolvedRevisionMetadata.Release
+		}
 		revisionAnnotations := map[string]string{
 			labels.BundleNameKey:      state.resolvedRevisionMetadata.Name,
 			labels.PackageNameKey:     state.resolvedRevisionMetadata.Package,
 			labels.BundleVersionKey:   state.resolvedRevisionMetadata.Version,
 			labels.BundleReferenceKey: state.resolvedRevisionMetadata.Image,
+			labels.BundleReleaseKey:   releaseValue,
```
pedjak: **Unconditional `BundleReleaseKey` write causes a roundtrip correctness bug during controller upgrade.**

When `Release` is nil (old install, no release annotation), `releaseValue` defaults to `""` and `BundleReleaseKey` is written as an empty-string annotation. On next read at `GetRevisionStates`, the annotation key exists → `Release = ptr("")` (non-nil). This changes the code path taken in `SuccessorsOf` from legacy (correct) to explicit (broken for this case).

**Full failure scenario:**
1. Old controller installed `"1.0.0+2"` → annotation `BundleVersionKey="1.0.0+2"`, no `BundleReleaseKey`
2. New controller starts, catalogs temporarily unavailable → fallback at line 299 uses installed metadata
3. Apply writes: `BundleReleaseKey=""` (this line)
4. Next reconcile reads: `Version="1.0.0+2"`, `Release=ptr("")`
5. `SuccessorsOf`: `Release != nil` → explicit path → `bsemver.Parse("1.0.0+2")` = `1.0.0+2`, `NewRelease("")` = `nil`
6. → `installedVersionRelease = {Version: 1.0.0+2, Release: nil}`
7. Catalog parses same bundle via `GetVersionAndRelease` → `{Version: 1.0.0, Release: [2]}`
8. `ExactVersionRelease` compares: `Version.Compare` returns 0 (build metadata ignored per semver), but `Release.Compare(nil, [2])` returns -1
9. **Result: installed bundle cannot match itself** → resolution may fail if no explicit upgrade edge exists

**Fix:** Only write `BundleReleaseKey` when `Release` is non-nil:
```go
if state.resolvedRevisionMetadata.Release != nil {
    revisionAnnotations[labels.BundleReleaseKey] = *state.resolvedRevisionMetadata.Release
}
```

internal/operator-controller/controllers/boxcutter_reconcile_steps.go
```diff
@@ -99,11 +103,16 @@ func MigrateStorage(m StorageMigrator) ReconcileStepFunc {
...
+		releaseValue := ""
+		if state.resolvedRevisionMetadata.Release != nil {
+			releaseValue = *state.resolvedRevisionMetadata.Release
+		}
 		revisionAnnotations := map[string]string{
 			labels.BundleNameKey:      state.resolvedRevisionMetadata.Name,
 			labels.PackageNameKey:     state.resolvedRevisionMetadata.Package,
 			labels.BundleVersionKey:   state.resolvedRevisionMetadata.Version,
 			labels.BundleReferenceKey: state.resolvedRevisionMetadata.Image,
+			labels.BundleReleaseKey:   releaseValue,
```
pedjak: Same unconditional `BundleReleaseKey` write issue as in `ApplyBundle` — see the detailed roundtrip failure scenario in my comment there. The fix is the same: only add `BundleReleaseKey` to the annotations map when `Release` is non-nil.

internal/operator-controller/catalogmetadata/filter/successors.go
```diff
@@ -13,14 +13,36 @@ import (
...
 	}
```
pedjak: The branching on `Release != nil` is correct in steady state, but creates a roundtrip inconsistency during controller upgrade when `BundleReleaseKey` is written unconditionally (see my comment on `ApplyBundle`).

When old-format annotations (`Version="1.0.0+2"`, no release annotation) go through a fallback apply and get `BundleReleaseKey=""` added, the next read produces `Release=ptr("")`. This branch then fires, parsing version `"1.0.0+2"` with `NewRelease("")` = nil, yielding `{Version: 1.0.0+2, Release: nil}`. Meanwhile, the catalog parses the same bundle as `{Version: 1.0.0, Release: [2]}`. These don't compare equal because `Release nil != [2]`, even though `Version.Compare` returns 0 (blang ignores build metadata per semver spec).

This is fixed by not writing `BundleReleaseKey` unconditionally (see my other comments). But it may also be worth adding a defensive guard here: if the version contains build metadata and release is empty, fall through to the legacy path, since that's the scenario this explicit path can't handle correctly.

rashmigottipati: Not adding the defensive guard here as it'll broke legitimate cases where bundles have build metadata (e.g., "1.0.0+git.abc") with explicit empty release. Instead, added empty string handling that creates an empty slice to match catalog parsing behavior and ensure roundtrip correctness. The root cause is fixed by conditional writes of `BundleReleaseKey` in both controllers.

internal/operator-controller/controllers/boxcutter_reconcile_steps.go
rashmigottipati: Addressed in latest commit

internal/operator-controller/controllers/clusterextension_reconcile_steps.go
rashmigottipati: Fixed by feature gating `MetadataFor` itself. When BundleReleaseSupport is disabled, `MetadataFor` calls `vr.AsLegacyRegistryV1Version().String()` to reconstitute build metadata in the Version field, preserving backward compatibility for standard CRD users.

internal/operator-controller/controllers/clusterextension_reconcile_steps.go
rashmigottipati: Addressed in latest commit. Now only writes `BundleReleaseKey` when Release != nil, preventing the empty-string annotation that caused the roundtrip bug.

api/v1/clusterextension_types.go
```diff
@@ -466,6 +466,28 @@ type BundleMetadata struct {
...
+	//
+	// For bundles with explicit pkg.Release metadata, this field contains that release value.
+	// For registry+v1 bundles lacking an explicit release value, this field contains the release
+	// extracted from version's build metadata (e.g., '2' from '1.0.0+2').
+	// This field is omitted when the bundle's release value is unset.
+	//
+	// +optional
+	// <opcon:experimental>
+	// +kubebuilder:validation:XValidation:rule="self.matches(\"^$|^(0|[1-9][0-9]*|[0-9]*[A-Za-z-][0-9A-Za-z-]*)(\\\\.(0|[1-9][0-9]*|[0-9]*[A-Za-z-][0-9A-Za-z-]*))*$\")",message="release must be empty or consist of dot-separated identifiers (numeric without leading zeros, or alphanumeric)"
+	Release *string `json:"release,omitempty"`
```
grokspawn, line 490: I was planning to do this in a follow-up PR, because I think it ultimately should be based on op-reg Release type instead of op-con Release type. 
This is neither.  

Do we have any concern if a later PR makes this explicitly declcfg.Release type?

Please align the type+validation with [api](https://github.com/operator-framework/api/blob/master/pkg/operators/v1alpha1/clusterserviceversion_types.go#L280-L295) and my follow-on PR in [operator-registry](https://github.com/operator-framework/operator-registry/pull/1938/changes#diff-a03814ccc99b73651ac3cb69d994dda696d06af54b2785214b4d0f7360ac28bbR13).

grokspawn: Unless there is meaning between nil and empty, we generally prefer to do 

```suggestion
	Release string `json:"release,omitzero"`
```

I don't think there's any difference between these cases for this type.  If the release is empty or omitted then it equates to no release value.

internal/operator-controller/bundleutil/bundle.go
```diff
@@ -11,34 +11,99 @@ import (
...
+func parseExplicitRelease(version, releaseStr string) (*bundle.VersionRelease, error) {
+	vers, err := bsemver.Parse(version)
+	if err != nil {
+		return nil, fmt.Errorf("error parsing version %q: %w", version, err)
+	}
+
+	var rel bundle.Release
+	if releaseStr == "" {
+		// Explicit empty release: use empty slice (not nil)
+		rel = bundle.Release([]bsemver.PRVersion{})
```
grokspawn, line 74: This goes against bundle.NewRelease handling, which returns nil if given an empty string.  
I've also aligned with that approach [here](https://github.com/operator-framework/operator-registry/pull/1938/changes#diff-a03814ccc99b73651ac3cb69d994dda696d06af54b2785214b4d0f7360ac28bbR76-R78).

internal/operator-controller/bundleutil/bundle.go
```diff
@@ -11,34 +11,99 @@ import (
...
 
-// MetadataFor returns a BundleMetadata for the given bundle name and version.
-func MetadataFor(bundleName string, bundleVersion bsemver.Version) ocv1.BundleMetadata {
+func parseVersionRelease(pkgData json.RawMessage) (*bundle.VersionRelease, error) {
+	var pkg property.Package
+	if err := json.Unmarshal(pkgData, &pkg); err != nil {
+		return nil, fmt.Errorf("error unmarshalling package property: %w", err)
+	}
+
+	// Check if release field is explicitly present in JSON (even if empty).
```
grokspawn, line 32: Since `Package.Release` is defined in [operator-registry](https://github.com/operator-framework/operator-registry/blob/master/alpha/property/property.go#L41) as string with omitzero, isn't the salient test whether the unmarshaled Package at 28 has a non-empty Release or not? 

If we are worried about the scenario where the feature is disabled, but a bundle with a release version is in the catalog, then we can explicitly clear the field.

Is there another scenario that's important?

internal/operator-controller/bundleutil/bundle.go
```diff
@@ -11,34 +11,99 @@ import (
...
+
+// MetadataFor returns a BundleMetadata for the given bundle name and version/release.
+func MetadataFor(bundleName string, vr bundle.VersionRelease) ocv1.BundleMetadata {
+	if features.OperatorControllerFeatureGate.Enabled(features.BundleReleaseSupport) {
+		// New behavior: separate Version and Release fields
+		bm := ocv1.BundleMetadata{
+			Name:    bundleName,
+			Version: vr.Version.String(),
+		}
+		if vr.Release != nil {
```
grokspawn, line 96: This looks like we could skip it if we we move to `omitempty` instead of a pointer with `omitzero`, and furthermore could be omitted if we're using an explicit type instead of a string.  
(I don't think we can use an explicit type yet, unless the bundle.Release supports marshaling/unmarshaling.)

internal/operator-controller/controllers/boxcutter_reconcile_steps.go
```diff
@@ -71,6 +71,10 @@ func (d *BoxcutterRevisionStatesGetter) GetRevisionStates(ctx context.Context, e
 				Version: rev.Annotations[labels.BundleVersionKey],
 			},
 		}
+		// Only set Release if the annotation key exists (to distinguish "not set" from "explicitly empty")
```
grokspawn, line 74: I don't think these are different cases.  The value is only salient if it exists _and_ is non-empty.  Otherwise, there is no effective release.

api/v1/clusterextension_types.go
rashmigottipati: @grokspawn I was going to converge the type in the follow-up PR as well but based on this[ review comment ](https://github.com/operator-framework/operator-controller/pull/2543#discussion_r3054137664), I added `Release` field to the BundleMetadata struct so we can persist it and roundtrip it properly. 
No concern if a later PR makes this a declcfg.Release type though. 

I'll align the type+validation with [api](https://github.com/operator-framework/api/blob/master/pkg/operators/v1alpha1/clusterserviceversion_types.go#L280-L295) and your follow-on PR in [operator-registry](https://github.com/operator-framework/operator-registry/pull/1938/changes#diff-a03814ccc99b73651ac3cb69d994dda696d06af54b2785214b4d0f7360ac28bbR13) as you suggested.

rashmigottipati: > I don't think there's any difference between these cases for this type. If the release is empty or omitted then it equates to no release value.

Regarding using omitzero, @joelanford has identified an upgrade edge case when OLMv1 upgrades from a version without the Release field to one with it:
https://github.com/operator-framework/operator-controller/pull/2543#discussion_r3066023818

To fix this problem, I think we need to make Release a *string type to distinguish between:
 - Field not present (nil) - parse release from build metadata (legacy registry+v1 behavior for upgrade compatibility)
 - Field present but empty ("") - treat build metadata as actual build metadata, not release

  Without this distinction, when OLMv1 reconciles an existing ClusterExtension after upgrade, we can't tell if `Version: "1.2.0+1"` with no Release field means "parse +1 as release" (old behavior) or "keep +1 as build metadata" (new behavior).

 With omitzero on a value type, both cases serialize identically, breaking the upgrade path. The pointer lets us preserve the semantic difference between these two `olm.package` values:
 `{"version": "1.2.0+1"}`  --> `{Version: "1.2.0", Release: "1"}`
`{"version": "1.2.0+1", "release": ""}` --> `{Version: "1.2.0+1", Release: ""}`

internal/operator-controller/bundleutil/bundle.go
rashmigottipati: In my understanding, we need the empty slice (not nil) for roundtrip correctness.

If we use nil for explicit empty release, then in MetadataFor, the check `if vr.Release != nil` fails and we lose the Release field entirely. This breaks the distinction between "Release field not present" (legacy) vs "Release field explicitly empty".

The three-state representation is:
 - nil → field not present (parse from build metadata)
 - []bsemver.PRVersion{} → field explicitly empty
 - []bsemver.PRVersion{...} → field with value

Without this, we can't preserve the upgrade path semantics mentioned in [this comment ](https://github.com/operator-framework/operator-controller/pull/2543/changes#r3119827392)

Pls correct me if I am missing something here.

internal/operator-controller/bundleutil/bundle.go
rashmigottipati: If you agree that we need this distinction for the upgrade path based on the previous review comments, should we update `property.Package.Release` in operator-registry to `*string` instead of string with omitzero? That would preserve the field presence distinction at the source and eliminate the need for the helper struct workaround.

internal/operator-controller/bundleutil/bundle.go
rashmigottipati: Good point. I am thinking that when opreg PR merges and we do a follow-up PR in opcon to do the type convergence, we can then use `bundle.Release` directly instead of `*string`, since opreg PR already uses the explicit type with JSON marshaling support. Let's align with that in the follow-up PR?

api/v1/clusterextension_types.go
rashmigottipati: Validation is now aligned with MaxLength=20.

grokspawn: I guess I'm not clear on how this breaks the upgrade path.  My understanding is that currently we would encode a ClusterExtension (1.2.0+1) as a `VersionRelease{ Version: "1.2.0", Release: 1 }` via `NewLegacyRegistryV1VersionRelease` ([here](https://github.com/operator-framework/operator-controller/blob/7277a46af0d0cca9d7bd2adfa29ccaba65fb5654/internal/operator-controller/bundle/versionrelease.go#L21)). 

If our processing to populate the BundleMetadata with .version & .release comes from the VersionRelease, then we would already be honoring the build-metadata-based release.  

I can't get that comment link to resolve.  I'll have to track it down when I have more cycles.

internal/operator-controller/bundleutil/bundle.go
grokspawn: I don't think there is any semantic value in "release field explicitly empty".

api/v1/clusterextension_types.go
rashmigottipati: @grokspawn Attaching that comment here for easier access:
Screenshot 2026-04-24 at 5 43 10 PM

I've also unresolved that particular comment so it shows up

grokspawn: Concur that BundleMetadata needs to be sensitive to "installed before this version of OLM introduced the field considerations".

internal/operator-controller/bundleutil/bundle.go
grokspawn: So .... it's not a simple as I'd hoped, because we need to handle the case where existing ClusterExtensions (with BundleMetadata) were installed before OLMv1 upgraded itself and implemented this field.  Resolving.

internal/operator-controller/controllers/boxcutter_reconcile_steps.go
```diff
@@ -71,6 +71,10 @@ func (d *BoxcutterRevisionStatesGetter) GetRevisionStates(ctx context.Context, e
...
+		if releaseValue, ok := rev.Annotations[labels.BundleReleaseKey]; ok {
+			rm.Release = &releaseValue
```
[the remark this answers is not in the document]
rashmigottipati: Fixed in latest commit

internal/operator-controller/applier/boxcutter.go
```diff
@@ -85,6 +85,9 @@ func (r *SimpleRevisionGenerator) GenerateRevisionFromHelmRelease(
 		if v := helmRelease.Labels[labels.BundleVersionKey]; v != "" {
 			annotationUpdates[labels.BundleVersionKey] = v
 		}
+		if v := helmRelease.Labels[labels.BundleReleaseKey]; v != "" {
+			annotationUpdates[labels.BundleReleaseKey] = v
+		}
```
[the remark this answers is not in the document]
rashmigottipati: Makes sense.  Removing the `v != ""` to preserve empty values.

internal/operator-controller/applier/boxcutter.go
```diff
@@ -96,12 +99,16 @@ func (r *SimpleRevisionGenerator) GenerateRevisionFromHelmRelease(
...
+	revisionAnnotations := map[string]string{
 		labels.BundleNameKey:      helmRelease.Labels[labels.BundleNameKey],
 		labels.PackageNameKey:     helmRelease.Labels[labels.PackageNameKey],
 		labels.BundleVersionKey:   helmRelease.Labels[labels.BundleVersionKey],
 		labels.BundleReferenceKey: helmRelease.Labels[labels.BundleReferenceKey],
-	})
+	}
+	if v := helmRelease.Labels[labels.BundleReleaseKey]; v != "" {
+		revisionAnnotations[labels.BundleReleaseKey] = v
+	}
```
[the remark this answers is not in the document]
rashmigottipati: Makes sense.  Removing the `v != ""` to preserve empty values.

internal/operator-controller/bundleutil/bundle.go
```diff
@@ -11,34 +11,99 @@ import (
...
+	var releaseField struct {
+		Release *string `json:"release"`
+	}
+	if err := json.Unmarshal(pkgData, &releaseField); err != nil {
+		return nil, fmt.Errorf("error unmarshalling package release field: %w", err)
+	}
+
+	// When BundleReleaseSupport is enabled and bundle has explicit release field, use it.
+	if features.OperatorControllerFeatureGate.Enabled(features.BundleReleaseSupport) && releaseField.Release != nil {
+		return parseExplicitRelease(pkg.Version, *releaseField.Release)
```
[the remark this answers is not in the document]
rashmigottipati: Will optimize this in follow-up PR

grokspawn (approved):

PR #2566, opened 2026-03-16
🐛 OCPBUGS-77829: fix(Boxcutter): Fix ClusterExtension spec changes ignored while revision is rolling out

camilamacedo86:
#### **What's wrong**

When a revision is rolling out, the controller skips the resolver — even if the CE consumer changes the spec. The CE consumer's only option to recover is manually deleting revisions.

#### **What this PR does**

Fixes Boxcutter rollout behavior so ClusterExtension spec updates (during an in-progress/stuck rollout) are no longer ignored, by persisting and comparing a catalog-spec fingerprint per revision.

**Changes:**
- Add a SHA-256 “source digest” annotation on each Boxcutter revision and plumb it through revision state reading.
- Reuse a rolling-out revision only when its stored digest matches the current ClusterExtension catalog spec; otherwise re-resolve.
- Add E2E + controller-unit coverage for “spec changes during rollout triggers re-resolution” and digest normalization behavior.

## Backward compatibility

Revisions created before this PR have no fingerprint. The controller detects this and falls back to the old behavior until a new revision is created.

RFC: https://docs.google.com/document/d/1qNjdal3zaz7QTGHPAnPRO1aRnaum5iup0rRyiFar8CM/edit?tab=t.0#heading=h.x3tfh25grvnv
Motivated by: https://redhat.atlassian.net/browse/OCPBUGS-77829

PS: :jira-blocker: by thread: https://redhat-internal.slack.com/archives/C06KP34REFJ/p1774251840822449?thread_ts=1773998183.972329&cid=C06KP34REFJ

### DIFF
--- a/internal/operator-controller/controllers/clusterextension_reconcile_steps.go
+++ b/internal/operator-controller/controllers/clusterextension_reconcile_steps.go
@@ -18,9 +18,15 @@ package controllers
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/json"
 	"errors"
 	"fmt"
+	"slices"
+	"sort"
 
+	bsemver "github.com/blang/semver/v4"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -32,6 +38,7 @@ import (
 
 	ocv1 "github.com/operator-framework/operator-controller/api/v1"
 	"github.com/operator-framework/operator-controller/internal/operator-controller/bundleutil"
+	"github.com/operator-framework/operator-controller/internal/operator-controller/catalogmetadata/compare"
 	"github.com/operator-framework/operator-controller/internal/operator-controller/labels"
 	"github.com/operator-framework/operator-controller/internal/operator-controller/resolve"
 	imageutil "github.com/operator-framework/operator-controller/internal/shared/util/image"
@@ -140,14 +147,7 @@ func ResolveBundle(r resolve.Resolver, c client.Client) ReconcileStepFunc {
 	return func(ctx context.Context, state *reconcileState, ext *ocv1.ClusterExtension) (*ctrl.Result, error) {
 		l := log.FromContext(ctx)
 
-		// If already rolling out, use existing revision and set deprecation to Unknown (no catalog check)
-		if len(state.revisionStates.RollingOut) > 0 {
-			installedBundleName := ""
-			if state.revisionStates.Installed != nil {
-				installedBundleName = state.revisionStates.Installed.Name
-			}
-			SetDeprecationStatus(ext, installedBundleName, nil, false)
-			state.resolvedRevisionMetadata = state.revisionStates.RollingOut[0]
+		if reuseRollingOutRevision(ctx, state, ext) {
 			return nil, nil
 		}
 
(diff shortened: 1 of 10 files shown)


test/e2e/features/update.feature
```diff
@@ -243,3 +243,34 @@ Feature: Update ClusterExtension
...
+            packageName: test
+            selector:
+              matchLabels:
+                "olm.operatorframework.io/metadata.name": test-catalog
+            version: 1.0.0
+            upgradeConstraintPolicy: SelfCertified
+      """
+    And ClusterExtension is rolled out
+    And ClusterExtension is available
+    When ClusterExtension is updated to version "1.0.2"
```
pedjak: you can make the test faster if you initially try to install "1.0.2" version that stucks and then you install 1.0.1

pedjak (commented):
/lgtm 

IMHO, the added e2e test could be made faster.

test/e2e/features/update.feature
camilamacedo86: Done 👍

internal/operator-controller/controllers/clusterextension_reconcile_steps.go
```diff
@@ -199,6 +207,103 @@ func ResolveBundle(r resolve.Resolver, c client.Client) ReconcileStepFunc {
...
+// constraints are invalid/unparseable, in which case we should re-resolve from the catalog.
+func rollingOutRevisionMatchesSpec(rm *RevisionMetadata, ext *ocv1.ClusterExtension) bool {
+	if ext.Spec.Source.Catalog == nil {
+		return false
+	}
+
+	if rm.Package != ext.Spec.Source.Catalog.PackageName {
+		return false
+	}
+
```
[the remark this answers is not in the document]
camilamacedo86: Out of scope for this one.

test/e2e/features/update.feature
```diff
@@ -243,3 +243,69 @@ Feature: Update ClusterExtension
...
+              matchLabels:
+                "olm.operatorframework.io/metadata.name": test-catalog
+            version: 1.0.2
+            upgradeConstraintPolicy: SelfCertified
+      """
+    And ClusterExtensionRevision "${NAME}-1" reports Available as False with Reason ProbeFailure
+    When ClusterExtension is updated to version "1.0.1"
+    Then ClusterExtension is rolled out
+    And ClusterExtension is available
+    And bundle "test-operator.1.0.1" is installed in version "1.0.1"
```
joelanford: This is a potential downgrade from 1.0.2 to 1.0.1. What if the deployment for 1.0.2 included an initContainer that performed a data migration, which succeeded, but then the container that runs the 1.0.2 controller fails the probe because of a silly typo the user made in an envvar in their deploymentConfig?

We'd downgrade to 1.0.1, which would be unable to read the migrated data, and maybe 1.0.1 would make matters worse.

If we instead required human intervention, a human would find the log in the failing deployment, see it complaining about the envvar typo, and then go and fix it in the CE.spec.config.deploymentConfig. And then 1.0.2 would progress cleanly.

internal/operator-controller/controllers/clusterextension_reconcile_steps.go
```diff
@@ -199,6 +207,108 @@ func ResolveBundle(r resolve.Resolver, c client.Client) ReconcileStepFunc {
...
+// constraints are invalid/unparseable, in which case we should re-resolve from the catalog.
+func rollingOutRevisionMatchesSpec(rm *RevisionMetadata, ext *ocv1.ClusterExtension) bool {
+	if ext.Spec.Source.Catalog == nil {
+		return false
+	}
+	if rm.Package != ext.Spec.Source.Catalog.PackageName {
+		return false
+	}
+	return versionMatchesSpec(rm.Version, ext)
+}
```
joelanford: To me, this check feels way too implicit and unintuitive. If I understand it correctly, it requires that I specifically craft an updated version range to exclude _all_ of the versions of the active releases?

But if I expand the range to include a new version, no dice.

joelanford:
/hold
this one please!

I'm concerned that this could cause unsafe roll-forward or roll-back without the explicit acknowledgement from the user that forcing a re-resolution is an unsafe thing to do. IMO, default behavior needs to remain "wait for human intervention".

internal/operator-controller/controllers/clusterextension_reconcile_steps.go
camilamacedo86: any edit to resolution-relevant spec fields triggers re-resolution now. 
that was a good point

camilamacedo86:
Hi @joelanford 

I removed the second trigger via deadline
So, now it is only reconciled if the CE consumer explicitly change the spec. 

In this way, I will

/hold cancel

test/e2e/features/update.feature
camilamacedo86: We changed tat.
Now we have as a trigger just the spec changes ( as first proposal )

tmshort:
Hmm... the tests _passes_ in the `experimental-e2e`, but some other issue caused a failure; re-running.

tmshort:
> Hmm... the tests _passes_ in the `experimental-e2e`, but some other issue caused a failure; re-running.

And still failing?

camilamacedo86:
The failure does not seems related: https://github.com/operator-framework/operator-controller/actions/runs/26450862021/job/77877287902?pr=2566

But I am looking on that

camilamacedo86:
@pedjak @tmshort now it is passing in the e2e tests please fell free to give a look.
I think it is good to get a review.

tmshort:
What's the status of this one @pedjak @perdasilva ? Do we want it? If so, someone needs to take it over.

PR #2610, opened 2026-03-31
[Blocked by PR 2643] - 🐛 fix: (boxcutter) Enable archival and spec changes for ProgressDeadlineExceeded revisions

camilamacedo86:
## How it works on main

The COS reconciler checks if a revision took too long to roll out. If it did, it sets
`Progressing=False/ProgressDeadlineExceeded`.

To avoid a reconcile loop after that, `main` has a watch predicate that blocks all
updates to COS objects with ProgressDeadlineExceeded. The predicate only allows
deletions through.

## What is the problem?

The predicate blocks too much. It blocks all updates, not just status updates.

Example: a new revision rolls out and tries to archive the old stuck one by patching
`lifecycleState: Archived`. The predicate sees ProgressDeadlineExceeded and drops the
event. The old revision never gets archived.

### Example scenario

1. User installs a ClusterExtension. The CE controller creates COS-rev-1.
2. COS-rev-1 gets stuck (e.g. a Deployment never becomes ready). After
   `ProgressDeadlineMinutes`, the reconciler sets `Progressing=False/ProgressDeadlineExceeded`.
3. User updates the ClusterExtension. The CE controller creates COS-rev-2.
4. COS-rev-2 rolls out successfully. It patches COS-rev-1 with `lifecycleState: Archived`
   so the old revision gets cleaned up.
5. The watch predicate sees COS-rev-1 has ProgressDeadlineExceeded and drops the event.
   COS-rev-1 never reconciles, never processes the archival, and stays stuck forever.

## Why does the reconcile loop happen?

Every time the reconciler runs on a deadline-exceeded COS, it calls `markAsProgressing()`
which sets `Progressing=True`. Then the deadline check immediately sets it back to
`Progressing=False/ProgressDeadlineExceeded`. This back-and-forth produces a status
update, which triggers another reconcile, and so on.

The predicate was added to break this loop. But it also breaks archiving.

## What is the fix?

**1. Make ProgressDeadlineExceeded sticky (prevents reconcile loop)**

Once ProgressDeadlineExceeded is set, markAsProgressing() will not overwrite
it with RollingOut for Active revisions. This prevents the status flip-flop
that caused the reconcile loop.

The guard only blocks RollingOut. Other reasons (Succeeded, Retrying) can
still update the condition because they represent meaningful state changes.

With no flip-flop, there are no unnecessary status updates, so no reconcile
loop. The watch predicate is removed.

**2. Skip progress deadline enforcement for Archived revisions**

The deadline check now skips Archived revisions:

```
if pd > 0 && lifecycleState != Archived {
    // deadline check logic
}
```

This allows Archived revisions to retry teardown operations and update their
status without being overridden by the deadline check.

**3. Detect spec changes during rollout (resolution digest)**

A SHA256 digest of all catalog filter inputs (packageName, version, channels,
selector, upgradeConstraintPolicy) is stored in the revision's
olm.operatorframework.io/resolution-digest annotation.

Before reusing a rolling-out revision, the controller compares the current
spec's digest with the revision's digest. If they differ, a new revision is
created.

This ensures spec changes are not ignored when a revision has
ProgressDeadlineExceeded.

## How it works

1. Archival: A new successful revision can archive old revisions with
   ProgressDeadlineExceeded
2. Deletion: Users can delete ClusterExtensions with stuck revisions
3. Spec changes: Users can update catalog filters even while a revision is rolling out
4. Error visibility: Retrying conditions are visible even after deadline exceeded
5. No reconcile loops: Status updates stabilize once deadline is exceeded

## Motivation

Addresses feedback from:
https://github.com/openshift/operator-framework-operator-controller/pull/687#discussion_r3013433613

### DIFF
--- a/internal/operator-controller/controllers/clusterextension_controller_test.go
+++ b/internal/operator-controller/controllers/clusterextension_controller_test.go
@@ -785,7 +785,12 @@ func TestClusterExtensionBoxcutterApplierFailsDoesNotLeakDeprecationErrors(t *te
 		// the same inputs the runtime would see.
 		d.RevisionStatesGetter = &MockRevisionStatesGetter{
 			RevisionStates: &controllers.RevisionStates{
-				RollingOut: []*controllers.RevisionMetadata{{}},
+				RollingOut: []*controllers.RevisionMetadata{{
+					// Different digest from current spec - ensures we call the resolver
+					// (This simulates a rolling-out revision from a previous reconcile)
+					ResolutionDigest: "different-digest-from-previous-reconcile",
+					BundleMetadata:   ocv1.BundleMetadata{Version: "1.0.0"},
+				}},
 			},
 		}
 		d.Resolver = resolve.Func(func(ctx context.Context, ext *ocv1.ClusterExtension, installedBundle *ocv1.BundleMetadata) (*declcfg.Bundle, *bundle.VersionRelease, *declcfg.Deprecation, error) {
@@ -845,21 +850,18 @@ func TestClusterExtensionBoxcutterApplierFailsDoesNotLeakDeprecationErrors(t *te
 
 	deprecatedCond := apimeta.FindStatusCondition(clusterExtension.Status.Conditions, ocv1.TypeDeprecated)
 	require.NotNil(t, deprecatedCond)
-	require.Equal(t, metav1.ConditionUnknown, deprecatedCond.Status, "no catalog data during rollout, so Unknown")
-	require.Equal(t, ocv1.ReasonDeprecationStatusUnknown, deprecatedCond.Reason)
-	require.Equal(t, "deprecation status unknown: catalog data unavailable", deprecatedCond.Message)
+	require.Equal(t, metav1.ConditionFalse, deprecatedCond.Status, "catalog data available from resolution, not deprecated")
+	require.Equal(t, ocv1.ReasonNotDeprecated, deprecatedCond.Reason)
 
 	packageCond := apimeta.FindStatusCondition(clusterExtension.Status.Conditions, ocv1.TypePackageDeprecated)
 	require.NotNil(t, packageCond)
-	require.Equal(t, metav1.ConditionUnknown, packageCond.Status, "no catalog data during rollout, so Unknown")
-	require.Equal(t, ocv1.ReasonDeprecationStatusUnknown, packageCond.Reason)
-	require.Equal(t, "deprecation status unknown: catalog data unavailable", packageCond.Message)
+	require.Equal(t, metav1.ConditionFalse, packageCond.Status, "catalog data available from resolution, package not deprecated")
+	require.Equal(t, ocv1.ReasonNotDeprecated, packageCond.Reason)
 
 	channelCond := apimeta.FindStatusCondition(clusterExtension.Status.Conditions, ocv1.TypeChannelDeprecated)
 	require.NotNil(t, channelCond)
-	require.Equal(t, metav1.ConditionUnknown, channelCond.Status, "no catalog data during rollout, so Unknown")
-	require.Equal(t, ocv1.ReasonDeprecationStatusUnknown, channelCond.Reason)
-	require.Equal(t, "deprecation status unknown: catalog data unavailable", channelCond.Message)
+	require.Equal(t, metav1.ConditionFalse, channelCond.Status, "catalog data available from resolution, channel not deprecated")
+	require.Equal(t, ocv1.ReasonNotDeprecated, channelCond.Reason)
 
 	bundleCond := apimeta.FindStatusCondition(clusterExtension.Status.Conditions, ocv1.TypeBundleDeprecated)
 	require.NotNil(t, bundleCond)
--- a/internal/operator-controller/controllers/clusterextension_reconcile_steps.go
+++ b/internal/operator-controller/controllers/clusterextension_reconcile_steps.go
@@ -18,8 +18,12 @@ package controllers
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/base64"
+	"encoding/json"
 	"errors"
 	"fmt"
+	"slices"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
@@ -36,6 +40,58 @@ import (
 	"github.com/operator-framework/operator-controller/internal/operator-controller/resolve"
 	imageutil "github.com/operator-framework/operator-controller/internal/shared/util/image"
 )
+
+// resolutionInputs captures the catalog filter fields that affect bundle resolution.
+// Changes to any of these fields require re-resolution.
+type resolutionInputs struct {
+	PackageName             string   `json:"packageName"`
+	Version                 string   `json:"version"`
+	Channels                []string `json:"channels,omitempty"`
+	Selector                string   `json:"selector,omitempty"`
+	UpgradeConstraintPolicy string   `json:"upgradeConstraintPolicy,omitempty"`
+}
+
+// calculateResolutionDigest computes a SHA256 hash of the catalog filter inputs
+// that affect bundle resolution. This digest enables detection of spec changes that
+// require re-resolution versus when a rolling-out revision can be reused.
+//
+// The digest is base64 URL-safe encoded (43 characters) to fit within Kubernetes
+// label value limits (63 characters max) when stored in Helm release metadata.
+func calculateResolutionDigest(catalog *ocv1.CatalogFilter) (string, error) {
+	if catalog == nil {
+		return "", nil
+	}
+
+	// Sort channels for deterministic hashing
+	channels := slices.Clone(catalog.Channels)
+	slices.Sort(channels)
+
+	inputs := resolutionInputs{
+		PackageName:             catalog.PackageName,
+		Version:                 catalog.Version,
+		Channels:                channels,
+		UpgradeConstraintPolicy: string(catalog.UpgradeConstraintPolicy),
+	}
+
+	// Convert selector to canonical string representation for deterministic hashing
+	if catalog.Selector != nil {
+		selector, err := metav1.LabelSelectorAsSelector(catalog.Selector)
+		if err != nil {
+			return "", fmt.Errorf("converting selector: %w", err)
+		}
+		inputs.Selector = selector.String()
+	}
+
+	// Marshal to JSON for consistent hashing
+	inputsBytes, err := json.Marshal(inputs)
+	if err != nil {
+		return "", fmt.Errorf("marshaling resolution inputs: %w", err)
+	}
+
+	// Compute SHA256 hash and encode as base64 URL-safe (43 chars, fits in label value limit)
+	hash := sha256.Sum256(inputsBytes)
+	return base64.RawURLEncoding.EncodeToString(hash[:]), nil
+}
 
 func HandleFinalizers(f finalizer.Finalizer) ReconcileStepFunc {
 	return func(ctx context.Context, state *reconcileState, ext *ocv1.ClusterExtension) (*ctrl.Result, error) {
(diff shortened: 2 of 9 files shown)


internal/operator-controller/controllers/clusterobjectset_controller.go
```diff
@@ -344,21 +344,44 @@ type Sourcoser interface {
...
+	}
+
+	// Suppress status-only updates
+	return false
+}
+
 func (c *ClusterObjectSetReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	skipProgressDeadlineExceededPredicate := predicate.Funcs{
 		UpdateFunc: func(e event.UpdateEvent) bool {
-			rev, ok := e.ObjectNew.(*ocv1.ClusterObjectSet)
```
pedjak, line 338: I would extract the whole update func into standalone function, not just the logic under `shouldAllowProgressDeadlineExceededUpdate`

camilamacedo86: Done

joelanford:
Maybe I misunderstand the progress deadline exceeded stuff. But I thought exceeding the deadline literally only meant that we set Progressing=False?

But otherwise all other functionality and decision making around reconciliation of the CE works the same as it does prior to the deadline exceeding.

If that's not the case, why?

internal/operator-controller/controllers/clusterobjectset_controller.go
camilamacedo86: @pedjak 
I think we might can remove the predicate.
See the current proposal now.
WDYT? I tried to update the PR desc as well

camilamacedo86:
Hi @joelanford 

I changed the proposal here. I think it might be a better way to solve it.
Also, with the Claude I update the description with a description of this.
I tried to answer your questions. Thank you for your time.

pedjak (commented):
**Three suggestions:**

1. **Add a comment explaining the `ReasonSucceeded` carve-out.** The guard uses `reason != ReasonSucceeded`, which means any new reason added in the future is silently blocked by default. That's the safe default, but it's a hidden constraint — a short comment explaining *why* only `Succeeded` is exempt would prevent a future contributor from accidentally breaking this.

2. **Add a debug log in the early return.** When the guard fires, it silently swallows a state transition with zero signal. A `V(1)` log line like `"skipping markAsProgressing: ProgressDeadlineExceeded is sticky"` would help debugging when someone is troubleshooting why a COS condition isn't updating.

3. **Consider an integration test for the archival scenario (follow-up).** The unit tests cover `markAsProgressing` well, but the motivating bug (archival blocked after deadline exceeded) isn't tested e2e. A test simulating deadline exceeded → `lifecycleState: Archived` → verify reconcile proceeds would cover the actual user-facing scenario.

joelanford:
> Only status-only updates (same generation) were suppressed.

IMO, these should not be suppressed either. If some other user or component changes the status of a COS after deadline exceeded, we should reconcile the COS and put the correct status back in place.

I propose that we fully remove the predicate and continue reconciling _ALL_ COS events no matter whether deadline exceeded or not.

internal/operator-controller/controllers/clusterextension_reconcile_steps.go
```diff
@@ -188,9 +261,17 @@ func ResolveBundle(r resolve.Resolver, c client.Client) ReconcileStepFunc {
 			return handleResolutionError(ctx, c, state, ext, err)
 		}
 
+		// Calculate digest of the resolution inputs for change detection
+		digest, err := calculateResolutionDigest(ext.Spec.Source.Catalog)
+		if err != nil {
+			l.Error(err, "failed to calculate resolution digest, continuing without it")
```
pedjak, line 267: If we can continue, perhaps this should be just info?

internal/operator-controller/controllers/clusterextension_reconcile_steps.go
```diff
@@ -37,6 +41,58 @@ import (
...
+	Selector                string   `json:"selector,omitempty"`
+	UpgradeConstraintPolicy string   `json:"upgradeConstraintPolicy,omitempty"`
+}
+
+// calculateResolutionDigest computes a SHA256 hash of the catalog filter inputs
+// that affect bundle resolution. This digest enables detection of spec changes that
+// require re-resolution versus when a rolling-out revision can be reused.
+//
+// The digest is base64 URL-safe encoded (43 characters) to fit within Kubernetes
+// label value limits (63 characters max) when stored in Helm release metadata.
```
pedjak, line 59: The doc comment says "to fit within Kubernetes label value limits (63 characters max) when stored in Helm release metadata" — but in the boxcutter path, `ResolutionDigestKey` is stored as an **annotation** (in `revisionAnnotations` maps), not a label. Annotations don't have the 63-char limit.

It *is* stored as a Helm release label via `rel.Labels[labels.ResolutionDigestKey]` in `clusterextension_controller.go`, so the 63-char constraint matters there. Consider clarifying that the size constraint comes from the Helm label path specifically.

internal/operator-controller/controllers/clusterextension_reconcile_steps.go
```diff
@@ -140,15 +196,32 @@ func ResolveBundle(r resolve.Resolver, c client.Client) ReconcileStepFunc {
...
+		//
+		// Note: RollingOut slice is sorted oldest→newest, so use the last element (most recent).
+		// This avoids re-resolving when a newer revision already matches the spec, even if an
+		// older revision is still stuck rolling out.
 		if len(state.revisionStates.RollingOut) > 0 {
-			installedBundleName := ""
-			if state.revisionStates.Installed != nil {
-				installedBundleName = state.revisionStates.Installed.Name
+			latestRollingOutRevision := state.revisionStates.RollingOut[len(state.revisionStates.RollingOut)-1]
+
```
pedjak, line 208: This changes from `RollingOut[0]` (oldest) to `RollingOut[len-1]` (newest). Since the slice is sorted ascending by `Spec.Revision`, this is correct for digest comparison — you want the most recent revision.

However, it's also a behavioral change beyond the digest feature: even when digests match, the *metadata* propagated to `state.resolvedRevisionMetadata` now comes from the newest rolling-out revision instead of the oldest. Worth calling out in the PR description.

internal/operator-controller/controllers/clusterextension_reconcile_steps.go
```diff
@@ -188,9 +261,17 @@ func ResolveBundle(r resolve.Resolver, c client.Client) ReconcileStepFunc {
 			return handleResolutionError(ctx, c, state, ext, err)
 		}
 
+		// Calculate digest of the resolution inputs for change detection
+		digest, err := calculateResolutionDigest(ext.Spec.Source.Catalog)
+		if err != nil {
+			l.Error(err, "failed to calculate resolution digest, continuing without it")
+			digest = ""
```
pedjak, line 268: If `calculateResolutionDigest` fails here, `digest` is set to `""` and stored in the annotation. On the next reconcile, `calculateResolutionDigest` for the same (valid) spec would succeed, producing a non-empty digest that won't match `""`, causing an unnecessary re-resolution. This is safe (self-healing) but might be worth a brief comment noting the behavior.

internal/operator-controller/controllers/clusterobjectset_controller.go
```diff
@@ -638,7 +623,33 @@ func setRetryingConditions(cos *ocv1.ClusterObjectSet, message string) {
...
+// - Retrying represents transient errors that need to be visible in status
+//
+// If the generation changed since ProgressDeadlineExceeded was recorded, the guard
+// allows all updates through so the condition reflects the new generation.
 func markAsProgressing(cos *ocv1.ClusterObjectSet, reason, message string) {
+	if cnd := meta.FindStatusCondition(cos.Status.Conditions, ocv1.ClusterObjectSetTypeProgressing); cnd != nil &&
+		cnd.Status == metav1.ConditionFalse && cnd.Reason == ocv1.ReasonProgressDeadlineExceeded &&
+		cnd.ObservedGeneration == cos.Generation &&
+		reason == ocv1.ReasonRollingOut &&
+		cos.Spec.LifecycleState != ocv1.ClusterObjectSetLifecycleStateArchived {
```
pedjak, line 648: Good design — the guard is narrow and well-targeted. Only `RollingOut` is blocked to prevent the specific flip-flop loop; `Succeeded` and `Retrying` pass through, and the `ObservedGeneration` check resets everything on new spec revisions.

One thought: if a new reason were added in the future without updating this guard, it could accidentally bypass the deadline. Consider adding a brief comment noting the assumption about the current set of possible reasons, so future authors know to check here.

internal/operator-controller/controllers/clusterobjectset_controller_internal_test.go
```diff
@@ -235,3 +236,113 @@ func (m *mockTrackingCacheInternal) Watch(ctx context.Context, user client.Objec
 func (m *mockTrackingCacheInternal) Source(h handler.EventHandler, predicates ...predicate.Predicate) source.Source {
 	return nil
 }
+
+func Test_markAsProgressing_doesNotOverwriteProgressDeadlineExceeded(t *testing.T) {
```
pedjak, line 240: Good table-driven coverage of the sticky behavior.

Suggestion: consider also adding unit tests for `calculateResolutionDigest` — e.g., nil catalog, empty channels, selector canonicalization, and determinism (same inputs → same digest across calls). That function is load-bearing for spec-change detection and currently has no direct tests.

internal/operator-controller/controllers/clusterextension_controller_test.go
```diff
@@ -845,21 +850,18 @@ func TestClusterExtensionBoxcutterApplierFailsDoesNotLeakDeprecationErrors(t *te
 
 	deprecatedCond := apimeta.FindStatusCondition(clusterExtension.Status.Conditions, ocv1.TypeDeprecated)
 	require.NotNil(t, deprecatedCond)
-	require.Equal(t, metav1.ConditionUnknown, deprecatedCond.Status, "no catalog data during rollout, so Unknown")
-	require.Equal(t, ocv1.ReasonDeprecationStatusUnknown, deprecatedCond.Reason)
-	require.Equal(t, "deprecation status unknown: catalog data unavailable", deprecatedCond.Message)
+	require.Equal(t, metav1.ConditionFalse, deprecatedCond.Status, "catalog data available from resolution, not deprecated")
```
pedjak, line 853: Worth noting: this is a **behavioral change** beyond the bug fix. Previously, when a revision was rolling out, deprecation status was always Unknown ("no catalog data during rollout"). Now, when the spec changes and triggers re-resolution, deprecation is set based on actual catalog data.

This is an improvement — the test updates are correct — but it's a side effect worth calling out in the PR description since it changes user-visible condition values.

joelanford:
I might still be missing something, but I feel like this is a ton of complexity being added to solve a problem that we've created for ourselves.

> Why does the reconcile loop happen?
>
> Every time the reconciler runs on a deadline-exceeded COS, it calls markAsProgressing()
> which sets Progressing=True. Then the deadline check immediately sets it back to
> Progressing=False/ProgressDeadlineExceeded. This back-and-forth produces a status
> update, which triggers another reconcile, and so on.
>
> The predicate was added to break this loop. But it also breaks archiving.

It seems like we should not actually mark a deadline-exceeded COS as `Progressing=True` unless the original reason or message for blocking the progression has changed. Or maybe unless the entire COS flips to Ready=True.

IMO, there should be no predicate that blocks reconciliation of deadline-exceeded COS or for any other reason. All informer events should always be fully reconciled. As we've discovered, this predicate has lead to more problems (like not properly processing an archival), and when we try to band-aid over those problems, we end up building more complexity that has the possibility of introducing more bugs, and it makes the entire system more difficult to maintain over time.

I think the way to solve the "not archiving properly" problem is to remove the predicate. If that causes a reconcile loop, let's fix the root cause of the reconcile loop.

joelanford:
Another thing I know @perdasilva and I have discussed, not sure if you all have talked about it as a group:

Get rid of the COS revisions other than `Ready`. The COS is either fully applied and passing all health probes, or it isn't. All the deadline progression logic then lives only in the CE logic/reconciler/status.

joelanford:
Hey @camilamacedo86, following up on my last comments — I went ahead and prototyped the approach I was describing (see #2643).

The idea is:

1. **Remove the predicate entirely** — no special-casing of any COS events.

2. **Fix the root cause of the reconcile loop in `markAsProgressing`** — when the deadline has been exceeded, non-terminal reasons (`RollingOut`, `Retrying`) set `ProgressDeadlineExceeded` instead. `Succeeded` always applies. This way the condition never flaps, so there's no loop to break.

3. **Continue reconciling after `ProgressDeadlineExceeded`** — previously, when the deadline was hit, we cleared the reconcile error and stopped requeueing. Now we let the normal error/backoff continue. This means if a transient error resolves itself, the revision can still succeed even after the deadline was exceeded.

4. **Add a deadline-aware rate limiter** to cap exponential backoff at the deadline, so `ProgressDeadlineExceeded` is set promptly even during error retries.

The fix is scoped entirely to the COS controller — no changes to the CE reconciler or resolution logic needed.

Let me know what you think!

camilamacedo86:
Closed in favor of: 2643

PR #2616, opened 2026-03-31
🐛 OCPBUGS-81452: prevent webhook rollout stalls by issuing certificates earlier

bandrade:
## What changed
- move cert-manager `Certificate` objects into the `infrastructure` phase
- keep `Issuer` in `infrastructure` and leave workload `Deployment` objects in `deploy`
- update phase sorting tests to cover the new ordering

## Why
Webhook installs can stall in `RollingOut` when certificate issuance and deployment availability are gated within the same phase. Starting certificate issuance earlier shortens the rollout critical path for webhook-backed operators.

## Impact
This reduces the chance that `ClusterExtension` installations for webhook operators remain stuck waiting for the generated deployment to become available.

## Root cause
The rollout path allowed cert-manager `Certificate` objects to be applied in the same phase as the webhook deployment. For operators that mount the serving cert secret, that can delay deployment availability long enough to hit external rollout timeouts.

## Validation
- `go test -tags containers_image_openpgp ./internal/operator-controller/applier -run Test_PhaseSort`

### DIFF
--- a/internal/operator-controller/applier/phase.go
+++ b/internal/operator-controller/applier/phase.go
@@ -120,10 +120,10 @@ var (
 		PhaseInfrastructure: {
 			{Kind: "Service"},
 			{Kind: "Issuer", Group: "cert-manager.io"},
+			{Kind: "Certificate", Group: "cert-manager.io"},
 		},
 
 		PhaseDeploy: {
-			{Kind: "Certificate", Group: "cert-manager.io"},
 			{Kind: "Deployment", Group: "apps"},
 		},
 
--- a/internal/operator-controller/applier/phase_test.go
+++ b/internal/operator-controller/applier/phase_test.go
@@ -168,6 +168,17 @@ func Test_PhaseSort(t *testing.T) {
 				{
 					Object: &unstructured.Unstructured{
 						Object: map[string]interface{}{
+							"apiVersion": "cert-manager.io/v1",
+							"kind":       "Certificate",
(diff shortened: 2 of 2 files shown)


camilamacedo86:
/lgtm

But would be nice get a PTAL from @perdasilva @pedjak @tmshort

pedjak (commented):
It would be really nice to craft an e2e test that demonstrate that the issue is fixed, or this is already covered by webhook-operator bundle?

tmshort:
> It would be really nice to craft an e2e test that demonstrate that the issue is fixed, or this is already covered by webhook-operator bundle?

There's already a unit test for this; which we prefer over e2e tests.

@bandrade ?

tmshort:
> It would be really nice to craft an e2e test that demonstrate that the issue is fixed, or this is already covered by webhook-operator bundle?

@bandrade 
If this is not already covered by an existing test, then a unit-test would be preferred, to avoid extending the time it takes to run e2e's. (@joelanford has expressed a preference for unit over e2e in this project.)

bandrade:
@pedjak @tmshort I added a focused unit test for this path so it’s covered more directly now: it verifies that the Certificate is applied in infrastructure before the webhook Deployment, and I kept it as a unit test rather than adding an e2e to stay aligned with the preference for faster, narrower coverage here.

pedjak (approved):
/approve

PR #2643, opened 2026-04-11
🐛 fix: allow reconciliation of deadline-exceeded ClusterObjectSets

joelanford:
# Description

The `skipProgressDeadlineExceededPredicate` blocked all update events for COS objects
with `ProgressDeadlineExceeded`, which prevented archival of stuck revisions — the
lifecycle state patch was silently dropped.

This PR:
- **Removes the predicate** so all COS events are fully reconciled
- **Updates `markAsProgressing`** to set `ProgressDeadlineExceeded` instead of
  `RollingOut`/`Retrying` when the deadline has been exceeded, preventing the reconcile
  loop the predicate was masking
- **Continues reconciling after `ProgressDeadlineExceeded`** rather than clearing the
  error and stopping requeue. This allows revisions to recover if a transient error
  resolves itself, even after the deadline was exceeded
- **Extracts `durationUntilDeadline`** as a shared helper for deadline computation
- **Adds a `deadlineAwareRateLimiter`** that caps exponential backoff at the deadline so
  `ProgressDeadlineExceeded` is set promptly even during error retries
- **Moves deadline requeue logic** into `reconcile`, using the remaining duration from
  `durationUntilDeadline` directly for `RequeueAfter`
- **Adds unit tests** for `durationUntilDeadline` and recovery from
  `ProgressDeadlineExceeded` to `Succeeded`
- **Adds e2e tests** for `ProgressDeadlineExceeded` archival/cleanup and for recovery
  from `ProgressDeadlineExceeded` to `Succeeded`

Addresses feedback from:
https://github.com/operator-framework/operator-controller/pull/2610#issuecomment-4226236991

## Reviewer Checklist

### DIFF
--- a/internal/operator-controller/controllers/clusterobjectset_controller.go
+++ b/internal/operator-controller/controllers/clusterobjectset_controller.go
@@ -14,6 +14,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -136,7 +137,7 @@ func (c *ClusterObjectSetReconciler) reconcile(ctx context.Context, cos *ocv1.Cl
 
 	phases, currentPhases, opts, err := c.buildBoxcutterPhases(ctx, cos)
 	if err != nil {
-		setRetryingConditions(cos, err.Error(), isDeadlineExceeded)
+		setRetryingConditions(l, cos, err.Error(), isDeadlineExceeded)
 		return ctrl.Result{}, fmt.Errorf("converting to boxcutter revision: %v", err)
 	}
 
@@ -150,7 +151,7 @@ func (c *ClusterObjectSetReconciler) reconcile(ctx context.Context, cos *ocv1.Cl
 
 	revisionEngine, err := c.RevisionEngineFactory.CreateRevisionEngine(ctx, cos)
 	if err != nil {
-		setRetryingConditions(cos, err.Error(), isDeadlineExceeded)
+		setRetryingConditions(l, cos, err.Error(), isDeadlineExceeded)
 		return ctrl.Result{}, fmt.Errorf("failed to create revision engine: %v", err)
 	}
 
@@ -176,7 +177,7 @@ func (c *ClusterObjectSetReconciler) reconcile(ctx context.Context, cos *ocv1.Cl
 
 	if err := c.establishWatch(ctx, cos, revision); err != nil {
 		werr := fmt.Errorf("establish watch: %v", err)
-		setRetryingConditions(cos, werr.Error(), isDeadlineExceeded)
+		setRetryingConditions(l, cos, werr.Error(), isDeadlineExceeded)
 		return ctrl.Result{}, werr
 	}
 
@@ -186,7 +187,7 @@ func (c *ClusterObjectSetReconciler) reconcile(ctx context.Context, cos *ocv1.Cl
 			// Log detailed reconcile reports only in debug mode (V(1)) to reduce verbosity.
 			l.V(1).Info("reconcile report", "report", rres.String())
 		}
-		setRetryingConditions(cos, err.Error(), isDeadlineExceeded)
+		setRetryingConditions(l, cos, err.Error(), isDeadlineExceeded)
 		return ctrl.Result{}, fmt.Errorf("revision reconcile: %v", err)
 	}
 
@@ -194,14 +195,14 @@ func (c *ClusterObjectSetReconciler) reconcile(ctx context.Context, cos *ocv1.Cl
 	// TODO: report status, backoff?
 	if verr := rres.GetValidationError(); verr != nil {
 		l.Error(fmt.Errorf("%w", verr), "preflight validation failed, retrying after 10s")
-		setRetryingConditions(cos, fmt.Sprintf("revision validation error: %s", verr), isDeadlineExceeded)
+		setRetryingConditions(l, cos, fmt.Sprintf("revision validation error: %s", verr), isDeadlineExceeded)
 		return ctrl.Result{RequeueAfter: 10 * time.Second}, nil
 	}
 
 	for i, pres := range rres.GetPhases() {
 		if verr := pres.GetValidationError(); verr != nil {
 			l.Error(fmt.Errorf("%w", verr), "phase preflight validation failed, retrying after 10s", "phase", i)
-			setRetryingConditions(cos, fmt.Sprintf("phase %d validation error: %s", i, verr), isDeadlineExceeded)
+			setRetryingConditions(l, cos, fmt.Sprintf("phase %d validation error: %s", i, verr), isDeadlineExceeded)
 			return ctrl.Result{RequeueAfter: 10 * time.Second}, nil
 		}
 
@@ -214,14 +215,14 @@ func (c *ClusterObjectSetReconciler) reconcile(ctx context.Context, cos *ocv1.Cl
 
 		if len(collidingObjs) > 0 {
 			l.Error(fmt.Errorf("object collision detected"), "object collision, retrying after 10s", "phase", i, "collisions", collidingObjs)
-			setRetryingConditions(cos, fmt.Sprintf("revision object collisions in phase %d\n%s", i, strings.Join(collidingObjs, "\n\n")), isDeadlineExceeded)
+			setRetryingConditions(l, cos, fmt.Sprintf("revision object collisions in phase %d\n%s", i, strings.Join(collidingObjs, "\n\n")), isDeadlineExceeded)
 			return ctrl.Result{RequeueAfter: 10 * time.Second}, nil
 		}
 	}
 
 	revVersion := cos.GetAnnotations()[labels.BundleVersionKey]
 	if rres.InTransition() {
-		markAsProgressing(cos, ocv1.ReasonRollingOut, fmt.Sprintf("Revision %s is rolling out.", revVersion), isDeadlineExceeded)
+		markAsProgressing(l, cos, ocv1.ReasonRollingOut, fmt.Sprintf("Revision %s is rolling out.", revVersion), isDeadlineExceeded)
 	}
 
 	//nolint:nestif
@@ -241,7 +242,7 @@ func (c *ClusterObjectSetReconciler) reconcile(ctx context.Context, cos *ocv1.Cl
 			}
 		}
 
-		markAsProgressing(cos, ocv1.ReasonSucceeded, fmt.Sprintf("Revision %s has rolled out.", revVersion), isDeadlineExceeded)
+		markAsProgressing(l, cos, ocv1.ReasonSucceeded, fmt.Sprintf("Revision %s has rolled out.", revVersion), isDeadlineExceeded)
 		markAsAvailable(cos, ocv1.ClusterObjectSetReasonProbesSucceeded, "Objects are available and pass all probes.")
 
 		// We'll probably only want to remove this once we are done updating the ClusterExtension conditions
@@ -286,7 +287,7 @@ func (c *ClusterObjectSetReconciler) reconcile(ctx context.Context, cos *ocv1.Cl
 		} else {
 			markAsUnavailable(cos, ocv1.ReasonRollingOut, fmt.Sprintf("Revision %s is rolling out.", revVersion))
 		}
-		markAsProgressing(cos, ocv1.ReasonRollingOut, fmt.Sprintf("Revision %s is rolling out.", revVersion), isDeadlineExceeded)
+		markAsProgressing(l, cos, ocv1.ReasonRollingOut, fmt.Sprintf("Revision %s is rolling out.", revVersion), isDeadlineExceeded)
 		if hasDeadline && !isDeadlineExceeded {
 			return ctrl.Result{RequeueAfter: remaining}, nil
 		}
@@ -307,14 +308,15 @@ func (c *ClusterObjectSetReconciler) delete(ctx context.Context, cos *ocv1.Clust
 }
 
 func (c *ClusterObjectSetReconciler) archive(ctx context.Context, revisionEngine RevisionEngine, cos *ocv1.ClusterObjectSet, revision boxcutter.RevisionBuilder) (ctrl.Result, error) {
+	l := log.FromContext(ctx)
 	tdres, err := revisionEngine.Teardown(ctx, revision)
 	if err != nil {
 		err = fmt.Errorf("error archiving revision: %v", err)
-		setRetryingConditions(cos, err.Error(), false)
+		setRetryingConditions(l, cos, err.Error(), false)
 		return ctrl.Result{}, err
 	}
 	if tdres != nil && !tdres.IsComplete() {
-		setRetryingConditions(cos, "removing revision resources that are not owned by another revision", false)
+		setRetryingConditions(l, cos, "removing revision resources that are not owned by another revision", false)
 		return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
 	}
 	// Ensure conditions are set before removing the finalizer when archiving
@@ -632,8 +634,8 @@ func buildProgressionProbes(progressionProbes []ocv1.ProgressionProbe) (probing.
 	return userProbes, nil
 }
 
-func setRetryingConditions(cos *ocv1.ClusterObjectSet, message string, isDeadlineExceeded bool) {
-	markAsProgressing(cos, ocv1.ClusterObjectSetReasonRetrying, message, isDeadlineExceeded)
+func setRetryingConditions(l logr.Logger, cos *ocv1.ClusterObjectSet, message string, isDeadlineExceeded bool) {
+	markAsProgressing(l, cos, ocv1.ClusterObjectSetReasonRetrying, message, isDeadlineExceeded)
 	if meta.FindStatusCondition(cos.Status.Conditions, ocv1.ClusterObjectSetTypeAvailable) != nil {
 		markAsAvailableUnknown(cos, ocv1.ClusterObjectSetReasonReconciling, message)
 	}
@@ -644,13 +646,13 @@ var nonTerminalProgressingReasons = map[string]struct{}{
 	ocv1.ClusterObjectSetReasonRetrying: {},
 }
 
-func markAsProgressing(cos *ocv1.ClusterObjectSet, reason, message string, isDeadlineExceeded bool) {
+func markAsProgressing(l logr.Logger, cos *ocv1.ClusterObjectSet, reason, message string, isDeadlineExceeded bool) {
 	switch reason {
 	case ocv1.ReasonSucceeded:
 		// Terminal — always apply.
 	default:
 		if _, known := nonTerminalProgressingReasons[reason]; !known {
-			log.Log.Error(fmt.Errorf("unregistered progressing reason: %q", reason), "treating as non-terminal for deadline enforcement")
+			l.Error(fmt.Errorf("unregistered progressing reason: %q", reason), "treating as non-terminal for deadline enforcement")
 		}
 		if isDeadlineExceeded {
 			markAsNotProgressing(cos, ocv1.ReasonProgressDeadlineExceeded,
--- a/internal/operator-controller/controllers/clusterobjectset_controller.go
+++ b/internal/operator-controller/controllers/clusterobjectset_controller.go
@@ -127,6 +127,7 @@ func (c *ClusterObjectSetReconciler) reconcile(ctx context.Context, cos *ocv1.Cl
 	remaining, hasDeadline := durationUntilDeadline(c.Clock, cos)
 	isDeadlineExceeded := hasDeadline && remaining <= 0
 
+	// Blocked takes precedence over ProgressDeadlineExceeded: it is more actionable for the user.
 	if err := c.verifyReferencedSecretsImmutable(ctx, cos); err != nil {
 		l.Error(err, "referenced Secret verification failed, blocking reconciliation")
 		markAsNotProgressing(cos, ocv1.ClusterObjectSetReasonBlocked, err.Error())
--- a/internal/operator-controller/controllers/clusterobjectset_controller.go
+++ b/internal/operator-controller/controllers/clusterobjectset_controller.go
@@ -639,17 +639,17 @@ func setRetryingConditions(cos *ocv1.ClusterObjectSet, message string, isDeadlin
 	}
 }
 
-func markAsProgressing(cos *ocv1.ClusterObjectSet, reason, message string, isDeadlineExceeded bool) {
-	nonTerminalReasons := map[string]struct{}{
-		ocv1.ReasonRollingOut:               {},
-		ocv1.ClusterObjectSetReasonRetrying: {},
-	}
+var nonTerminalProgressingReasons = map[string]struct{}{
+	ocv1.ReasonRollingOut:               {},
+	ocv1.ClusterObjectSetReasonRetrying: {},
+}
 
+func markAsProgressing(cos *ocv1.ClusterObjectSet, reason, message string, isDeadlineExceeded bool) {
 	switch reason {
 	case ocv1.ReasonSucceeded:
 		// Terminal — always apply.
 	default:
-		if _, known := nonTerminalReasons[reason]; !known {
+		if _, known := nonTerminalProgressingReasons[reason]; !known {
 			log.Log.Error(fmt.Errorf("unregistered progressing reason: %q", reason), "treating as non-terminal for deadline enforcement")
 		}
 		if isDeadlineExceeded {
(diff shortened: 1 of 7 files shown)


test/e2e/features/revision.feature
```diff
@@ -442,4 +442,71 @@ Feature: Install ClusterObjectSet
...
+                      image: does-not-exist:latest
+        revision: 1
+      """
+    Then resource "configmap/test-configmap" is installed
+    And resource "deployment/test-deployment" is installed
+    And ClusterObjectSet "${COS_NAME}" reports Progressing as False with Reason ProgressDeadlineExceeded
+    When ClusterObjectSet "${COS_NAME}" lifecycle is set to "Archived"
+    Then ClusterObjectSet "${COS_NAME}" is archived
+    And resource "configmap/test-configmap" is eventually not found
+    And resource "deployment/test-deployment" is eventually not found
```
camilamacedo86, line 512: It does not seems to be testing the same scenario @joelanford 
Could we ensure the same scenario here?

camilamacedo86: 1. User installs a ClusterExtension. The CE controller creates COS-rev-1.
2. COS-rev-1 gets stuck (e.g. a Deployment never becomes ready). After ProgressDeadlineMinutes, the reconciler sets Progressing=False/ProgressDeadlineExceeded.
4. User updates the ClusterExtension. The CE controller creates COS-rev-2.
5. COS-rev-2 rolls out successfully. It patches COS-rev-1 with lifecycleState: Archived so the old revision gets cleaned up.
7. The watch predicate sees COS-rev-1 has ProgressDeadlineExceeded and drops the event.
8. COS-rev-1 never reconciles, never processes the archival, and stays stuck forever.

joelanford: Is it not the same? Making a CE that stamps out the COS-1 and COS-2 such that the CE reconciler eventually tries to set COS-1 as archived is the same thing, but just more ceremony around a standalone COS that exceeds the deadline and then is directly archived by the test code.

I'll verify manually to make sure. Is there something else happening with the CE/COS interaction that would make the COS-only test in the PR different?

internal/operator-controller/controllers/clusterobjectset_controller.go
```diff
@@ -367,6 +346,51 @@ func (c *ClusterObjectSetReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	clk clock.Clock,
+) *deadlineAwareRateLimiter {
+	return &deadlineAwareRateLimiter{delegate: delegate, client: c, clock: clk}
+}
+
+func (r *deadlineAwareRateLimiter) When(item ctrl.Request) time.Duration {
+	backoff := r.delegate.When(item)
+
+	cos := &ocv1.ClusterObjectSet{}
+	if err := r.client.Get(context.Background(), item.NamespacedName, cos); err != nil {
```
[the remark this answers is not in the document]
joelanford: Our use of the rate limiter uses the informer cache which is in-memory and does not block.

internal/operator-controller/controllers/clusterobjectset_controller.go
```diff
@@ -333,29 +323,19 @@ type Sourcoser interface {
...
-			if !rev.DeletionTimestamp.IsZero() {
-				return true
-			}
-			if cnd := meta.FindStatusCondition(rev.Status.Conditions, ocv1.ClusterObjectSetTypeProgressing); cnd != nil && cnd.Status == metav1.ConditionFalse && cnd.Reason == ocv1.ReasonProgressDeadlineExceeded {
-				return false
-			}
-			return true
-		},
-	}
 	c.Clock = clock.RealClock{}
```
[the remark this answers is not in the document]
joelanford: Not something I changed in this PR, so out-of-scope. Regardless, caller's don't call SetupWithManager in tests, so this isn't a problem in practice.

internal/operator-controller/controllers/clusterobjectset_controller.go
```diff
@@ -367,6 +347,51 @@ func (c *ClusterObjectSetReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	clk clock.Clock,
+) *deadlineAwareRateLimiter {
+	return &deadlineAwareRateLimiter{delegate: delegate, client: c, clock: clk}
+}
+
+func (r *deadlineAwareRateLimiter) When(item ctrl.Request) time.Duration {
+	backoff := r.delegate.When(item)
+
+	cos := &ocv1.ClusterObjectSet{}
+	if err := r.client.Get(context.Background(), item.NamespacedName, cos); err != nil {
```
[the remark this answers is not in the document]
joelanford: Our use of the rate limiter uses the informer cache which is in-memory and does not block.

test/e2e/features/revision.feature
camilamacedo86: we are not upgrading, checking it stuck and then, checking that is possible to change the resource and the changes will be reconciled. WDY about we add a test with?

internal/operator-controller/controllers/clusterobjectset_controller.go
```diff
@@ -144,15 +122,18 @@ func (c *ClusterObjectSetReconciler) reconcile(ctx context.Context, cos *ocv1.Cl
 		return c.delete(ctx, cos)
 	}
 
+	remaining, hasDeadline := c.durationUntilDeadline(cos)
+	isDeadlineExceeded := hasDeadline && remaining <= 0
+
```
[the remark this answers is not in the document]
joelanford: This is a good point. `requeueForDeadline` should not perform the calculation again with a different "now" timestamp. Perhaps instead we could simply pass the result of the original calculation into requeueForDeadline.

camilamacedo86:
@joelanford can you please rebase this one and address the Copilot comments so that we can review and see if we can go with.

joelanford:
Ran out of time to get back to this today. Hopefully tomorrow!

This PR has some flaws around recomputation of the remaining time until deadline which can cause funky things to happen if we're reconciling just before the deadline and then cross over the deadline threshold prior to the rate limiter logic running, which can cause requeues not to happen right at the deadline.

Might need a bit more logic in the recomputation code before it's ready. Hopefully that extra code isn't too brittle. Maybe it will be fine as a stopgap if the plan is to eventually move the progress deadline stuff to be a concern only of the ClusterExtension (leaving ClusterObjectSet with just a single "Ready" or similar condition)

internal/operator-controller/controllers/clusterobjectset_controller.go
```diff
@@ -288,9 +267,7 @@ func (c *ClusterObjectSetReconciler) reconcile(ctx context.Context, cos *ocv1.Cl
 		} else {
 			markAsUnavailable(cos, ocv1.ReasonRollingOut, fmt.Sprintf("Revision %s is rolling out.", revVersion))
 		}
-		if meta.FindStatusCondition(cos.Status.Conditions, ocv1.ClusterObjectSetTypeProgressing) == nil {
-			markAsProgressing(cos, ocv1.ReasonRollingOut, fmt.Sprintf("Revision %s is rolling out.", revVersion))
-		}
+		markAsProgressing(cos, ocv1.ReasonRollingOut, fmt.Sprintf("Revision %s is rolling out.", revVersion))
```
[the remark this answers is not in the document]
joelanford: Ah right. I had fixed this in my original attempt, but reverted because I had forgotten about the flapping that a post-reconcile fixup was causing.

I put back the logic to set this correctly at the markAsProgressing call, with no need for a post-reconcile fixup.

joelanford:
~experimental e2e is timing out. I guess I need to bump e2e timeout again?~

Never mind: just needed to pick up the latest changes from `main`

internal/operator-controller/controllers/clusterobjectset_controller.go
```diff
@@ -659,14 +631,26 @@ func buildProgressionProbes(progressionProbes []ocv1.ProgressionProbe) (probing.
...
+	case ocv1.ReasonSucceeded:
+		// Terminal — always apply.
+	case ocv1.ReasonRollingOut, ocv1.ClusterObjectSetReasonRetrying:
+		if isDeadlineExceeded {
+			markAsNotProgressing(cos, ocv1.ReasonProgressDeadlineExceeded,
+				fmt.Sprintf("Revision has not rolled out for %d minute(s). Last status: %s", cos.Spec.ProgressDeadlineMinutes, message))
+			return
+		}
+	default:
+		panic(fmt.Sprintf("unregistered progressing reason: %q", reason))
```
pedjak: is panicking the right strategy here... it is going to restart the pod, but with the next reconcile of the problematic resource, it is going to crash again affecting reconciliation of other valid resources.

internal/operator-controller/controllers/clusterobjectset_controller.go
```diff
@@ -659,14 +631,26 @@ func buildProgressionProbes(progressionProbes []ocv1.ProgressionProbe) (probing.
...
 	if meta.FindStatusCondition(cos.Status.Conditions, ocv1.ClusterObjectSetTypeAvailable) != nil {
 		markAsAvailableUnknown(cos, ocv1.ClusterObjectSetReasonReconciling, message)
 	}
 }
 
-func markAsProgressing(cos *ocv1.ClusterObjectSet, reason, message string) {
+func markAsProgressing(cos *ocv1.ClusterObjectSet, reason, message string, isDeadlineExceeded bool) {
+	switch reason {
+	case ocv1.ReasonSucceeded:
+		// Terminal — always apply.
```
pedjak, line 652: **No test for the recovery-after-ProgressDeadlineExceeded path.**

The `ReasonSucceeded` case is marked "Terminal — always apply", which means a COS that hits `ProgressDeadlineExceeded` and later has its probes pass should transition back to `Succeeded`. The PR description claims this behavior: _"allows revisions to recover if a transient error resolves itself, even after the deadline was exceeded."_

But neither the unit tests nor the e2e test exercise this recovery path. A unit test showing a COS transitioning from `ProgressDeadlineExceeded` → `Succeeded` when probes start passing would document and protect this behavior.

internal/operator-controller/controllers/clusterobjectset_controller.go
```diff
@@ -146,6 +124,9 @@ func (c *ClusterObjectSetReconciler) reconcile(ctx context.Context, cos *ocv1.Cl
 		return c.delete(ctx, cos)
 	}
 
+	remaining, hasDeadline := durationUntilDeadline(c.Clock, cos)
+	isDeadlineExceeded := hasDeadline && remaining <= 0
```
pedjak, line 129: **Behavioral change: errors are no longer swallowed after deadline.**

The old code explicitly cleared errors when the deadline was exceeded:
```go
reconcileErr = nil
res = ctrl.Result{}
```
The new code propagates errors even after the deadline, relying on the sticky `ProgressDeadlineExceeded` condition to communicate state. This is intentional per the PR description, but it changes how the controller interacts with the workqueue — the COS will keep retrying with backoff rather than going quiet.

Worth confirming this doesn't cause excessive log noise or API server load for permanently broken COS objects that will never recover.

internal/operator-controller/controllers/clusterobjectset_controller.go
```diff
@@ -659,14 +631,26 @@ func buildProgressionProbes(progressionProbes []ocv1.ProgressionProbe) (probing.
...
 
-func markAsProgressing(cos *ocv1.ClusterObjectSet, reason, message string) {
+func markAsProgressing(cos *ocv1.ClusterObjectSet, reason, message string, isDeadlineExceeded bool) {
+	switch reason {
+	case ocv1.ReasonSucceeded:
+		// Terminal — always apply.
+	case ocv1.ReasonRollingOut, ocv1.ClusterObjectSetReasonRetrying:
+		if isDeadlineExceeded {
+			markAsNotProgressing(cos, ocv1.ReasonProgressDeadlineExceeded,
+				fmt.Sprintf("Revision has not rolled out for %d minute(s). Last status: %s", cos.Spec.ProgressDeadlineMinutes, message))
```
pedjak, line 659: **Condition message can grow large.**

The inner `message` can already be quite verbose — e.g., probe failure messages include full `Kind.Group Namespace/Name` strings, or multi-line collision reports from `strings.Join(collidingObjs, "\n\n")`. Concatenating it untruncated into `"Last status: %s"` could produce very long status condition messages. Consider truncating `message` before embedding, or capping the total length.

internal/operator-controller/controllers/progress_deadline.go
```diff
@@ -0,0 +1,110 @@
...
+// alone that a COS succeeded in the past. If Succeeded is removed or set to
+// False, this function will return a deadline and the reconciler will set
+// ProgressDeadlineExceeded even though the revision previously succeeded.
+//
+// Returns (0, false) when there is no active deadline:
+//   - progressDeadlineMinutes is 0
+//   - the revision has already succeeded
+//   - the revision is archived (deadline is irrelevant)
+//   - the revision is being deleted
+func durationUntilDeadline(clk clock.Clock, cos *ocv1.ClusterObjectSet) (time.Duration, bool) {
```
pedjak, line 94: **`durationUntilDeadline` has no direct unit tests.**

This function has explicit early-return branches for `DeletionTimestamp != zero`, `LifecycleState == Archived`, `Succeeded == True`, and `pd <= 0`, but `progress_deadline_test.go` only tests the rate limiter (which calls `durationUntilDeadline` indirectly). Adding targeted table-driven tests for each early-return branch would protect against regressions if this logic changes — especially since the PR description documents the Succeeded-latch semantics as an important design decision.

internal/operator-controller/controllers/clusterobjectset_controller.go
```diff
@@ -146,6 +124,9 @@ func (c *ClusterObjectSetReconciler) reconcile(ctx context.Context, cos *ocv1.Cl
 		return c.delete(ctx, cos)
 	}
 
+	remaining, hasDeadline := durationUntilDeadline(c.Clock, cos)
+	isDeadlineExceeded := hasDeadline && remaining <= 0
+
 	if err := c.verifyReferencedSecretsImmutable(ctx, cos); err != nil {
 		l.Error(err, "referenced Secret verification failed, blocking reconciliation")
 		markAsNotProgressing(cos, ocv1.ClusterObjectSetReasonBlocked, err.Error())
```
pedjak, line 134: **`verifyReferencedSecretsImmutable` bypasses deadline-aware conditions.**

When secret verification fails, this calls `markAsNotProgressing(cos, Blocked, ...)` directly, not through `setRetryingConditions`. This means a COS that is both blocked AND past its deadline shows `Blocked` rather than `ProgressDeadlineExceeded`.

This is probably the right priority ordering (Blocked is more actionable), but it's an implicit precedence that isn't documented or tested. A brief comment here, or a test case covering the blocked+deadline-exceeded scenario, would clarify the intent.

internal/operator-controller/controllers/clusterobjectset_controller.go
```diff
@@ -146,6 +124,9 @@ func (c *ClusterObjectSetReconciler) reconcile(ctx context.Context, cos *ocv1.Cl
 		return c.delete(ctx, cos)
 	}
 
+	remaining, hasDeadline := durationUntilDeadline(c.Clock, cos)
```
pedjak, line 128: **`isDeadlineExceeded` is computed once, used after potentially long operations.**

`durationUntilDeadline` is called here at the start of `reconcile()`, but the resulting `isDeadlineExceeded` flag is threaded through calls that happen after `buildBoxcutterPhases`, `establishWatch`, and `revisionEngine.Reconcile`. If those operations take significant time (especially `Reconcile`), the snapshot could be stale — a COS might cross the deadline mid-reconcile but still get `RollingOut` instead of `ProgressDeadlineExceeded`.

In practice the rate limiter handles the next iteration, and recomputing mid-reconcile risks the "different Now" problem already discussed. Worth acknowledging as an accepted trade-off.

internal/operator-controller/controllers/clusterobjectset_controller.go
```diff
@@ -304,8 +285,9 @@ func (c *ClusterObjectSetReconciler) reconcile(ctx context.Context, cos *ocv1.Cl
 		} else {
 			markAsUnavailable(cos, ocv1.ReasonRollingOut, fmt.Sprintf("Revision %s is rolling out.", revVersion))
 		}
-		if meta.FindStatusCondition(cos.Status.Conditions, ocv1.ClusterObjectSetTypeProgressing) == nil {
-			markAsProgressing(cos, ocv1.ReasonRollingOut, fmt.Sprintf("Revision %s is rolling out.", revVersion))
+		markAsProgressing(cos, ocv1.ReasonRollingOut, fmt.Sprintf("Revision %s is rolling out.", revVersion), isDeadlineExceeded)
```
pedjak: **Removed the "set progressing only if nil" guard.**

The old code had:
```go
if meta.FindStatusCondition(cos.Status.Conditions, ocv1.ClusterObjectSetTypeProgressing) == nil {
    markAsProgressing(cos, ocv1.ReasonRollingOut, ...)
}
```
This now unconditionally calls `markAsProgressing`, which means every reconcile where probes are failing will overwrite the Progressing condition. If the COS already has `ProgressDeadlineExceeded` (Progressing=False), this will re-set it via `markAsNotProgressing` inside `markAsProgressing`.

The behavior is correct because `meta.SetStatusCondition` only updates `LastTransitionTime` when status or reason actually changes — so this is idempotent. But it's a subtlety worth a brief comment, since the old guard existed for a reason and a future reader might wonder why it was removed.

pedjak (commented):
Additional review comments — focused on test coverage gaps, behavioral subtleties, and edge cases.

internal/operator-controller/controllers/clusterobjectset_controller.go
joelanford: Idea is that if we add a new reason, it needs to be explicitly mentioned here to decide if it is a "success" reason or a "still progressing" reason so that it gets the correct deadline exceeded override treatment.

Seemed better to me to fail loud and hard so that we catch it in development.

internal/operator-controller/controllers/clusterobjectset_controller.go
joelanford: That is intentional. Exponential backoff after the deadline will avoid excessive log noise and API server load. And it is a fallacy to assume that a COS that has exceeded it's deadline is permanently broken. They very well may recover after the deadline.

internal/operator-controller/controllers/clusterobjectset_controller.go
joelanford: This logic merely adds a fixed amount of length to the message that was passed into the function. Before, the intended message was swallowed. Now the underlying issue causing the errors is still surfaced. It might be essential for a user to be able to troubleshoot the issue.

internal/operator-controller/controllers/clusterobjectset_controller.go
joelanford: Fair point. Changed this to log an error and treat the unknown reason as non-terminal (i.e., it will still flip to ProgressDeadlineExceeded when the deadline is exceeded). This avoids crash-looping the pod while still making the issue visible in logs.

internal/operator-controller/controllers/clusterobjectset_controller.go
joelanford: Added a unit test: "recovery from ProgressDeadlineExceeded to Succeeded when revision completes". It sets up a COS with Progressing=False/ProgressDeadlineExceeded, returns a complete revision result, and verifies the condition transitions to Progressing=True/Succeeded.

internal/operator-controller/controllers/progress_deadline.go
joelanford: Added table-driven tests covering each early-return branch: pd<=0, Succeeded=True, Archived, DeletionTimestamp set, deadline not yet exceeded, and deadline already exceeded.

internal/operator-controller/controllers/clusterobjectset_controller.go
joelanford: Added a comment clarifying the intent: "Blocked takes precedence over ProgressDeadlineExceeded: it is more actionable for the user."

internal/operator-controller/controllers/clusterobjectset_controller.go
joelanford: Agreed this is an accepted trade-off. The rate limiter handles the next iteration promptly, and recomputing mid-reconcile would introduce the "different Now" problem discussed in the earlier Copilot thread.

internal/operator-controller/controllers/clusterobjectset_controller.go
joelanford: This is safe because `meta.SetStatusCondition` only updates `LastTransitionTime` when status or reason actually changes, making the call idempotent. The old guard prevented deadline enforcement from working correctly on subsequent reconciles. It is also an anti-pattern to take action based on a status read of the object we are reconciling.

internal/operator-controller/controllers/clusterobjectset_controller_test.go
```diff
@@ -1000,6 +1000,39 @@ func Test_ClusterObjectSetReconciler_Reconcile_ProgressDeadline(t *testing.T) {
 				require.Equal(t, ocv1.ReasonRollingOut, cnd.Reason)
 			},
 		},
+		{
+			name: "recovery from ProgressDeadlineExceeded to Succeeded when revision completes",
```
pedjak, line 1004: **Consider an e2e test for the recovery path.**

This unit test mocks away the workqueue, rate limiter, watches, and event pipeline — the exact components this PR changes. It proves the `markAsProgressing` switch logic works, but doesn't validate that a deadline-exceeded COS will actually get re-reconciled in a running controller.

A simple e2e test could cover this without any mid-test patching: use a deployment with a delayed readiness probe (`initialDelaySeconds: 80`) and `progressDeadlineMinutes: 1`. The COS hits `ProgressDeadlineExceeded` at ~60s, the deployment becomes ready at ~80s, and the COS should recover to `Succeeded` — testing the full loop (workqueue re-enqueue, rate limiter, reconcile, probe pass, condition flip).

```gherkin
Then ClusterObjectSet "${COS_NAME}" reports Progressing as False with Reason ProgressDeadlineExceeded
Then ClusterObjectSet "${COS_NAME}" reports Progressing as True with Reason Succeeded
```

Adds ~90 seconds to the suite, roughly the same as the archival scenario already in this PR.

internal/operator-controller/controllers/clusterobjectset_controller.go
```diff
@@ -659,14 +632,32 @@ func buildProgressionProbes(progressionProbes []ocv1.ProgressionProbe) (probing.
...
+func setRetryingConditions(cos *ocv1.ClusterObjectSet, message string, isDeadlineExceeded bool) {
+	markAsProgressing(cos, ocv1.ClusterObjectSetReasonRetrying, message, isDeadlineExceeded)
 	if meta.FindStatusCondition(cos.Status.Conditions, ocv1.ClusterObjectSetTypeAvailable) != nil {
 		markAsAvailableUnknown(cos, ocv1.ClusterObjectSetReasonReconciling, message)
 	}
 }
 
-func markAsProgressing(cos *ocv1.ClusterObjectSet, reason, message string) {
+func markAsProgressing(cos *ocv1.ClusterObjectSet, reason, message string, isDeadlineExceeded bool) {
+	nonTerminalReasons := map[string]struct{}{
```
pedjak: Nit: this map is allocated on every `markAsProgressing` call (multiple times per reconcile). Could be a package-level `var` for a free optimization.

internal/operator-controller/controllers/clusterobjectset_controller.go
```diff
@@ -659,14 +632,32 @@ func buildProgressionProbes(progressionProbes []ocv1.ProgressionProbe) (probing.
...
+		ocv1.ReasonRollingOut:               {},
+		ocv1.ClusterObjectSetReasonRetrying: {},
+	}
+
+	switch reason {
+	case ocv1.ReasonSucceeded:
+		// Terminal — always apply.
+	default:
+		if _, known := nonTerminalReasons[reason]; !known {
+			log.Log.Error(fmt.Errorf("unregistered progressing reason: %q", reason), "treating as non-terminal for deadline enforcement")
```
pedjak: Nit: `log.Log` is the global logger — the error message won't include structured context (e.g., COS name), making it harder to diagnose in production. Acceptable given the function signature lacks a `context.Context`, but worth noting.

pedjak (commented):
Follow-up comments on the updated PR — one test coverage suggestion and two minor nits.

internal/operator-controller/controllers/clusterobjectset_controller_test.go
joelanford: Added.

internal/operator-controller/controllers/clusterobjectset_controller.go
joelanford: Plumbed the context's logger.

internal/operator-controller/controllers/clusterobjectset_controller.go
joelanford: Done

pedjak (commented):
/approve

camilamacedo86 (approved):

PR #2652, opened 2026-04-15
:bug: Fix DNS1123 validation message to match rule semantics

pedjak:
## Summary
- The `XValidation` message for `dns1123Label()` on `ClusterObjectSetPhase.Name` and `ObservedPhase.Name` incorrectly stated values "must start with an alphabetic character"
- DNS1123 labels allow starting with a digit, so the message now reads "must start and end with an alphanumeric character" to match the actual rule

## Test plan
- [ ] Verify the updated validation message is accurate per [RFC 1123](https://tools.ietf.org/html/rfc1123)
- [ ] Confirm generated CRD manifests reflect the corrected message


rashmigottipati (approved):
/lgtm

PR #2732, opened 2026-05-29
✨ catalogd graphql shift to file-based cache

grokspawn:
# Description
This PR proposes a shift in the graphql service endpoint from in-memory to on-disk caching which leverages common architectural conventions like fan-out catalog index generation (graphql-schema.json) which is then accessed to fulfill queries to eliminate RSS increases from the feature initial implementation.

Included is a structured graphql validation package as a focus for future limits enforcement and adds several cold-cache/cache-miss concurrency tests. 

Referenced against a diverse set of existing catalogs, this change reduces RSS 39X on average:

  ```
┌───────────────┬────────┬─────────┬──────────┬──────────┬───────────┐
  │    Catalog    │  Raw   │ Objects │   OLD    │   NEW    │ Reduction │
  │               │  Size  │         │  Memory  │  Memory  │           │
  ├───────────────┼────────┼─────────┼──────────┼──────────┼───────────┤
  │ r-o-i/4.23    │ 31MB   │ 1,244   │ ~126MB   │ ~2MB     │ 63x       │
  ├───────────────┼────────┼─────────┼──────────┼──────────┼───────────┤
  │ ce-o-i/4.22   │ 8.6MB  │ 954     │ ~35MB    │ ~2MB     │ 17x       │
  ├───────────────┼────────┼─────────┼──────────┼──────────┼───────────┤
  │ co-o-i/4.22   │ 20MB   │ 4,958   │ ~79MB    │ ~2MB     │ 39x       │
  └───────────────┴────────┴─────────┴──────────┴──────────┴───────────┘
```

The new file-system storage is approximately 200KB/catalog for the new file offsets index data.

Query latency shifts from microsecond access to low millisecond for the same access approach, with a transient per-query allocation between ~1-10MB depending on the size of the object(s) returned in the query.  

NB:  I ran some benchmarks after investigating repeated failures of the st2ex e2e test, and determined that unbounded FBC evaluation during schema detection phase was causing excessive delays in catalogd service readiness.  Instituted a bound of ten (10) samples of a given Meta to determine the appropriate GraphQL schema type, which allows the functionality to complete operation on example catalogs above with a worst-case of 250ms on test hardware.  These changes are in a separate commit to help them stand out. 

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/catalogd/graphql/graphql.go
+++ b/internal/catalogd/graphql/graphql.go
@@ -49,11 +49,12 @@ type CatalogSchema struct {
 
 // serializableFieldInfo is a JSON-friendly representation of FieldInfo
 type serializableFieldInfo struct {
-	Name         string                            `json:"name"`
-	OriginalName string                            `json:"originalName"`
-	JSONType     string                            `json:"jsonType"`
-	IsArray      bool                              `json:"isArray"`
-	NestedFields map[string]*serializableFieldInfo `json:"nestedFields,omitempty"`
+	Name            string                            `json:"name"`
+	OriginalName    string                            `json:"originalName"`
+	JSONType        string                            `json:"jsonType"`
+	IsArray         bool                              `json:"isArray"`
+	GraphQLTypeName string                            `json:"graphqlTypeName,omitempty"`
+	NestedFields    map[string]*serializableFieldInfo `json:"nestedFields,omitempty"`
 }
 
 // serializableSchemaInfo is a JSON-friendly representation of SchemaInfo
@@ -101,12 +102,38 @@ func stringToKind(s string) reflect.Kind {
 	}
 }
 
+func graphqlTypeName(t graphql.Type) string {
+	if list, ok := t.(*graphql.List); ok {
+		return graphqlTypeName(list.OfType)
+	}
+	return t.Name()
+}
+
+func graphqlTypeFromName(name string, isArray bool) graphql.Type {
+	var base graphql.Type
+	switch name {
+	case "Int":
+		base = graphql.Int
+	case "Float":
+		base = graphql.Float
+	case "Boolean":
+		base = graphql.Boolean
+	default:
+		base = graphql.String
+	}
+	if isArray {
+		return graphql.NewList(base)
+	}
+	return base
+}
+
 func fieldInfoToSerializable(fi *FieldInfo) *serializableFieldInfo {
 	sfi := &serializableFieldInfo{
-		Name:         fi.Name,
-		OriginalName: fi.OriginalName,
-		JSONType:     kindToString(fi.JSONType),
-		IsArray:      fi.IsArray,
+		Name:            fi.Name,
+		OriginalName:    fi.OriginalName,
+		JSONType:        kindToString(fi.JSONType),
+		IsArray:         fi.IsArray,
+		GraphQLTypeName: graphqlTypeName(fi.GraphQLType),
 	}
 	if len(fi.NestedFields) > 0 {
 		sfi.NestedFields = make(map[string]*serializableFieldInfo)
@@ -119,12 +146,18 @@ func fieldInfoToSerializable(fi *FieldInfo) *serializableFieldInfo {
 
 func serializableToFieldInfo(sfi *serializableFieldInfo) *FieldInfo {
 	k := stringToKind(sfi.JSONType)
+	var gqlType graphql.Type
+	if sfi.GraphQLTypeName != "" {
+		gqlType = graphqlTypeFromName(sfi.GraphQLTypeName, sfi.IsArray)
+	} else {
+		gqlType = jsonTypeToGraphQL(k, sfi.IsArray, nil)
+	}
 	fi := &FieldInfo{
 		Name:         sfi.Name,
 		OriginalName: sfi.OriginalName,
 		JSONType:     k,
 		IsArray:      sfi.IsArray,
-		GraphQLType:  jsonTypeToGraphQL(k, sfi.IsArray, nil),
+		GraphQLType:  gqlType,
 	}
 	if len(sfi.NestedFields) > 0 {
 		fi.NestedFields = make(map[string]*FieldInfo)
--- a/internal/catalogd/graphql/graphql.go
+++ b/internal/catalogd/graphql/graphql.go
@@ -335,6 +335,7 @@ func determineFieldType(value interface{}) (reflect.Kind, bool) {
 }
 
 const maxSampleElements = 10
+const maxSampleValues = 10
 
 // analyzeFieldValue analyzes a field value and returns type info, sample value, and nested fields.
 // For slices, it examines up to maxSampleElements to detect heterogeneous element types
@@ -439,6 +440,9 @@ func mergeNestedFields(existing, new map[string]*FieldInfo) {
 		if existingInfo, ok := existing[fieldName]; ok {
 			// Merge sample values
 			for _, sample := range newInfo.SampleValues {
+				if len(existingInfo.SampleValues) >= maxSampleValues {
+					break
+				}
 				existingInfo.SampleValues = appendUnique(existingInfo.SampleValues, sample)
 			}
 		} else {
@@ -481,7 +485,9 @@ func analyzeJSONObject(obj map[string]interface{}, info *SchemaInfo) {
 		}
 
 		// Update existing field
-		existing.SampleValues = appendUnique(existing.SampleValues, sampleValue)
+		if len(existing.SampleValues) < maxSampleValues {
+			existing.SampleValues = appendUnique(existing.SampleValues, sampleValue)
+		}
 
 		// Merge nested fields if discovered
 		if nestedFields == nil {
--- a/internal/catalogd/graphql/graphql.go
+++ b/internal/catalogd/graphql/graphql.go
@@ -31,7 +31,6 @@ type FieldInfo struct {
 	GraphQLType  graphql.Type
 	JSONType     reflect.Kind
 	IsArray      bool
-	SampleValues []interface{}
 	NestedFields map[string]*FieldInfo // For array-of-objects, stores object structure
 }
 
@@ -39,7 +38,6 @@ type FieldInfo struct {
 type SchemaInfo struct {
 	Fields       map[string]*FieldInfo
 	TotalObjects int
-	SampleObject map[string]interface{}
 }
 
 // CatalogSchema holds the complete discovered schema
@@ -334,8 +332,9 @@ func determineFieldType(value interface{}) (reflect.Kind, bool) {
 	return reflect.TypeOf(firstElem).Kind(), true
 }
 
+// maxSampleElements limits how many slice elements are examined when inferring
+// GraphQL types from catalog data, bounding work on large arrays.
 const maxSampleElements = 10
-const maxSampleValues = 10
 
 // analyzeFieldValue analyzes a field value and returns type info, sample value, and nested fields.
 // For slices, it examines up to maxSampleElements to detect heterogeneous element types
@@ -427,7 +426,6 @@ func analyzeNestedObject(obj map[string]interface{}) map[string]*FieldInfo {
 			GraphQLType:  jsonTypeToGraphQL(jsonType, isArray, sampleValue),
 			JSONType:     jsonType,
 			IsArray:      isArray,
-			SampleValues: []interface{}{value},
 		}
 	}
 
@@ -437,15 +435,7 @@ func analyzeNestedObject(obj map[string]interface{}) map[string]*FieldInfo {
 // mergeNestedFields merges discovered nested fields into existing ones
 func mergeNestedFields(existing, new map[string]*FieldInfo) {
 	for fieldName, newInfo := range new {
-		if existingInfo, ok := existing[fieldName]; ok {
-			// Merge sample values
-			for _, sample := range newInfo.SampleValues {
-				if len(existingInfo.SampleValues) >= maxSampleValues {
-					break
-				}
-				existingInfo.SampleValues = appendUnique(existingInfo.SampleValues, sample)
-			}
-		} else {
+		if _, ok := existing[fieldName]; !ok {
 			existing[fieldName] = newInfo
 		}
 	}
@@ -472,7 +462,6 @@ func analyzeJSONObject(obj map[string]interface{}, info *SchemaInfo) {
 				GraphQLType:  jsonTypeToGraphQL(jsonType, isArray, sampleValue),
 				JSONType:     jsonType,
 				IsArray:      isArray,
-				SampleValues: []interface{}{sampleValue},
 				NestedFields: nestedFields,
 			}
 			continue
@@ -482,11 +471,6 @@ func analyzeJSONObject(obj map[string]interface{}, info *SchemaInfo) {
 		if existing.OriginalName != key {
 			klog.V(2).InfoS("field name collision: different JSON keys map to same GraphQL field",
 				"graphqlField", fieldName, "existingKey", existing.OriginalName, "newKey", key)
-		}
-
-		// Update existing field
-		if len(existing.SampleValues) < maxSampleValues {
-			existing.SampleValues = appendUnique(existing.SampleValues, sampleValue)
 		}
 
 		// Merge nested fields if discovered
@@ -501,70 +485,37 @@ func analyzeJSONObject(obj map[string]interface{}, info *SchemaInfo) {
 	}
 }
 
-// appendUnique adds a value to slice if not already present, using JSON string as key for uniqueness
-func appendUnique(slice []interface{}, value interface{}) []interface{} {
-	seen := make(map[string]struct{}, len(slice))
-
-	for _, existing := range slice {
-		key, err := json.Marshal(existing)
-		if err != nil {
-			continue // skip values that can't be marshaled
+// processMetaIntoSchema parses one meta blob and updates the catalog schema.
+// TotalObjects is incremented only on successful parse so pagination counts
+// reflect objects that can actually be returned by the ObjectLoader.
+func processMetaIntoSchema(catalogSchema *CatalogSchema, meta *declcfg.Meta) {
+	if meta.Schema == "" {
+		return
+	}
+	if catalogSchema.Schemas[meta.Schema] == nil {
+		catalogSchema.Schemas[meta.Schema] = &SchemaInfo{
+			Fields: make(map[string]*FieldInfo),
 		}
-		seen[string(key)] = struct{}{}
+	}
+	info := catalogSchema.Schemas[meta.Schema]
+
+	var obj map[string]interface{}
+	if err := json.Unmarshal(meta.Blob, &obj); err != nil {
+		klog.V(4).InfoS("skipping malformed meta blob during schema discovery",
+			"schema", meta.Schema, "name", meta.Name, "error", err)
+		return
 	}
 
-	valueKey, err := json.Marshal(value)
-	if err != nil {
-		return slice // skip value if it can't be marshaled
-	}
-
-	if _, exists := seen[string(valueKey)]; exists {
-		return slice
-	}
-
-	return append(slice, value)
+	info.TotalObjects++
+	analyzeJSONObject(obj, info)
 }
 
-// DiscoverSchemaFromMetas analyzes Meta objects to discover schema structure
+// DiscoverSchemaFromMetas analyzes Meta objects to discover schema structure.
 func DiscoverSchemaFromMetas(metas []*declcfg.Meta) (*CatalogSchema, error) {
-	catalogSchema := &CatalogSchema{
-		Schemas: make(map[string]*SchemaInfo),
-	}
-
-	// Process each meta object
+	catalogSchema := &CatalogSchema{Schemas: make(map[string]*SchemaInfo)}
 	for _, meta := range metas {
-		if meta.Schema == "" {
-			continue
-		}
-
-		// Ensure schema info exists
-		if catalogSchema.Schemas[meta.Schema] == nil {
-			catalogSchema.Schemas[meta.Schema] = &SchemaInfo{
-				Fields:       make(map[string]*FieldInfo),
-				TotalObjects: 0,
-			}
-		}
-
-		info := catalogSchema.Schemas[meta.Schema]
-		info.TotalObjects++
-
-		// Parse the JSON blob
-		var obj map[string]interface{}
-		if err := json.Unmarshal(meta.Blob, &obj); err != nil {
-			klog.V(4).InfoS("skipping malformed meta blob during schema discovery",
-				"schema", meta.Schema, "name", meta.Name, "error", err)
-			continue
-		}
-
-		// Store a sample object for reference
-		if info.SampleObject == nil {
-			info.SampleObject = obj
-		}
-
-		// Analyze general fields (including nested structures)
-		analyzeJSONObject(obj, info)
+		processMetaIntoSchema(catalogSchema, meta)
 	}
-
 	return catalogSchema, nil
 }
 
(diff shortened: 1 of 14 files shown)

### CODE REVIEW

File: internal/catalogd/storage/localdir.go [hunk 1]
```diff
@@ -273,6 +290,107 @@ func storeIndexData(catalogDir string, metas <-chan *declcfg.Meta) error {
...
+		return nil, fmt.Errorf("error loading index for catalog %q: %w", catalog, err)
+	}
+
+	catalogPath := catalogFilePath(s.catalogDir(catalog))
+
+	return func(schemaName string, offset, limit int) ([]map[string]interface{}, error) {
+		sections := idx.GetSchemaSections(schemaName)
+		if sections == nil {
+			return nil, nil
+		}
```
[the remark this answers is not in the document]
grokspawn: This _shouldn't_ happen, but even if it did it will handle it with the cache invalidation strategy which copilot is not including in its context.

internal/catalogd/storage/localdir.go
```diff
@@ -273,6 +290,107 @@ func storeIndexData(catalogDir string, metas <-chan *declcfg.Meta) error {
...
+			return nil, fmt.Errorf("error opening catalog file: %w", err)
+		}
+		defer f.Close()
+
+		results := make([]map[string]interface{}, 0, len(sections))
+		for _, sec := range sections {
+			buf := make([]byte, sec.Length)
+			if _, err := f.ReadAt(buf, sec.Offset); err != nil {
+				return nil, fmt.Errorf("error reading section at offset %d: %w", sec.Offset, err)
+			}
```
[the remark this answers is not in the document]
grokspawn: Hallucination.

internal/catalogd/storage/localdir.go
```diff
@@ -273,6 +290,107 @@ func storeIndexData(catalogDir string, metas <-chan *declcfg.Meta) error {
...
+
+			var obj map[string]interface{}
+			if err := json.Unmarshal(buf, &obj); err != nil {
+				continue
+			}
+			results = append(results, obj)
+		}
```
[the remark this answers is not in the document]
grokspawn: copilot is hallucinating

internal/catalogd/storage/localdir.go
```diff
@@ -273,6 +290,107 @@ func storeIndexData(catalogDir string, metas <-chan *declcfg.Meta) error {
...
+
+		if limit < len(sections) {
+			sections = sections[:limit]
+		}
+
+		f, err := os.Open(catalogPath)
+		if err != nil {
+			return nil, fmt.Errorf("error opening catalog file: %w", err)
+		}
+		defer f.Close()
```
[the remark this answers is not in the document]
grokspawn: Handled by cache invalidation on reconcile; the loader is short-lived and a stale read triggers re-sync.

internal/catalogd/graphql/validation.go
```diff
@@ -0,0 +1,92 @@
...
+			name := s.Name.Value
+			if c.visited[name] {
+				continue
+			}
+			c.visited[name] = true
+			if frag, ok := c.fragments[name]; ok {
+				if err := c.walkSelectionSet(frag.SelectionSet, depth+1); err != nil {
+					return err
+				}
+			}
```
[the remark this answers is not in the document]
grokspawn: Intentionally conservative — fragments  consuming depth budget is acceptable for DoS protection limits.

internal/catalogd/graphql/graphql.go
```diff
@@ -172,7 +334,12 @@ func determineFieldType(value interface{}) (reflect.Kind, bool) {
 	return reflect.TypeOf(firstElem).Kind(), true
 }
 
-// analyzeFieldValue analyzes a field value and returns type info, sample value, and nested fields
+const maxSampleElements = 10
```
perdasilva, line 337: would it be worth adding a comment to explain these parameters? I wonder if they'll need to be tweaked at some point in the future...?

internal/catalogd/graphql/validation.go
```diff
@@ -0,0 +1,92 @@
+package graphql
+
+import (
+	"fmt"
+
+	"github.com/graphql-go/graphql/language/ast"
+	"github.com/graphql-go/graphql/language/parser"
+)
+
+const (
```
perdasilva, line 10: do we need to export these fellas?

internal/catalogd/storage/localdir.go
```diff
@@ -74,38 +75,68 @@ func NewLocalDirV1(rootDir string, rootURL *url.URL, enableMetasHandler MetasHan
...
 	}
 	if enableGraphQLQueries == GraphQLQueriesEnabled {
-		s.graphqlSvc = service.NewCachedGraphQLService()
+		s.graphqlSvc = service.NewCachedGraphQLService(s)
 	}
 	return s
 }
 
 func (s *LocalDirV1) Store(ctx context.Context, catalog string, fsys fs.FS) error {
+	catalogDir, err := s.storeAtomicSwap(ctx, catalog, fsys)
```
perdasilva, line 85: could there be a race condition if there are two concurrent calls to `Store`?

First caller finishes `storeAtomicSwap` and starts `GetSchema`
Second caller starts `storeAtomicSwap` nuking the existing data
First caller `GetSchema` fails due to rug pull - rolls back destroying 2nd caller's `storeAtomicSwap` changes
Second caller soldiers on thinking his catalog data is where it should be - but 1st caller just rolled back!

internal/catalogd/storage/localdir.go
```diff
@@ -74,38 +75,68 @@ func NewLocalDirV1(rootDir string, rootURL *url.URL, enableMetasHandler MetasHan
...
+	// Concurrent queries during this window get a cache miss and trigger
+	// their own build via singleflight, which is safe — the data is on disk.
+	if s.graphqlSvc != nil {
+		s.graphqlSvc.InvalidateCache(catalog)
+
+		if _, err := s.graphqlSvc.GetSchema(ctx, catalog); err != nil {
+			// Schema build failed — remove the catalog to maintain consistency.
+			// Re-acquire the write lock for the rollback since it touches shared filesystem state.
+			s.m.Lock()
+			removeErr := os.RemoveAll(catalogDir)
```
perdasilva, line 101: should we `InvalidateCache` again here? `GetSchema` rebuilds the cache right?

internal/catalogd/storage/localdir.go
```diff
@@ -273,6 +290,107 @@ func storeIndexData(catalogDir string, metas <-chan *declcfg.Meta) error {
...
+		if offset >= len(sections) {
+			return nil, nil
+		}
+		sections = sections[offset:]
+
+		if limit < len(sections) {
+			sections = sections[:limit]
+		}
+
+		f, err := os.Open(catalogPath)
```
perdasilva, line 358: should we hold a lock on reading the catalog? Another thread somewhere could be rolling back?

internal/catalogd/graphql/graphql.go
grokspawn: Yes, this was a total oversight on my part.  You don't write down the tech debt before you go on holiday and it gets dropped sometimes.  Sorry!

internal/catalogd/graphql/validation.go
grokspawn: I didn't stress b/c these are internal anyway, but easy enough to do.

internal/catalogd/storage/localdir.go
grokspawn: It's theoretically possible, but I think in practice it's vanishingly unlikely.  In order for that to reasonably happen, a catalog would have to be unpacking when we notice that it needs to unpack (but fail to notice its unpacking condition), and that's not likely.  
However, it also is simple enough to broaden the lock scoping to encapsulate store+schema+rollback under that set of circumstances to eliminate that class of problem entirely.

internal/catalogd/storage/localdir.go
grokspawn: Yes, we should.  Nice catch.  
I think we mostly would've been fine here, but bookkeeping should be totally clean now.

internal/catalogd/storage/localdir.go
grokspawn: Totally reasonable.  Thanks.

internal/catalogd/storage/localdir.go
```diff
@@ -74,38 +75,64 @@ func NewLocalDirV1(rootDir string, rootURL *url.URL, enableMetasHandler MetasHan
...
+	}
+	s.m.Unlock()
+
+	if s.graphqlSvc != nil {
+		s.graphqlSvc.InvalidateCache(catalog)
 
+		if _, err := s.graphqlSvc.GetSchema(ctx, catalog); err != nil {
+			s.graphqlSvc.InvalidateCache(catalog)
+			s.m.Lock()
+			removeErr := os.RemoveAll(catalogDir)
```
perdasilva, line 101: **Bug — Store() rollback race**

After the write lock is released at line 90, a concurrent `Store("same-catalog", fsB)` can complete its own `storeAtomicSwap`, replacing the catalog directory with valid fsB data. If *this* goroutine's `GetSchema` then fails, `os.RemoveAll(catalogDir)` here deletes the *other* goroutine's valid catalog directory — `catalogDir` is a deterministic path (`rootDir/catalogName`), not the one this goroutine wrote.

The result: both Stores fail, and the catalog is left in a missing state despite fsB being valid.

Consider keeping the write lock held through the entire `Store()` operation, or using per-catalog locks.

perdasilva (commented):
## Code Review — 10 findings (3 correctness, 1 performance, 6 cleanup/structural)

The most impactful fix would be keeping the write lock held through the entire `Store()` operation (or using per-catalog locks), which would resolve findings 1–3 in one change.

### Finding not placeable inline:

**Performance — O(N×regex) field resolution on every query** (`graphql.go:678`)

`createFieldResolver` iterates *all* source keys calling `remapFieldName` (3 regex replacements each) to find the matching key. For a query returning 100 objects with 20 fields, requesting 5 fields: 5 × 100 × 20 × 3 = **30,000 regex operations per query**.

`FieldInfo.OriginalName` is already stored during schema discovery — the resolver could capture it and do a direct O(1) map lookup:
```go
func createFieldResolver(fieldName, originalName string) graphql.FieldResolveFn {
    return func(p graphql.ResolveParams) (interface{}, error) {
        if source, ok := p.Source.(map[string]interface{}); ok {
            if value, ok := source[originalName]; ok {
                return marshalComplexValue(value), nil
            }
        }
        return nil, nil
    }
}
```

Same issue in `createNestedFieldResolver` (line 698) and `createNestedObjectType` resolver (line 653).

internal/catalogd/storage/localdir.go
```diff
@@ -74,38 +75,64 @@ func NewLocalDirV1(rootDir string, rootURL *url.URL, enableMetasHandler MetasHan
...
 
 func (s *LocalDirV1) Store(ctx context.Context, catalog string, fsys fs.FS) error {
 	s.m.Lock()
-	defer s.m.Unlock()
+	catalogDir, err := s.storeAtomicSwap(ctx, catalog, fsys)
+	if err != nil {
+		s.m.Unlock()
+		return err
+	}
+	s.m.Unlock()
```
perdasilva: **Bug — stale cached schema serves queries against replaced catalog file**

Between `s.m.Unlock()` here and `InvalidateCache` at line 93, a concurrent GraphQL query can hit the *old* cached schema. That schema's `ObjectLoader` holds byte offsets from the *previous* index file. Since `storeAtomicSwap` already replaced `catalog.jsonl` with different content, the `ObjectLoader` reads the new file at old offsets → corrupted/wrong JSON objects.

This is the same root cause as the rollback race: the lock is released too early. Holding the lock through `InvalidateCache` (or through the entire pre-warm) would close this window.

internal/catalogd/storage/localdir.go
```diff
@@ -74,38 +75,64 @@ func NewLocalDirV1(rootDir string, rootURL *url.URL, enableMetasHandler MetasHan
...
 	tmpCatalogDir, err := os.MkdirTemp(s.RootDir, fmt.Sprintf(".%s-*", catalog))
 	if err != nil {
-		return err
+		return "", err
 	}
 	defer os.RemoveAll(tmpCatalogDir)
 
 	storeMetaFuncs := []storeMetasFunc{storeCatalogData}
-	if s.EnableMetasHandler {
+	if s.EnableMetasHandler || s.EnableGraphQLQueries == GraphQLQueriesEnabled {
```
perdasilva, line 131: **Bug — ContentExists doesn't verify GraphQL-required files**

`storeIndexData` is now enabled when GraphQL is enabled (even with `MetasHandlerDisabled`), and `discoverAndStoreSchema` writes `graphql-schema.json`. But `ContentExists` (line 230) only checks `index.json` when `EnableMetasHandler` is set — it never checks `graphql-schema.json`.

If a crash leaves `catalog.jsonl` intact but `graphql-schema.json` or `index.json` missing, `ContentExists` returns `true` while GraphQL queries fail with file-not-found from `LoadCatalogSchema`/`loadIndex`.

```go
if s.EnableGraphQLQueries == GraphQLQueriesEnabled {
    // check index.json and graphql-schema.json too
}
```

internal/catalogd/graphql/graphql.go
```diff
@@ -355,6 +570,46 @@ func DiscoverSchemaFromMetas(metas []*declcfg.Meta) (*CatalogSchema, error) {
...
+
+		if catalogSchema.Schemas[meta.Schema] == nil {
+			catalogSchema.Schemas[meta.Schema] = &SchemaInfo{
+				Fields:       make(map[string]*FieldInfo),
+				TotalObjects: 0,
+			}
+		}
+
+		info := catalogSchema.Schemas[meta.Schema]
+		info.TotalObjects++
```
perdasilva: **Bug — TotalObjects counts blobs that fail json.Unmarshal**

`TotalObjects` is incremented here *before* the `json.Unmarshal` at line 597. If a blob is malformed, `TotalObjects` includes it but the `ObjectLoader` (localdir.go:365) silently skips parse failures.

A client using `summary.totalObjects` for pagination will request beyond the actual data. Move the increment after the successful unmarshal, or count failures separately.

internal/catalogd/graphql/graphql.go
```diff
@@ -26,6 +27,7 @@ var (
 // FieldInfo represents discovered field information
 type FieldInfo struct {
 	Name         string
+	OriginalName string // The original JSON key before remapping
```
perdasilva, line 30: **Cleanup — SampleValues (line 34) and SampleObject (line 42) are dead state**

`OriginalName` is a great addition. However, `SampleValues` is populated during schema discovery (lines 432, 444–448, 477, 490–491) via `appendUnique` — which JSON-marshals *all* existing elements on every call — but is never read by any consumer: not by serialization (`serializableFieldInfo` omits it), not by `BuildDynamicGraphQLSchema`, not by resolvers, not by `PrintCatalogSummary`.

Same for `SampleObject` on `SchemaInfo` (line 42) — written at lines 562–563 and 603–604, never read.

Removing both fields and `appendUnique` would eliminate ~40 lines and the per-meta JSON marshaling overhead during every `Store` operation.

internal/catalogd/graphql/graphql.go
```diff
@@ -355,6 +570,46 @@ func DiscoverSchemaFromMetas(metas []*declcfg.Meta) (*CatalogSchema, error) {
 	return catalogSchema, nil
 }
 
+// DiscoverSchemaFromChannel performs streaming schema discovery, processing
+// one meta at a time through a channel. Each meta's blob is parsed, analyzed,
+// and then goes out of scope — avoiding accumulation of all blobs in memory.
+func DiscoverSchemaFromChannel(metasChan <-chan *declcfg.Meta) (*CatalogSchema, error) {
```
perdasilva, line 525: **Cleanup — duplicates DiscoverSchemaFromMetas body**

This function's body (lines 576–611) is nearly identical to `DiscoverSchemaFromMetas` (lines 531–571): init `CatalogSchema`, skip empty schema, ensure map entry, increment `TotalObjects`, unmarshal blob, set `SampleObject`, call `analyzeJSONObject`. Only the iteration source differs (channel vs slice).

A shared `processMeta(catalogSchema, meta)` helper would reduce both to a ~5-line loop, eliminating ~35 duplicated lines and the risk of divergence (e.g., if the `TotalObjects` counting bug above is fixed in one but not the other).

internal/catalogd/storage/localdir.go
```diff
@@ -273,6 +286,114 @@ func storeIndexData(catalogDir string, metas <-chan *declcfg.Meta) error {
...
+			results = append(results, obj)
+		}
+
+		return results, nil
+	}, nil
+}
+
+// loadIndex loads the index from disk using singleflight for efficiency.
+func (s *LocalDirV1) loadIndex(catalog string) (*index, error) {
+	s.m.RLock()
```
perdasilva, line 388: **Efficiency — RLock held across singleflight disk I/O**

`loadIndex` acquires `RLock` here and holds it through the entire `singleflight.Do` (line 379), which opens and JSON-decodes the index file. All concurrent callers sharing the singleflight also hold `RLock`s while waiting for the leader to finish.

During this time, `Store()` cannot acquire the write lock. For a large index or slow disk, this creates unbounded reader starvation of writers. Consider acquiring the `RLock` only for the path computation, or using a separate lock for the singleflight.

internal/catalogd/graphql/validation.go
```diff
@@ -0,0 +1,92 @@
...
+// depth, alias, or total field count thresholds.
+func ValidateQueryComplexity(query string) error {
+	doc, err := parser.Parse(parser.ParseParams{Source: query})
+	if err != nil {
+		return fmt.Errorf("query parse error: %w", err)
+	}
+
+	c := &queryComplexity{
+		fragments: make(map[string]*ast.FragmentDefinition),
+		visited:   make(map[string]bool),
```
perdasilva: **Minor bug — fragment visited map shared across operations**

The `visited` map is created once per `ValidateQueryComplexity` call and shared across all `OperationDefinition`s in the document (line 40–46). If two operations both spread the same fragment (`query A { ...F } query B { ...F }`), validating op A marks fragment F as visited (line 83). Validating op B finds F already visited (line 80) and skips counting its fields.

This under-counts complexity for multi-operation documents with shared fragments. Low impact since clients typically send one operation per request, but the fix is simple: reset `visited` per operation.

internal/catalogd/graphql/graphql.go
```diff
@@ -45,16 +47,176 @@ type CatalogSchema struct {
 	Schemas map[string]*SchemaInfo // schema name -> info
 }
 
+// serializableFieldInfo is a JSON-friendly representation of FieldInfo
+type serializableFieldInfo struct {
+	Name            string                            `json:"name"`
+	OriginalName    string                            `json:"originalName"`
```
perdasilva, line 51: **Structural — serializable shadow hierarchy adds maintenance burden**

`serializableFieldInfo`/`serializableSchemaInfo`/`serializableCatalogSchema` plus 6 converter functions (~130 lines) exist because `FieldInfo` stores non-JSON-serializable `reflect.Kind` and `graphql.Type`.

But both are derivable: `reflect.Kind` round-trips through a string, and `graphql.Type` is already reconstructed from `JSONType` + `IsArray` via `jsonTypeToGraphQL`. If `FieldInfo` stored a string enum for `JSONType` and dropped `GraphQLType` (reconstructing it lazily in `BuildDynamicGraphQLSchema`), this entire shadow hierarchy collapses into plain `json.Marshal`/`Unmarshal` with struct tags.

internal/catalogd/storage/localdir.go
grokspawn: moved to invalidate before unlocking, so that stale byte offsets are not used for live queries.

internal/catalogd/storage/localdir.go
grokspawn: Initially, the absence of this check was deliberate, because it would be generated during the unpack/store operation.  However, we cannot eliminate the possibility that the process could fail while leaving enough objects behind for the `/all` API to work while silently concealing that the `/graphql` API is irretrievably broken.  The design relies on some recognition that we have not yet unpacked, and can stale/regenerate the file, but it's not unreasonable to have a fail-fast check for the files.

internal/catalogd/graphql/graphql.go
grokspawn: moved it to count only unmarshal successes

internal/catalogd/graphql/graphql.go
grokspawn: pruned deadwood

internal/catalogd/graphql/graphql.go
grokspawn: unified

internal/catalogd/storage/localdir.go
grokspawn: theoretical problem.  impossible to resolve without profiling, which necessitates merging the feature to get the info.  Deferring to a later effort if it manifests.

internal/catalogd/graphql/validation.go
grokspawn: resetting the visited info per-iteration to better track

internal/catalogd/graphql/graphql.go
grokspawn: doing this would be a huge rise in complexity for this PR and the benefits aren't  evident.  I would defer any such investigation to a follow-on effort.  This PR already resolves a significant reduction in the RSS for catalogd with the feature enabled.

grokspawn: It's also worth noting that one of the design goals for this PR was to reduce the number of (duplicate) marshal/unmarshal operations, so we're attempting to carry a little temp storage around to reduce I/O.  It's a balancing act, and we won't know success for awhile.

internal/catalogd/storage/localdir.go
grokspawn: Right now this looks like a highly theoretical sequence of events, which would require that catalogd's polling/unpack phases are so delayed that they back up against one another so that we're attempting to unpack either two updated images simultaneously or the original + an updated image.  
Knowing the general update frequency of catalog images in the wild presently, I'm inclined to defer the work.  
It's also worth nothing that when we keep the lock through the store operation, the LLMs tell us to truncate the duration because it's a problem.  And if we hold it for a focused relevancy period we get LLM feedback that we should treat the whole operation as a critical section. 
We have new checks which help us detect if we were only partially through an unpack, and I'm inclined to see how those prove out before tilting again at this windmill. :)

internal/catalogd/storage/localdir.go
```diff
@@ -273,6 +298,114 @@ func storeIndexData(catalogDir string, metas <-chan *declcfg.Meta) error {
...
+		}
+		defer f.Close()
+
+		const maxEntrySize = 16 * 1024 * 1024 // 16 MiB
+		results := make([]map[string]interface{}, 0, len(sections))
+		for _, sec := range sections {
+			if sec.Length <= 0 || sec.Length > maxEntrySize {
+				return nil, fmt.Errorf("invalid section length %d at offset %d", sec.Length, sec.Offset)
+			}
+			buf := make([]byte, sec.Length)
```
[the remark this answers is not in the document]
grokspawn: still hallucinating

internal/catalogd/graphql/graphql.go
```diff
@@ -570,40 +752,28 @@ func BuildDynamicGraphQLSchema(catalogSchema *CatalogSchema, metasBySchema map[s
...
-				objects, ok := parsedObjects[currentSchemaName]
-				if !ok {
-					return []interface{}{}, nil
-				}
-
-				// Parse arguments
 				limit, _ := p.Args["limit"].(int)
 				if limit <= 0 || limit > 100 {
-					limit = 100 // Clamp to default/max to prevent DoS
+					limit = 100
```
tmshort, line 762: Nit: loss of context comment

internal/catalogd/graphql/graphql.go
```diff
@@ -570,40 +752,28 @@ func BuildDynamicGraphQLSchema(catalogSchema *CatalogSchema, metasBySchema map[s
...
 				}
 				offset, _ := p.Args["offset"].(int)
 				if offset < 0 {
-					offset = 0 // Negative offsets make no sense
+					offset = 0
```
tmshort, line 766: Nit: loss of context comment

tmshort (approved):
/lgtm

rashmigottipati (approved):

PR #2764, opened 2026-06-12
:seedling: Upgrade boxcutter to v0.14.0 and adapt sibling owners API

perdasilva:
## Description

Adapt the ClusterObjectSet controller to boxcutter v0.14.0's API:
- `WithPreviousOwners` is replaced by `WithSiblingOwners`, which takes all active sibling revisions (both lower and higher) so boxcutter can distinguish siblings from true collisions during handover.
- `NewObjectEngine` now requires a `managedBy` parameter.

### Changes

- **`listSiblingRevisions`** — returns all active revisions for the same ClusterExtension, excluding self, archived, and deleting revisions. Used by `buildBoxcutterPhases` with `WithSiblingOwners`.
- **`listPreviousRevisions`** — returns only lower-revision active siblings. Used for archiving after rollout completes.
- **`listOtherActiveRevisions`** — shared helper with predicate-based filtering used by both methods above.
- **`buildBoxcutterPhases`** — uses `WithSiblingOwners` instead of `WithPreviousOwners`.
- **`NewObjectEngine`** — passes `FieldOwnerPrefix` as the new `managedBy` parameter.
- **Dependency bumps** — boxcutter v0.13.1 → v0.14.0, plus transitive updates.
- **Unit tests** — separate table-driven tests for `listSiblingRevisions` and `listPreviousRevisions` with shared `callRevisionLister` helper.

The cross-CE collision regression e2e test has been split out into #2783.

## Reviewer Checklist

- [ ] API Go Documentation


internal/operator-controller/controllers/revision_engine_factory.go
```diff
@@ -69,7 +69,7 @@ func (f *defaultRevisionEngineFactory) CreateRevisionEngine(_ context.Context, r
 			machinery.NewObjectEngine(
 				f.Scheme, f.TrackingCache, scopedClient,
 				machinery.NewComparator(f.DiscoveryClient, f.Scheme, f.FieldOwnerPrefix),
-				f.FieldOwnerPrefix, f.FieldOwnerPrefix, scopedClient,
+				f.FieldOwnerPrefix, f.FieldOwnerPrefix, "", scopedClient,
 			),
```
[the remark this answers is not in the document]
perdasilva: Fixed — added an inline comment clarifying the `managedBy` parameter.

go.mod
```diff
@@ -256,7 +256,7 @@ require (
 	sigs.k8s.io/randfill v1.0.0 // indirect
 )
 
-retract v1.5.0 // contains filename with ':' which causes failure creating module zip file
+retract v1.5.0 // contains filename with ':' which causes failure creating module zip fileq
```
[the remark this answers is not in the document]
perdasilva: Good catch — stray keystroke during the dependency update. Fixed.

internal/operator-controller/controllers/clusterobjectset_controller.go
```diff
@@ -416,6 +416,38 @@ func (c *ClusterObjectSetReconciler) removeFinalizer(ctx context.Context, obj cl
 	return nil
 }
 
+// listSiblingRevisions returns all active revisions belonging to the same ClusterExtension, excluding the current one.
+// This includes both lower and higher revision numbers, enabling boxcutter to properly classify
```
pedjak, line 421: **suggestion (non-blocking):** `listSiblingRevisions` and `listPreviousRevisions` share ~15 lines of identical logic (label lookup, `TrackingCache.List`, skip-self, skip-archived, skip-deleting). The only difference is the `Revision >= cos.Spec.Revision` filter. Consider extracting a private helper to reduce the duplication:

```go
func (c *ClusterObjectSetReconciler) listActiveRevisions(
    ctx context.Context,
    cos *ocv1.ClusterObjectSet,
    include func(*ocv1.ClusterObjectSet) bool,
) ([]*ocv1.ClusterObjectSet, error)
```

Then `listSiblingRevisions` passes `func(*ocv1.ClusterObjectSet) bool { return true }` and `listPreviousRevisions` passes the revision-number check. Fine as a follow-up.

pedjak (commented):
**question (non-blocking):** The Copilot review on an earlier push mentioned an e2e `.feature` file change for collision persistence across upgrades, but it's absent from the current diff. Is that planned as a follow-up PR? The unit tests cover `listSiblingRevisions` well, but an e2e test exercising the higher-sibling handover path would add confidence that the behavioral change works end-to-end.

internal/operator-controller/controllers/clusterobjectset_controller.go
perdasilva: nice one! Fixed =D

internal/operator-controller/controllers/clusterobjectset_controller.go
```diff
@@ -416,12 +416,27 @@ func (c *ClusterObjectSetReconciler) removeFinalizer(ctx context.Context, obj cl
...
+}
+
 // listPreviousRevisions returns active revisions belonging to the same ClusterExtension with lower revision numbers.
-// Filters out the current revision, archived revisions, deleting revisions, and revisions with equal or higher numbers.
 func (c *ClusterObjectSetReconciler) listPreviousRevisions(ctx context.Context, cos *ocv1.ClusterObjectSet) ([]*ocv1.ClusterObjectSet, error) {
+	return c.listActiveRevisions(ctx, cos, func(r *ocv1.ClusterObjectSet) bool {
+		return r.Spec.Revision < cos.Spec.Revision
+	})
+}
+
```
pedjak, line 433: **suggestion:** `listActiveRevisions` implies it returns all active revisions, but it excludes the one passed as `cos`. Consider `listOtherActiveRevisions` to make the exclusion behavior obvious from call sites.

internal/operator-controller/controllers/clusterobjectset_controller.go
```diff
@@ -416,12 +416,27 @@ func (c *ClusterObjectSetReconciler) removeFinalizer(ctx context.Context, obj cl
...
+func (c *ClusterObjectSetReconciler) listActiveRevisions(
+	ctx context.Context,
+	cos *ocv1.ClusterObjectSet,
```
pedjak, line 436: **suggestion:** `include` reads like a collection, not a function. A name like `match`, `predicate`, or `filter` would better convey that this is a test applied to each revision.

pedjak (commented):
Review focused on boxcutter library update and coding style.

internal/operator-controller/controllers/clusterobjectset_controller_internal_test.go
```diff
@@ -30,6 +30,125 @@ import (
...
+				names = append(names, rev.GetName())
+			}
+
+			require.ElementsMatch(t, tc.expectedRevs, names)
+		})
+	}
+}
+
 func Test_ClusterObjectSetReconciler_listPreviousRevisions(t *testing.T) {
 	testScheme := runtime.NewScheme()
```
pedjak, line 131: **suggestion:** This test and `Test_ClusterObjectSetReconciler_listPreviousRevisions` below share nearly identical structure — same test case shapes, same helpers, same setup patterns for archived/deleting/cross-extension scenarios. The only real difference is which method is called and the expected results.

Since both now delegate to `listOtherActiveRevisions` (proposed rename), these could be consolidated into a single table-driven test that parameterizes the method under test, or tests `listOtherActiveRevisions` directly with different predicates, reducing ~120 lines of near-duplicate setup.

internal/operator-controller/controllers/revision_engine_factory.go
```diff
@@ -69,7 +69,9 @@ func (f *defaultRevisionEngineFactory) CreateRevisionEngine(_ context.Context, r
 			machinery.NewObjectEngine(
 				f.Scheme, f.TrackingCache, scopedClient,
 				machinery.NewComparator(f.DiscoveryClient, f.Scheme, f.FieldOwnerPrefix),
-				f.FieldOwnerPrefix, f.FieldOwnerPrefix, scopedClient,
+				f.FieldOwnerPrefix, f.FieldOwnerPrefix,
+				"", // managedBy: empty string defaults to "boxcutter"
```
pedjak: why not setting to something like `"olm"` or `"operator-controller"`? What if boxcutter changes the treatment of empty string in the future and sets to something else than boxcutter?

perdasilva: I went for using the field owner prefix `olm.operatorframework.io`

pedjak (commented):
/lgtm

some e2e test are failing though

test/e2e/features/install.feature
```diff
@@ -678,3 +678,100 @@ Feature: Install ClusterExtension
     Then ClusterExtension is available
     And ClusterExtension reports Progressing as True with Reason Succeeded
     And ClusterExtension reports Installed as True
+
```
pedjak: **suggestion:** This scenario has three When-Then cycles (install first CE → assert available, apply dup CE → assert collision, update first CE → assert still colliding). Multiple When-Then blocks in a single Gherkin scenario is generally considered an anti-pattern — each scenario should specify one behavior, not a multi-step integration script.

Having multiple transitions makes it harder to diagnose which step failed and why, and the scenario title ("does not take over resources") only describes the final assertion.

Consider splitting into two scenarios:
1. Conflicting CE gets collision on install
2. Conflicting CE with higher revision still gets collision after update

They could share setup via a `Background` or a reusable Given step.

Also, this scenario involves updating the CE spec (adding `DeploymentConfig`) and asserting behavior *after* the update. That makes it a better fit for `update.feature` than `install.feature`. The earlier Copilot review also referenced this being in `update.feature` — was there a reason to move it here?

internal/operator-controller/controllers/clusterobjectset_controller_internal_test.go
```diff
@@ -26,23 +26,38 @@ import (
 	"github.com/operator-framework/operator-controller/internal/operator-controller/labels"
 )
 
-func Test_ClusterObjectSetReconciler_listPreviousRevisions(t *testing.T) {
+func Test_ClusterObjectSetReconciler_listOtherActiveRevisions(t *testing.T) {
 	testScheme := runtime.NewScheme()
 	require.NoError(t, ocv1.AddToScheme(testScheme))
 
+	type methodUnderTest struct {
+		name string
```
pedjak: **suggestion:** The double loop (test cases × methods) with a `map[string][]string` for expected results works, but it obscures the arrange-act-assert structure. A reader has to mentally cross-reference the map key with the method definition at the top to understand what any single test case asserts.

An alternative structure that would be clearer:
- Test `listOtherActiveRevisions` directly with explicit predicates (the real unit under test). This covers the shared filtering logic once.
- Add a couple of trivial tests for `listSiblingRevisions` and `listPreviousRevisions` that just verify they delegate with the correct predicate (e.g., one passes all, the other filters by revision number).

This separates "does the shared logic work" from "do the wrappers delegate correctly" and makes each test read as a clean given-when-then.

internal/operator-controller/controllers/clusterobjectset_controller_internal_test.go
```diff
@@ -134,36 +145,41 @@ func Test_ClusterObjectSetReconciler_listPreviousRevisions(t *testing.T) {
...
-			require.NoError(t, err)
-
-			names := make([]string, 0, len(previous))
-			for _, rev := range previous {
-				names = append(names, rev.GetName())
-			}
-
-			require.ElementsMatch(t, tc.expectedRevs, names)
-		})
+		for _, m := range methods {
```
pedjak: **suggestion:** The name format `fmt.Sprintf("%s/%s", m.name, tc.name)` puts the method name first, but the outer loop iterates over test cases. This means Go test output interleaves methods:
```
listSiblingRevisions/should exclude self...
listPreviousRevisions/should exclude self...
listSiblingRevisions/should exclude archived...
listPreviousRevisions/should exclude archived...
```

Swapping to `tc.name/m.name` (or swapping the loop order) would group output by method, making it easier to scan all cases for one method when debugging a failure.

internal/operator-controller/controllers/clusterobjectset_controller_internal_test.go
```diff
@@ -73,35 +87,33 @@ func Test_ClusterObjectSetReconciler_listPreviousRevisions(t *testing.T) {
...
-			//   - We list previous revisions.
-			//   - The deleting one is filtered out.
-			name: "should drop deleting revisions when listing previous",
+			name: "should exclude deleting revisions",
 			existingObjs: func() []client.Object {
 				ext := newTestClusterExtensionInternal()
 				rev1 := newTestClusterObjectSetInternal(t, "rev-1")
+				rev1.Finalizers = []string{"test-finalizer"}
+				rev1.DeletionTimestamp = &metav1.Time{Time: time.Now()}
 				rev2 := newTestClusterObjectSetInternal(t, "rev-2")
```
pedjak, line 75: **suggestion:** The old test marked `rev-2` (middle revision) as deleting with `currentRev: "rev-3"`, which tested the case where a *middle* revision is deleting while both lower and higher revisions exist. The new test moves the deletion to `rev-1` (lowest revision). Both validate the "skip deleting" filter, but the old arrangement was a more interesting interleaving — it verified that a deleting revision between two active ones is correctly excluded regardless of its position.

Consider keeping deletion on `rev-2` and adjusting the expected results accordingly, or adding a second sub-case for middle-revision deletion.

internal/operator-controller/controllers/clusterobjectset_controller.go
```diff
@@ -417,12 +417,27 @@ func (c *ClusterObjectSetReconciler) removeFinalizer(ctx context.Context, obj cl
 	return nil
 }
 
+// listSiblingRevisions returns all active revisions belonging to the same ClusterExtension, excluding the current one.
```
pedjak, line 420: **suggestion:** The godoc here explains *why boxcutter needs siblings* (avoiding false collisions during handover). That is caller context — it belongs at the call site in `buildBoxcutterPhases` where `WithSiblingOwners` is passed. The method itself just returns "all active non-self revisions for the same owner," which is already clear from its name and signature.

Keeping method-level docs focused on what the method does (not why a specific caller needs it) prevents the comment from becoming stale if another caller is added.

test/e2e/features/install.feature
```diff
@@ -678,3 +678,100 @@ Feature: Install ClusterExtension
...
+            selector:
+              matchLabels:
+                "olm.operatorframework.io/metadata.name": ${CATALOG:test}
+      """
+    Then ClusterExtension "${NAME}" has 2 ClusterObjectSets
+    # The higher-revision COS (revision 2) should also collide, not take over resources
+    And ClusterExtension reports Progressing as True with Reason Retrying and Message includes:
+      """
+      revision object collisions
+      """
```
pedjak: **suggestion:** The scenario asserts that the dup CE shows collision and that the higher revision still shows collision after update. But it doesn't verify that the *original* `${NAME}` CE remains Available/Installed throughout the conflict. Adding something like `And ClusterExtension "${NAME}" reports Installed as True` after the dup's collision assertion would strengthen the test — confirming the original owner is unaffected is arguably the most important property to verify here.

pedjak (commented):
Review focused on test structure (unit + e2e) and a minor code comment placement nit.

test/e2e/features/install.feature
perdasilva: Moved the test to `update.feature` — agreed it fits better there.

On splitting: the "conflicting CE gets collision on install" case is already covered by the scenario right above ("Detect collision when a second ClusterExtension installs the same package after an upgrade"). The value of this scenario is the multi-step sequence: initial collision → spec change producing a higher-revision COS → still collision. Splitting would duplicate the first half of the existing test without adding coverage.

Also added assertions that the original CE remains `Installed=True` after both collision points, which was missing before.

internal/operator-controller/controllers/clusterobjectset_controller_internal_test.go
perdasilva: This is addressed — `listOtherActiveRevisions` and `listSiblingRevisions` have been removed. The current code only has `listPreviousRevisions` with a simple single-loop table-driven test. No more double loop or map cross-referencing.

internal/operator-controller/controllers/clusterobjectset_controller_internal_test.go
perdasilva: Also addressed — the double loop is gone entirely. Single loop over test cases calling `listPreviousRevisions` directly.

internal/operator-controller/controllers/clusterobjectset_controller_internal_test.go
perdasilva: Addressed — deletion is back on `rev-2` (middle revision) with `currentRev: "rev-3"`, matching the original arrangement.

internal/operator-controller/controllers/clusterobjectset_controller.go
perdasilva: Addressed — `listSiblingRevisions` and `WithSiblingOwners` are gone. The remaining `listPreviousRevisions` godoc only describes what the method does, no caller context.

test/e2e/features/install.feature
perdasilva: Agreed — added `ClusterExtension "ce-${SCENARIO_ID}" reports Installed as True` after both collision points (initial collision and post-update higher-revision collision).

test/e2e/features/update.feature
```diff
@@ -338,3 +338,97 @@ Feature: Update ClusterExtension
...
+          catalog:
+            packageName: ${PACKAGE:test}
+            version: 1.0.0
+            selector:
+              matchLabels:
+                "olm.operatorframework.io/metadata.name": ${CATALOG:test}
+      """
+    And ClusterExtension is rolled out
+    And ClusterExtension is available
+    And ClusterExtension "${NAME}" has 1 ClusterObjectSet
```
pedjak: perhaps: owns?

test/e2e/features/update.feature
```diff
@@ -338,3 +338,97 @@ Feature: Update ClusterExtension
...
+          catalog:
+            packageName: ${PACKAGE:test}
+            version: 1.0.0
+            selector:
+              matchLabels:
+                "olm.operatorframework.io/metadata.name": ${CATALOG:test}
+      """
+    Then ClusterExtension reports Progressing as True with Reason Retrying and Message includes:
+      """
+      revision object collisions
```
pedjak: could we assert here the full message?

test/e2e/features/update.feature
```diff
@@ -338,3 +338,97 @@ Feature: Update ClusterExtension
...
+        source:
+          sourceType: Catalog
+          catalog:
+            packageName: ${PACKAGE:test}
+            version: 1.0.0
+            selector:
+              matchLabels:
+                "olm.operatorframework.io/metadata.name": ${CATALOG:test}
+      """
+    Then ClusterExtension reports Progressing as True with Reason Retrying and Message includes:
```
pedjak: given that we have applied in this tests two ClusterExtension it would be helpful here to distinguish visually which one reports the condition:

```
ClusterExtension "${NAME}-dup" reports Progressing as True with Reason Retrying and Message includes:

```

test/e2e/features/update.feature
```diff
@@ -338,3 +338,97 @@ Feature: Update ClusterExtension
...
+      """
+      revision object collisions
+      """
+    # Verify the original ClusterExtension remains installed and unaffected
+    And ClusterExtension "ce-${SCENARIO_ID}" reports Installed as True
```
pedjak: should be 

```
ClusterExtension "${NAME}" reports Installed as True
```

because it is very internal detail that NAME has `ce-${SCENARIO_ID}` convention.

test/e2e/features/update.feature
```diff
@@ -338,3 +338,97 @@ Feature: Update ClusterExtension
...
+          catalog:
+            packageName: ${PACKAGE:test}
+            version: 1.0.0
+            selector:
+              matchLabels:
+                "olm.operatorframework.io/metadata.name": ${CATALOG:test}
+      """
+    Then ClusterExtension "${NAME}" has 2 ClusterObjectSets
+    # The higher-revision COS (revision 2) should also collide, not take over resources
+    And ClusterExtension reports Progressing as True with Reason Retrying and Message includes:
```
pedjak: which cluster extension report that?

test/e2e/features/update.feature
```diff
@@ -338,3 +338,97 @@ Feature: Update ClusterExtension
     Then ClusterExtension reports "${NAME}-1, ${NAME}-2" as active revisions
     And ClusterObjectSet "${NAME}-2" reports Progressing as True with Reason RollingOut
     And ClusterObjectSet "${NAME}-2" reports Available as False with Reason ProbeFailure
+
+  @BoxcutterRuntime
+  @DeploymentConfig
+  Scenario: A conflicting ClusterExtension with a higher revision does not take over resources from the original owner
```
pedjak: I find the scenario title to describe a low-level implementation details - we should focus ourselves here on user visible change, something like "Cannot install further ClusterExtension referring already installed bundle".

Given that we are adding this test in the PR which is about upgrading boxcutter version, I am curious to understand if before this update we did not have such protection? If we did, then I would add this tests in a separate PR. Does boxcutter upgrade gives us this functionality automatically or we need to so something on our side to achieve it?

internal/operator-controller/controllers/clusterobjectset_controller_internal_test.go
```diff
@@ -134,36 +145,41 @@ func Test_ClusterObjectSetReconciler_listPreviousRevisions(t *testing.T) {
...
+
+				result, err := m.call(reconciler, t.Context(), currentRev)
+				require.NoError(t, err)
+
+				names := make([]string, 0, len(result))
+				for _, rev := range result {
+					names = append(names, rev.GetName())
+				}
+
+				require.ElementsMatch(t, tc.expectedRevs[m.name], names)
```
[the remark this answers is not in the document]
perdasilva: No longer applicable — the `map[string][]string` and method dispatch loop are gone. Each test case now has a plain `expectedRevs []string` field checked directly against `listPreviousRevisions`.

test/e2e/features/update.feature
perdasilva: Renamed to `owns` — `ClusterExtension "${NAME}" owns 1 ClusterObjectSet`.

test/e2e/features/update.feature
perdasilva: The full message includes the phase index and colliding object details (GVK, namespace/name with scenario IDs), e.g.:

```
revision object collisions in phase 0
ConfigMap.v1 ns-<id>/test-configmap-<id> ...
```

Asserting the full message would be brittle since the object names are scenario-specific. The existing collision test at line 274 uses the same `Message includes` approach with just the fragment. Happy to add a more specific fragment like `"revision object collisions in phase"` if that helps.

test/e2e/features/update.feature
perdasilva: Can't use `${NAME}` here — after the dup CE is applied, `ResourceIsApplied` overwrites `sc.clusterExtensionName` to the dup's name, so `${NAME}` resolves to the dup (which is retrying, not installed). We need `ce-${SCENARIO_ID}` to reference the original CE.

`${SCENARIO_ID}` is a first-class template variable exposed by the framework, and the `ce-` prefix is established in `CreateScenarioContext` (`hooks.go:218`). The existing `TrackCurrentClusterExtensionForCleanup` pattern has the same constraint — once you apply a second CE, you lose the `${NAME}` reference to the first one.

An alternative would be adding a new template variable (e.g. `${ORIGINAL_NAME}`) that `TrackCurrentClusterExtensionForCleanup` populates, but that feels like over-engineering for one test.

test/e2e/features/update.feature
perdasilva: Added a comment above the assertion clarifying which CE is being checked: `# The conflicting ClusterExtension (${NAME}-dup, now tracked as ${NAME}) should be retrying`. The unnamed `ClusterExtension reports` steps use `sc.clusterExtensionName` which is the dup at that point, but that's not obvious to a reader.

A fully named variant like `ClusterExtension "${NAME}-dup" reports ... with Reason ... and Message includes` would require a new step definition that accepts name+type+status+reason+message — seemed like over-engineering for this one test.

test/e2e/features/update.feature
perdasilva: The conflicting ClusterExtension (the dup, tracked as `${NAME}` after `ResourceIsApplied`). The comment on the line above clarifies this: "The higher-revision COS (revision 2) should also collide, not take over resources." Same pattern as the first collision assertion where I added a clarifying comment in the latest push.

test/e2e/features/update.feature
perdasilva: Renamed to "Cannot install a ClusterExtension that refers to an already installed bundle".

On the scope question: cross-CE collision protection is existing boxcutter behavior — it was there before this PR. This PR changes the sibling owners API from `WithSiblingOwners` (all active revisions from the same CE) to `WithPreviousOwners` (only lower-revision ones). That's a same-CE handover concern, not cross-CE. The e2e test was added as a regression test to confirm that cross-CE collision protection still works correctly after the API change, but it's not testing new functionality. Happy to split it out into a separate PR if you prefer.

internal/operator-controller/controllers/clusterobjectset_controller_internal_test.go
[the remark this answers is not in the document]
perdasilva: No longer applicable — the map and method dispatch loop are gone. Each test case now has a plain `expectedRevs []string` checked directly.

internal/operator-controller/controllers/revision_engine_factory.go
```diff
@@ -69,7 +69,9 @@ func (f *defaultRevisionEngineFactory) CreateRevisionEngine(_ context.Context, r
 			machinery.NewObjectEngine(
 				f.Scheme, f.TrackingCache, scopedClient,
 				machinery.NewComparator(f.DiscoveryClient, f.Scheme, f.FieldOwnerPrefix),
-				f.FieldOwnerPrefix, f.FieldOwnerPrefix, scopedClient,
+				f.FieldOwnerPrefix, f.FieldOwnerPrefix,
+				f.FieldOwnerPrefix,
+				scopedClient,
```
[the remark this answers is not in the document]
perdasilva: Added `// managedBy` inline comment on the third `f.FieldOwnerPrefix` argument.

internal/operator-controller/controllers/clusterobjectset_controller_internal_test.go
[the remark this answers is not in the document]
perdasilva: No longer applicable — the map and method dispatch loop are gone. Each test case now has a plain `expectedRevs []string` checked directly against `listPreviousRevisions`.

test/e2e/features/update.feature
pedjak: > Can't use `${NAME}` here — after the dup CE is applied, `ResourceIsApplied` overwrites `sc.clusterExtensionName` to the dup's name, so `${NAME}` resolves to the dup (which is retrying, not installed). We need `ce-${SCENARIO_ID}` to reference the original CE.

We should update `ResourceIsApplied` or some other e2e test logic so that we can have clearer step descriptions.

test/e2e/features/update.feature
pedjak: > The conflicting ClusterExtension (the dup, tracked as `${NAME}` after `ResourceIsApplied`). The comment on the line above clarifies this: "The higher-revision COS (revision 2) should also collide, not take over resources." Same pattern as the first collision assertion where I added a clarifying comment in the latest push.

For reader it should be better if we could have `ClusterExtension ${NAME}-dup ...`

test/e2e/features/update.feature
pedjak: > Happy to split it out into a separate PR if you prefer.

It would make sense to me... and to merge that first, before upgrading the boxcutter.

test/e2e/features/update.feature
pedjak: ok, but given that we actually forbid installing further ClusterExtensions referring the same bundle, perhaps we could instead of `revision object collisions` message improve the wording so that user knows what is going on (cannot install ClusterExtension that refers already installed bundle by ClusterExtensino X`? wdyt?

perdasilva: this sounds like larger design work would be needed - collision is a general problem it that can happen whether or not your installing the same package. I can create an issue to track this but I think it goes beyond the intended scope. Would you like to pick this one up?

test/e2e/features/update.feature
perdasilva: moving the e2e into another PR and attempting to address this there

test/e2e/features/update.feature
perdasilva: moving the e2e into another PR and attempting to address this there

test/e2e/features/update.feature
perdasilva: moving the e2e into another PR

test/e2e/features/update.feature
perdasilva: #2784

test/e2e/features/update.feature
perdasilva: #2783

test/e2e/features/update.feature
perdasilva: #2783

test/e2e/features/update.feature
perdasilva: #2783

internal/operator-controller/controllers/clusterobjectset_controller_internal_test.go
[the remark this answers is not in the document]
perdasilva: Restructured the tests. The double loop and `map[string][]string` are gone — now two separate test functions (`Test_ClusterObjectSetReconciler_listSiblingRevisions` and `Test_ClusterObjectSetReconciler_listPreviousRevisions`) each with a clean table-driven structure and a shared `callRevisionLister` helper for the common setup/assert logic.

Also fixed the deleting revision test to use `rev-2` (middle revision) and synced the vendor directory.

internal/operator-controller/controllers/clusterobjectset_controller.go
```diff
@@ -472,7 +484,7 @@ func (c *ClusterObjectSetReconciler) buildBoxcutterPhases(ctx context.Context, c
 	}
 
 	opts := []boxcutter.RevisionReconcileOption{
-		boxcutter.WithPreviousOwners(previousObjs),
+		boxcutter.WithSiblingOwners(siblingObjs),
 		boxcutter.WithProbe(boxcutter.ProgressProbeType, progressionProbes),
```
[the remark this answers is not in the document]
perdasilva: The PR description was wrong — it's the other way around. Boxcutter v0.14.0 replaces `WithPreviousOwners` (v0.13.1) with `WithSiblingOwners` (v0.14.0). The code correctly uses `WithSiblingOwners` with `listSiblingRevisions`. Updated the PR description to match.

pedjak (commented):
/lgtm

rashmigottipati (approved):

PR #2770, opened 2026-06-16
✨ Deprecate ClusterExtension spec.serviceAccount

pedjak:
# Description

Deprecate `spec.serviceAccount` on ClusterExtension. OLMv1 is a single-tenant system where users with ClusterExtension write access are effectively delegated cluster-admin trust. With the service account field deprecated, operator-controller uses its own cluster-admin service account for all cluster interactions, eliminating the per-SA permission model and ~5,700 lines of supporting infrastructure.

## Changes

**API deprecation:**
- Mark `spec.serviceAccount` as deprecated with `Deprecated:` godoc convention and `omitzero` tag
- Add ValidatingAdmissionPolicy to emit kubectl deprecation warning when `serviceAccount` is set
- Emit controller log warning when deprecated field is used
- Relax `serviceAccount.name` immutability rule to allow clearing the deprecated field

**Remove per-SA infrastructure:**
- Remove `authorization/` package, `authentication/` package, SA-scoped config mappers
- Remove `k8s.io/kubernetes` dependency and transitive replace directives from go.mod
- Remove `hack/tools/k8smaintainer/` and `k8s-pin` Makefile target
- Remove `PreflightPermissions` and `SyntheticPermissions` feature gates
- Remove pre-auth wiring from appliers and main.go

**Simplify cache architecture:**
- Replace per-CE `contentmanager/` package with shared `TrackingCache` from boxcutter/managedcache
- Single global cache for drift detection instead of per-CE dynamic clients and informer factories
- Remove `RestConfigMapper` pattern

**E2E test cleanup:**
- Remove `serviceAccount` from all feature files, add explicit namespace setup step
- Add deprecation warning scenario
- Remove dead SA step code and templates

**Upgrade safety:**
- Delete old ClusterRoleBinding before apply in install script to handle immutable roleRef change

Removed ~5,700 lines of unneeded code.

## Reviewer Checklist

- [x] API Go Documentation


api/v1/clusterextension_types.go
```diff
@@ -404,8 +401,8 @@ type ServiceAccountReference struct {
 	//
 	// +kubebuilder:validation:MaxLength:=253
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="name is immutable"
-	// +kubebuilder:validation:XValidation:rule="self.matches(\"^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$\")",message="name must be a valid DNS1123 subdomain. It must contain only lowercase alphanumeric characters, hyphens (-) or periods (.), start and end with an alphanumeric character, and be no longer than 253 characters"
-	// +required
+	// +kubebuilder:validation:XValidation:rule="size(self) == 0 || self.matches(\"^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$\")",message="name must be a valid DNS1123 subdomain. It must contain only lowercase alphanumeric characters, hyphens (-) or periods (.), start and end with an alphanumeric character, and be no longer than 253 characters"
+	// +optional
```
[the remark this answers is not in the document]
dtfranz: This was fixed - the validation now allows the field to be cleared

docs/project/olmv1_design_decisions.md
```diff
@@ -176,14 +176,13 @@ OLM v1 will include primitives (e.g. templating) to make it possible to have mul
 
 However, it should be noted that the purpose of these primitives is not to enable multi-tenancy. It is to enable administrators to provide configuration for the installation of an extension. The fact that operators can be packaged as separate bundles and parameterized in a way that permits multiple controller installations is incidental, and not something that OLM v1 will encourage or promote.
 
-### Make OLM secure by default
+### Make OLM secure by safeguarding ClusterExtension permissions
```
dtfranz: Wanted to specifically draw attention to the rewrite here, if anybody can think of a better sub-title and wording in the below paragraph.

docs/project/olmv1_design_decisions.md
```diff
@@ -176,14 +176,13 @@ OLM v1 will include primitives (e.g. templating) to make it possible to have mul
...
+Initially, OLMv1 was designed as secure-by-default and required cluster administrators to create ServiceAccounts and associated RBAC to enable the installation of ClusterExtensions. However, any users granted permission to create ClusterExtension resources are effectively being delegated cluster-admin trust. Requiring administrators to create least-privileged RBAC for ClusterExtension installation does not remove this avenue of privilege escalation. Thus, OLMv1 adopts cluster-admin scope directly to eliminate the complexity and UX burden of requiring users to manage ServiceAccounts with difficult-to-determine permissions, and enables simpler internal operations for namespace management and content lifecycle.
 
-To make OLM secure by default:
+To safeguard against potential escalation attacks, cluster administrators should treat the ability to create ClusterExtensions as equivalent to having cluster-admin priveleges.
```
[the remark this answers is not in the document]
dtfranz: Spelled it right several times then spelled it wrong once 🙃 fixed now.

internal/operator-controller/controllers/clusterextension_reconcile_steps.go
```diff
@@ -98,16 +97,13 @@ func ValidateClusterExtension(validators ...ClusterExtensionValidator) Reconcile
 	}
 }
 
-// ServiceAccountValidator returns a validator that checks if the specified
-// ServiceAccount exists in the cluster by performing a direct Get call.
-func ServiceAccountValidator(saClient corev1client.ServiceAccountsGetter) ClusterExtensionValidator {
+// ServiceAccountDeprecationWarning returns a validator that checks if the deprecated
+// serviceAccount field is populated and logs a warning if found.
+func ServiceAccountDeprecationWarning(saClient corev1client.ServiceAccountsGetter) ClusterExtensionValidator {
```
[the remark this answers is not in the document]
pedjak: Fixed in 4fdb3a48 — removed the unused `saClient` parameter and `corev1client` import.

api/v1/clusterextension_types.go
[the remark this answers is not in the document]
pedjak: Fixed in 4fdb3a48 — relaxed the CEL rule to `self == oldSelf || size(self) == 0`, allowing the deprecated field to be cleared.

helm/olmv1/templates/rbac/clusterrolebinding-operator-controller-manager-rolebinding.yml
```diff
@@ -16,7 +16,7 @@ metadata:
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
-  name: operator-controller-manager-role
+  name: cluster-admin
 subjects:
```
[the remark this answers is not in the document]
pedjak: Fixed in 4fdb3a48 — added `kubectl delete clusterrolebinding operator-controller-manager-rolebinding --ignore-not-found` before `kubectl apply` in `scripts/install.tpl.sh`. Verified with `make test-upgrade-st2st-e2e`.

internal/operator-controller/controllers/suite_test.go
```diff
@@ -136,6 +171,38 @@ func TestMain(m *testing.M) {
 		log.Panic("expected cfg to not be nil")
 	}
 
+	cl, err := client.New(config, client.Options{})
+	utilruntime.Must(err)
+	ctx := context.Background()
```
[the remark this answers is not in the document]
pedjak: Fixed by Daniel in a subsequent commit.

docs/project/olmv1_design_decisions.md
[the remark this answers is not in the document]
pedjak: Fixed by Daniel in a subsequent commit.

internal/operator-controller/controllers/suite_test.go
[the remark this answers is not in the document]
pedjak: Not an issue — `client.New` with nil `Options.Scheme` defaults to `k8s.io/client-go/kubernetes/scheme.Scheme` which includes `admissionregistrationv1`. The types are registered.

internal/operator-controller/applier/helm.go
```diff
@@ -176,12 +153,8 @@ func (h *Helm) Apply(ctx context.Context, contentFS fs.FS, ext *ocv1.ClusterExte
...
-	cache, err := h.Manager.Get(ctx, ext)
-	if err != nil {
-		return true, "", err
-	}
-
-	if err := cache.Watch(ctx, h.Watcher, relObjects...); err != nil {
+	gvks := gvksForObjects(relObjects)
+	if err := h.TrackingCache.Watch(ctx, ext, gvks); err != nil {
 		return true, "", err
 	}
```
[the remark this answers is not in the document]
pedjak: Fixed — added nil guard that returns an error instead of panicking.

.claude/TODO-OCPSTRAT-3040.md
```diff
@@ -0,0 +1,96 @@
+# OCPSTRAT-3040: Adopt cluster-admin scope and deprecate ServiceAccount
```
pedjak: note to reviewers: this file is going to removed from the PR before the merge. It was used to track the work done, given that the work items were split between myself and @dtfranz .

.claude/memory/MEMORY.md
```diff
@@ -0,0 +1 @@
+- [Update OCPSTRAT-3040 TODO](update-ocpstrat-3040-todo.md) — Check off items in `.claude/TODO-OCPSTRAT-3040.md` whenever related work is completed
```
pedjak: note to reviewers: this file is going to removed from the PR before the merge. It was used to track the work done, given that the work items were split between myself and @dtfranz .

.claude/memory/update-ocpstrat-3040-todo.md
```diff
@@ -0,0 +1,12 @@
+---
```
pedjak: note to reviewers: this file is going to removed from the PR before the merge. It was used to track the work done, given that the work items were split between myself and @dtfranz .

joelanford:
> Delete old ClusterRoleBinding before apply in install script to handle immutable roleRef change

This might cause problems in systems that perform upgrades without using our upstream install scripts. Perhaps a safer alternative would be to continue using our existing role, but changing its rules to match what `cluster-admin`'s rules are?

api/v1/clusterextension_types.go
```diff
@@ -67,14 +66,13 @@ type ClusterExtensionSpec struct {
 	// +required
 	Namespace string `json:"namespace"`
 
-	// serviceAccount specifies a ServiceAccount used to perform all interactions with the cluster
-	// that are required to manage the extension.
-	// The ServiceAccount must be configured with the necessary permissions to perform these interactions.
-	// The ServiceAccount must exist in the namespace referenced in the spec.
-	// The serviceAccount field is required.
+	// Deprecated: serviceAccount is no longer used and will be removed in a future release.
```
joelanford: This new GoDoc may still run afoul of API conventions which expect the comment to begin with `serviceAccount ...`.

If so, perhaps something like:

```
serviceAccount is a deprecated field and is completely ignored. OLMv1 is a single-tenant system where users....

Deprecated: serviceAccount is no longer used and will be removed in a future release.
```

internal/operator-controller/controllers/clusterextension_reconcile_steps.go
```diff
@@ -98,16 +96,13 @@ func ValidateClusterExtension(validators ...ClusterExtensionValidator) Reconcile
...
 	return func(ctx context.Context, ext *ocv1.ClusterExtension) error {
-		_, err := saClient.ServiceAccounts(ext.Spec.Namespace).Get(ctx, ext.Spec.ServiceAccount.Name, metav1.GetOptions{})
-		if err != nil {
-			if apierrors.IsNotFound(err) {
-				return fmt.Errorf("service account %q not found in namespace %q", ext.Spec.ServiceAccount.Name, ext.Spec.Namespace)
-			}
-			return fmt.Errorf("error getting service account: %w", err)
+		l := log.FromContext(ctx)
+		if len(ext.Spec.ServiceAccount.Name) > 0 { //nolint:staticcheck // intentional read of deprecated field to emit deprecation warning
+			l.Info("spec.serviceAccount is deprecated, ignored, and will be removed in a future release - operator-controller's cluster-admin service account is used for all cluster interactions")
```
joelanford, line 105: Users will see admission-time deprecation warnings with the VAP, right? What is the audience for a deprecation warning in the logs? I'm wondering if we can just remove this step from the reconciler? Or perhaps change it to a gauge metric if we care about tracking the field's continued use?

cmd/operator-controller/main.go
```diff
@@ -460,15 +460,40 @@ func run() error {
...
+		return err
+	}
+
 	var ctrlBuilderOpts []controllers.ControllerBuilderOption
 	if features.OperatorControllerFeatureGate.Enabled(features.BoxcutterRuntime) {
 		ctrlBuilderOpts = append(ctrlBuilderOpts, controllers.WithOwns(&ocv1.ClusterObjectSet{}))
 	}
+	ctrlBuilderOpts = append(ctrlBuilderOpts, controllers.WithWatchesRawSource(
+		trackingCache.Source(
+			crhandler.EnqueueRequestForOwner(mgr.GetScheme(), mgr.GetRESTMapper(), &ocv1.ClusterExtension{}),
```
joelanford: The owner object for the managed objects is different between boxcutter and helm, right?
- For helm, the owner is the ClusterExtension
- For boxcutter, the owner is the ClusterObjectSet.

joelanford:
/hold

Due to the sensitivity of this change (we _are_ removing some behaviors that users may be expecting), let's make sure we have an approved design doc in place before this actually merges.

joelanford:
Why are we overriding the api-diff-lint failure? Those `New Issues` seem like things that are fixable?

pedjak:
> This might cause problems in systems that perform upgrades without using our upstream install scripts. 

What other installation method you have in mind, given that our releases notes points users to use install script?

> Perhaps a safer alternative would be to continue using our existing role, but changing its rules to match what `cluster-admin`'s rules are?

Rather than copying `cluster-admin`'s rules into a custom ClusterRole (which would be a redundant replica of a built-in role with wildcard permissions), or
 relying on install script workarounds, the cleanest solution is to **rename the ClusterRoleBinding**.

If we change the name to e.g. `operator-controller-cluster-admin-rolebinding`, the upgrade creates a new CRB with the correct roleRef instead of trying to mutate the immutable roleRef on the old one. The old CRB (`operator-controller-manager-rolebinding`) becomes orphaned — it points to the now-deleted  `operator-controller-manager-role` ClusterRole, so it's effectively a no-op and can be cleaned up separately.

  This works regardless of the install method (script, Helm, GitOps) without any pre-delete steps.

internal/operator-controller/controllers/clusterextension_reconcile_steps.go
pedjak: sure, we can remove it - no harm.

internal/operator-controller/controllers/suite_test.go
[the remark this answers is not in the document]
pedjak: Fixed in the latest push. The earlier run had 4 NEW issues:

**`commentstart` (2 issues)** — godoc for deprecated fields started with `Deprecated:` (Go convention) instead of the JSON field name (OpenShift convention). Restructured to follow the [openshift/api pattern](https://github.com/openshift/api/blob/master/operator/v1/types_network.go) for deprecated fields: field name first, `DEPRECATED:` marker in the body.

**`optionalfields` (2 issues)** — the linter suggests making `ServiceAccount` and `Name` pointers. Suppressed with `//nolint:kubeapilinter` because:
- The [OpenShift API conventions](https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md) explicitly advise against pointers for CRD fields unless needed to distinguish zero from unset
- `ServiceAccount` already uses `omitzero` (Go 1.24+) for correct zero-value serialization
- Making a deprecated field a pointer would be a breaking API change for no benefit

Current api-diff-lint result: **FIXED: 2, NEW: 0**.

cmd/operator-controller/main.go
pedjak: Good catch. The two runtimes are mutually exclusive (feature gate), so in practice:

- **Helm mode**: CE controller gets the TrackingCache Source with `EnqueueRequestForOwner(&ClusterExtension{})` — correct, Helm-managed objects have CE ownerRefs via the `DefaultPostRendererFunc`.
- **Boxcutter mode**: COS controller gets its own TrackingCache Source with `EnqueueRequestForOwner(&ClusterObjectSet{})` at `clusterobjectset_controller.go:353`. The CE controller's Source (line 483) is registered but effectively a no-op since no Helm `Watch()` calls populate it.

That said, the CE-owner Source registration should be gated behind `!BoxcutterRuntime` to avoid the unnecessary no-op wiring. Will fix.

test/e2e/features/revision.feature
```diff
@@ -14,8 +14,6 @@ Feature: Install ClusterObjectSet
       kind: ClusterObjectSet
       metadata:
         annotations:
-          olm.operatorframework.io/service-account-name: pvc-probe-sa
-          olm.operatorframework.io/service-account-namespace: ${TEST_NAMESPACE}
         name: ${COS_NAME}
```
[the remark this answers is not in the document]
dtfranz: Fixed all of these.

test/extension-developer-e2e/extension_developer_test.go
```diff
@@ -162,9 +162,6 @@ func TestExtensionDeveloper(t *testing.T) {
 				},
 			},
 			Namespace: installNamespace,
-			ServiceAccount: ocv1.ServiceAccountReference{
-				Name: sa.Name,
-			},
 		},
```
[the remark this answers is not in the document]
dtfranz: Fixed

test/e2e/features/recover.feature
```diff
@@ -8,7 +8,7 @@ Feature: Recover cluster extension from errors that might occur during its lifet
       | test    | 1.2.0   | beta    |          | CRD, Deployment, ConfigMap |
 
   Scenario: Restore removed resource
-    Given ServiceAccount "olm-sa" with needed permissions is available in test namespace
+    And namespace "${TEST_NAMESPACE}" is available
     And ClusterExtension is applied
```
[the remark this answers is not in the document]
dtfranz: Fixed

test/e2e/features/recover.feature
```diff
@@ -52,12 +48,12 @@ Feature: Recover cluster extension from errors that might occur during its lifet
...
     And ClusterExtension reports Progressing as True with Reason Retrying
-    When ServiceAccount "olm-sa" with needed permissions is available in test namespace
+    When namespace "${TEST_NAMESPACE}" is available
     Then ClusterExtension is available
     And ClusterExtension reports Progressing as True with Reason Succeeded
 
   Scenario: Install ClusterExtension after conflicting resource is removed
-    Given ServiceAccount "olm-sa" with needed permissions is available in test namespace
+    And namespace "${TEST_NAMESPACE}" is available
     And resource is applied
```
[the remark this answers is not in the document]
dtfranz: Fixed

test/e2e/features/recover.feature
```diff
@@ -205,7 +162,7 @@ Feature: Recover cluster extension from errors that might occur during its lifet
     # - Reconciliation completing (observedGeneration == generation) proves the spec was processed
     # - Progressing=Succeeded proves the controller didn't block on missing catalog
     # - Extension staying Available proves workload continues running
-    Given ServiceAccount "olm-sa" with needed permissions is available in test namespace
+    And namespace "${TEST_NAMESPACE}" is available
     And ClusterExtension is applied
```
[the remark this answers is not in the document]
dtfranz: Fixed

test/e2e/features/recover.feature
```diff
@@ -165,7 +124,7 @@ Feature: Recover cluster extension from errors that might occur during its lifet
     # - If the controller stopped reconciling, the configmap would stay deleted
     # - Resource restoration is an observable event that PROVES active reconciliation
     # - The deployment staying healthy proves the workload continues running
-    Given ServiceAccount "olm-sa" with needed permissions is available in test namespace
+    And namespace "${TEST_NAMESPACE}" is available
     And ClusterExtension is applied
```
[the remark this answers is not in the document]
dtfranz: Fixed

api/v1/clusterextension_types.go
pedjak: Adopted your suggested format — field description first, `Deprecated:` at the end. This satisfies both OpenShift convention (starts with field name) and Go convention (`Deprecated:` recognized by godoc/staticcheck). Updated in the squashed commits.

scripts/install.tpl.sh
```diff
@@ -116,6 +116,10 @@ if [ -f "${olmv1_manifest}" ]; then
     olmv1_manifest=file://localhost$(realpath ${olmv1_manifest})
 fi
 
+# Delete the old ClusterRoleBinding if it exists — Kubernetes does not allow
+# changing roleRef via kubectl apply, so we must delete before re-creating.
+kubectl delete clusterrolebinding operator-controller-manager-rolebinding --ignore-not-found
```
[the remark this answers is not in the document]
pedjak: Fixed — now also deletes the stale `operator-controller-manager-role` ClusterRole alongside the CRB.

test/extension-developer-e2e/extension_developer_test.go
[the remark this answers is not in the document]
pedjak: Already addressed — the entire file was deleted as part of the e2e cleanup.

api/v1/clusterextension_types.go
```diff
@@ -67,14 +66,15 @@ type ClusterExtensionSpec struct {
...
-	// that are required to manage the extension.
-	// The ServiceAccount must be configured with the necessary permissions to perform these interactions.
-	// The ServiceAccount must exist in the namespace referenced in the spec.
-	// The serviceAccount field is required.
+	// serviceAccount was previously used to specify the service account for
+	// extension management.
+	// DEPRECATED: serviceAccount is no longer used and will be removed in a future release.
+	// OLMv1 is a single-tenant system where users with ClusterExtension write access are
+	// effectively delegated cluster-admin trust. The operator-controller runs with
+	// cluster-admin privileges and uses its own service account for all cluster interactions.
```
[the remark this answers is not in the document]
pedjak: Fixed — restructured to use Go-standard `Deprecated:` at the end of the comment, per Joe's suggestion.

api/v1/clusterextension_types.go
```diff
@@ -378,12 +378,14 @@ type CatalogFilter struct {
 	UpgradeConstraintPolicy UpgradeConstraintPolicy `json:"upgradeConstraintPolicy,omitempty"`
 }
 
-// ServiceAccountReference identifies the serviceAccount used fo install a ClusterExtension.
+// ServiceAccountReference was previously used to reference a service account for
+// extension management.
+// DEPRECATED: ServiceAccountReference is no longer used and will be removed in a future release.
```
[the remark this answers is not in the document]
pedjak: Fixed — restructured to use Go-standard `Deprecated:` at the end of the comment, per Joe's suggestion.

api/v1/clusterextension_types.go
```diff
@@ -378,12 +378,14 @@ type CatalogFilter struct {
...
 type ServiceAccountReference struct {
-	// name is a required, immutable reference to the name of the ServiceAccount used for installation
-	// and management of the content for the package specified in the packageName field.
-	//
-	// This ServiceAccount must exist in the installNamespace.
+	// name was previously a required, immutable reference to the name of the
+	// ServiceAccount used for installation and management of the content for
+	// the package specified in the packageName field.
+	// DEPRECATED: name is no longer used and will be removed in a future release.
```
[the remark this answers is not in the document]
pedjak: Fixed — restructured to use Go-standard `Deprecated:` at the end of the comment, per Joe's suggestion.

test/e2e/features/revision.feature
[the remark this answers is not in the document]
pedjak: Not an issue in the current state — the `annotations:` line was removed entirely along with the SA annotation values. No empty annotations key remains.

test/e2e/features/recover.feature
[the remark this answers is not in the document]
pedjak: Fixed — Daniel corrected these in a subsequent commit.

test/e2e/steps/steps.go
```diff
@@ -1335,6 +1345,23 @@ func applyPermissionsToServiceAccount(ctx context.Context, serviceAccount, rbacT
...
+	_, thisFile, _, _ := runtime.Caller(0)
+	yaml, err := templateYaml(filepath.Join(filepath.Dir(thisFile), "testdata", "namespace-template.yaml"), map[string]string{
+		"NAMESPACE": ns,
+	})
+	if err != nil {
+		return fmt.Errorf("failed to template namespace yaml: %v", err)
+	}
+	if _, _, err := k8scliWithInput(yaml, "apply", "-f", "-"); err != nil {
+		return fmt.Errorf("failed to apply namespace %s: %w", ns, err)
+	}
```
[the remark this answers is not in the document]
pedjak: Valid suggestion for a follow-up. The stderr is captured but not surfaced in the error message. Not blocking for this PR since e2e tests pass.

api/v1/clusterextension_types.go
```diff
@@ -67,14 +66,15 @@ type ClusterExtensionSpec struct {
...
-	// serviceAccount specifies a ServiceAccount used to perform all interactions with the cluster
-	// that are required to manage the extension.
-	// The ServiceAccount must be configured with the necessary permissions to perform these interactions.
-	// The ServiceAccount must exist in the namespace referenced in the spec.
-	// The serviceAccount field is required.
+	// serviceAccount was previously used to specify the service account for
+	// extension management.
+	// DEPRECATED: serviceAccount is no longer used and will be removed in a future release.
+	// OLMv1 is a single-tenant system where users with ClusterExtension write access are
+	// effectively delegated cluster-admin trust. The operator-controller runs with
```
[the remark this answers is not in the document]
pedjak: Fixed — restructured to use Go-standard `Deprecated:` at the end of the comment.

api/v1/clusterextension_types.go
```diff
@@ -378,12 +378,14 @@ type CatalogFilter struct {
 	UpgradeConstraintPolicy UpgradeConstraintPolicy `json:"upgradeConstraintPolicy,omitempty"`
 }
 
-// ServiceAccountReference identifies the serviceAccount used fo install a ClusterExtension.
+// ServiceAccountReference was previously used to reference a service account for
+// extension management.
+// DEPRECATED: ServiceAccountReference is no longer used and will be removed in a future release.
 type ServiceAccountReference struct {
```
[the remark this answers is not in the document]
pedjak: Fixed — restructured to use Go-standard `Deprecated:` at the end of the comment.

api/v1/clusterextension_types.go
```diff
@@ -378,12 +378,14 @@ type CatalogFilter struct {
...
-	// name is a required, immutable reference to the name of the ServiceAccount used for installation
-	// and management of the content for the package specified in the packageName field.
-	//
-	// This ServiceAccount must exist in the installNamespace.
+	// name was previously a required, immutable reference to the name of the
+	// ServiceAccount used for installation and management of the content for
+	// the package specified in the packageName field.
+	// DEPRECATED: name is no longer used and will be removed in a future release.
 	//
```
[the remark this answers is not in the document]
pedjak: Fixed — restructured to use Go-standard `Deprecated:` at the end of the comment.

test/e2e/features/revision.feature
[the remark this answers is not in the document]
pedjak: ValidatingAdmissionPolicy is GA since Kubernetes 1.30. OLMv1 targets k8s 1.30+ (client-go v0.36.1). Adding a `.Capabilities.APIVersions.Has` guard is a reasonable follow-up but not needed for the current target versions.

test/extension-developer-e2e/extension_developer_test.go
```diff
@@ -162,126 +151,8 @@ func TestExtensionDeveloper(t *testing.T) {
...
-		},
-		RoleRef: rbacv1.RoleRef{
-			APIGroup: "rbac.authorization.k8s.io",
-			Kind:     "ClusterRole",
-			Name:     cr.Name,
 		},
 	}
-	require.NoError(t, c.Create(ctx, crb))
 
 	t.Logf("When creating an ClusterExtension that references a package with a %q bundle type", clusterExtension.Name)
```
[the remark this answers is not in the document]
dtfranz: Fixed

docs/project/olmv1_design_decisions.md
```diff
@@ -176,14 +176,13 @@ OLM v1 will include primitives (e.g. templating) to make it possible to have mul
 
 However, it should be noted that the purpose of these primitives is not to enable multi-tenancy. It is to enable administrators to provide configuration for the installation of an extension. The fact that operators can be packaged as separate bundles and parameterized in a way that permits multiple controller installations is incidental, and not something that OLM v1 will encourage or promote.
 
-### Make OLM secure by default
+### Secure acces to ClusterExtensions through API access RBAC and ValidatingAdmissionPolicy
```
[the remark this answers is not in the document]
dtfranz: Fixed

test/extension-developer-e2e/extension_developer_test.go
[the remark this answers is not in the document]
dtfranz: Fixed

helm/olmv1/templates/validatingadmissionpolicy-clusterextension-serviceaccount-deprecated.yml
```diff
@@ -0,0 +1,26 @@
...
+metadata:
+  name: clusterextension-serviceaccount-deprecated
+  labels:
+    app.kubernetes.io/name: operator-controller
+    {{- include "olmv1.labels" . | nindent 4 }}
+  annotations:
+    {{- include "olmv1.annotations" . | nindent 4 }}
+spec:
+  matchConstraints:
+    resourceRules:
```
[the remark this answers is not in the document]
dtfranz: Added

internal/operator-controller/controllers/suite_test.go
```diff
@@ -128,6 +164,38 @@ func TestMain(m *testing.M) {
...
+	utilruntime.Must(cl.Create(ctx, &admissionregistrationv1.ValidatingAdmissionPolicy{
+		ObjectMeta: metav1.ObjectMeta{Name: "clusterextension-serviceaccount-deprecated"},
+		Spec: admissionregistrationv1.ValidatingAdmissionPolicySpec{
+			MatchConstraints: &admissionregistrationv1.MatchResources{
+				ResourceRules: []admissionregistrationv1.NamedRuleWithOperations{{
```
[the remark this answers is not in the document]
dtfranz: Added - ignored the removed line though which breaks everything 🤔

docs/project/olmv1_design_decisions.md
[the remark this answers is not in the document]
dtfranz: Fixed

internal/operator-controller/controllers/clusterextension_admission_test.go
```diff
@@ -393,6 +380,9 @@ func TestClusterExtensionAdmissionServiceAccount(t *testing.T) {
 				require.Error(t, err)
 				require.Contains(t, err.Error(), tc.errMsg)
 			}
+			if tc.warnMsg != "" {
+				require.True(t, collector.hasWarning(tc.warnMsg), "expected deprecation warning containing %q", tc.warnMsg)
+			}
```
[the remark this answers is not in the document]
dtfranz: Fixed

helm/olmv1/templates/rbac/clusterrolebinding-operator-controller-manager-rolebinding.yml
```diff
@@ -8,15 +8,11 @@ metadata:
...
-  name: operator-controller-manager-admin-rolebinding
-{{- else }}
-  name: operator-controller-manager-rolebinding
-{{- end }}
+  name: operator-controller-cluster-admin-rolebinding
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
-  name: operator-controller-manager-role
+  name: cluster-admin
```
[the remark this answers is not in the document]
dtfranz: Fixed

api/v1/clusterextension_types.go
```diff
@@ -150,7 +150,7 @@ type SourceConfig struct {
 }
 
 // ClusterExtensionInstallConfig is a union which selects the clusterExtension installation config.
-// ClusterExtensionInstallConfig requires the namespace and serviceAccount which should be used for the installation of packages.
+// ClusterExtensionInstallConfig requires the namespace which should be used for the installation of packages.
```
joelanford: I think this entire sentence is stale and can be removed. At one point, SA and namespace were in the `spec.install` struct, but we promoted them to the `spec`.

internal/operator-controller/controllers/suite_test.go
```diff
@@ -128,6 +164,39 @@ func TestMain(m *testing.M) {
...
+			}},
+		},
+	}))
+	utilruntime.Must(cl.Create(ctx, &admissionregistrationv1.ValidatingAdmissionPolicyBinding{
+		ObjectMeta: metav1.ObjectMeta{Name: "clusterextension-serviceaccount-deprecated"},
+		Spec: admissionregistrationv1.ValidatingAdmissionPolicyBindingSpec{
+			PolicyName:        "clusterextension-serviceaccount-deprecated",
+			ValidationActions: []admissionregistrationv1.ValidationAction{admissionregistrationv1.Warn},
+		},
+	}))
```
joelanford: This feels fragile. Can we create these based on what's checked into our helm manifests? Similar to how we create the CRDs, I assume?

test/e2e/features/install.feature
```diff
@@ -420,13 +361,10 @@ Feature: Install ClusterExtension
       spec:
         namespace: ${TEST_NAMESPACE}
         progressDeadlineMinutes: 1
-        serviceAccount:
-          name: olm-sa
         source:
           sourceType: Catalog
           catalog:
             packageName: ${PACKAGE:test}
-            # bundle refers bad image references, so that the deployment never becomes available
```
joelanford, line 429: Why is this comment removed?

go.mod
```diff
@@ -275,54 +270,12 @@ replace k8s.io/cli-runtime => k8s.io/cli-runtime v0.36.1
 
 replace k8s.io/client-go => k8s.io/client-go v0.36.1
```
joelanford: I _think_ we should be able to remove _all_ of the replaces lines now, and update the `require` versions to all be consistently set to 0.36.1.

docs/project/olmv1_design_decisions.md
```diff
@@ -176,14 +176,13 @@ OLM v1 will include primitives (e.g. templating) to make it possible to have mul
...
+Initially, OLMv1 was designed as secure-by-default and required cluster administrators to create ServiceAccounts and associated RBAC to enable the installation of ClusterExtensions. However, any users granted permission to create ClusterExtension resources are effectively being delegated cluster-admin trust. Requiring administrators to create least-privileged RBAC for ClusterExtension installation does not remove this avenue of privilege escalation. Thus, OLMv1 adopts cluster-admin scope directly to eliminate the complexity and UX burden of requiring users to manage ServiceAccounts with difficult-to-determine permissions, and enables simpler internal operations for namespace management and content lifecycle.
```
joelanford: Maybe this is okay? But I also wonder if we should just use present tense here and just explain what the current/new security model is without referencing the past.

Maybe @michaelryanpeter would have some guidance for us here.

docs/howto/how-to-protect-olmv1-api-access.md
```diff
@@ -0,0 +1,276 @@
+# Protecting OLMv1 API Access
+
+Because the operator-controller uses `cluster-admin` level privileges in its normal operations, it is important to lock down access to its associated APIs in order to prevent unauthorized or potentially malicious package installations. Access to the `ClusterExtension` and `ClusterObjectSet` APIs should ideally be limited to cluster administrators and trusted users.
```
joelanford: We should start off here saying that by default, only Cluster Admins have any permission whatsoever to read/write/delete the ClusterExtension and ClusterObjectSet APIs.

Then we should say that cluster admins are generally discouraged from delegating permission to non-cluster admins fore $reasons.

_Then_ we can say, but if you are dead set on doing that, then here's how....

docs/howto/how-to-protect-olmv1-api-access.md
```diff
@@ -0,0 +1,276 @@
...
+spec:
+  policyName: ce-policy-team-monitoring
+  validationActions: ["Deny"]
+```
+
+The following custom validations can be applied through native Kubernetes' `ValidatingAdmissionPolicy` API. Each snippet shows what to add to the `validations:` list in the policy above; combine multiple entries to enforce several constraints in a single policy.
+
+### Allow a user to create/manage only specific ClusterExtensions
+
+To restrict which `ClusterExtension` names a group may manage, add a name-prefix validation. In this example, members of `team-monitoring` may only create or update `ClusterExtension` resources whose names begin with `monitoring-`:
```
joelanford: I'd drop this example. This doesn't meaningfully restrict anything because the CE name is arbitrary. This maybe makes sense as more of a policy thing to make it easier to identify the owners of CEs, but it isn't a security measure.

go.mod
pedjak: Investigated this. The main blocker is \`k8s.io/client-go\` — its module path uses \`v1.5.2\` in the require but the actual version is \`v0.36.x\`. The \`replace\` directive is the standard workaround for this Go module versioning quirk. Removing the replace causes \`go mod tidy\` to fail because Go tries to resolve \`v1.5.2\` which doesn't have the modern packages.

We could update the require to \`k8s.io/client-go v0.36.2\` and drop all replaces, but that's a broader dependency change that affects the whole project. I think that's better as a separate follow-up PR rather than part of this SA deprecation work. The replaces we *can* safely remove here are \`component-helpers\` and \`controller-manager\` (no longer needed since we dropped the \`k8s.io/kubernetes\` dependency), but leaving them is harmless.

api/v1/clusterextension_types.go
pedjak: Done — removed the stale sentence.

test/e2e/features/install.feature
pedjak: Restored — it was accidentally removed during the serviceAccount field cleanup. The comment explains why the deployment never becomes available (BadImage bundle contents).

docs/howto/how-to-protect-olmv1-api-access.md
pedjak: Done — restructured the intro. It now leads with "by default, only cluster administrators have permission", then discourages delegation, then presents the RBAC/VAP controls as "if you must".

docs/howto/how-to-protect-olmv1-api-access.md
pedjak: Agreed — dropped the name-prefix example.

internal/operator-controller/controllers/suite_test.go
pedjak: Done — replaced the hardcoded VAP/VAPB structs with a `test.LoadManifests(kind)` helper that parses `manifests/standard.yaml` and creates all resources matching the given kind. Same pattern as CRDs being loaded from the helm directory.

docs/project/olmv1_design_decisions.md
pedjak: Happy to rewrite this in present tense — waiting for @michaelryanpeter's guidance on the preferred framing before making the change.

michaelryanpeter: I agree with emphasizing the present. 

I like going a bit more aggressive with the reframing. WDYT of the following:

- Describe the present state of OLM v1 and the reasons behind the decisions without getting into v0 and the previous iterations of v1
- Add an admonition (maybe an INFO block?) that briefly describes the previous v1 stance.
- (optional) add an additional resources section that could include  the following links:
   - GH issues that explains the change
   - OLM v0 documentation that describes how it works
   - upstream Slack if people have questions

If questions come up in terms of the change, we could always add more documentation or add links to any ADR-like issues/docs. 

If this sounds like a good path forward, I would be happy to help shape the text.

joelanford: That sounds good to me. Thanks @michaelryanpeter!

go.mod
joelanford: I see other projects that directly `require k8s.io/client-go v0.36.1` without needing a replace, but I won't fuss about this now. We can investigate in a follow-up since this would be a no-op cleanup change.

docs/project/olmv1_design_decisions.md
```diff
@@ -176,14 +176,13 @@ OLM v1 will include primitives (e.g. templating) to make it possible to have mul
...
 
-To make OLM secure by default:
+Users who previously set `spec.serviceAccount` in their `ClusterExtension` manifests should remove that field. It is retained in the API for backwards compatibility — the API server will still accept it — but its value is silently ignored and will be removed in a future release. Because the operator-controller now always uses its own cluster-admin service account for all operations, cluster administrators should treat the ability to create or modify `ClusterExtension` resources as equivalent to granting cluster-admin privileges, and guard that access accordingly. To restrict which users may trigger installations, use Kubernetes RBAC on the `ClusterExtension` API; for finer-grained constraints such as limiting which packages, catalogs, namespaces, or upgrade policies a user may specify, pair RBAC with a `ValidatingAdmissionPolicy`. See [Protecting OLMv1 API Access](../howto/how-to-protect-olmv1-api-access.md) for concrete examples of both approaches.
```
michaelryanpeter:
```suggestion
OLMv1 adopts cluster-admin scope directly. Previously, OLM v1 required cluster administrators to create ServiceAccounts and associated RBAC to enable the installation of ClusterExtensions. The current approach provides the following benefits:
- Eliminates the complexity of difficult-to-determine permissions
-  Removes the UX burden of requiring users to manage ServiceAccounts 
- Enables simpler internal operations for namespace management and content lifecycle

NOTE:
The `spec.serviceAccount` field is depricated and will be removed in a future release. The field is retained in the API for backwards compatibility . The API server still accepts the field,  but its value is silently ignored. 

 For more information, see [Protecting OLMv1 API Access](../howto/how-to-protect-olmv1-api-access.md).
```

That might be a little too aggressive, but feel free to take what is useful and disregard what is not.

api/v1/clusterextension_types.go
```diff
@@ -402,11 +405,12 @@ type ServiceAccountReference struct {
 	//
 	// [RFC 1123]: https://tools.ietf.org/html/rfc1123
 	//
+	// +kubebuilder:validation:MinLength=1
 	// +kubebuilder:validation:MaxLength:=253
-	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="name is immutable"
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="name is immutable once set"
```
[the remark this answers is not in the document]
dtfranz: This change got dropped somehow - I'm adding it back in.
As an aside, the rule `self == null` fails validation when applying the CRD. I'm using `size(self) == 0` instead.

api/v1/clusterextension_types.go
```diff
@@ -403,10 +405,10 @@ type ServiceAccountReference struct {
 	// [RFC 1123]: https://tools.ietf.org/html/rfc1123
 	//
 	// +kubebuilder:validation:MaxLength:=253
-	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="name is immutable"
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf || size(self) == 0",message="name is immutable once set but may be cleared"
 	// +kubebuilder:validation:XValidation:rule="self.matches(\"^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$\")",message="name must be a valid DNS1123 subdomain. It must contain only lowercase alphanumeric characters, hyphens (-) or periods (.), start and end with an alphanumeric character, and be no longer than 253 characters"
-	// +required
-	Name string `json:"name"`
+	// +optional
```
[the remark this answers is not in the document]
dtfranz: The enhancement proposal covers why we're not allowing "" - see [here](https://github.com/openshift/enhancements/blob/master/enhancements/olm/service-account-deprecation.md#clusterextension-api-change). The only modification that can be made to this field is completely removing `spec.serviceAccount`.

api/v1/clusterextension_types.go
```diff
@@ -402,11 +404,12 @@ type ServiceAccountReference struct {
 	//
 	// [RFC 1123]: https://tools.ietf.org/html/rfc1123
 	//
+	// +kubebuilder:validation:MinLength:=1
 	// +kubebuilder:validation:MaxLength:=253
-	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="name is immutable"
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="name is immutable once set but may be cleared"
```
joelanford, line 409: I'm noticing that once I clear the field, I'm able to set a serviceAccount.name on it.  I'm guessing `self == oldSelf` is not catching that case because it only fires when there is an `oldSelf` to compare against.

docs/project/olmv1_design_decisions.md
```diff
@@ -176,14 +176,19 @@ OLM v1 will include primitives (e.g. templating) to make it possible to have mul
...
+OLMv1 adopts cluster-admin scope directly. Previously, OLMv1 required cluster administrators to create ServiceAccounts and associated RBAC to enable the installation of ClusterExtensions. The current approach provides the following benefits:
```
joelanford: I would suggest moving the `Previously... ` sentence into the field deprecation note.

And then I think the next sentence and bullet points are still sort of comparing/contrasting with the previous logic, which again I think is only relevant in the deprecation note (if at all).

I'll try to wordsmith more and use only present tense descriptions and motivations.

api/v1/clusterextension_types.go
joelanford: I realized the `absent -> set` case may be impossible to block since we need to keep allowing the initial create to set a service account. CEL cannot distinguish between these:
- initial create with `.spec.serviceAccount.name` set
- update from absent `.spec.serviceAccount` to `.spec.serviceAccount.name` set`

go.mod
joelanford: I was able to resolve this in 0f59083

go.mod
```diff
@@ -257,25 +256,3 @@ require (
...
-
-replace k8s.io/component-base => k8s.io/component-base v0.36.1
-
-replace k8s.io/component-helpers => k8s.io/component-helpers v0.36.1
-
-replace k8s.io/controller-manager => k8s.io/controller-manager v0.36.1
-
-replace k8s.io/kubectl => k8s.io/kubectl v0.36.1
-
-replace k8s.io/streaming => k8s.io/streaming v0.36.1
```
tmshort, line 328: Are these replaces no longer needed?

dtfranz: Yes - [this](https://github.com/operator-framework/operator-controller/pull/2770#discussion_r3553225090) copilot comment highlighted it as well.

api/v1/clusterextension_types.go
dtfranz: Couldn't we do this with another VAP? We could remove `UPDATE` from the current VAP's `operations` and add another VAP for it which gives errors instead of warnings.

api/v1/clusterextension_types.go
```diff
@@ -67,14 +66,15 @@ type ClusterExtensionSpec struct {
...
+	// OLMv1 is a single-tenant system where users with ClusterExtension write access are
+	// effectively delegated cluster-admin trust. The operator-controller runs with
+	// cluster-admin privileges and uses its own service account for all cluster interactions.
 	//
-	// +required
-	ServiceAccount ServiceAccountReference `json:"serviceAccount"`
+	// Deprecated: serviceAccount is no longer used and will be removed in a future release.
+	//
+	// +optional
+	ServiceAccount ServiceAccountReference `json:"serviceAccount,omitzero"`
```
perdasilva, line 77: **Bug:** `omitzero` is not recognized by `encoding/json` v1 — only by the upcoming `encoding/json/v2`. Since `ServiceAccountReference` is a value struct (not a pointer), a zero-value instance always serializes to `{}`, which fails the `// +kubebuilder:validation:MinProperties=1` constraint on the type.

Any client that creates or updates a `ClusterExtension` without explicitly setting `serviceAccount` will have the API server reject the request with a validation error, permanently blocking reconcile.

**Fix:** change the field to a pointer with `omitempty`:
```go
ServiceAccount *ServiceAccountReference `json:"serviceAccount,omitempty"`
```
This also matches the conventional way to represent optional/deprecated API fields in Kubernetes types.

internal/operator-controller/applier/helm.go
```diff
@@ -173,13 +150,12 @@ func (h *Helm) Apply(ctx context.Context, contentFS fs.FS, ext *ocv1.ClusterExte
...
-		return true, "", err
+	if h.TrackingCache == nil {
+		return true, "", fmt.Errorf("TrackingCache not initialized, cannot set up drift detection watches")
 	}
-
-	if err := cache.Watch(ctx, h.Watcher, relObjects...); err != nil {
+	klog.FromContext(ctx).Info("watching managed objects")
+	gvks := gvksForObjects(relObjects)
+	if err := h.TrackingCache.Watch(ctx, ext, gvks); err != nil {
 		return true, "", err
```
perdasilva: **Bug:** When `TrackingCache.Watch()` fails, `Apply()` returns `(true, "", err)`. The caller in `clusterextension_reconcile_steps.go` interprets `rolloutSucceeded=true` → sets `InstalledCondition=true`, then interprets `err != nil` → sets `ProgressingCondition=error` and requeues. The extension ends up in a permanently contradictory state (`Installed=true` + `Progressing=error`) with continuous retries.

If a Watch failure should block the install, return `(false, "", err)` so the reconciler retries cleanly without setting the Installed condition. If drift detection is non-fatal, log the error and return `(true, "", nil)` instead.

internal/operator-controller/applier/helm.go
```diff
@@ -342,73 +280,10 @@ func (p *postrenderer) Run(renderedManifests *bytes.Buffer) (*bytes.Buffer, erro
...
-				APIVersion:      ocv1.GroupVersion.Version,
-				Resource:        "clusterextensions/finalizers",
-				ResourceRequest: true,
-				Verb:            "update",
-			},
-		}
+func gvksForObjects(objs []client.Object) sets.Set[schema.GroupVersionKind] {
+	gvks := sets.New[schema.GroupVersionKind]()
+	for _, obj := range objs {
+		gvks.Insert(obj.GetObjectKind().GroupVersionKind())
```
perdasilva: **Potential bug:** `GetObjectKind().GroupVersionKind()` returns a zero-value `GroupVersionKind{}` for `Unstructured` objects whose `apiVersion`/`kind` fields were not set (e.g. a malformed Helm template that renders a YAML document without those fields). There is no guard here against inserting an empty GVK into the set.

A zero-value GVK passed to `TrackingCache.Watch` causes `ensureCacheSyncForGVK` to fail REST mapping and return an error, which propagates through `Apply()` as `(true, "", err)` — triggering the contradictory Installed+Error state described in the `line 159` comment.

**Fix:** skip zero-value GVKs before inserting, or return an error so callers can surface malformed manifests explicitly:
```go
if gvk := obj.GetObjectKind().GroupVersionKind(); !gvk.Empty() {
    gvks.Insert(gvk)
}
```

internal/operator-controller/controllers/clusterextension_reconcile_steps.go
```diff
@@ -98,16 +96,13 @@ func ValidateClusterExtension(validators ...ClusterExtensionValidator) Reconcile
 	}
 }
 
-// ServiceAccountValidator returns a validator that checks if the specified
-// ServiceAccount exists in the cluster by performing a direct Get call.
-func ServiceAccountValidator(saClient corev1client.ServiceAccountsGetter) ClusterExtensionValidator {
+// ServiceAccountDeprecationWarning returns a validator that checks if the deprecated
+// serviceAccount field is populated and logs a warning if found.
+func ServiceAccountDeprecationWarning() ClusterExtensionValidator {
```
perdasilva, line 101: **Bug:** The old `ServiceAccountValidator` returned an error when the referenced service account did not exist, surfacing a status condition to users. `ServiceAccountDeprecationWarning()` always returns `nil` — it only emits an `l.Info()` log entry.

A user who sets `spec.serviceAccount.name=nonexistent-sa` will see no error and no status condition; the extension installs with full cluster-admin privileges, silently ignoring the field.

At minimum the validator should emit a Kubernetes `Warning` response header (displayed by `kubectl` automatically) or set a status condition, rather than logging only at Info level.

internal/operator-controller/controllers/revision_engine_factory.go
```diff
@@ -40,87 +35,40 @@ type RevisionEngineFactory interface {
...
 // CreateRevisionEngine constructs a boxcutter RevisionEngine for the given ClusterObjectSet.
-// It reads the ServiceAccount from annotations and creates a scoped client.
 func (f *defaultRevisionEngineFactory) CreateRevisionEngine(_ context.Context, rev *ocv1.ClusterObjectSet) (RevisionEngine, error) {
-	saNamespace, saName, err := f.getServiceAccount(rev)
-	if err != nil {
-		return nil, err
-	}
-
-	scopedClient, err := f.createScopedClient(saNamespace, saName)
+	c, err := client.New(f.BaseConfig, client.Options{
```
perdasilva: **Performance:** `client.New(f.BaseConfig, ...)` is called inside `CreateRevisionEngine`, which is invoked once per `ClusterObjectSet` reconcile. This allocates a fresh non-caching REST client with its own connection pool and TLS stack on every reconcile cycle.

The old per-SA design justified this cost for identity isolation. Now that all engines share the same `BaseConfig`, the client can be built once.

**Fix:** call `client.New` in `NewDefaultRevisionEngineFactory`, store the result in the factory struct, and reference the pre-built client in `CreateRevisionEngine`.

test/e2e/steps/steps.go
```diff
@@ -420,18 +426,105 @@ func ResourceApplyFails(ctx context.Context, errMsg string, yamlTemplate *godog.
...
+	annotatedYAML, err := yaml.Marshal(res.Object)
+	if err != nil {
+		return fmt.Errorf("failed to marshal resource yaml: %w", err)
+	}
+	out, stdErr, err := k8scliWithInput(ctx, string(annotatedYAML), "apply", "-f", "-", "--as="+user, "--as-group="+group)
+	if err != nil {
+		return fmt.Errorf("failed to apply resource as %s/%s: %v; stderr: %s", user, group, out, stdErr)
+	}
+	sc.lastApplyStderr = stdErr
+	if res.GetKind() == "ClusterExtension" {
```
perdasilva, line 459: **Bug:** `ResourceIsAppliedAsUser` only tracks `ClusterExtension` resources in `sc.addedResources`. Compare with `ResourceIsApplied` (~line 617), which also handles `ClusterObjectSet` (setting `sc.clusterObjectSetName`) and a generic `else` branch for all other resource kinds with their namespace.

Any non-`ClusterExtension` resource applied via impersonation (e.g. a `Namespace`, `ClusterRole`, or other type used in the VAP access-control tests) will not be registered for cleanup, leaking it into subsequent test scenarios.

test/e2e/steps/steps.go
```diff
@@ -420,18 +426,105 @@ func ResourceApplyFails(ctx context.Context, errMsg string, yamlTemplate *godog.
...
+	// informer cache may not have the binding yet. Dry-run avoids persisting the resource on the first
+	// attempt, so we can safely poll until the admission chain starts denying.
+	waitFor(ctx, func() bool {
+		_, stdErr, err := k8scliWithInput(ctx, yamlContent, "apply", "--dry-run=server", "-f", "-", "--as="+user, "--as-group="+group)
+		if err == nil {
+			return false
+		}
+		return strings.Contains(stdErr, errMsg)
+	})
+	_, stdErr, err := k8scliWithInput(ctx, yamlContent, "apply", "-f", "-", "--as="+user, "--as-group="+group)
```
perdasilva, line 497: **Bug:** After the dry-run polling confirms the request is denied, this line performs a **real** `kubectl apply` (no `--dry-run`). If the VAP is momentarily inactive at write time (timing race between the dry-run poll succeeding and the real write), the apply succeeds and creates the resource. The function then returns an error (`"expected apply to fail"`), but the created resource is **never added to `sc.addedResources`**, so end-of-scenario cleanup does not delete it.

This can corrupt subsequent test scenarios if the leaked resource conflicts with setup. Consider tracking the resource before returning the error, or adding a deferred cleanup.

test/e2e/steps/steps.go
```diff
@@ -1576,6 +1665,36 @@ func ScenarioCatalogIsDeleted(ctx context.Context, catalogUserName string) error
...
+	}
+	return nil
+}
+
+// ValidatingAdmissionPolicyIsActive waits for the named ValidatingAdmissionPolicy to be compiled
+// and active by checking that status.observedGeneration matches metadata.generation.
+func ValidatingAdmissionPolicyIsActive(ctx context.Context, name string) error {
+	sc := scenarioCtx(ctx)
+	name = substituteScenarioVars(name, sc)
+	waitFor(ctx, func() bool {
```
perdasilva, line 1701: **Bug:** The return value of `waitFor()` is discarded. If the named `ValidatingAdmissionPolicy` never reaches `observedGeneration == generation` (e.g. wrong name, missing binding, or an old API server), `waitFor` polls until timeout — but the timeout/error result is silently dropped and the function returns `nil`.

All downstream steps that assume the VAP is enforcing access control then pass even though no admission is taking place, producing a false-green result.

**Fix:** capture and propagate the error:
```go
if err := waitFor(ctx, func() bool { ... }); err != nil {
    return fmt.Errorf("ValidatingAdmissionPolicy %q did not become active: %w", name, err)
}
```

perdasilva (commented):
**Finding — semantic mismatch on finalizer constant (not inline-commentable; line 60 is outside the diff hunks):**

`internal/operator-controller/controllers/clusterextension_controller.go:60`

The constant `ClusterExtensionCleanupContentManagerCacheFinalizer` still holds the string `"olm.operatorframework.io/cleanup-contentmanager-cache"`, but the finalizer handler now calls `trackingCache.Free` (Helm path) or is a no-op (Boxcutter path). Neither path involves a content manager.

This string is baked into etcd on all existing `ClusterExtension` objects. Operators debugging stuck finalizers and future maintainers will search for content-manager cleanup logic that no longer exists. Consider renaming the constant and updating the string (with a migration step for existing objects), or adding a comment explaining the historical name.

api/v1/clusterextension_types.go
dtfranz: The enhancement proposal outlines why we're not using a pointer [here](https://github.com/openshift/enhancements/blob/master/enhancements/olm/service-account-deprecation.md#clusterextension-api-change). In summary, it creates a breaking API change and goes against OpenShift API conventions.

Also, creating a ClusterExtension without `serviceAccount` is valid with these properties and does not run into the `MinProperties=1` validation from my testing (if that's what the comment is implying). You would need to explicitly set `serviceAccount: {}` in order to get that error.

pedjak: `omitzero` has been part of the standard `encoding/json` (v1) since **Go 1.24** (February 2025) — it is not exclusive to `encoding/json/v2`. This project uses Go 1.26.3, so `omitzero` is fully supported.

A zero-value `ServiceAccountReference{}` is **completely omitted** from serialized JSON — it does not produce `{}`. I verified this empirically under Go 1.26.3: marshaling a spec with `ServiceAccountReference{}` omits the `serviceAccount` field entirely, so `MinProperties=1` is never triggered.

`omitzero` is already used in 8 other struct fields in `api/v1/clusterobjectset_types.go` in this codebase. The [merged enhancement proposal](https://github.com/openshift/enhancements/blob/master/enhancements/olm/service-account-deprecation.md) explicitly chose `omitzero` over a pointer type, citing OpenShift API conventions that advise against pointer types for CRD fields.

`MinProperties=1` serves as a complementary server-side guard for raw JSON/YAML submissions (e.g. via `curl` or `kubectl`) that bypass Go client serialization — if someone sends `{"serviceAccount": {}}` directly, the CRD validation rejects it.

internal/operator-controller/controllers/clusterextension_reconcile_steps.go
pedjak: > **Bug:** The old `ServiceAccountValidator` returned an error when the referenced service account did not exist, surfacing a status condition to users. `ServiceAccountDeprecationWarning()` always returns `nil` — it only emits an `l.Info()` log entry.

`spec.serviceAccount.name` is not used anymore, and we emit such warning header right now through `ValidationAdmisionPolicy` that is a part of this PR. Not sure if there is an added value to report that the referred service account does not exist when we do not use it anyway. Let me know.

internal/operator-controller/applier/helm.go
pedjak: Fixed — Watch/TrackingCache failures in `Apply()` now log the error and return `(true, "", nil)`, matching the pattern in `reconcileExistingRelease()`. Drift detection is non-fatal; the install/upgrade succeeded.

internal/operator-controller/applier/helm.go
pedjak: Fixed — `gvksForObjects` now skips zero-value GVKs: `if gvk := obj.GetObjectKind().GroupVersionKind(); !gvk.Empty() { gvks.Insert(gvk) }`

internal/operator-controller/controllers/revision_engine_factory.go
pedjak: Fixed — `client.New` is now called once in `NewDefaultRevisionEngineFactory` and the pre-built client is stored in the factory struct. `CreateRevisionEngine` reuses it.

test/e2e/steps/steps.go
pedjak: Fixed — `ResourceIsAppliedAsUser` now has the same `else` branch as `ResourceIsApplied` to track non-CE resources (ClusterRole, CRB, VAP, VAPB, etc.) for cleanup.

test/e2e/steps/steps.go
pedjak: Fixed — if the real apply unexpectedly succeeds, the resource is now tracked in `sc.addedResources` before returning the error.

test/e2e/steps/steps.go
pedjak: `waitFor` uses `require.Eventually` which fails the scenario on timeout — the error is not silently swallowed. Added a clarifying comment to the function doc.

pedjak:
> Due to the sensitivity of this change (we _are_ removing some behaviors that users may be expecting), let's make sure we have an approved design doc in place before this actually merges.

/hold cancel

https://github.com/openshift/enhancements/pull/2054 got merged recently.

rashmigottipati (approved):
/lgtm

PR #2771, opened 2026-06-16
🌱 consolidate Release types with operator-registry

rashmigottipati:
# Description

  Removes duplicate `bundle.Release` and `bundle.VersionRelease` type definitions in favor of using `declcfg.Release` and `declcfg.VersionRelease` from operator-registry.

 This consolidates type definitions between operator-controller and operator-registry, eliminating code duplication and aligning with the vision from OPRUN-4548.

  **Changes:**
  - Delete `internal/operator-controller/bundle/versionrelease.go`  and the test file 
  - Update all internal code to use `declcfg.VersionRelease` and `declcfg.Release` from operator-registry
  - API boundary (`BundleMetadata.Release`) remains `*string` with conversion at:
    - `bundleutil.MetadataFor` (declcfg -> API)
    - `catalogmetadata/filter.parseInstalledBundleVersionRelease` (API -> declcfg)
    
## Reviewer Checklist

- [x] API Go Documentation


internal/operator-controller/bundleutil/bundle.go
```diff
@@ -50,43 +93,39 @@ func parseVersionRelease(pkgData json.RawMessage) (*bundle.VersionRelease, error
...
-		return nil, err
-	}
-	return vr, nil
+	return newLegacyRegistryV1VersionRelease(pkg.Version)
 }
 
 // parseExplicitRelease parses version and release from separate fields.
 // Build metadata is preserved in the version because with an explicit release field,
 // build metadata serves its proper semver purpose (e.g., git commit, build number).
 // In contrast, NewLegacyRegistryV1VersionRelease clears build metadata because it
```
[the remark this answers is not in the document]
perdasilva: might be worth applying this suggestion

manifests/experimental.yaml
```diff
@@ -578,7 +578,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.20.1
+    controller-gen.kubebuilder.io/version: v0.21.0
```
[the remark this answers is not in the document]
perdasilva: is this worth investigating?

internal/operator-controller/bundleutil/bundle.go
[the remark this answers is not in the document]
rashmigottipati: yup, fixed!

manifests/experimental.yaml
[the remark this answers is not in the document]
rashmigottipati: Fixed in the latest commit - all CRDs now use v0.20.1 consistently

internal/operator-controller/catalogmetadata/filter/successors.go
```diff
@@ -35,24 +30,53 @@ func parseInstalledBundleVersionRelease(installedBundle ocv1.BundleMetadata) (*b
...
+	return &declcfg.VersionRelease{
 		Version: version,
 		Release: release,
 	}, nil
 }
 
+// newLegacyRegistryV1VersionRelease parses a registry+v1 bundle version string and returns a
+// VersionRelease. Some registry+v1 bundles utilize the build metadata field of the semver version
+// as release information (a semver spec violation maintained for backward compatibility).
+func newLegacyRegistryV1VersionRelease(vStr string) (*declcfg.VersionRelease, error) {
```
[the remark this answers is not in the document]
rashmigottipati: agreed, although duplication existed in the original code. will create follow-up PR to extract this to a shared utility function. this PR focusses on type consolidation.

internal/operator-controller/catalogmetadata/filter/successors.go
```diff
@@ -35,24 +30,53 @@ func parseInstalledBundleVersionRelease(installedBundle ocv1.BundleMetadata) (*b
...
+	vers, err := bsemver.Parse(vStr)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get version and release of installed bundle: %w", err)
```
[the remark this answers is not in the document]
rashmigottipati: addressed in https://github.com/operator-framework/operator-controller/pull/2771#discussion_r3428471737

internal/operator-controller/bundleutil/bundle.go
```diff
@@ -23,7 +22,51 @@ func GetVersionAndRelease(b declcfg.Bundle) (*bundle.VersionRelease, error) {
...
-func parseVersionRelease(pkgData json.RawMessage) (*bundle.VersionRelease, error) {
+// newLegacyRegistryV1VersionRelease parses a registry+v1 bundle version string and returns a
+// VersionRelease. Some registry+v1 bundles utilize the build metadata field of the semver version
+// as release information (a semver spec violation maintained for backward compatibility). When the
+// bundle version includes build metadata that is parsable as a release, the returned
+// VersionRelease has the build metadata extracted into the Release field, and the Version field
+// has its Build metadata cleared. When the bundle version includes build metadata that is NOT
+// parseable as a release, the returned VersionRelease has its Version set to the full semver
+// version (with build metadata) and its Release left empty.
+func newLegacyRegistryV1VersionRelease(vStr string) (*declcfg.VersionRelease, error) {
```
tmshort: This function seems duplicated in internal/operator-controller/catalogmetadata/filter/successors.go

Is there a way to consolidate this function into a single location?

internal/operator-controller/catalogmetadata/compare/compare.go
```diff
@@ -51,7 +51,7 @@ func ByVersionAndRelease(b1, b2 declcfg.Bundle) int {
 	if err1 != nil || err2 != nil {
 		return compareErrors(err2, err1)
 	}
-	return vr2.Compare(*vr1)
+	return vr2.Compare(vr1)
```
tmshort, line 54: Why is this changing? Was it wrong before?

tmshort (commented):
Most of this is changing the namespace of an API, but there is some de-duplication that could be done.

internal/operator-controller/catalogmetadata/filter/successors.go
pedjak: **suggestion (non-blocking):** This is identical to the `newLegacyRegistryV1VersionRelease` added in `bundleutil/bundle.go`. Since the `filter` package already depends on `bundleutil` (via `bundle_predicates.go`), this could be deduplicated now rather than in a follow-up — the duplication is introduced by this PR, so cleaning it up here feels natural.

Concretely:
1. Export the function from `bundleutil` (e.g. as `ParseLegacyVersionRelease` — the current name is quite long; `LegacyVersionRelease` would also work)
2. Call `bundleutil.ParseLegacyVersionRelease` here
3. The error wrapping difference (`"failed to get version and release of installed bundle: %w"`) can stay at the call site in `parseInstalledBundleVersionRelease`

That's ~30 lines deleted with no structural risk.

grokspawn: +1, and I'd consider this a blocker since this PR introduces the duplication.

internal/operator-controller/bundleutil/bundle.go
```diff
@@ -23,7 +22,51 @@ func GetVersionAndRelease(b declcfg.Bundle) (*bundle.VersionRelease, error) {
...
+		Version: vers,
+	}
+
+	buildMetadata := ""
+	if len(vr.Version.Build) > 0 {
+		buildMetadata = vr.Version.Build[0]
+		for i := 1; i < len(vr.Version.Build); i++ {
+			buildMetadata += "." + vr.Version.Build[i]
+		}
+	}
```
grokspawn: This is logically
```suggestion
buildMetadata = strings.Join(vr.Version.Build, ".")
```

internal/operator-controller/catalogmetadata/filter/successors.go
```diff
@@ -35,24 +30,53 @@ func parseInstalledBundleVersionRelease(installedBundle ocv1.BundleMetadata) (*b
...
+		Version: vers,
+	}
+
+	buildMetadata := ""
+	if len(vr.Version.Build) > 0 {
+		buildMetadata = vr.Version.Build[0]
+		for i := 1; i < len(vr.Version.Build); i++ {
+			buildMetadata += "." + vr.Version.Build[i]
+		}
+	}
```
grokspawn:
```suggestion
	buildMetadata := strings.Join(vr.Version.Build, ".")
```

internal/operator-controller/catalogmetadata/compare/compare.go
grokspawn: It's now using the upstream [Compare](https://github.com/operator-framework/operator-registry/blob/48f73d6cf0108cc54a72612b34f87704ab804a53/alpha/model/versionrelease.go#L135) function, which takes a pointer instead of an instance.

internal/operator-controller/catalogmetadata/filter/successors.go
rashmigottipati: Fixed in latest commit

internal/operator-controller/catalogmetadata/filter/successors.go
rashmigottipati: This code was removed as part of the deduplication: the function now only exists in `bundleutil.ParseLegacyVersionRelease` where I've already applied your `strings.Join` suggestion. The duplicate in successors.go was deleted in this same commit.

internal/operator-controller/controllers/clusterextension_controller_test.go
```diff
@@ -775,11 +774,13 @@ func TestClusterExtensionBoxcutterApplierFailsDoesNotLeakDeprecationErrors(t *te
...
-			RollingOut: []*controllers.RevisionMetadata{{}},
-		}, nil)
-		d.Resolver = resolve.Func(func(ctx context.Context, ext *ocv1.ClusterExtension, installedBundle *ocv1.BundleMetadata) (*declcfg.Bundle, *bundle.VersionRelease, *declcfg.Deprecation, error) {
-			v := bundle.VersionRelease{
+		d.RevisionStatesGetter = &MockRevisionStatesGetter{
+			RevisionStates: &controllers.RevisionStates{
+				RollingOut: []*controllers.RevisionMetadata{{}},
+			},
+		}
+		d.Resolver = resolve.Func(func(ctx context.Context, ext *ocv1.ClusterExtension, installedBundle *ocv1.BundleMetadata) (*declcfg.Bundle, *declcfg.VersionRelease, *declcfg.Deprecation, error) {
```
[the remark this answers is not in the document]
rashmigottipati: outdated comment

internal/operator-controller/bundleutil/bundle.go
rashmigottipati: Fixed in latest commit

internal/operator-controller/bundleutil/bundle.go
```diff
@@ -23,7 +23,45 @@ func GetVersionAndRelease(b declcfg.Bundle) (*bundle.VersionRelease, error) {
...
+		return nil, err
+	}
+
+	vr := &declcfg.VersionRelease{
+		Version: vers,
+	}
+
+	buildMetadata := strings.Join(vr.Version.Build, ".")
+
+	rel, err := declcfg.NewRelease(buildMetadata)
```
perdasilva, line 47: `NewRelease` here introduces a length limit on buildMetadata (https://github.com/operator-framework/operator-registry/blob/master/alpha/model/versionrelease.go#L84) - this is technically a breaking change since we didn't have a limit before. Are we concerned about this at all? Could it break existing bundles? Should we call this out in docs somewhere or mark this PR as a breaking change? (cc @grokspawn)

grokspawn: All available catalogs were assessed before establishing the limit, and the limit is sufficient with some breathing room (https://github.com/operator-framework/operator-registry/pull/1852). 

IMHO it was an oops on our part to allow an unbounded construct in the first place (pointing finger at myself).  

It's not _impossible_ that there is an unknown case where someone has become dependent on using this in a way which would violate the length rule, but it would have to have been in the last six months, so I don't think I'm concerned here.  Happy to discuss further, in case there are specific cases/conditions of concern.

grokspawn: ... and great question.  I failed to show all my work here I guess.

rashmigottipati: @grokspawn thanks for clarifying! Good to know all catalogs were already assessed for this. 

@perdasilva can we proceed with this PR since this concern has been addressed?

perdasilva, line 47: `declcfg.NewRelease` enforces a 20-character max length (`versionReleaseMaxLength = 20` in operator-registry) that the old `bundle.NewRelease` did not have.

In `ParseLegacyVersionRelease`, the error is silently swallowed by the `if err == nil && len(rel) > 0` guard, so a bundle with build metadata > 20 chars (e.g. `1.0.0+build.2024.01.15.1234`) would change from `{Version:1.0.0, Release:[build,2024,01,15,1234]}` to `{Version:1.0.0+build.2024.01.15.1234, Release:nil}`. This affects sorting and `ExactVersionRelease` matching.

In practice unlikely since build-metadata-as-release typically uses short numeric values, but worth acknowledging as a behavioral change.

internal/operator-controller/catalogmetadata/filter/successors.go
```diff
@@ -35,19 +35,19 @@ func parseInstalledBundleVersionRelease(installedBundle ocv1.BundleMetadata) (*b
...
-	var release bundle.Release
+	var release declcfg.Release
 	if *installedBundle.Release == "" {
 		// Explicit empty release: use empty slice (not nil) to match catalog parsing behavior.
 		// NewRelease("") returns nil, but we need empty slice for roundtrip correctness.
-		release = bundle.Release([]bsemver.PRVersion{})
+		release = declcfg.Release([]bsemver.PRVersion{})
 	} else {
-		release, err = bundle.NewRelease(*installedBundle.Release)
+		release, err = declcfg.NewRelease(*installedBundle.Release)
```
perdasilva: Same 20-char limit from `declcfg.NewRelease` applies here. If `BundleReleaseSupport` was enabled and a `BundleMetadata` was persisted with a `Release` field > 20 chars, this will return a **hard error** (unlike the legacy path which silently falls back), causing `SuccessorsOf` to fail and blocking resolution for that `ClusterExtension`.

internal/operator-controller/catalogmetadata/filter/successors.go
```diff
@@ -35,19 +35,19 @@ func parseInstalledBundleVersionRelease(installedBundle ocv1.BundleMetadata) (*b
 		return nil, fmt.Errorf("failed to parse installed bundle version %q: %w", installedBundle.Version, err)
 	}
 
-	var release bundle.Release
+	var release declcfg.Release
```
perdasilva: This block (lines 33-51) is a semantic duplicate of `parseExplicitRelease` in `bundleutil/bundle.go` (lines 99-120): both parse version, branch on empty-vs-nonempty release, construct `VersionRelease`.

Consider exporting `parseExplicitRelease` as `ParseExplicitRelease` and calling it from here to avoid drift if either copy changes.

internal/operator-controller/bundleutil/bundle.go
```diff
@@ -104,6 +138,27 @@ func MetadataFor(bundleName string, vr bundle.VersionRelease) ocv1.BundleMetadat
...
+func asLegacyRegistryV1Version(vr declcfg.VersionRelease) bsemver.Version {
+	v := bsemver.Version{
+		Major: vr.Version.Major,
+		Minor: vr.Version.Minor,
+		Patch: vr.Version.Patch,
+		Pre:   vr.Version.Pre,
+		Build: vr.Version.Build,
+	}
+	if len(vr.Release) > 0 {
+		v.Build = make([]string, len(vr.Release))
```
perdasilva: nit: This manual loop duplicates the `slicesutil.Map` pattern already used in `compare/compare.go:35` for the identical `PRVersion`→`string` conversion. The old deleted code also used `slicesutil.Map` here.

perdasilva (commented):
## Code Review: Consolidate Release types with operator-registry

4 inline comments posted + 3 additional findings below.

---

### Double JSON unmarshal in `parseVersionRelease` (`bundleutil/bundle.go:66-76`)
`pkgData` is unmarshalled twice: once into `property.Package` (line 66) and again into a helper struct (line 76) just to detect `Release *string` presence. Only `pkg.Version` is actually used from `property.Package`. A single struct `struct{ Version string \`json:\"version\"\`; Release *string \`json:\"release\"\` }` would serve both purposes in one unmarshal pass.

### `SuccessorsOf` double-parses installed bundle version (`successors.go:57-75`)
`parseInstalledBundleVersionRelease` (line 57) parses `installedBundle.Version` into a `bsemver.Version`, then `legacySuccessor` (line 75) parses the exact same string again via `bsemver.Parse`. The already-parsed version could be passed to `legacySuccessor` instead.

### 33 comparison test cases deleted without equivalent upstream coverage
The deleted `bundle/versionrelease_test.go` contained `TestRelease_Compare` (17 cases) and `TestVersionRelease_Compare` (16 cases). operator-registry v1.72.0's `versionrelease_test.go` has **zero** dedicated `Compare` tests — only a single equality assertion inside a round-trip test. `ParseLegacyVersionRelease` also has no direct unit tests (only indirect coverage through `GetVersionAndRelease`). Consider either adding comparison tests upstream or keeping a subset here.

internal/operator-controller/bundleutil/bundle.go
perdasilva: I think we're good with this one looking at the previous comments - I'll resolve

internal/operator-controller/catalogmetadata/filter/successors.go
rashmigottipati: done - exported parseExplicitRelease as ParseExplicitRelease and called it from parseInstalledBundleVersionRelease to eliminate the duplication. Fixed in [latest commit](https://github.com/operator-framework/operator-controller/pull/2771/changes/74c294ce9f7c915aa85f911637946fb82492e007)

internal/operator-controller/bundleutil/bundle.go
rashmigottipati: fixed - replaced the manual loop with slicesutil.Map. addressed in [latest commit](https://github.com/operator-framework/operator-controller/pull/2771/changes/74c294ce9f7c915aa85f911637946fb82492e007).

PR #2779, opened 2026-06-24
:bug: fix catalogd panic, paste error on mac

grokspawn:
# Description
This PR fixes two issues, from least- to most-important: 
1. differing syntax for `paste` tool in linux, mac contexts.  `paste` on BSD requires an explicit indication of STDIN, whereas linux assumes STDIN if no file is specified.  POSIX behavior is to specify `-` if STDIN is indicated, which works for both platforms. 
2. `make run` and variants use the "linux" build targets, which claim in docs to use native GOARCH but actually hardcode it to amd64.  On Apple silicon the cross-compile now results in panics running these cross-compiled binaries using Rosetta (and Apple has stated it is sunsetting support for amd64 binaries on arm64 hardware, so it needs resolving anyway).  The GOARCH hardcoding was introduced during the change to helm templating (from kustomize) but appears to have been unnecessary.  

## Reviewer Checklist

- [ ] API Go Documentation


pedjak (commented):
/lgtm

rashmigottipati (approved):

PR #2782, opened 2026-06-24
🌱 move BuildHTTPClient from shared httputil to catalogmetadata/client

joelanford:
# Description

Move `BuildHTTPClient` from `internal/shared/util/http` into
`internal/operator-controller/catalogmetadata/client`, since it is only
used by the catalog client. This reduces the surface of the shared
httputil package and co-locates the HTTP client construction with its
sole consumer.

- Rename the package import from `httputil` to `catalogclient` in
  `main.go` and tests
- Move `httputil.go` and `httputil_test.go` into the `client` package
- Update test helpers to use the new import path

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/internal/operator-controller/catalogmetadata/client/httputil.go
+++ b/internal/operator-controller/catalogmetadata/client/httputil.go
@@ -1,4 +1,4 @@
-package http
+package client
 
 import (
 	"crypto/tls"
@@ -6,10 +6,11 @@ import (
 	"net/http"
 	"time"
 
+	httputil "github.com/operator-framework/operator-controller/internal/shared/util/http"
 	"github.com/operator-framework/operator-controller/internal/shared/util/tlsprofiles"
 )
 
-func BuildHTTPClient(cpw *CertPoolWatcher) (*http.Client, error) {
+func BuildHTTPClient(cpw *httputil.CertPoolWatcher) (*http.Client, error) {
 	httpClient := &http.Client{Timeout: 10 * time.Second}
 
 	pool, _, err := cpw.Get()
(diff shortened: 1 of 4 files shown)


internal/operator-controller/catalogmetadata/client/httputil.go
```diff
@@ -1,16 +1,17 @@
...
 	"time"
 
+	httputil "github.com/operator-framework/operator-controller/internal/shared/util/http"
 	"github.com/operator-framework/operator-controller/internal/shared/util/tlsprofiles"
 )
 
-func BuildHTTPClient(cpw *CertPoolWatcher) (*http.Client, error) {
-	httpClient := &http.Client{Timeout: 10 * time.Second}
+func BuildHTTPClient(cpw *httputil.CertPoolWatcher) (*http.Client, error) {
+	httpClient := &http.Client{Timeout: 5 * time.Minute}
```
pedjak: can we extract it into a constant?

pedjak (commented):
I would focus this PR solely on the timeout increase, and rest of refactoring would push into another PR.

joelanford:
Updated this PR to be just the refactoring. I'll make a follow-up for bumping the timeout.

internal/operator-controller/catalogmetadata/client/httputil.go
joelanford: I'll split it up.

On the constant, I don't know that we should have one constant shared between the client and server. The server's timeout needs to consider all of its potential client use cases more broadly than just the op-con client. The reason they were different before is that catd needed a longer timeout to accommodate off-cluster clients, which could theoretically be on high-latency/low-bandwidth connections).

We _could_ bump the op-con client timeout to something between 10s and 5m (e.g. like 2m), but that seems arbitrary, and 5m doesn't seem excessively long for op-con anyway. If we bump catalogd up again later (e.g. to 20m), I don't think we'd automatically want to bump op-con client to that as well, would we?

I could extract separate constants for the existing values, but 🤷‍♂️ , maybe not necessary if there's only one place they're used. Not opinionated either way on this.

PR #2783, opened 2026-06-25
:seedling: Add e2e regression test for cross-CE collision protection

perdasilva:
## Summary

- Add e2e regression test verifying that a second ClusterExtension targeting an already-installed bundle cannot take over resources from the original owner, even with a higher-revision ClusterObjectSet. The single merged scenario covers both initial collision detection after an upgrade and collision persistence across revisions.
- Use `${NAME}` / `${NAME}-dup` naming convention in the collision scenario so CE names stay abstract. Stop overwriting `sc.clusterExtensionName` in `ResourceIsApplied` so `${NAME}` remains constant throughout a scenario — the overwrite was always a no-op for existing tests since they all use `name: ${NAME}`.
- Auto-track every applied ClusterExtension for cleanup via `addedResources` in `ResourceIsApplied`, eliminating the manual `TrackCurrentClusterExtensionForCleanup` step and the redundant `sc.clusterExtensionName` cleanup in `ScenarioCleanup`.
- Add named CE condition steps (`NamedClusterExtensionReportsCondition`, `NamedClusterExtensionReportsConditionWithMessageFragment`) so step descriptions explicitly identify which CE is being asserted.
- Add `ClusterExtensionOwnsClusterObjectSets` step to verify the number of ClusterObjectSets owned by a named CE.
- Extract `messageFragmentComparison` helper to deduplicate the fragment-matching closure across condition steps.
- Preserve unresolved template variables as literal `${KEY}` instead of silently substituting empty string.

**Note:** ClusterObjectSets still use the legacy `sc.clusterObjectSetName` cleanup path rather than `addedResources` tracking. The same consolidation can be done in a follow-up.

## Test plan

- [x] `go vet` passes
- [x] New scenario passes against a kind cluster with BoxcutterRuntime and DeploymentConfig enabled
- [ ] CI e2e suite passes

### DIFF
--- a/test/e2e/features/update.feature
+++ b/test/e2e/features/update.feature
@@ -225,6 +225,7 @@ Feature: Update ClusterExtension
     And bundle "${PACKAGE:test}.1.0.4" is installed in version "1.0.4"
 
   @BoxcutterRuntime
+  @DeploymentConfig
   Scenario: Detect collision when a second ClusterExtension installs the same package after an upgrade
     Given ClusterExtension is applied
       """
@@ -251,7 +252,6 @@ Feature: Update ClusterExtension
     Then ClusterExtension is rolled out
     And ClusterExtension is available
     And bundle "${PACKAGE:test}.1.0.1" is installed in version "1.0.1"
-    And the current ClusterExtension is tracked for cleanup
     When ClusterExtension is applied
       """
       apiVersion: olm.operatorframework.io/v1
@@ -271,10 +271,44 @@ Feature: Update ClusterExtension
                 "olm.operatorframework.io/metadata.name": ${CATALOG:test}
             version: 1.0.1
       """
-    Then ClusterExtension reports Progressing as True with Reason Retrying and Message includes:
+    Then ClusterExtension "${NAME}-dup" reports Progressing as True with Reason Retrying and Message includes:
       """
       revision object collisions
       """
+    And ClusterExtension "${NAME}" reports Installed as True
+    # Force a second revision on the dup via env var change — collision must persist
+    When ClusterExtension is updated
+      """
+      apiVersion: olm.operatorframework.io/v1
+      kind: ClusterExtension
+      metadata:
+        name: ${NAME}-dup
+      spec:
+        namespace: ${TEST_NAMESPACE}
+        serviceAccount:
+          name: olm-sa
+        config:
+          configType: Inline
+          inline:
+            deploymentConfig:
+              env:
+                - name: MY_VAR
+                  value: "my-value"
+        source:
+          sourceType: Catalog
+          catalog:
+            packageName: ${PACKAGE:test}
+            selector:
+              matchLabels:
+                "olm.operatorframework.io/metadata.name": ${CATALOG:test}
+            version: 1.0.1
+      """
+    Then ClusterExtension "${NAME}-dup" owns 2 ClusterObjectSets
+    And ClusterExtension "${NAME}-dup" reports Progressing as True with Reason Retrying and Message includes:
+      """
+      revision object collisions
+      """
+    And ClusterExtension "${NAME}" reports Installed as True
 
   @BoxcutterRuntime
   Scenario: Each update creates a new revision and resources not present in the new revision are removed from the cluster
--- a/test/e2e/steps/hooks.go
+++ b/test/e2e/steps/hooks.go
@@ -268,9 +268,6 @@ func ScenarioCleanup(ctx context.Context, _ *godog.Scenario, err error) (context
 	}
 
 	forDeletion := sc.addedResources
-	if sc.clusterExtensionName != "" {
-		forDeletion = append(forDeletion, resource{name: sc.clusterExtensionName, kind: "clusterextension"})
-	}
 	if sc.clusterObjectSetName != "" && featureGates[features.BoxcutterRuntime] {
 		forDeletion = append(forDeletion, resource{name: sc.clusterObjectSetName, kind: "clusterobjectset"})
 	}
--- a/test/e2e/steps/steps.go
+++ b/test/e2e/steps/steps.go
@@ -182,7 +182,9 @@ func RegisterSteps(sc *godog.ScenarioContext) {
 
 	sc.Step(`^(?i)min value for (ClusterExtension|ClusterObjectSet) ((?:\.[a-zA-Z]+)+) is set to (\d+)$`, SetCRDFieldMinValue)
 
-	sc.Step(`^(?i)the current ClusterExtension is tracked for cleanup$`, TrackCurrentClusterExtensionForCleanup)
+	sc.Step(`^(?i)ClusterExtension "([^"]+)" owns (\d+) ClusterObjectSets?$`, ClusterExtensionOwnsClusterObjectSets)
+	sc.Step(`^(?i)ClusterExtension "([^"]+)" reports ([[:alnum:]]+) as ([[:alnum:]]+)$`, NamedClusterExtensionReportsCondition)
+	sc.Step(`^(?i)ClusterExtension "([^"]+)" reports ([[:alnum:]]+) as ([[:alnum:]]+) with Reason ([[:alnum:]]+) and Message includes:$`, NamedClusterExtensionReportsConditionWithMessageFragment)
 
 	// TLS profile enforcement steps — deployment configuration
 	sc.Step(`^(?i)the "([^"]+)" deployment is configured with custom TLS minimum version "([^"]+)"$`, ConfigureDeploymentWithCustomTLSVersion)
@@ -382,15 +384,37 @@ func ResourceApplyFails(ctx context.Context, errMsg string, yamlTemplate *godog.
 	return nil
 }
 
-// TrackCurrentClusterExtensionForCleanup saves the current ClusterExtension name in the cleanup list
-// so it gets deleted at the end of the scenario. Call this before applying a second ClusterExtension
-// in the same scenario, because ResourceIsApplied overwrites the tracked name.
-func TrackCurrentClusterExtensionForCleanup(ctx context.Context) error {
+// ClusterExtensionOwnsClusterObjectSets waits for the named ClusterExtension to own exactly the
+// expected number of ClusterObjectSets. Polls with timeout.
+func ClusterExtensionOwnsClusterObjectSets(ctx context.Context, extName string, expectedCount int) error {
 	sc := scenarioCtx(ctx)
-	if sc.clusterExtensionName != "" {
-		sc.addedResources = append(sc.addedResources, resource{name: sc.clusterExtensionName, kind: "clusterextension"})
-	}
+	extName = substituteScenarioVars(extName, sc)
+	waitFor(ctx, func() bool {
+		out, err := k8sClient("get", "clusterobjectsets",
+			"-l", fmt.Sprintf("olm.operatorframework.io/owner-name=%s", extName),
+			"-o", "jsonpath={.items[*].metadata.name}")
+		if err != nil {
+			return false
+		}
+		names := strings.Fields(strings.TrimSpace(out))
+		return len(names) == expectedCount
+	})
 	return nil
+}
+
+// NamedClusterExtensionReportsCondition waits for a specific ClusterExtension (by name) to have a condition
+// matching type and status. Polls with timeout.
+func NamedClusterExtensionReportsCondition(ctx context.Context, extName, conditionType, conditionStatus string) error {
+	sc := scenarioCtx(ctx)
+	extName = substituteScenarioVars(extName, sc)
+	return waitForCondition(ctx, "clusterextension", extName, conditionType, conditionStatus, nil, nil)
+}
+
+// NamedClusterExtensionReportsConditionWithMessageFragment waits for a specific ClusterExtension (by name)
+// to have a condition matching type, status, reason, with a message containing the specified fragment.
+func NamedClusterExtensionReportsConditionWithMessageFragment(ctx context.Context, extName, conditionType, conditionStatus, conditionReason string, msgFragment *godog.DocString) error {
+	extName = substituteScenarioVars(extName, scenarioCtx(ctx))
+	return waitForCondition(ctx, "clusterextension", extName, conditionType, conditionStatus, &conditionReason, messageFragmentComparison(ctx, msgFragment))
 }
 
 // ClusterExtensionVersionUpdate patches the ClusterExtension's catalog version to the specified value.
@@ -449,7 +473,7 @@ func ResourceIsApplied(ctx context.Context, yamlTemplate *godog.DocString) error
 		return fmt.Errorf("failed to apply resource %v; err: %w; stderr: %s", out, err, stderrOutput(err))
 	}
 	if res.GetKind() == "ClusterExtension" {
-		sc.clusterExtensionName = res.GetName()
+		sc.addedResources = append(sc.addedResources, resource{name: res.GetName(), kind: "clusterextension"})
 	} else if res.GetKind() == "ClusterObjectSet" {
 		sc.clusterObjectSetName = res.GetName()
 	} else {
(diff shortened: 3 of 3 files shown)


test/e2e/steps/steps.go
```diff
@@ -348,6 +351,7 @@ func substituteScenarioVars(content string, sc *scenarioContext) string {
 	vars := map[string]string{
 		"TEST_NAMESPACE": sc.namespace,
 		"NAME":           sc.clusterExtensionName,
+		"PREV_NAME":      sc.previousClusterExtensionName,
 		"COS_NAME":       sc.clusterObjectSetName,
```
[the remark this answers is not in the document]
perdasilva: Good catch — `PREV_NAME` was a known key with an empty value, so the `templateContent` literal-preservation fix didn't cover it. Moved it out of the static map: now only added when `previousClusterExtensionName` is non-empty.

test/e2e/features/update.feature
```diff
@@ -338,3 +338,93 @@ Feature: Update ClusterExtension
...
+            packageName: ${PACKAGE:test}
+            version: 1.0.0
+            selector:
+              matchLabels:
+                "olm.operatorframework.io/metadata.name": ${CATALOG:test}
+      """
+    And ClusterExtension is rolled out
+    And ClusterExtension is available
+    And ClusterExtension "${NAME}" owns 1 ClusterObjectSet
+    And the current ClusterExtension is tracked for cleanup
```
pedjak: I would perhaps replace this with:

```
ClusterExtension "${NAME}"  is tracked for cleanup
```

btw, I am not sure what does it mean from user perspective and if users need to know anything about it? Is that an implementation detail?

test/e2e/features/update.feature
```diff
@@ -338,3 +338,93 @@ Feature: Update ClusterExtension
...
+        source:
+          sourceType: Catalog
+          catalog:
+            packageName: ${PACKAGE:test}
+            version: 1.0.0
+            selector:
+              matchLabels:
+                "olm.operatorframework.io/metadata.name": ${CATALOG:test}
+      """
+    Then ClusterExtension "${NAME}" reports Progressing as True with Reason Retrying and Message includes:
```
pedjak: I assume `${NAME}-dup` should report Progressing, not the already installed CE `${NAME}`?

test/e2e/features/update.feature
```diff
@@ -338,3 +338,93 @@ Feature: Update ClusterExtension
...
+      """
+      revision object collisions
```
pedjak: can we make this message clearer to users, as we discussed in the original PR?

test/e2e/features/update.feature
```diff
@@ -338,3 +338,93 @@ Feature: Update ClusterExtension
...
+      """
+    And ClusterExtension "${PREV_NAME}" reports Installed as True
```
pedjak: it should be actually 

```
ClusterExtension "${NAME}" reports Installed as True
```

because this is the one first installed/created in this scenario?

test/e2e/features/update.feature
```diff
@@ -338,3 +338,93 @@ Feature: Update ClusterExtension
...
+    # Update the conflicting ClusterExtension with a deployment config env var. At the time
```
pedjak: are we updating here `${NAME}` or `${NAME}-dup`?

test/e2e/features/update.feature
```diff
@@ -338,3 +338,93 @@ Feature: Update ClusterExtension
...
+          sourceType: Catalog
+          catalog:
+            packageName: ${PACKAGE:test}
+            version: 1.0.0
+            selector:
+              matchLabels:
+                "olm.operatorframework.io/metadata.name": ${CATALOG:test}
+      """
+    Then ClusterExtension "${NAME}" owns 2 ClusterObjectSets
+    And ClusterExtension "${NAME}" reports Progressing as True with Reason Retrying and Message includes:
```
pedjak: should be `${NAME}-dup`?

test/e2e/features/update.feature
```diff
@@ -338,3 +338,93 @@ Feature: Update ClusterExtension
...
+      """
+      revision object collisions
+      """
+    And ClusterExtension "${PREV_NAME}" reports Installed as True
```
pedjak: how do we know that 2nd ClusterObjectSet is installed successfully?

test/e2e/features/update.feature
perdasilva: Removed this step entirely. `ResourceIsApplied` now auto-tracks every applied ClusterExtension for cleanup via `addedResources`, so there's no need for a manual tracking step. The `previousClusterExtensionName` / `${PREV_NAME}` machinery has also been removed.

test/e2e/features/update.feature
perdasilva: Fixed. Both collision scenarios now use explicit names — `ce-orig-${SCENARIO_ID}` for the original CE and `ce-dup-${SCENARIO_ID}` for the duplicate — so every assertion unambiguously identifies which CE it checks. No more implicit `${NAME}` context switching.

test/e2e/features/update.feature
perdasilva: The test asserts a substring match against what the controller actually emits — `revision object collisions` is the literal message from the ClusterObjectSet reconciler (`setRetryingConditions` in `clusterobjectset_controller.go`). Changing the wording would be a controller code change rather than a test change. Happy to open a follow-up to improve the message if you have a suggestion for clearer wording.

test/e2e/features/update.feature
perdasilva: Fixed — now explicitly `ClusterExtension "ce-orig-${SCENARIO_ID}" reports Installed as True`. Both CEs are referenced by their explicit names throughout the scenario.

test/e2e/features/update.feature
perdasilva: Fixed — the YAML now explicitly says `name: ce-dup-${SCENARIO_ID}` and the comment reads "Update the duplicate ClusterExtension with a deployment config env var to force a second revision."

test/e2e/features/update.feature
perdasilva: Fixed — now explicitly `ClusterExtension "ce-dup-${SCENARIO_ID}"` throughout.

test/e2e/features/update.feature
perdasilva: The test intentionally verifies that the duplicate CE *cannot* install — both its revisions collide with the original CE's objects. The assertions confirm this:

- `ce-dup` owns 2 ClusterObjectSets → revision 2 was created
- `ce-dup` reports Progressing/Retrying with "revision object collisions" → both revisions are blocked
- `ce-orig` reports Installed=True → the original CE is unaffected

I initially considered using the progression deadline timeout to verify the dup CE settles into a terminal failure state, but it started adding too much complexity and wait time for what is fundamentally a regression test. Would there be a better signal we could use here? For example, would asserting the dup stays in Retrying for 30s–60s be sufficient, or is there a cleaner way to verify the collision is persistent?

test/e2e/steps/steps.go
```diff
@@ -449,6 +473,9 @@ func ResourceIsApplied(ctx context.Context, yamlTemplate *godog.DocString) error
 		return fmt.Errorf("failed to apply resource %v; err: %w; stderr: %s", out, err, stderrOutput(err))
 	}
 	if res.GetKind() == "ClusterExtension" {
+		if sc.clusterExtensionName != "" && sc.clusterExtensionName != res.GetName() {
+			sc.addedResources = append(sc.addedResources, resource{name: sc.clusterExtensionName, kind: "clusterextension"})
+		}
 		sc.clusterExtensionName = res.GetName()
```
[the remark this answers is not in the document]
perdasilva: Fixed. Reverted to unconditionally adding the applied CE name to `addedResources` on every apply, rather than trying to save the previous name before overwriting. The previous approach would add the default placeholder name (`ce-${SCENARIO_ID}` from `CreateScenarioContext`) to cleanup even though no CE with that name was ever created.

The trade-off is that the last applied CE appears twice in cleanup (once from `addedResources`, once from `sc.clusterExtensionName`), but `--ignore-not-found=true` makes that harmless.

test/e2e/features/update.feature
```diff
@@ -231,7 +231,7 @@ Feature: Update ClusterExtension
       apiVersion: olm.operatorframework.io/v1
       kind: ClusterExtension
       metadata:
-        name: ${NAME}
+        name: ce-orig-${SCENARIO_ID}
```
pedjak: I would prefer that we keep using `${NAME}` - it is a very handy way to avoid leaking how we actually craft it under the hood.

test/e2e/features/update.feature
```diff
@@ -251,13 +251,12 @@ Feature: Update ClusterExtension
...
     And ClusterExtension is available
     And bundle "${PACKAGE:test}.1.0.1" is installed in version "1.0.1"
-    And the current ClusterExtension is tracked for cleanup
     When ClusterExtension is applied
       """
       apiVersion: olm.operatorframework.io/v1
       kind: ClusterExtension
       metadata:
-        name: ${NAME}-dup
+        name: ce-dup-${SCENARIO_ID}
```
pedjak: same here, what is an advantage of changing it to `ce-dup-${SCENARIO_ID}` ?

test/e2e/features/update.feature
```diff
@@ -271,7 +270,7 @@ Feature: Update ClusterExtension
                 "olm.operatorframework.io/metadata.name": ${CATALOG:test}
             version: 1.0.1
       """
-    Then ClusterExtension reports Progressing as True with Reason Retrying and Message includes:
+    Then ClusterExtension "ce-dup-${SCENARIO_ID}" reports Progressing as True with Reason Retrying and Message includes:
```
pedjak:
```suggestion
Then ClusterExtension "${NAME}-dup" reports Progressing as True with Reason Retrying and Message includes:
```

test/e2e/features/update.feature
```diff
@@ -338,3 +337,90 @@ Feature: Update ClusterExtension
     Then ClusterExtension reports "${NAME}-1, ${NAME}-2" as active revisions
     And ClusterObjectSet "${NAME}-2" reports Progressing as True with Reason RollingOut
     And ClusterObjectSet "${NAME}-2" reports Available as False with Reason ProbeFailure
+
+  @BoxcutterRuntime
+  @DeploymentConfig
+  Scenario: Cannot install a ClusterExtension that refers to an already installed bundle
```
pedjak: given that we have the previous test, do we really need this one at all? What is the added value?

test/e2e/features/update.feature
perdasilva: But once you have 2 resources, how do you disambiguate?

pedjak: > But once you have 2 resources, how do you disambiguate?

By name, we should track inside ScenarioContext resource kinds and their names

perdasilva: The problem is we track the CE last CE name as a way to drive unnamed steps, e.g. `ClusterExtension is Available`. Maybe we could remove the unnamed steps, keep `NAME` static for the execution and let templating produce the CE name. This would require we always provide the CE name in those steps. I'm not entirely against that - I don't think it hurts readability (just makes things more precise, and anyway, if the idea if to in some sense reproduce what the user would do, every kubectl command would take the CE name anyway). Wdyt?

perdasilva: I'd probably do this in a separate PR and rebase this one.

test/e2e/features/update.feature
perdasilva: I think the main differentiation is that in this one we ensure that when the conflicting CE has a higher revision COS it doesn't take over the resources (the regression). I think it would be ok to just merge these two since the first just checks same version collisions.

test/e2e/features/update.feature
pedjak: > The problem is we track the CE last CE name as a way to drive unnamed steps, e.g. `ClusterExtension is Available`

such steps should work as-is if there is just single CE in the context, then it fully clear which one we are talking about.

perdasilva: True, but then order becomes a dependency and the mechanics become obscure for the developer. They need to dig in to the code to understand why they can't reference a previous CE in a particular step. What do we lose by adding the additional clarity in each step? 

- It doesn't impact readability and is inline with user behavior w.r.t input
- It reduces state tracking as much as a possible - simplifying each step (reducing complexity)
- Not even the extra key strokes are a concern anymore since it will all probably be generated XD

perdasilva:
overriding as the changes in api are the e2e level - we are removing `TrackCurrentClusterExtensionForCleanup: removed` and reworking the resource cleanup process

test/e2e/features/update.feature
```diff
@@ -225,7 +225,8 @@ Feature: Update ClusterExtension
     And bundle "${PACKAGE:test}.1.0.4" is installed in version "1.0.4"
 
   @BoxcutterRuntime
-  Scenario: Detect collision when a second ClusterExtension installs the same package after an upgrade
+  @DeploymentConfig
+  Scenario: Detect cross-ClusterExtension collision protection
```
pedjak: I would revert this line to the previous version from above.

perdasilva: sorry - forgot about that!

pedjak (commented):
/approve

tmshort:
`/override` doesn't work here...

test/e2e/steps/hooks.go
```diff
@@ -268,9 +268,6 @@ func ScenarioCleanup(ctx context.Context, _ *godog.Scenario, err error) (context
 	}
 
 	forDeletion := sc.addedResources
-	if sc.clusterExtensionName != "" {
-		forDeletion = append(forDeletion, resource{name: sc.clusterExtensionName, kind: "clusterextension"})
-	}
```
tmshort, line 273: _NOT_ removing cluster extension? Will this cause issues for other tests?

test/e2e/steps/steps.go
```diff
@@ -449,7 +473,7 @@ func ResourceIsApplied(ctx context.Context, yamlTemplate *godog.DocString) error
 		return fmt.Errorf("failed to apply resource %v; err: %w; stderr: %s", out, err, stderrOutput(err))
 	}
 	if res.GetKind() == "ClusterExtension" {
-		sc.clusterExtensionName = res.GetName()
+		sc.addedResources = append(sc.addedResources, resource{name: res.GetName(), kind: "clusterextension"})
```
tmshort, line 476: Should you still update `sc.clusterExtensionName` here?

test/e2e/steps/hooks.go
tmshort: Added later in steps.go?

perdasilva: No, ClusterExtensions are still cleaned up. `ResourceIsApplied` now adds every applied CE to `sc.addedResources` (line 476 in steps.go), which `ScenarioCleanup` iterates over for deletion (line 270). This removed block was redundant — it just added `sc.clusterExtensionName` a second time, resulting in a duplicate delete (harmless due to `--ignore-not-found=true`, but unnecessary).

test/e2e/steps/steps.go
perdasilva: No, removing this is intentional. `sc.clusterExtensionName` is initialized once in `CreateScenarioContext` as `ce-<SCENARIO_ID>`, and `${NAME}` resolves to it. By not overwriting it on each apply, `${NAME}` stays constant throughout the scenario. This lets multi-CE scenarios use `${NAME}` for the original CE and `${NAME}-dup` for the duplicate without the implicit pointer drifting to the last-applied CE.

The overwrite was always a no-op for existing tests since they all apply CEs with `name: ${NAME}`, which resolves to the same value `sc.clusterExtensionName` already holds. Unnamed steps like `ClusterExtension is available` continue to work because they read `sc.clusterExtensionName`, which is still set — just never changed.

perdasilva: In a follow up I'll make the same changes to the COS handling

rashmigottipati (approved):
looks like all the review comments have been addressed and there will be a follow-up PR of COS handling based on the [review comment ](https://github.com/operator-framework/operator-controller/pull/2783#discussion_r3504051499). I am good with these changes. 

/lgtm

PR #2799, opened 2026-07-02
:seedling: Align ClusterObjectSet e2e cleanup with addedResources pattern

perdasilva:
## Summary
- Follow-up to #2783 — applies the same `addedResources`-based cleanup pattern to ClusterObjectSets in the e2e test harness
- `ResourceIsApplied` now appends ClusterObjectSets to `addedResources` (alongside setting `clusterObjectSetName` for `${COS_NAME}` substitution)
- Removes the special-case `clusterObjectSetName` cleanup block from `ScenarioCleanup`, so multiple ClusterObjectSets per scenario are all cleaned up correctly

## Test plan
- [ ] Verify e2e tests compile (`go build ./test/e2e/...`)
- [ ] Run BoxcutterRuntime e2e scenarios that create ClusterObjectSets and confirm cleanup succeeds
- [ ] Verify no leaked ClusterObjectSets remain after scenario completion

### DIFF
--- a/test/e2e/steps/steps.go
+++ b/test/e2e/steps/steps.go
@@ -475,7 +475,8 @@ func ResourceIsApplied(ctx context.Context, yamlTemplate *godog.DocString) error
 	if res.GetKind() == "ClusterExtension" {
 		sc.addedResources = append(sc.addedResources, resource{name: res.GetName(), kind: "clusterextension"})
 	} else if res.GetKind() == "ClusterObjectSet" {
-		sc.clusterObjectSetName = res.GetName()
+		sc.clusterObjectSetName = res.GetName() // used for ${COS_NAME} variable substitution in YAML templates
+		sc.addedResources = append(sc.addedResources, resource{name: res.GetName(), kind: "clusterobjectset"})
 	} else {
 		namespace := res.GetNamespace()
 		if namespace == "" {
--- a/test/e2e/steps/hooks.go
+++ b/test/e2e/steps/hooks.go
@@ -268,9 +268,6 @@ func ScenarioCleanup(ctx context.Context, _ *godog.Scenario, err error) (context
 	}
 
 	forDeletion := sc.addedResources
-	if sc.clusterObjectSetName != "" && featureGates[features.BoxcutterRuntime] {
-		forDeletion = append(forDeletion, resource{name: sc.clusterObjectSetName, kind: "clusterobjectset"})
-	}
 	for _, catalogName := range sc.catalogs {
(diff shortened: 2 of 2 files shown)


pedjak (commented):
## suggestion: Improve PR title and description for clarity

The current title describes the mechanism ("addedResources pattern") but not the problem being solved. Reviewers who haven't read #2783 won't know why alignment matters.

**Suggested title:**
> :seedling: Fix e2e cleanup to handle multiple ClusterObjectSets per scenario

**Suggested description:**

> Follow-up to #2783, which consolidated ClusterExtension cleanup into `addedResources`
> but left ClusterObjectSets on the legacy single-field path.
> 
> The old cleanup tracked only `sc.clusterObjectSetName` — a scalar that gets overwritten
> on each apply. Scenarios that create multiple ClusterObjectSets would only clean up the
> last one, leaking the rest. This PR applies the same fix #2783 used for ClusterExtensions:
> append each applied COS to `addedResources` at apply time and drop the special-case
> cleanup block.
> 
> **What changed:**
> - `ResourceIsApplied` now appends every ClusterObjectSet to `addedResources`
>   (the `clusterObjectSetName` field is still set for `${COS_NAME}` substitution)
> - `ScenarioCleanup` drops the single-COS cleanup block — `addedResources` handles it

test/e2e/steps/steps.go
```diff
@@ -476,6 +476,7 @@ func ResourceIsApplied(ctx context.Context, yamlTemplate *godog.DocString) error
 		sc.addedResources = append(sc.addedResources, resource{name: res.GetName(), kind: "clusterextension"})
 	} else if res.GetKind() == "ClusterObjectSet" {
 		sc.clusterObjectSetName = res.GetName()
+		sc.addedResources = append(sc.addedResources, resource{name: res.GetName(), kind: "clusterobjectset"})
```
pedjak, line 479: **nitpick (non-blocking):** `clusterObjectSetName` is still assigned here but cleanup has moved to `addedResources`. A short inline comment noting this field now only serves `${COS_NAME}` substitution would help future readers understand why it is still set.

pedjak:
**thought (non-blocking):** If a scenario applies the same COS name twice (e.g., updating it), `addedResources` will have duplicate entries and cleanup will attempt two deletes. Harmless due to `--ignore-not-found=true`, but noting for awareness.

/lgtm

rashmigottipati (approved):
/lgtm

test/e2e/steps/steps.go
perdasilva: updated ^^

pedjak (approved):
/approve

PR #2810, opened 2026-07-08
🐛 set Progressing condition to False on successful reconciliation

vdurgam10:
## Summary

- The `Progressing` condition on `ClusterExtension` resources was incorrectly set to `status: True` with `reason: Succeeded` after successful reconciliation. This is semantically wrong — `Progressing=True` means "still working," which contradicts the `Succeeded` reason and "Desired state reached" message.
- This fix changes the default `Progressing` status to `ConditionFalse` (done progressing) and only sets `ConditionTrue` when there is an active non-terminal error causing retries. Terminal errors correctly remain `ConditionFalse` with reason `Blocked`.

### Before
```
$ oc get clusterextension metallb-operator
NAME               INSTALLED BUNDLE                        VERSION               INSTALLED   PROGRESSING   AGE
metallb-operator   metallb-operator.v4.22.0-202607011454   4.22.0-202607011454   True        True          30h
```

### After
```
$ oc get clusterextension metallb-operator
NAME               INSTALLED BUNDLE                        VERSION               INSTALLED   PROGRESSING   AGE
metallb-operator   metallb-operator.v4.22.0-202607011454   4.22.0-202607011454   True        False         30h
```

### Condition semantics

| Scenario | Status | Reason |
|----------|--------|--------|
| Success (no error) | `False` | `Succeeded` |
| Non-terminal error (retrying) | `True` | `Retrying` |
| Terminal error (blocked) | `False` | `Blocked` / specific reason |

## Test plan

- [x] Unit tests updated and passing (`TestSetStatusProgressing` — all 8 sub-tests)
- [x] Integration tests updated and passing (`TestClusterExtensionInstallationSucceeds`, `TestResolutionFallbackToInstalledBundle`)
- [x] Full controllers test suite passing
- [x] Validated on OCP 4.22.2 cluster with MetalLB operator installed via ClusterExtension

Made with [Cursor](https://cursor.com)

### DIFF
--- a/internal/operator-controller/controllers/clusterextension_controller_test.go
+++ b/internal/operator-controller/controllers/clusterextension_controller_test.go
@@ -1352,7 +1352,7 @@ func TestClusterExtensionInstallationSucceeds(t *testing.T) {
 	t.Log("By checking the expected progressing conditions")
 	progressingCond := apimeta.FindStatusCondition(clusterExtension.Status.Conditions, ocv1.TypeProgressing)
 	require.NotNil(t, progressingCond)
-	require.Equal(t, metav1.ConditionTrue, progressingCond.Status)
+	require.Equal(t, metav1.ConditionFalse, progressingCond.Status)
 	require.Equal(t, ocv1.ReasonSucceeded, progressingCond.Reason)
 
 	require.NoError(t, cl.DeleteAllOf(ctx, &ocv1.ClusterExtension{}))
@@ -2674,7 +2674,7 @@ func TestResolutionFallbackToInstalledBundle(t *testing.T) {
 
 		progCond := apimeta.FindStatusCondition(ext.Status.Conditions, ocv1.TypeProgressing)
 		require.NotNil(t, progCond)
-		require.Equal(t, metav1.ConditionTrue, progCond.Status)
+		require.Equal(t, metav1.ConditionFalse, progCond.Status)
 		require.Equal(t, ocv1.ReasonSucceeded, progCond.Reason)
 
 		// Verify all conditions are present and valid after first reconcile
@@ -2695,7 +2695,7 @@ func TestResolutionFallbackToInstalledBundle(t *testing.T) {
 		// Progressing should be Succeeded (apply completed successfully)
 		progCond = apimeta.FindStatusCondition(ext.Status.Conditions, ocv1.TypeProgressing)
 		require.NotNil(t, progCond)
-		require.Equal(t, metav1.ConditionTrue, progCond.Status)
+		require.Equal(t, metav1.ConditionFalse, progCond.Status)
 		require.Equal(t, ocv1.ReasonSucceeded, progCond.Reason)
 
 		// Installed should be True (maintaining current version)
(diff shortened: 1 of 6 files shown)


perdasilva:
@vdurgam10 this was technically the original design. Internally, we are quite happy to walk this back. So, this change would also need an update to the documentation in the CR (https://github.com/operator-framework/operator-controller/blob/main/api/v1/clusterextension_types.go#L521). Since this technically constitutes a breaking change, we need to check internally whether we can push it through - this PR hopefully gives us enough data to show it would make sense to walk this design choice back...

perdasilva:
We'd also need to address the same issue in the ClusterObjectSet (https://github.com/operator-framework/operator-controller/blob/main/api/v1/clusterobjectset_types.go#L496) and probably update the CE controller logic for the change.

vdurgam10:
> @vdurgam10 this was technically the original design. Internally, we are quite happy to walk this back. So, this change would also need an update to the documentation in the CR (https://github.com/operator-framework/operator-controller/blob/main/api/v1/clusterextension_types.go#L521). Since this technically constitutes a breaking change, we need to check internally whether we can push it through - this PR hopefully gives us enough data to show it would make sense to walk this design choice back...

@perdasilva Thanks for looking into it. Although, I am not sure why this was the case. If we are to make the behaviour consistent across cluster components this does not match the behaviour. Do let me know if any inputs required from my side.

grokspawn:
OLMv1 Progressing condition is [documented](https://operator-framework.github.io/operator-controller/draft/concepts/clusterobjectsets/#progressing) to be in line with k8s deployments' [progressing condition](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#deployment-status). 

ClusterCatalog [also follows the same pattern](https://operator-framework.github.io/operator-controller/api-reference/olmv1-api-reference/#clustercatalogstatus).

Maintainers have had this discussion a few times and always rejected an impulse to change it.  What is the new argument for making this change?
/hold

joelanford:
> What is the new argument for making this change?

I thought of a pretty good one. Consider this scenario:
1. At `T0`, user creates a CE, which leads to a successful installation:
  - `Progressing=True/Succeeded`
  - `lastTransitionTime=T0`
2. Much later at `T100`, user modifies the CE to ask for a non-existent version:
  - `Progressing=True/Retrying`
  - `lastTransitionTime=T0` (because this only changes when status changes)

What currently happens at `T100+progressDeadlineMinutes` is.... nothing. We should exceed the deadline at that point, but there is nothing in the CE API anywhere that records when we've _actually_ started progressing from steady state again.

If we change (1) such that it results in `Progressing=False/Succeeded`, then (2) would transition to `Progressing=True`, which means `lastTransitionTime` would also be set to `T100`, which would mean would could close _that scenario's_ progressing deadline bug by setting `Progressing=False/ProgressDeadlineExceeded`, which we wouldn't subsequently change until we successfully resolve/apply the COS again.

EDIT: I opened a separate issue about that bug here: https://github.com/operator-framework/operator-controller/issues/2861

PR #2819, opened 2026-07-17
:bug:OCPBUGS-97746: Add fsync to storeCatalogData and storeIndexData

dis016:
: bug : OCPBUGS-97746 : Add fsync to storeCatalogData and storeIndexData

## Summary by CodeRabbit

* **Reliability**
  * Improved durability of catalog and index storage by synchronizing written data and renamed directories before completion.
  * Storage operations now consistently report errors encountered while encoding and synchronizing catalog data.
* **Bug Fixes**
  * Reduced the risk of incomplete persistence after catalog updates by ensuring metadata and index files are flushed before operations complete.

### DIFF
--- a/internal/catalogd/controllers/core/clustercatalog_controller.go
+++ b/internal/catalogd/controllers/core/clustercatalog_controller.go
@@ -270,6 +270,12 @@ func (r *ClusterCatalogReconciler) reconcile(ctx context.Context, catalog *ocv1.
 	}
 	baseURL := r.Storage.BaseURL(catalog.Name)
 
+	if err := r.Storage.VerifyAndSync(catalog.Name); err != nil {
+		verifyErr := fmt.Errorf("error verifying catalog content before serving: %w", err)
+		updateStatusProgressing(&catalog.Status, catalog.GetGeneration(), verifyErr)
+		return ctrl.Result{}, verifyErr
+	}
+
 	updateStatusProgressing(&catalog.Status, catalog.GetGeneration(), nil)
 	updateStatusServing(&catalog.Status, canonicalRef, unpackTime, baseURL, catalog.GetGeneration())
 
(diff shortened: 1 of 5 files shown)


tmshort:
@dis016 please prefix your PR title with `:bug:` (before the OCPBUGS jira space name).
Also address the copilot comments.

internal/catalogd/controllers/core/clustercatalog_controller.go
```diff
@@ -270,6 +270,12 @@ func (r *ClusterCatalogReconciler) reconcile(ctx context.Context, catalog *ocv1.
 	}
 	baseURL := r.Storage.BaseURL(catalog.Name)
 
+	if err := r.Storage.VerifyAndSync(catalog.Name); err != nil {
```
joelanford, line 273: It seems like the existing `Store` method can be responsible for calling `f.Sync()` since we call this only right after we've called `Store`.

And looking further, it seems like we'd want `storeCatalogData` and `storeIndexData` to sync after they've finished writing.

I'm not sure we need anything more than that to fix the bug. For example, is there truly a case where Sync succeeds, but then we can't actually read it?

I think I would expect `Store` to guarantee the initial write is fully synced for _all_ written files. And same thing for `ContentExists` checking previously written files.

tmshort:
This has been sitting around for 6 weeks without much activity.

internal/catalogd/storage/localdir.go
```diff
@@ -179,6 +179,13 @@ func (s *LocalDirV1) storeAtomicSwap(ctx context.Context, catalog string, fsys f
 		return "", err
 	}
 
+	if err := syncDir(catalogDir); err != nil {
```
perdasilva, line 182: would it be worthwhile to add a comment around why we're explicitly sync'ing the directories?

internal/catalogd/controllers/core/clustercatalog_controller.go
dis016: Hi @joelanford made the changes as suggested, please review.

PR #2825, opened 2026-07-22
⚠ make spec.namespace optional with managed namespace support and PSA support

nader-ziada:
# Description

When spec.namespace is omitted, operator-controller resolves a managed namespace from bundle metadata using the fallback chain: suggested-namespace-template > suggested-namespace > <packageName>-system.

The managed namespace is included as a ClusterObjectSet object with collision protection. Pod Security Admission labels from the bundle's suggested-namespace-template annotation are applied to the managed namespace, enabling operators to declare their PSA requirements

## Reviewer Checklist

- [ ] API Go Documentation

## Summary by CodeRabbit

* **New Features**
  * Extensions can omit the installation namespace to use an automatically managed namespace.
  * Managed namespaces are resolved from bundle metadata and can receive appropriate Pod Security labels.
  * User-specified namespaces continue to be supported without automatic label changes.
* **Bug Fixes**
  * Improved validation ensures specified namespaces exist, use valid names, and cannot be changed after creation.
  * Missing installation namespaces now report a retryable progress status.
* **Documentation**
  * Added guidance for managed namespaces and bundle author configuration.


internal/operator-controller/controllers/clusterextension_reconcile_steps.go
```diff
@@ -431,6 +502,8 @@ func ApplyBundle(a Applier) ReconcileStepFunc {
 		//     The only way to eventually recover from permission errors is to keep retrying).
 		rolloutSucceeded, rolloutStatus, err := a.Apply(ctx, state.imageFS, ext, objLbls, revisionAnnotations)
```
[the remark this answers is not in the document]
nader-ziada: the feature is for boxcutter path only

internal/operator-controller/controllers/clusterextension_admission_test.go
```diff
@@ -285,8 +285,8 @@ func TestClusterExtensionAdmissionInstallNamespace(t *testing.T) {
 		errMsg    string
 	}{
 		{"just alphanumeric", "justalphanumberic1", ""},
-		{"hyphen-separated", "hyphenated-name", ""},
-		{"no install namespace", "", regexMismatchError},
+		{"hypen-separated", "hyphenated-name", ""},
```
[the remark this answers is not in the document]
nader-ziada: already exisitng

internal/operator-controller/controllers/clusterextension_controller_test.go
```diff
@@ -979,6 +982,92 @@ func TestValidateClusterExtension(t *testing.T) {
...
 }
 
+func TestResolveNamespace(t *testing.T) {
+	tests := []struct {
+		name                 string
+		specNamespace        string
+		namespaceObjects     []runtime.Object
+		expectError          bool
+		errorMessageIncludes string
+	}{
```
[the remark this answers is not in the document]
nader-ziada: can add as a follow up later

internal/operator-controller/controllers/clusterobjectset_controller.go
```diff
@@ -580,6 +582,27 @@ func (c *ClusterObjectSetReconciler) resolveObjectRef(ctx context.Context, ref o
...
+				return fmt.Sprintf("namespace %q is already managed by %s %q", name, owner.Kind, ownerName)
+			}
+			return fmt.Sprintf("%s %q is already managed by %s %q", gvk.Kind, name, owner.Kind, ownerName)
+		}
+	}
+
+	if gvk.Kind == "Namespace" {
+		return fmt.Sprintf("namespace %q is already managed by another controller", name)
+	}
+	return fmt.Sprintf("%s.%s %s/%s collision: %s", gvk.Kind, gvk.GroupVersion(), obj.GetNamespace(), name, ores.String())
```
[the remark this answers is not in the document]
nader-ziada: fixed

internal/operator-controller/rukpak/render/namespace.go
```diff
@@ -0,0 +1,91 @@
...
+		}
+		if len(template.Annotations) > 0 {
+			ns.Annotations = template.Annotations
+		}
+	}
+
+	unstructuredObj, err := runtime.DefaultUnstructuredConverter.ToUnstructured(&ns)
+	if err != nil {
+		panic(fmt.Sprintf("failed to convert namespace to unstructured: %v", err))
+	}
```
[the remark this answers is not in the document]
nader-ziada: fixed

internal/operator-controller/controllers/clusterextension_reconcile_steps.go
```diff
@@ -402,6 +406,73 @@ func UnpackBundle(i imageutil.Puller, cache imageutil.Cache) ReconcileStepFunc {
...
+			}
+			state.resolvedNamespace = ext.Spec.Namespace
+			state.namespaceManaged = false
+			return nil, nil
+		}
+
+		// Managed mode: resolve name from bundle annotations
+		if state.imageFS == nil {
+			return nil, nil
+		}
```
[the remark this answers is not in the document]
nader-ziada: fixed

internal/operator-controller/controllers/clusterextension_reconcile_steps.go
```diff
@@ -402,6 +406,73 @@ func UnpackBundle(i imageutil.Puller, cache imageutil.Cache) ReconcileStepFunc {
...
+			if !apierrors.IsNotFound(getErr) {
+				return nil, fmt.Errorf("error checking namespace %q: %w", resolvedName, getErr)
+			}
+		}
+
+		state.resolvedNamespace = resolvedName
+		state.namespaceManaged = true
+		state.namespaceTemplate = template
+		l.Info("resolved managed namespace", "namespace", resolvedName)
+		return nil, nil
```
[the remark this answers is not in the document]
nader-ziada: fixed

internal/operator-controller/controllers/clusterextension_reconcile_steps.go
```diff
@@ -402,6 +406,79 @@ func UnpackBundle(i imageutil.Puller, cache imageutil.Cache) ReconcileStepFunc {
...
+	return func(ctx context.Context, state *reconcileState, ext *ocv1.ClusterExtension) (*ctrl.Result, error) {
+		l := log.FromContext(ctx)
+
+		if ext.Spec.Namespace != "" {
+			l.V(1).Info("validating user-provided namespace exists", "namespace", ext.Spec.Namespace)
+			_, err := nsClient.Namespaces().Get(ctx, ext.Spec.Namespace, metav1.GetOptions{})
+			if apierrors.IsNotFound(err) {
+				err := fmt.Errorf("namespace %q not found; spec.namespace must reference an existing namespace", ext.Spec.Namespace)
+				setStatusProgressing(ext, reconcile.TerminalError(err))
+				return nil, err
```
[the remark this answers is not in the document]
nader-ziada: fixed

internal/operator-controller/controllers/clusterextension_reconcile_steps.go
```diff
@@ -402,6 +406,79 @@ func UnpackBundle(i imageutil.Puller, cache imageutil.Cache) ReconcileStepFunc {
...
+		if err != nil {
+			return nil, fmt.Errorf("error reading bundle annotations: %w", err)
+		}
+
+		packageName := getPackageName(ext)
+		resolvedName, template, err := applier.ResolveNamespaceName(bundleAnnotations, packageName)
+		if err != nil {
+			err := fmt.Errorf("error resolving namespace from bundle annotations: %w", err)
+			setStatusProgressing(ext, reconcile.TerminalError(err))
+			return nil, err
```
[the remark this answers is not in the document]
nader-ziada: fixed

internal/operator-controller/controllers/clusterextension_reconcile_steps.go
```diff
@@ -402,6 +406,79 @@ func UnpackBundle(i imageutil.Puller, cache imageutil.Cache) ReconcileStepFunc {
...
+
+		// On first install, verify managed namespace does not already exist.
+		// On upgrade (status.Namespace already set), skip this check — the namespace was created by us.
+		if ext.Status.Namespace == "" {
+			l.V(1).Info("checking managed namespace does not already exist", "namespace", resolvedName)
+			_, getErr := nsClient.Namespaces().Get(ctx, resolvedName, metav1.GetOptions{})
+			if getErr == nil {
+				err := fmt.Errorf("managed namespace %q already exists; use spec.namespace to install into an existing namespace", resolvedName)
+				setStatusProgressing(ext, reconcile.TerminalError(err))
+				return nil, err
```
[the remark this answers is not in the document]
nader-ziada: fixed

internal/operator-controller/controllers/clusterextension_reconcile_steps.go
```diff
@@ -402,6 +406,79 @@ func UnpackBundle(i imageutil.Puller, cache imageutil.Cache) ReconcileStepFunc {
...
+		if err != nil {
+			err := fmt.Errorf("error resolving namespace from bundle annotations: %w", err)
+			setStatusProgressing(ext, reconcile.TerminalError(err))
+			return nil, err
+		}
+
+		if ext.Status.Namespace != "" && ext.Status.Namespace != resolvedName {
+			err := fmt.Errorf("bundle upgrade changes managed namespace from %q to %q; this is not supported", ext.Status.Namespace, resolvedName)
+			setStatusProgressing(ext, reconcile.TerminalError(err))
+			return nil, err
```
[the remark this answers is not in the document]
nader-ziada: fixed

internal/operator-controller/applier/namespace.go
```diff
@@ -0,0 +1,91 @@
...
+	// Try suggested-namespace annotation
+	if csvAnnotations != nil {
+		if name, ok := csvAnnotations[AnnotationSuggestedNamespace]; ok && name != "" {
+			return name, template, nil
+		}
+	}
+
+	// Fallback: <packageName>-system
+	return fmt.Sprintf("%s-system", packageName), template, nil
+}
```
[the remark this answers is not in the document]
nader-ziada: fixed

cmd/operator-controller/main.go
```diff
@@ -746,6 +747,7 @@ func (c *helmReconcilerConfigurator) Configure(ceReconciler *controllers.Cluster
 		controllers.RetrieveRevisionStates(revisionStatesGetter),
 		controllers.ResolveBundle(c.resolver, c.mgr.GetClient()),
 		controllers.UnpackBundle(c.imagePuller, c.imageCache),
+		controllers.ResolveNamespace(coreClient),
 		controllers.ApplyBundle(appl),
 	}
```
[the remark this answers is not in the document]
nader-ziada: fixed

api/v1/clusterextension_types.go
```diff
@@ -545,6 +549,13 @@ type ClusterExtensionStatus struct {
 	// +optional
 	Install *ClusterExtensionInstallStatus `json:"install,omitempty"`
 
+	// namespace is the resolved namespace where the extension is installed.
+	// For user-provided namespaces, this mirrors spec.namespace.
+	// For managed namespaces, this shows the name resolved from bundle metadata.
+	//
+	// +optional
+	Namespace string `json:"namespace,omitempty"`
```
joelanford: This field only seems relevant for bundles that assume a single namespace as an install namespace. That is true for registry+v1 bundles, but is not generally true at the cluster scope.

For now, I'd suggest dropping this. The set of managed objects (including the registry+v1 install namespace will be discoverable in the ClusterObjectSet.

We can wait to see what feedback we get from real-world usage to help us decide what if anything to surface in the CE status.

docs/concepts/managed-namespaces.md
```diff
@@ -0,0 +1,45 @@
+# Managed Namespaces
+
+## What is a managed namespace?
+
+When you create a ClusterExtension without specifying `spec.namespace`, operator-controller automatically creates and manages a namespace for the operator. The namespace name comes from the bundle's metadata or defaults to `<packageName>-system`.
```
joelanford: True for registry+v1, but other bundles may just directly include one (or more!) namespaces. I think that's why @perdasilva was suggesting that we move this field to be in the config and deprecate it in the spec.

docs/concepts/managed-namespaces.md
```diff
@@ -0,0 +1,45 @@
...
+
+When you specify `spec.namespace`, the namespace must already exist on the cluster and operator-controller installs into it without managing its lifecycle.
+
+The mode is locked at creation time: you cannot switch between managed and user-provided after the ClusterExtension is created.
+
+## Namespace resolution
+
+In managed mode, the namespace name is resolved from bundle CSV annotations in this order:
```
joelanford: We should call out that this is the logic for registry+v1.

And then a line that says other formats are like to have different behaviors for namespace handling (again this points more toward namespace as bundle-specific config rather than CE-level specification)

docs/draft/concepts/managed-namespaces.md
joelanford, line 1: Let's move docs to the drafts folder. We have a process where @michaelryanpeter reviews our drafts and incorporated into the live docs site as a separate step.

joelanford (commented):
I think we may want to refactor to deprecate spec.namespace, register a new field in the registry+v1 config schema for namespace. And then implement the logic on the bundle converter, which already reads and applies the config.

Would that work?

nader-ziada:
> I think we may want to refactor to deprecate spec.namespace, register a new field in the registry+v1 config schema for namespace. And then implement the logic on the bundle converter, which already reads and applies the config.

we had originally planned the deprecation and removal to be phase 2, once we confirm everything else looks okay, will go ahead with that change

docs/draft/concepts/managed-namespaces.md
nader-ziada: moved

docs/concepts/managed-namespaces.md
nader-ziada: explicitly called out registry+v1 scope and that other formats will likely differ

docs/concepts/managed-namespaces.md
nader-ziada: yes, will update doc

api/v1/clusterextension_types.go
nader-ziada: removed

tmshort:
Any status upstates here? This PR has been idle for 2 weeks.

nader-ziada:
> Any status upstates here? This PR has been idle for 2 weeks.

We have a meeting scheduled for next week to discuss

joelanford:
Mentioned in Slack, but repeating here in case this is easier to consume:

> I reviewed as well and agree with Per's Slack comments. We should not think of the `Namespace` as being a special thing that gets special conflict treatment. It is just any other object that may or not may pop out of a bundle rendering. That being said, improving the overall UX of object conflicts is possibly a worthwhile thing to do, but I'd defer that until we get specific user feedback about it.
>
> Tangibly, my suggestions are:
>
> 1. Drop `nsConfig` and its plumbing entirely
> 2. Put all of the namespace extraction logic into the BundleRenderer as an implementation detail
> 3. Change the signature of the renderer to drop the install namespace parameter
> 4. Add a WithSelfManagedInstallNamespace() option to the BundleRenderer. If `ce.spec.namespace` is set, we pass that option in as `WithSelfManagedInstallNamespace(ce.Spec.Namespace)`  (self-managed means managed by the user/caller, NOT the library)
> 5. If `WithSelfManagedInstallNamespace` is NOT included in the options, then the renderer generates a system managed namespace object based on our precedence logic.

internal/operator-controller/applier/provider.go
```diff
@@ -132,7 +151,9 @@ type RegistryV1HelmChartProvider struct {
 }
 
 func (r *RegistryV1HelmChartProvider) Get(bundleFS fs.FS, ext *ocv1.ClusterExtension) (*chart.Chart, error) {
-	objs, err := r.ManifestProvider.Get(bundleFS, ext)
+	// The Helm path never manages the namespace lifecycle, so it renders into the
+	// user-provided namespace and never emits a Namespace object.
+	objs, err := r.ManifestProvider.Get(bundleFS, ext, NamespaceConfig{Target: ext.Spec.Namespace})
```
joelanford: Since (if?) we're making the `ce.Spec.Namespace` field optional for both standard and experimental builds, we'll need to make the helm chart provider understand what to do when `ce.Spec.Namespace` is unset/empty. I assume there will also be implications on the Helm-to-Boxcutter migrator.

nader-ziada: The Helm chart provider never sees an unset namespace and the migrator's spec.namespace == "" skip stays valid. Doesn't change by this PR, but supporting managed namespaces in the standard/Helm build is a bigger change, the Helm applier needs the resolved namespace for the release itself (it passes ext.Spec.Namespace to Install/Upgrade with CreateNamespace=false), and resolution now lives inside the renderer, should that be part of this PR as well?

internal/operator-controller/rukpak/render/render.go
```diff
@@ -126,22 +143,34 @@ type BundleRenderer struct {
...
 		TargetNamespaces:    defaultTargetNamespacesForBundle(&rv1),
 		UniqueNameGenerator: DefaultUniqueNameGenerator,
 		CertificateProvider: nil,
-	}).apply(opts...).validate(&rv1)
+	}).apply(opts...)
+
+	// When the install namespace is not caller-managed, resolve a system-managed
+	// namespace from the bundle and emit a Namespace object as part of the set.
+	var systemNamespace client.Object
+	if !genOpts.selfManagedNamespace {
```
perdasilva: maybe we could make a self-managed namespace generator function and add it to the rest collection of generator and basically use this implementation within it? (i.e. selfManagedNamespace is set - produce the object, if not, bail). Wdyt?

internal/operator-controller/rukpak/render/render.go
```diff
@@ -156,34 +181,14 @@ func (r BundleRenderer) Render(rv1 bundle.RegistryV1, opts ...Option) ([]client.
...
+		if err := mutate(ctx); err != nil {
 			return nil, err
 		}
-		genOpts.InstallNamespace = name
-		if systemNamespace, err = buildNamespaceObject(name, template); err != nil {
-			return nil, err
-		}
-	}
-
-	if _, errs := genOpts.validate(&rv1); len(errs) > 0 {
```
perdasilva: we no longer validate the options?

nader-ziada: fixed

internal/operator-controller/controllers/clusterextension_reconcile_steps.go
```diff
@@ -402,6 +405,33 @@ func UnpackBundle(i imageutil.Puller, cache imageutil.Cache) ReconcileStepFunc {
...
+		l := log.FromContext(ctx)
+
+		if ext.Spec.Namespace == "" {
+			return nil, nil
+		}
+
+		l.V(1).Info("validating user-provided namespace exists", "namespace", ext.Spec.Namespace)
+		_, err := nsClient.Namespaces().Get(ctx, ext.Spec.Namespace, metav1.GetOptions{})
+		if apierrors.IsNotFound(err) {
+			termErr := reconcile.TerminalError(fmt.Errorf("namespace %q not found; spec.namespace must reference an existing namespace", ext.Spec.Namespace))
```
perdasilva: In principle, it probably doesn't need to be a terminal error, i.e. the user could create the namespace and everything works on the following retry

internal/operator-controller/controllers/clusterobjectset_controller.go
```diff
@@ -200,23 +200,25 @@ func (c *ClusterObjectSetReconciler) reconcile(ctx context.Context, cos *ocv1.Cl
 		return ctrl.Result{RequeueAfter: 10 * time.Second}, nil
 	}
 
-	for i, pres := range rres.GetPhases() {
+	for _, pres := range rres.GetPhases() {
 		if verr := pres.GetValidationError(); verr != nil {
-			l.Error(fmt.Errorf("%w", verr), "phase preflight validation failed, retrying after 10s", "phase", i)
-			setRetryingConditions(l, cos, fmt.Sprintf("phase %d validation error: %s", i, verr), isDeadlineExceeded)
+			phaseName := pres.GetName()
```
perdasilva: nice - thank you!

internal/operator-controller/controllers/clusterobjectset_controller.go
```diff
@@ -580,6 +582,30 @@ func (c *ClusterObjectSetReconciler) resolveObjectRef(ctx context.Context, ref o
 	return obj, nil
 }
 
+func collisionMessage(ores machinery.ObjectResult) string {
```
perdasilva: awesome <3

internal/operator-controller/applier/provider.go
```diff
@@ -187,7 +187,7 @@ func extensionConfigBytes(ext *ocv1.ClusterExtension) []byte {
 	return nil
 }
 
-func getBundleAnnotations(bundleFS fs.FS) (map[string]string, error) {
+func GetBundleAnnotations(bundleFS fs.FS) (map[string]string, error) {
```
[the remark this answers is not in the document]
perdasilva: might be worth adding the godoc =D

internal/operator-controller/controllers/clusterextension_reconcile_steps.go
nader-ziada: fixed

api/v1/clusterextension_types.go
```diff
@@ -49,21 +49,32 @@ const (
...
+	// <opcon:experimental:validation:XValidation:rule="oldSelf != '' || self == ''",message="namespace cannot be set after creation; mode is locked at creation time">
+	//
 	// +kubebuilder:validation:MaxLength:=63
-	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="namespace is immutable"
-	// +kubebuilder:validation:XValidation:rule="self.matches(\"^[a-z0-9]([-a-z0-9]*[a-z0-9])?$\")",message="namespace must be a valid DNS1123 label"
-	// +required
+	// +kubebuilder:validation:XValidation:rule="self == '' || self.matches(\"^[a-z0-9]([-a-z0-9]*[a-z0-9])?$\")",message="namespace must be a valid DNS1123 label"
+	// +kubebuilder:validation:XValidation:rule="oldSelf == '' || self == oldSelf",message="namespace is immutable once set"
+	// +optional
 	Namespace string `json:"namespace"`
```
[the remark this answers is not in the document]
nader-ziada: fixed

applyconfigurations/api/v1/clusterextensionspec.go
```diff
@@ -22,15 +22,26 @@ package v1
...
+	// creation time and cannot be changed. Omitting namespace requires the experimental
+	// feature set (BoxcutterRuntime).
+	// </opcon:experimental:description>
+	//
+	// The namespace field follows the DNS label standard as defined in [RFC 1123].
 	// It must contain only lowercase alphanumeric characters or hyphens (-), start and end with an alphanumeric character,
 	// and be no longer than 63 characters.
 	//
+	// <opcon:standard:validation:Required>
+	// <opcon:experimental:validation:XValidation:rule="oldSelf != ” || self == ”",message="namespace cannot be set after creation; mode is locked at creation time">
```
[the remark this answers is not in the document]
nader-ziada: fixed

docs/api-reference/olmv1-api-reference.md
```diff
@@ -358,7 +358,7 @@ _Appears in:_
...
...ce` _string_ | namespace references an existing namespace where namespace-scoped resources<br />for the extension are applied. The namespace must already exist on the cluster.<br /><opcon:standard:description><br />namespace is required.<br /></opcon:standard:description><br /><opcon:experimental:description><br />namespace is optional. When omitted, operator-controller resolves and creates a<br />managed namespace from bundle metadata. The mode (set vs omitted) is locked at<br />creation time and cannot be changed. Omitting namespace requires the experimental<br />feature set (BoxcutterRuntime).<br /></opcon:experimental:description><br />The namespace field follows the DNS label standard as defined in [RFC 1123].<br />It must contain only lowercase alphanumeric characters or hyphens (-), start and end with an alphanumeric character,<br />and be no longer than 63 characters.<br />[RFC 1123]: https://tools.ietf.org/html/rfc1123<br /><opcon:standard:validation:Required><br /><opcon:experimental:validation:XValidation:rule="oldSelf != '' \|\| self == ''",message="namespace cannot be set after creation; mode is locked at creation time"> |  | MaxLength: 63 <br />Optional: \{\} <br /> |
```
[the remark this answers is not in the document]
nader-ziada: fixed

docs/draft/concepts/managed-namespaces.md
```diff
@@ -0,0 +1,53 @@
...
+
+1. `operatorframework.io/suggested-namespace-template`: the `metadata.name` field from the JSON template
+2. `operatorframework.io/suggested-namespace`: a plain string with the preferred name
+3. `<packageName>-system`: convention fallback
+
+## What belongs in a managed namespace
+
+- The operator's own workloads (deployments, services, configmaps)
+- The operator's RBAC resources (service accounts, roles, role bindings)
+- CRDs and webhooks installed by the operator
```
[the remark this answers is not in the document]
nader-ziada: fixed

internal/operator-controller/rukpak/render/namespace.go
```diff
@@ -0,0 +1,118 @@
...
+	case template != nil && template.Name != "":
+		name = template.Name
+	case csvAnnotations[AnnotationSuggestedNamespace] != "":
+		name = csvAnnotations[AnnotationSuggestedNamespace]
+	default:
+		name = fmt.Sprintf("%s-system", rv1.PackageName)
+	}
+
+	if err := validateNamespaceName(name); err != nil {
+		return "", nil, err
```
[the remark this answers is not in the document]
nader-ziada: fixed

api/v1/clusterextension_types.go
```diff
@@ -49,21 +49,34 @@ const (
...
+	// namespace selects the namespace that namespace-scoped resources for the extension
+	// are applied to.
 	//
-	// The namespace field is required, immutable, and follows the DNS label standard as defined in [RFC 1123].
+	// <opcon:standard:description>
+	// In the standard configuration, namespace is required and must reference an existing
+	// namespace on the cluster.
+	// </opcon:standard:description>
+	// <opcon:experimental:description>
+	// In the experimental configuration (BoxcutterRuntime feature set), namespace is optional.
```
joelanford: Nit: Remove `In the experimental configuration`. When we promote this to standard, we can ideally use the experimental text verbatim, and not have to remember to update the text as well.

nader-ziada: fixed

internal/operator-controller/controllers/clusterextension_admission_test.go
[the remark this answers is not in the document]
tmshort: No, it wasn't already existing. The diff indicates it was previously `hyphen-separated`

tmshort:
There's a "hypen"-typo that was introduced in this that ought to be fixed (it was a Copilot comment originally)

tmshort:
closed/reopened to trigger CI?

api/v1/clusterextension_types.go
```diff
@@ -49,21 +49,34 @@ const (
...
-	// It designates the default namespace where namespace-scoped resources for the extension are applied to the cluster.
-	// Some extensions may contain namespace-scoped resources to be applied in other namespaces.
-	// This namespace must exist.
+	// namespace selects the namespace that namespace-scoped resources for the extension
+	// are applied to.
 	//
-	// The namespace field is required, immutable, and follows the DNS label standard as defined in [RFC 1123].
+	// <opcon:standard:description>
+	// In the standard configuration, namespace is required and must reference an existing
+	// namespace on the cluster.
```
joelanford: Sorry didn't notice this on my first comment about the CRD changes: Let's leave the standard configuration CRD completely unchanged.  When we re-gen the CRDs, I'd expect `helm/olmv1/base/operator-controller/crd/standard/olm.operatorframework.io_clusterextensions.yaml` to be unchanged.

api/v1/clusterextension_types.go
```diff
@@ -49,21 +49,34 @@ const (
...
+	// </opcon:standard:description>
+	// <opcon:experimental:description>
+	// BoxcutterRuntime feature set, namespace is optional.
```
joelanford: Nit again, sorry: Ignore feature gates, experimental vs. standard, etc. Just plainly state what the field is/means, etc. for users of the experimental release. There should be no appearance of "when experimental is enabled" or "with the BoxcutterRuntime feature gate", etc.

Imagine this feature has been promoted to standard and it just _is_ this way in GA. Use _that_ description. That way when we do promote to standard, all we need to do is drop the `</opcon>` markers.

helm/olmv1/base/operator-controller/crd/standard/olm.operatorframework.io_clusterextensions.yaml
```diff
@@ -445,8 +448,8 @@ spec:
                   rule: 'has(self.sourceType) && self.sourceType == ''Catalog'' ?
                     has(self.catalog) : !has(self.catalog)'
             required:
-            - namespace
             - source
+            - namespace
```
joelanford: There should be no changes to the standard CRD, right? Aren't we isolating our changes to the experimental CRD?

cmd/operator-controller/main.go
```diff
@@ -746,6 +748,7 @@ func (c *helmReconcilerConfigurator) Configure(ceReconciler *controllers.Cluster
 		controllers.RetrieveRevisionStates(revisionStatesGetter),
 		controllers.ResolveBundle(c.resolver, c.mgr.GetClient()),
 		controllers.UnpackBundle(c.imagePuller, c.imageCache),
+		controllers.ValidateInstallNamespace(coreClient),
```
joelanford: I thought we decided on no changes to the helm reconciler/applier path?

docs/draft/howto/namespace-configuration-for-authors.md
joelanford, line 1: Generally docs start out in `drafts` and then we have an docs-focused community member review and move to the "main" docs.

Not sure how consistent we've been with that though. For now, I'd suggest moving to drafts, and we should then separately review our drafts and make a push for them to be moved to the main docs.

internal/operator-controller/applier/boxcutter.go
```diff
@@ -125,7 +126,7 @@ func (r *SimpleRevisionGenerator) GenerateRevision(
 	}
 
 	// add bundle properties of interest to revision annotations
-	bundleAnnotations, err := getBundleAnnotations(bundleFS)
+	bundleAnnotations, err := GetBundleAnnotations(bundleFS)
```
joelanford: Does this need to be exported?

internal/operator-controller/applier/boxcutter_test.go
```diff
@@ -92,66 +92,30 @@ func Test_SimpleRevisionGenerator_GenerateRevisionFromHelmRelease(t *testing.T)
...
+
+	// The Helm-release migration path never injects a namespace (the release's
+	// namespace already exists), so only the configuration phase is present.
+	require.Len(t, rev.Spec.Phases, 1)
+
+	configPhase := rev.Spec.Phases[0]
+	assert.Equal(t, "configuration", *configPhase.Name)
+	require.Len(t, configPhase.Objects, 2)
+	assert.Equal(t, "ConfigMap", configPhase.Objects[0].Object.GetKind())
+	assert.Equal(t, "Secret", configPhase.Objects[1].Object.GetKind())
```
joelanford: This change seems unrelated to this PR?

internal/operator-controller/applier/boxcutter_test.go
```diff
@@ -414,10 +378,17 @@ func Test_SimpleRevisionGenerator_AppliesObjectLabelsAndRevisionAnnotations(t *t
...
 		for _, revObj := range phase.Objects {
-			require.Equal(t, map[string]string{
-				"app":  "test-obj",
-				"some": "value",
-			}, revObj.Object.GetLabels())
+			// Namespace objects only have objectLabels, not bundle object labels
+			if revObj.Object.GetKind() == "Namespace" {
+				require.Equal(t, map[string]string{
+					"some": "value",
+				}, revObj.Object.GetLabels())
```
joelanford: The CE sets `spec.Namespace`, so I'd not expect a `Namespace` object in the phases, right?

internal/operator-controller/applier/boxcutter_test.go
```diff
@@ -1651,3 +1622,167 @@ func TestBoxcutterStorageMigrator(t *testing.T) {
...
+	t.Log("by checking that the namespaces phase contains exactly one namespace object")
+	require.Len(t, firstPhase.Objects, 1, "namespaces phase should contain exactly one object")
+
+	t.Log("by checking that the namespace object has the correct name")
+	nsObj := firstPhase.Objects[0].Object
+	require.Equal(t, "Namespace", nsObj.GetKind())
+	require.Equal(t, "test-namespace", nsObj.GetName(), "namespace name should match ext.Spec.Namespace")
+}
+
+func Test_GenerateRevision_COSHasOwnerLabels(t *testing.T) {
```
joelanford: Is this unrelated to the scope of the PR?

Was this a code gap that you found and fixed in this PR, or just additional test coverage?

internal/operator-controller/applier/provider.go
```diff
@@ -34,6 +35,7 @@ type RegistryV1ManifestProvider struct {
 	IsWebhookSupportEnabled     bool
 	IsSingleOwnNamespaceEnabled bool
 	IsDeploymentConfigEnabled   bool
+	IsBoxcutterRuntimeEnabled   bool
```
joelanford: Nit: I'd suggest renaming this to something more semantic related to the feature it gates in the manifest provider rather than the feature gate we decided to put it behind.

Maybe `IsNamespaceManagementEnabled`, but still driven in `main.go` from the BoxcutterRuntime feature gate.

internal/operator-controller/applier/provider.go
```diff
@@ -67,10 +69,23 @@ func (r *RegistryV1ManifestProvider) Get(bundleFS fs.FS, ext *ocv1.ClusterExtens
...
 	}
 
+	// When the user set spec.namespace, render into that (already-existing) namespace and do
+	// not emit a Namespace object. Otherwise default to the bundle's system-managed namespace
+	// and have the renderer emit the Namespace object for it.
+	if ext.Spec.Namespace != "" {
+		opts = append(opts, render.WithInstallNamespace(ext.Spec.Namespace))
+	} else {
+		opts = append(opts, render.RenderInstallNamespace())
+	}
```
joelanford, line 85: What happens if neither are provided? What happens if both are provided?

internal/operator-controller/applier/provider.go
```diff
@@ -67,10 +69,23 @@ func (r *RegistryV1ManifestProvider) Get(bundleFS fs.FS, ext *ocv1.ClusterExtens
 		return nil, fmt.Errorf("unsupported bundle: bundle must support at least one of [AllNamespaces SingleNamespace OwnNamespace] install modes")
 	}
 
+	if ext.Spec.Namespace == "" && !r.IsBoxcutterRuntimeEnabled {
+		return nil, errorutil.NewTerminalError(ocv1.ReasonInvalidConfiguration, fmt.Errorf("spec.namespace is required unless the BoxcutterRuntime feature gate is enabled"))
+	}
```
joelanford: Nit: not sure this is really necessary since the namespace can only ever be empty when the feature is enabled.

internal/operator-controller/controllers/clusterextension_admission_test.go
```diff
@@ -325,9 +325,87 @@ func TestClusterExtensionAdmissionInstallNamespace(t *testing.T) {
 	}
 }
 
-// TestClusterExtensionAdmissionServiceAccount validates the deprecated spec.serviceAccount field:
-// - CRD-level validation (format, length) still works
-// - ValidatingAdmissionPolicy emits a deprecation warning for valid non-empty values
+func TestClusterExtensionAdmissionNamespaceImmutability(t *testing.T) {
```
joelanford, line 329: I wonder if experimental should just let go of this immutability and allow changes: set to unset, unset to set, and set to something else.

We imagined that we'd eventually deprecate `spec.namespace` and move the field to be sourced from `spec.config`.  Now that we don't have an SA field, I'm thinking the need for immutable namespace field is now much lower, maybe non-existent.

internal/operator-controller/controllers/clusterobjectset_controller.go
```diff
@@ -200,23 +200,25 @@ func (c *ClusterObjectSetReconciler) reconcile(ctx context.Context, cos *ocv1.Cl
 		return ctrl.Result{RequeueAfter: 10 * time.Second}, nil
 	}
 
-	for i, pres := range rres.GetPhases() {
+	for _, pres := range rres.GetPhases() {
 		if verr := pres.GetValidationError(); verr != nil {
-			l.Error(fmt.Errorf("%w", verr), "phase preflight validation failed, retrying after 10s", "phase", i)
-			setRetryingConditions(l, cos, fmt.Sprintf("phase %d validation error: %s", i, verr), isDeadlineExceeded)
+			phaseName := pres.GetName()
+			l.Error(fmt.Errorf("%w", verr), "phase preflight validation failed, retrying after 10s", "phase", phaseName)
```
joelanford: Nit: Unrelated to this PR? If we leave this change in (not opposed to it really), I'd suggest leaving the phase index in the error and retrying condition message.

internal/operator-controller/controllers/clusterobjectset_controller.go
```diff
@@ -580,6 +582,30 @@ func (c *ClusterObjectSetReconciler) resolveObjectRef(ctx context.Context, ref o
...
+	obj := ores.Object()
+	gvk := obj.GetObjectKind().GroupVersionKind()
+	name := obj.GetName()
+
+	if collision, ok := ores.(machinery.ObjectResultCollision); ok {
+		if owner, hasOwner := collision.ConflictingOwner(); hasOwner {
+			ownerName := owner.Name
+			if gvk.Kind == "Namespace" {
+				return fmt.Sprintf("namespace %q is already managed by %s %q", name, owner.Kind, ownerName)
+			}
```
joelanford: This feels like a slippery slope to me. Why would we give special treatment to namespaces, but not any other object type that has a conflict?

internal/operator-controller/rukpak/render/registryv1/generators/generators_test.go
```diff
@@ -62,6 +62,45 @@ func Test_ResourceGenerators_Errors(t *testing.T) {
...
+	t.Run("seeds labels and annotations from the template", func(t *testing.T) {
+		objs, err := generators.BundleInstallNamespaceGenerator(&bundle.RegistryV1{}, render.Options{
+			InstallNamespace:         "install-namespace",
+			GenerateInstallNamespace: true,
+			InstallNamespaceTemplate: &corev1.Namespace{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels:      map[string]string{"pod-security.kubernetes.io/enforce": "privileged"},
+					Annotations: map[string]string{"example.com/foo": "bar"},
+				},
+			},
```
joelanford: Any reason we need both `InstallNamespace` and `InstallNamespaceTemplate`. Could we just have `InstallNamespace *corev1.Namespace`, and if all we have is a name (and no metadata), we'd just populate/read from `InstallNamespace.Metadata.Name`?

Otherwise, it seems like there's the awkward possibility of `InstallNamespace` and `InstallNamespaceTemplate` disagreeing on the name.

api/v1/clusterextension_types.go
nader-ziada: to let experimental users skip namespace, this PR loosened the validation rules on the namespace field, but those rules are shared, so the change showed up in both the standard and experimental CRDs. That's why standard changed.

api/v1/clusterextension_types.go
nader-ziada: Done, reworded the experimental description as plain GA-style prose with no feature-gate or channel language, so promotion is just dropping the <opcon> markers.

helm/olmv1/base/operator-controller/crd/standard/olm.operatorframework.io_clusterextensions.yaml
nader-ziada: you're right, they should be isolated, but they aren't yet. To let experimental users skip namespace, the PR loosened the shared field markers (validation + required + description), and since those markers feed both channels, the change leaked into standard too. I can keep standard byte-identical by moving the relaxation entirely into experimental (managed mode = field omitted rather than ""), if we're OK rejecting an explicit namespace: "".

cmd/operator-controller/main.go
nader-ziada: fixed this, ValidateInstallNamespace is only wired into the boxcutter configurator now

docs/draft/howto/namespace-configuration-for-authors.md
nader-ziada: moved them

internal/operator-controller/applier/boxcutter.go
nader-ziada: un-exported it

internal/operator-controller/applier/boxcutter_test.go
nader-ziada: Reverted the unrelated test churn.

internal/operator-controller/applier/boxcutter_test.go
nader-ziada: removed it, it was extra coverage unrelated to this PR's scope, part of the refactor churn i guess

internal/operator-controller/applier/provider.go
nader-ziada: renamed

internal/operator-controller/applier/provider.go
nader-ziada: In this code, "neither" and "both" can't happen. It's a simple either/or on one thing: is spec.namespace set or not?

internal/operator-controller/applier/provider.go
nader-ziada: Kept as defense-in-depth in case the provider is invoked without CRD validation (e.g. direct/unit-test callers), but happy to drop it if you'd prefer.

internal/operator-controller/controllers/clusterextension_admission_test.go
nader-ziada: Agreed the SA-driven rationale is gone, but immutability still guards against stranding already-installed resources when the namespace changes, relaxing it safely needs GC/re-apply-on-change handling. I'd prefer to keep it immutable here and revisit when we move the field to spec.config.

internal/operator-controller/controllers/clusterobjectset_controller.go
nader-ziada: Reverted

internal/operator-controller/controllers/clusterobjectset_controller.go
nader-ziada: removed the namespace special-casing entirely

internal/operator-controller/rukpak/render/registryv1/generators/generators_test.go
nader-ziada: Rather than collapse to *corev1.Namespace (which forces every namespace-name read to .Name), I dropped the typed template and kept only the labels/annotations it actually contributes, the name now has a single source of truth in InstallNamespace and there's no second .Name to disagree with.

nader-ziada:
> There's a "hypen"-typo that was introduced in this that ought to be fixed (it was a Copilot comment originally)

fixed this in the last commit

internal/operator-controller/applier/boxcutter_test.go
nader-ziada: fixed

internal/operator-controller/applier/provider.go
```diff
@@ -67,10 +69,23 @@ func (r *RegistryV1ManifestProvider) Get(bundleFS fs.FS, ext *ocv1.ClusterExtens
 		return nil, fmt.Errorf("unsupported bundle: bundle must support at least one of [AllNamespaces SingleNamespace OwnNamespace] install modes")
 	}
 
+	if ext.Spec.Namespace == "" && !r.IsNamespaceManagementEnabled {
```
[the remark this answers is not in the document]
nader-ziada: low chance of happening and i don't think i want to change the api to fix this at this point

api/v1/clusterextension_types.go
joelanford: Validation rules do not need to be shared though. We can have separate sets of validation for standard vs experimental.

internal/operator-controller/applier/provider.go
joelanford: I'm asking from the perspective of the library, not the perspective of the caller. We should try to make the library itself cohesive and ergonomic (which would imply that options are in fact optional).

My suggestion here would be to remove `render.RenderInstallNamespace()` entirely. That would become the default. And then if `WithInstallNamespace` is passed as an option, then we get the existing self-managed behavior.

And with that change, I'd suggest `WithSelfManagedInstallNamespace` as the name to help illustrate that the default behavior is _not_ a self-managed namespace.

api/v1/clusterextension_types.go
nader-ziada: The actual blocker is narrower: the DNS1123 matches rule contains double quotes, and the opcon tag parser (rule="([^"]*)") can't capture a quoted CEL rule, so that one rule has to stay a base marker both channels inherit. Two ways forward: 
(1) keep the base rules identical to main and make managed mode omit the field rather than set "" (CEL only runs on present fields, so standard stays byte-identical), or 
(2) extend the generator's regex to accept quoted CEL so we can split the DNS rule into separate standard/experimental opcon tags. 
Happy to go whichever way you prefer:
(1) is no generator change but rejects an explicit namespace: ""; 
(2) keeps "" valid at the cost of a small generator change.

internal/operator-controller/applier/provider.go
joelanford: Tracing the code.. Let's say we dropped this check and we did somehow have an empty namespace passed into this code without the `BoxcutterRuntime` feature gate enabled. We'd ultimately end up calling our rendering library in a way where the namespace is a managed object, which would violate the intent of the feature gate being disable.

So yeah, let's keep it!

api/v1/clusterextension_types.go
joelanford: Ah, a limitation in the generator! Let's do (2). That seems like something we'll need regardless.

internal/operator-controller/controllers/clusterextension_admission_test.go
joelanford: I don't think we'd strand anything though, would we? We'd regenerate the entire set of manifests and anything in the old install namespace would be re-created in the new install namespace. 

_Unmanaged_ objects in the old namespace, potentially related to the ClusterExtension, wouldn't move. But I think that is a general problem that we can't really solve anyway, especially when OLM will have support for helm charts that can essentially do _anything_ using their own config schema.

Ultimately, I guess the question is "if this field had started out as a field in `spec.config` and was not required due to SA existence, would we have made it immutable then?"

I think the answer is "probably not", but maybe we can dig up some of those discussions to help inform us.

As far as I can tell, this is not a PR-blocking discussion. It seems like this PR's immutability rules still only allow `foo` -> `foo` and `""` -> `""`, and disallow setting/unsetting/changing. Which is the same semantic we have today, just with the relaxing of allowing unset on creation.

internal/operator-controller/rukpak/render/registryv1/generators/generators_test.go
joelanford: From a compiler/library standpoint though, `InstallNamespaceTemplate.Metadata.Name` is a valid field (that will always be unset, it sounds like). So there's still some ambiguity in the actual types/fields even if we (right now) are being careful to ignore it. Mainly I'm worried about this being something that could be a point of confusion to our future selves.

Can we think of a way to eliminate the ambiguity at the struct level?

internal/operator-controller/applier/provider.go
nader-ziada: reworked it so the default (no option) is the OLM-managed namespace that emits its own Namespace object, and added WithSelfManagedInstallNamespace(ns) for the bring-your-own-namespace case. Dropped WithInstallNamespace/RenderInstallNamespace. The provider's if/else collapses to a single option, and the zero-option call now does the sensible thing.

api/v1/clusterextension_types.go
nader-ziada: Done. Extended the generator's CEL regex to accept quoted rules, then split the namespace validation into per-channel opcon tags. Standard now regenerates byte-identical to main (empty diff on both the standard CRD and standard manifests), and experimental carries the relaxed rules. Added generator test cases for quoted rules since we'll want that capability generally.

internal/operator-controller/rukpak/render/registryv1/generators/generators_test.go
nader-ziada: Agreed, that ambiguity is exactly why I dropped the typed *corev1.Namespace template. render.Options now carries InstallNamespace string plus InstallNamespaceLabels/InstallNamespaceAnnotations map[string]string, so there's no second .Metadata.Name field to be silently ignored. The struct expresses only what it actually uses.

api/v1/clusterextension_types.go
```diff
@@ -59,10 +60,31 @@ type ClusterExtensionSpec struct {
...
+	// and be no longer than 63 characters.
+	//
+	// [RFC 1123]: https://tools.ietf.org/html/rfc1123
+	// </opcon:experimental:description>
+	//
+	// <opcon:standard:validation:XValidation:rule="self == oldSelf",message="namespace is immutable">
+	// <opcon:standard:validation:XValidation:rule="self.matches("^[a-z0-9]([-a-z0-9]*[a-z0-9])?$")",message="namespace must be a valid DNS1123 label">
+	// <opcon:experimental:validation:XValidation:rule="self == '' || self.matches("^[a-z0-9]([-a-z0-9]*[a-z0-9])?$")",message="namespace must be a valid DNS1123 label">
+	// <opcon:experimental:validation:XValidation:rule="oldSelf == '' || self == oldSelf",message="namespace is immutable once set">
+	// <opcon:experimental:validation:XValidation:rule="oldSelf != '' || self == ''",message="namespace cannot be set after creation; mode is locked at creation time">
```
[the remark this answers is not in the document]
nader-ziada: the experimental rules block add-after-create, remove, and value changes, and TestClusterExtensionAdmissionNamespaceImmutability has API-server test cases for all of them.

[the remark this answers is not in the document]
nader-ziada: fixed it

tmshort:
Force triggering of CI?

tmshort:
Seems like the CI is wedged. I can't trigger it via close/open, so it may require a (small) code change, or a change to the commit description such that a new SHA is generated.

nader-ziada:
> Seems like the CI is wedged. I can't trigger it via close/open, so it may require a (small) code change, or a change to the commit description such that a new SHA is generated.

updated the last commit message to generate a new sha

api/v1/clusterextension_types.go
```diff
@@ -49,6 +49,7 @@ const (
 
 // ClusterExtensionSpec defines the desired state of ClusterExtension
 type ClusterExtensionSpec struct {
+	// <opcon:standard:description>
```
tmshort, line 52: The API checker is complaining about this:
```
api/v1/clusterextension_types.go:52:2:kubeapilinter:commentstart: godoc for field ClusterExtensionSpec.Namespace should start with 'namespace ...'
```

But I think it's "ok" since the field has been moved back to experimental.

tmshort:
@joelanford; have all your concerns been addressed?

internal/operator-controller/rukpak/render/namespace.go
```diff
@@ -0,0 +1,168 @@
...
+	switch {
+	case template != nil && template.Name != "":
+		name = template.Name
+	case csvAnnotations[AnnotationSuggestedNamespace] != "":
+		name = csvAnnotations[AnnotationSuggestedNamespace]
+	default:
+		// The auto-derived default must always be a valid namespace, even for package names
+		// with disallowed characters (e.g. dots) or names that are too long.
+		name = defaultInstallNamespace(rv1.PackageName)
+	}
```
tmshort, line 64: When an extension is installed without spec.namespace, this resolves the namespace from each newly selected bundle. If a later bundle changes its suggested namespace or adds a template, the upgrade moves all namespaced resources and archives the old revision, which can delete the previous managed namespace and its contents.

Do we want to persist or otherwise retain the initially resolved namespace for the ClusterExtension lifecycle? Is it acceptable to change the namespace?

internal/operator-controller/rukpak/render/namespace.go
```diff
@@ -0,0 +1,170 @@
...
+// and enforces the namespace length limit. When the normalized name must be truncated, a short
+// hash of the original package name is appended to preserve deterministic collision resistance.
+func defaultInstallNamespace(packageName string) string {
+	const suffix = "system"
+
+	base := sanitizeDNS1123Label(packageName)
+
+	// Fast path: an already-valid, short base keeps the historical "<package>-system" name.
+	if base != "" && len(base)+1+len(suffix) <= maxNamespaceNameLength {
+		return base + "-" + suffix
```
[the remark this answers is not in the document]
tmshort: This needs to be addressed.

tmshort:
This comment is hiding inside a CodeRabbit comment, and ought to be addressed:
https://github.com/operator-framework/operator-controller/pull/2825/changes#r3982872948

internal/operator-controller/rukpak/render/namespace.go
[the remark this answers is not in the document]
nader-ziada: applied fix

internal/operator-controller/rukpak/render/namespace.go
nader-ziada: Yes, it is not acceptable to change the namespace. It's new to the managed-namespace path (before this PR spec.namespace was required and immutable).

No persistence needed though: since we should reject the change rather than follow it, we only need to detect it, and the namespace is already discoverable via the owner-kind/owner-name labels OLM puts on every rendered object. So `ValidateInstallNamespace` now fills its empty `spec.namespace == ""` branch: find the owned namespace, resolve from the incoming bundle, terminal error if they differ.

nader-ziada: fixed in the last commit

joelanford: I'm not sure I agree with this. If the install namespace is system managed, what's wrong with the bundle author having an opinion about moving it to a different namespace?

Another perspective: the spec is the spec is the spec. What happened in the past (e.g. an old bundle was installed in namespace `foo`) has no bearing on what the spec directly or indirectly (via a newly resolved bundle) says should happen in the future.

We should not use current state as input to future state.

nader-ziada:
Splitting this into three smaller PRs since it got too large to review.

 - #2925: rukpak/render only: namespace resolution from bundle metadata and the Namespace generator. No behavior change.
 - #2926  applier + controller: the BoxcutterRuntime gating and managed-namespace runtime handling.
 - #2929  API + docs + e2e: making spec.namespace optional.

Keeping this open for now for reference, but will be closed

PR #2838, opened 2026-07-31
:bug: Fix flaky generate-demos CI job

pedjak:
# Description

The `generate-demos` CI job fails ~45% of the time (9 of last 20 runs) with `jq` exit status 5 on the `ClusterCatalog Quickstart` scenario. The failure is always at the `catalog "operatorhubio" contains some packages` step.

Several issues contribute to the flakiness:

1. **`jq -s` (slurp mode) buffers the entire operatorhubio FBC response in memory** before processing, risking system errors (exit code 5) on the large operatorhubio catalog
2. **Catalog content queries run exactly once with no retry**, so any transient port-forward drop or network hiccup fails the step immediately
3. **`bash()` does not attach stderr to `ExitError`**, making failures opaque in CI logs
4. **With CatalogdHA, `kubectl port-forward` to the service deterministically picks the same pod** via `GetFirstPod` sorting; if that pod is not the leader, it returns 404 (empty local cache) for every retry — non-leader pods serve HTTP but only the leader reconciler downloads catalog content

This PR:
- Removes `jq -s` (slurp mode) so each JSON object is processed in constant memory via streaming, prefixing filters with `objects` to skip non-object values in the FBC stream
- Wraps `CatalogContainsSomePackages`, `PackageHasSomeChannels`, and `PackageHasSomeBundles` in `waitFor` for retry on transient errors (matching the pattern used by `PodHasContainerCount` and dozens of other steps)
- Adds `curl --compressed` to request gzip and stream-decompress (catalogd uses gzhttp), `--fail` with `set -o pipefail` to detect HTTP errors through the pipe
- Resolves the catalogd leader pod via the `catalogd-operator-lock` Lease and port-forwards directly to it on the container port (8443), falling back to the service when the lease cannot be read
- Resets port-forwards on query failure and re-establishes dead ones via liveness checks
- Injects stderr into `ExitError` in `bash()` to match the `k8sClient` diagnostics pattern
- Logs catalog query errors at V(0) so CI timeout failures are diagnosable

Tested with 10 consecutive CI passes after the fix (previously ~45% failure rate).

## Reviewer Checklist

- [ ] API Go Documentation

### DIFF
--- a/test/e2e/steps/demo_steps.go
+++ b/test/e2e/steps/demo_steps.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"crypto/tls"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net"
 	"net/http"
@@ -41,6 +42,10 @@ func bash(ctx context.Context, script string) (string, error) {
 
 	if err != nil {
 		logger.V(1).Info("Failed to run", "command", script, "stderr", stderr, "error", err)
+		var exitErr *exec.ExitError
+		if errors.As(err, &exitErr) {
+			exitErr.Stderr = stderrBuf.Bytes()
+		}
 	}
 	logger.V(1).Info("Output", "command", script, "output", stdout)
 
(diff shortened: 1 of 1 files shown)


test/e2e/steps/demo_steps.go
```diff
@@ -115,45 +120,45 @@ func catalogCurlJq(ctx context.Context, catalogName, jqFilter string) (string, e
...
-		return fmt.Errorf("catalog %q contains no packages", catalogName)
-	}
+	waitFor(ctx, func() bool {
+		out, err := catalogCurlJq(ctx, catalogName,
+			`select(.schema == "olm.package") | .name`)
+		if err != nil {
+			return false
+		}
+		return strings.TrimSpace(out) != ""
+	})
```
tmshort, line 199: These changed functions will no longer return an error. Should the context be checked for a timeout so that an error _can_ be returned?

test/e2e/steps/demo_steps.go
```diff
@@ -115,45 +120,45 @@ func catalogCurlJq(ctx context.Context, catalogName, jqFilter string) (string, e
 		return "", err
 	}
 	script := fmt.Sprintf(
-		`curl -s -k https://%s/catalogs/%s/api/v1/all | jq -s '%s'`,
+		`curl -s -k --retry 3 --retry-delay 2 --retry-all-errors https://%s/catalogs/%s/api/v1/all | jq '%s'`,
 		addr, catalogName, jqFilter,
 	)
```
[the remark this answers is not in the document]
pedjak: Good catch — updated to `curl -sS` so transport/TLS errors are surfaced on stderr while keeping normal output quiet. This pairs well with the new stderr-to-ExitError propagation in `bash()`.

test/e2e/steps/demo_steps.go
pedjak: The `waitFor` function uses `require.Eventually(godog.T(ctx), ..., timeout, tick)` which calls `t.FailNow()` on timeout — the function never returns on failure, it hard-fails the test via `runtime.Goexit()`. This matches the established pattern used by `PodHasContainerCount` in the same file (line 209) and dozens of other step functions throughout the test suite. So from a test perspective, the behavior is: either the condition is met within 5 minutes, or the scenario fails immediately.

tmshort:
Not a good sign when the update-demos is still failing with this.

test/e2e/steps/demo_steps.go
```diff
@@ -115,45 +120,45 @@ func catalogCurlJq(ctx context.Context, catalogName, jqFilter string) (string, e
 		return "", err
 	}
 	script := fmt.Sprintf(
-		`curl -s -k https://%s/catalogs/%s/api/v1/all | jq -s '%s'`,
```
grokspawn, line 118: We _need_ `jq -s` because the catalog format is in JSONLines (jsonl) format, not monolithic JSON, and slurp-mode forces it to re-represent it inside a monolithic object.  Without this, jq will error.

grokspawn:
Thanks for looking into this @pedjak!
Looking at the failure logs, it also looks like we still generate a demo 'cast even if the test failed.  I haven't investigated to see if the demo is successful in this case, but it casts doubt.

test/e2e/steps/demo_steps.go
pedjak: jq handles JSON Lines (JSONL/NDJSON) natively — each JSON object on a separate line is read as a separate input and the filter is applied to each one independently. No `-s` needed:

```bash
$ echo '{"schema":"olm.package","name":"foo"}
{"schema":"olm.channel","name":"bar"}' | jq 'select(.schema == "olm.package") | .name'
"foo"
```

The filters were adjusted to remove the `.[] |` prefix that was only needed in slurp mode (where jq wraps all inputs into a single array). Without `-s`, jq iterates over inputs automatically, so `select(...)` applies directly to each object.

The benefit of dropping `-s` is constant-memory processing — jq processes each JSON object as it arrives from the pipe instead of buffering the entire catalog (~100+ MB for operatorhubio) into memory before processing. The original `exit status 5` (jq system error) was likely caused by this memory pressure. The full test suite (all 4 scenarios, 30 steps) passes locally with this change.

tmshort:
I do question the need for the demo stuff. Is there really any reason to keep it?

pedjak:
> I do question the need for the demo stuff. Is there really any reason to keep it?

According to the discussion we had on CCSIG meeting on June 17 we wanted to keep them (and fix them). Let's merge this and as folllowup on the upcoming CCSIG we can chat about the possible removal.

rashmigottipati (approved):

PR #2848, opened 2026-08-04
🐛 fix: collect operator pod logs across restarts and rolling updates

pedjak:
# Description

The experimental-e2e CI job runs parallel tests then serial tests on the same kind cluster. Serial tests (HA, TLS, proxy scenarios) patch operator-controller and catalogd Deployments, triggering rolling updates that delete the original pods. The support bundle is collected only at the end (on failure), so it only captures replacement pod logs — parallel test failure logs are lost.

Since Kubernetes does not support retaining logs from deleted pods ([kubernetes#123279](https://github.com/kubernetes/kubernetes/issues/123279)), the only way to preserve them is to collect logs before the pods are removed.

This PR deploys Fluent Bit (gated on `ARTIFACT_PATH`) to continuously tail `olmv1-system` container logs to a persistent directory on the node filesystem. A `copyFromHost` collector in the support bundle spec extracts these persisted logs into the `operator-logs/` directory, preserving the complete log history across all pod generations including deleted and restarted instances.

## Changes

- **`testdata/fluentbit/values.yaml`** — Minimal Fluent Bit Helm values: tails only `olmv1-system` pod logs, writes to `/var/log/fluentbit/output/<namespace>/<pod>.<container>`
- **`Makefile`** — New `fluentbit-%` target (parallel with `prometheus-%`), added as dependency of `e2e-run-%`
- **`test/e2e/support-bundle.yaml`** — Added `copyFromHost` collector to extract persisted logs from nodes

## Reviewer Checklist

- [x] API Go Documentation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -383,6 +383,18 @@ wait-%: lint-deployed-%
 	KUBECONFIG=$(E2E_KUBECONFIG) kubectl wait --for=condition=Available --namespace=$(CATD_NAMESPACE) deployment/catalogd-controller-manager --timeout=60s
 	KUBECONFIG=$(E2E_KUBECONFIG) kubectl wait --for=condition=Ready --namespace=$(CATD_NAMESPACE) certificate/catalogd-service-cert
 
+.PHONY: fluentbit-%
+fluentbit-%: wait-% $(HELM)
+ifeq ($(strip $(ARTIFACT_PATH)),)
+	@echo "ARTIFACT_PATH unset; skipping fluent-bit deployment"
+else
+	KUBECONFIG=$(E2E_KUBECONFIG) $(HELM) upgrade --install fluent-bit oci://ghcr.io/fluent/helm-charts/fluent-bit \
+	  --namespace $(FLUENTBIT_NAMESPACE) --create-namespace \
+	  --version $(FLUENTBIT_CHART_VERSION) \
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -383,6 +383,18 @@ wait-%: lint-deployed-%
 	KUBECONFIG=$(E2E_KUBECONFIG) kubectl wait --for=condition=Available --namespace=$(CATD_NAMESPACE) deployment/catalogd-controller-manager --timeout=60s
 	KUBECONFIG=$(E2E_KUBECONFIG) kubectl wait --for=condition=Ready --namespace=$(CATD_NAMESPACE) certificate/catalogd-service-cert
 
+.PHONY: fluentbit-%
+fluentbit-%: wait-% $(HELM)
```
tmshort, line 387: Both `prometheus-%` and `fluentbit-%` have a dependency on `wait-%`. The `wait-%` recipe will only be run once. This should still work, but I want to ensure that you aren't expecting `wait-%` to occur twice.

pedjak: Correct — `wait-%` runs once, then both `prometheus-%` and `fluentbit-%` can proceed. That's intentional: they share the same prerequisite but don't depend on each other, so with `make -j` they install in parallel.

rashmigottipati (approved):
solid improvement. thanks @pedjak 

/lgtm

PR #2885, opened 2026-08-21
:seedling: OPRUN-4723: registry+v1: add APIService renderer support

tmshort:
## Summary

Adds the rendering infrastructure for APIService objects to the registry+v1 bundle renderer.
The implementation is **complete but not yet activated** — `BundleCSVAPIServiceGenerator` and
`CheckAPIServiceDeploymentReferentialIntegrity` are **not registered** in `ResourceGenerators`
or `BundleValidator` pending end-to-end Boxcutter path validation. The code is retained as
infrastructure for a future release.

The OLMv0→OLMv1 migration tool ([library-olm](https://github.com/operator-framework/library-olm))
continues to hard-block operators with owned APIService definitions (C3) until this is fully
wired and confirmed working end-to-end.

---

## Changes

**`generators/generators.go`**
- `BundleCSVAPIServiceGenerator` — generates per owned APIService:
  - `APIService` object (`GroupPriorityMinimum=2000`, `VersionPriority=15`, service reference,
    CA bundle via cert provider) — validates port is in `[1, 65535]`
  - `ClusterRoleBinding <service>-system:auth-delegator` — delegates TokenReview/SubjectAccessReview
    to the extension API server (required for aggregation auth)
  - `RoleBinding <service>-auth-reader` in `kube-system` — allows reading
    `extension-apiserver-authentication` ConfigMap (required for client CA config)
  - RBAC deduplicated per deployment (multiple group+versions sharing one deployment get one set)
  - APIService descriptions deduplicated by group+version via `GetOwnedAPIServiceDescriptions()`
- `BundleCSVDeploymentGenerator` extended to inject `apiservice-cert` volumes into deployments
  serving APIServices
- `BundleDeploymentServiceResourceGenerator` extended to create Services for APIService-serving
  deployments, with port conflict detection vs webhook ports
- `resolveAPIServicePort` — shared helper enforcing valid port range `[1, 65535]`

**`validators/validator.go`**
- `CheckAPIServiceDeploymentReferentialIntegrity` — validates each owned APIService has a
  non-empty `deploymentName` that references an existing deployment in the install spec
  (intentionally uses raw `Owned` slice to fail-closed on every declared entry, including
  duplicates)

**`certproviders/certmanager.go` + `openshift_serviceca.go`**
- Added `*apiregistrationv1.APIService` case to `InjectCABundle` so both cert providers
  annotate APIService objects for CA bundle injection

**`registryv1/registryv1.go`**
- `BundleCSVAPIServiceGenerator` and `CheckAPIServiceDeploymentReferentialIntegrity` are
  implemented but **intentionally NOT registered** here. Comments explain why.

**`go.mod` / `go.sum`**
- `k8s.io/kube-aggregator` promoted to direct dependency

## Test plan

- [x] `go build ./internal/operator-controller/rukpak/...` passes
- [x] `go test ./internal/operator-controller/rukpak/...` — all pass
  - `BundleCSVAPIServiceGenerator`: nil bundle, no owned, success (APIService + RBAC), default
    port, invalid port (-1, 65536), dedup by group+version, RBAC dedup per deployment
  - `CheckAPIServiceDeploymentReferentialIntegrity`: no owned, empty deploymentName, unknown
    deploymentName, valid deploymentName
  - `CertProviderResourceGenerator`: extended tests for APIService-serving deployments
  - Enumeration tests updated to reflect the unregistered status
- [ ] E2E: install an operator with owned APIServices via OLMv1 — deferred until registered

## Related

- Migration tool C3 block restoration: [library-olm PR #22](https://github.com/operator-framework/library-olm/pull/22)


internal/operator-controller/rukpak/render/registryv1/generators/generators.go
```diff
@@ -414,6 +424,113 @@ func BundleMutatingWebhookResourceGenerator(rv1 *bundle.RegistryV1, opts render.
...
+				Service: &apiregistrationv1.ServiceReference{
+					Namespace: opts.InstallNamespace,
+					Name:      certProvisioner.ServiceName,
+					Port:      &containerPort,
+				},
+				InsecureSkipTLSVerify: false,
+			},
+		}
+
+		if err := certProvisioner.InjectCABundle(apiService); err != nil {
```
pedjak, line 484: **issue (blocking):** `CertProviderResourceGenerator` (line 621) only generates Issuer/Certificate resources for deployments that serve webhooks. When an operator has owned APIService definitions backed by a deployment with _no_ webhooks, no cert-manager resources will be created — yet this line annotates the APIService with `cert-manager.io/inject-ca-from` referencing that non-existent Certificate.

Result: cert-manager has nothing to inject, the APIService has no CA bundle, `InsecureSkipTLSVerify` is false, and API aggregation fails with TLS errors.

The fix is to also include APIService deployment names in `CertProviderResourceGenerator`:

```go
deploymentsNeedingCerts := sets.Set[string]{}
for _, wh := range rv1.CSV.Spec.WebhookDefinitions {
    deploymentsNeedingCerts.Insert(wh.DeploymentName)
}
for _, desc := range rv1.CSV.GetOwnedAPIServiceDescriptions() {
    if desc.DeploymentName != "" {
        deploymentsNeedingCerts.Insert(desc.DeploymentName)
    }
}
```

(The downstream openshift-service-ca path is unaffected since its `AdditionalObjects()` is a no-op.)

internal/operator-controller/rukpak/render/registryv1/generators/generators.go
```diff
@@ -422,14 +539,44 @@ func BundleDeploymentServiceResourceGenerator(rv1 *bundle.RegistryV1, opts rende
 		return nil, fmt.Errorf("bundle cannot be nil")
 	}
 
-	// collect webhook service ports
+	// collect service ports from webhooks and owned APIService definitions
 	webhookServicePortsByDeployment := map[string]sets.Set[corev1.ServicePort]{}
```
pedjak: **nitpick:** This variable now holds ports from both webhooks and APIService definitions but the name still says `webhook`. Consider renaming to `servicePortsByDeployment` to match the updated comment on line 541.

internal/operator-controller/rukpak/render/registryv1/validators/validator.go
```diff
@@ -362,3 +362,30 @@ func CheckObjectSupport(rv1 *bundle.RegistryV1) []error {
...
+// references a deployment present in the CSV's install spec. An empty deploymentName
+// is rejected because the APIService would have no backend to serve requests.
+func CheckAPIServiceDeploymentReferentialIntegrity(rv1 *bundle.RegistryV1) []error {
+	deploymentNames := sets.New[string]()
+	for _, dep := range rv1.CSV.Spec.InstallStrategy.StrategySpec.DeploymentSpecs {
+		deploymentNames.Insert(dep.Name)
+	}
+
+	var errs []error
+	for _, desc := range rv1.CSV.Spec.APIServiceDefinitions.Owned {
```
pedjak, line 381: **thought (non-blocking):** This iterates `rv1.CSV.Spec.APIServiceDefinitions.Owned` (raw, with potential duplicates) while all generators use `rv1.CSV.GetOwnedAPIServiceDescriptions()` (deduplicated, sorted). The choice seems intentionally fail-closed — validate every declared entry, even duplicates — but a brief comment noting why the raw field is used here (vs. the deduplicated method elsewhere) would help future readers.

tmshort:
/hold
Need to reduce scope

internal/operator-controller/rukpak/render/registryv1/generators/generators.go
```diff
@@ -422,22 +539,52 @@ func BundleDeploymentServiceResourceGenerator(rv1 *bundle.RegistryV1, opts rende
...
-			webhookServicePortsByDeployment[wh.DeploymentName] = sets.Set[corev1.ServicePort]{}
+		if _, ok := servicePortsByDeployment[wh.DeploymentName]; !ok {
+			servicePortsByDeployment[wh.DeploymentName] = sets.Set[corev1.ServicePort]{}
 		}
-		webhookServicePortsByDeployment[wh.DeploymentName].Insert(getWebhookServicePort(wh))
+		servicePortsByDeployment[wh.DeploymentName].Insert(getWebhookServicePort(wh))
+	}
+	// GetOwnedAPIServiceDescriptions() deduplicates by group+version identity.
+	for _, desc := range rv1.CSV.GetOwnedAPIServiceDescriptions() {
+		if desc.DeploymentName == "" {
```
perdasilva, line 562: Could it be a problem that here we are skipping descriptions without a set Deployment name but in `BundleDeploymentServiceResourceGenerator` (line 454) we're not?

internal/operator-controller/rukpak/render/registryv1/generators/generators.go
```diff
@@ -414,6 +424,113 @@ func BundleMutatingWebhookResourceGenerator(rv1 *bundle.RegistryV1, opts render.
...
+			return nil, err
+		}
+		objs = append(objs, apiService)
+
+		// Emit RBAC once per deployment — multiple group+versions may share one deployment,
+		// and creating duplicate ClusterRoleBinding/RoleBinding names would be rejected by k8s.
+		if !generatedRBACForDeployment.Has(desc.DeploymentName) {
+			generatedRBACForDeployment.Insert(desc.DeploymentName)
+
+			saName := saNameOrDefault(depSAName[desc.DeploymentName])
```
perdasilva: is it a worry that we're binding system:auth-delegator to default if the APIService doesn't refer to a defined deployment?

internal/operator-controller/rukpak/render/registryv1/generators/generators.go
```diff
@@ -422,22 +539,52 @@ func BundleDeploymentServiceResourceGenerator(rv1 *bundle.RegistryV1, opts rende
...
+			Port:       port,
+			TargetPort: intstr.FromInt32(port),
+		}
+		// Detect conflicts: if the deployment already has a port with the same port
+		// number but different configuration (e.g., a webhook with a different TargetPort),
+		// reject rather than generate an invalid Service with duplicate port numbers.
+		for existing := range servicePortsByDeployment[desc.DeploymentName] {
+			if existing.Port == port && existing != apiSvcPort {
+				return nil, fmt.Errorf(
+					"deployment %q has a Service port conflict: APIService port %d conflicts with an existing webhook port configuration",
```
perdasilva: it could be conflicting with an existing APIService port too, no?


```suggestion
					"deployment %q has a Service port conflict: APIService port %d conflicts with an existing webhook or APIService port configuration",
```

internal/operator-controller/rukpak/render/registryv1/registryv1.go
```diff
@@ -47,5 +50,8 @@ var ResourceGenerators = []render.ResourceGenerator{
 	generators.BundleValidatingWebhookResourceGenerator,
 	generators.BundleMutatingWebhookResourceGenerator,
 	generators.BundleDeploymentServiceResourceGenerator,
+	// NOTE: BundleCSVAPIServiceGenerator is implemented but NOT registered here.
```
perdasilva: I think its fine to enable the new generators and validators - we just need to make sure the applier manifest provider bounces bundles with APIServices

tmshort: @joelanford suggested we just add the code, and not actually support it yet.

internal/operator-controller/rukpak/render/registryv1/generators/generators.go
tmshort: We skip on line 454 now.

internal/operator-controller/rukpak/render/registryv1/generators/generators.go
tmshort: We now error on this case.

internal/operator-controller/rukpak/render/registryv1/registryv1.go
perdasilva: I think it should be ok to support the behavior at this level - we are gating [at the provider level](https://github.com/operator-framework/operator-controller/blob/main/internal/operator-controller/applier/provider.go#L45)

perdasilva: @joelanford you good with this?

joelanford: Yeah, +1 on hooking it up within the rukpak tree.

PR #1418, opened 2024-10-30
✨ WIP: DO NOT MERGE: Shift Week Prototypes

joelanford:
# Description
- Direct bundle install
- Parameter passing with schema validation in ClusterExtension
- Parameterizing registry+v1 bundles by converting to helm charts with values.schema.json
- CLI to convert registry+v1 bundle to helm chart
- CLI to convert helm chart repo to an FBC catalog
- Support for bundles that require webhooks
- Incorporate new conversion and plumbing of values into the operator-controller's Helm applier that is used to install content.

## Reviewer Checklist

- [ ] API Go Documentation
- [ ] Tests: Unit Tests (and E2E Tests, if appropriate)
- [ ] Comprehensive Commit Messages
- [ ] Links to related GitHub Issue(s)

Issue #597, opened 2024-01-24
[epic] Support direct bundle installation


ncdc:
After #596 goes in, we can implement support for installing a bundle from a URL instead of resolving it from a catalog. To do this, we'll have a very strict set of guardrails:

1. The bundle must be a registry+v1 bundle
2. We will do minimal or no constraint/dependency checking - a direct bundle installation is considered "satisfied"

ncdc:
FYI @varshaprasad96

grokspawn:
A candidate brief started (but not reviewable) [here](https://docs.google.com/document/d/1UEoHNgMnONYahJ2_TOFaCKn04lmEUXUXKY0y_Oy_UYs/)

joelanford:
I implemented a proof of concept as part of https://github.com/operator-framework/operator-controller/pull/1418. We should revisit that PR when we have time to pick this issue up.

ankitathomas:
[RFC: Direct Bundle Installation](https://docs.google.com/document/d/1fNeEpixSX_D3IHjl-ewb_W79Il4D0eikMAWJupXkmc8/edit?tab=t.0)

We have an RFC tracking this work under review

PR #2907, opened 2026-09-04
✨ implementation of a registry+v1 direct bundle installer

grokspawn:
# Description
An implementation of a direct bundle install capability to Operator Controller, currently only supporting registry+v1 bundles.
This implementation adds an OCI image resolver to existing resolver architecture to handle and validate direct bundle attempts for registry+v1 bundles, bypassing catalog resolution phases.  
~~This is intended as a basis for doing additional type sniffing for other content types in the future, for e.g. helm charts.~~

solves #2686 and non-docs portions of epic #597 

This is based off predecessor proof of concept implementations 
- https://github.com/operator-framework/operator-controller/pull/1418/changes#diff-387659018fbee6e27fcf5ab561b45489e14b62ae73ecf00fed3a0f4c017714d1
- https://github.com/ankitathomas/operator-controller/tree/POC_direct_bundle_install
- https://github.com/perdasilva/operator-controller/tree/direct-bundle-install

and the RFC at https://docs.google.com/document/d/1fNeEpixSX_D3IHjl-ewb_W79Il4D0eikMAWJupXkmc8/

## Reviewer Checklist

- [ ] API Go Documentation

## Summary by CodeRabbit

* **New Features**
  * ClusterExtensions can be installed directly from OCI image bundles using an image reference.
  * Direct OCI image sources are available with the Boxcutter runtime and skip catalog-based dependency resolution.
* **Bug Fixes**
  * Improved validation for image references and OCI bundle package metadata.
  * OCI image sources now require image details and reject conflicting catalog configuration.
* **Documentation**
  * Updated API documentation and resource descriptions to explain OCI image installation support and runtime requirements.


api/v1/clusterextension_types.go
```diff
@@ -122,23 +121,30 @@ type ClusterExtensionSpec struct {
...
+	// When set to "OCIImage", the bundle image is used directly. Direct sources do not perform
+	// dependency resolution and are only supported by the Boxcutter runtime.
 	//
 	// When set to "Catalog", information for determining the appropriate bundle of content to install
 	// is fetched from ClusterCatalog resources on the cluster.
 	// When using the Catalog sourceType, the catalog field must also be set.
 	//
 	// +unionDiscriminator
-	// +kubebuilder:validation:Enum:="Catalog"
+	// +kubebuilder:validation:Enum:="Catalog";"OCIImage"
```
joelanford, line 147: Separate enum validation and the type addition between standard and experimental, such that only experimental adds the support for the new enum/type since this appears to be behind the experimental only `BoxcutterRuntime` feature gate?

internal/operator-controller/controllers/clusterextension_reconcile_steps.go
```diff
@@ -146,6 +159,27 @@ func ResolveBundle(r resolve.Resolver, c client.Client) ReconcileStepFunc {
 			return nil, nil
 		}
 
+		// Direct OCIImage sources have no catalog metadata, so resolve them
+		// without running catalog fallback or deprecation handling.
+		if ext.Spec.Source.SourceType == ocv1.SourceTypeOCIImage {
```
joelanford, line 164: This seems to at least partially defeat the purpose of the `MultiResolver`. Is it not possible to leave this function untouched and let the existing (theoretically resolver-implementation-agnostic) logic just do the general thing with whatever the resolver returns?

If not, it seems like we have a bad resolver abstraction, and maybe we should make a change there?

internal/operator-controller/controllers/clusterextension_reconcile_steps.go
joelanford, line 196: Replace with a call to the new `installedBundleName` helper?

internal/operator-controller/resolve/ociimage.go
```diff
@@ -0,0 +1,118 @@
...
+	Puller    imageutil.Puller
+	Cache     imageutil.Cache
+	Detectors []BundleContentDetector
+}
+
+// BundleContentDetector identifies and loads a supported bundle format from
+// already-unpacked image content.
+type BundleContentDetector interface {
+	Detect(fs.FS, string) (*declcfg.Bundle, error)
+}
```
joelanford, line 33: A couple of notes here: 
1. We'll end up needing more than an fs.FS-based detector. The thing that pulls the image from the registry will need to be able to unpack a ref, and not all refs are standard container images.
2. The thing that actually assesses the OCI layers will very likely be the thing that can do the detection, well before anything exists on disk.

I have another iteration of the image pulling stuff here: https://github.com/joelanford/library-olm/tree/main/image

The idea is that you register a client with a bunch of `Handler`s and then start throwing refs at it. A `Handler` does two things:
1. Given an OCI descriptor and a repository client (in case it needs to pull more stuff), detect whether the given OCI artifact matches the kind the handler supports
2. Given an OCI descriptor and a repository client, actually unpack whatever is there into a directory on disk, with whatever logic the handler wants (i.e. no need to pull all layers if not all layers are relevant)


For now, maybe we drop the detector stuff and just continue assuming registry+v1 only? When we add Helm support, I'm keen on getting this library in place. The library will make it easy for us to add support for more formats in the future because it is just "implement handler and register it".

internal/operator-controller/resolve/ociimage.go
```diff
@@ -0,0 +1,118 @@
...
+	propertiesJSON := registryBundle.CSV.Annotations[bundlesource.PropertyOLMProperties]
+	if propertiesJSON == "" {
+		return nil, fmt.Errorf("bundle %q has no %q package property", bundle.Name, bundlesource.PropertyOLMProperties)
+	}
+	if err := json.Unmarshal([]byte(propertiesJSON), &bundle.Properties); err != nil {
+		return nil, fmt.Errorf("failed to parse bundle properties: %w", err)
+	}
+	if !hasPackageProperty(bundle.Properties) {
+		return nil, fmt.Errorf("bundle %q has no package property", bundle.Name)
+	}
```
joelanford, line 107: What's this part doing? Seems like we already have a package name from `registryBundle.PackageName`?

joelanford: Seems like it should be a concern of the bundle parser to validate this?

internal/operator-controller/resolve/ociimage.go
grokspawn: Dropped.  I didn't want to do a full registry but I _did_ want to jump-start some conversations on the topic early.  We can continue to discuss.

api/v1/clusterextension_types.go
grokspawn: Done, with mods.  Here it's clean, but in the descriptions we have mutually-exclusive info, so we have to wrap the old in standard tags and the new stuff in experimental tags.

api/v1/clusterextension_types.go
```diff
@@ -147,6 +164,26 @@ type SourceConfig struct {
...
 	// +optional
 	Catalog *CatalogFilter `json:"catalog,omitempty"`
+
+	// ociImage configures a bundle image to install directly.
+	// <opcon:experimental:description>
+	// They do not provide catalog dependency resolution or upgrade safety.
+	// </opcon:experimental:description>
+	// <opcon:experimental>
+	// +optional
+	OCIImage OCIImageSource `json:"ociImage,omitzero"`
```
joelanford, line 174: Should we make this a pointer with `omitempty` to match `*CatalogFilter`. Then the XValidation check can also follow the same pattern?

I know `omitzero` makes it possible to have a non-pointer here, which is generally preffered, but I think consistency among the union members is probably more important than using the new Go 1.24+ features just for new union members.

Another thing _I think_ we are free to do is change `CatalogFilter` to a non-pointer. I don't think that would break (de-)serialization, and our public API guarantee is our Kuberentes API, not our Go types.

api/v1/clusterextension_types.go
```diff
@@ -122,23 +121,41 @@ type ClusterExtensionSpec struct {
...
+const (
+	SourceTypeCatalog  = "Catalog"
+	SourceTypeOCIImage = "OCIImage"
+)
 
 // SourceConfig is a discriminated union which selects the installation source.
 //
 // +union
 // +kubebuilder:validation:XValidation:rule="has(self.sourceType) && self.sourceType == 'Catalog' ? has(self.catalog) : !has(self.catalog)",message="catalog is required when sourceType is Catalog, and forbidden otherwise"
+// <opcon:experimental:validation:XValidation:rule="has(self.sourceType) && self.sourceType == 'OCIImage' ? self.ociImage.ref.size() != 0 : self.ociImage.ref.size() == 0",message="ociImage is required when sourceType is OCIImage, and forbidden otherwise">
```
joelanford, line 133: If feels like `self.ociImage.ref.size() != 0` is misplaced in _this_ rule. I'd expect that on the validation of the OCIImage type itself. At this level, what we want to do conceptually is say "if `sourceType` is OCIImage, then `ociImage` needs to be set." Without a pointer, that means we need to look at the contents of `ociImage`. The best way to do that here would be: `self.ociImage.size() > 0`

api/v1/clusterextension_types.go
```diff
@@ -147,6 +164,26 @@ type SourceConfig struct {
...
+	// </opcon:experimental:description>
+	// <opcon:experimental>
+	// +optional
+	OCIImage OCIImageSource `json:"ociImage,omitzero"`
+}
+
+// OCIImageSource identifies a bundle image to install directly from an OCI registry.
+// +kubebuilder:validation:MinProperties:=1
+type OCIImageSource struct {
+	// ref is a Docker-style image reference with a tag or digest.
```
joelanford, line 180: If the ref uses a tag, what is our behavior when the tag is moved to a different digest in the image registry? And then the follow-up question would be: is that the behavior that users would expect/that we want to support?

We should document that behavior and test for it. Alternatively, we could require a digest at least to start. I know the UX of that is worse, but it is simpler for us to deal with and easier for readers of this API to reason about.

cmd/operator-controller/main.go
```diff
@@ -654,6 +663,8 @@ func (c *boxcutterReconcilerConfigurator) Configure(ceReconciler *controllers.Cl
 		controllers.HandleFinalizers(c.finalizers),
 		controllers.ValidateClusterExtension(
 			controllers.ServiceAccountDeprecationWarning(),
+			controllers.DirectBundleRequiresBoxcutter(),
```
joelanford, line 666: Nit: I wonder if we should combine the functionality of `DirectBundleRequiresBoxcutter` and `ValidateDirectBundleSource`? That way, when we drop the feature gate, we just remove that one conditional check from the function. That would keep this call site tidy (just one DBI check instead of two) and unchanged when the feature gate is removed.

api/v1/clusterextension_types.go
```diff
@@ -147,6 +164,26 @@ type SourceConfig struct {
...
+	//
+	// +required
+	// +kubebuilder:validation:MaxLength:=1000
+	// +kubebuilder:validation:MinLength:=1
+	// +kubebuilder:validation:XValidation:rule="self.matches(\"^[a-zA-Z0-9]([a-zA-Z0-9.-]*[a-zA-Z0-9])?(:[0-9]+)?/[a-z0-9]+([._-][a-z0-9]+)*(/[a-z0-9]+([._-][a-z0-9]+)*)*(:[A-Za-z0-9_][A-Za-z0-9_.-]{0,126}|@[A-Za-z][A-Za-z0-9+._-]*:[0-9A-Fa-f]{32,})$\")",message="must be a complete image reference with a valid repository and tag or digest"
```
joelanford, line 185: Just checking that this is the exact same validation that we are using for the format of the catalogd image source?

I know we have an extra check there for "digest disallowed with poll interval", but other than that, I'd expect to duplicate the validation.

internal/operator-controller/controllers/clusterextension_admission_test.go
```diff
@@ -74,6 +74,62 @@ func TestClusterExtensionSourceConfig(t *testing.T) {
...
 }
 
+func TestClusterExtensionOCIImageSourceConfig(t *testing.T) {
+	t.Parallel()
+	testCases := []struct {
+		name      string
+		source    ocv1.SourceConfig
+		wantError bool
+	}{
+		{
```
joelanford, line 84: Let's also check for supporting IPv4 and IPv6 addresses for the host?

internal/operator-controller/controllers/clusterextension_reconcile_steps.go
```diff
@@ -108,6 +109,35 @@ func ServiceAccountDeprecationWarning() ClusterExtensionValidator {
...
+	return func(_ context.Context, ext *ocv1.ClusterExtension) error {
+		if ext.Spec.Source.SourceType == ocv1.SourceTypeOCIImage && !features.OperatorControllerFeatureGate.Enabled(features.BoxcutterRuntime) {
+			return fmt.Errorf("sourceType %q requires the %s feature gate", ocv1.SourceTypeOCIImage, features.BoxcutterRuntime)
+		}
+		return nil
+	}
+}
+
+// ValidateDirectBundleSource validates the direct source fields that cannot be
+// expressed in the standard CRD because OCIImage is experimental-only.
```
joelanford, line 125: I'm confused by this GoDoc comment. What does standard vs experimental have to do with it? My mental model is:
1. CRD controls what field combinations are even possible (standard: not possible, experimental: possible)
2. Validations here are basically just sanity checks and aren't truly necessary if we've done our job correctly in describing this in the CRD validation. If the closure returns an error in production, that means we've screwed up our CRD validation or feature gate mapping.

internal/operator-controller/controllers/clusterextension_reconcile_steps.go
```diff
@@ -175,11 +197,19 @@ func ResolveBundle(r resolve.Resolver, c client.Client) ReconcileStepFunc {
 		//   the deprecation status to unknown? Or perhaps we somehow combine the deprecation information from
 		//   all catalogs? This needs a follow-up discussion and PR.
 		hasCatalogData := err == nil || resolvedDeprecation != nil
+		if behavior, ok := r.(resolve.ResolverBehavior); ok {
+			hasCatalogData = behavior.HasCatalogData(ext)
+		}
```
joelanford, line 202: Clearly our resolver API is falling over here, and is in need of some re-work. We should specifically talk about how to fix this, likely as part of our planned work on a centralized resolver service.

In the meantime, I think we can get by without changes in this area. With the current state of the art [1] for detecting bundle deprecations, we implicitly _know_ what the deprecations are: none. Which means for the bundle source, we always have definitive knowledge that `Deprecated=False` across the board. And that translates to "no deprecations, ever", which I think then translates to the DBI resolver returning a non-nil `resolvedDeprecation` where nothing is actually deprecated`.

Regardless, I don't think we should extend this broken resolver API further with new optional interface implementations and checking, as I think that would further entrench the idea and make it more difficult for our future selves to understand how to unwind it.

[1]: Deprecations are only conveyed in catalogs. In the future, if we have an OCI Referrers-based bundle deprecation mechanism, that might be possible

api/v1/clusterextension_types.go
grokspawn: I originally did, and coderabbit flagged for omitzero instead.  
Since there is not a significance between nil and unconfigured, omitzero makes sense here.  Otherwise we just introduce a bunch of nil checks for no benefit. 
We're already failing crdiff because of description delta, so we would have to override it anyway.  I think we might be able to go value for CatalogFilter+omitzero and achieve the same goals while being consistent in the overall approach.

api/v1/clusterextension_types.go
grokspawn: The check does this, but instead of ">0" it's "!=0".  This is just looking at the first field in the first field.

api/v1/clusterextension_types.go
grokspawn: This was a minimal lift-and-shift from earlier work.  I've adopted the more-robust catalogsource ref validation here now.

internal/operator-controller/controllers/clusterextension_reconcile_steps.go
grokspawn: Combined

api/v1/clusterextension_types.go
joelanford: > Otherwise we just introduce a bunch of nil checks for no benefit.

Ehh, if we have API level validation that says "if sourceType is OCIImage, then `ociImage` must be set", I feel like it is perfectly reasonable to skip `nil` checks if we're already switching on sourceType.

Also, I'd go out on a (short?) limb and say "coderabbit is wrong" to suggest that we follow a different type pattern among different union members. I feel like the only right answers are:
- all non-pointers with `omitzero`, OR
- all pointers with `omitempty`

I _think_ we'd see go-apidiff fail but no mention in crddiff (I don't think the CRD schema would actually change, but I might be wrong).

api/v1/clusterextension_types.go
joelanford: Yeah, this is mainly a separation of concerns comment.
1. At this level, it is not the concern of the discriminated union that there is a field called `ref` that is set. This level only cares that the field itself is non-zero.
2. At the level of "validate this struct when it is set", that's where we'd put validations related to the `ref` field.

PR #2925, opened 2026-09-16
✨ feat(render): resolve system-managed install namespace from bundle metadata

nader-ziada:
# Description
Derive the install namespace from the bundle's CSV rather than requiring the caller to supply one. Resolution order is the suggested-namespace-template name, then the suggested-namespace annotation, then <packageName>-system. Package names that are not usable DNS1123 labels are normalized and given a short hash suffix so distinct packages cannot collide on one namespace.

```
First of three PRs splitting #2825. This one is rukpak/render only. applier in PR 2, API/docs/e2e in PR 3.
```

summary:
 - New: render/namespace.go
Resolves an install namespace from the bundle's CSV. Order: suggested-namespace-template name → suggested-namespace → <packageName>-system. If the package name isn't a usable DNS1123 label (dots, too long), it's normalized and gets a short hash appended so two packages can't land on the same namespace. Also BuildNamespaceObject, which builds the Namespace with labels/annotations from the template and strips empty spec/status.
 
 - render/render.go
Three new Options fields (SelfManagedInstallNamespace, InstallNamespaceLabels, InstallNamespaceAnnotations) and a WithSelfManagedInstallNamespace(ns) option. Render() loses its positional installNamespace argument — callers pass the option instead. If the option isn't set, Render resolves the namespace itself.

 - generators/generators.go + registryv1.go
New BundleInstallNamespaceGenerator emits the Namespace object, registered first in the generator list. Returns nil when the caller used WithSelfManagedInstallNamespace.

 - applier/provider.go
Two lines: add the option, drop the positional arg. Always passes the option, so nothing changes.

 - test/regression/convert/generate-manifests.go
Same caller update. Namespaces there are hardcoded, so no change.

Net effect: the renderer can resolve and emit a namespace, but nothing asks it to yet.

## Reviewer Checklist

- [ ] API Go Documentation

## Summary by CodeRabbit

- **New Features**
  - Added automatic installation namespace management for bundles.
  - Namespace names can be derived from suggested metadata, with validated deterministic fallbacks.
  - Generated installation namespaces preserve configured labels and annotations.
  - Bundles can manage their own installation namespace; the renderer does not create one in this mode.
  - Namespaced resources are consistently assigned to the resolved installation namespace.
- **Bug Fixes**
  - Invalid or unsuitable namespace names are now rejected or normalized to safe deterministic alternatives.


internal/operator-controller/rukpak/render/namespace.go
```diff
@@ -0,0 +1,172 @@
...
+		return base + "-" + suffix
+	}
+
+	// Otherwise keep the name deterministic and collision-resistant: append a short hash of the
+	// original package name and truncate the base to fit within the length limit.
+	hash := hashutil.DeepHashObject(packageName)[:8]
+	maxBase := maxNamespaceNameLength - len(suffix) - len(hash) - 2 // account for two '-' separators
+	if len(base) > maxBase {
+		base = base[:maxBase]
+	}
```
joelanford, line 119: Could maxBase ever be less than zero here? Would that cause a panic?

internal/operator-controller/rukpak/render/namespace.go
```diff
@@ -0,0 +1,172 @@
...
+			Name: name,
+		},
+	}
+
+	if len(labels) > 0 {
+		ns.Labels = labels
+	}
+	if len(annotations) > 0 {
+		ns.Annotations = annotations
+	}
```
joelanford: Probably no need for the conditionals, right? We could directly assign:
```go 
ns := corev1.Namespace{
    Name: name,
    Labels: labels,
    Annotations: annotations,
  }
}
```

internal/operator-controller/rukpak/render/namespace.go
```diff
@@ -0,0 +1,172 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name: name,
+		},
+	}
+
+	if len(labels) > 0 {
+		ns.Labels = labels
+	}
+	if len(annotations) > 0 {
+		ns.Annotations = annotations
```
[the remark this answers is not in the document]
joelanford: Disagree. I think it is safe to delegate this validation to the apiserver when something ultimately tries to apply this to the cluster.

internal/operator-controller/rukpak/render/namespace.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	unstructuredObj, err := runtime.DefaultUnstructuredConverter.ToUnstructured(&ns)
+	if err != nil {
+		return nil, fmt.Errorf("failed to convert namespace to unstructured: %w", err)
+	}
+	delete(unstructuredObj, "status")
+	delete(unstructuredObj, "spec")
+
+	return &unstructured.Unstructured{Object: unstructuredObj}, nil
```
joelanford: Do we need to convert it to unstructured? `*corev1.Namespace` already implements `client.Object`.

nader-ziada: fixed

internal/operator-controller/rukpak/render/namespace.go
nader-ziada: fixed

internal/operator-controller/rukpak/render/namespace.go
nader-ziada: Made it an actual const (with hashLength)

perdasilva (approved):
nice! thank you ^^

PR #2926, opened 2026-09-16
✨ support a system-managed install namespace at runtime

nader-ziada:
# Description

When spec.namespace is empty the applier stops passing WithSelfManagedInstallNamespace, so the renderer resolves the install namespace from bundle metadata and emits the Namespace object itself. This is gated on BoxcutterRuntime; with the gate off an empty spec.namespace is a terminal configuration error rather than a silent fallback.

_second of three PRs splitting #2825. This one is applier.

validateSystemManagedNamespaceUnchanged from the original branch is deliberately omitted while the requirement is being discussed. It blocked an upgrade whose bundle resolves to a different namespace, by listing namespaces with the extension's owner labels and comparing._

summary
 - applier/provider.go
New IsNamespaceManagementEnabled field. If spec.namespace is empty and the gate is off, Get() returns a terminal config error. If spec.namespace is set, it passes WithSelfManagedInstallNamespace as before; if empty, it omits the option so the renderer resolves the namespace and emits the Namespace object.

 - applier/boxcutter.go
Migrate() returns early when spec.namespace is empty — a managed-namespace extension never had a Helm release, so there's nothing to migrate.

 - controllers/clusterextension_reconcile_steps.go
New ValidateInstallNamespace step. Checks a user-provided spec.namespace exists and fails retryably if not, so the user can create it and the next reconcile succeeds. No-ops when the namespace is system-managed.

 - cmd/operator-controller/main.go
Sets IsNamespaceManagementEnabled from the BoxcutterRuntime gate, and inserts ValidateInstallNamespace(coreClient) into the boxcutter step list between UnpackBundle and ApplyBundleWithBoxcutter.

 - Tests
Gate coverage in provider_test.go, the migration skip in boxcutter_test.go, and TestValidateInstallNamespace plus a direct test of the system-managed short-circuit in clusterextension_controller_test.go.

Only the namespace existence check is live today. Everything else waits on the CRD change in PR 3.

## Reviewer Checklist

- [ ] API Go Documentation

## Summary by CodeRabbit

* **New Features**
  * Added support for system-managed installation namespaces when the BoxcutterRuntime feature is enabled.
  * Install namespaces provided by users are now validated before deployment.
  * Missing user-provided namespaces report a retryable progress status.
  * Managed-namespace installations now skip unnecessary storage migration.


internal/operator-controller/applier/boxcutter_test.go
```diff
@@ -1141,7 +1141,7 @@ func TestBoxcutterStorageMigrator(t *testing.T) {
 		require.NoError(t, ocv1.AddToScheme(testScheme))
 
 		ext := &ocv1.ClusterExtension{
-			ObjectMeta: metav1.ObjectMeta{Name: "test123"},
+			ObjectMeta: metav1.ObjectMeta{Name: "test123"}, Spec: ocv1.ClusterExtensionSpec{Namespace: "test-namespace"},
```
tmshort, line 1144: Nit, put the new `Spec:` on a separate line for readability.

rashmigottipati (approved):

internal/operator-controller/controllers/clusterextension_reconcile_steps.go
```diff
@@ -402,6 +404,35 @@ func UnpackBundle(i imageutil.Puller, cache imageutil.Cache) ReconcileStepFunc {
...
+		if ext.Spec.Namespace == "" {
+			return nil, nil
+		}
+
+		l := log.FromContext(ctx)
+		l.V(1).Info("validating user-provided namespace exists", "namespace", ext.Spec.Namespace)
+
+		_, err := nsClient.Namespaces().Get(ctx, ext.Spec.Namespace, metav1.GetOptions{})
+		if apierrors.IsNotFound(err) {
+			nsErr := fmt.Errorf("namespace %q not found; spec.namespace must reference an existing namespace", ext.Spec.Namespace)
```
joelanford, line 425: Nit: This might imply that `spec.namespace` is required to be set to an existing namespace. Another option would be to unset it and let the system manage it.

Maybe:
```suggestion
			nsErr := fmt.Errorf("namespace %q not found; when set, spec.namespace must reference an existing namespace", ext.Spec.Namespace)
```

PR #2929, opened 2026-09-17
⚠ make spec.namespace optional in the experimental channel

nader-ziada:
# Description

Makes ClusterExtension.spec.namespace optional in the experimental API to support system-managed install namespaces.

When omitted, the controller uses the bundle’s system-managed namespace instead of requiring callers to provide one.

## Changes

  - Marks spec.namespace optional in the experimental API and generated CRDs.
  - Updates API references and namespace-management documentation.
  - Adds admission coverage for omitted namespaces.
  - Adds e2e coverage for installing an extension without spec.namespace.
  - Updates generated apply configurations and CRD tooling.

## Testing

  - Added/updated unit tests for API generation and admission validation.
  - Added namespace.feature e2e coverage for the omitted-namespace flow.

Depends on #2926 for the runtime handling of an omitted namespace.

## Reviewer Checklist

- [ ] API Go Documentation

## Summary by CodeRabbit

- **New Features**
  - `ClusterExtension.spec.namespace` is now optional. When omitted, a managed namespace can be created automatically from bundle metadata.
  - Managed namespaces support configurable Pod Security Admission labels.
  - Namespace selection remains fixed after creation and cannot be added, removed, or changed.

- **Documentation**
  - Added guidance for managed namespaces and namespace configuration.
  - Clarified namespace resolution, upgrade stability, deletion behavior, and administrator overrides.

- **Bug Fixes**
  - Corrected generated validation rules and improved handling of validation expressions containing quotes.


docs/draft/concepts/managed-namespaces.md
```diff
@@ -0,0 +1,53 @@
...
+
+1. `operatorframework.io/suggested-namespace-template`: the `metadata.name` field from the JSON template
+2. `operatorframework.io/suggested-namespace`: a plain string with the preferred name
+3. `<packageName>-system`: convention fallback
+
+## What belongs in a managed namespace
+
+- The operator's own workloads (deployments, services, configmaps)
+- The operator's RBAC resources (service accounts, roles, role bindings)
+- CRDs and webhooks installed by the operator
```
joelanford: CRDs are cluster-scoped, which aren't in the managed namespace. 

I'd put "webhooks" in the workload category, unless we're talking about `ValidatingWebhookConfiguration` and `MutatingWebhookConfiguration` (in which case, those are cluster-scoped as well)

docs/draft/concepts/managed-namespaces.md
```diff
@@ -0,0 +1,53 @@
...
+
+- User application workloads
+- Shared services used by multiple operators
+- Persistent data that should survive operator uninstallation
+
+## Deletion behavior
+
+Deleting a ClusterExtension with a managed namespace **deletes the entire namespace and everything in it.** If you have created resources in the managed namespace that are not part of the operator, they will be lost.
+
+If you need the namespace to persist beyond the operator's lifecycle, use `spec.namespace` to point at an existing namespace you manage yourself.
```
joelanford, line 74: We may also want to mention that it is possible for the bundle-derived namespace name to change. If that happens with a managed namespace, OLM will create the new namespace and install the new bundle version into it, and then delete the old namespace. So it is possible for the namespace to be deleted without the ClusterExtension being deleted. We should note that OLM considers this a breaking change, and would expect bundle authors to rev their major version number when changing the bundle-derived namespace.

docs/draft/howto/namespace-configuration-for-authors.md
```diff
@@ -0,0 +1,63 @@
...
+
+If neither annotation is present, operator-controller uses `<packageName>-system` as the namespace name.
+
+## Priority
+
+If both annotations are present, `suggested-namespace-template` takes priority.
+
+## Guidelines
+
+- Always include PSA labels if your operator runs privileged containers.
```
joelanford:
```suggestion
- Always include explicit PSA labels, even if your pods align with the `restricted` PSA profile.
```

docs/draft/howto/namespace-configuration-for-authors.md
```diff
@@ -0,0 +1,63 @@
...
+- Use a descriptive, unique namespace name that includes your package name to avoid collisions.
+- Do not assume the namespace name will be exactly what you suggest as cluster admins can override it by setting `spec.namespace`.
+- The namespace name from the template is used only when `spec.namespace` is omitted. When set, the admin's choice takes precedence and no namespace object is created.
+
+## Consistency across bundle formats
+
+The `operatorframework.io/suggested-namespace-template` and `operatorframework.io/suggested-namespace` annotations are the canonical way to declare namespace preferences. Future bundle formats should use the same annotation keys to avoid divergence across the ecosystem.
```
joelanford: I would drop this line. Other formats may already have fields for this, may specify namespaces directly as templated manifests, or may not even have the concept of a single "namespace".

I would actually call this a relic of registry+v1 and something we should avoid entirely when considering support for other formats.

internal/operator-controller/applier/boxcutter.go
```diff
@@ -273,6 +273,11 @@ type boxcutterStorageMigratorClient interface {
 // Migrate creates a ClusterObjectSet from an existing Helm release if no revisions exist yet.
 // The migration is idempotent and skipped if revisions already exist or no Helm release is found.
 func (m *BoxcutterStorageMigrator) Migrate(ctx context.Context, ext *ocv1.ClusterExtension, objectLabels map[string]string) error {
+	// Managed namespace mode (spec.namespace empty) means this is a new-style extension
+	// that never had a Helm release, so there's nothing to migrate.
+	if ext.Spec.Namespace == "" {
+		return nil
+	}
```
joelanford, line 283: The reason we can make this assumption is because:
1. Old CRD disallows empty namespace
2. New CRD disallows unsetting the namespace

This means that the only way to end up with a CE with an unset spec.namespace is for it to be created _after_ the new CRD is applied.

Is that right?

I do still wonder if we should have fewer (no?) opinions about the transitions we allow here though. I haven't thought deeply enough about this question yet, but I think we should.
- Should we allow a user who was forced to specify a managed namespace to transition their existing CE to a managed namespace?
- Should we allow a user who was using a managed namespace to instead opt to start managing the namespace?

But I think we can discuss that question separate from this PR, as that would just be a further relaxation of the CRD schema.

api/v1/clusterextension_types.go
```diff
@@ -59,12 +59,36 @@ type ClusterExtensionSpec struct {
...
+	//
+	// namespace is optional. When set, it must reference an existing namespace on the cluster.
+	// When omitted, operator-controller resolves and creates a managed namespace from the
+	// bundle's metadata. Whether namespace is set or omitted is fixed at creation time and
+	// cannot be changed afterwards.
+	//
+	// Keep the namespace resolved by bundle metadata stable across bundle upgrades. A change
+	// to that metadata can cause a later revision to use a different managed namespace. When
+	// the previous revision is archived, its namespace is deleted with all of its contents.
+	//
```
perdasilva: I wonder if we should include this here - this seems to be more author focused, maybe?

docs/draft/concepts/managed-namespaces.md
perdasilva: It may also be worth noting that this should not impact any workloads (outside of the namespace being deleted) since the CRD will not be deleted through this migration.

api/v1/clusterextension_types.go
nader-ziada: agreed, removed it from here

docs/draft/concepts/managed-namespaces.md
nader-ziada: changed it

docs/draft/concepts/managed-namespaces.md
nader-ziada: updated the comments

docs/draft/howto/namespace-configuration-for-authors.md
nader-ziada: applied change

docs/draft/howto/namespace-configuration-for-authors.md
nader-ziada: fixed

docs/draft/concepts/managed-namespaces.md
```diff
@@ -0,0 +1,79 @@
...
+
+Until managed-namespace relocation has defined safe behavior, publish bundle upgrades that
+resolve to the same namespace. To move an operator to another namespace, treat it as a
+separate migration rather than changing the bundle's namespace metadata in place.
+
+## What belongs in a managed namespace
+
+- The operator's own workloads (deployments, services, configmaps)
+- The operator's RBAC resources (service accounts, roles, role bindings)
+- CRDs and webhooks installed by the operator
```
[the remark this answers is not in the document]
nader-ziada: fixed

Makefile
```diff
@@ -685,6 +685,15 @@ crd-ref-docs: $(CRD_REF_DOCS) #EXHELP Generate the API Reference Documents.
 	$(CRD_REF_DOCS) --source-path=$(ROOT_DIR)/api/ \
 	--config=$(API_REFERENCE_DIR)/crd-ref-docs-gen-config.yaml \
 	--renderer=markdown --output-path=$(API_REFERENCE_DIR)/$(API_REFERENCE_FILENAME);
+	# crd-ref-docs renders doc-comment text verbatim, including internal <opcon:...> generator
+	# directives. The reference covers both channels at once, so label the channel-specific
+	# description blocks rather than dropping the tags silently -- otherwise a field documented
+	# per channel reads as self-contradictory. Remaining directives are stripped.
+	sed -E -e 's#<opcon:standard:description>#**Standard channel:** #g' \
+		-e 's#<opcon:experimental:description>#**Experimental channel:** #g' \
+		-e 's#</?opcon:[^>]*>##g' \
```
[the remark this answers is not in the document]
nader-ziada: fixed

test/e2e/features/namespace.feature
```diff
@@ -0,0 +1,62 @@
...
+      | key                                      | value      |
+      | pod-security.kubernetes.io/enforce        | privileged |
+      | pod-security.kubernetes.io/audit          | privileged |
+      | pod-security.kubernetes.io/warn           | privileged |
+
+  Scenario: User-provided namespace does not get PSA labels
+    Given namespace "${TEST_NAMESPACE}" is available
+    And a catalog "test" with packages:
+      | package | version | channel | replaces | contents                   |
+      | test    | 1.0.0   | stable  |          | CRD, Deployment, ConfigMap |
```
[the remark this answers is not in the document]
nader-ziada: fixed

internal/operator-controller/applier/boxcutter.go
nader-ziada: Yes, that assumption is correct for supported CRD transitions.

  - The old CRD required spec.namespace, and its DNS-label rule rejected "".
  - An existing extension therefore always has a non-empty namespace.
  - The new experimental CRD’s spec-level CEL rule prevents changing presence, so an existing non-empty field cannot be removed.
  - Therefore an object with omitted spec.namespace must have been created after the new CRD was installed; it cannot represent a legacy Helm-backed extension, so skipping
    Helm→ClusterObjectSet migration is valid.

tmshort (approved):

PR #2936, opened 2026-09-21
:seedling: OPRUN-4716: fix: adopt migrated ClusterObjectSets

tmshort:
## Summary

Adopts the revision 1 ClusterObjectSet created by the OLM v0-to-v1 migration tool when its subsequently created ClusterExtension resolves to the same bundle.

Without this change, Boxcutter recognizes the existing COS but compares it with a freshly rendered initial revision, finds a difference, and creates revision 2. That leaves two active revisions and fails the migration adoption prerequisite.

This change:

- identifies a successful revision 1 marked with the olm.operatorframework.io/migrated-from-subscription annotation;
- requires its package, bundle name, version, reference, and release annotations to match the resolved bundle;
- assigns the new ClusterExtension as controller owner of the existing COS; and
- returns the adopted COS as the initial installed revision without generating a duplicate.

A later bundle resolution with different metadata follows the normal upgrade path and creates a new revision.

## Validation

- go test ./internal/operator-controller/applier -count=1
- Rebuilt the controller image locally and ran the library-olm migration COS adoption test against Kind. The test creates and waits for a migration COS before creating the CE, verifies the original COS UID remains the only COS, and passed.

## Summary by CodeRabbit

* **New Features**
  * Existing migrated subscription revisions are now adopted as the initial installed state when they match the desired configuration.
  * Prevents duplicate revision creation during migration.
  * Migrated revisions are linked to the appropriate extension for continued management.
  * Adoption is safely rejected when a migrated revision is already managed by another extension, preventing conflicting ownership.
  * Matching migrated revisions can now be recognized automatically during installation and upgrade workflows.


tmshort:
Not needed

